PR #41119, opened 2025-09-24
Separate docker images for Nvidia and AMD in benchmarking

ahadnagy:
# What does this PR do?

This PR adds the option to run the benchmarks with different Docker images and options for each hardware platform.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/.github/workflows/benchmark_v2_a10_caller.yml
+++ b/.github/workflows/benchmark_v2_a10_caller.yml
@@ -13,6 +13,8 @@ jobs:
     uses: ./.github/workflows/benchmark_v2.yml
     with:
       runner: aws-g5-4xlarge-cache-use1-public-80
+      container_image: huggingface/transformers-pytorch-gpu
+      container_options: --gpus all --privileged --ipc host --shm-size "16gb"
       commit_sha: ${{ github.sha }}
       run_id: ${{ github.run_id }}
       benchmark_repo_id: hf-internal-testing/transformers-daily-benchmarks
--- a/.github/workflows/benchmark_v2.yml
+++ b/.github/workflows/benchmark_v2.yml
@@ -7,6 +7,14 @@ on:
         description: 'GH Actions runner group to use'
         required: true
         type: string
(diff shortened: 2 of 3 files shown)


ahadnagy:
cc @ydshieh

ydshieh (approved):
For sure, but before merge, I have a question?

For workflow running on AMD runners, should we have it on the repository

https://github.com/huggingface/hf-workflows

(just like the AMD CI workflow files, for the security reason.).

You don't need to address this in this PR, but it's an important question (to be addressed in a separate PR if it makes sense)

.github/workflows/benchmark_v2_a10_caller.yml
```diff
@@ -13,6 +13,8 @@ jobs:
     uses: ./.github/workflows/benchmark_v2.yml
     with:
       runner: aws-g5-4xlarge-cache-use1-public-80
```
ydshieh, line 15: and one minor question, why here it's `runner: aws-g5-4xlarge-cache-use1-public-80` instead of

```
    runs-on:
      group: aws-g5-4xlarge-cache
```

(that is what is used so far in many workflow files ..?)

ahadnagy: This was a workaround until we got the AMD runner reconfigured as well to be a proper runner group. I'll clean this up in a separate PR.

ahadnagy:
> For workflow running on AMD runners, should we have it on the repository

We can certainly create a wrapper in the hf-workflows repo. Or do we need to create a copy of the whole workflow?

ydshieh:
> We can certainly create a wrapper in the hf-workflows repo. Or do we need to create a copy of the whole workflow?

The jobs that will actually use the AMD runners should be on `hf-workflows`. The caller to them could be in `transformers`. But you can talk to @ivarflakstad for the details, he knows better than me regarding this (and I am not 100% sure if I am  right about the necessity of doing so)

PR #41121, opened 2025-09-24
fix: resolve the unexpected video frame drop issue of the InternVL model with multiple video inputs

WesKwong:
# What does this PR do?

This PR addresses a bug where the InternVL preprocessor would incorrectly drop one video frame from subsequent videos in a multi-video input.

This issue was reported in: https://github.com/OpenGVLab/InternVL/issues/1178


Rocketknight1:
cc @zucchini-nlp

PR #41123, opened 2025-09-24
[v5] Remove deprecated prediction loop

SunMarc:
# What does this PR do?

This PR removes the legacy prediction loop that was deprecated 4 years ago. I think that it should be fine to deprecate it even without any warning for v5 which it would have been better to have. I checked files on github and no one really seemed to use the legacy one. Also `SequentialDistributedSampler` was bound to be deprecated in v5. 

If we still prefer to have it for one version, happy to submit a PR for a deprecation msg.


qgallouedec (approved):
lgtm!

Cyrilvallez (approved):
LGTM! 🧹

PR #41124, opened 2025-09-24
Revert "Fix DeepSpeed mixed precision precedence over Accelerate defaults"

SunMarc:
Reverts huggingface/transformers#39856

Not the intended fix. This breaks the logic in deepspeed and makes things complicated to maintain.


Cyrilvallez (approved):
Don't have the background on the original issue, but this indeed looks much simpler! Feel free to merge if you're confident, otherwise we can wait for @S1ro1!

PR #41127, opened 2025-09-24
[v5] Remove train kwargs

SunMarc:
# What does this PR do?

This PR removes train kwargs. There were used for deprecated arguments but since we are removing them, we can also safely remove the kwargs.


Cyrilvallez (approved):
Nice!

PR #41128, opened 2025-09-24
[v5] Remove tokenizer from Trainer

SunMarc:
# What does this PR do?

This PR removes tokenizers from Trainer as it will be deprecated for v5. Since we removed it from the init, I think it should also make sense to remove the setter and the property method for tokenizer


qgallouedec (approved):
lgtm!

PR #41134, opened 2025-09-24
Deprecate `half_precision_backend`

SunMarc:
# What does this PR do?

This PR deprecates `half_precision_backend` arg in Trainer. We don't really need cpu and apex backend.
For apex, they even removed the amp folder from their project since they've upstreamed everything to torch a long time ago.
Also the installation process for apex seems way too complicated, so i'm pretty sure there is no real usage. 
For cpu, we were using `torch.autocast` but we don't really need that as this is supported in accelerate. By the way, the fp16 case with cpu was already handled by accelerate.
In the end, this shouldn't impact users at all. I left a deprecation message but I don't think this is really needed.


cyyever:
apex can be dropped IMO..

src/transformers/training_args.py
```diff
@@ -1029,20 +1021,10 @@ class TrainingArguments:
...
-            "help": (
-                "For fp16: Apex AMP optimization level selected in ['O0', 'O1', 'O2', and 'O3']. "
-                "See details at https://nvidia.github.io/apex/amp.html"
-            )
-        },
-    )
-    half_precision_backend: str = field(
-        default="auto",
+    half_precision_backend: Optional[str] = field(
+        default=None,
```
SunMarc, line 1023: I left that here but I think we can potentially just remove it

qgallouedec (approved):
no problem on my side

cyyever (approved):
We can remove the `if not isinstance(autocast_ctx, contextlib.nullcontext)` branch in `compute_loss_context_manager`

SunMarc:
> We can remove the if not isinstance(autocast_ctx, contextlib.nullcontext) branch in compute_loss_context_manager

It's more for backwards compatibility in case someone overwrote `autocast_smart_context_manager` in their codebase

SunMarc:
cc @winglian for visibility

src/transformers/training_args.py
SunMarc: maybe we can keep it for the last release of v4 then remove in v5

Cyrilvallez: No strong opinions either way on whether you want to keep it for last v4 version or not - but yeah let's clean it for v5 at least! 🤗

Cyrilvallez (approved):
Nice, LGTM!

PR #41137, opened 2025-09-24
:rotating_light: [`v5`] Remove legacy cache

vasqu:
As per title, however this really is only a draft for now. Only checked for basic removal

Legacy cache still exists but only for a few special cases atm (whisper, rag, ...)


vasqu:
Yea looks like this needs a good amount of more effort to check for specific more exotic models :eyes: leaving it as is for now

vasqu:
Was too ambitious :eyes: might reopen this when I have time

PR #41138, opened 2025-09-24
Make quantizers good citizens loading-wise

Cyrilvallez:
# What does this PR do?

The quantizers should NEVER need to access the `state_dict` during `param_needs_quantization` or `create_quantized_param`. This is mostly for the following reasons:

- Most models have a lot of different state dict (due to restricting the size of each file to 5GB), so relying on other params of the state dict for quantizing a given param is a very bad idea -> if the needed params happen to live on different saved state dicts (representing the same model), then the whole loading logic collapses without any possibility of fallback
- Without quantizers, we load parameters one by one without ever materializing the whole state dict at once, and it's quite annoying to have to change that logic only for a few quantizers, as it's a good thing in general

If the quantizers need several serialized keys to retrieve a given quantized param, then they should cache them individually, and then quantize back once all the needed keys are retrieved. Note that this DOES NOT incur a memory cost, as previously the whole state dict was loaded on memory, so the memory footprint was always higher before (-> a param will necessarily get quantized back, and its serialized tensors cleaned, before the whole state dict is materialized, assuming they all reside on the same state_dict as was done previously). So we actually have a gain of memory during loading with this PR.

Note that the current PR has less slow tests failing than main, especially for bitsandbytes, i.e. it actually solves a few issues in addition to simplifying the whole loading logic of the library.

A following PR will come to unify the `update_expected_keys`, `update_missing_keys`, and `update_unexpected_keys` methods, as some are currently redundant/not practical regarding the whole loading logic.

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -120,7 +120,7 @@ from .utils.quantization_config import BitsAndBytesConfig, QuantizationMethod
 
 
 if is_torchao_available():
-    from torchao.quantization import Int4WeightOnlyConfig
+    pass
 
 if is_accelerate_available():
     from accelerate import dispatch_model, infer_auto_device_map
@@ -699,11 +699,8 @@ def _load_state_dict_into_meta_model(
         device_map_regex = "|".join([re.escape(k) for k in sorted(device_map.keys(), reverse=True)])
 
     is_quantized = hf_quantizer is not None
-    is_ao = is_quantized and hf_quantizer.quantization_config.quant_method in {
-        QuantizationMethod.TORCHAO,
-    }
     is_safetensors = shard_file.endswith(".safetensors")
-    is_meta_state_dict = is_safetensors and not is_ao
+    is_meta_state_dict = is_safetensors
     file_pointer = safe_open(shard_file, framework="pt", device=tensor_device) if is_meta_state_dict else None
     params_to_load = list(state_dict.keys())
 
@@ -737,8 +734,7 @@ def _load_state_dict_into_meta_model(
                     device_mesh.get_local_rank(),
                     device_mesh,
                 )
-            else:
-                # we have a device mesh but the param needs to be quantized, so we shard inside create_quantized_param
+            else:  # we have a device mesh but the param needs to be quantized, so we shard inside create_quantized_param:
                 sharding_kwargs = {
                     "empty_param": empty_param,
                     "casting_dtype": casting_dtype,
@@ -751,6 +747,7 @@ def _load_state_dict_into_meta_model(
                     param,
                     param_name,
                     device_mesh.get_local_rank(),
+                    state_dict,
                     **sharding_kwargs,
                 )
         else:
@@ -779,7 +776,8 @@ def _load_state_dict_into_meta_model(
                 _load_parameter_into_model(model, param_name, param.to(param_device))
 
             else:
-                hf_quantizer.create_quantized_param(model, param, param_name, param_device)
+                # TODO naming is stupid it loads it as well
+                hf_quantizer.create_quantized_param(model, param, param_name, param_device, state_dict)
 
                 # For quantized modules with FSDP/DeepSpeed Stage 3, we need to quantize the parameter on the GPU
                 # and then cast it to CPU to avoid excessive memory usage on each GPU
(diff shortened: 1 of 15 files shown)


src/transformers/modeling_utils.py
```diff
@@ -4400,9 +4367,6 @@ def from_pretrained(
             force_download (`bool`, *optional*, defaults to `False`):
                 Whether or not to force the (re-)download of the model weights and configuration files, overriding the
                 cached versions if they exist.
-            resume_download:
-                Deprecated and ignored. All downloads are now resumed by default when possible.
-                Will be removed in v5 of Transformers.
```
SunMarc, line 4405: maybe remove this in another PR ?

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -113,6 +106,20 @@ class TorchAoHfQuantizer(HfQuantizer):
...
+            config_name = self.quantization_config.quant_type.__class__.__name__
+            is_int_4 = fuzzy_match_size(config_name) == "4"
+
+        # TODO: better way to get the serialized key names? Hard to read from torchao codebase
+        if is_int_4:
+            self.weight_ao_keys = ["qdata", "scale", "zero_point"]
+        else:
+            self.weight_ao_keys = ["qdata", "scale"]
+        # Instead of serializing the simple torch.Tensor like usual, torchao adds a `:_data` suffix so we need this
+        self.full_ao_keys = self.weight_ao_keys + ["_data"]
```
SunMarc, line 122: cc @jerryzh168 is there a better way to get serialized key names with safetensors ? We need those to cache the values before unflattening them as we don't store anymore the entire state dict.

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -313,16 +338,6 @@ def create_quantized_param(
 
             quantize_(module, self.quantization_config.get_apply_tensor_subclass())
 
-    def update_state_dict_with_metadata(self, state_dict, metadata):
-        """
-        If the metadata contains torchao tensor subclass information, we reconstruct the tensor subclass state dict
-        from the provided state_dict and metadata.
-        """
-        if TORCHAO_VERSION >= version.parse("0.14.0") and is_metadata_torchao(metadata):
```
SunMarc, line 321: can we keep the version check and put that in `create_quantized_param` + raise a an error ? Otherwise, the user might get a error about unflatten_tensor_state_dict missing from the lib

SunMarc (approved):
Thanks for this work ! This is really nice. Left some minor comments

SunMarc:
run-slow: bnb

src/transformers/modeling_utils.py
Cyrilvallez: It's just the docstring (the arg was removed long ago)

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -229,61 +237,82 @@ def _process_model_before_weight_loading(
...
 
-        param_device = kwargs.pop("param_device", None)
         # check if the param_name is not in self.modules_to_not_convert
-        if any((key + "." in param_name) or (key == param_name) for key in self.modules_to_not_convert):
-            return False
-        elif param_device == "cpu" and self.offload:
-            # We don't quantize weights that we offload
+        if any(key + "." in param_name or key == param_name for key in self.modules_to_not_convert):
             return False
+        elif any(param_name.endswith(f":{x}") for x in self.full_ao_keys):
```
jerryzh168, line 250: if full_ao_keys is only used here, we can update this to `":" in param_name` I think

src/transformers/quantizers/quantizer_torchao.py
jerryzh168: we'll have different key names depending on different dtype (int4, int8, float8) and packing format (plain, preshuffled etc.) a fixed list won't work for everything I think. is this always required to load the params?

SunMarc: It will indeed be required in order to load safetensors file. If you check the PR, we need to know when all the tensors are cached in order to unflatten them. Also this is for dealing with missing / unexpected keys.

Cyrilvallez: Indeed, exactly as @SunMarc said! It should be fairly easy to retrieve which keys are needed from the config though, so create the list at __init__ time!

PR #41142, opened 2025-09-25
Fix typing of train_args

cyyever:
# What does this PR do?

Some `Optional` typing is quite confusing because their default values is not `None`.


Rocketknight1 (approved):
Yep, these all make sense. Thank you!

Rocketknight1 (approved):
Yes, LGTM. I think this is all just misuse of `Optional`!

cyyever:
@Rocketknight1 I have fixed `fsdp` typing. Otherwise HfArgumentParser could generate `None` as its default value.

PR #41143, opened 2025-09-25
Adapt to the SDPA interface to enable the NPU to call FlashAttentionScore

frozenleaves:
# What does this PR do?

Enable the torhc native SDPA interface  to invoke the FlashAttentionScore operator of NPU.

----

### Restrictions on the use of torch native interface calling FA operator: 
1. When is_causal calculation is enabled, attn_mask must be None; when is_causal is not enabled, if attn_mask enters valid data, the input data type must be Bool type.
2. When entering the data types of query, key, value, and enables requirements_grad, executes the FA operator.

### DIFF
--- a/src/transformers/integrations/sdpa_attention.py
+++ b/src/transformers/integrations/sdpa_attention.py
@@ -2,7 +2,7 @@ from typing import Optional
 
 import torch
 
-from ..utils import is_torch_xpu_available, logging
+from ..utils import is_torch_npu_available, is_torch_xpu_available, logging
 from ..utils.import_utils import is_torch_greater_or_equal
 
 
@@ -12,6 +12,7 @@ logger = logging.get_logger(__name__)
 _is_torch_greater_or_equal_than_2_5 = is_torch_greater_or_equal("2.5", accept_dev=True)
 _is_torch_greater_or_equal_than_2_8 = is_torch_greater_or_equal("2.8", accept_dev=True)
 _is_torch_xpu_available = is_torch_xpu_available()
+_is_torch_npu_available = is_torch_npu_available()
 
 
 def repeat_kv(hidden_states: torch.Tensor, n_rep: int) -> torch.Tensor:
@@ -35,8 +36,12 @@ def use_gqa_in_sdpa(attention_mask: Optional[torch.Tensor], key: torch.Tensor) -
     # 2.xpu
     #   - torch version >= 2.8
     #   - key is not a torch.fx.Proxy (otherwise it will fail with a tracing error)
+    # 3.npu
+    #   - npu is not supported gqa currently
     if _is_torch_xpu_available:
         return _is_torch_greater_or_equal_than_2_8 and not isinstance(key, torch.fx.Proxy)
+    if _is_torch_npu_available:
+        return False
     return _is_torch_greater_or_equal_than_2_5 and attention_mask is None and not isinstance(key, torch.fx.Proxy)
 
 
@@ -79,6 +84,14 @@ def sdpa_attention_forward(
     # We convert it to a bool for the SDPA kernel that only accepts bools.
     if torch.jit.is_tracing() and isinstance(is_causal, torch.Tensor):
         is_causal = is_causal.item()
+
+    # When `is_causal = False` and the `attention_mask` is not of boolean type, the Ascend NPU's SDPA interface cannot utilize the FlashAttentionScore operator，
+    # and falls back to small-operator concatenation. To invoke the FlashAttentionScore, the attention_mask must be converted to boolean type.
+    # This adaptation ensures the `attention_mask` meets the requirement for using FlashAttentionScore.
+    if _is_torch_npu_available:
+        if attention_mask is not None and attention_mask.dtype != torch.bool:
+            # Convert to boolean type, making sdpa to force call FlashAttentionScore to improve performance.
+            attention_mask = torch.logical_not(attention_mask.bool()).to(query.device)
 
     attn_output = torch.nn.functional.scaled_dot_product_attention(
         query,


src/transformers/integrations/sdpa_attention.py
```diff
@@ -80,6 +81,17 @@ def sdpa_attention_forward(
     if torch.jit.is_tracing() and isinstance(is_causal, torch.Tensor):
         is_causal = is_causal.item()
 
+    # When is_causal=False and the dtype of attention_mask is not bool, the sdpa interface of Ascend NPU cannot invoke
+    # the FlashAttentionScore operator but instead uses internal small operator concatenation. To enter FlashAttentionScore,
+    # the following conditions must be met:enable is_causal and set attention_mask to None; or disable is_causal and
+    # set attention_mask to a boolean type.So we adapt the parameters to allow it entry the FlashAttentionScore.
+    if _is_torch_npu_available:
+        if is_causal:
+            attention_mask = None
```
vasqu: We already have proper checks for `is_causal` over here https://github.com/huggingface/transformers/blob/44682e71314f004c8cf157f76e6956e58f6f730f/src/transformers/integrations/sdpa_attention.py#L70-L76

The flag itself really is barely usable tbh but in rare cases it can happen

src/transformers/integrations/sdpa_attention.py
```diff
@@ -80,6 +81,17 @@ def sdpa_attention_forward(
...
+        else:
+            if attention_mask.dtype != torch.bool:
+                attention_mask = torch.logical_not(attention_mask.bool()).to(query.device)
```
vasqu: Imo, this seems like the only real change needed here, i.e. `bool` is a requirement instead of float (for some older models).

tests/test_executorch.py
```diff
@@ -112,18 +112,21 @@ def test_decoder_only_lm_export(self):
...
-        torch.testing.assert_close(eager_output_ids, exported_output_ids, atol=1e-4, rtol=1e-4)
-
-        # Test export with inputs_embeds
-        exported_program_embeds = module.export(inputs_embeds=self.inputs_embeds, cache_position=self.cache_position)
-        eager_output_embeds = self.model(inputs_embeds=self.inputs_embeds, use_cache=False).logits
-        exported_output_embeds = exported_program_embeds.module()(
-            inputs_embeds=self.inputs_embeds, cache_position=self.cache_position
-        )
-        torch.testing.assert_close(eager_output_embeds, exported_output_embeds, atol=1e-4, rtol=1e-4)
+        with torch.no_grad():
```
vasqu: Why was this change needed?

src/transformers/utils/import_utils.py
```diff
@@ -1518,6 +1519,10 @@ def is_jinja_available() -> Union[tuple[bool, str], bool]:
     return _jinja_available
 
 
+def is_jieba_available():
+    return _jieba_available
```
vasqu: Shouldnt be needed

vasqu (commented):
Some questions since imo we only need to guarantee bool when npu is used, the rest should already be handled

Can you also check if there is a version requirement (it's only available from version xxx)? And last comment: We enable the `use_gqa` kwarg if the conditions are met https://github.com/huggingface/transformers/blob/44682e71314f004c8cf157f76e6956e58f6f730f/src/transformers/integrations/sdpa_attention.py#L29-L40
Is this supported on npu or should it default to False in this case.

cc @FightingZhen if you also have time to look. You're the only one I know who works with NPU

tests/test_executorch.py
frozenleaves: this is to fix the CI faild error: 
```
self = <tests.test_executorch.ExecutorchTest testMethod=test_decoder_only_lm_export>
mock_import_utils_npu = <MagicMock name='is_torch_npu_available' id='281458784650192'>

    @patch('transformers.utils.import_utils.is_torch_npu_available')
    def test_decoder_only_lm_export(self, mock_import_utils_npu):
        """Test TorchExportableModuleForDecoderOnlyLM export with both input types"""
        mock_import_utils_npu.return_value = False
        module = TorchExportableModuleForDecoderOnlyLM(self.model)

        # Test export with input_ids
>       exported_program_ids = module.export(input_ids=self.input_ids, cache_position=self.cache_position)

test_executorch.py:119:

......


>                   raise RuntimeError(
                        f"""\
    Found a graph input that requires gradients, and received a mutation.
    This is currently banned in the aot_export workflow. If you need this functionality, please file a github issue.

    fw_metadata={str(fw_metadata)}"""
                    )
E                   RuntimeError: Found a graph input that requires gradients, and received a mutation.

```

note:  this is not the original CI  Traceback ,  just show the error message.

src/transformers/utils/import_utils.py
frozenleaves: ok, i'll remove it.

src/transformers/integrations/sdpa_attention.py
frozenleaves: yes, just here is the real  necessary change。

tests/test_executorch.py
vasqu: Ah ic, this seems related to NPU accelerators only then. For now, we do not support proper checks on our CI for NPU so let's revert it for now / skip it with a proper comment.

frozenleaves: uhmm,  using the` @require_torch_multi_gpu` to decorate this test case, what do you think?

vasqu: Nope, shouldnt require multi-gpu tbh. Let's either ignore it for now (unrelated to the PR) or skip it, e.g. `if is_torch_npu_available: self.skip(reason="NPU not supported yet")`

Or does it introduce failures on cuda accelerators? I can check as well then, let's focus on the sdpa part for now

src/transformers/integrations/sdpa_attention.py
```diff
@@ -80,6 +81,17 @@ def sdpa_attention_forward(
     if torch.jit.is_tracing() and isinstance(is_causal, torch.Tensor):
         is_causal = is_causal.item()
 
+    # When is_causal=False and the dtype of attention_mask is not bool, the sdpa interface of Ascend NPU cannot invoke
```
FightingZhen: Two questions:
1. For Ascend NPU, `torch.nn.functional.scaled_dot_product_attention` has two downstream ways (as you mentioned 1) going FlashAttentionScore operator or 2) series of small operation concatenation), it should be automatically handled by `torch_npu`, **what is the reason for trying to force using FlashAttentionScore?**
2. In below code, setting `attention_mask` param from a pre-calculated value to `None` **by force is very dangerous, which is highly not recommanded**, for Ascend NPU, even if you put `attention_mask` param without `None,` the `scaled_dot_product_attention` api in `torch_npu` should handle this situation automatically.

tests/test_executorch.py
FightingZhen: CI in transformers not include NPU backend, it is not recommanded to modify existing GPU side CI script

frozenleaves: ok, I have already modify the pr, it just include the necessary part about sdpa

src/transformers/integrations/sdpa_attention.py
```diff
@@ -80,6 +81,14 @@ def sdpa_attention_forward(
...
+    # the FlashAttentionScore operator but instead uses internal small operator concatenation. To enter FlashAttentionScore,
+    # the following conditions must be met:enable is_causal and set attention_mask to None; or disable is_causal and
+    # set attention_mask to a boolean type.So we adapt the parameters to allow it entry the FlashAttentionScore.
+    if _is_torch_npu_available:
+        if attention_mask is not None and attention_mask.dtype != torch.bool:
```
FightingZhen, line 89: The implementation is different from https://www.hiascend.com/document/detail/zh/Pytorch/710/ptmoddevg/trainingmigrguide/performance_tuning_0034.html, have a check!

src/transformers/integrations/sdpa_attention.py
vasqu: I think the second question should be handled now? (as we don't set the mask to `None` anymore, only change the dtype if necessary). I think we need to update the comment here, seems like it still contains the prev logic about `is_causal`

Still curious about 1 cc @frozenleaves

vasqu (commented):
Thanks for iterating
- Can you look at https://github.com/huggingface/transformers/pull/41143#pullrequestreview-3267237606 --> just want to make sure that `use_gqa` is also supported on npu and if there are any restrictions
- And a last comment on the questions from @FightingZhen --> I think one question has been handled, but the other should be addressed imo

src/transformers/integrations/sdpa_attention.py
frozenleaves: sure, I just update the comment, please review it again.

FightingZhen:
> Thanks for iterating
> 
> * Can you look at [#41143 (review)](https://github.com/huggingface/transformers/pull/41143#pullrequestreview-3267237606) --> just want to make sure that `use_gqa` is also supported on npu and if there are any restrictions
> * And a last comment on the questions from @FightingZhen --> I think one question has been handled, but the other should be addressed imo

@vasqu I've received question 1. It's currently under internal review with the team, and I will provide an update in the PR once we have a conclusion.

frozenleaves:
**compare the loss between original sdpa and this pr:**
internlmVL3-1B-hf:
image

qwen3-1.7B:
image


**compare the loss between original eager and this pr:**
internlmVL3-1B-hf:
image

qwen3-1.7B:
image


This is part of the test data, showing the comparison of loss curves before and after adaptation， include eager and sdpa mode.

src/transformers/integrations/sdpa_attention.py
frozenleaves: > Two questions:
> 
> 1. For Ascend NPU, `torch.nn.functional.scaled_dot_product_attention` has two downstream ways (as you mentioned 1) going FlashAttentionScore operator or 2) series of small operation concatenation), it should be automatically handled by `torch_npu`, **what is the reason for trying to force using FlashAttentionScore?**
> 2. In below code, setting `attention_mask` param from a pre-calculated value to `None` **by force is very dangerous, which is highly not recommanded**, for Ascend NPU, even if you put `attention_mask` param without `None,` the `scaled_dot_product_attention` api in `torch_npu` should handle this situation automatically.

For the first question,  the aim of this pr is to improve the performance for commonly used model， such as Qwen, Llama, et.  Just as you are concerned, in some scenarios on the NPU, the small operator concatenation might perform better than FlashAttentionScore. However, for the majority of current scenarios, FlashAttentionScore should be the better choice. I think this trade-off is worth leaning towards FlashAttentionScore.

FightingZhen:
> > Thanks for iterating
> > 
> > * Can you look at [#41143 (review)](https://github.com/huggingface/transformers/pull/41143#pullrequestreview-3267237606) --> just want to make sure that `use_gqa` is also supported on npu and if there are any restrictions
> > * And a last comment on the questions from @FightingZhen --> I think one question has been handled, but the other should be addressed imo
> 
> @vasqu I've received question 1. It's currently under internal review with the team, and I will provide an update in the PR once we have a conclusion.

@vasqu `enable_gqa` is not available for Ascend NPU for now, I will try my best to push inner team make it available. Maybe we need to add some code to make func `use_gqa_in_sdpa` return `False` for Ascend NPU @frozenleaves

src/transformers/integrations/sdpa_attention.py
vasqu: Thanks a lot, then let's keep it this way. The comment is also very detailed so that other contributors can see why we do this.

vasqu (approved):
LGTM, then let's add the proper `return False` within https://github.com/huggingface/transformers/blob/cdba28c3443bc95b8631d73e33579e47f754c198/src/transformers/integrations/sdpa_attention.py#L29-L40
then we are good, i.e. add comment and make a conditional check at the beginning similar to xpu. No worries about `gqa` not being available yet @FightingZhen @frozenleaves, we can always change it in the future when it becomes a feature in npu as well

frozenleaves:
> > > Thanks for iterating
> > > 
> > > * Can you look at [#41143 (review)](https://github.com/huggingface/transformers/pull/41143#pullrequestreview-3267237606) --> just want to make sure that `use_gqa` is also supported on npu and if there are any restrictions
> > > * And a last comment on the questions from @FightingZhen --> I think one question has been handled, but the other should be addressed imo
> > 
> > 
> > @vasqu I've received question 1. It's currently under internal review with the team, and I will provide an update in the PR once we have a conclusion.
> 
> @vasqu `enable_gqa` is not available for Ascend NPU for now, I will try my best to push inner team make it available. Maybe we need to add some code to make func `use_gqa_in_sdpa` return `False` for Ascend NPU @frozenleaves

No problem, I've updated the code to make sure that in npu environment `use_gqa_in_sdpa` returns False. Please review this PR again, thanks. @vasqu @FightingZhen

vasqu:
Looks good, would wait for @FightingZhen if he has any additional comments then merge :hugs:

FightingZhen:
> Looks good, would wait for @FightingZhen if he has any additional comments then merge 🤗

@vasqu Codes and comments all LGTM :)

vasqu:
Thanks a lot both of you, merging now (docs building)

PR #41147, opened 2025-09-25
add rotary kernel support to Qwen3 model

kaixuanliu:
Adds Rotary kernels from https://huggingface.co/kernels-community/rotary to Qwen3 series models

Here are Some benchmarks comparing perfs between rotary kernels and `apply_rotary_pos_emb` func in transformers:
For A100,
rotary_a100_combined_comparison
And for Intel XPU:
rotary_xpu_combined_comparison


kaixuanliu:
I made benchmark for `Qwen/Qwen3-4B-Instruct-2507` model, and on Intel XPU, it will get ~10% performance improvement for E2E time. While on A100, there is no obvious performance improvement or drop. Pls let me know if it is OK using this manner to apply rotary kernel, and then I will add the support for more models.

Rocketknight1:
cc @arthurzucker

src/transformers/models/qwen3/modeling_qwen3.py
```diff
@@ -477,6 +513,10 @@ def forward(
         >>> tokenizer.batch_decode(generate_ids, skip_special_tokens=True, clean_up_tokenization_spaces=False)[0]
         "Hey, are you conscious? Can you talk to me?\nI'm not conscious, but I can talk to you."
         ```"""
+        # Set global use_kernels flag based on model's kernel usage
+        global use_kernels
+        use_kernels = getattr(self, "use_kernels", False)
+
```
MekkCyber: It's better to have an attention kwarg passed `use_rotary_kernel` for example than defining a global variable like this

src/transformers/models/qwen3/modeling_qwen3.py
```diff
@@ -28,6 +28,7 @@
 from ...cache_utils import Cache, DynamicCache
 from ...generation import GenerationMixin
 from ...integrations import use_kernel_forward_from_hub
+from ...integrations.hub_kernels import rotary_kernel
```
MekkCyber: I think we need to lazily load the kernel, because here we are loading it before even knowing if the user wants to use kernels or not

src/transformers/models/qwen3/modeling_qwen3.py
```diff
@@ -117,6 +122,34 @@ def apply_rotary_pos_emb(q, k, cos, sin, position_ids=None, unsqueeze_dim=1):
...
+    q2 = q_rotated[..., half_dim:]
+    k1 = k_rotated[..., :half_dim]
+    k2 = k_rotated[..., half_dim:]
+    if cos.shape[-1] != half_dim:
+        # Trim cos/sin to match half_dim
+        cos = cos[..., :half_dim]
+        sin = sin[..., :half_dim]
+
+    # Apply rotary embedding using our kernel
+    rotary_kernel.apply_rotary(q1, q2, cos, sin, q1, q2, False)
```
MekkCyber: Did you try to benchmark the performance with and without this kernel ?

MekkCyber (commented):
Thanks for this integration @kaixuanliu ! I left few nits to consider

src/transformers/models/qwen3/modeling_qwen3.py
kaixuanliu: Yes, on Intel XPU, one single rotary op needs 0.22 ms, and it drops to 0.1 ms after applying this patch. above 2x speedup.

src/transformers/models/qwen3/modeling_qwen3.py
kaixuanliu: Thx for your advice! Have updated related code

src/transformers/models/qwen3/modeling_qwen3.py
kaixuanliu: You mean add a param called `use_rotary_kernel` to [kwargs](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3/modeling_qwen3.py#L488) here, and passed it down to `Qwen3Attention`?

src/transformers/models/qwen3/modeling_qwen3.py
IlyasMoutawwakil: @kaixuanliu cool but how much faster was the inference with a real model ?

ArthurZucker (commented):
hey! unfortunately this is not how we want to be adding support for kernels in general! 
There should be 0 modeling changes involved, especially here it does not even seem to be required! 

We'd rather import once from kernels to replace the rotary embed if the function is defined or something, but in the broad scheme of things, we want a mapping for function ! Like we do for classes

kaixuanliu:
@ArthurZucker Thx for the comment, it makes sense.Since for function level, we do not have a shema for mapping like classes in kernels, we will add related support and then based on this, I will adjust this PR.

yao-matrix:
@ArthurZucker  @danieldk , could you comment the feasibility of Kaixuan's proposal?

ArthurZucker:
I think @MekkCyber is working on that feature specifically!

kaixuanliu:
> I think @MekkCyber is working on that feature specifically!

@ArthurZucker @MekkCyber , do you mean this PR: https://github.com/huggingface/transformers/pull/41577 ?

MekkCyber:
Hey @kaixuanliu, yes we will start using the hub mapping in the PR you linked, but the kernel needs to be a drop in replacement for the function in the modeling so we don't have to change the modeling files apart from lazily loading the kernel, in case you need a special function for example in the case of `rotary`, we can expose it directly in the kernel

kaixuanliu:
@MekkCyber @ArthurZucker I have adjusted the code using the function mapping in PR 41577, and put the rotary adpat function in https://github.com/huggingface/kernels-community/pull/53, Can you help review again? Thx!

yao-matrix:
@MekkCyber, does the refactored solution meet your expectation? Thx very much.

src/transformers/models/qwen3/modular_qwen3.py
```diff
@@ -139,7 +153,33 @@ def forward(
...
+        hidden_states = outputs.last_hidden_state
+        # Only compute necessary logits, and do not upcast them to float if we are not computing the loss
+        slice_indices = slice(-logits_to_keep, None) if isinstance(logits_to_keep, int) else logits_to_keep
+        logits = self.lm_head(hidden_states[:, slice_indices, :])
+
+        loss = None
+        if labels is not None:
+            loss = self.loss_function(logits=logits, labels=labels, vocab_size=self.config.vocab_size, **kwargs)
+
+        return CausalLMOutputWithPast(
```
MekkCyber: could you revert the unrelated changes please ?

src/transformers/models/qwen3_moe/modeling_qwen3_moe.py
```diff
@@ -148,6 +148,11 @@ def __init__(self, config: Qwen3MoeConfig, layer_idx: int):
         self.k_norm = Qwen3MoeRMSNorm(self.head_dim, eps=config.rms_norm_eps)  # thus post q_norm does not need reshape
         self.sliding_window = getattr(config, "sliding_window", None)
 
+        from ...integrations.hub_kernels import lazy_load_kernel
```
MekkCyber: let's have the import at the start of the file with other imports

src/transformers/models/qwen3_next/modular_qwen3_next.py
```diff
@@ -221,11 +221,34 @@ class Qwen3NextRMSNorm(Gemma3RMSNorm):
...
+            config.hidden_size, config.num_key_value_heads * self.head_dim, bias=config.attention_bias
+        )
+        self.v_proj = nn.Linear(
+            config.hidden_size, config.num_key_value_heads * self.head_dim, bias=config.attention_bias
+        )
+        self.o_proj = nn.Linear(
+            config.num_attention_heads * self.head_dim, config.hidden_size, bias=config.attention_bias
+        )
+        self.q_norm = Qwen3NextRMSNorm(self.head_dim, eps=config.rms_norm_eps)
+        self.k_norm = Qwen3NextRMSNorm(self.head_dim, eps=config.rms_norm_eps)
```
MekkCyber: seems unrelated too, we should only load the kernel and use it with very minimal changes to the modeling files

MekkCyber (commented):
Thanks @kaixuanliu @yao-matrix, I have a few comments. We should also include some benchmarks in the PR description to confirm that the kernel actually improves performance — both standalone kernel benchmarks and end-to-end ones (e.g., for the full forward pass).

src/transformers/models/qwen3/modular_qwen3.py
kaixuanliu: Here the changes are aimed to pass `python utils/check_modular_conversion.py` in CI process.

src/transformers/models/qwen3_next/modular_qwen3_next.py
kaixuanliu: Same as above.

src/transformers/models/qwen3_moe/modeling_qwen3_moe.py
kaixuanliu: Fixed

yao-matrix:
@MekkCyber , could you pls take another look at it? Thx very much.

IlyasMoutawwakil:
@kaixuanliu can you please add end-to-end perf improvements

kaixuanliu:
@IlyasMoutawwakil ,Hi, I made a benchmark using `Qwen/Qwen3-4B-Instruct-2507` model, and here is the E2E latency result:
| Device           | Configuration         | bs=4   | bs=8   |
|------------------|----------------------|-------:|-------:|
| **A100**         | w/o rotary kernel    | 84.3 s | 98.8 s |
|                  | w/ rotary kernel     | 75.1 s | 92.2 s |
| **Intel XPU (PVC)** | w/o rotary kernel    | 96.2 s | 97.3 s |
|                  | w/ rotary kernel     | 90.8 s | 92.4 s |

MekkCyber (approved):
lgtm ! Pinging @ArthurZucker for the unrelated modular changes, result of running `make fix-copies`

ArthurZucker (commented):
sorry late review here! 
@MekkCyber I think we want to have a much more "invisible" change. `self.rotary_embedding` is a nice direction, in that case we want to allow the decorator to do this part: 
```python 
        rotary_kernel = lazy_load_kernel("rotary_emb")
        self.rotary_fn = (
            rotary_kernel.apply_rotary_transformers
            if rotary_kernel is not None
            and hasattr(rotary_kernel, "apply_rotary_transformers")
            and rotary_kernel.apply_rotary_transformers is not None
            else apply_rotary_pos_emb
        )
```
because we want users to just see the default naive function. 

Appart from that, very happy to merge! Let's strive to make the api similar to hub kernels for modules!

yao-matrix:
@kaixuanliu , could you do that?

kaixuanliu:
> @kaixuanliu , could you do that?

Sure, WIP.

kaixuanliu:
@ArthurZucker Thx for your advice! I added a decorator called `use_kernel_func_from_hub` to do the function level kernel mapping in https://github.com/huggingface/kernels/pull/183. And w/ this the change should be much more concise now. Pls LMK if it is OK. And I will do some refactor work for PR https://github.com/huggingface/kernels/pull/183 if you think this way is acceptable.

MekkCyber:
Hi @kaixuanliu

Thanks a lot for the `kernels` PR! I really appreciate the effort, but I think it adds a bit of extra complexity to how we load kernels for now. I think we could handle the function mapping inside Transformers directly (we can discuss this)

Let's keep the PR in `kernels` open, we can discuss if we can add it or just add the logic in transformers

Sorry again and thanks for your work on this! Let me know what you think

kaixuanliu:
@ArthurZucker @MekkCyber @Rocketknight1 Since Daniel has helped re-implement `use_kernel_func_from_hub` API in https://github.com/huggingface/kernels/pull/188, I updated related code in this PR. Can you help review it again? Thx!

src/transformers/models/dots1/modeling_dots1.py
```diff
@@ -229,6 +230,7 @@ def __init__(self, config: Dots1Config, layer_idx: int):
         self.q_norm = Dots1RMSNorm(self.head_dim, eps=config.rms_norm_eps)  # unlike olmo, only on the head dim!
         self.k_norm = Dots1RMSNorm(self.head_dim, eps=config.rms_norm_eps)  # thus post q_norm does not need reshape
         self.sliding_window = config.sliding_window if self.layer_type == "sliding_attention" else None
+        self.rotary_fn = apply_rotary_pos_emb
```
ArthurZucker: do we need self? if not we can just directly use the func? (i did not follow precisely!)

src/transformers/integrations/hub_kernels.py
```diff
@@ -34,22 +34,52 @@
...
-
             return _kernels_use_kernel_forward_from_hub(layer_name)
         else:
             logger.warning_once(
                 f"kernels hub usage is disabled through the environment USE_HUB_KERNELS={_TRANSFORMERS_USE_HUB_KERNELS}"
             )
             return lambda cls: cls
 
+    def use_kernel_func_from_hub(func_name: str):
+        if _kernels_enabled and _has_use_kernel_func_from_hub:
```
ArthurZucker, line 69: @MekkCyber we need some docs here on usage etc!

ArthurZucker (approved):
Very very nice! IDK if we "have" to use self.rotary func? if not would be perfect hehe 

Kudos everyone 🚀

src/transformers/models/dots1/modeling_dots1.py
kaixuanliu: Here is generated by modular, if we modify this, it will fail for `utils/check_modular_conversion.py`

ArthurZucker: I know I mean for the original model!

src/transformers/models/dots1/modeling_dots1.py
```diff
@@ -200,6 +173,34 @@ def eager_attention_forward(
     return attn_output, attn_weights
 
 
+@use_kernel_func_from_hub("rotary_pos_emb")
```
ArthurZucker: lets put it on llama and all models that have the same no?

kaixuanliu: Thanks for the advice! Since current implemetations will not use the kernels for functions by default as the former version. I think it is ok to add this to all models. Have updated the code.

MekkCyber (approved):
Thanks for your patience @kaixuanliu ! lgtm

src/transformers/models/dots1/modeling_dots1.py
MekkCyber: self here is necessary, the kernelized function must be included in the module so that calling kernelize on the model can detect it.

Cyrilvallez:
Humm, this adds a random `self.rotary_fn` in the module, which is not used... IMO `forward` should be changed to use `self.rotary_fn` then!

kaixuanliu:
Hi, @Cyrilvallez , `self.rotary_fn` is needed here, as curent design for `use_kernel_func_from_hub` decorator need to bind the function to a Module. Maybe we can add a line of comment here?

MekkCyber:
It's not really necessary to use the `self.rotary_fn` since it's only used to make the function discoverable by the `kernelize` process. Btw i'm trying to think of a better way to do that

Cyrilvallez:
Yeah I know it's needed, was just saying that it's a bit awkward rn as it's not being used! But all good if @MekkCyber is looking for a better way then it can wait in the meantime!

kaixuanliu:
Yes, I agree. It would be much better if we use `self.rotary_fn` to replace `apply_rotary_pos_emb` forward. But let's wait to see if  @MekkCyber has some better design.

PR #41156, opened 2025-09-25
Fix inaccurate train_tokens_per_second when resuming from checkpoint

### LINKED ISSUE #40560: `train_tokens_per_second` is wrong after continuing from checkpoint
### System Info

- `transformers` version: 4.55.4
- Platform: macOS-15.6.1-arm64-arm-64bit
- Python version: 3.12.2
- Safetensors version: 0.5.3
- Accelerate version: 1.9.0
- PyTorch version (accelerator?): 2.7.1 (NA)
- Jax version: not installed

### Who can help?

 @zach-huggingface @SunMarc

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

When continuing `run_clm.py` from a checkpoint, it starts with insane tokens per second numbers, then slowly, eventually, goes down. 

Image

Image c

### Expected behavior

It should be calculated with the local training run information, not the global one. The numbers should roughly be the same since it is the same exact setup.

lilin-1:
# Fix inaccurate `train_tokens_per_second` when resuming from checkpoint

## What does this PR do?
This PR addresses the issue where `train_tokens_per_second` calculates incorrectly when resuming training from a checkpoint. 

### Problem Background
Previously, `speed_metrics` used the **cumulative total token count** (`self.state.num_input_tokens_seen`) to compute training speed. However, when resuming from a checkpoint:
1. The cumulative token count includes tokens processed in prior sessions.
2. `start_time` is reset to the current session’s start time.
This mismatch led to abnormally high initial `train_tokens_per_second` values (as the cumulative tokens were divided by a short new time window), which gradually decreased but never accurately reflected the current session’s real throughput (reported in #40560).

### Solution
To ensure speed metrics reflect only the current training session:
1. **Capture baseline token count**: In the `_inner_training_loop` method (right after `start_time = time.time()`), save the initial token count of the current session as `initial_num_input_tokens_seen_for_session`.
2. **Calculate session-specific token increment**: When updating logs with `speed_metrics`, compute the number of tokens processed *in the current session* (via `self.state.num_input_tokens_seen - initial_num_input_tokens_seen_for_session`), and pass this increment to `speed_metrics` instead of the cumulative total.

This change ensures `train_tokens_per_second` consistently represents the real-time throughput of the active training session—whether starting fresh or resuming from a checkpoint.

Fixes #40560

## Before submitting
  Links: #40560
- [ ] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and here are tips on formatting docstrings.  


Rocketknight1:
cc @sunmarc

SunMarc (approved):
Thanks, makes sense !

lilin-1:
Tests keep failing. Only modified training speed calculation in logs，without touching previously failed core modules. Why do failed test cases differ each time and seem unrelated to this change? @SunMarc

SunMarc:
> Tests keep failing. Only modified training speed calculation in logs，without touching previously failed core modules. Why do failed test cases differ each time and seem unrelated to this change? @SunMarc

Some tests are flaky unfortunately ! Don't worry, i will make sure to merge this PR

PR #41158, opened 2025-09-25
fix qwen text config

### LINKED ISSUE #41020: Config rope_scaling and text_config.rope_scaling might be the same or different dict objects
### System Info

- `transformers` version: 4.57.0.dev0
- Platform: Linux-5.15.0-153-generic-x86_64-with-glibc2.31
- Python version: 3.12.9
- Huggingface_hub version: 0.34.4
- Safetensors version: 0.6.2
- Accelerate version: 1.4.0
- Accelerate config: 	not found
- PyTorch version (accelerator?): 2.8.0+cu128 (NA)
- Jax version: not installed

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

After investigation of an issue in `trl`, I found a weird behavior of `transformers` config `rope_scaling`: the `config.rope_scaling` (at the root config level) and `config.text_config.rope_scaling` (under `text_config`) might be the same or different dict objects depending on whether we pass `text_config` param to `AutoConfig.from_pretrained`
- if we don't pass `text_config` param, the 2 `rope_scaling` point to the same dict object
- if we pass `text_config` param, the 2 `rope_scaling` are different dict objects
```python
In [1]: from transformers import AutoConfig

In [2]: model_id = "Qwen/Qwen2.5-VL-3B-Instruct"

In [3]: config1 = AutoConfig.from_pretrained(model_id)
In [4]: config1.text_config.rope_scaling
Out[4]: {'type': 'default', 'mrope_section': [16, 24, 24], 'rope_type': 'default'}
In [5]: config1.rope_scaling
Out[5]: {'type': 'default', 'mrope_section': [16, 24, 24], 'rope_type': 'default'}
In [6]: id(config1.text_config.rope_scaling)
Out[6]: 140211029392000
In [7]: id(config1.rope_scaling)
Out[7]: 140211029392000
# Both are the same dict object

In [8]: config2 = AutoConfig.from_pretrained(model_id, text_config={})
In [9]:config2.text_config.rope_scaling
Out[9]: {'type': 'default', 'mrope_section': [16, 24, 24], 'rope_type': 'default'}
In [10]: config2.rope_scaling
Out[10]: {'type': 'default', 'mrope_section': [16, 24, 24], 'rope_type': 'default'}
In [11]: id(config2.text_config.rope_scaling)
Out[11]: 140210801100608
In [12]: id(config2.rope_scaling)
Out[12]: 140211029786688
# Both are different dict objects
```

Is this expected?

We discovered this while investigating why changing (after initialization) the `config.text_config.rope_scaling` will or will not change the `config.rope_scaling` as well.

### Expected behavior

- Either they should be the same dict object in any case
- Or they should be different dict objects in any case

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/41020 and ensures constency in Qwen-VL text config. Side note: prev we had flat dict structure in Qwen and for BC passed kwargs to `super()` and to `text_config`. This caused confusion in TRL which apparently resets some text attributes manually when training

In this PR, Qwen will set/get text related attributes only through text config. The attributes are obtainable from nested config as `config.text_config.vocab_size` and from root as `config.vocab_size` (BC)


zucchini-nlp:
Test failures are not related

src/transformers/models/qwen2_5_vl/configuration_qwen2_5_vl.py
```diff
@@ -314,8 +309,12 @@ def __init__(
         vision_config=None,
         image_token_id=151655,
         video_token_id=151656,
+        vision_start_token_id=151652,
+        vision_end_token_id=151653,
         **kwargs,
     ):
+        super().__init__(**kwargs)
+
```
Cyrilvallez, line 320: I believe this should stay after setting the subconfigs no? Otherwise the `__setattr__` won't work as it won't see the subcinfigs?

src/transformers/models/qwen2_5_vl/configuration_qwen2_5_vl.py
```diff
@@ -329,8 +328,30 @@ def __init__(
...
+    def __setattr__(self, key, value):
+        if "text_config" in super().__getattribute__("__dict__") and key not in [
+            "dtype",
+            "_attn_implementation_internal",
+        ]:
+            text_config = super().__getattribute__("text_config")
+            if key in text_config.__dict__:
+                return setattr(text_config, key, value)
+
+        return super().__setattr__(key, value)
```
Cyrilvallez: no `return` for `setattr`!

Cyrilvallez (approved):
Arf, very sad we did not use subconfigs when the model was added 🥲

src/transformers/models/qwen2_5_vl/configuration_qwen2_5_vl.py
zucchini-nlp: Ah, I added it for extra kwargs that users can add in a flat dict. Otherwise those are getting serialized in text config because we pass all kwargs to the text config 

For usage it has no effects, only if someone has non-common kwarg which is not supposed to be in text config. I think supporting correct attention is more important than a rare edge case, so I'll revert it

PR #41161, opened 2025-09-25
[v5] Remove old sagemaker api support 

SunMarc:
# What does this PR do?
This PR cleans a bit the sagemaker code. We remove code related to the old sagemaker API (<1.10) as it is 5 years old. In our setup file, the recommended version is >2.0 also.


Cyrilvallez (approved):
🧼🧽🧼

PR #41163, opened 2025-09-25
:rotating_light: [`DistilBert`] Refactor Attention

vasqu:
Same as in bert models, refactor to the attn interface + several other utilities (output xxx, kwargs, ...). It didn't have any dependency on bert which is why this was missed. Should enable vLLM support on our side

cc @hmellor @ArthurZucker


vasqu:
run-slow: distilbert

ArthurZucker (approved):
LGTM 🤗

PR #41165, opened 2025-09-25
[Trainer] deprecate `num_train_tokens`

SunMarc:
# What does this PR do?

This PR deprecates `num_train_tokens` as we already have `include_num_input_tokens_seen` that is better supported.


Cyrilvallez (approved):
lgtm!

PR #41166, opened 2025-09-25
[v5] Remove `model_parallel` deprecated feature 

SunMarc:
# What does this PR do?
This PR removes deprecates code related to `model_parallel` feature in the modeling as well as in Trainer. This was a feature that was added only for a few models like t5 and gpt2 but eventually we switched to using device_map and tp now. 
For the models, we removes `is_parallelizable` and `model_parallel` attributes and some deprecated methods.
For Trainer, we remove a small section related to `model_parallel`.


src/transformers/trainer.py
```diff
@@ -497,26 +497,13 @@ def __init__(
...
                 self.is_model_parallel = self.args.device != torch.device(devices[0])
-            else:
-                self.is_model_parallel = False
-
-            # warn users
-            if self.is_model_parallel:
-                logger.info(
-                    "You have loaded a model on multiple GPUs. `is_model_parallel` attribute will be force-set"
-                    " to `True` to avoid any unexpected behavior such as device placement mismatching."
-                )
```
SunMarc, line 519: this is what changed in trainer cc @qgallouedec

Cyrilvallez (approved):
Nice cleanup, thanks a lot! 🤗 Happy to see it disappear!

PR #41170, opened 2025-09-25
:rotating_light: [`v5`] Remove relative position embeddings (for bert like models)

vasqu:
These embedding types are barely used and make the modeling files just more complex without justifying their existence. Position embedding types still exist in a few models; this PR just addresses the `relative_key(_query)` ones.

Some stats:
- None of the slow tests use them except bert
- The respective models in those tests together have less than 2k downloads in the last month

cc @hmellor this should remove any clashes with the kwargs you encountered in vLLM :D


vasqu:
run-slow: flava, instructblib, mra

examples/modular-transformers/modeling_dummy_bert.py
vasqu, line 1: This is mostly due to me forgetting to update them in my bert refactor PR --> big diff because the whole refactor is included (same for the roberta example)

vasqu:
run-slow: instructblip

vasqu:
run-slow: bert, roberta, albert, mra, instructblip, blip_2, flava

vasqu:
Failing slow tests are the same as in main :eyes:

examples/modular-transformers/modeling_dummy_bert.py
vasqu: Updated: Only includes the changes here now

src/transformers/models/bert/modeling_bert.py
zucchini-nlp, line 120: i think now we can copy bert from llama or another big model group? 👀  Keeping less sources of truth makes it easier to submit PRs

src/transformers/models/instructblip/modeling_instructblip.py
```diff
@@ -502,22 +498,6 @@ def forward(
...
-            positional_embedding = positional_embedding.to(dtype=query_layer.dtype)  # fp16 compatibility
-
-            if self.position_embedding_type == "relative_key":
-                relative_position_scores = torch.einsum("bhld,lrd->bhlr", query_layer, positional_embedding)
-                attention_scores = attention_scores + relative_position_scores
-            elif self.position_embedding_type == "relative_key_query":
-                relative_position_scores_query = torch.einsum("bhld,lrd->bhlr", query_layer, positional_embedding)
-                relative_position_scores_key = torch.einsum("bhrd,lrd->bhlr", key_layer, positional_embedding)
-                attention_scores = attention_scores + relative_position_scores_query + relative_position_scores_key
-
```
zucchini-nlp, line 520: happy to see it, I assumed that BLIP models use relative positions haha. Now it can support attention implementation API in qformer 🙌🏻

zucchini-nlp (approved):
Thanks, super nice clean-up! 🧼

src/transformers/models/instructblip/modeling_instructblip.py
vasqu: yea, it's honestly a bit baffling how many models have this while not using it at all :eyes:

src/transformers/models/bert/modeling_bert.py
vasqu: Let me make a follow-up PR for that, would like to sync bert and bart instead tho since llama would indicate causal masks which is not the case here + unnecessary gqa dependency from llama

vasqu: Opened #41248 for the sync

ArthurZucker (approved):
Longdue ! Thanks

vasqu:
Merging as nothing has changed, just forgot to merge :sweat_smile:

PR #41174, opened 2025-09-26
🚨 [v5] Delete feature extractors used for vision

zucchini-nlp:
# What does this PR do?

As per title, let's clean up for v5

These were supposed to be deleted anyway and we had the warning logged for a long time, even before I joined. Feature Extractors are now reserved for audio models only


molbap:
Nice, IIRC there were a few utils in processing handling that involved using the `feature_extractor` keyword as well

zucchini-nlp:
Yeah, the tests are failing, I will do one more round of cleaning

zucchini-nlp:
Done

gante (approved):
LGTM 🤗 🧹 🧹 🧹 

(probably needs a few more deletions, looking at the CI issues)

PR #41176, opened 2025-09-26
Optimize rope_deltas propagation logic in Qwen2.5-VL

Xqle:
# What does this PR do?

This PR fixes the propagation of `rope_deltas` in the Qwen2.5-VL model during generation. 

Currently, the `forward()` method of `Qwen2_5_VLForConditionalGeneration` accepts `rope_deltas` as an argument, but the value is never passed to the underlying `Qwen2_5_VLModel`. As a result, users providing `rope_deltas` directly to `forward()` would see no effect. 

**Modifications**

1. `Qwen2_5_VLForConditionalGeneration.forward()` now passes `rope_deltas` to `Qwen2_5_VLModel.forward()`.
2. `Qwen2_5_VLModel.forward()` now accepts `rope_deltas` and updates its internal state accordingly.
3. Updates `prepare_inputs_for_generation()` to store calculated rope_deltas
in model_inputs, aligning its handling with position_ids.

**Impact**

- Users can now explicitly provide `rope_deltas` during generation and have them correctly applied.
- Ensures consistency between prefill and decoding phases in multi-modal generation.
- Aligns the API behavior with documentation.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


Rocketknight1:
cc @zucchini-nlp for VLMs, @ArthurZucker for rope

src/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py
```diff
@@ -1282,26 +1282,28 @@ def forward(
                 (cache_position is not None and cache_position[0] == 0)
                 or (past_key_values is None or past_key_values.get_seq_length() == 0)
             )
-            if (prefill_compiled_stage or prefill_noncompiled_stage) or self.rope_deltas is None:
+            if (prefill_compiled_stage or prefill_noncompiled_stage) or rope_deltas is None:
```
molbap, line 1285: Isn't it enough to move above the line  `self.rope_deltas = rope_deltas`?

molbap (commented):
Indeed rope_deltas is not propagated from forward, but I think we can catch it earlier and not touch the rest

molbap:
Btw could this influence https://github.com/huggingface/transformers/issues/41180 ? If the rope_deltas state is never modified in the forward

zucchini-nlp (commented):
Hey @Xqle , thanks for the PR! Indeed they are not being propagated to forward. Though we can also see that the forward does not use `rope_deltas` so the current changes will not be enough to make it work

We had a PR in the past which was very close to how it should look like for Qwen-VL series in https://github.com/huggingface/transformers/pull/39756. It got stale so we could not merge it. I'd suggest to make changes accordingly so that the deltas are actually used by the `forward` (pls take a look at comments under https://github.com/huggingface/transformers/pull/39756) and add a small test

zucchini-nlp:
LMK if you have bandwidth to apply the changes, otherwise I will add it to my TODO so we don't forget adding it in the future :)

PR #41183, opened 2025-09-26
Trainer: Pass `num_items_in_batch` to `compute_loss` in `prediction_step`

pramodith:
# What does this PR do?

Ensures that `num_items_in_batch` is passed to the `compute_loss` function in the `prediction_step` to ensure that loss is calculated the same way both at train and eval time.

Fixes #41108 

## Before submitting
      Pull Request section?
      to it if that's the case.
- [X] Did you write any new necessary tests?

## Who can review?

@SunMarc


pramodith:
I think the test case could be better open to hearing suggestions to add tests (for multi-gpu?) or modify the test.

SunMarc (approved):
LGTM ! Thanks for this nice PR !

src/transformers/trainer.py
```diff
@@ -4897,7 +4897,10 @@ def prediction_step(
             else:
                 if has_labels or loss_without_labels:
                     with self.compute_loss_context_manager():
-                        loss, outputs = self.compute_loss(model, inputs, return_outputs=True)
+                        num_items_in_batch = self._get_num_items_in_batch([inputs], inputs["input_ids"].device)
+                        loss, outputs = self.compute_loss(
```
SunMarc: using "input_ids" might lead to potential issues, just use args.device just like in the training. This is why the CI is not passing

pramodith: lemme change that now.

SunMarc:
Hmmm these tests are still failing: tests/trainer/test_trainer.py::TrainerIntegrationTest::test_evaluate_with_jit, can you quickly check why ? I guess the simplest solution would be to check if we have `self.args.jit_mode_eval` or not. Btw, we should probably deprecate this arg also

pramodith:
> Hmmm these tests are still failing: tests/trainer/test_trainer.py::TrainerIntegrationTest::test_evaluate_with_jit, can you quickly check why ? I guess the simplest solution would be to check if we have `self.args.jit_mode_eval` or not. Btw, we should probably deprecate this arg also

Will take a look in a few hours once I'm off work.

pramodith:
The issue was with `RegressionPretrainedModel's` signature not having `num_items_in_batch` as an argument, which leads to `jit` compile mode not knowing about this argument. I've fixed those test cases in the least intrusive way possible by setting `trainer.model_accepts_loss_kwargs=False`. A bit hacky, let me know if that's not okay and if we should pass an argument to `get_regression_trainer` instead.

SunMarc:
For now, let's do that !

PR #41194, opened 2025-09-29
Don't convert to `safetensors` on the fly if the call is from testing

ydshieh:
# What does this PR do?

Don't convert to `safetensors` on the fly if the call is from testing

### DIFF
--- a/.circleci/create_circleci_config.py
+++ b/.circleci/create_circleci_config.py
@@ -29,6 +29,7 @@ COMMON_ENV_VARIABLES = {
     "RUN_PIPELINE_TESTS": False,
     # will be adjust in `CircleCIJob.to_dict`.
     "RUN_FLAKY": True,
+    "DISABLE_SAFETENSORS_CONVERSION": False,
 }
 # Disable the use of {"s": None} as the output is way too long, causing the navigation on CircleCI impractical
 COMMON_PYTEST_OPTIONS = {"max-worker-restart": 0, "vvv": None, "rsfE":None}
--- a/.circleci/create_circleci_config.py
+++ b/.circleci/create_circleci_config.py
@@ -29,7 +29,7 @@ COMMON_ENV_VARIABLES = {
     "RUN_PIPELINE_TESTS": False,
     # will be adjust in `CircleCIJob.to_dict`.
     "RUN_FLAKY": True,
-    "DISABLE_SAFETENSORS_CONVERSION": False,
+    "DISABLE_SAFETENSORS_CONVERSION": True,
 }
 # Disable the use of {"s": None} as the output is way too long, causing the navigation on CircleCI impractical
 COMMON_PYTEST_OPTIONS = {"max-worker-restart": 0, "vvv": None, "rsfE":None}
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -1073,7 +1073,6 @@ def _get_resolved_checkpoint_files(
                     if resolved_archive_file is not None:
                         # In a CI environment (CircleCI / Github Actions workflow runs) or in a pytest run,
                         # we set `DISABLE_SAFETENSORS_CONVERSION=true` to prevent the conversion.
-                        breakpoint()
                         if (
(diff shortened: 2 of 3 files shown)


src/transformers/modeling_utils.py
```diff
@@ -1071,7 +1071,12 @@ def _get_resolved_checkpoint_files(
                         is_sharded = True
                 if not local_files_only and not is_offline_mode():
                     if resolved_archive_file is not None:
-                        if filename in [WEIGHTS_NAME, WEIGHTS_INDEX_NAME]:
+                        # In a CI environment (CircleCI / Github Actions workflow runs) or in a pytest run, don't perform the conversion
+                        if (
+                            filename in [WEIGHTS_NAME, WEIGHTS_INDEX_NAME]
+                            and os.getenv("CI", "false") != "true"
```
ydshieh: CircleCI / Github Actions workflow runs will set `CI` to `true`

Cyrilvallez:
Hey @ydshieh! Unless I'm mistaken, the conversion is ever run once: if a PR was already opened by the converter bot, then it will use it instead of re-running conversion. 
So I don't really see the use of this PR?

ydshieh:
Hi ✋ 

A few lines below the change

> not has_file(pretrained_model_name_or_path, safe_weights_name, **has_file_kwargs)

This `has_file` will check the server (if it didn't find it on local cache). If the hub has it, it won't perform the conversion, but the call to the server (to check the existence) is still performed, which is what this PR is to prevent (in order to reduce the traffic).

Also, this existence check (and conversion) is not meant to be run within a test (or a CI environment).

Cyrilvallez:
I see - however I don't really like to inject code specific to CI/tests directly in very general loading code like that. Do you mind if we instead check something like `and os.getenv("ALLOW_CONVERSION", "true") == "true"), and set that variable to False on the CI runners/conftest? 😬

ydshieh:
YES, your suggestion looks better!

I will use `DISABLE_SAFETENSORS_CONVERSION` however.

(i.e. if nothing specified, it will do what is currently doing, i.e. conversion. Only if specified as `DISABLE_SAFETENSORS_CONVERSION =true` will change the current behavior)

ydshieh:
@Cyrilvallez updated, thanks for the suggestion

.circleci/create_circleci_config.py
```diff
@@ -29,6 +29,7 @@
     "RUN_PIPELINE_TESTS": False,
     # will be adjust in `CircleCIJob.to_dict`.
     "RUN_FLAKY": True,
+    "DISABLE_SAFETENSORS_CONVERSION": False,
```
Cyrilvallez: If we call the variable "disable", we should set it to True no? Otherwise logic does not match the name!

src/transformers/modeling_utils.py
```diff
@@ -1071,7 +1071,12 @@ def _get_resolved_checkpoint_files(
                         is_sharded = True
                 if not local_files_only and not is_offline_mode():
                     if resolved_archive_file is not None:
-                        if filename in [WEIGHTS_NAME, WEIGHTS_INDEX_NAME]:
+                        # In a CI environment (CircleCI / Github Actions workflow runs) or in a pytest run,
+                        # we set `DISABLE_SAFETENSORS_CONVERSION=false` to prevent to conversion.
+                        if (
+                            filename in [WEIGHTS_NAME, WEIGHTS_INDEX_NAME]
+                            and os.getenv("DISABLE_SAFETENSORS_CONVERSION", None) != "false"
```
Cyrilvallez:
```suggestion
                            and os.getenv("DISABLE_SAFETENSORS_CONVERSION", None) != "true"
```

Cyrilvallez (approved):
Nice, thanks for making the change! I just think we should invert the logic and set it to True no? Otherwise we do not match between the name ("disable") and the actual logic!

.circleci/create_circleci_config.py
ydshieh: you are right

ydshieh:
My brain must be broken this morning ...

ydshieh:
Merge as failing test is irrelevant

PR #41195, opened 2025-09-29
Use accelerator API to free device memory

cyyever:
# What does this PR do?

To simplify device related code.

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -216,6 +216,7 @@ if is_accelerate_available():
         DataLoaderConfiguration,
         DistributedDataParallelKwargs,
         DistributedType,
+        clear_device_cache,
         load_fsdp_model,
         load_fsdp_optimizer,
         release_memory,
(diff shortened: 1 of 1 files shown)


Rocketknight1:
cc @sunmarc

src/transformers/trainer.py
```diff
@@ -4023,22 +4022,7 @@ def training_step(
...
-                    torch.npu.empty_cache()
-                elif is_torch_mps_available():
-                    torch.mps.empty_cache()
-                elif is_torch_hpu_available():
-                    logger.warning(
-                        "`torch_empty_cache_steps` is set but HPU device/backend does not support empty_cache()."
-                    )
-                else:
-                    torch.cuda.empty_cache()
+                release_memory()
```
SunMarc: use clear_device_cache instead

SunMarc (commented):
thanks, just a nit

src/transformers/trainer.py
cyyever: clear_device_cache is not available on accelerator 0.26.0, which is the minimum supported version. So we have to use release_memory. Bump accelerator?

SunMarc: thanks for checking. Hmmm let's keep `release_memory` then and I will change that when we bump accelerate version

cyyever:
@SunMarc It was switched to use `clear_device_cache` and I cleaned up another place using `release_memory`

SunMarc (approved):
Nice

PR #41198, opened 2025-09-29
Remove unnecessary Optional typing

cyyever:
# What does this PR do?

As the title says

### DIFF
--- a/src/transformers/audio_utils.py
+++ b/src/transformers/audio_utils.py
@@ -321,9 +321,7 @@ def mel_to_hertz(mels: Union[float, np.ndarray], mel_scale: str = "htk") -> Unio
     return freq
 
 
-def hertz_to_octave(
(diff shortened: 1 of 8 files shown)


ydshieh (commented):
Is `Optional` only to be used when the default value is `None`?

LGTM, but since I am less familiar with typing stuff, just want to get some confirmation.

Thank you as always

cyyever:
@ydshieh Sometime there is special handling of `is None` for some Optional types, but I can't find it for the changes.

ydshieh (approved):

Tialo:
Actually, `Optional` can be used even without a default value. As official [documentation](https://docs.python.org/3/library/typing.html#typing.Optional) says, `Optional[X]` is equivalent to `Union[X, None]`
They also point out that `def foo(x: int = 0)` does not require `Optional`

cyyever:
@Tialo That is correct. I have checked that the annotated variables in this change don't really have `None` semantics.

ydshieh:
Thanks a lot for the information !

PR #41188, opened 2025-09-28
feat: support cwm modeling

xgal:
# What does this PR do?

Support CWM modeling https://huggingface.co/facebook/cwm/tree/main

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #41199, opened 2025-09-29
Add Code World Model (CWM)

jacobkahn:
Adds the Code World Model (CWM) - https://ai.meta.com/research/publications/cwm-an-open-weights-llm-for-research-on-code-generation-with-world-models/

High-level implementation details:
- This is a GQA + local/global sliding window attention model
- Implemented in HF Llama3 + interleaved sliding window attention
- Inheriting from Gemma2/3 requires weight remapping which breaks VLLM compatibility and other components, so this is implemented using the existing causal mask utils from HF

The model repos are:
- https://huggingface.co/facebook/cwm
- https://huggingface.co/facebook/cwm-sft
- https://huggingface.co/facebook/cwm-pretrain

Note that for VLLM compatibility, model `config.json` still refer to `Llama3ForCausalLM` and a `llama` `model_type` — see [example](https://huggingface.co/facebook/cwm/blob/main/config.json). https://github.com/vllm-project/vllm/pull/25611 adds support mapping `CwmForCausalLM` to the Llama3 model class in VLLM since VLLM supports `Llama3` + `layer_types` with local/global attention - see [docs](https://docs.vllm.ai/en/latest/contributing/model/basic.html#how-to-support-models-with-interleaving-sliding-windows). The model type in the `config.json` will be updated on HF (and the special automapping condition removed) once this PR is merged and a Transformers release has happened containing the `CwmForCausalLM` model class.

@ArthurZucker, @zucchini-nlp

Supersedes https://github.com/huggingface/transformers/pull/41188 due to some fork misery


docs/source/en/model_doc/cwm.md
```diff
@@ -0,0 +1,186 @@
...
+    enable_thinking=True, # Switches between thinking and non-thinking modes. Default is True.
+    preserve_previous_think=True, # Switches between keeping thinking blocks from previous messages or not. Default is True.
+)
+
+
+model_inputs = tokenizer([text], return_tensors="pt").to(model.device)
+
+generated_ids = model.generate(
+    **model_inputs,
+    max_new_tokens=32768
```
zucchini-nlp: re from prev review: not sure if we can keep it this long, can we make it 1024 tokens or less

src/transformers/models/auto/configuration_auto.py
```diff
@@ -1349,6 +1351,16 @@ def from_pretrained(cls, pretrained_model_name_or_path: Union[str, os.PathLike[s
...
+            # Apply heuristic: if model_type is llama but layer_types is present, treat as cwm
+            # TODO: remove this when VLLM 0.11 is released with https://github.com/vllm-project/vllm/pull/25611
+            # Model repos to be updated when requisite transformers version is released with cwm model type
+            if config_dict["model_type"] == "llama" and "layer_types" in config_dict:
+                logger.info(
+                    "Detected llama model with layer_types, treating as cwm for interleaved sliding window "
+                    "attention compatibility. "
+                )
+                config_dict["model_type"] = "cwm"
+
```
zucchini-nlp: this can have undesired side effects. For vLLM we'll use the cwm class after this PR is merged, so lets delete it

src/transformers/models/cwm/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+_import_structure = {
+    "configuration_cwm": ["CwmConfig"],
+}
+
+try:
+    if not is_torch_available():
+        raise OptionalDependencyNotAvailable()
+except OptionalDependencyNotAvailable:
```
zucchini-nlp: init follows a different format and we should not need to explicitly check `is_torch_available()`, can you rewrite similar to llama

https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/__init__.py

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+def _validate_layer_types(layer_types: list[str], num_hidden_layers: int) -> None:
+    if len(layer_types) != num_hidden_layers:
+        raise ValueError(
+            f"layer_types must be a list of length {num_hidden_layers} for each "
+            f"hidden layer, got length {len(layer_types)}"
+        )
+    if any(t not in ("full_attention", "sliding_attention") for t in layer_types):
+        raise ValueError("Layer types must be either 'full_attention' or 'sliding_attention'")
```
zucchini-nlp: we can import it from `modeling_utils`

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+class CwmTextConfig(LlamaConfig):
+    """
+    Llama3-compatible configuration with layer-interleaved sliding-window attention
```
zucchini-nlp: let's add in the docstring that the config is for CWM

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,378 @@
...
+    if len(layer_types) != num_hidden_layers:
+        raise ValueError(
+            f"layer_types must be a list of length {num_hidden_layers} for each "
+            f"hidden layer, got length {len(layer_types)}"
+        )
+    if any(t not in ("full_attention", "sliding_attention") for t in layer_types):
+        raise ValueError("Layer types must be either 'full_attention' or 'sliding_attention'")
+
+
+class CwmTextConfig(LlamaConfig):
```
zucchini-nlp: why `text` config? We can keep as `CwmConfig`. The "text" is usually added in multimodal models to differentiate between configs for different modalities

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,378 @@
...
+        intermediate_size: int = 21504,
+        num_hidden_layers: int = 64,
+        num_attention_heads: int = 48,
+        num_key_value_heads: int = 8,
+        head_dim: int = 128,
+        hidden_act: str = "silu",
+        max_position_embeddings: int = 131072,
+        initializer_range: float = 0.02,
+        rms_norm_eps: float = 1e-5,
+        use_cache: bool = True,
```
zucchini-nlp, line 119: nit: ig these are all optional args?

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,378 @@
...
+            rope_theta=rope_theta,
+            attention_bias=attention_bias,
+            attention_dropout=attention_dropout,
+            rope_scaling=rope_scaling,
+            pretraining_tp=pretraining_tp,
+            mlp_bias=mlp_bias,
+            **kwargs,
+        )
+
+        self.sliding_window = int(sliding_window)
```
zucchini-nlp: we could get `sliding_window=None` (in test models or for users who are experimenting with config params), in which case it will fail to turn to integer

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,378 @@
...
+    based on config.layer_types / sliding_window / global_window before calling attention
+    """
+
+    def __init__(self, config: CwmTextConfig, layer_idx: int):
+        super().__init__(config, layer_idx)
+        self.layer_idx = layer_idx  # Ensure layer_idx is stored as instance attribute
+        self._cwm_layer_types = getattr(config, "layer_types", None)
+        self.layer_type = None
+        if self._cwm_layer_types is not None:
+            self.layer_type = self._cwm_layer_types[self.layer_idx]
```
zucchini-nlp: nit, we can directly access because the config creates layer types even when not provided:
```python
self.layer_type = config.layer_types[self.layer_idx]
```

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,378 @@
...
+
+        residual = hidden_states
+        hidden_states = self.post_attention_layernorm(hidden_states)
+
+        hidden_states = self.mlp(hidden_states)
+        hidden_states = residual + hidden_states
+
+        if use_cache:
+            outputs = (hidden_states,) + outputs
+            return outputs
```
zucchini-nlp: not needed, the `@check_model_inputs` will handle extra output from layers

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+class CwmModelOutputWithPast(BaseModelOutputWithPast):
+    pass
+
+
+class CwmPreTrainedModel(LlamaPreTrainedModel):
+    config_class = CwmTextConfig
+    base_model_prefix = "model"
+
```
zucchini-nlp: nit: we don't need to explicitly define them. Modular will copy it from llama as long as we have the LlamaModel below

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,378 @@
...
+
+class CwmModel(LlamaModel):
+    config_class = CwmTextConfig
+
+    def __init__(self, config: CwmTextConfig):
+        # Validate layer types at model creation time
+        if hasattr(config, "layer_types") and config.layer_types is not None:
+            _validate_layer_types(config.layer_types, config.num_hidden_layers)
```
zucchini-nlp: we should not validate again since it was validated in `config.__init__`. We will also add validation for setting config values manually before v5 (hopefully)

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,378 @@
...
+        )
+
+
+class CwmForCausalLM(LlamaForCausalLM):
+    config_class = CwmTextConfig
+
+    def __init__(self, config: CwmTextConfig):
+        super().__init__(config)
+        self.model = CwmModel(config)
+
```
zucchini-nlp, line 286: we can just `pass`, modular can copy it correctly after we rename the "TextConfig"

tests/models/cwm/test_modeling_cwm.py
```diff
@@ -0,0 +1,341 @@
...
+    def get_config(self):
+        # Add CWM-specific parameters
+        kwargs = {}
+        model_name_to_common_name = {v: k for k, v in self.config_class.attribute_map.items()}
+        for k in self.config_args + self.forced_config_args:
+            if hasattr(self, k) and k != "self":
+                kwargs[k] = getattr(self, k)
+            elif k in model_name_to_common_name and hasattr(self, model_name_to_common_name[k]):
+                kwargs[k] = getattr(self, model_name_to_common_name[k])
+
```
zucchini-nlp: let's call `super()` so that future changes in base class are always propagated to CWM. Then we can re-assign the rope `config.rope_scaling = scaling_dict_cwm`

tests/models/cwm/test_modeling_cwm.py
```diff
@@ -0,0 +1,341 @@
...
+    def test_cwm_config_invalid_layer_types(self):
+        config = self.model_tester.get_config()
+
+        with self.assertRaises(ValueError):
+            config.layer_types = ["invalid_attention"] * config.num_hidden_layers
+            CwmModel(config)
+
+        with self.assertRaises(ValueError):
+            config.layer_types = ["full_attention"] * (config.num_hidden_layers - 1)
+            CwmModel(config)
```
zucchini-nlp: these tests are quite redundant. Layer types are being used by many models and we have a verification before setting them. We should not need to test it much

tests/models/cwm/test_modeling_cwm.py
```diff
@@ -0,0 +1,341 @@
...
+        model.eval()
+
+        # input longer than sliding window
+        seq_length = config.sliding_window + 10
+        input_ids = torch.randint(0, config.vocab_size, (1, seq_length), device=torch_device)
+
+        with torch.no_grad():
+            outputs = model(input_ids)
+
+        self.assertEqual(outputs.last_hidden_state.shape, (1, seq_length, config.hidden_size))
```
zucchini-nlp: better if we can make an integration test and check that the generated ids are correct

tests/models/cwm/test_modeling_cwm.py
```diff
@@ -0,0 +1,341 @@
...
+        model.eval()
+
+        input_ids = torch.tensor([[1, 2, 3]], device=torch_device)  # Simple input
+
+        with torch.no_grad():
+            generated = model.generate(input_ids, max_new_tokens=5, do_sample=False)
+
+        self.assertEqual(generated.shape[0], 1)
+        self.assertGreater(generated.shape[1], input_ids.shape[1])
+
```
zucchini-nlp: not needed, generation functionality is tested already by `GenerationTesterMixin`

tests/models/cwm/test_modeling_cwm.py
```diff
@@ -0,0 +1,341 @@
...
+        model.eval()
+
+        seq_length = 20
+        input_ids = torch.randint(0, config.vocab_size, (1, seq_length), device=torch_device)
+
+        with torch.no_grad():
+            outputs = model(input_ids)
+
+        # no errors
+        self.assertIsNotNone(outputs.last_hidden_state)
```
zucchini-nlp: not clear what we are testing for here, last hidden state can never be `None`, no?

tests/models/cwm/test_modeling_cwm.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+class CwmIntegrationTest(unittest.TestCase):
+    def setUp(self):
+        cleanup(torch_device, gc_collect=True)
+
+    def tearDown(self):
+        cleanup(torch_device, gc_collect=True)
+
+    def test_cwm_small_model_forward(self):
```
zucchini-nlp: let's use actual model for all below tests and check generated token ids or logit values for a few important cases (sliding window, simple generation etc)

zucchini-nlp (commented):
Thanks, left some comments to clean up. Btw, do we have converted weights already which we can use for the integration tests?

src/transformers/models/auto/configuration_auto.py
jacobkahn: Ok — is the strategy then to update the models (i.e. https://huggingface.co/facebook/cwm) to have the right architecture class (`CwmForCausalLM`) and `model_type: "cwm"`, then just have VLLM fail for the time being (when not using main)? This is fine if there there aren't better options

src/transformers/models/cwm/modular_cwm.py
jacobkahn: 👍 will fix -- worth noting that `add-new-model-like` created a bunch of stub code that defaulted to `CwmTextConfig`  (everything had a `CwmText*` prefix) and made removing the old naming very difficult

src/transformers/models/auto/configuration_auto.py
zucchini-nlp: Yep, we can do a pre-release tag for CWM so that users who want to run cwm inference with vLLM can install it. That is what happens for most models

src/transformers/models/cwm/modular_cwm.py
jacobkahn: don't see a validator there or in `masking_utils` - should I create one?

edit: found `from ...configuration_utils import layer_type_validation`

src/transformers/models/cwm/modular_cwm.py
jacobkahn: edit: yes, they're defaults per the config - is there a better way to do this?

src/transformers/models/cwm/modular_cwm.py
zucchini-nlp: Oh, I confused file names, it is here

https://github.com/huggingface/transformers/blob/1f1e93e09537ff148956cb602c85d319e972f190/src/transformers/configuration_utils.py#L1366-L1374

tests/models/cwm/test_modeling_cwm.py
```diff
@@ -89,253 +77,161 @@ class CwmModelTest(CausalLMModelTest, unittest.TestCase):
...
-            sliding_window=32,
-        )
+    @slow
+    def test_cwm_integration(self):
+        from transformers import AutoTokenizer
 
-        model = CwmForCausalLM(config)
-        model.to(torch_device)
-        model.eval()
+        path = "/checkpoint/amaia/codegen/jacobkahn/cwm/v1/release_checkpoints/hf/cwm"
```
zucchini-nlp: we will need to update them before merging to hub repo ids

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -255,42 +231,31 @@ def forward(
...
+    _no_split_modules = ["CwmDecoderLayer"]
+    _skip_keys_device_placement = ["past_key_values"]
+    _supports_flash_attn = True
+    _supports_sdpa = True
+    _supports_flex_attn = True
 
-class CwmRotaryEmbedding(LlamaRotaryEmbedding):
-    pass
+    _can_compile_fullgraph = True
+    _supports_attention_backend = True
```
zucchini-nlp: nit: i think we don't need to redefine them as well, looks similar to llama

zucchini-nlp (approved):

zucchini-nlp:
Can we update slow test ids and then I will trigger the slow CI? Overall lgtm and we can merge

zucchini-nlp:
run-slow: cwm

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,329 @@
...
+    """
+    Same as LlamaDecoderLayer, but we inject an additive mask (local or causal) per layer
+    based on config.layer_types / sliding_window / global_window before calling attention
+    """
+
+    def __init__(self, config: CwmConfig, layer_idx: int):
+        super().__init__(config, layer_idx)
+        self.layer_idx = layer_idx
+        self.layer_type = config.layer_types[self.layer_idx]
+        self.sliding_window = int(getattr(config, "sliding_window", 0))
```
ArthurZucker: there is a better way to handle sliding window! I would use qwen2Attention for example that has slidign

src/transformers/models/cwm/modular_cwm.py
ArthurZucker, line 1: LGTM tho the decoder layer should not change, only the attention layer!

tests/models/cwm/test_modeling_cwm.py
ArthurZucker, line 1: let's add a generation tests

ArthurZucker (commented):
LGTM I think you can completely inherit from `Qwen2Model` tho

src/transformers/models/cwm/modular_cwm.py
jacobkahn: fixed 👍

tests/models/cwm/test_modeling_cwm.py
jacobkahn: Anything in particular that isn't already covered by Llama3?

src/transformers/models/cwm/modular_cwm.py
```diff
@@ -0,0 +1,299 @@
...
+        super().__init__(config=config, layer_idx=layer_idx)
+        self.q_proj = torch.nn.Linear(
+            config.hidden_size, config.num_attention_heads * self.head_dim, bias=config.attention_bias
+        )
+        self.k_proj = torch.nn.Linear(
+            config.hidden_size, config.num_key_value_heads * self.head_dim, bias=config.attention_bias
+        )
+        self.v_proj = torch.nn.Linear(
+            config.hidden_size, config.num_key_value_heads * self.head_dim, bias=config.attention_bias
+        )
```
ArthurZucker:
```suggestion
        self.q_proj = torch.nn.Linear(
            config.hidden_size, config.num_attention_heads * self.head_dim, bias=False
        )
        self.k_proj = torch.nn.Linear(
            config.hidden_size, config.num_key_value_heads * self.head_dim, bias=False
        )
        self.v_proj = torch.nn.Linear(
            config.hidden_size, config.num_key_value_heads * self.head_dim, bias=False
        )
```

I have no idea what your model actually has, but if its never has bias lets remove it entirely!

ArthurZucker (approved):
Let's go! 🤗

src/transformers/models/cwm/modular_cwm.py
jacobkahn: Fixed + removed config param + explicitly deleted for codegen

PR #41204, opened 2025-09-29
Fix multi-video timestamp bug in Qwen-3-VL

tim120526:
# What does this PR do?

This PR fixes a timestamp placeholder error that occurs when multiple videos are included in a single text input.

In `Qwen3-VLProcessor`, frames are sampled from videos, and the relative timestamp placeholder `"<{curr_time:.1f} seconds>"` is constructed for each frame to indicate its corresponding time information. Previously, the code used the index `i` to retrieve the video metadata (`metadata = video_metadata[i]`), which only worked correctly for a single video input. When multiple videos were provided, the metadata of the first video was incorrectly used to generate timestamp placeholders for all the other videos, resulting in errors.

This PR replaces the index `i` with the correct video index `index` when retrieving metadata, ensuring that each video use the appropriate timestamp placeholders.

Hi @zach-huggingface @SunMarc,   I am just following up on this bug and the proposed fix.  It is a one-line change, and I would be happy to assist with any questions or further clarifications to help get it merged.

Thank you very much!


zucchini-nlp (commented):
Thanks, we have a similar issue in [GLM4V](https://github.com/huggingface/transformers/blob/main/src/transformers/models/glm4v/processing_glm4v.py) with the index, can you include the fix as well?

tim120526:
> Thanks, we have a similar issue in [GLM4V](https://github.com/huggingface/transformers/blob/main/src/transformers/models/glm4v/processing_glm4v.py) with the index, can you include the fix as well?

Done. We have fixed the issue in `processing_glm4v.py`.

zucchini-nlp (approved):
Thanks! Can you also make sure the commit history is correct and doesn't include unrelated commits and then run `make fix-copies`?

Btw GLM tests are failing `tests/models/glm4v/test_processor_glm4v.py::Glm4vProcessorTest::test_apply_chat_template_video_1`

tim120526:
> Thanks! Can you also make sure the commit history is correct and doesn't include unrelated commits and then run `make fix-copies`?
> 
> Btw GLM tests are failing `tests/models/glm4v/test_processor_glm4v.py::Glm4vProcessorTest::test_apply_chat_template_video_1`

Sorry, I closed the previous PR and opened a new one here: https://github.com/huggingface/transformers/pull/41229
. The GLM4V error has also been fixed.

PR #20792, opened 2022-12-16
Add Mask2Former

alaradirik:
# What does this PR do?
Adds Mask2Former to transformers.
Original repo: https://github.com/facebookresearch/Mask2Former/
Paper: https://arxiv.org/abs/2112.01527

Co-authored with @shivalikasingh95 

To Do:
- [x] Fix model tests (hidden state shapes, loading the config)
- [X] Test model, visualize outputs
- [X] Update model cards

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [X ] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests),
      Pull Request section?
- [X ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [X ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [X] Did you write any new necessary tests?

PR #41212, opened 2025-09-30
Add EoMT with DINOv3 backbone

NielsRogge:
# What does this PR do?

This PR adds EoMT with a DINOv3 backbone. The authors of EoMT released new checkpoints which swap the DINOv2 backbone by the newer DINOv3: https://github.com/tue-mps/eomt/blob/master/model_zoo/dinov3.md

## Disclaimer

This PR was implemented using OpenAI Codex and further improved by me.

## Who can review?

@yonigozlan


src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,530 @@
...
+        embeddings = self.patch_embeddings(pixel_values.to(dtype=target_dtype))
+        embeddings = embeddings.flatten(2).transpose(1, 2)
+
+        cls_tokens = self.cls_token.expand(batch_size, -1, -1)
+        register_tokens = self.register_tokens.expand(batch_size, -1, -1)
+
+        embeddings = torch.cat([cls_tokens, register_tokens, embeddings], dim=1)
+        embeddings = self.dropout(embeddings)
+
+        return embeddings
```
yonigozlan: It looks like we wouldn't have to override this if we put `num_prefix_tokens` and dropout  in EomtDinov3ForUniversalSegmentation. Same for the Base Eomt

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,530 @@
...
+
+        return mask_logits, class_logits
+
+    @staticmethod
+    def _disable_attention_mask(attn_mask, prob, num_query_tokens, encoder_start_tokens, device):
+        if prob < 1:
+            random_queries = torch.rand(attn_mask.shape[0], num_query_tokens, device=device) > prob
+            attn_mask[:, :num_query_tokens, encoder_start_tokens:][random_queries] = 1
+
+        return attn_mask
```
yonigozlan: I don't think any of this need to be overwritten if we inherit from `EomtForUniversalSegmentation` directly?

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,530 @@
...
+                )
+                loss += self.get_loss(loss_dict)
+
+        return EomtDinov3ForUniversalSegmentationOutput(
+            loss=loss,
+            masks_queries_logits=masks_queries_logits,
+            class_queries_logits=class_queries_logits,
+            last_hidden_state=sequence_output,
+            patch_offsets=patch_offsets,
+        )
```
yonigozlan, line 447: Most of this is very similar to the base Eomt, I'm wondering if we could support the dinov3 version without adding a new model? This would mean defining a backbone instead of copying submodules of dinov2/dinov3, and make position_embeddings optional in the backbone layer call, so not sure how feasible this is, also in terms of BC. Wdyt @NielsRogge ?

yonigozlan (commented):
Hey @NielsRogge ! Looking clean but I wonder if we can simplify even further, or if we even need a new model for this (we would still need to do some refactoring to Eomt though)

src/transformers/models/eomt_dinov3/convert.sh
```diff
@@ -0,0 +1,5 @@
+python convert_eomt_dinov3_to_hf.py \
```
NielsRogge: To be removed

NielsRogge:
@yonigozlan I've addressed your comments. I've updated `modular_eomt_dinov3.py` to inherit from `EomtForUniversalSegmentation` instead. As can be seen, the difference with `modeling_eomt.py` is quite minimal. But we cannot add a self.backbone attribute to `EomtForUniversalSegmentation` as that would be a breaking change with respect to the DINOv2 backbone.

src/transformers/models/eomt/modeling_eomt.py
```diff
@@ -1031,7 +1028,11 @@ def __init__(self, config: EomtConfig):
         super().__init__(config)
         self.config = config
         self.num_hidden_layers = config.num_hidden_layers
+        self.num_prefix_tokens = 1 + config.num_register_tokens
+        self.dropout = nn.Dropout(config.hidden_dropout_prob)
+
         self.embeddings = EomtEmbeddings(config)
+        self.embeddings.num_prefix_tokens = self.num_prefix_tokens
```
yonigozlan: I don't think we need to set this then no?

src/transformers/models/eomt/modeling_eomt.py
```diff
@@ -1190,11 +1196,14 @@ def forward(
     def get_input_embeddings(self):
         return self.embeddings.patch_embeddings
 
+    def get_position_embeddings(self, pixel_values: Tensor) -> Optional[tuple[Tensor, Tensor]]:
+        return None
```
yonigozlan: Not really a fan of this. Maybe let's just unroll the forward in EOMTDinov3 an remove this here

yonigozlan (approved):
I think we can just unroll the forward pass instead of adding a function that always returns None to EoMT, didn't realize that was the only way to make it work.
After that, OK to merge for me!

yonigozlan:
Het @NielsRogge ! I'm ok to merge on the eomt dinov3 part, but it looks like there might have been an issue with branch merging or something?😅

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,468 @@
+# coding=utf-8
+# Copyright 2025 the HuggingFace Team. All rights reserved.
```
molbap: Happy new year! 
```suggestion
# Copyright 2026 the HuggingFace Team. All rights reserved.
```

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,468 @@
...
+class EomtDinov3ForUniversalSegmentationOutput(EomtForUniversalSegmentationOutput):
+    pass
+
+
+class EomtDinov3HungarianMatcher(EomtHungarianMatcher):
+    pass
+
+
+class EomtDinov3Loss(EomtLoss):
+    pass
```
molbap: These inheritances can be cleaned a bit, `EomtDinov3HungarianMatcher` is in `EomtDinov3Loss` so automatic renaming ought to work no? is it because of the prefix EomtDinov3, renaming issue? If it's the latter, maybe doing some 

```python
from ..dinov3.modeling_dinov3 import (
      DINOv3ViTLayer as EomtDinov3Layer,
      DINOv3ViTAttention as EomtDinov3Attention,
  )
```
might work with the converter

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,468 @@
...
+        self.embeddings.register_parameter("mask_token", None)
+
+        self.rope_embeddings = EomtDinov3RopePositionEmbedding(config)
+        self.layers = nn.ModuleList([EomtDinov3Layer(config) for _ in range(config.num_hidden_layers)])
+
+        self.post_init()
+
+    @check_model_inputs()
+    @auto_docstring
+    def forward(
```
molbap, line 344: forward method seems identical to that of `EomtForUniversalSegmentation` except the `rope_embeddings`, since we're already redefining some modeling in the parent model,  separating the embeddings methods would simplify this even more (if we're not changing eomt, disregard that)

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
```diff
@@ -0,0 +1,416 @@
...
+
+        image = Image.open(requests.get("http://images.cocodataset.org/val2017/000000039769.jpg", stream=True).raw)
+
+        inputs = processor(images=image, return_tensors="pt").to(model.device)
+
+        with torch.inference_mode():
+            outputs = model(**inputs)
+
+        self.assertTrue(outputs.class_queries_logits.shape == (1, 200, 134))
+        self.assertTrue(outputs.masks_queries_logits.shape == (1, 200, 160, 160))
```
molbap, line 265: These two tests are just checking for output shape, is that intended? (supposed to be integration)

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
```diff
@@ -0,0 +1,416 @@
...
+        image = Image.open(requests.get("http://images.cocodataset.org/val2017/000000039769.jpg", stream=True).raw)
+
+        pipe = pipeline(model=self.model_id, subtask="panoptic", device=torch_device)
+        output = pipe(image)
+
+        self.assertTrue(len(output) > 0)
+        for segment in output:
+            self.assertIn("score", segment)
+            self.assertIn("label", segment)
+            self.assertTrue(0.0 <= segment["score"] <= 1.0)
```
molbap: same, would be nice to actually measure against some expected score

src/transformers/models/eomt_dinov3/convert.sh
molbap: still to remove :grin:

src/transformers/models/eomt_dinov3/convert_eomt_dinov3_to_hf.py
molbap, line 1: I looked at the converter here - wouldn't that be a job for the  `conversion_mapping`, so that we can load directly from the checkpoints listed? https://github.com/huggingface/transformers/blob/main/src/transformers/conversion_mapping.py

molbap (commented):
Thanks! left some comments related to modular & conversion

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
NielsRogge: I tried this, but the modular converter does not support aliases.

molbap: and do we need both `EomtDinov3HungarianMatcher` and `EomtDinov3Loss` ?

NielsRogge:
All checkpoints have been converted: https://huggingface.co/models?library=transformers&other=eomt_dinov3&sort=trending.

src/transformers/models/eomt/modeling_eomt.py
molbap: read the line and was surprised as well :laughing: to remove @NielsRogge ?

src/transformers/models/eomt_dinov3/convert_eomt_dinov3_to_hf.py
NielsRogge: Not feasible due to the custom `rope_embeddings`

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
NielsRogge: Both are required to compute the loss.

NielsRogge:
@molbap failing tests are unrelated, all checkpoints have been transferred :)

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
molbap: of course, I meant - one is contained in another so it was strange to need both

src/transformers/models/siglip2/modeling_siglip2.py
```diff
@@ -610,6 +610,8 @@ def __init__(self, config: Siglip2TextConfig):
         self.head = nn.Linear(embed_dim, config.projection_size)
         self.post_init()
 
+        self.post_init()
```
molbap, line 606:
```suggestion
        self.post_init()
```

src/transformers/models/siglip/modeling_siglip.py
```diff
@@ -519,6 +519,8 @@ def __init__(self, config: SiglipTextConfig):
         self.head = nn.Linear(embed_dim, config.projection_size)
         self.post_init()
 
+        self.post_init()
```
molbap, line 519:
```suggestion
```

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
molbap: still true - no way to have an actual score? (it's always between 0 and 1 no? so not sure if needed)

molbap (approved):
@vasqu for core review before merge :wink: 
model is a eomt+ dinov3 so weights are reuploaded, the weight converter is not designed for that. @NielsRogge I left a couple comments still!

CLAUDE.md
vasqu, line 1: I guess we remove the agents file as it's more for private usage

docs/source/en/model_doc/eomt_dinov3.md
```diff
@@ -0,0 +1,91 @@
+<!--Copyright 2025 Mobile Perception Systems Lab at TU/e and The Hugging Face team. All rights reserved.
```
vasqu:
```suggestion
<!--Copyright 2026 Mobile Perception Systems Lab at TU/e and The Hugging Face team. All rights reserved.
```
happy new year :D

docs/source/en/model_doc/eomt_dinov3.md
```diff
@@ -0,0 +1,91 @@
...
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+*This model was released on 2025-09-09 and added to Hugging Face Transformers on 2026-01-15.*
+
+# EoMT-DINOv3
+
+<div class="flex flex-wrap space-x-1">
+<img alt="PyTorch" src="https://img.shields.io/badge/PyTorch-DE3412?style=flat&logo=pytorch&logoColor=white">
```
vasqu: We can also add the other attention flavors? E.g. SDPA

```md
    <div class="flex flex-wrap space-x-1">
        <img alt="PyTorch" src="https://img.shields.io/badge/PyTorch-DE3412?style=flat&logo=pytorch&logoColor=white">
        <img alt="SDPA" src="https://img.shields.io/badge/SDPA-DE3412?style=flat&logo=pytorch&logoColor=white">
    </div>
```

src/transformers/models/auto/image_processing_auto.py
```diff
@@ -98,6 +98,7 @@
             ("efficientnet", ("EfficientNetImageProcessor", "EfficientNetImageProcessorFast")),
             ("emu3", ("Emu3ImageProcessor", None)),
             ("eomt", ("EomtImageProcessor", "EomtImageProcessorFast")),
+            ("eomt_dinov3", ("EomtImageProcessor", "EomtImageProcessorFast")),
```
vasqu, line 106: Do we still want to support the slow image processor?

src/transformers/models/eomt/modular_eomt.py
```diff
@@ -525,7 +526,7 @@ def forward(
         if pixel_values is None:
             raise ValueError("You have to specify pixel_values")
 
-        hidden_states = self.embeddings(pixel_values)
+        hidden_states = self.dropout(self.embeddings(pixel_values))
```
vasqu: This change is for modular I suppose?

src/transformers/models/eomt_dinov3/preprocessor_config.json
vasqu, line 1: To remove

src/transformers/models/eomt_dinov3/convert_eomt_dinov3_to_hf.py
vasqu: Rope embeddings should be saved as non-persistent buffers so they should not affect the model weights? Am I missing something? It would be nicer to have the conversion mapping tbh, but dw if not

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,424 @@
...
+        del self.qkv_bias
+        del self.pooler_act
+        del self.pooler_output_size
+        del self.encoder_stride
+        del self.attention_probs_dropout_prob
+        del self.mlp_ratio
+        del self.use_swiglu_ffn
+
+        if intermediate_size is None:
+            intermediate_size = hidden_size * 4
```
vasqu: Not a fan of these ifs, wdyt about directly giving the correct value? Modular always struggles with these so it's IMO better to avoid it as much as we can

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,424 @@
...
+class EomtDinov3Layer(DINOv3ViTLayer):
+    pass
+
+
+class EomtDinov3LayerScale(DINOv3ViTLayerScale):
+    pass
+
+
+class EomtDinov3RopePositionEmbedding(DINOv3ViTRopePositionEmbedding):
+    pass
```
vasqu: Just looked into the modeling code, this is super custom. I'd love to have this refactored at some point to allow custom RoPE more easily - I'll work on that but want to keep you in the loop cc @yonigozlan @molbap

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,424 @@
...
+        self.class_weight = class_weight
+        self.mask_weight = mask_weight
+        self.dice_weight = dice_weight
+        self.train_num_points = train_num_points
+        self.oversample_ratio = oversample_ratio
+        self.importance_sample_ratio = importance_sample_ratio
+        self.num_queries = num_queries
+        self.num_register_tokens = num_register_tokens
+
+        self.rope_theta = rope_theta
```
vasqu: Will refactor along this with rope parameters then

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,424 @@
...
+            init.trunc_normal_(module.weight, mean=0.0, std=std)
+            if module.bias is not None:
+                init.zeros_(module.bias)
+        elif isinstance(module, nn.LayerNorm):
+            init.ones_(module.weight)
+            init.zeros_(module.bias)
+        elif isinstance(module, nn.Embedding):
+            init.normal_(module.weight, mean=0.0, std=1)
+            if module.padding_idx is not None:
+                init.zeros_(module.weight[module.padding_idx])
```
vasqu: These look fairly standard, can we no use `super()._init_weights()` to reduce some code here?

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,424 @@
...
+        self.pos_embed_shift = pos_embed_shift
+        self.pos_embed_jitter = pos_embed_jitter
+        self.pos_embed_rescale = pos_embed_rescale
+
+
+class EomtDinov3ForUniversalSegmentationOutput(EomtForUniversalSegmentationOutput):
+    pass
+
+
+class EomtDinov3PreTrainedModel(PreTrainedModel):
```
vasqu: Why not inherit from e.g. `EomtPreTrainedModel`?

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,424 @@
...
+        elif isinstance(module, EomtDinov3ForUniversalSegmentation):
+            init.ones_(module.attn_mask_probs)
+
+
+@auto_docstring(
+    custom_intro="""
+    The EoMT-DINOv3 model with head on top for instance/semantic/panoptic segmentation.
+    """,
+)
+class EomtDinov3ForUniversalSegmentation(EomtDinov3PreTrainedModel, EomtForUniversalSegmentation):
```
vasqu, line 326: Imo, now it's too late because the base eomt model has it but it should have been split into `EomtDinov3Model` and the `ForUniversalSegmentation`. Maybe a consideration for the future, makes the current model prefix in `PreTrainedModel` useless

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
```diff
@@ -0,0 +1,496 @@
+# Copyright 2025 the HuggingFace Team. All rights reserved.
```
vasqu:
```suggestion
# Copyright 2026 the HuggingFace Team. All rights reserved.
```

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
```diff
@@ -0,0 +1,496 @@
...
+
+@require_torch
+class EomtDinov3ForUniversalSegmentationTest(ModelTesterMixin, PipelineTesterMixin, unittest.TestCase):
+    all_model_classes = (EomtDinov3ForUniversalSegmentation,) if is_torch_available() else ()
+    pipeline_model_mapping = {"image-segmentation": EomtDinov3ForUniversalSegmentation} if is_torch_available() else {}
+    is_encoder_decoder = False
+    test_pruning = False
+    test_head_masking = False
+    test_missing_keys = False
+    test_torch_exportable = False
```
vasqu, line 110: Let's double check these flags, e.g. `test_pruning` and `test_head_masking` are no longer used

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
```diff
@@ -0,0 +1,496 @@
...
+            "mask_labels": torch.randn((2, 10, *size), device=torch_device),
+            "class_labels": torch.zeros(2, 10, device=torch_device).long(),
+        }
+        config = self.model_tester.get_config()
+
+        model = EomtDinov3ForUniversalSegmentation(config).to(torch_device)
+        outputs = model(**inputs)
+        self.assertTrue(outputs.loss is not None)
+
+    def test_attention_outputs(self):
```
vasqu: If we override, let's add some short comment why (also for the other tests)

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
```diff
@@ -0,0 +1,496 @@
...
+        ], device=model.device)
+        expected_masks_logits_slice = torch.tensor([
+            [-1.6251, -1.1417, -1.0285],
+            [ 2.5673,  5.3380,  6.2132],
+            [ 3.7562,  7.1667,  8.1707]
+        ], device=model.device)
+        # fmt: on
+
+        torch.testing.assert_close(
+            outputs.class_queries_logits[0, :3, :3].float(), expected_class_logits_slice, rtol=1e-2, atol=1e-2
```
vasqu: `1e-2` is too high, is it a GPU diff to our CIs? Or something else causing this - are these working as expected locally

We can also update them ourselves afterwards cc @yonigozlan @molbap

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
vasqu: ^ agreeing, let's seed tho to have a reasonable reproducability

vasqu (commented):
It's pretty much ready, just added some comments to
- Remove some local things (agents readme, local json)
- Some pointers what we could improve on our side (not for this PR, e.g. RoPE refactor for custom variants)
- There are some high atols, is this expected? I'd rather we adjust to the A10 CI than have high atols

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
molbap: indeed several VLMs have this, due to multimodal projections compounding errors, I think some models even have it at 1e-1 :skull: 
I think it's still working as intended but @NielsRogge can confirm

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
NielsRogge: I agree on this, curious to know why @yonigozlan approved the original modeling_eomt.py file, I would have also expected it to be split up into `xxxModel` and `xxxForUniversalSegmentation` classes

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
NielsRogge: I have applied this by making `intermediate_size` default to 4096, but modular for some reason added an additional indentation to the docstring of the config.

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
NielsRogge: I have removed the hungarian matcher

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
NielsRogge: Added these!

src/transformers/models/eomt_dinov3/convert_eomt_dinov3_to_hf.py
NielsRogge: I verified it, the conversion mapping is not possible here due to architectural changes, mostly for this reason:
> The DINOv3 attention path applies rotary position embeddings (RoPE) to the query/key tensors at runtime (apply_rotary_pos_emb), and uses a RoPE-specific position embedding module. This is part of the forward pass logic, not just a checkpoint shape difference. 
> The EoMT (DINOv2 backbone) attention does not apply RoPE and is architecturally different at runtime. The attention layer is a standard QKV attention without RoPE logic.

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
NielsRogge: Turns out this test does not need to be overridden

docs/source/en/model_doc/glm4_moe_lite.md
molbap, line 1: wrong rebase? seems unrelated

src/transformers/models/eomt_dinov3/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+from typing import TYPE_CHECKING
+
+from ...utils import _LazyModule
+from ...utils.import_utils import define_import_structure
+
+
+if TYPE_CHECKING:
+    from .configuration_eomt_dinov3 import *
+    from .image_processing_eomt_dinov3 import *
+    from .image_processing_eomt_dinov3_fast import *
```
molbap: these don't exist, no? the automapping is enough
```suggestion
```

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
molbap: bumping this question

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
molbap: First thing: this happens, and there's no need to be pointing fingers. We know how `git blame` works 😉 
Second, looks like it inherits from `Mask2FormerForUniversalSegmentation` which has the same non-split issue and is 3-years old, so it's a long overdue technical debt, which seems to be from this PR and even mentioned in this discussion https://github.com/huggingface/transformers/pull/20792#discussion_r1061343490 - and was approved by you 😛 

So it's a whole series of models to update/improve in a subsequent refactor, and likely also can be absorbed with the dynamic weight loader IMO

src/transformers/models/eomt_dinov3/convert_eomt_dinov3_to_hf.py
molbap: good to take note for an improvement over the conversion mapper tbh!

molbap (approved):
seems good to me - @vasqu for core review! Left some comments.

docs/source/en/model_doc/glm4_moe_lite.md
NielsRogge: When running `make fix-repo`, it updated this file.

src/transformers/models/eomt/modeling_eomt.py
NielsRogge: Have removed this indeed

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
NielsRogge: Haven't used git blame for this, just genuinely curious, it might be there's a good reason for not having `xxxModel` and  `xxxForUniversalSegmentation`.

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
NielsRogge: When I do that, make fix-repo adds an entire new `EomtDinov3PatchEmbeddings` class and updates the init_weights of `EomtDinov3PreTrainedModel` where it doesn't leverage `super()._init_weights(module)` anymore.

docs/source/en/model_doc/glm4_moe_lite.md
vasqu: I think we can revert these changes fully?

src/transformers/models/eomt_dinov3/convert_eomt_dinov3_to_hf.py
vasqu: I'm fine with keeping this script but it's just wrong to say it's because of RoPE as it should not be saved within the weights in the first place --> we have non-persistent buffers, not weights that are serialized. The whole goal of RoPE is just to cache the frequencies so it is not needed to be recalculated

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,415 @@
...
+        no_object_weight: float = 0.1,
+        class_weight: float = 2.0,
+        mask_weight: float = 5.0,
+        dice_weight: float = 5.0,
+        train_num_points: int = 12544,
+        oversample_ratio: float = 3.0,
+        importance_sample_ratio: float = 0.75,
+        num_queries=200,
+        num_register_tokens=4,
+        rope_theta: float = 100.0,
```
vasqu: I thought about this a bit more, while I will refactor rope a bit to allow more custom implementations, I think it makes sense to also adopt the standard style for newer models, i.e. let's remove this + add a `default_theta` + `rope_parameters` (similar to llama)

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
vasqu: would be `rope_parameters` then

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
vasqu: You can inherit from both and do something along https://github.com/huggingface/transformers/blob/14eed387ee6a546529711958731c4267a1f7d94a/src/transformers/models/gemma3/modular_gemma3.py#L579

That would allow the best of both:
- flags from the original eomt
- init with normal overwrite

vasqu: My main reason to inherit is to have an easier time on refactoring these models which Yoni will already have enough work as it is

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
vasqu: Let's add a small comment for the main difference then, i.e. rope

tests/models/eomt/test_modeling_eomt.py
```diff
@@ -128,78 +128,6 @@ def test_model_with_labels(self):
         outputs = model(**inputs)
         self.assertTrue(outputs.loss is not None)
 
-    def test_attention_outputs(self):
```
vasqu, line 131: No longer needed? That's a welcome surprise

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
```diff
@@ -0,0 +1,450 @@
...
+        return config, inputs_dict
+
+
+@require_torch
+class EomtDinov3ForUniversalSegmentationTest(ModelTesterMixin, PipelineTesterMixin, unittest.TestCase):
+    all_model_classes = (EomtDinov3ForUniversalSegmentation,) if is_torch_available() else ()
+    pipeline_model_mapping = {"image-segmentation": EomtDinov3ForUniversalSegmentation} if is_torch_available() else {}
+    is_encoder_decoder = False
+    test_missing_keys = False
+    test_torch_exportable = False
```
vasqu, line 110:
```suggestion
```
I think it was made export friendly with recent commits, can you recheck (might also need to check what changed eomt in case you needed to overwrite certain parts that changed now)

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
```diff
@@ -0,0 +1,450 @@
...
+
+    @unittest.skip(reason="EoMT is not a generative model")
+    def test_generate_without_input_ids(self):
+        pass
+
+    @unittest.skip(reason="EoMT does not use token embeddings")
+    def test_resize_tokens_embeddings(self):
+        pass
+
+    def test_training(self):
```
vasqu, line 152: I don't really see a reason for overwrite?

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
vasqu: bumping

vasqu (commented):
Close to finishing, mostly nits left

I think the biggest thing might be rope --> let's use the power of our rope mixin and start forcing it to be used

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
NielsRogge: Yes there's a specific reason, added a comment on top of the test

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
NielsRogge: Ok, added `rope_parameters`

NielsRogge:
Thanks for your reviews, addressed all comments.

docs/source/en/model_doc/eomt_dinov3.md
```diff
@@ -0,0 +1,92 @@
...
+Below is a minimal example showing how to run panoptic segmentation with a DINOv3-backed EoMT model. The same
+image processor can be reused for semantic or instance segmentation simply by swapping the checkpoint.
+
+```python
+import matplotlib.pyplot as plt
+import requests
+import torch
+from PIL import Image
+
+from transformers import AutoImageProcessor, EomtDinov3ForUniversalSegmentation
```
vasqu: Nit: do we have an auto model for `ForUniversalSegmentation`?

src/transformers/models/eomt_dinov3/convert_eomt_dinov3_to_hf.py
```diff
@@ -0,0 +1,657 @@
+"""Conversion script for EoMT-DINOv3 checkpoints.
```
vasqu, line 14: Licence?

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,436 @@
...
+from ..eomt.configuration_eomt import EomtConfig
+from ..eomt.modeling_eomt import (
+    EomtForUniversalSegmentation,
+    EomtForUniversalSegmentationOutput,
+    EomtLoss,
+    EomtPreTrainedModel,
+)
+
+
+class EomtDinov3Config(EomtConfig):
```
vasqu, line 48: I made the rotary mixin explicit in #43410, but modular doesn't recognize inheriting additional modules (yet)

Wrote a small PR to change this behavior in #43556

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,436 @@
...
+        self.mask_weight = mask_weight
+        self.dice_weight = dice_weight
+        self.train_num_points = train_num_points
+        self.oversample_ratio = oversample_ratio
+        self.importance_sample_ratio = importance_sample_ratio
+        self.num_queries = num_queries
+        self.num_register_tokens = num_register_tokens
+
+        if rope_parameters is None:
+            rope_parameters = {"rope_theta": self.default_theta}
```
vasqu: With the mixin, it will be handled automatically https://github.com/huggingface/transformers/blob/a63032b614dc38f67947a4e588bc385bc7d71e26/src/transformers/modeling_rope_utils.py#L632

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,436 @@
...
+        nn.Module.__init__(self)
+
+        self.config = config
+        self.base = config.rope_parameters["rope_theta"]
+        self.head_dim = config.hidden_size // config.num_attention_heads
+        self.num_patches_h = config.image_size // config.patch_size
+        self.num_patches_w = config.image_size // config.patch_size
+
+        inv_freq = 1 / self.base ** torch.arange(0, 1, 4 / self.head_dim, dtype=torch.float32)  # (head_dim / 4,)
+        self.register_buffer("inv_freq", inv_freq, persistent=False)
```
vasqu: Will be more to overwrite but could we follow llama https://github.com/huggingface/transformers/blob/a63032b614dc38f67947a4e588bc385bc7d71e26/src/transformers/models/llama/modeling_llama.py#L75-L119

I.e. register both `inv_freq` and `original_inv_freq` + `compute_default_rope_parameters` --> I think we can remove the extra care in `_init_weights` then

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,436 @@
...
+    config_class = EomtDinov3Config
+    base_model_prefix = "eomt"
+    main_input_name = "pixel_values"
+    supports_gradient_checkpointing = False
+    _no_split_modules = ["EomtDinov3Layer"]
+    _supports_sdpa = True
+    _can_record_outputs = {
+        "hidden_states": EomtDinov3Layer,
+        "attentions": EomtDinov3Attention,
+    }
```
vasqu: A lot of these attributes should be inheritable so no need to copy all (only unsure about the prefixes not working out maybe, e.g. in `_can_record_outputs`)

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,436 @@
...
+        self.embeddings.register_parameter("mask_token", None)
+
+        self.rope_embeddings = EomtDinov3RopePositionEmbedding(config)
+        self.layers = nn.ModuleList([EomtDinov3Layer(config) for _ in range(config.num_hidden_layers)])
+
+        self.post_init()
+
+    # We redefine forward here because EoMT-DINOv3 uses DINOv3 backbone components (RoPE embeddings, layers)
+    # which require different integration than the base EoMT model that uses a separate encoder.
+    @check_model_inputs()
```
vasqu:
```suggestion
    @check_model_inputs
```

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,436 @@
...
+        mask_labels (`list[torch.Tensor]`, *optional*):
+            list of mask labels of shape `(num_labels, height, width)` to be fed to a model
+        class_labels (`list[torch.LongTensor]`, *optional*):
+            list of target class labels of shape `(num_labels, height, width)` to be fed to a model. They identify the
+            labels of `mask_labels`, e.g. the label of `mask_labels[i][j]` if `class_labels[i][j]`.
+        patch_offsets (`list[torch.Tensor]`, *optional*):
+            list of tuples indicating the image index and start and end positions of patches for semantic segmentation.
+        """
+        masks_queries_logits_per_layer, class_queries_logits_per_layer = (), ()
+        attention_mask = None
```
vasqu:
```suggestion
```
is this needed?

vasqu (commented):
Looks super solid, thanks a lot - only a few small nits. We are kinda depending on #43556 to make modular also allow inheriting the rotary mixin

src/transformers/models/eomt_dinov3/convert_eomt_dinov3_to_hf.py
NielsRogge: Added the license

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
NielsRogge: Done

vasqu:
#43556 is now merged so inheritance should properly work now with modular re the rope mixin

NielsRogge:
@vasqu I have used it, however I cannot have the config inherit from `EomtConfig`, I had to inherit from `PreTrainedConfig`, as modular otherwise introduces additional attributes - seems like a bug

src/transformers/models/eomt_dinov3/__init__.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2025 the HuggingFace Team. All rights reserved.
```
vasqu:
```suggestion
# Copyright 2026 the HuggingFace Team. All rights reserved.
```

src/transformers/models/eomt_dinov3/modular_eomt_dinov3.py
```diff
@@ -0,0 +1,427 @@
...
+
+        self.config = config
+        self.base = config.rope_parameters["rope_theta"]
+        self.head_dim = config.hidden_size // config.num_attention_heads
+        self.num_patches_h = config.image_size // config.patch_size
+        self.num_patches_w = config.image_size // config.patch_size
+
+        inv_freq = 1 / self.base ** torch.arange(0, 1, 4 / self.head_dim, dtype=torch.float32)  # (head_dim / 4,)
+        self.register_buffer("inv_freq", inv_freq, persistent=False)
+        self.register_buffer("original_inv_freq", inv_freq.clone(), persistent=False)
```
vasqu: Can you rewrite it similarly to https://github.com/huggingface/transformers/blob/071e178be163917777dac272e8e26525bc20db08/src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py#L59-L116

The goal is to allow default rope with our own custom logic on init --> we then also don't need overwrite the rope portion in `_init_weights`

vasqu (approved):
Don't have much to add anymore, only one last nit in regards to rope (let's make it a tad more standard/aligned to other models)

vasqu:
run-slow: eomt, eomt_dinov3

vasqu:
run-slow: eomt, eomt_dinov3, vit

vasqu:
run-slow: eomt, eomt_dinov3, vit

vasqu:
@NielsRogge I've fixed the last few smaller things, can you check the failing tests at https://github.com/huggingface/transformers/pull/41212#issuecomment-3824812558 

Moreso the integration tests, if that's a GPU diff (A10) lmk I can update as well in that case - sdpa dispatch one is probably since we are forced to use a mask --> suffices to skip with a comment

vasqu:
run-slow: eomt_dinov3

NielsRogge:
@vasqu feel free to handle the integration tests, I assume this is because of the GPU difference

vasqu:
run-slow: eomt, eomt_dinov3, vit

vasqu:
Merging now, just fixed the last few tests due to GPU diffs resulting in slightly higher atols - updated to a10 numbers where needed (2 tests)

PR #41216, opened 2025-09-30
[generate] cache missing custom generate file

gante:
# What does this PR do?

Reorders code in `load_custom_generate` such that we load the `.py` file first, and check whether it is to be used later. This way, we can cache the result of the Hub request even when the file doesn't exist.

⚠️ Since we try to load `custom_generate` at model-loading time, this results in much fewer Hub requests in situations like CIs. Some downstream CIs were hitting rate limits, in part due to this feature's uncached requests prior to this change.

______________________

✅ ran `custom_generation` tests, including existing slow tests.

### DIFF
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -392,23 +392,20 @@ class GenerationMixin(ContinuousMixin):
         Returns:
             A callable that can be used to generate text.
         """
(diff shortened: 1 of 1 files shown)


Wauplin (approved):
Conceptual LGTM!

BenjaminBossan:
Using on a subset of PEFT tests, this significantly reduces the number of requests (by ~75%) :rocket:

PR #41217, opened 2025-09-30
[`FA3`] Fix masking and loading logic in same process

vasqu:
There are 2 issues that are addressed
1. Fa3 did not use any masking
    - Now registered correctly
    - Tested indirectly in a rewritten parity check between fa2 and fa3 that already exists 
2. Loading logic for anything fa related - kernels, fa2, fa3 - did not work in the same process: Only the first loaded fa variation will be loaded while now it will be the latest.
    - Only forced to reload if an attention type is set to anything fa.

Small benchmark in the parity test also now has a 1.1x speedup (A100) instead of being around 1x (no improvement).


Cyrilvallez (approved):
Ouhh, indeed FA3 was not registered as a mask! Thanks a lot!

PR #41219, opened 2025-09-30
Add missing ModelOutput subclass return type hints

tomaarsen:
# What does this PR do?

This PR adds missing ModelOutput subclass return type hints. I'm running an analysis on the Model classes to see what kind of outputs are used for the various modalities, and I noticed a handful of architectures were missing them.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker 

- Tom Aarsen


tomaarsen:
I fixed the merge conflicts. I totally forgot about this PR, but it should still be valuable to improve the typings somewhat.

- Tom Aarsen

vasqu (approved):
LGTM, I'm just wondering whether you only want to target the base model with this PR / how consistent this needs to be, e.g. https://github.com/huggingface/transformers/blob/b25d98770edf1a2af7377ca5e8bd188ccc394bbc/src/transformers/models/mm_grounding_dino/modeling_mm_grounding_dino.py#L2394 has no output type

tomaarsen:
@vasqu The script I wrote to check the type hints indeed only checked the base models, also because Sentence Transformers will in the future rely on the type hints of the base models. Extending it to the non-base classes would be nice, but I don't have the bandwidth for that now, as I'm aiming to get #42564 ready before the full v5.

- Tom Aarsen

vasqu:
Gotcha, makes sense then. Let's merge, having more typing is better than nothing in the first place

PR #41222, opened 2025-09-30
Fix pylint warnings

cyyever:
# What does this PR do?

More fixes

### DIFF
--- a/src/transformers/tokenization_utils_base.py
+++ b/src/transformers/tokenization_utils_base.py
@@ -2138,7 +2138,7 @@ class PreTrainedTokenizerBase(SpecialTokensMixin, PushToHubMixin):
             if template_file is None:
                 continue  # I think this should never happen, but just in case
             template_name = extra_chat_template.removeprefix("chat_template_")
-            with open(template_file) as chat_template_handle:
+            with open(template_file, "rt", encoding="utf8") as chat_template_handle:
                 chat_templates[template_name] = chat_template_handle.read()
         if len(chat_templates) == 1 and "default" in chat_templates:
             init_kwargs["chat_template"] = chat_templates["default"]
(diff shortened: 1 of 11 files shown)


src/transformers/tokenization_utils_base.py
```diff
@@ -2138,7 +2138,7 @@ def _from_pretrained(
             if template_file is None:
                 continue  # I think this should never happen, but just in case
             template_name = extra_chat_template.removeprefix("chat_template_")
-            with open(template_file) as chat_template_handle:
+            with open(template_file, "rt", encoding="utf8") as chat_template_handle:
```
Rocketknight1: I don't understand why we need these args! Aren't they the default?

Rocketknight1 (approved):
LGTM with one comment!

src/transformers/tokenization_utils_base.py
cyyever: FYI
```
open(file, mode='r', buffering=-1, encoding=None, errors=None, newline=None, closefd=True, opener=None)
The default encoding is platform dependent (whatever [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding) returns)
```

cyyever: But in almost all cases, the code assumes utf8.

Rocketknight1: Good point - specifying utf8 makes sense then. `rt` should be the default, though, according to the documentation for `open()` 
image

Rocketknight1: Or, to put it another way, `r` and `rt` should have the same behaviour, since "t" is assumed.

cyyever: Nice, let's remove `rt` and keep `utf8`.

Rocketknight1 (approved):
Yep, LGTM now!

Cyrilvallez:
Please be very careful with your PRs @cyyever. This one broke gpt-oss dequantization due to the removed param

PR #41223, opened 2025-09-30
🚨 [v5] remove deprecated `generate` classes (constraints and beam scorers)

gante:
# What does this PR do?

The decoding methods using these classes were recently moved to the hub, and these classes were deprecated. They were scheduled to be removed in minor versions that won't exist, so they get tagged as v5 removals :)

(cc @manueldeprada )


zucchini-nlp (approved):
Super cool, thanks

PR #41225, opened 2025-09-30
[voxtral] language detection + skipping lang:xx

eustlb:
# What does this PR do?

Adds the possibility of setting `language=None` in `apply_transcription_request` for automatic language detection.

> [!NOTE]  
> This is **not** breaking. Instead of being required, `language` is now optional.

It was a bit hidden, but Voxtral supports language detection by omitting the language token, e.g.
```
- <s>[INST][BEGIN_AUDIO][AUDIO]...[AUDIO][/INST]lang:en[TRANSCRIBE]
+ <s>[INST][BEGIN_AUDIO][AUDIO]...[AUDIO][/INST][TRANSCRIBE]
```

see [here](https://github.com/mistralai/mistral-common/blob/f88454face1d2b4ae062ca53b7b0272cee6d4511/src/mistral_common/tokens/tokenizers/instruct.py#L955-L958):
```python
tokens = [*prefix, *tokens]
if request.language is not None:
    language_string = f"lang:{request.language}"  # no space.
    tokens += self.tokenizer.encode(language_string, bos=False, eos=False)
```

Nevertheless we should remove it to ease users life.

Added:
- skipping logic in `MistralCommonTokenizer`'s `decode`
- associated `test_decode_transcription_mode`


eustlb:
run-slow: voxtral

src/transformers/tokenization_mistral_common.py
```diff
@@ -471,6 +471,13 @@ def decode(
         if clean_up_tokenization_spaces:
             decoded_string = PreTrainedTokenizerBase.clean_up_tokenization(decoded_string)
 
+        # in the specific case of Voxtral, the added f"lang:xx" (always a two char language code since it follows ISO 639-1 alpha-2 format)
+        # is not considered as a special token by mistral-common and is encoded/ decoded as normal text.
+        # Nevertheless we should remove it to ease users life.
+        lang_prefix = "lang:xx"
+        if skip_special_tokens and decoded_string.startswith("lang:"):
+            decoded_string = decoded_string[len(lang_prefix) :]
+
```
ArthurZucker: if it always starts with it (no other token in between) fine, otherwise would use regex

ArthurZucker (approved):
Nice catch 🤗

src/transformers/tokenization_mistral_common.py
eustlb: it should! but just in case I switched to using regex

PR #41227, opened 2025-09-30
Unify is_torchvision_v2_available with is_torchvision_available

cyyever:
# What does this PR do?

The minimum recommended torchvision version for torch 2.2 is 0.17 .


src/transformers/models/segformer/modular_segformer.py
```diff
@@ -17,6 +17,7 @@
 from typing import Optional, Union
 
 import torch
+from torchvision.transforms.v2 import functional as F
```
Rocketknight1, line 20: In some cases, this results in unguarded `torchvision` imports. Do these still need to be protected by `is_torchvision_available`?

cyyever: Before the change, there is always torchvision.transformes imported, no matter v1 or v2. So the import is unconditional.

Rocketknight1: Okay, makes sense in that case!

Rocketknight1 (approved):

Issue #41184, opened 2025-09-27
Continuous Batching sliding window attention mask is wrong


NixGD:
### System Info

- `transformers` version: 4.57.0.dev0
- Platform: Linux-6.8.0-52-generic-x86_64-with-glibc2.35
- Python version: 3.11.11
- Huggingface_hub version: 1.0.0.rc1
- Safetensors version: 0.6.2
- Accelerate version: 1.10.1
- Accelerate config:    not found
- DeepSpeed version: not installed
- PyTorch version (accelerator?): 2.8.0+cu128 (CUDA)
- Using distributed or parallel set-up in script?: no
- Using GPU in script?: no
- GPU type: NVIDIA A100-SXM4-80GB

### Who can help?

@remi-or 

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

Here's a minimal example showing that the sliding attention leaves more key-value pairs able to be attended to.

```
import torch
from transformers.generation.continuous_batching.continuous_api import build_attention_mask

seq_len = 16
sliding_window = 4
cumulative_seqlens_q = torch.tensor([0, seq_len])
cumulative_seqlens_k = torch.tensor([0, seq_len])

# initialize masks to all ones -- attention is allowed everywhere
window_mask = torch.ones((1, 1, seq_len, seq_len), dtype=torch.float32)
full_mask = torch.ones((1, 1, seq_len, seq_len), dtype=torch.float32)

# build_attention_mask converts this to 0 & -inf
build_attention_mask(window_mask, cumulative_seqlens_q, cumulative_seqlens_k, sliding_window=sliding_window)
build_attention_mask(full_mask, cumulative_seqlens_q, cumulative_seqlens_k, sliding_window=1)

# entries that are still 0 allow the model to attend to that query-key pair
print("Key/Query pairs the model can attend to (full mask):", (full_mask == 0).sum().item())
print("Key/Query pairs the model can attend to (window mask):", (window_mask == 0).sum().item())
```

This outputs
```
Key/Query pairs the model can attend to (full mask): 136
Key/Query pairs the model can attend to (window mask): 202
```

### Expected behavior

Using Continuous Batching with `gpt-oss` gives meaningless results. I believe this is the reason.

[...]

PR #41228, opened 2025-09-30
Fix sliding window attn mask

remi-or:
It was reporedt in #41184 that the sliding window attention mask has errors in CB. This PR fixes this and adds a test for the `build_attention_mask` function, so it does not happen again.


remi-or:
@bot /style

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -63,8 +63,8 @@ def build_attention_mask(
         masked = torch.triu(minus_inf, diagonal=causal_diagonal)
         # Apply sliding window mask if needed
         if sliding_window > 1:
-            sliding_diagonal = seqlen_k - seqlen_q + sliding_window
-            masked = torch.tril(masked, diagonal=sliding_diagonal)
+            sliding_diagonal = seqlen_k - seqlen_q - sliding_window
+            masked += torch.tril(minus_inf, diagonal=sliding_diagonal)
```
ArthurZucker, line 116: Mmmm the sliding indeed has to be substracted but only if seqlen_q > sliding

ArthurZucker (commented):
good that we are having a look. 
If we have:
```
k = 200
q = 200
sliding = 496
k - q - 496 = -496 
```
Probably noop? but not sure 
```
k = 800 
q = 1
sliding = 200
k - q - sliding = 599
```
the 600 index will be attended to, this does look correct

``` 
k = 200 
q = 1
sliding = 400
k - sliding - q = -201
```
not sur that's what we want?

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: I am not sure of this, because the formula works with all those configs: 
```
k = 10, q = 10, sliding = 20
k = 10, q = 1, sliding = 4
k = 10, q = 1, sliding = 20
```

remi-or: Added ASCII drawing, should make things clearer

ArthurZucker (approved):
thanks a lot easier to read

PR #41229, opened 2025-09-30
Fix multi-video timestamp bug in Qwen-3-VL and GLM4V

tim120526:
# What does this PR do?

This PR fixes a timestamp placeholder error that occurs when multiple videos are included in a single text input.
In `Qwen3-VLProcessor,` frames are sampled from videos, and the relative timestamp placeholder `<{curr_time:.1f} seconds>` is constructed for each frame to indicate its corresponding time information. Previously, the code used the index `i` to retrieve the video metadata (`metadata= video_metadata[i]` ), which only worked correctly for a single video input. When multiple videos were provided, the metadata of the first video was incorrectly used to generate timestamp placeholders for all the other videos, resulting in errors.

This PR replaces the index `i` with the correct video index `index` when retrieving metadata, ensuring that each video use the appropriate timestamp placeholders.

Hi [@zach-huggingface](https://github.com/zach-huggingface) [@SunMarc](https://github.com/SunMarc) [@zucchini-nlp](https://github.com/zucchini-nlp), I am just following up on this bug and the proposed fix. It is a one-line change, and I would be happy to assist with any questions or further clarifications to help get it merged.

Thank you very much!

### DIFF
--- a/src/transformers/models/glm4v/modular_glm4v.py
+++ b/src/transformers/models/glm4v/modular_glm4v.py
@@ -1623,7 +1623,7 @@ class Glm4vProcessor(Qwen2_VLProcessor):
                     num_frames = video_grid_thw[video_index][0]
                     video_structure = ""
 
-                    metadata = video_metadata[i]
+                    metadata = video_metadata[video_index]
                     if metadata.fps is None:
                         logger.warning_once(
                             "SmolVLM requires frame timestamps to construct prompts, but the `fps` of the input video could not be inferred. "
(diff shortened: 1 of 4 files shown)


zucchini-nlp:
Amazing thanks, can you update the modular files for models so that the CI turns green?

tim120526:
> Amazing thanks, can you update the modular files for models so that the CI turns green?

done.

zucchini-nlp (approved):

PR #41231, opened 2025-09-30
[repo utils] Update `models_to_deprecate.py`

gante:
# What does this PR do?

Updates the script:
- Fixes hub usage bugs;
- Stop counting downloads after hitting thresholds, to spare the hub from requests;
- Better prints, so it doesn't seem like the script has hung.
- Add support for aliases and folder<>model_tag mapping


utils/models_to_deprecate.py
```diff
@@ -124,19 +128,21 @@ def get_list_of_models_to_deprecate(
         }
 
         # We make successive calls to the hub, filtering based on the model tags
-        n_seen = 0
-        for model, model_info in models_info.items():
+        print("Making calls to the hub to find models below the threshold number of downloads")
+        num_models = len(models_info)
+        for i, (model, model_info) in enumerate(models_info.items()):
+            if max_num_models != -1 and i > max_num_models:
+                break
```
ydshieh: I think this is redundant: it's already controlled when we build `models_info` at the beginning

utils/models_to_deprecate.py
```diff
@@ -124,19 +128,21 @@ def get_list_of_models_to_deprecate(
...
+            print(f"{i+1}/{num_models}: getting hub downloads for {model_info['tags']}")
             for model_tag in model_info["tags"]:
                 model_list = HubModelLister(tags=model_tag)
-                for i, hub_model in enumerate(model_list):
-                    n_seen += 1
-                    if i % 100 == 0:
-                        print(f"Processing model {i} for tag {model_tag}")
-                    if max_num_models != -1 and i > n_seen:
-                        break
+                for hub_model in model_list:
```
ydshieh, line 261: Not a big deal, but do you think it's ever possible a single repository could have more than 1 (model arch.) tags?

ydshieh: and regarding

```
            for model_tag in model_info["tags"]:
                model_list = HubModelLister(tags=model_tag)
```

I think `api.list_models` could take a list of tags

```
    def list_models(
        self,
        *,
        # Search-query parameter
        filter: Union[str, Iterable[str], None] = None,
```
so maybe we just do

```
            model_list = HubModelLister(tags=model_info["tags"])
```
without the for loop?

ydshieh (approved):
Just a few nit questions, but yes, let's do it. Thank you.

utils/models_to_deprecate.py
gante: good point 👍 

Yeah, repos can have more than one tag, but usually they are aliases (I've added aliases in the latest commit). Example: gemma3 models may have `gemma3` and `gemma3_text`, both are aliases of `gemma3`

gante: actually, the script fails if we pass multiple filters together (`filter` must apply an AND, and not an OR), so we need to do one at a time

ydshieh:
Just in case: my concern is more about if we might count the same hub repository multiple times due to the fact we use multiple tags ` for model_tag in model_info["tags"`

```
            for model_tag in model_info["tags"]:
                model_list = HubModelLister(tags=model_tag)  # different `model_tag` might give overlapping repositories ..?
```

It's nit, but could be nice to check / fix when we have the bandwidth to avoid this (if my concern is valid)

gante:
I could double-count yes, we can fix it in the next iteration :D (imo duplicate model tags are very uncommon)

PR #41232, opened 2025-09-30
[kernels] Kernel Config 

MekkCyber:
# What does this PR do?

Adding a kernel config in `from_pretrained` to allow users to use custom kernels. This requires though that the layer is already registered to be used with kernels like `RMSNorm`. Other prs will follow to allow to register kernels without doing that in the modeling.
For `RMSNorm` it looks like this: 
```python
from transformers import AutoModelForCausalLM, AutoTokenizer, KernelConfig
import torch

model_id = "meta-llama/Llama-3.2-1B"

kernel_mapping = {
    "RMSNorm": "kernels-community/layer_norm:LlamaRMSNorm"
}

kernel_config = KernelConfig(kernel_mapping)
    model_id,
    device_map="auto",
    torch_dtype=torch.bfloat16,
    use_kernels=True,
    kernel_config=kernel_config
)

prompts = ["The capital of France is"]
token_dict = tokenizer(prompts, return_tensors="pt").to(model.device)
outputs = model.generate(**token_dict, max_new_tokens=10)
print(tokenizer.decode(outputs[0], skip_special_tokens=True))
```

### DIFF
--- a/src/transformers/utils/kernel_config.py
+++ b/src/transformers/utils/kernel_config.py
@@ -11,11 +11,43 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from ..utils import is_kernels_available
+from ..utils import is_kernels_available, is_torch_available
 
 
 if is_kernels_available():
-    from kernels import LayerRepository, Mode
+    from kernels import Device, LayerRepository, Mode
+
+if is_torch_available():
+    import torch
+
+
+def infer_device(model):
+    EXAMPLE_MAPPING = """
+    {
+        "RMSNorm": {
+            "cuda":
+                "kernels-community/layer_norm:LlamaRMSNorm",
+            ...
+        },
+        ...
+    }
+    """
+    try:
+        param = next(model.parameters())
+    except StopIteration:
+        raise ValueError(
+            f"Cannot determine model device, please provide a device to the mapping. Example: {EXAMPLE_MAPPING}"
+        )
+
+    dev_type = param.device.type
+    if dev_type == "cuda":
+        # Refine based on actual platform
+        if torch.version.hip is not None:
+            return Device(type="rocm")
+        elif torch.version.cuda is not None:
+            return Device(type="cuda")
+
+    return Device(type=dev_type)
 
 
 class KernelConfig:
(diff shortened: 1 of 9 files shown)


ArthurZucker (commented):
Its missing a test! 
Also: 
```python
kernel_mapping = {
    "RMSNorm": {
        "cuda": {
            Mode.INFERENCE: LayerRepository(
                repo_id="kernels-community/layer_norm",
                layer_name="LlamaRMSNorm",
            )
        }
    }
}
```
is there a way to kinda auto detect the devices available? if not provided: 
```python
kernel_mapping = {
    "RMSNorm": "kernels-community/layer_norm"
}
```

MekkCyber:
I refactored the api to allow the users to pass simpler mappings. Tests are coming in an other pr https://github.com/huggingface/transformers/pull/41281 where i will enable the ci for kernels too.

src/transformers/utils/kernel_config.py
```diff
@@ -0,0 +1,207 @@
...
+                    revision=revision,
+                )
+            }
+        }
+
+    def check_if_layer_name_registered(self, layer_name, model):
+        for _, module in model.named_modules():
+            if hasattr(module, "kernel_layer_name") and module.kernel_layer_name == layer_name:
+                return True
+        return False
```
ArthurZucker: we should do this only once, store a mapping module -> layer or something !

src/transformers/utils/kernel_config.py
```diff
@@ -0,0 +1,207 @@
...
+        }
+        """
+        # Validate that the kernel mapping is a dict
+        if not isinstance(self.kernel_mapping, dict):
+            raise ValueError(
+                f"Kernel mapping must be a dict of the following format: {MAPPING_FORMAT}, got: {type(self.kernel_mapping)}"
+            )
+
+        for layer_name, kernel in self.kernel_mapping.items():
+            if not self.check_if_layer_name_registered(layer_name, model):
```
ArthurZucker: re: storing the mapping once

src/transformers/utils/kernel_config.py
```diff
@@ -0,0 +1,207 @@
...
+                repo_layer_name = repo_name.split(":")[1]
+                repo_id = repo_name.split(":")[0]
+                compatible_mapping[layer_name] = {
+                    device: {
+                        mode: LayerRepository(
+                            repo_id=repo_id,
+                            layer_name=repo_layer_name,
+                        )
+                    }
+                }
```
ArthurZucker: no need for nested func definitions

src/transformers/utils/kernel_config.py
```diff
@@ -0,0 +1,207 @@
...
+    dev_type = param.device.type
+    if dev_type == "cuda":
+        # Refine based on actual platform
+        if torch.version.hip is not None:
+            return "rocm"
+
+    return dev_type
+
+
+class KernelConfig:
```
ArthurZucker: why not inherit from ConfigMixin? This one is not push to hub / not even sure its serializable?

ArthurZucker (commented):
Nice nice qol updates!

src/transformers/utils/kernel_config.py
MekkCyber: indeed this is far better! updated

ArthurZucker (approved):
thanks for iterating 😉

PR #41233, opened 2025-09-30
Remove all instances of `is_safetensors_available`

SunMarc:
# What does this PR do?

This PR removes all the checks for safetensors availability as safetensors became a core dep of transformers 2 years ago


src/transformers/modeling_utils.py
```diff
@@ -4816,18 +4801,6 @@ def from_pretrained(
...
-        # Just a helpful message in case we try to load safetensors files coming from old Transformers tf/flax classes
-        if is_safetensors_available() and is_from_file and checkpoint_files[0].endswith(".safetensors"):
-            with safe_open(checkpoint_files[0], framework="pt") as f:
-                metadata = f.metadata()
-            if metadata is not None and metadata.get("format") in ["tf", "flax"]:
-                logger.warning(
-                    "The safetensors checkpoint found has format `tf` or `flax`. This mean that the keys will very"
-                    "likely not match to the model you are trying to load, and will be newly initialized. If it's the case "
-                    "another warning will be raised later. Consider converting your checkpoint to the correct format."
-                )
```
SunMarc, line 4790: should be fine to remove that no ?

Cyrilvallez: I added it when deprecating tf/flax recently as a similar check was performed. I think we should keep it for the first few versions of v5, as some people may have still been using the `from_tf/flax` arguments when loading. Will help them in such a case - but yeah, then we want to get rid of it for sure

Cyrilvallez (approved):
Very nice, thanks! Indeed no need to bother if it's a core dep! I pushed a small branch simplification as well as it was easier than a comment about it!

src/transformers/modeling_utils.py
SunMarc: SG, I will put it back for now !

PR #41234, opened 2025-09-30
[v5] Bump accelerate to 1.1.0 

SunMarc:
# What does this PR do?

This PR bump min version of accelerate to 1.1.0. This should be safe as trl already bump it to 1.4.0.


matthewdouglas:
LGTM! Would you mind updating the other quantizers that check accelerate version? In particular, bnb 4bit / torchao / quanto. Thanks!

SunMarc:
> LGTM! Would you mind updating the other quantizers that check accelerate version? In particular, bnb 4bit / torchao / quanto. Thanks!

It should be good for them as it relies on an older version of accelerate !

matthewdouglas:
> > LGTM! Would you mind updating the other quantizers that check accelerate version? In particular, bnb 4bit / torchao / quanto. Thanks!
> 
> It should be good for them as it relies on an older version of accelerate !

I understand it's safe to keep, but I was thinking it may make sense to clean up if it's now unnecessary, since the overall minimum for transformers will be newer than the versions that are being guarded against.

SunMarc:
Oh indeed, i removed the checks. We should at some point increase the min version of bnb

Cyrilvallez (approved):
Ha nice! Happy to bump and remove all the old checks!

PR #41236, opened 2025-09-30
fix TrainerIntegrationDeepSpeed UT failures

yao-matrix:
when run `TrainerIntegrationDeepSpeed` cases like `pytest -rA tests/deepspeed/test_deepspeed.py::TrainerIntegrationDeepSpeed::test_can_resume_training_normal_zero2_bf16_ds_optim_ds_scheduler`, it will fail w/

>         if "optimizer" in config:
>           if args.adafactor:
> E           AttributeError: 'RegressionTrainingArguments' object has no attribute 'adafactor'
> 
> src/transformers/integrations/deepspeed.py:359: AttributeError

Fix it by using `args.optim`

@ydshieh, @SunMarc  pls help review, thx very much.

### DIFF
--- a/src/transformers/integrations/deepspeed.py
+++ b/src/transformers/integrations/deepspeed.py
@@ -356,7 +356,7 @@ def deepspeed_optim_sched(trainer, hf_deepspeed_config, args, num_training_steps
 
     optimizer = None
     if "optimizer" in config:
-        if args.adafactor:
+        if args.optim == "adafactor":
             raise ValueError(
                 "--adafactor was passed, but also found `optimizer` configured in the DeepSpeed config. "
                 "Only one optimizer can be configured."


ydshieh (approved):
Searching the codebase, this change makes sense to me.

But would prefer @SunMarc to confirme.

Thank you @yao-matrix !

src/transformers/integrations/deepspeed.py
```diff
@@ -356,7 +356,7 @@ def deepspeed_optim_sched(trainer, hf_deepspeed_config, args, num_training_steps
 
     optimizer = None
     if "optimizer" in config:
-        if args.adafactor:
+        if args.optim == "adafactor":
```
ydshieh, line 359: BTW, it's better the message in the `ValueError` below to being updated too.

SunMarc (approved):
Indeed thanks !

PR #41237, opened 2025-09-30
Fix binding of video frames to video placeholder in `InternVL` model

daskol:
This PR fixes broken frame patch indexing in `InternVLProcessor`. If batch is non-trivial (more than one sample) and contains videos and video placeholder in text then the processor returns `pixel_values` of wrong shape. Specifically, it returns `(total_num_frames - 1, ...)` instead of `(total_num_frames, ...)`. The simple code snippet to reproduces follows.

```python
proc = AutoProcessor.from_pretrained('OpenGVLab/InternVL3_5-1B-HF')
texts = [
    '<video>\nAre there any cyan objects that enter the scene?\nno',
    '<video>\nAre there any red spheres that enter the scene?\nno',
]
frames = np.ones((4, 448, 448, 3), np.float32)
inputs = proc(text=texts,
              videos=[frames, frames], return_tensors='pt',
              videos_kwargs={'size': (448, 448)})
print(inputs.pixel_values.shape)  # (7, 3, 448, 448)
assert inputs.pixel_values.shape[0] == 2 * frames.shape[0]  # FAIL: 7 != 8
```

The original lines are authored by @yonigozlan. Could you take a look?

FYI @Rocketknight1 @zucchini-nlp

### DIFF
--- a/src/transformers/models/internvl/processing_internvl.py
+++ b/src/transformers/models/internvl/processing_internvl.py
@@ -132,10 +132,10 @@ class InternVLProcessor(ProcessorMixin):
                     # Get the slice of patches corresponding to the current video
                     # Here we need to account for both the multiple video frames and the potential multiple patches per frame
                     # As of now, InternVL only supports one patch per frame, but we keep the code flexible for future updates
-                    current_patch_index = video_patch_indices[video_index - 1] if video_index > 0 else 0
-                    end_patch_index = video_patch_indices[video_index]
-                    start_index = video_num_patches_indices[current_patch_index] if video_index > 0 else 0
-                    end_index = video_num_patches_indices[end_patch_index - 1]
+                    current_patch_index = video_patch_indices[video_index]
+                    end_patch_index = video_patch_indices[video_index + 1]
+                    start_index = video_num_patches_indices[current_patch_index]
+                    end_index = video_num_patches_indices[end_patch_index]
                     image_video_patches.append(video_pixel_values[start_index:end_index])
                     # Get the number of patches per frame and replace the video placeholder with the correct number of image tokens
                     num_patches = list(video_num_patches[current_patch_index:end_patch_index])
@@ -206,13 +206,8 @@ class InternVLProcessor(ProcessorMixin):
 
         # Process images and videos separately, as videos don't support crop_to_patches
         image_num_patches = []
-        video_num_patches = []
-        image_videos_inputs = {}
         image_pixel_values = None
-        video_pixel_values = None
         image_num_patches_indices = np.array([0])
-        video_patch_indices = np.array([0])
-        video_num_patches_indices = np.array([0])
         if images is not None:
             images = self.image_processor.fetch_images(images)
             images = make_flat_list_of_images(images)
@@ -220,17 +215,29 @@ class InternVLProcessor(ProcessorMixin):
             image_num_patches = image_inputs.pop("num_patches")
             image_pixel_values = image_inputs.pop("pixel_values")
             image_num_patches_indices = np.cumsum(image_num_patches)
+
+        video_num_patches = []  # per frame
+        video_pixel_values = None
+        video_patch_indices = np.array([0])
+        video_num_patches_indices = np.array([0])
         if videos is not None:
-            video_inputs = self.video_processor(videos=videos, **output_kwargs["videos_kwargs"])
+            video_kwargs = {**output_kwargs["videos_kwargs"], 'return_tensors': 'pt'}
+            video_inputs = self.video_processor(videos=videos, **video_kwargs)
             video_pixel_values = video_inputs.pop("pixel_values_videos")
 
-            # Obtain per frame information first and then flatten to (BS * T, ...)
-            num_frames_per_video = [len(video) for video in video_pixel_values]
-            video_num_patches = [1 for frames in num_frames_per_video for _ in range(frames)]
-            video_patch_indices = np.cumsum(num_frames_per_video)
-            video_num_patches_indices = np.cumsum(video_num_patches)
+            batch_size, num_frames, *_ = video_pixel_values.shape
+            num_frames_per_video = np.full(batch_size, num_frames)
+            num_frames = sum(num_frames_per_video)  # total
+            video_patch_indices = np.empty(batch_size + 1, int)
+            video_patch_indices[0] = 0
+            video_patch_indices[1:] = np.cumsum(num_frames_per_video)
+            video_num_patches = [1] * num_frames
+            video_num_patches_indices = np.empty(num_frames + 1, int)
+            video_num_patches_indices[0] = 0
+            video_num_patches_indices[1:] = np.cumsum(video_num_patches)
             video_pixel_values = video_pixel_values.flatten(0, 1)
 
+        image_videos_inputs = {}
         if images is not None or videos is not None:
             text, image_video_patches, image_index, video_index = self._insert_media_placeholders(
                 text,
--- a/src/transformers/models/internvl/processing_internvl.py
+++ b/src/transformers/models/internvl/processing_internvl.py
@@ -221,7 +221,7 @@ class InternVLProcessor(ProcessorMixin):
         video_patch_indices = np.array([0])
         video_num_patches_indices = np.array([0])
         if videos is not None:
-            video_kwargs = {**output_kwargs["videos_kwargs"], 'return_tensors': 'pt'}
+            video_kwargs = {**output_kwargs["videos_kwargs"], "return_tensors": "pt"}
             video_inputs = self.video_processor(videos=videos, **video_kwargs)
             video_pixel_values = video_inputs.pop("pixel_values_videos")
 
--- a/src/transformers/models/internvl/processing_internvl.py
+++ b/src/transformers/models/internvl/processing_internvl.py
@@ -40,7 +40,9 @@ class InternVLProcessorKwargs(ProcessingKwargs, total=False):
         "images_kwargs": {
             "crop_to_patches": True,
         },
-        "videos_kwargs": {},
+        "videos_kwargs": {
+            "return_tensors": "pt",
+        },
     }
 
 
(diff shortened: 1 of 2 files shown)


src/transformers/models/internvl/processing_internvl.py
```diff
@@ -206,31 +206,50 @@ def __call__(
...
-            video_inputs = self.video_processor(videos=videos, **output_kwargs["videos_kwargs"])
+            # Here, `videos` are batch of frames (images) that is a list of 4D
+            # arrays. Since spatial dimensions of a video in a batch are
+            # arbitrary, `videos` is a list. However, `video_pixel_values` is
+            # scaled/cropped/resized/etc in order to constitute a 5D array of
+            # frames.
+            #
+            # NOTE There is no multi-patch support per frame.
+            video_kwargs = {**output_kwargs["videos_kwargs"], 'return_tensors': 'pt'}
+            video_inputs = self.video_processor(videos=videos, **video_kwargs)
```
yonigozlan: `return_tensors` should already be included in output_kwargs["videos_kwargs"]
```suggestion
            video_inputs = self.video_processor(videos=videos, **output_kwargs["videos_kwargs"])
```

src/transformers/models/internvl/processing_internvl.py
```diff
@@ -206,31 +206,50 @@ def __call__(
...
+        video_num_patches_indices = np.array([0])
         if videos is not None:
-            video_inputs = self.video_processor(videos=videos, **output_kwargs["videos_kwargs"])
+            # Here, `videos` are batch of frames (images) that is a list of 4D
+            # arrays. Since spatial dimensions of a video in a batch are
+            # arbitrary, `videos` is a list. However, `video_pixel_values` is
+            # scaled/cropped/resized/etc in order to constitute a 5D array of
+            # frames.
+            #
+            # NOTE There is no multi-patch support per frame.
```
yonigozlan:
```suggestion
```

src/transformers/models/internvl/processing_internvl.py
```diff
@@ -206,31 +206,50 @@ def __call__(
...
-            # Obtain per frame information first and then flatten to (BS * T, ...)
-            num_frames_per_video = [len(video) for video in video_pixel_values]
-            video_num_patches = [1 for frames in num_frames_per_video for _ in range(frames)]
-            video_patch_indices = np.cumsum(num_frames_per_video)
-            video_num_patches_indices = np.cumsum(video_num_patches)
+            # Now, we need to build a multi-index on all patches in a batch
+            # across videos and frames. We need this data structure to
+            # efficiently build an input data with respect to prompt. The issue
+            # is that a prompt in a batch does not necessary refer to a
+            # corresponding video.
```
yonigozlan: Too verbose and unclear

tests/models/internvl/test_processing_internvl.py
```diff
@@ -377,3 +379,22 @@ def _test_apply_chat_template(
...
+        videos = [frames, frames]
+
+        processor = self.get_processor()
+        inputs = processor(text=texts[:batch_size], return_tensors='pt',
+                           videos=videos[:batch_size],
+                           videos_kwargs={'size': (448, 448)})
+
+        actual_num_frames = inputs.pixel_values.pixel_values.shape[0]
+        expected_num_frames = sum(x.shape[0] for x in videos[:batch_size])
+        assert actual_num_frames == expected_num_frames
```
yonigozlan, line 408: Thanks for adding a test! But let's find a better/more descriptive name for it

yonigozlan (approved):
Thanks for fixing! I suggested some small changes, then LGTM!

src/transformers/models/internvl/processing_internvl.py
daskol: Well, but it is `None` if it is not set explicitly to `'pt'`. I just checked now.

src/transformers/models/internvl/processing_internvl.py
daskol: Fixed.

src/transformers/models/internvl/processing_internvl.py
daskol: Fixed.

(I see. Only maintainers are allowed to leave a comment J)

tests/models/internvl/test_processing_internvl.py
daskol: Fixed.

src/transformers/models/internvl/processing_internvl.py
yonigozlan: It might be because of the way you call the processor in your example:
```python
inputs = proc(text=texts,
              videos=[frames, frames], return_tensors='pt',
              videos_kwargs={'size': (448, 448)})
```
Here you can just use size and it will automatically be used in video_kwargs, with return_tensors="pt":
```python
inputs = proc(text=texts,
              videos=[frames, frames], 
              size=(448, 448),
              return_tensors='pt',
```
Or you can add the return_tensors arg to the videos_kwargs as well, otherwise it's overwritten to None, and not set to the global return_tensors:
```python
inputs = proc(text=texts,
              videos=[frames, frames], return_tensors='pt',
              videos_kwargs={'size': (448, 448), return_tensors='pt',})
```

yonigozlan:
@daskol  Looks like some InternVL processor tests are failing, you can run them locally using:
`RUN_SLOW=1 python -m pytest tests/models/internvl/test_processing_internvl.py` ;)

src/transformers/models/internvl/processing_internvl.py
daskol: No, it tried different combinations for `return_tensors`. The original code fails if `return_tensors` is not specified on flattening `videos_pixel_values.faltten(0, 1)` in the end of this branch (i.e. the code implies that `return_tensors='pt'` is enforced).

You can easily reproduce this case with the code in the initial post.

src/transformers/models/internvl/processing_internvl.py
```diff
@@ -206,31 +206,38 @@ def __call__(
...
+            video_kwargs = {**output_kwargs["videos_kwargs"], "return_tensors": "pt"}
+            video_inputs = self.video_processor(videos=videos, **video_kwargs)
             video_pixel_values = video_inputs.pop("pixel_values_videos")
 
-            # Obtain per frame information first and then flatten to (BS * T, ...)
-            num_frames_per_video = [len(video) for video in video_pixel_values]
-            video_num_patches = [1 for frames in num_frames_per_video for _ in range(frames)]
-            video_patch_indices = np.cumsum(num_frames_per_video)
-            video_num_patches_indices = np.cumsum(video_num_patches)
+            batch_size, num_frames, *_ = video_pixel_values.shape
```
yonigozlan: Actually, I'm not sure this is a good idea, as there can be videos with different number of frames. In this case, `"return_tensors": "pt"`, will fail. let's keep the previous:
```python
num_frames_per_video = [len(video) for video in video_pixel_values]
```
And let's not force  `"return_tensors": "pt"`

daskol: The original code has already forced (see `video_pixes_values.flatten(0, 1)` and comment above https://github.com/huggingface/transformers/pull/41237#discussion_r2392808678).

yonigozlan: Yes my bad, but this is a problem I think.
In my opinion, it would be better to force `return_tensors` to `None` in the `video_processor` call, so that we can support batch processing of videos with different number of frames.
We could have something like this in the loop:
```python
            return_tensors = output_kwargs["videos_kwargs"].pop(return_tensors, None)
            video_inputs = self.video_processor(videos=videos, **output_kwargs["videos_kwargs"], return_tensors=None)
            video_pixel_values = video_inputs.pop("pixel_values_videos")
            num_frames_per_video = [len(video) for video in video_pixel_values]
            num_frames = sum(num_frames_per_video)  # total
            video_patch_indices = np.empty(batch_size + 1, int)
            video_patch_indices[0] = 0
            video_patch_indices[1:] = np.cumsum(num_frames_per_video)
            video_num_patches = [1] * num_frames
            video_num_patches_indices = np.empty(num_frames + 1, int)
            video_num_patches_indices[0] = 0
            video_num_patches_indices[1:] = np.cumsum(video_num_patches)
            # flatten
            video_pixel_values = [video_frame for video in video_pixel_values for video_frame in video]
            video_pixel_values = torch.stack(video_pixel_values) if return_tensors=="pt" else video_pixel_values 
```

What do you think @zucchini-nlp ? Am I missing something?

daskol: I suggest to keep it as simple as possible and work on the better solution later.

From my perspective, the best solution is to consider content processing as a binding prepared statement to actual variable. I mean that we should to parse prompt, extract content placeholders and load content (e.g. video from a file) on demand. At the moment, all content is loaded greedily. It is quite common pattern in content processing. Hence, it should live in base processor.

Moreover, there is no proper support for multi-patch frames. This makes the situation even more complicated.

daskol:
@yonigozlan Thanks for the oneliner to run tests. All jobs are green now.

src/transformers/models/internvl/processing_internvl.py
zucchini-nlp: I agree that we need to support variable number of frames when possible. Some models support it by patchifying or by padding

I am fine with leaving it since it has been working with torch-assumption for a while and not all models support variable frame count. Let's try to fix it in the subsequent PR though

src/transformers/models/internvl/processing_internvl.py
```diff
@@ -206,31 +206,38 @@ def __call__(
...
             image_pixel_values = image_inputs.pop("pixel_values")
             image_num_patches_indices = np.cumsum(image_num_patches)
+
+        video_num_patches = []  # per frame
+        video_pixel_values = None
+        video_patch_indices = np.array([0])
+        video_num_patches_indices = np.array([0])
         if videos is not None:
-            video_inputs = self.video_processor(videos=videos, **output_kwargs["videos_kwargs"])
+            video_kwargs = {**output_kwargs["videos_kwargs"], "return_tensors": "pt"}
```
zucchini-nlp: i think we should enforce it in the `ModelKwargs` defaults

daskol: I didn't get it. What is `ModelKwargs`?

zucchini-nlp: sorry, I meant the a typed dict - 'InternVLKwargs', which has defaults for each modality

daskol: Fixed.

zucchini-nlp (approved):
Oke, lets merge it. Thanks

daskol:
Thank you for your time.

PR #41240, opened 2025-10-01
Use removeprefix and removesuffix

cyyever:
# What does this PR do?

To simplify code

### DIFF
--- a/src/transformers/models/rag/retrieval_rag.py
+++ b/src/transformers/models/rag/retrieval_rag.py
@@ -509,10 +509,8 @@ class RagRetriever:
         def cat_input_and_doc(doc_title, doc_text, input_string, prefix):
             # TODO(Patrick): if we train more RAG models, I want to put the input first to take advantage of effortless truncation
             # TODO(piktus): better handling of truncation
-            if doc_title.startswith('"'):
-                doc_title = doc_title[1:]
-            if doc_title.endswith('"'):
-                doc_title = doc_title[:-1]
+            doc_title = doc_title.removeprefix('"')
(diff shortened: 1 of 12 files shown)


src/transformers/models/rag/retrieval_rag.py
```diff
@@ -509,10 +509,8 @@ def postprocess_docs(self, docs, input_strings, prefix, n_docs, return_tensors=N
         def cat_input_and_doc(doc_title, doc_text, input_string, prefix):
             # TODO(Patrick): if we train more RAG models, I want to put the input first to take advantage of effortless truncation
             # TODO(piktus): better handling of truncation
-            if doc_title.startswith('"'):
-                doc_title = doc_title[1:]
-            if doc_title.endswith('"'):
-                doc_title = doc_title[:-1]
+            doc_title = doc_title.removeprefix('"')
+            doc_title = doc_title.removesuffix('"')
```
ydshieh: could be chain in the same line, but up to you

ydshieh (approved):
I learn so much python today ... thanks as always.

@Cyrilvallez Since core files (like `src/transformers/modeling_utils.py`) are touched, you might want to take a look too despite the change is obvious.

src/transformers/models/rag/retrieval_rag.py
cyyever: Fixed

Rocketknight1 (approved):
LGTM! I don't think we need a core maintainer since the changes are equivalent, we're not modifying any of the functional code

PR #41241, opened 2025-10-01
Use math.log2

cyyever:
# What does this PR do?

This PR changes `math.log(XXX, 2)` to `math.log2(XXX)`, to make the invocation more clear.


ydshieh (approved):
Wow, you know too much python function 😄 ❤️ 💯

PR #41248, opened 2025-10-01
[`v5`] Sync Bert and Bart eager attention

vasqu:
Syncs the eager attentions if those model groups to have less intertwined versions of essentially the same; sync goes from bert as being the primary source

NOTE: Since it has a hard dependency on #41170, it should only come with v5 as well - do we still use the labels or just go for it since we have v4.57?

cc @zucchini-nlp


vasqu:
Rather wait for #41170 first otherwise the review here gets unnecessarily convoluted :D

ArthurZucker:
Waiting for the orther PR to be merged before I review!

src/transformers/models/bart/modeling_bart.py
```diff
@@ -124,18 +126,21 @@ def eager_attention_forward(
...
+    **kwargs: Unpack[TransformersKwargs],
 ):
     if scaling is None:
         scaling = query.size(-1) ** -0.5
 
+    # Take the dot product between "query" and "key" to get the raw attention scores.
     attn_weights = torch.matmul(query, key.transpose(2, 3)) * scaling
-    if attention_mask is not None:
+
+    if attention_mask is not None and attention_mask.ndim == 4:
```
Cyrilvallez: Do we need the dim check? In the function, the mask should only be None or 4d, otherwise something is wrong upstream!

Cyrilvallez (approved):
LGTM on principle, nice to keep only 1 source, but we should not need a dim check!

src/transformers/models/bart/modeling_bart.py
zucchini-nlp: just curious: in which cases we can get 2D mask and do we need to raise error/warning to let users know that their inputs will not be masked?

zucchini-nlp (approved):
Nice!

src/transformers/models/bart/modeling_bart.py
vasqu: Yes true, I kinda had the old llama mask logic in my mind :sweat_smile: if we encounter a non 4D mask here something went seriously wrong...

vasqu: Users should only pass either 2D masks (we handle all logic -> 4D) or 4D masks (we expect the user to apply the correct logic and only pass it). If we get a non-4D mask (not None) at that point, we have a bigger issue at hand as Cyril also said it.

Side note:
Currently, 3D is supported in Bert (like) models which imo is a bad design as in those cases a simple expand suffices, e.g. `(bsz, q_len, kv_len)` to `(bsz, 1, q_len, kv_len)`. #41265 will remove the 3D support tho - it's a weird use-case and power users should be able to infer 4D by themselves :man_shrugging:

PR #40352, opened 2025-08-21
Bug Fix: Dynamically set return_lse flag in FlexAttention

amd-lalithnc:
# What does this PR do?

Dynamically sets return_lse depending on output_attentions from model config. This enables the models to run on CPU devices as well. 

Fixes # (issue)
#40345 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #41083, opened 2025-09-23
Fix attention sink implementation in flex attention

SamuelBarryCS:
## What
- Fixes #41026 where @jonny-so correctly identified that attention sinks were being incorrectly applied within the [score_mod] function of flex attention
- The previous implementation attempted to apply attention sinks by manipulating pre-softmax scores within score_mod, which is incorrect because attention sinks require access to the full attention matrix after softmax normalization ==> this fix moves attention sink application after flex attention computation to renormalize the output correctly
- As a result, we can now run flex attention with attention sink (which was impossible before because it would crash when s_aux was provided)

## Test performed
- No major logic change, existing tests are thus still passing
- Created a small script `test_attention_sink.py` to test the old vs. new version of the attention sink implementation in flex attention.` flex_attention_old.py` is simply a copy/paste from the file `flex_attention.py` on main, and this script + `test_attention_sink.py`  are of course both meant to be deleted before merging. The fix is confirmed by the script output:

```
Testing attention sinks implementation...
Input shapes: query torch.Size([1, 2, 4, 16]), key torch.Size([1, 2, 4, 16]), value torch.Size([1, 2, 4, 16])
s_aux shape: torch.Size([2])

--- Old Implementation ---
Status: FAILED - ['LoweringException: AssertionError: wrong ndim () [2]', '  target: flex_attention']

--- New Implementation ---
Before calling flex_attention_new...
torch.Size([1, 2, 4, 16]) torch.Size([1, 2, 4]) torch.Size([2])
After calling flex_attention_new...
Status: SUCCESS
Output: (tensor([[[[-0.4479, -0.0579,  0.0891, -0.1978,  0.4921, -0.0836, -0.4697,
            0.2943,  0.4493, -0.2208, -0.1416, -0.0743, -0.2410,  0.4453,
           -0.3093,  0.0605],
[...]
```

## How to review
- Check diff
- Run `test_attention_sink.py`

Issue #40345, opened 2025-08-21
Flexattention Implementation in Gemma Failing on CPU


amd-lalithnc:
### System Info

- `transformers` version: 4.55.3
- Platform: Linux-5.15.0-119-generic-x86_64-with-glibc2.35
- Python version: 3.10.18
- Huggingface_hub version: 0.34.4
- Safetensors version: 0.6.2
- Accelerate version: not installed
- Accelerate config: not found
- DeepSpeed version: not installed
- PyTorch version (accelerator?): 2.8.0+cpu (NA)
- Tensorflow version (GPU?): not installed (NA)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
- Using distributed or parallel set-up in script?: No

### Who can help?

@ArthurZucker 

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

Loading a flex-attention supported model such as Gemma fails - returning the error - _detailed error logs below_
```bash
torch._inductor.exc.InductorError: LoweringException: NotImplementedError: torch.compile on CPU only supports inference and `return_lse` is not supported yet.
  target: flex_attention
```

```py
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer

def main():
    model_id = "google/gemma-2-2b"  # any flex-attn supported model is fine
    tokenizer = AutoTokenizer.from_pretrained(model_id)

    text = "The quick brown fox jumps over the lazy dog. " * 10
    inputs = tokenizer(
        text,
        return_tensors="pt",
        max_length=256,
        truncation=True,
        padding=True,
    )

    model = AutoModelForCausalLM.from_pretrained(
        model_id,
        torch_dtype=torch.bfloat16,          # matches failing environment; use float32 if bf16 unsupported
        attn_implementation="flex_attention",
        output_attentions=False,
    )

    with torch.no_grad():
        _ = model(**inputs)

if __name__ == "__main__":
    main()
```

### Expected behavior

Traceback (most recent call last):
  File "/proj/xcohdstaff2/lalithnc/flex-attention/benchmark-scripts/ZenDNN_tools/benchmark/flexattention_bench/../../../../experiments/3_llm.py", line 187, in <module>
    main()
[...]

PR #41249, opened 2025-10-01
[`Flex Attn`] Fix lse x attention sinks logic  

vasqu:
We always return lse except for CPU #40352 (it simply isn't supported in torch yet)

Let's rather raise a proper error for this than encountering a verbose torch exception

### DIFF
--- a/src/transformers/integrations/flex_attention.py
+++ b/src/transformers/integrations/flex_attention.py
@@ -282,10 +282,10 @@ def flex_attention_forward(
     # On CPU we must skip returning LSE due to a runtime issue; elsewhere, follow PyTorch API and return it
     return_lse = query.device.type != "cpu"
 
-    # Validate that s_aux is not silently ignored
     if not return_lse and s_aux is not None:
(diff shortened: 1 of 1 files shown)


ArthurZucker (commented):
haha I am pretty sure when I reviewed I mentioned this: we have to force return_lse for models that "require" it :wink: if saux -> set return lse (it won't be used anyways)

vasqu:
But it will lead to a torch error when we force it, I'd rather let the user know explicitly what's going on than to get a verbose torch error (#40345). We kinda intentionally run into that error that way

ArthurZucker:
Ah no for CPU absolutely, I was thinking about this one: #41083

vasqu:
Ah, yes but there is the issue:
- We only do NOT return lse if we are on cpu
- Attention sinks are detected
  - We force return lse and give a warning
  - We will encounter the torch issue in the end either way

With this PR:
- Change the warning to a proper error (as it can only happen on cpu)
- Avoid a mixture of warnings/errors

ArthurZucker (approved):
Ah yeah absolutely, MB! 
thanks

PR #41250, opened 2025-10-01
:rotating_light: Validate config attributes

zucchini-nlp:
# What does this PR do?

As per title. Continues from https://github.com/huggingface/transformers/pull/40793 and supersedes https://github.com/huggingface/transformers/pull/36534

NOTE: config classes can't accept positional args anymore! I don't think anyone would use pos args anyway but marring the PR as breaking

---

> [!NOTE]
> **High Risk**
> Refactors `PreTrainedConfig` and many model config classes to `@dataclass` + `huggingface_hub` `@strict` validation, which can change initialization/serialization behavior and reject previously-accepted configs. Also enforces save-time validation and updates defaults/deprecations (e.g., `use_return_dict`), risking backward-compatibility across model loading and downstream integrations.
> 
> **Overview**
> **Adds strict config validation.** `PreTrainedConfig` is converted to a `@dataclass` with `huggingface_hub`’s `@strict`, introduces built-in validators (architecture consistency, special token id ranges, layer type checks, `output_attentions` vs `attn_implementation`), and runs `validate()` automatically on `save_pretrained`.
> 
> **Modernizes and standardizes model configs.** Many model configuration classes are migrated from custom `__init__` logic to dataclass fields + `__post_init__`, moving compatibility logic (e.g., defaulting sub-configs, key/value casting for JSON) into post-init and adding model-specific `validate_architecture` where needed.
> 
> **API/behavior tweaks.** Deprecates `use_return_dict` in favor of `return_dict` (and updates multiple model forward paths accordingly), adjusts RoPE validation ignore-key handling, narrows AutoTokenizer fallback exception handling, and bumps the minimum `huggingface-hub` requirement to `>=1.5.0`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 07095f3575c2417e6cdf96b165cc99bae9df4601. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/transformers/models/bart/configuration_bart.py
```diff
@@ -109,67 +114,35 @@ class BartConfig(PreTrainedConfig):
...
+    activation_dropout: Optional[float] = 0.0
+    init_std: Optional[float] = 0.02
+    classifier_dropout: Optional[float] = 0.0
+    scale_embedding: Optional[bool] = False
+    use_cache: Optional[bool] = True
+    pad_token_id: Optional[int] = 1
+    bos_token_id: Optional[int] = 0
+    eos_token_id: Optional[int] = 2
+    is_encoder_decoder: Optional[bool] = True
+    decoder_start_token_id: Optional[int] = 2
```
ArthurZucker: can't wait to see this merged!

zucchini-nlp:
Blocked by https://github.com/huggingface/transformers/pull/41541#issuecomment-3632877790 for now

zucchini-nlp:
Tieme to revive this branch

zucchini-nlp:
Nice, much better and easy to maintain BC with remote code now!

src/transformers/configuration_utils.py
```diff
@@ -143,75 +149,60 @@ class PreTrainedConfig(PushToHubMixin, RotaryEmbeddingConfigMixin):
...
+        # Keys are always strings in JSON so convert ids to int here for id2label and pruned_heads
+        if self.id2label is None:
+            self._create_id_label_maps(kwargs.get("num_labels", 2))
+        else:
+            if kwargs.get("num_labels") is not None and len(self.id2label) != kwargs.get("num_labels"):
+                logger.warning(
+                    f"You passed `num_labels={kwargs.get('num_labels')}` which is incompatible to "
+                    f"the `id2label` map of length `{len(self.id2label)}`."
+                )
+            self.id2label = {int(key): value for key, value in self.id2label.items()}
```
ArthurZucker: is it a good time to get rid of these general attributes and only have them for models that actually require them?

src/transformers/configuration_utils.py
```diff
@@ -690,37 +707,40 @@ def from_dict(
...
+        # To remove arg here are those passed along for our internal telemetry but we still need to remove them
+        to_remove = ["_from_auto", "_from_pipeline"]
 
-        # We remove them from kwargs so that they do not appear in `return_unused_kwargs`.
-        config_dict["attn_implementation"] = kwargs.pop("attn_implementation", None)
-        config_dict["experts_implementation"] = kwargs.pop("experts_implementation", None)
+        # Case 1: dataclass → get dataclass fields
+        if is_dataclass(cls):
+            valid_fields = [field.name for field in fields(cls)]
+        # Case 2: for BC with remote code configs → inspect __init__ params
```
ArthurZucker: very nice to have BC on this TBH!

src/transformers/configuration_utils.py
```diff
@@ -1228,21 +1251,13 @@ def recursive_diff_dict(dict_a, dict_b, config_obj=None):
...
     "chunked_attention",
     "linear_attention",  # used in minimax
-)
-
-ALLOWED_MLP_LAYER_TYPES = (
+    "conv",  # used in LFMv2
+    "mamba",
+    "attention",
     "sparse",
     "dense",
```
ArthurZucker: I don't mind not splitting but it is a bit weird to keep ATTENTIOON_LAYER_TYPE as the name

src/transformers/models/bart/configuration_bart.py
```diff
@@ -13,13 +13,19 @@
...
+from huggingface_hub.dataclasses import strict
+
 from ...configuration_utils import PreTrainedConfig
 from ...utils import logging
 
 
 logger = logging.get_logger(__name__)
 
 
+@strict(accept_kwargs=True)
```
ArthurZucker, line 23: Here I am just wondering if: 
```
@strict(accept_kwargs=True)
@dataclass(repr=False)
class PreTrainedConfig
```
would also work (inherit the properties?

src/transformers/models/bart/configuration_bart.py
```diff
@@ -97,70 +103,38 @@ class BartConfig(PreTrainedConfig):
...
+    is_encoder_decoder: bool | None = True
+    decoder_start_token_id: int | None = 2
+    forced_eos_token_id: int | None = 2
+    is_decoder: bool | None = False
+    tie_word_embeddings: bool | None = True
+    num_labels: int | None = 3
+
+    def __post_init__(self, **kwargs):
+        self.num_hidden_layers = self.encoder_layers
+        super().__post_init__(**kwargs)
```
ArthurZucker, line 83: I think we should use attribute map instead for this

src/transformers/models/bart/configuration_bart.py
```diff
@@ -97,70 +103,38 @@ class BartConfig(PreTrainedConfig):
...
-        self.pad_token_id = pad_token_id
-        self.bos_token_id = bos_token_id
-        self.eos_token_id = eos_token_id
-        self.decoder_start_token_id = decoder_start_token_id
-        super().__init__(
-            num_labels=num_labels,
-            is_encoder_decoder=is_encoder_decoder,
-            **kwargs,
-        )
+    vocab_size: int | None = 50265
```
ArthurZucker: not really sure any of these support None

src/transformers/models/esm/configuration_esm.py
```diff
@@ -13,17 +13,93 @@
...
+        if self.sequence_state_dim != sequence_num_heads * self.sequence_head_width:
+            raise ValueError(
+                "`sequence_state_dim` should be equal to `sequence_num_heads * sequence_head_width, got"
+                f" {self.sequence_state_dim} != {sequence_num_heads} * {self.sequence_head_width}."
+            )
+        if self.pairwise_state_dim != pairwise_num_heads * self.pairwise_head_width:
+            raise ValueError(
+                "`pairwise_state_dim` should be equal to `pairwise_num_heads * pairwise_head_width, got"
+                f" {self.pairwise_state_dim} != {pairwise_num_heads} * {self.pairwise_head_width}."
+            )
```
ArthurZucker, line 129: no `validate` here?

ArthurZucker (approved):
Very very nice!

src/transformers/configuration_utils.py
zucchini-nlp: yep, I was planning to merge them all one day in any case. We also have different names for that `self.layer_types` and `self.layer_block_types` in linear attn models. Will merge them in one

src/transformers/models/bart/configuration_bart.py
zucchini-nlp: haha, you're right! I am just not sure how many tests will start complaining if we enforce `int`. Still gotta check, but indeed we should not have a None vocab

src/transformers/models/esm/configuration_esm.py
zucchini-nlp: will move

src/transformers/configuration_utils.py
```diff
@@ -252,14 +217,10 @@ def __init__(
...
-        # Attention implementation to use, if relevant (it sets it recursively on sub-configs)
+        # Attention/Experts implementation to use, if relevant (it sets it recursively on sub-configs)
+        self._output_attentions: bool | None = kwargs.pop("output_attentions", False)
         self._attn_implementation: str | None = kwargs.pop("attn_implementation", None)
-
-        # Experts implementation to use, if relevant (it sets it recursively on sub-configs)
-        self._experts_implementation = kwargs.pop("experts_implementation", None)
-
-        # Drop the transformers version info
-        self.transformers_version = kwargs.pop("transformers_version", None)
```
ArthurZucker, line 262: `self.transformers_version = kwargs.pop("transformers_version", None)` is missing

src/transformers/configuration_utils.py
```diff
@@ -375,6 +326,75 @@ def torch_dtype(self, value):
...
+        # if (
+        #     hasattr(self, "hidden_size")
+        #     and hasattr(self, "num_attention_heads")
+        #     and self.hidden_size % self.num_attention_heads != 0
+        # ):
+        #     raise ValueError(
+        #         f"The hidden size ({self.hidden_size}) is not a multiple of the number of attention "
+        #         f"heads ({self.num_attention_heads})."
+        #     )
+
```
ArthurZucker: to remove?

Wauplin:
love it 

image

src/transformers/configuration_utils.py
```diff
@@ -375,6 +326,75 @@ def torch_dtype(self, value):
...
+        if (
+            hasattr(self, "head_dim")
+            and hasattr(self, "num_heads")
+            and hasattr(self, "embed_dim")
+            and self.head_dim * self.num_heads != self.embed_dim
+        ):
+            raise ValueError(
+                f"The embed_dim ({self.embed_dim}) is not a multiple of the number of attention "
+                f"heads ({self.num_heads})."
+            )
```
ArthurZucker, line 381: we did not have that before no? for non transformers model this is not great?

zucchini-nlp:
omg, even cursor is commenting now. Oke, rebasing and should be done. Slow CI triggered on all model with YihDar and I fixed type hints to match official checkpoints

image

zucchini-nlp:
So, I really want to stop adding `@strict` or `@dataclass(repr=False)` on each model config class but seems like with Python's execution order when it comes to decorators, this is not easily done. I can't infiltrate these via `init_subclass` as I planned initially, so I had to revert it back
One options which is not 100% safe is to turn all subclasses into a `dataclass` only if the subclass has `__annotations__`. For ex if the model class has a cls attr with any type hint, it will be cast into strict dataclass. This will work 99% of cases because current configs don't type annotate anything. But  itmight break some custom code if they type annotated class attributes and their hints aren't matching their checkpoint (we never know what custom code is doing so very much possible). WDYT @ArthurZucker , worth it to reduce LOC or too risky?

src/transformers/models/afmoe/configuration_afmoe.py
```diff
@@ -64,85 +67,47 @@ class AfmoeConfig(PreTrainedConfig):
...
+    max_position_embeddings: int = 16384
+    initializer_range: float = 0.02
+    rms_norm_eps: float = 1e-5
+    use_cache: bool = True
+    tie_word_embeddings: bool = False
+    rope_parameters: RopeParameters | dict | None = None
+    num_experts: int | None = 64
+    num_experts_per_tok: int | None = 6
+    num_shared_experts: int | None = 2
+    route_scale: float | None = 10.0
```
ArthurZucker:
```suggestion
    route_scale: float | None = 1.0
```

tests/test_backbone_common.py
```diff
@@ -58,6 +58,7 @@ def test_config(self):
         self.assertEqual(config.out_indices, [0, 2])
 
         # Error raised when out_indices do not correspond to out_features
+        print("START")
```
ArthurZucker:
```suggestion
```

src/transformers/configuration_utils.py
```diff
@@ -54,6 +57,21 @@
...
+    "full_attention",
+    "sliding_attention",
+    "chunked_attention",
+    "linear_attention",  # used in minimax
+    "conv",  # used in LFMv2
+    "mamba",
+    "attention",
+    "sparse",
+    "dense",
+)
```
ArthurZucker, line 70: while we are moving could be self, but let's keep for another PR

ArthurZucker (approved):
First: huge work,  really important and reduces the LOC. Makes a lot of sense ty!!!

A few small nits! Asking code for a script that checks each and every default value will tell you in a bit what it found!

src/transformers/configuration_utils.py
zucchini-nlp: oh, it's now a ClassVar, next to `model_type, etc`  and will be handled by dataclass's own init

src/transformers/configuration_utils.py
zucchini-nlp: we had it before in several configs, not in common. From what we have, looks like if we follow consistent naming further, it will be fine

Personally I don't know many other arch where we use heads and hidden size

ArthurZucker (approved):

zucchini-nlp:
@bot /repo

tomaarsen:
I'm getting failures here due to this PR:
```python

config = AutoConfig.from_pretrained("tiny-random/glm-4v")
print(type(config))
```

```
  File "[sic]\demo_glm4v_config.py", line 4, in <module>
    config = AutoConfig.from_pretrained("tiny-random/glm-4v")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[sic]\src\transformers\models\auto\configuration_auto.py", line 1484, in from_pretrained
    return config_class.from_dict(config_dict, **unused_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[sic]\src\transformers\configuration_utils.py", line 757, in from_dict
    config = cls(**config_dict)
             ^^^^^^^^^^^^^^^^^^
  File "[sic]\transformers\Lib\site-packages\huggingface_hub\dataclasses.py", line 280, in init_with_validate
    cls.validate(self)  # type: ignore [attr-defined]
    ^^^^^^^^^^^^^^^^^^
  File "[sic]\transformers\Lib\site-packages\huggingface_hub\dataclasses.py", line 255, in validate
    validator(self)
  File "[sic]\src\transformers\modeling_rope_utils.py", line 723, in validate_rope
    validation_fn(rope_parameters, ignore_keys=self.ignore_keys_at_rope_validation)
  File "[sic]\src\transformers\modeling_rope_utils.py", line 733, in _validate_default_rope_parameters
    self._check_received_keys(rope_type, received_keys, required_keys, ignore_keys=ignore_keys)
  File "[sic]\src\transformers\modeling_rope_utils.py", line 921, in _check_received_keys
    raise KeyError(f"Missing required keys in `rope_parameters` for 'rope_type'='{rope_type}': {missing_keys}")
KeyError: "Missing required keys in `rope_parameters` for 'rope_type'='default': {'rope_theta'}"
```

Note that the remote config uses the old-style top-level rope parameters, including rope_theta: https://huggingface.co/tiny-random/glm-4v/blob/main/config.json#L28-L37

- Tom Aarsen

imstevenpmwork:
Hey @zucchini-nlp   👋 Just a quick heads-up that the changes in this PR (shipped in the 5.4.0 release) introduced a breaking change for us. Specifically, the introduction of the `@dataclass ...` decorators to the `PreTrainedConfig` class.

Over at [lerobot](https://github.com/huggingface/lerobot), we were decorating the classes that inherit from `PreTrainedConfig` with `@dataclass`, but now given that `PretrainedConfig.__init_subclass__` applies it, on the second pass the `field(init=False)` objects have already been consumed, so the fields appear as non-default arguments after the parent's default arguments. Just wanted to flag this to save some debugging time in case anyone else is suddenly staring at a [red CI](https://github.com/huggingface/lerobot/actions/runs/23653401027)!

~Maybe we can add it into the Release Notes: subclasses of `PretrainedConfig` should no longer be decorated with `dataclass`.~ -> My bad, it was already included in the release notes, thanks!

Cheers!

zucchini-nlp:
Interesting, I haven't thought that users might be already using a dataclass. TIL that we have subclasses which are dataclasses and also some that are pydantic BaseClasses

Yea, I think it is the best to remove the decorator. We've been making all subclasses a `dataclass` since it makes little sense for them to be non-dataclass objects after this PR

PR #26407, opened 2023-09-26
[`PEFT`] Fix PEFT multi adapters support

younesbelkada:
# What does this PR do?

With https://github.com/huggingface/peft/pull/905 being merged, it added the support for multi-adapter inference (combining multiple adapters at inference). 

The PR has introduced some changes that are not compatible anymore with peft integration of transformers, for example `active_adapter` is now a property method and returns a list. This PR adds the corresponding fixes to make everything compatible with the newest features of PEFT, while preserving backward compatiblity

cc @BenjaminBossan @pacman100

PR #41252, opened 2025-10-01
FIX: Bug in PEFT integration delete_adapter method

BenjaminBossan:
# What does this PR do?

The main content of this PR is to fix a bug in the `delete_adapter` method of the `PeftAdapterMixin`. Previously, it did not take into account auxiliary modules from PEFT, e.g. those added by `modules_to_save`. This PR fixes this oversight.

Note that the PR uses a new functionality from PEFT that exposes integration functions like delete_adapter. Those will be contained in the next PEFT release, 0.18.0 (yet unreleased). Therefore, the bug is only fixed when users have a PEFT version fulfilling this requirement. I ensured that with old PEFT versions, the integration still works the same as previously. The newly added test for this is skipped if the PEFT version is too low.

(Note: I tested locally with that the test will pass with PEFT 0.18.0)

While working on this, I also cleaned up the following:

- The `active_adapter` property has been deprecated for more than 2 years (#26407). It is safe to remove it now.
- There were numerous small errors or outdated pieces of information in the docstrings, which have been addressed.

## Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/src/transformers/integrations/peft.py
+++ b/src/transformers/integrations/peft.py
@@ -15,7 +15,6 @@
 import importlib
 import inspect
 import re
-import warnings
 from typing import Any, Optional, Union
 
 from packaging import version
--- a/src/transformers/integrations/peft.py
+++ b/src/transformers/integrations/peft.py
@@ -70,14 +70,9 @@ class PeftAdapterMixin:
     more details about adapters and injecting them on a transformer-based model, check out the documentation of PEFT
     library: https://huggingface.co/docs/peft/index
 
-    Currently supported PEFT methods are all non-prefix tuning methods. Below is the list of supported PEFT methods
-    that anyone can load, train and run with this mixin class:
-    - Low Rank Adapters (LoRA): https://huggingface.co/docs/peft/conceptual_guides/lora
-    - IA3: https://huggingface.co/docs/peft/conceptual_guides/ia3
-    - AdaLora: https://huggingface.co/papers/2303.10512
-
-    Other PEFT models such as prompt tuning, prompt learning are out of scope as these adapters are not "injectable"
-    into a torch module. For using these methods, please refer to the usage guide of PEFT library.
+    Currently supported PEFT methods are all non-prompt learning methods (LoRA, IA³, etc.). Other PEFT models such as
+    prompt tuning, prompt learning are out of scope as these adapters are not "injectable" into a torch module. For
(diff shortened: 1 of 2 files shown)


src/transformers/integrations/peft.py
```diff
@@ -564,34 +550,36 @@ def _dispatch_accelerate_model(
...
-        import torch
-
-        pipeline = AutoPipelineForText2Image.from_pretrained(
-            "stabilityai/stable-diffusion-xl-base-1.0", dtype=torch.float16
-        ).to("cuda")
-        pipeline.load_lora_weights(
-            "jbilcke-hf/sdxl-cinematic-1", weight_name="pytorch_lora_weights.safetensors", adapter_names="cinematic"
-        )
-        pipeline.delete_adapters("cinematic")
-        ```
```
ydshieh, line 586: we don't want to keep a doc anymore here?

src/transformers/integrations/peft.py
```diff
@@ -564,34 +550,36 @@ def _dispatch_accelerate_model(
...
+            from peft.tuners.tuners_utils import BaseTunerLayer
+
+            for module in model.modules():
+                if isinstance(module, BaseTunerLayer):
+                    if hasattr(module, "delete_adapter"):
+                        module.delete_adapter(adapter_name)
+                    else:
+                        raise ValueError(
+                            "The version of PEFT you are using is not compatible, please use a version that is greater than 0.6.1"
+                        )
```
ydshieh, line 582: I'm ok to have this as it is. But just wondering if the old version is not working correctly, should we at least warn the user, or we want to be silent about that?

ydshieh (approved):
Thanks !

src/transformers/integrations/peft.py
BenjaminBossan: The doc was unfitting, as it used diffusers, not transformers. Moreover, the API is so simple that it really doesn't need an example.

BenjaminBossan (commented):
> But just wondering if the old version is not working correctly, should we at least warn the user, or we want to be silent about that?

Good idea, I now detect this behavior and issue a warning. I added a test for the warning as well. Locally, I also ran the test with PEFT 0.18.0 to ensure it still passes once that version is released.

BenjaminBossan:
@ydshieh Thanks for the review, the comments should be addressed. Feel free to merge or let me know if I should merge.

PR #41253, opened 2025-10-01
transformers serve quantization docs + some api fixes for bitsandbytes

SunMarc:
# What does this PR do?

This PR adds some docs for transformers serve x quantization and fixes the quantization api for bitsandbytes. 
Following PR will add more support for other methods and I will add a section in each quantization method to show what is the preferred setup (compile ? dtype ?) when using serve !


SunMarc:
let's merge this first and i'll do follow up pr

docs/source/en/serving.md
```diff
@@ -383,6 +383,30 @@ transformers serve \
...
+Make sure to install the required libraries listed in the quantization documentation.
+
+> [!TIP]
+> Pre-quantized models generally provide the best balance of performance and accuracy.
+
+#### On the fly quantization
+
+If you want to quantize a model at runtime, you can specify the --quantization flag in the CLI. Note that not all quantization methods support on-the-fly conversion. The full list of supported methods is available in the quantization [overview](https://huggingface.co/docs/transformers/main/quantization/overview). 
+
+Currently, with transformers serve, we only supports the some methods: ["bitsandbytes-4bit", "bitsandbytes-8bit"]
```
MekkCyber:
```suggestion
Currently, with transformers serve, we only supports some methods: ["bitsandbytes-4bit", "bitsandbytes-8bit"]
```

src/transformers/cli/serve.py
```diff
@@ -1688,19 +1681,14 @@ def get_quantization_config(self) -> Optional["BitsAndBytesConfig"]:
...
-                bnb_4bit_compute_dtype=self.dtype,
-                bnb_4bit_quant_type=self.bnb_4bit_quant_type,
-                bnb_4bit_use_double_quant=self.use_bnb_nested_quant,
-                bnb_4bit_quant_storage=self.dtype,
-            )
-        elif self.load_in_8bit:
-            quantization_config = BitsAndBytesConfig(
-                load_in_8bit=True,
+                bnb_4bit_quant_type="nf4",
+                bnb_4bit_use_double_quant=True,
```
MekkCyber, line 1688: tiny nit : maybe we can add a `logger.info` here to specify that we are using nested quantization, and `nf4` types by default

MekkCyber (approved):
Sounds good! Thanks

src/transformers/cli/serve.py
SunMarc: I think it will be easier to just print the config

PR #41256, opened 2025-10-01
path validation for security reason

ydshieh:
# What does this PR do?

See the comment in file change tab

### DIFF
--- a/src/transformers/convert_slow_tokenizers_checkpoints_to_fast.py
+++ b/src/transformers/convert_slow_tokenizers_checkpoints_to_fast.py
@@ -15,6 +15,7 @@
 
 import argparse
 import os
+from pathlib import Path
 
 import transformers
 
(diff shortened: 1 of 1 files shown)


src/transformers/convert_slow_tokenizers_checkpoints_to_fast.py
```diff
@@ -69,6 +70,15 @@ def convert_slow_checkpoint_to_fast(tokenizer_name, checkpoint_name, dump_path,
             if "/" in checkpoint:
                 checkpoint_directory, checkpoint_prefix_name = checkpoint.split("/")
                 dump_path_full = os.path.join(dump_path, checkpoint_directory)
+
+                # Security check
+                try:
+                    Path(dump_path_full).resolve().relative_to(Path(dump_path).resolve())
```
ydshieh, line 76: if `checkpoint` starts with `..`, the path `dump_path_full` will be the parent of the user specified `dump_path`, which is not good itself, and potentially a security issue (although unlikely due to the usage of this script).

ydshieh: this is reported on huntr

ArthurZucker (approved):

PR #41261, opened 2025-10-01
Remove `test_initialization`

Cyrilvallez:
# What does this PR do?

As per the title. The test is very old and not useful anymore.

We have the much more recent and accurate `test_can_init_all_missing_weights` I added some time ago to actually test that all weights are correctly taken into account in `_init_weights`.

What distribution is then used on those weights is not really important to be tested, and way too hard for a general test as it can be anything. This can be seen by the fact that the test is basically skipped/overwritten for more than half the models in the lib, which is a *strong* sign that the test is quite badly designed to start with 🙃

cc @ArthurZucker @gante

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -54,7 +54,6 @@ NOT_DEVICE_TESTS = {
     "test_gradient_checkpointing_backward_compatibility",
     "test_gradient_checkpointing_enable_disable",
     "test_torch_save_load",
-    "test_initialization",
     "test_forward_signature",
     "test_model_get_set_embeddings",
     "test_model_main_input_name",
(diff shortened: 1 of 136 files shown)


gante (approved):
LGTM 👍 

Agreed with the removal:
1. indeed, if it's frequently overwritten, then it's a bad test
2. `test_can_init_all_missing_weights` looks superior, it checks that random parameters can be seeded and missing data is randomly initialized as needed
3. `test_initialization` checks that all parameters are in the `[0, 1]` range, which seems overly restrictive

Suggestions (this PR or following PRs):
1. `test_can_init_all_missing_weights` is also skipped on many models, it would be cool to alleviate the skip constraint (`addition_year`)
2. We now have no checks for the initialization values themselves. Some sanity checks would be nice (infs, nans, and even perhaps values outside large ranges like [-1e3, 1e3])?

Cyrilvallez:
For sure, I'm planning on going back and fixing more older models so that `test_can_init_all_missing_weights` would be run on all models!

ydshieh:
No failure of `test_can_init_all_missing_weights` since Jul 29th , very nice 👍

PR #41263, opened 2025-10-01
Remove some previous team members from allow list of triggering Github Actions

ydshieh:
# What does this PR do?

Better security practice.


gante (approved):
Thank you for the update 🤗 

(I think we can also remove `muellerzr`?)

PR #41264, opened 2025-10-01
Use regex defailed flags

cyyever:
# What does this PR do?

This PR uses full names of regex flags for better code readability.


ydshieh (approved):
Readability is indeed important!

PR #41265, opened 2025-10-01
:rotating_light: [`Attention Masks`] Bidirectional masks for encoder and encoder-decoder models

vasqu:
Introduces attention masks for encoder and encoder-decoder models, aka full bidirectional masks, fully compatible with our interface system. Some exceptions, like autoformer, have been ignored for now + there are probably a lot more models that have some weird update patterns but these are imo the most important (related) models here, i.e. Bert and Bart models.

Breaking changes:
- Executorch now also needs a mask workaround due to vmapping in our general interface
- 3D masks are no longer supported (looking at you grounding dino models :eye:) - they just needed to expand to the head dim as well...
- Torch fx will no longer work for these models, ig vmapping again (?)


vasqu:
run-slow: bert,bart,llama

vasqu:
run-slow: bert,bart,llama

vasqu:
run-slow: bert

vasqu:
run-slow: mm_grounding_dino

vasqu:
run-slow: mm_grounding_dino, grounding_dino

src/transformers/integrations/executorch.py
ArthurZucker, line 1: very nice!

src/transformers/masking_utils.py
```diff
@@ -836,6 +876,87 @@ def create_causal_mask(
...
+        allow_is_bidirectional_skip = False
+    if and_mask_function is not None:
+        if not _is_torch_greater_or_equal_than_2_6:
+            raise ValueError("Using `or_mask_function` or `and_mask_function` arguments require torch>=2.6")
+        mask_factory_function = and_masks(mask_factory_function, and_mask_function)
+        allow_is_bidirectional_skip = False
+
+    # Due to the logic surrounding `cache_position` in inferring query-related information, we
+    # construct a dummy tensor imitating initial positions
+    cache_position = torch.arange(input_embeds.shape[1], device=input_embeds.device, dtype=torch.long)
```
ArthurZucker: should work with cudagraph? but let's add a test to make sure please !

src/transformers/masking_utils.py
```diff
@@ -836,6 +876,87 @@ def create_causal_mask(
...
+    embeds = encoder_hidden_states if encoder_hidden_states is not None else input_embeds
+    # We ignore a few irrelevant arguments at the end as we do not have a (growing) cache here
+    early_exit, attention_mask, _, kv_length, kv_offset = _preprocess_mask_arguments(
+        config, embeds, attention_mask, None, None, None, 0
+    )
+    if early_exit:
+        return attention_mask
+
+    batch_size, dtype = embeds.shape[0], embeds.dtype
+    mask_factory_function = bidirectional_mask_function
```
ArthurZucker, line 923: i would remove this line probably

ArthurZucker: making it explicit that we are using bidrectionnal mask function

src/transformers/masking_utils.py
```diff
@@ -262,6 +269,21 @@ def _ignore_causal_mask_sdpa(
     return False
 
 
+def _ignore_bidirectional_mask_sdpa(padding_mask: Optional[torch.Tensor]) -> bool:
+    """
+    Detects whether the bidirectional mask can be ignored in case PyTorch's SDPA is used, i.e. when there is full
+    attention with no padding.
+    """
```
ArthurZucker, line 276: i am guessing this is quite general for non padding cases!

ArthurZucker (commented):
Mega nice! I would be paranoiac and add an extra test / a masking utils test for it !

src/transformers/masking_utils.py
```diff
@@ -720,7 +760,7 @@ def _preprocess_mask_arguments(
 
     # Move the mask to correct device, and potentially switch dtype for efficiency
     if attention_mask is not None and attention_mask.ndim == 2:
-        attention_mask = attention_mask.to(device=cache_position.device, dtype=torch.bool)
+        attention_mask = attention_mask.to(device=input_embeds.device, dtype=torch.bool)
```
Cyrilvallez: Why did you change which device we use here?
Also see my last comment

src/transformers/masking_utils.py
```diff
@@ -78,6 +78,13 @@ def causal_mask_function(batch_idx: int, head_idx: int, q_idx: int, kv_idx: int)
     return kv_idx <= q_idx
 
 
+def bidirectional_mask_function(batch_idx: int, head_idx: int, q_idx: int, kv_idx: int) -> bool:
+    """
+    This creates a full bidirectional mask.
+    """
+    return q_idx >= 0
```
Cyrilvallez:
```suggestion
def bidirectional_mask_function(batch_idx: int, head_idx: int, q_idx: int, kv_idx: int) -> bool:
    """
    This creates a full bidirectional mask.
    """
    return q_idx.new_ones((), dtype=torch.bool)
```

This would be more explicit IMO

src/transformers/integrations/executorch.py
```diff
@@ -1199,3 +1295,79 @@ def sdpa_mask_without_vmap(
...
+    # but without data-dependent slicing (i.e. torch.compile friendly)
+    kv_arange = torch.arange(kv_length, device=cache_position.device)[None, ...]
+    kv_arange += kv_offset
+
+    # Simplest and most efficient way to obtain a bidirectional mask
+    bidirectional_mask = kv_arange >= 0
+
+    bidirectional_mask = bidirectional_mask[None, None, :, :].expand(batch_size, -1, -1, -1)
+    if padding_mask is not None:
+        bidirectional_mask = bidirectional_mask * padding_mask[:, None, None, :]
```
Cyrilvallez: All this is simply the padding mask 👀 (or a fresh mask of True if None, but then it should be skipped)

src/transformers/masking_utils.py
Cyrilvallez: If we need to construct them anyway, let's do it at the beginning and revert changes to the device used in! Also in the comments and docs!

Cyrilvallez (commented):
Nice! Just left a few comments to simplify!

src/transformers/masking_utils.py
vasqu: Hmm, I thought we keep the flexibility with and/or masks which is why I need this line explicitly, i.e. https://github.com/huggingface/transformers/blob/a99ed6b49592f51a7e92342b46fc4e9b3878394f/src/transformers/masking_utils.py#L929-L941

Would like to keep things somewhat aligned with the causal version

src/transformers/masking_utils.py
vasqu: Good point, I moved this and reverted a bunch of unnecessary logic (e.g. your first comment on device). I think I wanted to avoid using cache positions completely but, at the end, I couldn't.

Should now touch the existing function way less

src/transformers/masking_utils.py
vasqu: See https://github.com/huggingface/transformers/pull/41265#discussion_r2406955793 

Now reverted and we shouldn't have changes in the core code/docs

src/transformers/integrations/executorch.py
vasqu: I was so free to simplify this a lot more, lmk if that's better. But yes, you're right in essence it's the padding mask only.

src/transformers/masking_utils.py
vasqu: Added a test at https://github.com/huggingface/transformers/blob/911752d0257bec006a75b3818f96fc39d8f973aa/tests/utils/test_masking_utils.py#L266

Cyrilvallez (approved):
Nice, thanks for reverting unnecessary changes! I'm now fully on board with the new mask function implementation.

From what I can see, we don't need to update the general `create_masks_for_generate` because bidirectional masks were never prepared in advance, and models loosing their `_prepare_4d_causal_attention_mask_with_cache_position` (thus now relying on `create_masks_for_generate`) have the corresponding `create_causal_mask` call, meaning only the causal one used to be prepared in advance. Could you just confirm that before merging?

ArthurZucker (approved):
Very very nice!

vasqu:
The current logic during generate is to preinfer the encoder hidden states already so that mask is only used once either way. I think the cross mask is always created on the fly (at least that is currently the case even before the PR). 

So yes it's as you said @Cyrilvallez, only the causal mask is prepared and we do not need to update anything there - it automatically switches to the new way when we do not have `_prepare_4d_causal_attention_mask_with_cache_position` in the file. This also happens in `Dia` for example ;)

PR #41267, opened 2025-10-01
Remove SageMakerTrainer

cyyever:
# What does this PR do?

It is replaced by Trainer.


tests/sagemaker/scripts/pytorch/run_glue_model_parallelism.py
ydshieh, line 1: It seems,  after the changes in this PR, this test file doesn't contain any sagemaker stuff, so unclear to me if this file still makes sense.

ydshieh: Let's delete this file instead.

ydshieh (approved):
👍

tests/sagemaker/scripts/pytorch/run_glue_model_parallelism.py
cyyever: Wait, after examining the `tests/sagemaker` folder, we can remove more...

cyyever:
@ydshieh sagemaker is still supported, although not in its own Trainer. Unless we feel that the test code is outdated, `tests/sagemaker/scripts/pytorch/run_glue_model_parallelism.py` can' be deleted (it is invoked by other scripts of `tests/sagemaker/scripts`, deleting it and all its invokers means eventually we have to delete the whole folder..).

But I have deleted `tensorflow` related stuff...

ydshieh:
Thanks for pointing out. Let's keep it then 👍

PR #41268, opened 2025-10-01
🚨 Bump to Python 3.10 and rework how we check 3rd-party libraries existence

Cyrilvallez:
# What does this PR do?

Python 3.9 is officially EOL starting this month, so let's bump it! Took the opportunity of the new features available in 3.10 to simplify quite a bit our library checks

Note that it is breaking BC a bit as I removed some old and unused library checks that were public members


ydshieh (approved):
Nice.

Let's wait python 3.9 reaches EOL and merge.

(or if v5 is released before that date, we can also drop the support of 3.9 at the same day)

Cyrilvallez:
Yeah, the exact date was not super clear to me @ydshieh 🥲, do you have more infos from experience? Official website only says October, without a given day, so I guess it's already EOL? Or at the end of the month? (but very unlikely they would have a last release in the last month anyway I guess)

ydshieh:
> do you have more infos from experience?

No. From [previous python versions](https://devguide.python.org/versions/#unsupported-versions), it is definitely not something predictable (i.e. not always the end of the month):


```
3.8 2024-10-07
3.7 2023-06-27
3.6 2021-12-23
3.5 2020-09-30
```

src/transformers/utils/import_utils.py
```diff
@@ -117,217 +82,54 @@ def _is_package_available(pkg_name: str, return_version: bool = False) -> Union[
...
-_ipex_available, _ipex_version = _is_package_available("intel_extension_for_pytorch", return_version=True)
-_jinja_available = _is_package_available("jinja2")
-_kenlm_available = _is_package_available("kenlm")
-_keras_nlp_available = _is_package_available("keras_nlp")
-_levenshtein_available = _is_package_available("Levenshtein")
-_librosa_available = _is_package_available("librosa")
-_natten_available = _is_package_available("natten")
-_nltk_available = _is_package_available("nltk")
-_onnx_available = _is_package_available("onnx")
-_openai_available = _is_package_available("openai")
```
ArthurZucker, line 172: ehh ahah

ArthurZucker (approved):
not sure lru cache is required no?

Cyrilvallez:
Not _strictly_ required, but as the output of those functions cannot change anyway, it's an easy and nice way to cache the result without having to define 10k global variables as was done previously. About one third/half of them had it previously already
It's basically a tiny performance optimization

Cyrilvallez:
Merging as we released last v4 version already!

lhoestq:
(note that `transformers@main` can still be installed on 3.9 since it's not completely dropped in setup.py and raises `AttributeError: module 'importlib.metadata' has no attribute 'packages_distributions'`)

Cyrilvallez:
Indeed, I forgot it! It was fixed in https://github.com/huggingface/transformers/pull/41410 already, thanks for checking-in!

PR #41272, opened 2025-10-01
feat: Add HRM Model

zbloss:
# What does this PR do?

Adds the new Hierarchical Reasoning Model.

Fixes #41271

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker @Cyrilvallez hoping you can take a look. The model is currently implemented as a text model though further research hints it may be helpful for understanding images as well.


ArthurZucker:
Nice! cc @molbap for an initial review!

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,972 @@
...
+            z = (b - a) / 2
+            c = (2 * math.pi) ** -0.5
+            pdf_u = c * math.exp(-0.5 * lower**2)
+            pdf_l = c * math.exp(-0.5 * upper**2)
+            comp_std = std / math.sqrt(1 - (upper * pdf_u - lower * pdf_l) / z - ((pdf_u - pdf_l) / z) ** 2)
+            tensor.uniform_(a, b)
+            tensor.erfinv_()
+            tensor.mul_(sqrt2 * comp_std)
+            tensor.clip_(lower * comp_std, upper * comp_std)
+    return tensor
```
molbap, line 373: specific initialization scheme here should be handled through a method of type `_init_weights_` to be handled correctly by `from_pretrained` and so on, see https://github.com/huggingface/transformers/blob/5abfa43f0252f7e069c2de99fe940138eb33600f/src/transformers/models/rwkv/modeling_rwkv.py#L390-L463

docs/source/en/model_doc/hrm.md
molbap, line 1: Very comprehensive documentation, thanks :raised_hands: !

One needed file that would bring immense value to the PR would be a conversion script, converting the checkpoint contained here https://huggingface.co/sapientinc/HRM-checkpoint-ARC-2/tree/main into a safentors-ingestable version such that one can simply run `AutoModel.from_pretrained('path/to/converted/hrm') and verify the results/run the model

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,439 @@
...
+            num_attention_heads=8,
+            max_position_embeddings=81,
+            h_cycles=2,
+            l_cycles=2,
+            halt_max_steps=8,
+            dtype="float32",
+        )
+
+        model = HrmForCausalLM(config).to(torch_device)
+        model.eval()
```
molbap, line 417: Related to previous comment - the `IntegrationTest` here would have to load, instead of a random config, the converted already learned weights, so we can have some form of a ground truth associated with the commit adding this model.

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,972 @@
...
+
+        return HrmCausalLMOutput(
+            loss=loss,
+            logits=logits,
+            q_halt_logits=q_halt_logits,
+            q_continue_logits=q_continue_logits,
+            carry=carry_out,
+            hidden_states=hidden_states,
+            attentions=attentions,
+        )
```
molbap, line 1074: The `carry` is very similar to the `state` carried for instance in `rwkv`, see here https://github.com/huggingface/transformers/blob/5abfa43f0252f7e069c2de99fe940138eb33600f/src/transformers/models/rwkv/modeling_rwkv.py#L689-L800. I think in this instance it should be able to be [modularisable](https://huggingface.co/docs/transformers/v4.45.1/en/modular_transformers), and a conversion script would suffice to handle mismatching keys

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,972 @@
+# coding=utf-8
+# Copyright 2025 The HRM Team and The HuggingFace Inc. team. All rights reserved.
```
molbap: nit, would be more Sapient inc team no?

molbap (commented):
Done a shallow first pass to leave some pointers - feel free to ask questions back and congrats for this endeavour! Exciting!

src/transformers/models/hrm/modular_hrm.py
zbloss: Yep good call, I shorthanded this

zbloss:
@molbap I believe I fixed everything, let me know when you get a chance to take another look.

I converted the author's original weights and uploaded them here after verifying that they work: 
* https://huggingface.co/zbloss/HRM-maze-30x30-hard
* https://huggingface.co/zbloss/HRM-sudoku-extreme

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,474 @@
+# Copyright 2025 The HRM Team and HuggingFace Inc. team. All rights reserved.
```
molbap: Should be Sapient Inc too (in all license comments)

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,474 @@
...
+
+            # Verify carry state is present
+            self.assertIsNotNone(outputs.carry)
+
+            # Verify Q-values for ACT mechanism
+            self.assertIsNotNone(outputs.q_halt_logits)
+            self.assertIsNotNone(outputs.q_continue_logits)
+
+        except Exception as e:
+            self.skipTest(f"Pre-trained checkpoint not available: {e}")
```
molbap: we should not add a try/catch here, and always test
```suggestion
            model = HrmForCausalLM.from_pretrained("zbloss/HRM-sudoku-extreme")
            model.eval()

            # Test model can perform inference
            # Create a simple input for ARC-like reasoning task (30x30 grid max)
            input_ids = torch.randint(0, 11, (1, 81), device=torch_device)
            model = model.to(torch_device)

            with torch.no_grad():
                outputs = model(input_ids=input_ids)

            # Verify outputs have expected structure
            self.assertIsNotNone(outputs)
            self.assertIsNotNone(outputs.logits)
            self.assertEqual(outputs.logits.shape[0], 1)  # batch size
            self.assertEqual(outputs.logits.shape[1], 81)  # sequence length
            self.assertEqual(outputs.logits.shape[2], 11)  # vocab size

            # Verify carry state is present
            self.assertIsNotNone(outputs.carry)

            # Verify Q-values for ACT mechanism
            self.assertIsNotNone(outputs.q_halt_logits)
            self.assertIsNotNone(outputs.q_continue_logits)


```

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,474 @@
...
+    """Integration tests for HRM with real use cases."""
+
+    @slow
+    @require_torch
+    def test_pretrained_model_loading(self):
+        """Test loading pre-trained HRM checkpoint from HuggingFace Hub."""
+        # This test requires the converted checkpoint to be available
+        # Skip if the model is not yet available on the Hub
+        try:
+            model = HrmForCausalLM.from_pretrained("zbloss/HRM-sudoku-extreme")
```
molbap: Are these weights converted from the original ones? If so, the script used to convert them should be also committed, under `hrm/convert_original_hrm_weights_to_hf.py`, like most models where the original weights are not loadable

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,474 @@
...
+
+        # Create a simple input (e.g., partial Sudoku-like pattern)
+        input_ids = torch.randint(0, 11, (2, 81), device=torch_device)
+
+        with torch.no_grad():
+            outputs = model(input_ids=input_ids)
+
+        self.assertIsNotNone(outputs)
+        self.assertIsNotNone(outputs.logits)
+        self.assertEqual(outputs.logits.shape, (2, 81, 11))
```
molbap: This test does not test "simple reasoning", it just asserts that some outputs are not None

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,474 @@
...
+            self.assertEqual(outputs.logits.shape[0], 1)  # batch size
+            self.assertEqual(outputs.logits.shape[1], 81)  # sequence length
+            self.assertEqual(outputs.logits.shape[2], 11)  # vocab size
+
+            # Verify carry state is present
+            self.assertIsNotNone(outputs.carry)
+
+            # Verify Q-values for ACT mechanism
+            self.assertIsNotNone(outputs.q_halt_logits)
+            self.assertIsNotNone(outputs.q_continue_logits)
```
molbap: Same issue, here; the test does not actually give a verifiable execution (we do not compare to actual values)

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,474 @@
...
+        """Skip cache test - HRM uses carry state instead of KV cache."""
+        self.skipTest("HRM uses carry state instead of standard KV cache")
+
+    def test_contrastive_generate_dict_outputs_use_cache(self):
+        """Skip cache test - HRM uses carry state instead of KV cache."""
+        self.skipTest("HRM uses carry state instead of standard KV cache")
+
+    def test_sample_generate_dict_outputs_use_cache(self):
+        """Skip cache test - HRM uses carry state instead of KV cache."""
+        self.skipTest("HRM uses carry state instead of standard KV cache")
```
molbap: mamba-like models have no PKV either, they have a flag inside to skip these tests as a whole!

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,984 @@
...
+                if batch_val.shape != carry_val.shape:
+                    # During generation, batch may have different seq length
+                    # Pad or truncate to match carry's sequence length
+                    if batch_val.shape[1] < carry_val.shape[1]:
+                        # Pad batch to carry size
+                        pad_size = carry_val.shape[1] - batch_val.shape[1]
+                        batch_val = F.pad(batch_val, (0, pad_size), value=0)
+                    elif batch_val.shape[1] > carry_val.shape[1]:
+                        # Truncate batch to carry size (take last tokens)
+                        batch_val = batch_val[:, -carry_val.shape[1] :]
```
molbap: Is this truncation expected? Won't it cause issues?

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,984 @@
...
+            carry = self.initial_carry(batch)
+
+        # Update data and carry
+        new_inner_carry = self.inner.reset_carry(carry.halted, carry.inner_carry)
+        new_steps = torch.where(carry.halted, torch.zeros_like(carry.steps), carry.steps)
+
+        # Handle size mismatch during generation - only update keys that exist in both batch and carry
+        new_current_data = {}
+        for k in carry.current_data.keys():
+            if k in batch:
```
molbap: ops in these loops should be vectorizable using torch scatter ops overall

src/transformers/models/hrm/modular_hrm.py
molbap, line 1: This file is not yet a `modular` file in the sense that it does not inherit from the other models I pointed out in last review, have you checked them out? Did you encounter issues?

molbap (commented):
Thanks for iterating! left additional comments

src/transformers/models/hrm/modular_hrm.py
zbloss: Good catch, I missed this issue in the original codebase. This would cause an issue and should actually fail if. Adding a raise ValueError to catch this

src/transformers/models/hrm/modular_hrm.py
zbloss: I did find a few modules I could reuse, but am adding comments and details in the docstrings addressing why certain modules behave differently

src/transformers/models/hrm/modular_hrm.py
zbloss: Good catch, fixing

tests/models/hrm/test_modeling_hrm.py
zbloss: Yeah I agree, the naming of the test is completely misleading

tests/models/hrm/test_modeling_hrm.py
zbloss: Yes, adding the conversion script

tests/models/hrm/test_modeling_hrm.py
zbloss: Good catch

molbap:
Reviewing now!

docs/source/en/model_doc/hrm.md
```diff
@@ -0,0 +1,309 @@
...
+### Using Puzzle-Specific Embeddings
+
+```python
+config = HrmConfig(
+    vocab_size=11,
+    hidden_size=512,
+    max_position_embeddings=81,
+    puzzle_emb_ndim=512,  # Enable puzzle embeddings
+    num_puzzle_identifiers=100,  # Number of unique puzzles
+    h_cycles=2,
```
molbap: naming (apply everywhere): instead of single-letters `h` and `l` for readability, `high` and `low` would be preferrable

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1040 @@
...
+        self.model.set_input_embeddings(value)
+
+    def get_output_embeddings(self):
+        return None  # HRM doesn't tie output embeddings
+
+    def set_output_embeddings(self, new_embeddings):
+        pass  # HRM doesn't tie output embeddings
+
+    def prepare_inputs_for_generation(
+        self, input_ids, carry=None, past_key_values=None, attention_mask=None, **kwargs
```
molbap: no need for these I think
```suggestion
        self, input_ids, carry=None, **kwargs
```

molbap: also nit, `state` is a more self-explanatory name than `carry`

src/transformers/models/hrm/modular_hrm.py
molbap: This is not resolved: this should still be called through an `_init_weights_` function, look in the codebase for similar patterns please!

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1040 @@
...
+            elif batch_val.shape[1] > carry_val.shape[1]:
+                # This should not happen in normal usage - carry is initialized from batch,
+                # so they should match initially. During generation, batch gets smaller (1 token),
+                # never larger. If this occurs, it indicates incorrect carry state reuse.
+                raise ValueError(
+                    f"Unexpected shape mismatch: batch sequence length ({batch_val.shape[1]}) "
+                    f"is greater than carry sequence length ({carry_val.shape[1]}) for key '{k}'. "
+                    f"This suggests carry state is being reused incorrectly across different input sizes. "
+                    f"Please initialize a new carry state for inputs with different sequence lengths."
+                )
```
molbap: I think this should go, if it's a corner case we want to avoid, it should be by construction

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1040 @@
...
+        model_inputs.update(filtered_kwargs)
+        return model_inputs
+
+    def _update_model_kwargs_for_generation(self, outputs, model_kwargs, **kwargs):
+        """Update model kwargs for next generation step."""
+        # Update carry for next generation step
+        model_kwargs["carry"] = outputs.carry
+        # Remove attention_mask as HRM doesn't use it
+        model_kwargs.pop("attention_mask", None)
+        return model_kwargs
```
molbap, line 1017: Same, we can get rid of these by fixing the signatures of the `forward` and the `prepare`

conftest.py
molbap, line 1: It's weird that this pops up - `make fixup` on latest `pip install -e .[quality]` after a rebase on main should remove it

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,432 @@
...
+        model.eval()
+
+        input_ids = torch.tensor([[1, 2, 3]], device=torch_device)
+
+        with torch.no_grad():
+            generated = model.generate(input_ids, max_new_tokens=5)
+
+        self.assertIsNotNone(generated)
+        self.assertEqual(generated.shape[0], 1)
+        self.assertGreater(generated.shape[1], input_ids.shape[1])
```
molbap: This does not verify that the model performs as expected.

 It's simply checking that the model has generated something that is not None. It loads the checkpoint, that's good, but it should also verify e.g. on a single puzzle that it does manage to output the correct answer.

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,432 @@
...
+            config.h_cycles = h_cycles
+            config.l_cycles = l_cycles
+            model = HrmModel(config).to(torch_device)
+            model.eval()
+
+            with torch.no_grad():
+                outputs = model(input_ids=input_ids)
+
+            self.assertIsNotNone(outputs)
+            self.assertIsNotNone(outputs.logits)
```
molbap, line 342: same comment, this just tests that the outputs are not `None`, it doesn't bring any insight on how the model reasons.

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1040 @@
...
+            hidden_states = outputs.hidden_states
+            attentions = outputs.attentions
+        else:
+            # When return_dict=False, outputs is a tuple
+            logits = outputs[0]
+            q_halt_logits = outputs[1] if len(outputs) > 1 else None
+            q_continue_logits = outputs[2] if len(outputs) > 2 else None
+            carry_out = outputs[3] if len(outputs) > 3 else None
+            hidden_states = outputs[4] if len(outputs) > 4 else None
+            attentions = outputs[5] if len(outputs) > 5 else None
```
molbap: let's avoid these patterns, they are highly unusual (and not needed)

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1040 @@
...
+                for _L_step in range(self.config.l_cycles):
+                    if not ((_H_step == self.config.h_cycles - 1) and (_L_step == self.config.l_cycles - 1)):
+                        z_L = self.L_level(z_L, z_H + input_embeddings, cos_sin=cos_sin)
+                if _H_step != self.config.h_cycles - 1:
+                    z_H = self.H_level(z_H, z_L, cos_sin=cos_sin)
+
+        # 1-step grad
+        z_L = self.L_level(z_L, z_H + input_embeddings, cos_sin=cos_sin)
+        z_H = self.H_level(z_H, z_L, cos_sin=cos_sin)
+
```
molbap: Thanks for implementing this - seems ok, would benefit massively from clearer variable naming. We can spare the characters and get some slightly more explanatory names for later code readers who are unfamiliar with the logic of that particular model

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1040 @@
...
+            pad_count = self.puzzle_emb_len * self.config.hidden_size - puzzle_embedding.shape[-1]
+            if pad_count > 0:
+                puzzle_embedding = F.pad(puzzle_embedding, (0, pad_count))
+            embedding = torch.cat(
+                (puzzle_embedding.view(-1, self.puzzle_emb_len, self.config.hidden_size), embedding), dim=-2
+            )
+
+        # Position embeddings
+        if self.config.pos_encodings == "learned":
+            embedding = 0.707106781 * (embedding + self.embed_pos.weight.to(self.forward_dtype))
```
molbap: no hardcoded values

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1040 @@
...
+        self.register_buffer(
+            "L_init",
+            truncated_normal_init_(torch.empty(config.hidden_size, dtype=self.forward_dtype), std=1),
+            persistent=True,
+        )
+
+        # Q head special init
+        with torch.no_grad():
+            self.q_head.weight.zero_()
+            self.q_head.bias.fill_(-5)
```
molbap: this should be in an init_weights as well

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1040 @@
...
+    return tensor
+
+
+def rms_norm(hidden_states: torch.Tensor, variance_epsilon: float) -> torch.Tensor:
+    """Root Mean Square Layer Normalization."""
+    input_dtype = hidden_states.dtype
+    hidden_states = hidden_states.to(torch.float32)
+    variance = hidden_states.square().mean(-1, keepdim=True)
+    hidden_states = hidden_states * torch.rsqrt(variance + variance_epsilon)
+    return hidden_states.to(input_dtype)
```
molbap: `rms_forward` in `modeling_falcon_mamba` is identical AFAIU

molbap (commented):
Left several comments - one very important is that currently there is no way to _actually_ test the model, and that's a bigger blocker than the rest. When you have an end-to-end test (using the weights you converted) we'll be able to move a bit faster.

The rest is more about being inline with the rest of the library. 
- Variable names should be explicit.
- we should avoid if/else codepaths in the modeling. For instance for rope, it should be clearly defined in one place.
- for init_weights, I understand the truncated normal init schema is needed, but it should be interfaced in a familiar way (especially for an "exotic" model with many new components, "usual" components have an even bigger motivation to stay, well, usual.

Finally, please let's only resolve the conversations when a reviewer had the time to check it or it's really obvious, else it makes it hard for us to track iterations :pray: 

Thanks and let me know if you face issues!

molbap:
Hey @zbloss, reviewing now!

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,704 @@
...
+        with torch.no_grad():
+            # Initialize state
+            state = model.model.initial_state({"input_ids": partial_puzzle})
+
+            # Run through multiple reasoning steps (ACT mechanism)
+            for _ in range(model.config.halt_max_steps):
+                outputs = model(input_ids=partial_puzzle, state=state)
+                state = outputs.carry
+
+                # Check if halted
```
molbap: these comments are not needed, as the code they comment is self-explanatory
```suggestion
```

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,704 @@
...
+        model.eval()
+
+        input_ids = torch.tensor([[1, 2, 3]], device=torch_device)
+
+        with torch.no_grad():
+            generated = model.generate(input_ids, max_new_tokens=5)
+
+        self.assertIsNotNone(generated)
+        self.assertEqual(generated.shape[0], 1)
+        self.assertGreater(generated.shape[1], input_ids.shape[1])
```
molbap: This does not test generation capabilities, it just tests that "something" is generated - it's not an integration test

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,704 @@
...
+
+
+if __name__ == "__main__":
+    unittest.main()
```
molbap:
```suggestion
```

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1017 @@
...
+hierarchical and multi-timescale processing with Adaptive Computation Time (ACT), achieving strong performance
+on tasks like Sudoku solving, maze navigation, and abstract reasoning with minimal training data.
+
+This implementation follows the architecture described in the paper: https://arxiv.org/abs/2506.21734
+
+## Modular Design Notes
+
+HRM follows the modular transformers pattern by inheriting reusable components from Llama where appropriate:
+- `apply_rotary_pos_emb`, `rotate_half`: Inherited from Llama (identical implementations)
+- `LlamaRotaryEmbedding`: Available for import but not used (see HrmRotaryEmbedding below)
```
molbap: nit, we don't inherit from llama in particular, just from whatever closest architecture there is

Overall we can remove the "modular design notes" as they are not specific to this model

docs/source/en/model_doc/hrm.md
```diff
@@ -0,0 +1,309 @@
...
+- Rotary Position Embeddings (RoPE) or learned positional encodings
+
+### Forward Pass
+
+```
+for h_cycle in range(H_cycles):
+    for l_cycle in range(L_cycles):
+        z_L = L_level(z_L, z_H + input_embeddings)
+    z_H = H_level(z_H, z_L)
+```
```
molbap, line 75: Some explanations around this lone code block would be nice :hugs:

src/transformers/models/hrm/configuration_hrm.py
```diff
@@ -0,0 +1,179 @@
...
+        dtype="bfloat16",
+        initializer_range=0.02,
+        use_cache=False,
+        **kwargs,
+    ):
+        self.vocab_size = vocab_size
+        self.hidden_size = hidden_size
+        self.num_hidden_layers = num_hidden_layers
+        self.h_layers = h_layers if h_layers is not None else num_hidden_layers
+        self.l_layers = l_layers if l_layers is not None else num_hidden_layers
```
molbap: why is this if/else needed? if the config.json is obtained during conversion

src/transformers/models/hrm/convert_original_hrm_weights_to_hf.py
```diff
@@ -0,0 +1,188 @@
...
+    checkpoint = torch.load(input_checkpoint, map_location="cpu")
+
+    # Load or create config
+    if config_file is not None:
+        print(f"Loading config from {config_file}")
+        with open(config_file) as f:
+            config_dict = json.load(f)
+        config = HrmConfig(**config_dict)
+    else:
+        # Try to extract config from checkpoint metadata
```
molbap, line 83: we can do without all these comments
```suggestion
```

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1017 @@
...
+    that is critical to HRM's training dynamics. This custom initialization differs from
+    standard PyTorch initialization and is applied during __init__.
+    """
+
+    def __init__(self, in_features: int, out_features: int, bias: bool = False):
+        super().__init__()
+        self.weight = nn.Parameter(
+            truncated_normal_init_(torch.empty((out_features, in_features)), std=1.0 / (in_features**0.5))
+        )
+        self.bias = None
```
molbap: this linear layer can be inherited from pre-existing ones! you just need to capture ` HrmLinear` name in the ` init_weights` module, which preserves the needed initialization scheme

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,704 @@
...
+                    0,
+                    0,
+                    0,
+                    2,
+                    0,
+                    0,
+                    0,
+                    6,  # Row 6
+                    0,
+                    6,
```
molbap: perfect, thanks!! let's use #noqa  or turn off formatting for a block to save some rows haha

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -0,0 +1,704 @@
...
+                outputs = model(input_ids=input_ids)
+
+            self.assertIsNotNone(outputs)
+
+    def test_save_and_load(self):
+        """Test model save and load functionality."""
+        config, input_ids, *_ = self.model_tester.prepare_config_and_inputs()
+        model = HrmForCausalLM(config).to(torch_device)
+
+        import tempfile
```
molbap: no imports in defs

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1017 @@
...
+        inv_freq = 1.0 / (base ** (torch.arange(0, dim, 2, dtype=torch.float32, device=device) / dim))
+        t = torch.arange(max_position_embeddings, dtype=torch.float32, device=device)
+        freqs = torch.outer(t, inv_freq)
+        emb = torch.cat((freqs, freqs), dim=-1)
+        self.register_buffer("cos_cached", emb.cos(), persistent=False)
+        self.register_buffer("sin_cached", emb.sin(), persistent=False)
+
+    def forward(self, position_ids=None) -> tuple[torch.Tensor, torch.Tensor]:
+        """Forward pass - position_ids parameter is accepted for compatibility but not used."""
+        return self.cos_cached, self.sin_cached
```
molbap: agree it's simpler but we're going to overhaul RoPE soon, for the time being let's stick to the existing design please!

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1017 @@
...
+        if self.causal:
+            mask = torch.triu(torch.ones(seq_len, seq_len, device=query.device, dtype=torch.bool), diagonal=1)
+            scores = scores.masked_fill(mask, float("-inf"))
+
+        # Softmax and apply to values
+        attn = torch.softmax(scores, dim=-1)
+        output = torch.matmul(attn, value)
+
+        # Transpose back to (batch, seq_len, num_heads, head_dim)
+        return output.transpose(1, 2).contiguous()
```
molbap: This should be a helper function outside of the class, and aligned with existing ` eager_attention_forward`, why not use directly the one from llama for instance?

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -0,0 +1,1017 @@
...
+                key_states=key,
+                value_states=value,
+                attention_mask=None,  # HRM doesn't use attention masks
+                query_length=seq_len,
+                is_causal=self.causal,
+                dropout=0.0,  # HRM doesn't use attention dropout
+            )
+        except (ValueError, ImportError):
+            # Fall back to eager attention if Flash Attention is not available
+            attn_output = self._eager_attention(query, key, value)
```
molbap: This is already handled by our attention interface, and should be of the form

```python
```
so the selection mechanism here is not needed

molbap (commented):
Left another review with some points to improve, thanks!

molbap:
Let me know when you want eyes on this @zbloss , looking forward to see this (and TRM) merged :rocket:

tests/models/hrm/test_modeling_hrm.py
```diff
@@ -671,34 +511,73 @@ def test_pretrained_model_reasoning(self):
...
+
+        # The model should get at least 50% of empty cells correct to show it's reasoning
+        # (Random guessing would give ~11% accuracy for digits 1-9)
+        self.assertGreater(
+            correct_predictions.item(),
+            0.88,
+            f"Model only got {correct_predictions.item():.1%} of empty cells correct. "
+            f"This suggests the model is not performing meaningful reasoning. "
+            f"Expected at least 50% accuracy on empty cells.",
+        )
```
molbap, line 507: Cool test! Isn't it redundant with the one above though? It seems they are measuring the same thing, one with a threshold above 50% and the other above 88%, am I missing something? We can remove one of the two IMO

tests/models/hrm/test_modeling_hrm.py
molbap: Same comment

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -363,94 +335,71 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
...
+
+    # Apply causal mask if needed
+    if module.causal:
+        mask = torch.triu(torch.ones(seq_len, seq_len, device=query.device, dtype=torch.bool), diagonal=1)
+        scores = scores.masked_fill(mask, float("-inf"))
+
+    # Softmax and apply to values
+    attn_weights = torch.softmax(scores, dim=-1, dtype=torch.float32).to(query.dtype)
+    attn_weights = F.dropout(attn_weights, p=dropout, training=module.training)
+    output = torch.matmul(attn_weights, value)
```
molbap: Given that q k v states are already computed and separated, this is standard: this `eager_attention_forward` can simply be imported from e.g. `llama`

src/transformers/models/hrm/modular_hrm.py
```diff
@@ -573,42 +522,52 @@ def __init__(self, config: HrmConfig):
...
+                (object,),
+                {
+                    "hidden_size": config.hidden_size,
+                    "num_attention_heads": config.num_attention_heads,
+                    "max_position_embeddings": config.max_position_embeddings + self.puzzle_embedding_length,
+                    "rope_theta": config.rope_theta,
+                    "rope_scaling": None,
+                },
+            )()
+            self.rotary_embedding = LlamaRotaryEmbedding(rope_config)
```
molbap: Are we ever using this code block? If no, we can remove it, if yes, we need to define above
```python
class HrmRotaryEmbedding(LlamaRotaryEmbedding):
    pass
```
and then do 
```suggestion
        if self.positional_encoding_type == "rope":
            # Create a temporary config for LlamaRotaryEmbedding with adjusted max_position_embeddings
            rope_config = type(
                "obj",
                (object,),
                {
                    "hidden_size": config.hidden_size,
                    "num_attention_heads": config.num_attention_heads,
                    "max_position_embeddings": config.max_position_embeddings + self.puzzle_embedding_length,
                    "rope_theta": config.rope_theta,
                    "rope_scaling": None,
                },
            )()
            self.rotary_embedding = HrmRotaryEmbedding(rope_config)
```

src/transformers/models/hrm/modular_hrm.py
molbap: Now that `init_weights` captures correctly `HrmLinear` and uses the custom truncated init, HrmLinear can be inherited from a standard linear now as it's only the initialization that changes. Sorry for these changes but we try to reduce the read diff as much as we can, especially for a model with a very different core system, what is similar _must_ be inherited when possible for clarity. If it seems impossible/doesn't work, tell me what breaks and we can also update modular accordingly.

molbap (commented):
Left some comments, ping me for a full review!

tests/models/hrm/test_modeling_hrm.py
zbloss: Yep  this was code left over from some copy-paste work on my part. Removing and updating the comments

src/transformers/models/hrm/modular_hrm.py
zbloss: Nope, good catch. Removed

src/transformers/models/hrm/modular_hrm.py
zbloss: Yep, this is now a straight forward eager attention forward pass, no need to duplicate the codebase.

molbap:
Thanks for the update, reviewing soon!

zbloss:
Just saw the release notes for v5.0.0rc0

Are there any changes I'm going to need to make in this PR to prepare for v5.0.0, or are we aiming to get this merged before the big v5.0.0 release?

ArthurZucker:
No should be alright! Just let me review! 🤗

vasqu:
We can close this, HRM has been merged in #46025

PR #41273, opened 2025-10-02
fix(quantization): Skip weight initialization for quantized models

### LINKED ISSUE #39366: RuntimeError when loading llmcompressor W8A8 quantized model: int8 dtype in weight initialization
I'm trying to load the quantized model `RedHatAI/Qwen2.5-VL-7B-Instruct-quantized.w8a8` but encountering a dtype compatibility issue during model initialization. The model appears to be quantized using `llmcompressor` with W8A8 quantization scheme.

**Note**: I need to load this model without vLLM because I may need to add custom hooks for my research, so I'm looking for a direct loading method using transformers/llmcompressor.

## Error Message

```python
RuntimeError: expected a floating-point or complex dtype, but got dtype=torch.int8
```

**Full Stack Trace:**
```python
File "/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py", line 366, in _init_weights
    module.weight.data.normal_(mean=0.0, std=std)
File "/torch/_refs/__init__.py", line 6214, in normal_
    return normal(mean, std, self.shape, out=self, generator=generator)
...
RuntimeError: expected a floating-point or complex dtype, but got dtype=torch.int8
```

## Traceback

The error occurs during model weight initialization where transformers tries to call `normal_()` on int8 tensors. The `normal_()` function in PyTorch only works with floating-point tensors, but the quantized model contains int8 weights.

**Specific failure point:**
- File: `modeling_qwen2_5_vl.py`, line 366
- Function: `_init_weights()` 
- Operation: `module.weight.data.normal_(mean=0.0, std=std)`
- Issue: Trying to apply normal distribution to int8 tensors

## Model Information

Based on the model's `config.json`:
- **Quantization method**: `compressed-tensors`
- **Format**: `int-quantized` 
- **Scheme**: W8A8 (8-bit weights and activations)
- **Base model**: `Qwen/Qwen2.5-VL-7B-Instruct`
- **Compression ratio**: ~1.2x
- **Ignored layers**: All visual layers (`visual.blocks.*`, `visual.merger.*`, `lm_head`)

### 1. llmcompressor methods:
```python
# Method 1: TraceableQwen2_5_VLForConditionalGeneration
from llmcompressor.transformers.tracing import TraceableQwen2_5_VLForConditionalGeneration
model = TraceableQwen2_5_VLForConditionalGeneration.from_pretrained(
    model_path, device_map="auto", torch_dtype="auto", trust_remote_code=True
)

# Method 2: SparseAutoModelForCausalLM  
from llmcompressor.transformers import SparseAutoModelForCausalLM
model = SparseAutoModelForCausalLM.from_pretrained(
    model_path, device_map="auto", torch_dtype="auto", trust_remote_code=True
)
```

### 2. Standard transformers methods:
```python
# Method 3: Various dtype configurations
model = Qwen2_5_VLForConditionalGeneration.from_pretrained(
    model_path,
    torch_dtype=torch.bfloat16,  # Also tried: torch.float16, "auto", None
    trust_remote_code=True,
    device_map="auto"
)

# Method 4: AutoModelForCausalLM
    model_path, trust_remote_code=True, torch_dtype="auto"
)
```

**All methods fail at the same weight initialization step, so I wonder should the model be loaded with `_fast_init=False` or other special parameters?**

## Additional Observations

1. **Warning about ignored layers**: The loader warns about missing visual layers, but this seems expected since they were ignored during quantization
2. **Model files exist**: The quantized model directory contains the expected `.safetensors` files and configuration
3. **Original model works**: The base `Qwen/Qwen2.5-VL-7B-Instruct` loads and works perfectly

## Environment

- **Python**: 3.10
- **PyTorch**: 2.7.0+cu126
- **Transformers**: 4.52.4
- **LLMCompressor**: 0.6.0
- **Compressed-tensors**: 0.10.2

This model was likely created using llmcompressor's oneshot quantization:
```python
from llmcompressor.modifiers.quantization import GPTQModifier
from llmcompressor.transformers import oneshot

recipe = [
    GPTQModifier(
        targets="Linear",
        scheme="W8A8", 
        sequential_targets=["Qwen2_5_VLDecoderLayer"],
        ignore=["lm_head", "re:visual.*"],
    ),
]
```
If this is more of an llmcompressor-specific model loading issue rather than a transformers compatibility issue, please let me know and I'll file this issue in the llmcompressor repository instead.

mauricioharley:
## Description

This Pull Request fixes a `RuntimeError` that occurs when loading `llmcompressor` W8A8 quantized models (e.g., `RedHatAI/Qwen2.5-VL-7B-Instruct-quantized.w8a8`) due to an attempt to initialize `int8` weights using `torch.nn.init.normal_()`, which only supports floating-point dtypes.

The issue was identified in `modeling_utils.py` within the `_initialize_missing_keys` method. When `is_quantized` is `True`, the `else` branch was still calling `self.initialize_weights()`, leading to the `RuntimeError`.

## Proposed Change

Added a conditional check `if not is_quantized:` before the call to `self.initialize_weights()` in the `else` branch of the `_initialize_missing_keys` method. This ensures that weight initialization is skipped for quantized models, as their weights are either already defined or will be loaded from a pretrained state dictionary, making the initialization redundant and problematic.

## Related Issue

Closes #39366

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -730,6 +730,7 @@ def _load_state_dict_into_meta_model(
                     device_mesh.get_local_rank(),
                     **sharding_kwargs,
                 )
+                hf_quantizer.register_loaded_quantized_param(model, param_name)
         else:
             param = param[...]
             if casting_dtype is not None:
@@ -758,6 +759,7 @@ def _load_state_dict_into_meta_model(
             else:
                 # TODO naming is stupid it loads it as well
                 hf_quantizer.create_quantized_param(model, param, param_name, param_device)
+                hf_quantizer.register_loaded_quantized_param(model, param_name)
 
                 # For quantized modules with FSDP/DeepSpeed Stage 3, we need to quantize the parameter on the GPU
                 # and then cast it to CPU to avoid excessive memory usage on each GPU
--- a/src/transformers/quantizers/base.py
+++ b/src/transformers/quantizers/base.py
@@ -61,6 +61,8 @@ class HfQuantizer(ABC):
         # -- Handle extra kwargs below --
         self.modules_to_not_convert = kwargs.pop("modules_to_not_convert", [])
         self.pre_quantized = kwargs.pop("pre_quantized", True)
+        # Track quantized parameters we create so they do not trigger reinitialization later on.
+        self._loaded_quantized_keys: set[str] = set()
 
         if not self.pre_quantized and self.requires_calibration:
             raise ValueError(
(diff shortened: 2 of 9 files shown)


ydshieh:
cc @MekkCyber @SunMarc

src/transformers/modeling_utils.py
```diff
@@ -5665,7 +5665,8 @@ def set_is_initialized_for_modules(module):
             with deepspeed.zero.GatheredParameters(not_initialized_parameters, modifier_rank=0):
                 self.initialize_weights()
         else:
-            self.initialize_weights()
+            if not is_quantized:
+                self.initialize_weights()
```
SunMarc: quantized models shouldn't be handled differently. Maybe we need to set `_is_hf_initialized` somewhere in the modules that are impacted.

mauricioharley: Thanks for the catch, @SunMarc! I restored the regular `initialize_weights()` path so quantized models no longer skip it, and added a hook in the quantizers to mark the tensors we create as `_is_hf_initialized`. That way the initialization logic still runs, but the quantized parameters stay untouched.

ArthurZucker (commented):
hey! Sorry for the radio silence, if you still want to work on this we merged #41580 which changes this a lot!

PR #41280, opened 2025-10-02
Use max/min

cyyever:
# What does this PR do?

Some code can directly use `max/min` to avoid sorting.


Rocketknight1 (approved):
LGTM! Though I might have to stop and think for a second when I see the `max()` of a list of files, lol

PR #41285, opened 2025-10-02
feat: use `aws-highcpu-32-priv` for amd docker img build

McPatate:
change the runner group to something more powerful, amd docker image build is very greedy in resources


LysandreJik (approved):
ok, cc @ydshieh

PR #41286, opened 2025-10-02
[docs] remove references to recently deleted classes in non-`en` docs (onnx, feature processors)

gante:
# What does this PR do?

See title. These are causing our doc builder to crash.

_______________________

how to locally test the doc builder:
1. clone and install the doc builder
2. run `doc-builder build transformers docs/source/ja/ --language ja --clean` (replace `ja` by the target language of your choice)


docs/source/es/_toctree.yml
```diff
@@ -64,8 +64,6 @@
     title: Entrenador
   - local: sagemaker
     title: Ejecutar el entrenamiento en Amazon SageMaker
-  - local: serialization
```
gante, line 67: in `es`, `it`, and `pt`: `serialization.md` was very outdated and doesn't match our recent practices -> delete

zucchini-nlp (approved):
Oh god, we really need to organize non-English language docs. Forgot they existed 🙃

ydshieh:
Should we try to avoid the doc build workflow run failing because of a failure from the build for other language?

i.e. always upload the built `en` doc (if it builds successfully).

gante:
@ydshieh yeah that would be nice (since it's the most relevant language, and we always check its CI at merge time)

ydshieh:
OK, will try to see if I can make that work

PR #41287, opened 2025-10-02
[v5] remove load_in_4bit and load_in_8bit

SunMarc:
# What does this PR do?

This PR removes `load_in_4bit` and `load_in_8bit` deprecated arguments for v5. This is something that I didn't want to do in a minor release as this might be breaking for a lot of users.


ydshieh:
Could you run a few integration tests to make sure 🙏 Thank you.

MekkCyber (approved):
Finallyyyy 🔥 ! Great clean up

src/transformers/quantizers/auto.py
```diff
@@ -121,11 +121,7 @@ class AutoQuantizationConfig:
     @classmethod
     def from_dict(cls, quantization_config_dict: dict):
         quant_method = quantization_config_dict.get("quant_method")
-        # We need a special care for bnb models to make sure everything is BC ..
-        if quantization_config_dict.get("load_in_8bit", False) or quantization_config_dict.get("load_in_4bit", False):
-            suffix = "_4bit" if quantization_config_dict.get("load_in_4bit", False) else "_8bit"
-            quant_method = QuantizationMethod.BITS_AND_BYTES + suffix
-        elif quant_method is None:
+        if quant_method is None:
```
matthewdouglas, line 124: I'm not sure this is safe to remove; do we still need the `_4bit` and `_8bit` suffixes to match the mapping in `AUTO_QUANTIZER_MAPPING` and `AUTO_QUANTIZATION_CONFIG_MAPPING`?

Bitsandbytes is a little bit different here in that we have two separate `HfQuantizer` selected based on `load_in_4bit` vs `load_in_8bit` instead of just `quant_method=="bitsandbytes"`

SunMarc: good catch ! indeed we shouldn't remove that

SunMarc:
run-slow: bnb

SunMarc:
run-slow: gemma

ArthurZucker:
hahha have it in #41445 as well for from_pretrained

PR #41292, opened 2025-10-02
Fix mxfp4 dequantization

Cyrilvallez:
# What does this PR do?

https://github.com/huggingface/transformers/pull/41222 changed the signature of a caller in TP without updating the function calls, so `dequantization` is currently broken. This fixes it.


ArthurZucker (commented):
thanks 😢

ArthurZucker (approved):

Rocketknight1:
My bad - I approved that PR! I didn't realize that callers were setting the unused function argument

PR #41293, opened 2025-10-02
Tiny Cleanup - Removed duplicate class field definition's

Sai-Suraj-27:
# What does this PR do?
Tiny cleanup. Used [Ruff-PIE794](https://docs.astral.sh/ruff/rules/duplicate-class-field-definition/). Will be happy to add the rule to [existing rules](https://github.com/huggingface/transformers/blob/e4930d6bde027ed21fd3dd91dbbfdc95e0221dd4/pyproject.toml#L36)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@ydshieh @Rocketknight1


Rocketknight1 (approved):
Yes, good fix! I also think this would be a nice rule to add to the ruff rules, cc @arthurzucker @cyrilvallez if you're okay with it.

src/transformers/pipelines/text_to_audio.py
```diff
@@ -84,7 +84,6 @@ class TextToAudioPipeline(Pipeline):
     _load_processor = True
 
     _pipeline_calls_generate = True
-    _load_processor = False
```
Cyrilvallez, line 87: The wrong value has been kept here - only False should remain (the last value that was set)

Cyrilvallez (commented):
LGTM but the wrong one was kept in a file

src/transformers/pipelines/text_to_audio.py
Sai-Suraj-27: Thanks for the review @Cyrilvallez @Rocketknight1. Updated & I've added the `"PIE794"` ruff rule as well.

Cyrilvallez (approved):
LGTM, thanks for cleaning!

PR #41297, opened 2025-10-02
ENH: Add support for LoRA hotswapping

BenjaminBossan:
# What does this PR do?

LoRA hotswapping has been available in PEFT since 0.15.0. There is already a diffusers
integration (https://github.com/huggingface/diffusers/pull/9453), but the transformers integration was still missing this feature. This PR remedies this. It sticks closely to the diffusers PR, both implementation-wise and API-wise. 

Hotswapping allows to swap different LoRA adapters in-place instead of loading multiple adapters and switching between them. Not only can this be advantageous to safe memory and potentially for quicker loading, the biggest advantage is that if the model is compiled, we can hotswap without triggering recompilation (loading a separate adapter would require recompilation).

There are some caveats to using this feature, most notably that only LoRA is supported. This was fine for diffusers, as it only works with LoRA, but the transformers integration works with other PEFT methods too. However, LoRA should be by far the most common method, so this should be fine for now. This and other caveats have been documented.

Note that testing is not super deep, there could be more edge cases being tested. However, as this is mainly about calling PEFT functionality, which is extensively tested in PEFT, I focused the tests mostly on the integration, with only a couple of tests for the functionality itself (namely the tests that call `_check_model_hotswap`).

## Before submitting
      Pull Request section?
      to it if that's the case.


docs/source/en/peft.md
```diff
@@ -151,3 +151,45 @@ model.enable_adapters()
...
+# now you can hotswap the 2nd adapter, use the same name as for the 1st
+model.load_adapter(file_name_2, adapter_name="default", hotswap=True)
+output_2 = model(...)
+```
+
+The `target_rank=max_rank` argument is important for setting the maximum rank among all LoRA adapters that will be loaded. If you have one adapter with rank 8 and another with rank 16, pass `target_rank=16`. You should use a higher value if in doubt. By default, this value is 128.
+
+However, there can be situations where recompilation is unavoidable. For example, if the hotswapped adapter targets more layers than the initial adapter, then recompilation is triggered. Try to load the adapter that targets the most layers first. Refer to the PEFT docs on [hotswapping](https://huggingface.co/docs/peft/main/en/package_reference/hotswap#peft.utils.hotswap.hotswap_adapter) for more details about the limitations of this feature.
+
+> [!Tip]
```
sayakpaul, line 196: We could also redirect the interested users to our diffusion model benchmark numbers so that they know the speed is significant.

tests/peft_integration/test_peft_integration.py
```diff
@@ -940,3 +941,175 @@ def test_peft_pipeline_no_warning(self):
 
             # Generate text to verify pipeline works
             _ = lora_generator(text, max_new_tokens=20)
+
+
+@require_peft
+@require_torch
+@slow
+class PeftHotswapIntegrationTester(unittest.TestCase):
```
sayakpaul: @ydshieh okay for you?

tests/peft_integration/test_peft_integration.py
```diff
@@ -940,3 +941,175 @@ def test_peft_pipeline_no_warning(self):
...
+    def tearDown(self):
+        # It is critical that the dynamo cache is reset for each test. Otherwise, if the test re-uses the same model,
+        # there will be recompilation errors, as torch caches the model when run in the same process.
+        torch._dynamo.reset()
```
sayakpaul:
```suggestion
        torch.compiler.reset()
```

We could also introduce `torch._inductor.utils.fresh_inductor_cache()`. Example: https://github.com/huggingface/diffusers/blob/7242b5ff627fad93dd85834b0278267b6cbe2d6d/tests/models/test_modeling_common.py#L2061C13-L2061C57

sayakpaul (approved):
LGTM 🔥 

Let's maybe update `diffusers` to uplift our current constraint that this feature isn't available for text encoder LoRAs once this is merged.

tests/peft_integration/test_peft_integration.py
BenjaminBossan: Done, also added the use of `torch._inductor.utils.fresh_inductor_cache()`.

tests/peft_integration/test_peft_integration.py
BenjaminBossan: Just a note, these are the same decorators as for `PeftIntegrationTester`, but I didn't want to use it here (nor `PeftTesterMixin`), as the hotswap tests don't make use of the test matrix defined there.

docs/source/en/peft.md
BenjaminBossan: Which benchmark numbers do you mean? These: https://huggingface.co/datasets/diffusers/benchmarks? They don't contain numbers for hotswapping, do they?

BenjaminBossan:
> Let's maybe update diffusers to uplift our current constraint that this feature isn't available for text encoder LoRAs once this is merged.

Yes, that would be the next step once this PR is merged.

docs/source/en/peft.md
sayakpaul: I am referring to the numbers we gathered in https://github.com/huggingface/lora-fast

BenjaminBossan: Done

BenjaminBossan:
@ydshieh How to proceed with this PR? Sayak approved but I guess we need a transformers dev's approval too?

ArthurZucker:
sorry cc @Cyrilvallez can you reviiew?

src/transformers/integrations/peft.py
```diff
@@ -159,12 +161,63 @@ def load_adapter(
...
+                `hotswap=False`.
+
+                Note that hotswapping comes with a couple of limitations documented here:
+                https://huggingface.co/docs/peft/main/en/package_reference/hotswap
             adapter_kwargs (`dict[str, Any]`, *optional*):
                 Additional keyword arguments passed along to the `from_pretrained` method of the adapter config and
                 `find_adapter_config_file` method.
         """
+        from peft import PeftType
+
```
Cyrilvallez: Does that import exist even for the `MIN_VERSION` below? If not, let's import after the check, so that we have gracious crash

src/transformers/integrations/peft.py
```diff
@@ -159,12 +161,63 @@ def load_adapter(
...
         check_peft_version(min_version=MIN_PEFT_VERSION)
 
+        if hotswap == "auto":
+            # if user called model.enable_peft_hotswap and this is not the first adapter, enable hotswap
+            hotswap_enabled = getattr(self, "_hotswap_enabled", False)
+            not_first_adapter = bool(self._hf_peft_config_loaded and (adapter_name in self.peft_config))
```
Cyrilvallez, line 207: We should not need bool casting here, do we?
```suggestion
            not_first_adapter = self._hf_peft_config_loaded and adapter_name in self.peft_config
```

src/transformers/integrations/peft.py
```diff
@@ -159,12 +161,63 @@ def load_adapter(
...
+            min_version_hotswap = "0.15.0"
+            if version.parse(importlib.metadata.version("peft")) < version.parse(min_version_hotswap):
+                raise ValueError(f"To hotswap the adapter, you need PEFT >= v{min_version_hotswap}.")
+            if (not self._hf_peft_config_loaded) or (adapter_name not in self.peft_config):
+                raise ValueError(
+                    "To hotswap an adapter, there must already be an existing adapter with the same adapter name."
+                )
+            if any(conf.peft_type != PeftType.LORA for conf in self.peft_config.values()):
+                raise ValueError("Hotswapping is currently only supported for LoRA, please set `hotswap=False`.")
+
```
Cyrilvallez, line 220: Just concerned about that, maybe `False` would be a better default no? Here looks like current code would crash without reason when loading more adapters that are not lora
Or maybe a `None` value, that becomes `auto` if we use lora, and False otherwise?

src/transformers/integrations/peft.py
```diff
@@ -261,12 +322,47 @@ def load_adapter(
...
+                # For hotswapping of compiled models or adapters with different ranks.
+                # If the user called enable_peft_hotswap, we need to ensure it is called:
+                # - after the first adapter was loaded
+                # - before the model is compiled and the 2nd adapter is being hotswapped in
+                # Therefore, it needs to be called here
+                from peft.utils.hotswap import prepare_model_for_compiled_hotswap
+
+                prepare_model_for_compiled_hotswap(self, config=peft_config, **self._prepare_peft_hotswap_kwargs)
+                # We only want to call prepare_model_for_compiled_hotswap once
+                self._prepare_peft_hotswap_kwargs = None
```
Cyrilvallez, line 351: Should this be under the condition `if not hotswap`? 🤔

Cyrilvallez (commented):
Very nice! Just mostly a bit concerned about the default which seem to be able to crash currently code that is valid?

src/transformers/integrations/peft.py
BenjaminBossan: Done, the import should work with any PEFT version but I agree it's cleaner that way.

src/transformers/integrations/peft.py
BenjaminBossan: Right now, `_hf_peft_config_loaded` is a bool, so coercing is not needed. But if its type changes in the future, the type of `not_first_adapter` could also change, so the cast makes this line future proof. Just an example:

```python
>>> _hf_peft_config_loaded = {"foo": 1}
>>> adapter_name = "default"
>>> peft_config = {"default": 2}
>>> not_first_adapter = _hf_peft_config_loaded and adapter_name in peft_config
>>> not_first_adapter
True
>>> _hf_peft_config_loaded = {}  # falsy value short-circuits the conditional
>>> not_first_adapter = _hf_peft_config_loaded and adapter_name in peft_config
>>> not_first_adapter
{}
```

As you can see, in the last line, we suddenly have a different type for `not_first_adapter`.

src/transformers/integrations/peft.py
BenjaminBossan: This would only crash if the user tries to load non-LoRA **and** if they:

- passed `hotswap=True` or
- called `enable_peft_hotswap`

In either case, the user intent is to use hotswapping. Therefore, I think that raising is the better choice, otherwise the user would think they used hotswapping successfully when it's not actually being used.

src/transformers/integrations/peft.py
BenjaminBossan: It is, see line 335.

BenjaminBossan (commented):
@Cyrilvallez Thanks a lot for your review. I addressed your comments, please check again.

src/transformers/integrations/peft.py
Cyrilvallez: Yes but I meant it the other way haha, like is it intended? If seems strange to me to have calls such as `prepare_model_for_compiled_hotswap` under a conditional `if not hotswap`, which should mean we don't do anything about hotswapping right?

BenjaminBossan: I see, yes it can be a bit confusing. We only arrive here if the user called `enable_peft_hotswap`, so we can be sure that the intent is to hotswap later. Then, if we load the **first** adapter, which by definition cannot be hotswapped, we call `prepare_model_for_compiled_hotswap` so that the model is ready for hotswapping once we get to the **second** adapter.

There is a very specific order in which things need to be called, which I tried to document in the comment above and the docs. If there is a way to make it clearer, please let me know.

Cyrilvallez (approved):
Alright then, thanks for tge added clarifications! And sorry for the delay! Looks like you have a small conflict, but feel free to merge afterwards!

PR #41298, opened 2025-10-02
fix resample in asr pipeline

yhzx233:
# What does this PR do?

Fixes duplicated line in ASR pipeline that passed `in_sampling_rate` twice, causing `F.resample` to no-op.

### DIFF
--- a/src/transformers/pipelines/automatic_speech_recognition.py
+++ b/src/transformers/pipelines/automatic_speech_recognition.py
@@ -415,7 +415,6 @@ class AutomaticSpeechRecognitionPipeline(ChunkPipeline):
                 inputs = F.resample(
                     torch.from_numpy(inputs) if isinstance(inputs, np.ndarray) else inputs,
                     in_sampling_rate,
-                    in_sampling_rate,
                     self.feature_extractor.sampling_rate,
                 ).numpy()
                 ratio = self.feature_extractor.sampling_rate / in_sampling_rate


Rocketknight1:
Ah, good catch! Is there an example input that fails before this PR?

yhzx233:
> Ah, good catch! Is there an example input that fails before this PR?

here’s a minimal repro.

```python
from datasets import load_dataset
import torch
from torchaudio import functional as F

speech_recognizer = pipeline("automatic-speech-recognition", model="openai/whisper-tiny", num_beams=1)
ds = load_dataset("hf-internal-testing/librispeech_asr_dummy", "clean", split="validation[:1]")

sr16000 = ds[0]['audio']['array']
sr48000 = F.resample(torch.from_numpy(sr16000), 16000, 48000).numpy()

print(speech_recognizer({"array": sr16000, "sampling_rate": 16000}))
print(speech_recognizer({"array": sr48000, "sampling_rate": 48000}))
```

Expected output (after the fix):

```
{'text': ' Mr. Quilter is the apostle of the middle classes and we are glad to welcome his gospel.'}
{'text': ' Mr. Quilter is the apostle of the middle classes and we are glad to welcome his gospel.'}
```

Actual output (before the fix):

```
{'text': ' Mr. Quilter is the apostle of the middle classes and we are glad to welcome his gospel.'}
{'text': ' you'}
```

Rocketknight1:
Thanks! cc @eustlb @ebezzam, you can merge if you're happy with it

eustlb (approved):
Nice catch, thanks! 🤗

PR #41300, opened 2025-10-02
Standardize `PretrainedConfig` to `PreTrainedConfig`

Cyrilvallez:
# What does this PR do?

As per the title! This is very very long due.


src/transformers/configuration_utils.py
```diff
@@ -1335,21 +1335,60 @@ def recursive_diff_dict(dict_a, dict_b, config_obj=None):
...
+        candidates = {cls, PreTrainedConfig}
+        return any(c in candidates for c in sub.mro())
+
+
+class PretrainedConfig(PreTrainedConfig, metaclass=DummyMeta):
+    def __init__(self, *args, **kwargs):
+        logger.warning_once(
+            "`PretrainedConfig` is deprecated and will be removed in v5. Please use `PreTrainedConfig` instead!"
+        )
+        super().__init__(*args, **kwargs)
```
Cyrilvallez: This is a bit fancy but it's the most simple way to keep full BC on the old `PretrainedConfig` object, while still showing deprecation warning when it's used. The other solution would have been a simple class alias such as

```python
PretrainedConfig = PreTrainedConfig
```

but then it's impossible to surface warnings, and users are not advised of the deprecation going on on the object!

A simple class inheritance was not enough IMO, as I feel that patterns such as `isinstance(my_config, PretrainedConfig)` are quite common, and would otherwise break old code because all configs are now `PreTrainedConfig` descendants, not `PretrainedConfig` descendants

Current code ensures that using the old class in `isinstance/issubclass` checks still works as expected, i.e.

```python
from transformers import PretrainedConfig, LlamaConfig

# This raises a warning and returns True
isinstance(LlamaConfig(), PretrainedConfig)
>>> `PretrainedConfig` is deprecated and will be removed in the future. Please use `PreTrainedConfig` instead!
>>> True  # True even though LlamaConfig inherits from PreTrainedConfig, not PretrainedConfig
```

ArthurZucker: Mmmm interesting. This will be annoying for a lot of libraries I think, I would recommend against, as it won't hurt if people keep using it TBH.

ArthurZucker (commented):
Happy, but IMO let's not warn for something that is not really valuable. We won't ever remove the mapping / bc as it's not costly and its like a typo that can stick

Cyrilvallez:
Alright, will revert and just make the alias then!

Issue #41296, opened 2025-10-02
--lr-scheduler-kwargs: invalid dict value: '{"min_lr": 1e-06}'


stephankoe:
### System Info

- `transformers` version: 4.57.0.dev0
- Platform: Linux-5.10.134-008.18.kangaroo.al8.x86_64-x86_64-with-glibc2.39
- Python version: 3.12.7+gc
- Huggingface_hub version: 1.0.0.rc2
- Safetensors version: 0.5.3
- Accelerate version: 1.7.0
- Accelerate config:    not found
- DeepSpeed version: 0.16.9+ali
- PyTorch version (accelerator?): 2.6.0+ali.7.post2.ppu1.5.2.cu126 (CUDA)
- Using distributed or parallel set-up in script?: <fill in>
- Using GPU in script?: <fill in>
- GPU type: PPU-ZW810E

### Who can help?

When using the `--lr-scheduler-kwargs` option with a dict argument, such as `'{"min_lr": 1e-06}'`, the parser reports the following error:

```text
test.py: error: argument --lr_scheduler_kwargs/--lr-scheduler-kwargs: invalid dict value: '{"min_lr": 1e-06}'
```

It appears that this error was introduced in commit #a61fc6a0 titled "Fix typing of train_args". It works as expected when reversing the data type of `lr_scheduler_kwargs` back to `Optional[Union[dict[str, Any], str]]`.

### Information

- [x] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

You can simply reproduce it with a short script, such as:

```python
from transformers import TrainingArguments, HfArgumentParser

parser = HfArgumentParser((TrainingArguments,))
training_args, = parser.parse_args_into_dataclasses()
```

and then run it with `python test.py --lr-scheduler-kwargs '{"min_lr": 1e-06}'`

It will report the following error:

```
usage: test.py [-h] [--output_dir OUTPUT_DIR] [--overwrite_output_dir [OVERWRITE_OUTPUT_DIR]] [--do_train [DO_TRAIN]] [--do_eval [DO_EVAL]]
[...]

PR #41302, opened 2025-10-02
Fix HfArgumentParser parsing for Union[dict, str] fields (e.g., --lr_scheduler_kwargs)

gmmkmtgk:
…scheduler_kwargs)

# What does this PR do?

This PR fixes the bug where `Union[dict, str]` fields, such as `--lr_scheduler_kwargs` in `TrainingArguments`, could not be parsed correctly from the command line.

**Root Cause:**
The issue was introduced when the type of `lr_scheduler_kwargs` changed from `Optional[Union[dict[str, Any], str]]` to `Union[dict[str, Any], str]`. `HfArgumentParser` previously filtered out `str` from Union types when `None` was not present, which caused `argparse` to attempt parsing the argument as a `dict` directly, leading to errors when passing JSON strings via CLI.

**Fix:**

* Modified `src/transformers/hf_argparser.py` to detect `Union[dict, str]` types and keep the type as `str` for `argparse`.
* Conversion from JSON string → `dict` now happens later in `TrainingArguments.post_init_`.
* Maintains backward compatibility with `Optional[Union[dict, str]]` fields.

**Tests added:**

* `test_17_union_dict_str_parsing()`: verifies basic Union[dict, str] CLI parsing.
* `test_18_lr_scheduler_kwargs_parsing()`: regression test specifically for the `lr_scheduler_kwargs` issue.

Fixes the behavior reported in [#41296](https://github.com/huggingface/transformers/issues/41296).

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/transformers/hf_argparser.py
+++ b/src/transformers/hf_argparser.py
@@ -176,9 +176,19 @@ class HfArgumentParser(ArgumentParser):
                     f" Problem encountered in field '{field.name}'."
                 )
             if type(None) not in field.type.__args__:
-                # filter `str` in Union
-                field.type = field.type.__args__[0] if field.type.__args__[1] is str else field.type.__args__[1]
-                origin_type = getattr(field.type, "__origin__", field.type)
+                # For Union[dict, str] or Union[str, dict], keep str type for argparse
+                # The conversion from JSON string to dict happens in the dataclass __post_init__
+                non_str_type = (field.type.__args__[0] if field.type.__args__[1] is str else field.type.__args__[1])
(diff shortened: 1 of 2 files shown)


Rocketknight1:
cc @sunmarc I think!

SunMarc (commented):
Thanks for this PR ! While the fix seems to work, I prefer to [revert](https://github.com/huggingface/transformers/pull/41322) the the previous dtype as this is more in line with other args like fsdp_config.

gmmkmtgk:
@SunMarc any way this contribution can be counted in hacktoberfest?

PR #41308, opened 2025-10-03
Subconfig is a class attribute

zucchini-nlp:
# What does this PR do?

As per title, it should be a class attribute for vision-only models. VLMs already use class attribute. If it is a property, it is unavailable before we init the class. Having it as a class attribute will be needed for integrating type validation and  transforming all configs to a dataclass

Also cleaned up some duplicate code. An attribute map already maps those keys, so an extra property is not needed


src/transformers/models/conditional_detr/configuration_conditional_detr.py
```diff
@@ -135,6 +135,7 @@ class ConditionalDetrConfig(PretrainedConfig):
     ```"""
 
     model_type = "conditional_detr"
+    sub_configs = {"backbone_config": AutoConfig}
```
ydshieh, line 138: I am somehow worried that this will cause the problem of permanent changes if a user change something in `sub_configs` in order to initialize a specific instance.

That happens before and quite difficult to debug.

Would it be possible to do like

```
class ConditionalDetrConfig:
    _ sub_configs  = ...

    # could make this class method if necessary
    def sub_configs():
        return copy.deepcopy(self._sub_configs)
```


```

Comment (zucchini-nlp) [reply src/transformers/models/conditional_detr/configuration_conditional_detr.py:138]: yeah, we could. Though I would strongly discourage changing it, as it is designed for us to know beforehand what subconfigs exist. And chnaging it does not make sense

Comment (zucchini-nlp) [reply src/transformers/models/conditional_detr/configuration_conditional_detr.py:138]: > That happens before and quite difficult to debug.

Oh just saw this. Do you remember when people would change it manually? Curious because it has been a cls attr for VLMs and I haven't seen issues

Comment (ydshieh) [reply src/transformers/models/conditional_detr/configuration_conditional_detr.py:138]: I read it too quickly. Indeed the attribute `sub_configs` is not meant to be changed, so ok for the current design.

Previous one is something about keeping in fp32 modules (which makes sense to be changed)

**Review by Cyrilvallez — COMMENTED** [review]
Looks very nice! You have a few conflicts due to https://github.com/huggingface/transformers/pull/41300 though, sorry!

**Comment by zucchini-nlp** [comment]
Approve pls 🙏🏻

File: src/transformers/modeling_utils.py [hunk 2]
```diff
@@ -1187,13 +1187,13 @@ def _get_dtype(
                 dtype = getattr(torch, dtype)
                 config.dtype = dtype
                 for sub_config_key in config.sub_configs:
-                    sub_config = getattr(config, sub_config_key)
-                    sub_config.dtype = dtype
+                    if (sub_config := getattr(config, sub_config_key)) is not None:
+                        sub_config.dtype = dtype
```
Cyrilvallez, line 1191: I think this check is needed only if `super().__init__()` is called before subconfigs are set in the config no? Let's quickly move the call to the bottom on all changed configs if so instead of this!

src/transformers/modeling_utils.py
```diff
@@ -1187,13 +1187,13 @@ def _get_dtype(
...
         elif isinstance(dtype, torch.dtype):
             config.dtype = dtype
             for sub_config_key in config.sub_configs:
-                sub_config = getattr(config, sub_config_key)
-                sub_config.dtype = dtype
+                if (sub_config := getattr(config, sub_config_key)) is not None:
+                    sub_config.dtype = dtype
```
Cyrilvallez, line 1196: same

src/transformers/modeling_utils.py
```diff
@@ -1218,8 +1218,8 @@ def _get_dtype(
         default_dtype = torch.get_default_dtype()
         config.dtype = default_dtype
         for key in config.sub_configs:
-            value = getattr(config, key)
-            value.dtype = default_dtype
+            if (sub_config := getattr(config, key)) is not None:
+                sub_config.dtype = default_dtype
```
Cyrilvallez, line 1222: same

src/transformers/modeling_utils.py
```diff
@@ -2673,34 +2673,34 @@ def set_attn_implementation(self, attn_implementation: Union[str, dict]):
...
+                    subconfig._attn_implementation_internal = sub_implementation
+                    logger.warning(
+                        f"We set the attention implementation for the sub-config `{subconfig_key}` to `{sub_implementation}` "
+                        "without finding the associated sub-model. For this reason we could not check if the model supports it. "
+                        "You may encounter undefined behavior."
+                    )
+                # Unset the attribute in this case, to avoid issues in the future
+                else:
+                    if hasattr(subconfig, "_attn_was_changed"):
+                        del subconfig._attn_was_changed
```
Cyrilvallez, line 2703: same here, we could revert everything

Cyrilvallez (approved):
Let's just revert all changes to `modeling_utils.py`, and instead fix where `super().__init__` is called in configs (should be at the bottom, after subconfigs are set) - see e.g. dpt!
Feel free to merge afterwards!

src/transformers/modeling_utils.py
zucchini-nlp: not only, some models do not initialize a sub config at run time and either 1) infer when modeling in [`load_backbone`](https://github.com/huggingface/transformers/blob/7aa888b7fa477d13153ffbfe107dfbd6c696014a/src/transformers/utils/backbone_utils.py#L306) from a variety of config keys 2) do not need the path with sub-model so they don't need a config for that

These are all vision models with backbones, and maybe we can update them to follow the common standards with nested configs. I would prefer to check and do standardization much later, after making and finishing the config type validation PR. So I'll merge it for now and take note to myself to come back and dig into backbone-vision configs :)

PR #40132, opened 2025-08-13
MoE + vllm = 😻 

ArthurZucker:
# What does this PR do?
Starting with vllm, but also towards better training

TODO:
- DBRX, GPTSAN, ERNIE4.5, Jamba (ça été fait par ia je crois), Minmax pareil,  NllbMoe, PhiMoe (AI based), switch et vitpose)

PR #41309, opened 2025-10-03
Fix jamba

Cyrilvallez:
# What does this PR do?

Jamba was mostly destroyed in https://github.com/huggingface/transformers/pull/40132, this PR fixes it!

Slow tests are good

### DIFF
--- a/src/transformers/models/jamba/modeling_jamba.py
+++ b/src/transformers/models/jamba/modeling_jamba.py
@@ -169,7 +169,7 @@ class HybridMambaAttentionDynamicCache:
         """Returns the sequence length of the cached states. A layer index can be optionally passed."""
         # take any layer that contains cache and not empty tensor
         layer_idx = self.transformer_layers[0] if layer_idx not in self.transformer_layers else layer_idx
-        if len(self.key_cache) <= layer_idx:
+        if len(self.key_cache) <= layer_idx or self.key_cache[layer_idx].shape[-1] == 0:
             return 0
         return self.key_cache[layer_idx].shape[-2]
 
--- a/src/transformers/models/jamba/modeling_jamba.py
+++ b/src/transformers/models/jamba/modeling_jamba.py
@@ -121,6 +121,12 @@ class HybridMambaAttentionDynamicCache:
         self.key_cache = [torch.tensor([[]] * batch_size, device=device) for _ in range(config.num_hidden_layers)]
         self.value_cache = [torch.tensor([[]] * batch_size, device=device) for _ in range(config.num_hidden_layers)]
 
+    def __len__(self):
+        return len(self.key_cache)
+
+    def __getitem__(self, layer_idx):
+        return self.key_cache[layer_idx], self.value_cache[layer_idx]
+
     def update(
         self,
         key_states: torch.Tensor,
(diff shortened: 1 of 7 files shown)


src/transformers/models/jamba/modeling_jamba.py
```diff
@@ -148,7 +169,7 @@ def get_seq_length(self, layer_idx: Optional[int] = 0) -> int:
         """Returns the sequence length of the cached states. A layer index can be optionally passed."""
         # take any layer that contains cache and not empty tensor
         layer_idx = self.transformer_layers[0] if layer_idx not in self.transformer_layers else layer_idx
-        if len(self.key_cache) <= layer_idx:
+        if len(self.key_cache) <= layer_idx or self.key_cache[layer_idx].shape[-1] == 0:
```
Cyrilvallez, line 172: Without this check, `get_seq_length` would return the batch size when the cache is empty, because layers are initialized as `torch.tensor([[]] * batch_size, device=device)`, which is batch size dim for `shape[-2]`.... 🫠🫠🫠🫠

src/transformers/models/jamba/modular_jamba.py
```diff
@@ -561,21 +535,24 @@ def forward(
...
             use_cache=use_cache,
             cache_position=cache_position,
+            **kwargs,
         )
         hidden_states = residual + hidden_states
         residual = hidden_states
         hidden_states = self.pre_ff_layernorm(hidden_states)
         hidden_states = self.feed_forward(hidden_states)
+        if isinstance(hidden_states, tuple):
+            hidden_states, _ = hidden_states
```
ArthurZucker: not needed, router logits not needed to be returned

src/transformers/models/jamba/modular_jamba.py
```diff
@@ -533,12 +507,12 @@ def route_tokens_to_experts(self, hidden_states, router_logits):
...
     def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
         batch_size, sequence_length, hidden_dim = hidden_states.shape
-        hidden_states = hidden_states.view(-1, hidden_states.shape[-1])
+        hidden_states = hidden_states.view(-1, hidden_dim)
         router_logits = self.router(hidden_states)
         top_k_index, top_k_weights = self.route_tokens_to_experts(hidden_states, router_logits)
         hidden_states = self.experts(hidden_states, top_k_index, top_k_weights)
         hidden_states = hidden_states.reshape(batch_size, sequence_length, hidden_dim)
-        return hidden_states
+        return hidden_states, router_logits
```
ArthurZucker:
```suggestion
        return hidden_states
```

ArthurZucker (approved):
ty for cleaning up

src/transformers/models/jamba/modular_jamba.py
```diff
@@ -624,27 +603,16 @@ class JambaPreTrainedModel(PreTrainedModel):
     _skip_keys_device_placement = "past_key_values"
     _supports_flash_attn = True
     _supports_sdpa = True
-    # Note: only supports HybridMambaAttentionDynamicCache
     _is_stateful = True
     _can_record_outputs = {
         "hidden_states": [JambaAttentionDecoderLayer, JambaMambaDecoderLayer],
         "attentions": JambaAttention,
-        "router_logits": OutputRecorder(nn.Linear, layer_name="router"),
+        "router_logits": OutputRecorder(JambaSparseMoeBlock, index=1),
```
ArthurZucker: no needed either, you need `feed_forward.router` as the name tho

PR #35847, opened 2025-01-22
Fix Jitter Noise Passing to Experts in Switch Transformers #33969

sambhavnoobcoder:
## Issue Description

This pull request addresses a bug in the Switch Transformers architecture where the jitter noise (intended to be applied only for routing decisions) was also being unintentionally applied to the expert inputs.

Fixes : #33969 

## Problem Statement

In Switch Transformers, a small amount of jitter noise is added to the inputs at routing time to ensure route diversity. However, these jittered inputs were incorrectly passed along to the experts, which contradicts the original paper’s design and led to unexpected discrepancies in outputs.

## Root Cause

It was discovered that the code used the same hidden states for both routing and expert processing. When jitter noise was enabled in training mode, it directly modified these states in place, causing the experts to receive noisy inputs.

## Implementation

1. We now clone the original hidden states before applying jitter noise.  
2. A separate copy is used exclusively for computing router logits and probabilities.  
3. The unchanged hidden states are then fed to the experts to maintain the original semantics.

## Screenshot

Screenshot 2025-01-23 at 1 14 26 AM

## Test Cases

1. **test_router_training_mode**  
   • Objective: Ensures that jitter noise is only applied during training.  
   • Checks that outputs differ between consecutive runs (due to noise) but original inputs remain unchanged.

2. **test_router_jitter_noise_separation **  
   • Objective: Verifies that jitter noise affects only the router’s internal computations and not the expert inputs.  
   • Confirms the logits differ when jitter is applied, while the main input stays the same.

3. **test_expert_inputs_consistency**  
   • Objective: Asserts that all expert inputs remain consistent, even when jitter is applied during training.  
   • Uses a forward hook on the first expert to capture its inputs across multiple runs and compares them.

With these changes and test additions, we ensure that Switch Transformers adhere to the original design while preserving backward compatibility and correctness.

**cc :** @ArthurZucker

PR #41313, opened 2025-10-03
Jitter noise PR

### LINKED ISSUE #33969: Jitter Noise added to input being passed to experts in Switch Transformers
### System Info

System Info

- transformers version: 4.44.2
- Platform: Linux-6.5.0-35-generic-x86_64-with-glibc2.35
- Python version: 3.10.14
- Huggingface_hub version: 0.24.6
- Safetensors version: 0.4.5
- Accelerate version: 0.34.2
- PyTorch version (GPU?): 2.3.0 (True)
- Jax version: not installed
- Using GPU in script?: No

### Who can help?

@ArthurZucker @younesbelkada 

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

```python
import torch
import torch.nn as nn
from transformers import (
    SwitchTransformersConfig,
    SwitchTransformersTop1Router,
)
from transformers.models.switch_transformers.modeling_switch_transformers import SwitchTransformersDenseActDense


class MySwitchTransformersSparseMLP(nn.Module):
    r"""
    Implementation of the Switch Transformers Sparse MLP module.
    """

    def __init__(self, config: SwitchTransformersConfig, expert_class: nn.Module = SwitchTransformersDenseActDense):
        super().__init__()
        # Step 1: Get the correct router according to its class
        self.router = SwitchTransformersTop1Router(config)

        # Step 2: Get the experts
        self.experts = nn.ModuleDict()
        for idx in range(config.num_experts):
            self.experts[f"expert_{idx}"] = expert_class(config)

    def forward(self, hidden_states):
        r"""
        Hold on, this will be slightly tricky to understand In the correct order, a MoE layer does the following:

        1- Gets the `router_mask` from the router. The shape of the mask is `(batch_size, sequence_length, num_expert)`
        and corresponds to the argmax of the `router_probs`. The probabilities are needed in the computation of the
        hidden states : they are broadcasted to the hidden states values (can be interpreted as a scaling factor).

        2- Dispatch the tokens to its associated experts. We do a classic for loop over the experts and assign for each
        expert the corresponding hidden states.

        """

        prev_save = hidden_states.clone()

        # Step 1: Get the router_mask from the router as wel as the probabilities
        router_mask, router_probs, router_logits = self.router(hidden_states)
        expert_index = torch.argmax(router_mask, dim=-1)

        print(torch.allclose(prev_save, hidden_states))
        print(torch.mean(prev_save - hidden_states))

        # The routers introduced might not always map all the tokens, to a router, which means that some hidden states
        # can be unchanged from one layer to another. That is why the hidden states are cloned before updating only the seleced ones.

        next_states = hidden_states.clone()

        router_mask = router_mask.bool()
        batch_size, seq_len, num_experts = router_mask.shape
        idx_mask = router_mask.transpose(1, 2).reshape(batch_size * seq_len, num_experts).sum(dim=0)
        idx_mask = torch.nonzero(idx_mask, as_tuple=True)[
            0
        ].tolist()  # length: number of "activated" expert / value: index
        for idx in idx_mask:
            next_states[router_mask[:, :, idx]] = getattr(self.experts, "expert_{}".format(idx))(
                hidden_states[router_mask[:, :, idx]]
            )

        hidden_states = router_probs * next_states
        return hidden_states, (router_logits, expert_index)

config = SwitchTransformersConfig()
model = MySwitchTransformersSparseMLP(config)

model.train()
in_data = torch.ones(1, 1, 768)
out = model(in_data)
```

The output is 
```bash
False
tensor(-0.0001)
```
which ideally should give True and the mean difference should be zero.

This is because in `SwitchTransformersTop1Router`, the `hidden_states` are multiplied with jitter noise which persists even when you pass it to the experts.

https://github.com/huggingface/transformers/blob/e71a01a104dd663c730e494eb0b6467bb51df357/src/transformers/models/switch_transformers/modeling_switch_transformers.py#L159-L161

### Expected behavior

Ideally, no jitter noise should be present when passing the input to the experts, returning True and the mean difference as 0.

sambhavnoobcoder:
## Issue Description

This pull request addresses a bug in the Switch Transformers architecture where the jitter noise (intended to be applied only for routing decisions) was also being unintentionally applied to the expert inputs.

Fixes : #33969 

## Problem Statement

In Switch Transformers, a small amount of jitter noise is added to the inputs at routing time to ensure route diversity. However, these jittered inputs were incorrectly passed along to the experts, which contradicts the original paper’s design and led to unexpected discrepancies in outputs.

## Root Cause

It was discovered that the code used the same hidden states for both routing and expert processing. When jitter noise was enabled in training mode, it directly modified these states in place, causing the experts to receive noisy inputs.

## Implementation

1. We now clone the original hidden states before applying jitter noise.  
2. A separate copy is used exclusively for computing router logits and probabilities.  
3. The unchanged hidden states are then fed to the experts to maintain the original semantics.

## Screenshot

Screenshot 2025-01-23 at 1 14 26 AM

## Test Cases

1. **test_router_training_mode**  
   • Objective: Ensures that jitter noise is only applied during training.  
   • Checks that outputs differ between consecutive runs (due to noise) but original inputs remain unchanged.

2. **test_router_jitter_noise_separation **  
   • Objective: Verifies that jitter noise affects only the router’s internal computations and not the expert inputs.  
   • Confirms the logits differ when jitter is applied, while the main input stays the same.

3. **test_expert_inputs_consistency**  
   • Objective: Asserts that all expert inputs remain consistent, even when jitter is applied during training.  
   • Uses a forward hook on the first expert to capture its inputs across multiple runs and compares them.

With these changes and test additions, we ensure that Switch Transformers adhere to the original design while preserving backward compatibility and correctness.

**cc :** @ArthurZucker 

original pr refrence : #35847

### DIFF
--- a/src/transformers/models/switch_transformers/modeling_switch_transformers.py
+++ b/src/transformers/models/switch_transformers/modeling_switch_transformers.py
@@ -629,7 +629,7 @@ class SwitchTransformersPreTrainedModel(PreTrainedModel):
             module.weight.data.fill_(factor * 1.0)
         elif isinstance(
             module,
-            (SwitchTransformersModel, SwitchTransformersForConditionalGeneration, SwitchTransformersEncoderModel),
+            SwitchTransformersModel | SwitchTransformersForConditionalGeneration | SwitchTransformersEncoderModel,
         ):
             module.shared.weight.data.normal_(mean=0.0, std=factor * 1.0)
             if hasattr(module, "lm_head") and not self.config.tie_word_embeddings:
--- a/src/transformers/models/switch_transformers/modeling_switch_transformers.py
+++ b/src/transformers/models/switch_transformers/modeling_switch_transformers.py
@@ -102,11 +102,17 @@ class SwitchTransformersTop1Router(nn.Module):
         # https://huggingface.co/papers/2101.03961.
         # We also store the previous dtype to cast back the output to the previous dtype
         self.input_dtype = hidden_states.dtype
-        hidden_states = hidden_states.to(self.dtype)
+
+        # Create a copy for applying jitter noise
+        routing_states = hidden_states.clone()
+        routing_states = routing_states.to(self.dtype)
+
         if self.training and self.jitter_noise > 0:
-            # Multiply the token inputs by the uniform distribution - adding some noise
-            hidden_states *= torch.empty_like(hidden_states).uniform_(1.0 - self.jitter_noise, 1.0 + self.jitter_noise)
-        router_logits = self.classifier(hidden_states)
+            # Apply jitter noise only to the routing copy
+            routing_states *= torch.empty_like(routing_states).uniform_(1.0 - self.jitter_noise, 1.0 + self.jitter_noise)
+
+        # Use jittered states for routing decisions
+        router_logits = self.classifier(routing_states)
 
         # Apply Softmax and cast back to the original `dtype`
         router_probs = nn.functional.softmax(router_logits, dim=-1, dtype=self.dtype).to(self.input_dtype)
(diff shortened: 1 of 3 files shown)


sambhavnoobcoder:
@ArthurZucker can you please have a look at this ? this a copy of an already pre approved branch https://github.com/huggingface/transformers/pull/35847 , looking to get this merged .

PR #41314, opened 2025-10-03
🚨 [v5] Rendundant code in nested configs

zucchini-nlp:
# What does this PR do?

This PR deletes `from_xxx_config` and similar methods from everywhere and instead ensures that the config init can accept subconfigs of all formats (dict or object)

Along the way, I found that not all configs call super init in the end and thus they can't set recursively attn implementation on subconfigs. I added a test for it and fixed all cases


src/transformers/models/bark/configuration_bark.py
```diff
@@ -251,53 +251,40 @@ def __init__(
         **kwargs,
     ):
         if semantic_config is None:
-            semantic_config = {}
-            logger.info("semantic_config is None. initializing the semantic model with default values.")
+            semantic_config = BarkSemanticConfig()
+            logger.info("`semantic_config` is `None`. Initializing the `BarkSemanticConfig` with default values.")
```
Cyrilvallez, line 255: Let's remove those logs instead wdyt? One of the v5 goals is to have fewer and more informatives warnings and logs!

src/transformers/models/mm_grounding_dino/modular_mm_grounding_dino.py
```diff
@@ -288,6 +290,19 @@ def __init__(
...
+    @property
+    def sub_configs(self):
+        sub_configs = {}
+        backbone_config = getattr(self, "backbone_config", None)
+        text_config = getattr(self, "text_config", None)
+        if isinstance(backbone_config, PretrainedConfig):
+            sub_configs["backbone_config"] = type(backbone_config)
+        if isinstance(text_config, PretrainedConfig):
+            sub_configs["text_config"] = type(self.text_config)
+        return sub_configs
```
Cyrilvallez: Was just removed by your other PR no? We should not add it again!

Cyrilvallez (approved):
Nice, happy to clean them up, they are not really useful indeed!
Let's just remove the `logger.info` entries everywhere (I only flagged it once in the comments) at the same time!

src/transformers/models/bark/configuration_bark.py
zucchini-nlp: oke, they are not super informative anyway

src/transformers/models/mm_grounding_dino/modular_mm_grounding_dino.py
zucchini-nlp: yeah, the two PRs will conflict, I will merge the other first

PR #41316, opened 2025-10-03
tests: add dtype/device parity check for GPT‑2

arvinder004:
#### Improve test reliability and coverage:
- Add a minimal dtype/device parity test to catch obvious precision/device regressions.

#### What’s changed- Added dtype/device parity test for GPT‑2:
  - Compares CPU fp32 vs CUDA fp16 logits for identical weights (tolerance ≤ 5e‑3).
  - Gated by `@require_torch_gpu`.
 
#### Files touched
- tests/models/gpt2/test_modeling_gpt2.py

#### Test plan
- Focused: `pytest -q tests/models/gpt2/test_modeling_gpt2.py::GPT2ModelTest::test_dtype_device_parity_logits_fp32_cpu_vs_fp16_cuda`
- Smoke: `pytest -q tests/models/gpt2/test_modeling_gpt2.py -k parity`

Checklist
[x] Minimal diff, follows repo conventions
[x] Tests added/updated
[x] Style/lint passes locally
[x] No runtime code changes

@ydshieh @Rocketknight1 @CyrilVallez

### DIFF
--- a/tests/models/gpt2/test_modeling_gpt2.py
+++ b/tests/models/gpt2/test_modeling_gpt2.py
@@ -281,6 +281,35 @@ class GPT2ModelTest(CausalLMModelTest, unittest.TestCase):
         super().test_training_gradient_checkpointing_use_reentrant_false()
         self.all_model_classes = self.original_all_model_classes
 
+    @require_torch_gpu
+    def test_dtype_device_parity_logits_fp32_cpu_vs_fp16_cuda(self):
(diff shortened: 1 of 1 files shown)


Rocketknight1:
I'm not sure we want extra tests for GPT-2 right now! It's a very old model. Also, most newer models use `bfloat16`, and we don't really guarantee that their `float32` logits are similar (in many cases those models were never designed/tested with `float32`), so I don't know if this works as a global test either!

PR #41317, opened 2025-10-03
Minor security fix for `ssh-runner.yml`

ydshieh:
# What does this PR do?

It's minor as it could not be triggered by external people. But just to be more cautious.


paulinebm (approved):
LGTM

PR #41318, opened 2025-10-03
[causallm tester] automate pipeline mappings + bloom tests

gante:
# What does this PR do?

👉 `CausalLMModelTest`-level changes:
- automate `pipeline_model_mapping` whenever possible -- when `all_model_classes` contains the default classes, then `pipeline_model_mapping` must also contain the default mappings. This should reduce room for human error 🙏 

👉 `Bloom` (the original target of this PR)
- `CausalLMModelTest` refactor
- moves integration tests into the correct tester

👉 misc
- removes `get_large_model_config` in multiple model testers (this function was never being used)

### DIFF
--- a/tests/causal_lm_tester.py
+++ b/tests/causal_lm_tester.py
@@ -144,6 +144,23 @@ class CausalLMModelTester:
             if model_class is not None
         ]
 
+    @property
+    def pipeline_model_mapping(self):
+        # This is the default pipeline mapping.
+        mapping = {
+            "feature-extraction": self.base_model_class,
+            "text-generation": self.causal_lm_class,
+        }
+        if self.question_answering_class is not None:
+            mapping["question-answering"] = self.question_answering_class
+        if self.sequence_classification_class is not None:
+            mapping["text-classification"] = self.sequence_classification_class
+        if self.token_classification_class is not None:
+            mapping["token-classification"] = self.token_classification_class
+        if self.sequence_classification_class is not None:
+            mapping["zero-shot"] = self.sequence_classification_class
+        return mapping
+
     def __init__(
         self,
         parent,
@@ -298,12 +315,20 @@ class CausalLMModelTest(ModelTesterMixin, GenerationTesterMixin, PipelineTesterM
             )
         self.model_tester = self.model_tester_class(self)
         self.config_tester = ConfigTester(self, config_class=self.model_tester.config_class)
+
+        if self.pipeline_model_mapping is None:
+            # If `all_model_classes` is not the default, maybe there are more pipeline mappings to be set.
+            if self.all_model_classes is not None:
+                raise ValueError(
+                    "Testes that inherit from `CausalLMModelTest` and set `all_model_classes` must manually set "
+                    "`pipeline_model_mapping`."
+                )
+            # Otherwise, we know the pipeline mapping is the default.
+            else:
+                self.pipeline_model_mapping = self.model_tester.pipeline_model_mapping
+
         if self.all_model_classes is None:
             self.all_model_classes = self.model_tester.all_model_classes
-        if self.pipeline_model_mapping is None:
-            raise ValueError(
-                "You have inherited from CausalLMModelTest but did not set the pipeline_model_mapping attribute."
-            )
 
     def test_config(self):
         self.config_tester.run_common_tests()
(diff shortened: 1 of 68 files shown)


tests/models/bloom/test_modeling_bloom.py
```diff
@@ -363,177 +188,43 @@ def test_bloom_lm_head_model(self):
...
-    def test_bloom_gradient_checkpointing(self):
-        config_and_inputs = self.model_tester.prepare_config_and_inputs()
-        self.model_tester.create_and_check_forward_and_backwards(*config_and_inputs, gradient_checkpointing=True)
-
     def test_bloom_weight_initialization(self):
         config_and_inputs = self.model_tester.prepare_config_and_inputs()
         self.model_tester.create_and_check_bloom_weight_initialization(*config_and_inputs)
 
-    @slow
-    def test_model_from_pretrained(self):
```
gante, line 383: this test is silly -- it's a slow test that's implicitly tested in all integration tests

tests/models/bloom/test_modeling_bloom.py
```diff
@@ -363,177 +188,43 @@ def test_bloom_lm_head_model(self):
...
-        model_name = "bigscience/bigscience-small-testing"
-        model = BloomModel.from_pretrained(model_name)
-        self.assertIsNotNone(model)
-
-    @slow
-    @require_torch_accelerator
-    def test_simple_generation(self):
```
gante, line 390: tests below: they are integration tests, moved to the other class

tests/models/bloom/test_modeling_bloom.py
```diff
@@ -363,177 +188,43 @@ def test_bloom_lm_head_model(self):
...
-        self.assertListEqual(generated_text, EXPECTED_GENERATIONS)
-
     @unittest.skip("Bloom needs a 2D attention for alibi")
     def test_custom_4d_attention_mask(self):
         pass
 
 
 @require_torch
-class BloomEmbeddingTest(unittest.TestCase):
-    """
```
gante, line 510: this comment pertains the first test only, not the whole tester.

also, the tester name was innacurate

ydshieh:
We had `pipeline_model_mapping` being explicitly defined as we wanted it to be transparent and easily to see. But I think it's worth the change now. Thanks

tests/causal_lm_tester.py
```diff
@@ -144,6 +144,23 @@ def all_model_classes(self):
...
+        }
+        if self.question_answering_class is not None:
+            mapping["question-answering"] = self.question_answering_class
+        if self.sequence_classification_class is not None:
+            mapping["text-classification"] = self.sequence_classification_class
+        if self.token_classification_class is not None:
+            mapping["token-classification"] = self.token_classification_class
+        if self.sequence_classification_class is not None:
+            mapping["zero-shot"] = self.sequence_classification_class
+        return mapping
```
ydshieh, line 162: not a big thing, but I am starting to think why we made all these `all_model_classes` and `pipeline_model_mapping` (in Tester or Test classes) as instance method/property.

These are more at class level, right?

tests/causal_lm_tester.py
```diff
@@ -299,12 +316,20 @@ def setUp(self):
...
+        if self.pipeline_model_mapping is None:
+            # If `all_model_classes` is not the default, maybe there are more pipeline mappings to be set.
+            if self.all_model_classes is not None:
+                raise ValueError(
+                    "Testes that inherit from `CausalLMModelTest` and set `all_model_classes` must manually set "
+                    "`pipeline_model_mapping`."
+                )
+            # Otherwise, we know the pipeline mapping is the default.
+            else:
+                self.pipeline_model_mapping = self.model_tester.pipeline_model_mapping
```
ydshieh, line 328: same here, this is more at class level instead of instance level

gante:
@ydshieh yes, they should be class-level attributes -- currently, all this logic is re-run at each test, as opposed to once per class.

I can take care of that in a subsequent PR :D (note that this PR doesn't introduce this behavior)

ydshieh:
ok in a separate PR :-)

Cyrilvallez (approved):
Very nice! Feel free to fix the conflicts and merge!

PR #41319, opened 2025-10-03
Use torch._check instead of a test in Gemma3Model

xadupre:
# What does this PR do?

``torch.export.export`` fails because of a test checking two dimension are equal. This PR replaces the test by ``torch._check``.


ydshieh:
> torch.export.export fails 

does it fails even if the dimension are equal ..?

ArthurZucker (approved):
Yeah very happy to fix export! Can we ship this to all other models that use the check?

PR #41320, opened 2025-10-03
Security/fuyu

molbap:
# What does this PR do?

Remove a reference to a compromised github repo. 

For reference, all conversion scripts are excluded from releases. https://github.com/huggingface/transformers/blob/217ff1e4ef6b006d601d0fa834c2bd68c79b66cd/utils/release.py#L135-L143


ydshieh (approved):
👍

PR #41321, opened 2025-10-03
[v5] Delete left traces of feature extractor

zucchini-nlp:
# What does this PR do?

After rebasing https://github.com/huggingface/transformers/pull/40793 and getting new changes on `TypedDict`, I found out that some traces from feature extractor were still left. Thanks for type validation we can catch inconsistencies easily :)

This PR goes over the deprecated properties as well and deletes them. Also it deletes `current_processor` which was also deprecated for v5


gante (approved):
🧹 🧹 🧹

PR #41322, opened 2025-10-03
Fix lr_scheduler_parsing

SunMarc:
# What does this PR do?

This PR fixes `lr_scheduler_kwargs` type to `Optional[Union[dict, str]] `as HfArgumentParser doesn't handled well args that needs to parsed into dict. For now, it is better to just follow what is done for fsdp_config for example.


MekkCyber (approved):
Okay sounds good!

PR #41323, opened 2025-10-03
deprecate `overwrite_output_dir`

SunMarc:
# What does this PR do?
This PR deprecates `overwrite_output_dir`. This arg has no utility at all tbh as this was kind of replaced by `resume_from_checkpoint`. It was used to overwrite the checkpoint when we didn't want to resume the training. It was only used in examples. Sylvain also mentioned that in this thread https://discuss.huggingface.co/t/no-skipping-steps-after-loading-from-checkpoint/6880/14


Cyrilvallez (approved):
lgtm!

PR #41326, opened 2025-10-03
Deprecate warmup_ratio

SunMarc:
# What does this PR do?
This PR deprecates `warmup_ratio`. Instead we ask the user to use `warmup_steps` instead as we now allow passing float values. This is something that was already done for a few args like `eval_steps` .

### DIFF
--- a/ISSUES.md
+++ b/ISSUES.md
@@ -153,7 +153,7 @@ You are not required to read the following guidelines before opening an issue. H
     cd examples/seq2seq
     torchrun --nproc_per_node=2 ./finetune_trainer.py \
     --model_name_or_path sshleifer/distill-mbart-en-ro-12-4 --data_dir wmt_en_ro \
-    --output_dir output_dir --overwrite_output_dir \
+    --output_dir output_dir \
     --do_train --n_train 500 --num_train_epochs 1 \
     --per_device_train_batch_size 1  --freeze_embeds \
     --src_lang en_XX --tgt_lang ro_RO --task translation \
--- a/ISSUES.md
+++ b/ISSUES.md
@@ -153,7 +153,7 @@ You are not required to read the following guidelines before opening an issue. H
     cd examples/seq2seq
     torchrun --nproc_per_node=2 ./finetune_trainer.py \
     --model_name_or_path sshleifer/distill-mbart-en-ro-12-4 --data_dir wmt_en_ro \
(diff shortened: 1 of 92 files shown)


qgallouedec:
we use `warmup_ratio` a lot in our recipes/configs, but when I think about it, it might make more sense to use `warmup_step` instead., I'd be curious to hear the feedback of @edbeeching @lewtun on this.

SunMarc:
> we use warmup_ratio a lot in our recipes/configs, but when I think about it, it might make more sense to use warmup_step instead., I'd be curious to hear the feedback of @edbeeching @lewtun on this.

I don't mind keeping it if you think this will lead to too many complications !

edbeeching:
> > we use warmup_ratio a lot in our recipes/configs, but when I think about it, it might make more sense to use warmup_step instead., I'd be curious to hear the feedback of @edbeeching @lewtun on this.
> 
> I don't mind keeping it if you think this will lead to too many complications !

I think it is fine as long as the warmup is the same for `warmup_ratio=0.1` vs `warmup_steps=0.1` (for example)

lewtun:
I agree with @edbeeching - this is a simple change we can propagate to our recipes

SunMarc:
Perfect, I'll do the change for v5 then. 

After this PR, if users are trying to pass a `float` value for `warmup_steps`:

- transformers v4: they will receive the following error `("warmup_steps must be of type int and must be 0 or a positive integer.")`. For clarify, maybe we can catch this in TRL and raise an error ? @qgallouedec 

- transformers v5: same behavior as we had with `warmup_ratio `

MekkCyber (approved):
Thanks for the cleaning

BenjaminBossan:
@SunMarc I noticed that the deprecation messages are confusing, one says deprecation happens in v5 and another in v5.2.

Also, for clarification, if I have `warmup_ratio=0.1`, I could change it to `warmup_steps=0.1` but only if I have transformers v5, in v4, float would not be valid?

SunMarc:
> Also, for clarification, if I have warmup_ratio=0.1, I could change it to warmup_steps=0.1 but only if I have transformers v5, in v4, float would not be valid?

Yeah, float would not be valid as this would only work with transformers v5. Also indeed, the deprecation msg are confusing. It should be v5.2.

PR #41332, opened 2025-10-03
fix asr ut failures

yao-matrix:
4 tests failed with AttributeError: 'AudioDecoder' object has no attribute 'copy':

`pytest -rA tests/pipelines/test_pipelines_automatic_speech_recognition.py::AutomaticSpeechRecognitionPipelineTests::test_speculative_decoding_whisper_non_distil
pytest -rA tests/test_pipeline_mixin.py::AutomaticSpeechRecognitionPipelineTests::test_speculative_decoding_whisper_non_distil
pytest -rA tests/pipelines/test_pipelines_automatic_speech_recognition.py::AutomaticSpeechRecognitionPipelineTests::test_whisper_prompted
pytest -rA tests/test_pipeline_mixin.py::AutomaticSpeechRecognitionPipelineTests::test_whisper_prompted`

after fixing, they all passed

@SunMarc  @ydshieh , pls help review, thx very much.


src/transformers/generation/logits_process.py
```diff
@@ -1915,7 +1915,7 @@ def __init__(self, suppress_tokens, device: str = "cpu"):
     @add_start_docstrings(LOGITS_PROCESSOR_INPUTS_DOCSTRING)
     def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> torch.FloatTensor:
         vocab_tensor = torch.arange(scores.shape[-1], device=scores.device)
-        suppress_token_mask = isin_mps_friendly(vocab_tensor, self.suppress_tokens)
+        suppress_token_mask = isin_mps_friendly(vocab_tensor, self.suppress_tokens.to(scores.device))
         scores = torch.where(suppress_token_mask, -float("inf"), scores)
```
yao-matrix, line 1919: In multi-device cases(like put 2 devices to run):
in current implementation, in assistant decoding case, assistant model will reuse main model's `SuppressTokensLogitsProcessor`, which place the `suppress_tokens` in the same device as `input_tensor` (which is `device 0`).  assistant model will ingest `encoder_outputs` of the main model and do the decoder(in whisper case), while `encoder_outputs` may in `device 1` but main model's `suppress_tokens` which is main model's is in `device 0`, so lead to `RuntimeError`: 

> RuntimeError: Expected all tensors to be on the same device, but got test_elements is on xpu:0, different from other tensors on xpu:1 (when checking argument in method wrapper_XPU_isin_Tensor_Tensor)

So based on current implementation(that assistant model shares main model's `SuppressTokensLogitsProcessor`), I move `suppress_tokens` to `scores.device` while doing `isin`.

SunMarc (approved):
Thanks for fixing ! cc @eustlb for final checks

tests/pipelines/test_pipelines_automatic_speech_recognition.py
```diff
@@ -1133,8 +1133,8 @@ def test_speculative_decoding_whisper_non_distil(self):
             num_beams=1,
         )
 
-        transcription_non_ass = pipe(sample.copy(), generate_kwargs={"assistant_model": assistant_model})["text"]
-        transcription_ass = pipe(sample)["text"]
+        transcription_ass = pipe(sample.clone().detach(), generate_kwargs={"assistant_model": assistant_model})["text"]
+        transcription_non_ass = pipe(sample)["text"]
```
eustlb, line 1137: thanks for catching the incorrect inversion here!

eustlb (approved):
LGTM! thanks for your help 🤗

PR #41335, opened 2025-10-03
Add `logits_to_keep` to many older CausalLM models

### LINKED ISSUE #40984: Adding `logits_to_keep` to older models
Hi, while working with BioGPT I came across #39016 to not compute logits of the entire sequence when it's not needed. I was wondering if you'd be open to a PR applying the same improvement to BioGPT and, while I'm at it, any other `GenerationMixin` models that would benefit from this. I understand some of these models might be a bit obsolete and some could benefit from a refactor to modular instead, but I figure while they're there it's worth doing.

philiproeleveld:
# What does this PR do?

Fixes #40984 

Adds `logits_to_keep` to many (older) `ForCausalLM` models that inherit from `GenerationMixin`.
Also consistently renames to `loss` and `logits`, and removes some code for float casting and mapping labels to the logits' device for models where that is already handled by the loss function (e.g. `gpt_neo`).

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

members/contributors who may be interested in your PR. @Rocketknight1

### DIFF
--- a/src/transformers/models/blip/modeling_blip.py
+++ b/src/transformers/models/blip/modeling_blip.py
@@ -996,7 +996,6 @@ class BlipForQuestionAnswering(BlipPreTrainedModel, GenerationMixin):
         attention_mask: Optional[torch.LongTensor] = None,
         labels: Optional[torch.LongTensor] = None,
         interpolate_pos_encoding: bool = False,
-        logits_to_keep: Union[int, torch.Tensor] = 0,
         **kwargs: Unpack[TransformersKwargs],
     ) -> Union[tuple, BlipTextVisionModelOutput]:
         r"""
@@ -1068,7 +1067,6 @@ class BlipForQuestionAnswering(BlipPreTrainedModel, GenerationMixin):
             encoder_attention_mask=attention_mask,
             labels=labels,
             reduction="mean",
-            logits_to_keep=logits_to_keep,
             **kwargs,
         )
 
--- a/src/transformers/models/blip/modeling_blip.py
+++ b/src/transformers/models/blip/modeling_blip.py
@@ -826,6 +826,7 @@ class BlipForConditionalGeneration(BlipPreTrainedModel, GenerationMixin):
         attention_mask: Optional[torch.LongTensor] = None,
         labels: Optional[torch.LongTensor] = None,
         interpolate_pos_encoding: bool = False,
+        logits_to_keep: Union[int, torch.Tensor] = 0,
         **kwargs: Unpack[TransformersKwargs],
     ) -> Union[tuple, BlipForConditionalGenerationModelOutput]:
         r"""
@@ -862,6 +863,7 @@ class BlipForConditionalGeneration(BlipPreTrainedModel, GenerationMixin):
             encoder_hidden_states=image_embeds,
             labels=labels,
             reduction="mean",
+            logits_to_keep=logits_to_keep,
             **kwargs,
         )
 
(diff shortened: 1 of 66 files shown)


philiproeleveld:
I was wondering if I could also add this to the many seq2seq models that also inherit from `GenerationMixin`? T5 for example. In theory they would benefit when the user provides a very large `decoder_input_ids`, but that's not really what they're designed for...

src/transformers/models/xlm_roberta/modeling_xlm_roberta.py
```diff
@@ -989,23 +990,18 @@ def forward(
...
         if labels is not None:
-            # move labels to correct device
-            labels = labels.to(prediction_scores.device)
-            lm_loss = self.loss_function(
-                prediction_scores,
-                labels,
-                vocab_size=self.config.vocab_size,
-                **kwargs,
-            )
+            loss = self.loss_function(logits=logits, labels=labels, vocab_size=self.config.vocab_size, **kwargs)
```
Rocketknight1, line 829: Some of these cases slightly change other behaviour (e.g. not moving labels to `logits.device`). Have you checked that this is equivalent?

Rocketknight1 (commented):
Mostly looks good, with one comment that applies to a couple of models!

This is a fairly big change that standardizes a lot of older models with the modern API so cc core maintainers @arthurzucker @cyrilvallez

philiproeleveld:
I'd like some guidance on the failing test `GitModelTest::test_forward_with_logits_to_keep`. The problem is that test assumes sequence length of logits is equal to sequence length of input_ids, but that's not true for Git (because it concatenates image tokens). I see three potential fixes, but not sure what's preferable:
1. Copy the test to `GitModelTest` and modify
2. Add an exception to `GenerationTesterMixin`
3. Skip it (If so, what reason do I give?)

src/transformers/models/xlm_roberta/modeling_xlm_roberta.py
philiproeleveld: Yes, I removed the device mapping and float casting in these cases because they use `self.loss_function` (I.e. `ForCausalLMLoss`) and the same things are already done in that function too. I assumed it was added there later and these models were just missed in the refactor.
But I haven't actually ran all the code to check physically. Should I do that?

Rocketknight1:
@philiproeleveld I'd just skip it in that case! For the reason you can say that the sequence lengths are not equivalent and so the test doesn't work.

src/transformers/models/blip/modeling_blip.py
```diff
@@ -994,6 +996,7 @@ def forward(
         attention_mask: Optional[torch.LongTensor] = None,
         labels: Optional[torch.LongTensor] = None,
         interpolate_pos_encoding: bool = False,
+        logits_to_keep: Union[int, torch.Tensor] = 0,
```
Cyrilvallez: Not sure it makes much sense here in the ForQuestionAnswering model - let's revert this one!

src/transformers/models/kosmos2/modeling_kosmos2.py
```diff
@@ -1325,14 +1326,12 @@ def forward(
             `[-100, 0, ..., config.vocab_size]` (see `input_ids` docstring) Tokens with indices set to `-100` are
             ignored (masked), the loss is only computed for the tokens with labels in `[0, ..., config.vocab_size]`
         """
-        return_dict = return_dict if return_dict is not None else self.config.use_return_dict
-
```
Cyrilvallez, line 1329: Nice catch! Can be removed from the signature as well

Cyrilvallez (approved):
Very very nice PR! Very clean, not much to say 🤗 Only 2 comments to address before we can merge!

src/transformers/models/kosmos2/modeling_kosmos2.py
philiproeleveld: Done! Thank you for the kind words 😄

Cyrilvallez:
Merging! Thanks a lot!

PR #41347, opened 2025-10-04
Fix flash_attention.py: wrong argument passing for attn_implementation

TKONIY:
# What does this PR do?
The name of the attn type argument for `_flash_attention_forward()` should be `implementation`, instead of `attn_implementation` which currently uses in the function call. This would result in wrong type specification.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## CC

### DIFF
--- a/src/transformers/integrations/flash_attention.py
+++ b/src/transformers/integrations/flash_attention.py
@@ -76,7 +76,7 @@ def flash_attention_forward(
         softcap=softcap,
         use_top_left_mask=_use_top_left_mask,
         target_dtype=target_dtype,
-        attn_implementation=module.config._attn_implementation,
+        implementation=module.config._attn_implementation,
         layer_idx=module.layer_idx if hasattr(module, "layer_idx") else None,
         **kwargs,
     )
--- a/src/transformers/integrations/flash_attention.py
+++ b/src/transformers/integrations/flash_attention.py
@@ -76,7 +76,7 @@ def flash_attention_forward(
         softcap=softcap,
         use_top_left_mask=_use_top_left_mask,
         target_dtype=target_dtype,
-        implementation=module.config._attn_implementation,
+        attn_implementation=module.config._attn_implementation,
         layer_idx=module.layer_idx if hasattr(module, "layer_idx") else None,
         **kwargs,
     )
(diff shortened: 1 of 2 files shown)


Cyrilvallez:
Oh good catch there is indeed the wrong kwarg being passed on! I would prefer if you can change the name of the kwarg in `_flash_attention_forward` though, it would be more explicit!

TKONIY:
> Oh good catch there is indeed the wrong kwarg being passed on! I would prefer if you can change the name of the kwarg in `_flash_attention_forward` though, it would be more explicit!

Thank you. I have changed that. Please check.

vasqu (approved):
Can you also fix the docs naming then at https://github.com/huggingface/transformers/blob/0452f28544f3626273d25f07f83c0e5f7da2d47a/src/transformers/modeling_flash_attention_utils.py#L567-L568

LGTM otherwise :hugs:

TKONIY:
> Can you also fix the docs naming then at
> 
> https://github.com/huggingface/transformers/blob/0452f28544f3626273d25f07f83c0e5f7da2d47a/src/transformers/modeling_flash_attention_utils.py#L567-L568
> 
> LGTM otherwise 🤗

Thank you! Done!

Cyrilvallez (approved):
Thanks!! 🤗

ArthurZucker:
https://github.com/huggingface/transformers/blob/update-from-pretrained/src/transformers/integrations/hub_kernels.py#L214-L214 is where we use `implementation`.  If you do this it won't fallback to `kernels`, we need to make sure we use the one passed in load and register

TKONIY:
> https://github.com/huggingface/transformers/blob/update-from-pretrained/src/transformers/integrations/hub_kernels.py#L214-L214 is where we use `implementation`.  If you do this it won't fallback to `kernels`, we need to make sure we use the one passed in load and register
> 
> 

So would it be better if I rollback to the fix that simply change `attn_implementation=` to `implementation=`？

vasqu:
The link is broken, I think Arthur meant https://github.com/huggingface/transformers/blob/caa14e7dabb086f167c14b7eecadc2ba9db25eb6/src/transformers/integrations/hub_kernels.py#L214-L215

Yes, we should also change the kwarg there, totally forgot there, e.g. `kernel_function = partial(attention_wrapper, attn_implementation=kernel) `. But it's a bit messier tbh and I don't think we actually use the kwarg much at all except on first call (which would happen if someone does something custom with fa interface we have) --> the forced lazy import should load the correct kernel (checking in a second) and as we already loaded it, we never change it again there.

Edit: Still loads the correct kernel implementation, checked with `kernels-community/flash-attn3`

ArthurZucker:
Sounds good! I think we need a single point where we do the kernel loading so let' s clear it up!

vasqu:
Kind of a followup to this PR with #41427 where loading is now more flexible

PR #41354, opened 2025-10-05
Fixed tiny incorrect import in `gemma3`

Sai-Suraj-27:
# What does this PR do?
Fixes tiny incorrect import in `gemma3`. This is the correct class name https://github.com/huggingface/transformers/blob/e11a00a16f925b7d3b52f5007bdce3464edb361f/src/transformers/models/paligemma/modeling_paligemma.py#L67

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@gante @zucchini-nlp

### DIFF
--- a/src/transformers/models/gemma3/modular_gemma3.py
+++ b/src/transformers/models/gemma3/modular_gemma3.py
@@ -44,7 +44,7 @@ from ..gemma2.modeling_gemma2 import (
     eager_attention_forward,
 )
 from ..paligemma.modeling_paligemma import (
-    PaligemmaCausalLMOutputWithPast,
+    PaliGemmaCausalLMOutputWithPast,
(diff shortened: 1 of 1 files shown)


zucchini-nlp (approved):
Thanks, interesting how modular import worked with a typo haha, looks like it is case insensitive? 

just fyi @Cyrilvallez , in case you think it is a bug in modular and should be fixed :)

gante:
indeed, surprised this didn't blow things up :o 

thank you for the fix @Sai-Suraj-27 🙏

Cyrilvallez:
Ouhh, indeed quite interesting! Thanks for the ping, will investigate! (it's probably a regex that is somehow case-insensitive indeed)

PR #41355, opened 2025-10-05
Better typehints for `apply_chat_template`

Samoed:
# What does this PR do?
Make better typehints for `apply_chat_template`. Now IDEs and type checkers will correctly infer types for this code.

```py
tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen3-0.6B")

# Returns raw string when tokenize=False
raw_prompt: str = tokenizer.apply_chat_template(
    [{"role": "user", "content": "Hello!"}],
    tokenize=False,
)

# Returns token ids when tokenize=True (default) and return_dict=False
token_ids: list[int] = tokenizer.apply_chat_template(
    [{"role": "user", "content": "Hello!"}],
)

# Returns BatchEncoding when return_dict=True
batch_encoding: BatchEncoding = tokenizer.apply_chat_template(
    [{"role": "user", "content": "Hello!"}],
    return_dict=True,
)
```

Similarly, can be done for [`ProcessorMixin.apply_chat_template`](https://github.com/huggingface/transformers/blob/e11a00a16f925b7d3b52f5007bdce3464edb361f/src/transformers/processing_utils.py#L1559), but I'm not sure how to add to it, because it's using kwargs instead of arguments   

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker and @itazap

### DIFF
--- a/src/transformers/models/auto/tokenization_auto.py
+++ b/src/transformers/models/auto/tokenization_auto.py
@@ -940,7 +940,9 @@ class AutoTokenizer:
 
     @classmethod
     @replace_list_option_in_docstrings(TOKENIZER_MAPPING_NAMES)
-    def from_pretrained(cls, pretrained_model_name_or_path, *inputs, **kwargs):
+    def from_pretrained(
+        cls, pretrained_model_name_or_path, *inputs, **kwargs
+    ) -> Union[PreTrainedTokenizer, PreTrainedTokenizerFast]:
         r"""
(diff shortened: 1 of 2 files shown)


Rocketknight1 (approved):
We generally don't use `@overload` too much as it bloats files, but this is a case where it makes sense, since `apply_chat_template` is a common function but has very different return types depending on how it's called.

Rocketknight1:
(as always, core maintainers yell at me if you don't like it and i'll revert :sweat_smile: )

PR #41359, opened 2025-10-06
Fix trainer for py3.9

SunMarc:
# What does this PR do?

This PR fixes trainer compatibility with py3.9. To be added in a patch for the latest release. Asked by optimum folks cc @IlyasMoutawwakil

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -5206,7 +5206,7 @@ class Trainer:
                     self.model.hf_quantizer.quantization_config.bnb_4bit_quant_storage, override=True
                 )
 
-    def _get_num_items_in_batch(self, batch_samples: list, device: torch.device) -> int | None:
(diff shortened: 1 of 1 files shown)


Cyrilvallez (approved):
I just merged https://github.com/huggingface/transformers/pull/41268 on main, if you prefer ton only add it on the v4 branch.
No strong opinion though, we can merge on main as well since it's only typing!

SunMarc:
For this one, let's merge on main and cherry pick it on v4 branch as this is just typing !

Samoed:
Can you close this issue https://github.com/huggingface/transformers/issues/41339 after merge?

Rocketknight1:
Hi, this one was my bad! We've been updating for Py3.10 because Py3.9 is EOL next week. Do we really want to revert to support it?

SunMarc:
Don't worry @Rocketknight1, it was for a patch that was released last week

Issue #41363, opened 2025-10-06
RT-Detr docs should reflect fixed 640x640 input size


konstantinos-p:
The authors of RT-Detr mention that the model was trained on 640x640 images and was meant to be used for inference on 640x640 images. Also, the current implementation has certain quirks that make training/inferring on images of different sizes problematic. For example, the pixel masks used for batching images of varying sizes are discarded.

https://github.com/huggingface/transformers/blob/0452f28544f3626273d25f07f83c0e5f7da2d47a/src/transformers/models/rt_detr/modeling_rt_detr.py#L1645

The above are not clear in the current docs. I'll open a PR which adds a few lines in the docs to notify users about these issues.

YashKerkarTech04:
Hii, I like to work on this issue. Issue number #41363

konstantinos-p:
Thanks @YashKerkarTech04 however as you can see I already have a PR open for this issue.

Rocketknight1:
Fixed by #41363 I think!

PR #41364, opened 2025-10-06
Change RT-Detr docs to reflect fixed 640x640 input size

konstantinos-p:
# What does this PR do?

The authors of RT-Detr mention that the model was trained on 640x640 images and was meant to be used for inference on 640x640 images. Also, the current implementation has certain quirks that make training/inferring on images of different sizes problematic. For example, the pixel masks used for batching images of varying sizes are discarded. I've added a few lines in the docs to notify the user about these issues.

``
proj_feats = [self.encoder_input_proj[level](source) for level, (source, mask) in enumerate(features)]
``

Fixes # (issue)
#41363

## Before submitting

## Who can review?

Models:

- vision models: @yonigozlan @molbap
- documentation: @stevhliu


stevhliu (approved):
LGMT, we can merge once @yonigozlan has also had a chance to take a look :)

docs/source/en/model_doc/rt_detr.md
```diff
@@ -40,7 +40,7 @@ The model version was contributed by [rafaelpadilla](https://huggingface.co/rafa
...
+Initially, an image is processed using a pre-trained convolutional neural network, specifically a Resnet-D variant as referenced in the original code. This network extracts features from the final three layers of the architecture. Following this, a hybrid encoder is employed to convert the multi-scale features into a sequential array of image features. Then, a decoder, equipped with auxiliary prediction heads is used to refine the object queries. This process facilitates the direct generation of bounding boxes, eliminating the need for any additional post-processing to acquire the logits and coordinates for the bounding boxes. The model is meant to be used on images resized to a size 640x640 with the corresponding ImageProcessor. Reshaping to other sizes will generally degrade performance. Also, batching images of variable size is not supported in the current implementation.
```
yonigozlan: > Also, batching images of variable size is not supported in the current implementation.

Not sure if this is worth mentioning, as this is the case for all models, since models expect tensor batches in inputs. Unless I'm misunderstanding what you mean here?

yonigozlan (commented):
Agreed to add that the model usually expects images of size 640x640, however I'm not sure I understand what you mean with:

> For example, the pixel masks used for batching images of varying sizes are discarded. I've added a few lines in the docs to notify the user about these issues.

> `proj_feats = [self.encoder_input_proj[level](source) for level, (source, mask) in enumerate(features)]`

This seems to show that `pixel_mask` is actually useless as an input to the model regardless of the masks size, not sure why it was added in the first place, maybe a relic from Detr... I'll have to check if they are any APIs that use it before removing it entirely, but I think this is out of scope for this PR ;)

docs/source/en/model_doc/rt_detr.md
konstantinos-p: Regarding batching with variable image sizes, I've encountered the following problem: 

The standard pipeline as I understand it (from notebooks etc) is to fix some max image dimensions for the batch then crop larger images and padd smaller images while generating a mask to ignore the padded areas. At the end, all the images in the batch will have the same dimensions and you should be able to ignore the padded areas in the loss using the masks. 

I think this happens correctly in the vanilla detr but the masks are ignored for rt-detr. What ends up happening is that you train on the padded areas and from my experience this introduces spurious correlations when doing inference. I think the current docs don't accurately reflect this because for rt-detr you would always resize to 640x640 and this wouldn't be a problem + some code just carried over from the vanilla detr.

I could always move this to another PR if you think code changes are necessary. Let me know how you think about it!

konstantinos-p: @yonigozlan I removed the reference to batching so as to make merging this PR easier. However, I still believe that the behavior of the model when batching with images of different sizes inside the same batch is not the what we'd expect from the docs. Happy to address this in another PR.

yonigozlan: > I think this happens correctly in the vanilla detr but the masks are ignored for rt-detr. What ends up happening is that you train on the padded areas and from my experience this introduces spurious correlations when doing inference. I think the current docs don't accurately reflect this because for rt-detr you would always resize to 640x640 and this wouldn't be a problem + some code just carried over from the vanilla detr.
> 
> I could always move this to another PR if you think code changes are necessary. Let me know how you think about it!

Ah I see what you mean. Indeed that's an issue that should be either clarified in the docs and code, or even better fixed. At the moment accepting masks as inputs but not using them is misleading, so in any case a change to the code would be needed. But yes this is a bit outside the scope of this PR, so feel free to open another one to address the problem, or open an issue to signal it :)

yonigozlan (approved):
Looks good to merge with the last changes to me!

PR #40884, opened 2025-09-15
Any to any pipeline and auto-mapping

zucchini-nlp:
# What does this PR do?

Adds any-to-any as a pipeline and in auto classes so that we can have a single mapping for all multimodal models. The model mapping is almost same as image-text-to-text, with inclusion of audio-LLM and omni-LLM. I hope I added all audio models, but lmk if anything is missing from recent ones

Fixes https://github.com/huggingface/transformers/issues/40302 and fixes https://github.com/huggingface/transformers/issues/37794

PR #41366, opened 2025-10-06
Add in-out modalities as class attribute per model

zucchini-nlp:
# What does this PR do?

Branches out from https://github.com/huggingface/transformers/pull/40884#discussion_r2400964015 to make review and merge faster

Adds a class attr for each model to indicate the supported input and output modalities. Out modalities will be `None` in case the model is not generative and "text" in most other cases. We have only a few models that can generate audio and image in the output. Note that for encoder decoder models that whisper the input modalities will contain both encoder ("audio") and decoder ("text") modalities

This will be used firstly for the pipeline and we can extend usage later to better testing suite and in preparing inputs better in generation with multimodal LLMs (e.g. if we move multimodal encoding to `GenrationMixin._prepare_multimodal_encodings`). No test added at this point, because there is nothing to test


src/transformers/models/aimv2/modeling_aimv2.py
```diff
@@ -393,6 +393,7 @@ class Aimv2PreTrainedModel(PreTrainedModel):
 
     config: Aimv2Config
     base_model_prefix = "aimv2"
+    input_modalities = "image"
```
gante, line 397: suggestion: use a `tuple` always, even when length = 1

- single possible type = simpler usage
- immutable
- `tuples` can be used as dictionary keys. In the future, this might be useful to do modality-specific operations (e.g. `SOME_MAPPING_OF_FUNCTIONS[model.input_modalities](**kwargs)`)

src/transformers/modeling_utils.py
```diff
@@ -2224,6 +2228,20 @@ def base_model(self) -> nn.Module:
...
+        Returns a list of output modalities that a model can generate. For non-generative models
+        returns a `None`. Multimodal models that can output several modalities or non-text modalities
+        should overwrite this method.
+
+        Returns:
+            `Union[str, list[str]]`: Output modalities supported for models that can call `.generate()`.
+        """
+        if cls.can_generate():
+            return "text"
+        return None
```
gante: I think one of two things should happen:
1. [my preference] `output_modalities` can be used on all models, i.e. it is not limited to generative models. I suspect this is a useful piece of info to have in model-agnostic code, enabling better error handling and other functionality. (unimplemented cases could throw an exception for now?)
2. If we truly only want to use this in models that inherit `GenerationMixin`, then this function should be moved to `GenerationMixin`. Otherwise, we're tangling the classes (= bad practice).

gante (commented):
I wonder if we can automate these variables, instead of having to manually define them. E.g. can we look at the signature of `forward` and, based on arguments present / type hints, determine modalities?

(fewer manual flags = smaller odds of human error = fewer bugs)

zucchini-nlp:
> E.g. can we look at the signature of forward and, based on arguments present / type hints, determine modalities?

yeah, also thought of it. It is doable for most models but there are some tricky ones as well, For example we don't have a consistent naming habit for video modality or we have no way to say what is being output by a model that has an overwritten `generate()`. We can have have a default to `input_modalities` as well similar to `output_modalities`, but then manually overwrite all models where the pattern does not match

src/transformers/modeling_utils.py
zucchini-nlp: Hmm, if we define it for all models what would it mean for models that output encodings (e.g. CLIP)? I could not think of use-case for that tbh

gante:
> We can have have a default to input_modalities as well similar to output_modalities, but then manually overwrite all models where the pattern does not match

imo this would be an improvement :) also an incentive to nudge contributors towards standard names and definitions!

gante:
But check with @ArthurZucker before committing code!

(See [tenet](https://huggingface.co/spaces/transformers-community/Transformers-tenets) number 4: standardization for model definitions, abstractions for infra. I would place this under infra)

adarshxs:
I think this is a decent approach. waiting on this PR as it helps quite a lot - rather than depending on heuristics or some sort of registry we want to get the input-output modalities supported by the model via config. I hope we can adopt this as a standard soon @ArthurZucker

zucchini-nlp:
@bot /style

src/transformers/modeling_utils.py
ArthurZucker: yeah In that case let's move it to `GenerateMixin` ?

ArthurZucker (approved):
Explicit and non automatic looks better for now IMO 😉

src/transformers/modeling_utils.py
zucchini-nlp: done! If model is not generative, it will not have class attribute `output_modalities`

PR #41370, opened 2025-10-06
[`CB`] Refactors the way we access paged

ArthurZucker:
# What does this PR do?
The user decides on the function, but CB handles itself which "interface" wrapper to use. This should make stuff easier

### DIFF
--- a/src/transformers/generation/continuous_batching/continuous_api.py
+++ b/src/transformers/generation/continuous_batching/continuous_api.py
@@ -27,6 +27,7 @@ from tqdm import tqdm
 
 from ...configuration_utils import PreTrainedConfig
 from ...generation.configuration_utils import GenerationConfig
+from ...integrations.hub_kernels import load_and_register_kernel
 from ...utils.logging import logging
 from ...utils.metrics import ContinuousBatchProcessorMetrics, attach_tracer, traced
 from .cache import PagedAttentionCache
@@ -241,7 +242,10 @@ class ContinuousBatchProcessor:
         self.reset_static_tensors(full_reset=True)
 
     def return_attention_mask(self) -> bool:
-        return self.config._attn_implementation != "paged_attention"  # we set `is_causal` to True in paged call
+        return self.config._attn_implementation in [
+            "paged|eager",
+            "paged|sdpa",
+        ]  # we set `is_causal` to True in paged call
 
     @traced
     @torch.no_grad()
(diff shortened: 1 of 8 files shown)


src/transformers/modeling_utils.py
```diff
@@ -5904,10 +5902,10 @@ class AttentionInterface(GeneralInterface):
         "flash_attention_3": flash_attention_forward,
         "flash_attention_2": flash_attention_forward,
         "flex_attention": flex_attention_forward,
-        "paged_attention": paged_attention_forward,
         "sdpa": sdpa_attention_forward,
-        "sdpa_paged": sdpa_attention_paged_forward,
-        "eager_paged": eager_paged_attention_forward,
+        "paged|flash_attention2": paged_attention_forward,
+        "paged|sdpa": sdpa_attention_paged_forward,
+        "paged|eager": eager_paged_attention_forward,
```
remi-or, line 5908: should `paged|flex_attention` be an option as well? I see it listed below in the tests

tests/generation/test_continuous_batching.py
```diff
@@ -328,6 +328,15 @@ def test_continuous_batching_parity_gpt_oss_flash(self) -> None:
             "openai/gpt-oss-20b", "paged_attention|kernels-community/flash-attn", expected_outputs
         )
 
+    def test_attn_implementation(self) -> None:
+        model = AutoModelForCausalLM.from_pretrained("gpt2")
+        manager = model.init_continuous_batching()
+        assert "paged|sdpa" in manager.model.config._attn_implementation
```
remi-or: Why not  `==` here? Same question for below

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -604,6 +608,9 @@ def __init__(
             streaming: Whether to stream tokens as they are generated
         """
         self.model = model.eval()
+        attn_implementation = "paged|" + self.model.config._attn_implementation
+        load_and_register_kernel(attn_implementation)
```
remi-or: Not familiar with this, so just resolve if this will work with non-kernelized attn like `paged|sdpa` or `paged|eager`

tests/generation/test_paged_attention.py
```diff
@@ -44,10 +44,10 @@ def setUpClass(cls):
 
     @parameterized.expand(
         [
-            ("eager_paged", 64, 128, 64),
-            ("sdpa_paged", 32, 256, 128),
-            ("paged_attention", 16, 512, 256),
-            ("flex_paged", 64, 128, 64),
+            ("paged|eager", 64, 128, 64),
+            ("paged|sdpa", 32, 256, 128),
+            ("paged|flash_attention2", 16, 512, 256),
```
remi-or: usualy it's `flash_attention_2` in `transformers` i think

tests/generation/test_paged_attention.py
```diff
@@ -89,10 +89,10 @@ def test_generate_batch_consistency(self, attn_impl, num_blocks, block_size, max
...
```
remi-or: same

remi-or (approved):
LGTM! Not sure we can run slow tests for this, but it's worth making sure the example still works with `paged|eager`, `paged|sdpa` or with `paged|flash_attention_2` -- kernels is not available on AMD so we need the classic FA package.

src/transformers/modeling_utils.py
ArthurZucker: Not supported yet AFAIK

src/transformers/generation/continuous_batching/continuous_api.py
ArthurZucker: It won't cuz not expected, but yeah I can fix this

tests/generation/test_paged_attention.py
ArthurZucker: arf

tests/generation/test_paged_attention.py
ArthurZucker: okay checking

src/transformers/modeling_utils.py
remi-or: Ok good to know :+1: ty

PR #41372, opened 2025-10-06
Enable modular files from other libraries

regisss:
# What does this PR do?

This PR extends the modular model converter so that one can use a modular file in an external library. The generated files (modeling, configuration, etc) will located in the same folder as the modular file. This requires users to use the new `--source-library` flag.

Another PR will come so that one can use model classes that are outside of the Transformers library in their modular file.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


Rocketknight1:
cc @arthurzucker @Cyrilvallez!

utils/modular_model_converter.py
```diff
@@ -505,6 +506,12 @@ def augmented_dependencies_for_class_node(
     "feature_extraction",
 )
 
+# Files from external libraries that should not be tracked
+# E.g. for habana, we don't want to track the dependencies from `modeling_all_models.py` as it is not part of the transformers library
+EXCLUDED_EXTERNAL_FILES = {
+    "habana": [{"name": "modeling_all_models", "type": "modeling"}],
+}
+
```
ArthurZucker: Thinking about having a `modular_config.json` which we can read from !

utils/modular_model_converter.py
```diff
@@ -1693,15 +1893,51 @@ def convert_modular_file(modular_file: str) -> dict[str, str]:
...
+        if source_library != "transformers":
+
+            class AbsoluteImportTransformer(cst.CSTTransformer):
+                def leave_ImportFrom(
+                    self, original_node: cst.ImportFrom, updated_node: cst.ImportFrom
+                ) -> cst.ImportFrom:
+                    return convert_relative_import_to_absolute(
+                        import_node=updated_node, file_path=relative_path, package_name=source_library
+                    )
+
```
ArthurZucker: would be better to go with a global define Transformer please

utils/modular_model_converter.py
```diff
@@ -1693,15 +1893,51 @@ def convert_modular_file(modular_file: str) -> dict[str, str]:
...
-                ).group(1)
+                if source_library != "transformers":
+                    # Convert back all absolute imports from the source library to relative ones
+                    class RelativeImportTransformer(cst.CSTTransformer):
+                        def leave_ImportFrom(self, original_node, updated_node):
+                            return convert_to_relative_import(
+                                updated_node,
+                                relative_path,
+                                source_library,
+                            )
```
ArthurZucker: same here

ArthurZucker (approved):
Looks very nice 🫡

utils/modular_model_converter.py
Cyrilvallez: Let's move it to integration file!

utils/modular_model_converter.py
```diff
@@ -1666,13 +1705,174 @@ def create_modules(modular_mapper: ModularFileMapper) -> dict[str, cst.Module]:
...
+    if stripped_name == "":
+        new_module = None
+    else:
+        name_parts = stripped_name.split(".")[i:]
+        new_module = cst.Name(name_parts[0])
+        for part in name_parts[1:]:
+            new_module = cst.Attribute(value=new_module, attr=cst.Name(part))
+
+    return import_node.with_changes(module=new_module, relative=relative)
+
```
Cyrilvallez: Let's move to integration as well!

utils/modular_model_converter.py
```diff
@@ -1693,15 +1893,51 @@ def convert_modular_file(modular_file: str) -> dict[str, str]:
...
+        # Convert all source library relative imports to absolute ones
+        if source_library != "transformers":
+
+            class AbsoluteImportTransformer(cst.CSTTransformer):
+                def leave_ImportFrom(
+                    self, original_node: cst.ImportFrom, updated_node: cst.ImportFrom
+                ) -> cst.ImportFrom:
+                    return convert_relative_import_to_absolute(
+                        import_node=updated_node, file_path=relative_path, package_name=source_library
+                    )
```
Cyrilvallez: Same, would be better in the integration!

utils/modular_model_converter.py
Cyrilvallez: And this one as well!

utils/modular_model_converter.py
```diff
@@ -1209,10 +1218,13 @@ def visit_ImportFrom(self, node: cst.ImportFrom) -> None:
...
-                    rf"(?:transformers\.models\.)|(?:\.\.\.models\.)|(?:\.\.)\w+\.({self.match_patterns}).*",
-                    import_statement,
-                )
+                if any(external_file["name"] in import_statement for external_file in self.excluded_external_files):
+                    _import = None
+                else:
+                    _import = re.search(
+                        rf"(?:transformers\.models\.)|(?:\.\.\.models\.)|(?:\.\.)\w+\.({self.match_patterns}).*",
+                        import_statement,
+                    )
```
Cyrilvallez: Let's simply skip the end of the loop here directly
```suggestion
                # If we match here, it's an import from 3rd party lib that we need to skip
                if any(external_file["name"] in import_statement for external_file in self.excluded_external_files):
                    continue
                _import = re.search(
                    rf"(?:transformers\.models\.)|(?:\.\.\.models\.)|(?:\.\.)\w+\.({self.match_patterns}).*",
                    import_statement,
                )
```

utils/modular_model_converter.py
```diff
@@ -1666,13 +1705,174 @@ def create_modules(modular_mapper: ModularFileMapper) -> dict[str, cst.Module]:
...
         new_body = [k[1]["node"] for k in sorted(body.items(), key=lambda x: x[1]["insert_idx"])]
         needed_imports = get_needed_imports(body, all_imports)
+
+        if package_name != "transformers":
+            # Convert all transformers relative imports to absolute ones
+            for imp in needed_imports:
+                if m.matches(imp, m.SimpleStatementLine(body=[m.ImportFrom()])):
+                    imp.body[0] = convert_relative_import_to_absolute(
+                        import_node=imp.body[0], file_path=file_path, package_name="transformers"
+                    )
```
Cyrilvallez, line 1731: We should not need that since we already convert all of them at the beginning with the new AbsoluteImportTransformer no?

Cyrilvallez (commented):
Very nice! Overall, I think we just mostly want to separate utils for clarity and keep what we can outside the base modular! Let's create a new file such as `modular_integrations` or `modular_external_libraries` and import the objects here!

utils/modular_model_converter.py
ArthurZucker: Yep fine by me

utils/modular_model_converter.py
regisss: New (relative) imports from `transformers` are added in the process, e.g. adding the following in the modular file
```py
GaudiLlamaDecoderLayer(LlamaDecoderLayer):
    ...
```
will lead to the following new import in the modeling file:
```py
from ...modeling_layers import GradientCheckpointingLayer
```
since `LlamaDecoderLayer` inherits from `GradientCheckpointingLayer`.

utils/modular_model_converter.py
regisss: Will keep it like that for now and will add a `modular_config.json` when `EXCLUDED_EXTERNAL_FILES ` is more used, WDYT?

Cyrilvallez: Sounds good!

utils/modular_model_converter.py
Cyrilvallez: Ha indeed, forgot about this one sorry!

Cyrilvallez (approved):
Very nice, thanks a lot for this!!

PR #41375, opened 2025-10-06
`report_to` default changed to "none" + cleaning deprecated env var

SunMarc:
# What does this PR do?

This PR changes the default for reporting to "none" as we move to v5. Moreover, we clean some deprecations messages around some env env + we deprecate `logging_dir` that was solely used for tensorboard. Instead the user needs to pass a global variable.

### DIFF
--- a/docs/source/en/tasks/knowledge_distillation_for_image_classification.md
+++ b/docs/source/en/tasks/knowledge_distillation_for_image_classification.md
@@ -108,7 +108,6 @@ training_args = TrainingArguments(
     output_dir="my-awesome-model",
     num_train_epochs=30,
     fp16=True,
-    logging_dir=f"{repo_name}/logs",
     logging_strategy="epoch",
     eval_strategy="epoch",
     save_strategy="epoch",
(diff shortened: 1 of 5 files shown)


SunMarc:
WDYT about the default change @qgallouedec ? I didn't know about this potential change unless I actually started cleaning some deprecation msg.

qgallouedec (approved):
Yes, I agree with this change. It is common to have several trackers installed (in my case, WandB and Trackio), and I don't think it's common to want to use multiple trackers simultaneously. Also, having `"none"` as the default seems to make more sense to me.

Cyrilvallez (approved):
Happy to go with it if everyone agrees!

BenjaminBossan:
@SunMarc I'm concerned about this change because countless training scripts that have `logging_dir` as an argument would start failing now without any deprecation period. Would it not be better to keep the argument, give a `FutureWarning`, and then ignore it?

tomaarsen:
I just discovered this after my `codecarbon` stopped being used automatically in Sentence Transformers. I'm not the biggest fan of this change. Previously I could publish training scripts that worked regardless of what callbacks users had configured, and now I have to make the choice of callbacks for the user. 

I'm even a bit tempted to override the default back to `"all"` for Sentence Transformers, but I also really don't like to change the defaults of `transformers` arguments except for backwards compatibility, as it can be unexpected for users familiar with `transformers`.

- Tom Aarsen

PR #41376, opened 2025-10-06
deprecate `jit_mode_eval`

SunMarc:
# What does this PR do?

This PR deprecates `jit_mode_eval` arguments. Going on, it is better to fully rely on torch compile.


SunMarc:
cc @yao-matrix for visibility

src/transformers/training_args.py
```diff
@@ -1004,7 +1002,10 @@ class TrainingArguments:
     seed: int = field(default=42, metadata={"help": "Random seed that will be set at the beginning of training."})
     data_seed: Optional[int] = field(default=None, metadata={"help": "Random seed to be used with data samplers."})
     jit_mode_eval: bool = field(
-        default=False, metadata={"help": "Whether or not to use PyTorch jit trace for inference"}
+        default=False,
+        metadata={
+            "help": "`jit mode eval is deprecated and will be removed in v5.2. Please use torch_compile instead. "
+        },
     )
```
Cyrilvallez: If we drop the support right now, we might as well remove the arg accordingly as well IMO - then we present the v5 with a list of all training args that were removed

Cyrilvallez (approved):
Nice, but we can simply remove the arg IMO!

src/transformers/training_args.py
SunMarc: let's do that !

PR #41377, opened 2025-10-06
[testing] Fix `JetMoeIntegrationTest`

ydshieh:
# What does this PR do?

> tests/models/jetmoe/test_modeling_jetmoe.py::JetMoeIntegrationTest::test_model_8b_generation 

cause the pytest process being killed (and sometimes hangs for hours), see the log below.

It loads the model on  `cpu`. Let's use "auto".

The outputs changes after #41324 (which is a fix of #40132), so the tests still fail but not causing the pytest process being killed and job hanging. 

There is a new warning showing 

>  The attention mask is not set and cannot be inferred from input because pad token is same as eos token. As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.

cc @ArthurZucker for this output change issue. 

Log when running before this PR: 

> 2025-09-30T16:33:22.8891725Z tests/models/jetmoe/test_modeling_jetmoe.py::JetMoeIntegrationTest::test_model_8b_batched_generation 
2025-09-30T16:33:22.8892381Z -------------------------------- live log call ---------------------------------
2025-09-30T16:33:22.8893314Z WARNING  transformers.generation.configuration_utils:logging.py:328 The following generation flags are not valid and may be ignored: ['temperature']. Set `TRANSFORMERS_VERBOSITY=info` for more details.
2025-09-30T16:33:23.3206200Z FAILED                                                                   [ 99%]
2025-09-30T16:39:21.4805460Z Killed
2025-09-30T16:39:21.4847256Z tests/models/jetmoe/test_modeling_jetmoe.py::JetMoeIntegrationTest::test_model_8b_generation


vasqu:
Something is fishy, the outputs shouldnt change tbh checking

tests/models/jetmoe/test_modeling_jetmoe.py
```diff
@@ -127,10 +126,16 @@ def test_flash_attn_2_inference_equivalence_right_padding(self):
...
+        cleanup(torch_device, gc_collect=True)
+
+    def tearDown(self):
+        cleanup(torch_device, gc_collect=True)
+
     @slow
     def test_model_8b_logits(self):
         input_ids = [1, 306, 4658, 278, 6593, 310, 2834, 338]
-        model = JetMoeForCausalLM.from_pretrained("jetmoe/jetmoe-8b")
+        model = JetMoeForCausalLM.from_pretrained("jetmoe/jetmoe-8b", device_map="auto", torch_dtype=torch.bfloat16)
```
vasqu, line 138: My only nit, don't we want to use fp16 for more consistency or was the model saved in bf16?

vasqu (approved):
Thx lgtm! Just a small nit. Outputs have changed but I submitted a PR for that in #41423

tests/models/jetmoe/test_modeling_jetmoe.py
ydshieh: I would love to use fp16, but there is memory issue when using "auto", see this internal discussion

https://huggingface.slack.com/archives/C01NE71C4F7/p1758532074369679

and it seems nowadays, many model weights are saved with bf16 (which is normal as we want to train with it)

vasqu: Argh, gotcha. Fair enough but that's not a good bug :sweat: guess we are fine since bf16 dominates

PR #41378, opened 2025-10-06
🚨🚨 Remove all traces of legacy cache format

Cyrilvallez:
# What does this PR do?

As per the title. Remove all legacy format and deprecated classes/methods.


gante:
@Cyrilvallez haha we had the same idea, started working on exactly the same things :D I am going to
- keep #41406 (because of the PEFT dependency issue)
- close #41405 [rm cache as tuple support], as this PR is more complete

gante:
actually #41405 has a few extra generate-side changes, but I'll wait for this one to be merged and open a new PR afterwards to prevert needless merge conflicts 🤗

Cyrilvallez:
Alright, thanks a lot! Indeed did not update `generate` as much as could be done in this one, wanted to ask you afterwards (should have done before haha to avoid duplicated work 🥲) - will revert cache classes in this one to avoid friction with peft

src/transformers/models/cpmant/modeling_cpmant.py
```diff
@@ -796,12 +789,5 @@ def get_input_embeddings(self):
     def set_input_embeddings(self, embeddings):
         self.cpmant.input_embedding = embeddings
 
-    def _reorder_cache(self, past_key_values, beam_idx):
```
gante, line 799: (good catch!)

src/transformers/models/whisper/generation_whisper.py
```diff
@@ -1178,26 +1178,14 @@ def split_by_batch_index(values, key, batch_idx, is_shortform, beam_indices=None
...
-                    for v in range(len(values)):
-                        layer_past_key_values = []
-                        for w in values[v]:
-                            if len(w) != 0:
-                                layer_past_key_values.append(w[batch_idx][None].cpu())
-                            else:
-                                layer_past_key_values.append(w)
-                        all_past_key_values.append(tuple(layer_past_key_values))
-                    return tuple(all_past_key_values)
+                all_past_key_values = []
```
gante, line 1181: (I'm assuming you double-checked that slow whisper tests have no regressions 👍 )

tests/generation/test_utils.py
```diff
@@ -2576,6 +2529,24 @@ def _check_sequence_inside_sequence(self, tensor_1, tensor_2):
 
         self.assertTrue(flag)
 
+    def _check_caches_are_equal(self, cache1: Cache, cache2: Cache):
```
gante, line 2532: 💛 💛 💛

gante (approved):
Thank you for this thorough cleanup 🫶

src/transformers/models/whisper/generation_whisper.py
Cyrilvallez: Yes, all good!

PR #41382, opened 2025-10-06
Remove `local_rank` arg from `TrainingArguments`

SunMarc:
# What does this PR do?

This PR removes `local_rank` from `TrainingArguments`. This arg can be safely replaced by `local_process_index`. This arg was never meant to be set and was supposed to be there to be overwritten by torch distributed.


Cyrilvallez (approved):
Yes, again happy to remove those old legacy arguments!

SunMarc:
Reverted in the end here : https://github.com/huggingface/transformers/pull/41504

PR #41383, opened 2025-10-06
remove `tpu_num_cores`

SunMarc:
# What does this PR do?

This PR removes `tpu_num_cores` arguments in `TrainingArguments`. This arg is somewhat confusing because the value is either `1` or `None` which means that we are using all devices. Instead, we remove the arg to make it clearer. Users who wants to control the number of devices can use `TPU_NUM_DEVICES` as specified in xla docs. Note that this was only used in example scripts


src/transformers/training_args.py
```diff
@@ -1056,7 +1054,10 @@ class TrainingArguments:
         },
     )
     tpu_num_cores: Optional[int] = field(
-        default=None, metadata={"help": "TPU: Number of TPU cores (automatically passed by launcher script)"}
+        default=None,
+        metadata={
+            "help": "This argument is deprecated and will be removed in v5.2. Please use `TPU_NUM_DEVICES` instead. By default, we will use all available devices. "
+        },
```
Cyrilvallez: We don't remove it right now? Why is that?

examples/legacy/seq2seq/finetune_tpu.sh
```diff
@@ -16,8 +16,7 @@ export TPU_NUM_CORES=8
 
 # the proper usage is documented in the README, you need to specify data_dir, output_dir and model_name_or_path
 # run ./finetune_tpu.sh --help to see all the possible options
-python xla_spawn.py --num_cores $TPU_NUM_CORES \
-    finetune_trainer.py \
+python xla_spawn.py finetune_trainer.py \
```
MekkCyber, line 20:
```suggestion
# To specify the number of cores to use, use the TPU_NUM_DEVICES environment variable
python xla_spawn.py finetune_trainer.py \
```

MekkCyber (approved):
Looks great ! Thanks for cleaning. For the sake of completeness I added a small comment

PR #41384, opened 2025-10-06
Remove `past_index`

SunMarc:
# What does this PR do?

This PR removes `past_index`. This was supposed to allow training a few old + specific arch models like transformersxl but I don't think anyone actually did it as there is no documentation at all. I think it is fine to remove it without any deprecation msg but happy to add it if needed. Maybe some users just copy paste Trainer code and try to maintain that.


qgallouedec (approved):
lgtm

Cyrilvallez (approved):
I think it's fine to remove without warning for v5. We will however need a nice list of all removed args along with the reason and how to replace so that users can refer to it when we release!

PR #41385, opened 2025-10-06
[`Ernie 4.5 Moe`] Fix Moe and offloading

vasqu:
The model is broken
- Moe was initialized with wrong dims (loading the models already failed)
- Gate needs the autocast decorator as well (we need to guarantee fp32)
- Not sure why but using the forward with the moe statics works with offloading while retrieving the parameter etc does not
- The router seems like an artefact


src/transformers/models/ernie4_5_moe/modeling_ernie4_5_moe.py
```diff
@@ -350,38 +325,33 @@ def __init__(self, config):
         if config.moe_num_shared_experts > 0:
             self.shared_experts = Ernie4_5_MoeMLP(config, config.moe_intermediate_size * config.moe_num_shared_experts)
 
-    def route_tokens_to_experts(self, hidden_states, router_logits):
+    def route_tokens_to_experts(self, hidden_states):
```
vasqu, line 327: Unsure if this is ok for the vLLM side as I change the signature here

src/transformers/models/ernie4_5_moe/modeling_ernie4_5_moe.py
```diff
@@ -287,37 +287,12 @@ def forward(self, hidden_states):
...
-        return router_logits, selected_experts, routing_weights
-
-
 class Ernie4_5_MoeExperts(nn.ModuleList):
     def __init__(self, config):
         super().__init__()
         self.num_experts = config.moe_num_experts
         for _ in range(self.num_experts):
-            self.append(Ernie4_5_MoeMLP(config))
+            self.append(Ernie4_5_MoeMLP(config, config.moe_intermediate_size))
```
vasqu, line 294: This is needed otherwise we cant load any of the pretrained models

src/transformers/models/ernie4_5_moe/modeling_ernie4_5_moe.py
```diff
@@ -350,38 +325,33 @@ def __init__(self, config):
...
             if isinstance(hidden_states.device.type, str) and hidden_states.device.type != "mps"
             else "cpu"
         )
 
         with torch.autocast(device_type=device_type, enabled=False):  # Force float32
+            router_logits = self.gate(hidden_states.float())
             routing_weights = F.softmax(router_logits, dim=1, dtype=torch.float)
-            routing_bias = self.moe_statics.e_score_correction_bias.squeeze()
-            _, selected_experts = torch.topk(routing_weights + routing_bias, self.top_k, dim=-1)
+            _, selected_experts = torch.topk(self.moe_statics(routing_weights), self.top_k, dim=-1)
```
vasqu, line 337: Idk why this fixes offloading tbh :sweat_smile:

vasqu:
run-slow: ernie4_5_moe

src/transformers/models/ernie4_5_moe/modeling_ernie4_5_moe.py
ArthurZucker: it should be fine, they rely on our self.expert tho

ArthurZucker (approved):
LGTM

PR #41395, opened 2025-10-07
Enable  FURB rules in ruff

cyyever:
# What does this PR do?

ruff contains the below `FURB` rules:
```
FURB101	open and read should be replaced by Path({filename}).{suggestion}
FURB103	open and write should be replaced by Path({filename}).{suggestion}
FURB105	Unnecessary empty string passed to print
FURB110	Replace ternary if expression with or operator
FURB113	Use {suggestion} instead of repeatedly calling {name}.append()
FURB116	Replace {function_name} call with {display}
FURB118	Use operator.{operator} instead of defining a {target}
FURB122	Use of {}.write in a for loop
FURB129	Instead of calling readlines(), iterate over file object directly
FURB131	Prefer clear over deleting a full slice
FURB132	Use {suggestion} instead of check and remove
FURB136	Replace if expression with {min_max} call
FURB140	Use itertools.starmap instead of the generator
FURB142	Use of set.{}() in a for loop
FURB145	Prefer copy method over slicing
FURB148	enumerate value is unused, use for x in range(len(y)) instead
FURB152	Replace {literal} with math.{constant}
FURB154	Use of repeated consecutive {}
FURB156	Use of hardcoded string charset
FURB157	Verbose expression in Decimal constructor
FURB161	Use of bin({existing}).count('1')
FURB162	Unnecessary timezone replacement with zero offset
FURB163	Prefer math.{log_function}({arg}) over math.log with a redundant base
FURB164	Verbose method {method_name} in {constructor} construction
FURB166	Use of int with explicit base={base} after removing prefix
FURB167	Use of regular expression alias re.{}
FURB168	Prefer is operator over isinstance to check if an object is None
FURB169	When checking against None, use {} instead of comparison with type(None)
FURB171	Membership test against single-item container
FURB177	Prefer Path.cwd() over Path().resolve() for current-directory lookups
FURB180	Use of metaclass=abc.ABCMeta to define abstract base class
FURB181	Use of hashlib's .digest().hex()
FURB187	Use of assignment of reversed on list {name}
FURB188	Prefer str.removeprefix() over conditionally replacing with slice.
FURB189	Subclassing {subclass} can be error prone, use collections.{replacement} instead
FURB192	Prefer min over sorted() to compute the minimum value in a sequence
```

In my opinion these rules are useful and many of their fixed have been applied ion previous PRs.


examples/legacy/token-classification/scripts/preprocess.py
```diff
@@ -31,7 +31,6 @@
             continue
 
         if (subword_len_counter + current_subwords_len) > max_len:
-            print("")
```
Rocketknight1, line 34: This seems functional - it prints a newline!

Rocketknight1 (approved):
Most of this looks good, with one comment! cc @cyrilvallez @arthurzucker since it enables more Ruff rules

examples/legacy/token-classification/scripts/preprocess.py
cyyever: Yes, let's keep it.

ydshieh:
Thank you.

Do you know if we need to be more careful to close the file descriptor in statements like

> [len(x) for x in Path(data_file).open()]

?

This is not the topic of this PR, but if it's better to close them, let's do it in a follow PR.

cyyever:
@ydshieh We can check all cases in code using
```
ruff check --select SIM115 src | more
SIM115 Use a context manager for opening files
   --> src/transformers/commands/serving.py:945:40
    |
943 |                                 image = Image.open(BytesIO(base64.b64decode(image_data)))
944 |
945 |                                 file = tempfile.NamedTemporaryFile(suffix=".png", delete=False)
    |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
946 |                                 url = file.name
    |

SIM115 Use a context manager for opening files
  --> src/transformers/integrations/tensor_parallel.py:88:22
   |
86 |         import sys
87 |
88 |         sys.stdout = open(os.devnull, "w")
   |                      ^^^^
89 |         sys.stderr = open(os.devnull, "w")
   |

SIM115 Use a context manager for opening files
  --> src/transformers/integrations/tensor_parallel.py:89:22
   |
88 |         sys.stdout = open(os.devnull, "w")
89 |         sys.stderr = open(os.devnull, "w")
   |                      ^^^^
90 |
91 |     device_map = tp_device
   |
```
I will propose another PR to fix them...

ydshieh:
@cyyever Let me know if we can merge this one now 🙏

cyyever:
@ydshieh CI is green now.

ydshieh (approved):
🤗

PR #41400, opened 2025-10-07
Automatically infer attention_mask when pad_token_id == eos_token_id

tanuj-rai:
# What does this PR do?

This PR updates the _prepare_special_tokens in the generation utilities to automatically infer a correct attention mask when none is provided and pad_token_id == eos_token_id.

Fixes ##41392

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker @vasqu

### DIFF
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -2084,16 +2084,34 @@ class GenerationMixin(ContinuousMixin):
             raise ValueError(
                 "`decoder_start_token_id` or `bos_token_id` has to be defined for encoder-decoder generation."
             )
-        if (
-            eos_token_tensor is not None
-            and isin_mps_friendly(elements=eos_token_tensor, test_elements=pad_token_tensor).any()
-        ):
-            if kwargs_has_attention_mask is not None and not kwargs_has_attention_mask:
-                logger.warning_once(
-                    "The attention mask is not set and cannot be inferred from input because pad token is same as "
-                    "eos token. As a consequence, you may observe unexpected behavior. Please pass your input's "
-                    "`attention_mask` to obtain reliable results."
-                )
+		if (
+		    eos_token_tensor is not None
+		    and isin_mps_friendly(elements=eos_token_tensor, test_elements=pad_token_tensor).any()
+		):
+		    if kwargs_has_attention_mask is not None and not kwargs_has_attention_mask:
+		        # Instead of warning, automatically create an attention mask
+		        logger.info_once(
+		            "Automatically inferring attention mask: treating the first EOS token as end-of-sequence "
+		            "and subsequent EOS tokens as padding (since pad_token == eos_token)."
+		        )
+		
+		        # Infer the attention mask from the input_ids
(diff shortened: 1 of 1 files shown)


yonikremer:
Really happy to see a PR request fast!, the code looks good, def better than what I did locally.

tanuj-rai:
> Really happy to see a PR request fast!, the code looks good, def better than what I did locally.

Thanks @yonikremer!

vasqu (commented):
Thanks for the PR. I have some thoughts below but cc @gante for generation.

Tbh, I'm not really sure if we want this as tokenizers behave differently
- Not always pad == eos
- Not always are special tokens used (most of the time they should)
- Padding side will influence the logic here, i.e. left padding behaves differently than right padding

This makes it imo hard to make this generalization that we can infer the mask ourselves. The tokenizer should always return the attention mask instead - when do you only pass the input ids or more like how do you preprocess the text?

vasqu:
Sorry, I think I've passed judgement too fast. I see that it only goes into the specific branch of pad == eos so that handles a bit of my issues I had. 

There still is the question about: Are special tokens used (bos/eos)? And do we adjust to the padding side?

gante (changes requested):
(apologies in advance for the assertive tone, I want to make sure the message gets passed across)

Not approving: when PAD = EOS, we don't know the intent behind the token. Maybe it is indeed an intentional sequence of EOS tokens that is encoding some special model-specific property.

This is actually a very old question that has been popping up in the library for the past few years (even before I joined HF). The consensus is that we want to avoid using `_prepare_special_tokens`, which is model-agnostic, and instead let the model-specific tokenizer prepare the correct mask.

As such, instead of adding 30 LOC with error-prone code, this warning can be easily skipped by calling generate with the attention mask, i.e.

```py
model_inputs = tokenizer(...)
model.generate(**model_inputs)
```

👉 always pass the attention mask when possible, it will prevent complex bugs 🤗

tanuj-rai:
> (apologies in advance for the assertive tone, I want to make sure the message gets passed across)
> 
> Not approving: when PAD = EOS, we don't know the intent behind the token. Maybe it is indeed an intentional sequence of EOS tokens that is encoding some special model-specific property.
> 
> This is actually a very old question that has been popping up in the library for the past few years (even before I joined HF). The consensus is that we want to avoid using `_prepare_special_tokens`, which is model-agnostic, and instead let the model-specific tokenizer prepare the correct mask.
> 
> As such, instead of adding 30 LOC with error-prone code, this warning can be easily skipped by calling generate with the attention mask, i.e.
> 
> ```python
> model_inputs = tokenizer(...)
> model.generate(**model_inputs)
> ```
> 
> 👉 always pass the attention mask when possible, it will prevent complex bugs 🤗

Thank you @gante for the detailed explanation — that makes sense.
I understand now why this behavior shouldn’t be handled globally in _prepare_special_tokens.

vasqu:
Closing as discussed here and in the linked issue. Thanks for understanding 🤗

PR #41402, opened 2025-10-07
[v5] rm `utils/tf_ops/`

gante:
# What does this PR do?

TF-related folder that flew under the radar :) (seems unused in the lib, latest commit was 4 years ago)


Cyrilvallez (approved):
Oh nice catch!

PR #41403, opened 2025-10-07
[Trainer] deprecate ray scope

SunMarc:
# What does this PR do?

This PR removes `ray_scope` from `TrainingArguments`. Instead, users will have to set the env var RAY_SCOPE if they want to change its value.


Cyrilvallez:
What's the rationale to switch to env var instead? Is it to align with other practices?

SunMarc:
> What's the rationale to switch to env var instead? Is it to align with other practices?

Yeah and I feel like this arg is not that useful tbh + related to third party lib so I prefer to remove it from training args.

Cyrilvallez (approved):
Alright, fine by me then! Once again, I think it's best to fully remove, and add it to the list of removed v5 args (along with the new env to set instead!)

PR #41405, opened 2025-10-07
🚨 [v5] remove deprecated cache tuple input support

gante:
# What does this PR do?

This PR:
- Removes the last remnants (🤞) of tuple cache input support, scheduled to be removed in v4.58
- Removes other minor items scheduled to be removed in v4.58, caught in the search
- Updates a few incorrect `past_key_values` type hints
- ⚠️ EDIT: add cache initialization in `forward` on models that were missing it


gante:
CI is failing for models that don't contain cache initialization in their forward pass, working on it

src/transformers/generation/utils.py
```diff
@@ -2006,17 +1996,13 @@ def _prepare_cache_for_generation(
...
-        # initialization should happen inside the model at prefill time.
-        else:
-            model_kwargs[cache_name] = DynamicCache(**dynamic_cache_kwargs)
-
         # TODO (joao): this logic is incomplete, e.g. `offloaded` should apply to both caches. Refactor this function
         # to correctly pass parameterization to both caches.
-        if requires_cross_attention_cache and not isinstance(model_kwargs[cache_name], EncoderDecoderCache):
+        if (
+            requires_cross_attention_cache
+            and cache_name in model_kwargs
```
zucchini-nlp: ig this is for special models where name isn't "past_key_values". Is it not going to fail later in model, if we leave the non-EncoderDecoderCache cache?

src/transformers/models/bert_generation/modeling_bert_generation.py
```diff
@@ -564,16 +564,6 @@ def forward(
         else:
             use_cache = False
 
-        return_legacy_cache = False
```
zucchini-nlp, line 567: imo we should prepare cache from scratch if it is None, similar to other models. Same for other bert-like models

src/transformers/models/llama4/configuration_llama4.py
```diff
@@ -126,18 +125,6 @@ def __init__(
 
         self._vision_feature_layer = kwargs.get("vision_feature_layer", -1)
 
-        @property
```
zucchini-nlp, line 129: lets delete the line 126 as well, it is not used in modeling

zucchini-nlp (approved):
Great

src/transformers/models/bert/modeling_bert.py
```diff
@@ -691,7 +691,7 @@ def forward(
         inputs_embeds: Optional[torch.Tensor] = None,
         encoder_hidden_states: Optional[torch.Tensor] = None,
         encoder_attention_mask: Optional[torch.Tensor] = None,
-        past_key_values: Optional[Union[list[torch.FloatTensor], Cache]] = None,
+        past_key_values: Optional[Cache] = None,
```
vasqu, line 694: Finally :pray:

src/transformers/generation/utils.py
gante: This branch, which converts an decoder-only cache into an encoder-decoder cache, is meant to kick in in the following case:
- the model is encoder-decoder
- the user has specified that they want a specific cache implementation

I'll replace `cache_name` by `"past_key_values"` to make it clearer that only models with the default cache name will use this branch

src/transformers/models/bert_generation/modeling_bert_generation.py
gante: yes, that was indeed missing 👍 adding in this PR

src/transformers/models/blip/modeling_blip_text.py
```diff
@@ -396,16 +396,9 @@ def forward(
...
-                    "Passing a tuple of `past_key_values` is deprecated and will be removed in Transformers v4.58.0. "
-                    "You should pass an instance of `EncoderDecoderCache` instead, e.g. "
-                    "`past_key_values=EncoderDecoderCache.from_legacy_cache(past_key_values)`."
-                )
-                past_key_values = EncoderDecoderCache.from_legacy_cache(past_key_values)
             # The model acts as encoder decoder but is not an encoder decoder. So we cast all cache objects to
             # `EncoderDecoderCache` type assuming that the incoming cache is from `self_attention`
-            elif isinstance(past_key_values, DynamicCache):
+            if isinstance(past_key_values, DynamicCache):
                 past_key_values = EncoderDecoderCache(past_key_values, DynamicCache(config=self.config))
```
vasqu, line 402: Not super important but we still need this if generation checks for cross attn cache necessity and converts if necessary  (also slightly changed in this PR for other cache names ig). Might be that this needs this custom logic, not familiar with the model here.

Relevant lines in main https://github.com/huggingface/transformers/blob/0464d9eb37b7e384f8d96b7c9bb9cfb61fcdfd1e/src/transformers/generation/utils.py#L2019-L2023

src/transformers/models/bridgetower/modeling_bridgetower.py
```diff
@@ -1069,16 +1069,6 @@ def forward(
                 )
                 use_cache = False
 
-        return_legacy_cache = False
```
vasqu, line 1072: We have a lot of deletions for a lot of models similar to here. Do all of them need a cache init or is it more model dependent? (Sorry for the mess here on the dependencies in the case we need them all :cry:)

vasqu (commented):
Don't have any major comments but we should sync with #41378 (which I noticed halfway through here)

Seems like a bit of duplicated effort atm. So it's hard to say how to proceed without proper comms with @Cyrilvallez here. Otherwise LGTM (when the CI failures are fixed)

gante:
@vasqu indeed duplicated work. I'm going to sync with @Cyrilvallez

gante:
Closing in place of #41378 , which is more complete!

PR #41406, opened 2025-10-07
🚨 [v5] Remove deprecated cache classes

gante:
# What does this PR do?

🚨🚨🚨 DO NOT MERGE 🚨🚨🚨
- Depends on #41889

Merging this PR will break `from transformers import Trainer` on the latest PEFT release (`0.17.1`), which in turn breaks our `make fixup` when PEFT is installed. PEFT `main` already sorted this issue ([here](https://github.com/huggingface/peft/commit/41c07f0445879da0e73ca3c682a0c359cd027f50)), but it is yet unreleased as of 7-oct-2025.

cc @BenjaminBossan (chatted on slack -- when `transformers` has a concrete timeline for v5, we can best evaluate an adequate timing for a PEFT release)
cc @Cyrilvallez FYI 

_______________________________

This PR removes the deprecated cache classes in `cache_utils.py`.

### DIFF
--- a/docs/source/en/internal/generation_utils.md
+++ b/docs/source/en/internal/generation_utils.md
@@ -253,21 +253,7 @@ A [`StoppingCriteria`] can be used to change when to stop generation (other than
 
 [[autodoc]] QuantizedCache
 
-[[autodoc]] QuantoQuantizedCache
-
-[[autodoc]] HQQQuantizedCache
-
-[[autodoc]] OffloadedCache
-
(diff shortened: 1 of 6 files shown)


Cyrilvallez:
Oh damn, I had https://github.com/huggingface/transformers/pull/41378 as well doing the same, so I will revert the classes and just remove support for legacy in order to wait for peft! Thanks for letting me know!

PR #41408, opened 2025-10-07
Benchmark overhaul

remi-or:
This PR overhauls the benchmarking suite that is included in transformers. 
The benchmarking suite is now based around three main components:

- `BenchmarkingConfig` is a dataclass-like object which contains everything needed to reproduce a benchmark on the same machine: input length, generation length, whether to use `kernels` or `compile`, attention implementation, etc. (subject to name change)
- `BenchmarkRunner` is the class that runs the benchmarks defined by the configs, with a given number of measurement iterations, warmup iterations, and a model-id. The runner takes care of setting up the runs in a way that ensures no run interacts with the downstream ones: the model is reloaded, the cache is emptied and the GPU memory is flushed. It also saves the results, the config, and any additional metadata needed to reproduce the benchmark, like hardware information and package versions. 
- The created results files, which contain enough informations to induces (to my knowledge) most of the metrics used to evaluate a model: e2e_atency, tpot, ttft, even inter-token latency. Results also include a sample of what has been generated, which is useful to check if it was gibberish. The results files are in json format and are made to be easily created from the dataclass-like objects and vice versa. 

For now, the new benchmarking suite replaces the `benchmark_v2` part of `transformers` but it could also overwrite the `benchmark` (v1) part. It would be good to make that decision in this PR. And update the CI workflows that rely on the current `benchmark_v2` (putting the PR in draft mode until then).
An example of how to use the new benchmarking suite can be found in `run_benchmarks.py`.

The format of the results file can (and may be bound to) change as we develop tools to analyze them. 
If there is a metric you want to see measured in `transformers`, please leave a comment before this is merged :slightly_smiling_face:


benchmark_v2/framework/benchmark_config.py
```diff
@@ -0,0 +1,211 @@
...
+        # Flash attention does not support compile mode, so we turn it off # FIXME: it would be better to support it
+        is_fa = self.attn_implementation == "flash_attention_2"
+        is_fa |= self.attn_implementation == "sdpa" and self.sdpa_backend == "flash_attention"
+        if is_fa:
+            logger.warning("Flash attention does not support compile mode. Turning off compile mode.")
+            self.compile_mode = None
+
+    @property
+    def device(self) -> str:
+        return "cuda"
```
McPatate: Image

benchmark_v2/framework/benchmark_config.py
```diff
@@ -0,0 +1,211 @@
...
+
+    @property
+    def dtype(self) -> str:
+        return "torch.bfloat16"
```
McPatate: same here, is it always set to this value?

benchmark_v2/framework/benchmark_config.py
```diff
@@ -0,0 +1,211 @@
...
+
+    @property
+    def hash(self) -> str:
+        return hashlib.sha256(json.dumps(self.to_dict()).encode()).hexdigest()
+
+    def infer_name(self) -> str:
+        """Infer a human-readable name for the benchmark config."""
```
McPatate: > human-readable

hmm...

Joke aside a bit verbose for a name imo, I'd rather do something with the first letters of the hash + date

benchmark_v2/framework/benchmark_config.py
```diff
@@ -0,0 +1,211 @@
...
+            "num_tokens_to_generate": self.num_tokens_to_generate,
+            "attn_implementation": self.attn_implementation,
+            "sdpa_backend": self.sdpa_backend,
+            "compile_mode": self.compile_mode,
+            "compile_options": self.compile_options,
+            "kernelize": self.kernelize,
+        }
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> "BenchmarkConfig":
```
McPatate: typing doesn't match with `to_dict`
```py
def to_dict(self) -> dict[str, Union[None, int, float, str]]:
```

benchmark_v2/framework/benchmark_config.py
```diff
@@ -0,0 +1,211 @@
...
+        warmup_iterations=warmup_iterations,
+        measurement_iterations=measurement_iterations,
+        batch_size=batch_size,
+        sequence_length=sequence_length,
+        num_tokens_to_generate=num_tokens_to_generate,
+        gpu_monitoring=gpu_monitoring,
+    )
+
+
+def generate_useful_configs(
```
McPatate:
```suggestion
def generate_default_configs(
```

benchmark_v2/framework/benchmark_config.py
McPatate:
```suggestion
    def from_dict(cls, data: dict[str, Union[None, int, float, str]]) -> "BenchmarkConfig":
```

benchmark_v2/framework/benchmark_config.py
```diff
@@ -0,0 +1,211 @@
...
+            "attn_implementation": self.attn_implementation,
+            "sdpa_backend": self.sdpa_backend,
+            "compile_mode": self.compile_mode,
+            "compile_options": self.compile_options,
+            "kernelize": self.kernelize,
+        }
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> "BenchmarkConfig":
+        return cls(
```
McPatate, line 113: perhaps we can wrap this in a `try: ... except: ...` block to catch missing keys in the `data` dict?

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -0,0 +1,359 @@
...
+                result.accumulate(wall_time_0, e2e_latency, dt_tokens, decoded_output, gpu_metrics)
+            self.logger.info("Benchmarking done. Cleaning up.")
+
+            # Profile if needed
+            if num_tokens_to_profile > 0:
+                self.profile_generate(num_tokens_to_profile, config.name)
+
+            return {
+                "metadata": BenchmarkMetadata(model_id=model_id, commit_id=self.commit_id),
+                "measures": result,
```
McPatate:
```suggestion
                "measurements": result,
```
wdyt?

benchmark_v2/framework/benchmark_runner.py
McPatate, line 1: have we got a pretty print formatter for bench results?

benchmark_v2/run_benchmarks.py
```diff
@@ -19,477 +19,60 @@
...
+        batch_size=args.batch_size,
+        sequence_length=args.sequence_length,
+        num_tokens_to_generate=args.num_tokens_to_generate,
+    )
+    random.shuffle(benchmark_configs)
 
-if __name__ == "__main__":
-    sys.exit(main())
+    runner = BenchmarkRunner(logger, args.output_dir, args.commit_id)
+    results = runner.run_benchmarks(args.model_id, benchmark_configs, args.num_tokens_to_profile)
```
McPatate: would be down for a pretty console print enabled by default for the results

benchmark_v2/framework/hardware_metrics.py
```diff
@@ -0,0 +1,172 @@
...
+            "gpu_memory_total_mb": self.gpu_memory_total_mb,
+            "python_version": self.python_version,
+            "torch_version": self.torch_version,
+        }
+
+
+# Functions to get information about the GPU
+def get_amd_gpu_stats() -> tuple[int, float]:
+    """Returns the utilization and memory used of an AMD GPU, both in percent"""
+    rocm_smi_output = subprocess.check_output(["rocm-smi", "--json", "--showuse", "--showmeminfo", "VRAM"])
```
McPatate, line 57: ouch, I feel the pain lol

benchmark_v2/framework/hardware_metrics.py
```diff
@@ -0,0 +1,172 @@
...
+    gpu_stats = json.loads(rocm_smi_output.decode("utf-8"))
```
McPatate, line 58: glad the output is json at least

benchmark_v2/framework/hardware_metrics.py
```diff
@@ -0,0 +1,172 @@
...
+        if "amd" in self.device_name.lower():
+            self.get_utilization_and_memory_used = get_amd_gpu_stats
+        elif "nvidia" in self.device_name.lower():
+            self.get_utilization_and_memory_used = get_nvidia_gpu_stats
+        else:
+            raise RuntimeError(f"Unsupported GPU: {self.device_name}")
+
+    def get_utilization_and_memory_used(self) -> tuple[int, float]:
+        """Get the utilization and memory used of the GPU, both in percent"""
+        raise NotImplementedError("This method is meant to be monkey patched during __init__")
```
McPatate, line 89: \~meta\~

benchmark_v2/framework/hardware_metrics.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def get_nvidia_gpu_stats() -> tuple[int, float]:
+    """Returns the utilization and memory used of an NVIDIA GPU, both in percent"""
+    gpu_stats = gpustat.GPUStatCollection.new_query()
+    gpu_stats = gpu_stats[0]
+    return int(gpu_stats["utilization.gpu"]), float(gpu_stats["memory.used"]) / 1024**3
+
+
+class GPUStatsGetter:
```
McPatate:
```suggestion
class GPUStatsCollector:
```

benchmark_v2/framework/hardware_metrics.py
```diff
@@ -0,0 +1,172 @@
...
+        self.device_name, self.device_memory_total = get_device_name_and_memory_total()
+        # Monkey patch the get_utilization_and_memory_used method based on the GPU type
+        if "amd" in self.device_name.lower():
+            self.get_utilization_and_memory_used = get_amd_gpu_stats
+        elif "nvidia" in self.device_name.lower():
+            self.get_utilization_and_memory_used = get_nvidia_gpu_stats
+        else:
+            raise RuntimeError(f"Unsupported GPU: {self.device_name}")
+
+    def get_utilization_and_memory_used(self) -> tuple[int, float]:
```
McPatate:
```suggestion
    def get_measurements(self) -> tuple[int, float]:
```

benchmark_v2/run_benchmarks.py
McPatate, line 1: nice job sir, clean and straight to the point 🫡

McPatate (approved):
btw to disable the associated CI workflows while I rework them later:
```yml
on:
  workflow_dispatch:
```
and you can delete the rest of the triggers

A few naming comments overall, but minor stuff I believe, feel free to ignore, gj 👌🏻

benchmark_v2/framework/benchmark_config.py
remi-or: Made it a constant parameter instead

benchmark_v2/framework/benchmark_config.py
remi-or: Same as device for now, constant but no longer a `proprety`

benchmark_v2/framework/benchmark_config.py
remi-or: Not sure we want the hash + date in the name -- the config is meant to be re-usable for any benchmark with the same framework, so the date adds little information. The date belongs with the measures taken in the benchmark parametrixed by the config, The hash is already used for indexing. As for human-readable, I admit it's a long name but it contains the main informations and can be understood, unlike a hash: 
```
w5_i20-unmonitored-b32_s128_n128-sdpa_math-compiled_max-autotune-kernelized
```
I can expand the w/i (warmup / iterations) and bsn (batch size, seqlen, ntok to generate) if you want. Or if you have an idea in mind for another use of the config name, Im all hears!

benchmark_v2/framework/benchmark_config.py
remi-or: Good catch, changed it

benchmark_v2/framework/benchmark_config.py
remi-or: There should not be any missing keys at this stage, but I can use `.get` with default values indeed. Will do.

benchmark_v2/framework/benchmark_config.py
remi-or: already changed in the other direction because type could be a lot more complicated (dict, etc)

benchmark_v2/framework/hardware_metrics.py
remi-or: also it's super-slow :(

benchmark_v2/framework/hardware_metrics.py
remi-or: I was pleasantly surprised

benchmark_v2/framework/benchmark_config.py
McPatate: Ok gotcha, what is the config name used for? Do you want to create a config from its name?

benchmark_v2/framework/hardware_metrics.py
remi-or: OOP FTW

benchmark_v2/framework/benchmark_config.py
remi-or: Looking at the results mostly

benchmark_v2/framework/benchmark_runner.py
remi-or: now yes

benchmark_v2/run_benchmarks.py
remi-or: done

benchmark_v2/run_benchmarks.py
remi-or: thank you!

benchmark_v2/framework/benchmark_config.py
McPatate: 👁️👄👁️

McPatate: In that case I'd go with a longer spelled out version. It's clear for us in this form, but may not be for end users at first glance.
You could have two methods, one compact and the other just regular.

McPatate: Went too fast, just saw the code below 🤓

benchmark_v2/framework/benchmark_config.py
```diff
@@ -0,0 +1,220 @@
...
+        return "-".join([
+            f"w{self.warmup_iterations}_i{self.measurement_iterations}",
+            "monitored" if self.gpu_monitoring else "unmonitored",
+            f"b{self.batch_size}_s{self.sequence_length}_n{self.num_tokens_to_generate}",
+            attn_code,
+            f"compiled_{self.compile_mode}" if self.compile_mode is not None else "uncompiled",
+            "kernelized" if self.kernelize else "unkernelized",
+        ])
+
+    def infer_pretty_name(self) -> str:
```
McPatate: Even the code is more readable ahah

Maybe `infer_readable_name` and `infer_compact_name` would be best?

benchmark_v2/framework/benchmark_config.py
remi-or: image

benchmark_v2/framework/benchmark_config.py
```diff
@@ -0,0 +1,220 @@
...
+            gpu_monitoring=data.get("gpu_monitoring", False),
+            batch_size=data.get("batch_size", 1),
+            sequence_length=data.get("sequence_length", 128),
+            num_tokens_to_generate=data.get("num_tokens_to_generate", 128),
+            attn_implementation=data.get("attn_implementation", "eager"),
+            sdpa_backend=data.get("sdpa_backend"),
+            compile_mode=data.get("compile_mode"),
+            compile_options=data.get("compile_options"),
+            kernelize=data.get("kernelize", False),
+            name=data.get("name"),
```
McPatate, line 125: No defaults?

Would try except if there is no sane default, to catch the error and properly format it to display to the user.
But perhaps I didn't check how data is initialised and these values are always set.

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -0,0 +1,389 @@
...
+        # Get the array content
+        content = match.group(1)
+        # Remove extra whitespace but keep commas
+        compact_content = re.sub(r"\s+", " ", content).strip()
+        return f"[{compact_content}]"
+
+    return re.sub(pattern, replace_numeric_array, json.dumps(data, indent=4, default=str), flags=re.DOTALL)
+
+
+def get_git_revision() -> str:
```
McPatate, line 68: Are we sure this is more robust than a git cmd?

McPatate, line 68: Are we sure this is more robust than a git cmd?

McPatate, line 68: Are we sure this is more robust than a git cmd?

McPatate, line 68: Are we sure this is more robust than a git cmd?

McPatate, line 68: Are we sure this is more robust than a git cmd?

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -0,0 +1,389 @@
...
+            self.profile_dir = self.output_dir + "_profiles"
+            os.makedirs(self.profile_dir, exist_ok=True)
+        prof.export_chrome_trace(f"{self.profile_dir}/{config_name}.json")
+
+    def run_benchmarks(
+        self,
+        model_id: str,
+        benchmark_configs: list[BenchmarkConfig],
+        num_tokens_to_profile: int = 0,
+        pretty_print_summary: bool = False,
```
McPatate: Not `True` by default?

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -0,0 +1,389 @@
...
+            except Exception as e:
+                self.logger.error(f"Error running with scenario: {config.name}:\n{repr(e)}")
+            # Cleanup model and save results
+            self.cleanup()
+            self.save_results(model_id, all_results, timestamp=timestamp)
+        self.logger.info("All benchmarks done.")
+
+        if pretty_print_summary:
+            print()
+            print("-" * 120)
```
McPatate: Not sure about the separator, I assume there is a tqdm bar from the trange, correct? We can just do one or two new lines, display the config for all benches and then sure a small separator works, though I'd rather it be a title than a `"-" * 120`.

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -0,0 +1,389 @@
...
+        input_tokens = self.inputs["input_ids"].size(-1)
+        batch_size, output_tokens = outputs.shape
+        new_tokens = output_tokens - input_tokens
+        if new_tokens != max_new_tokens:
+            raise RuntimeError(f"Generated {new_tokens} tokens, expected {max_new_tokens}")
+        # Decode outputs
+        decoded_output = self.tokenizer.decode(outputs[0, input_tokens:], skip_special_tokens=True)
+        # Compute intermediate quantities
+        e2e_latency = wall_time_1 - wall_time_0
+        dt_tokens = [t - wall_time_0 for t in streamer.timestamps[1:]]
```
McPatate: What does dt stand for? Time delta?

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -0,0 +1,389 @@
...
+                results = self.run_one_benchmark(model_id, config, num_tokens_to_profile)
+                if results is not None:
+                    all_results[config.hash] = results
+
+            except Exception as e:
+                self.logger.error(f"Error running with scenario: {config.name}:\n{repr(e)}")
+            # Cleanup model and save results
+            self.cleanup()
+            self.save_results(model_id, all_results, timestamp=timestamp)
+        self.logger.info("All benchmarks done.")
```
McPatate: Do we really need this log line? Especially with pretty print (which imo should be on by default)

benchmark_v2/framework/data_classes.py
```diff
@@ -0,0 +1,152 @@
...
+            "hardware_info": self.hardware_info.to_dict(),
+        }
+
+
+class BenchmarkResult:
+    """Result from a series of benchmark runs."""
+
+    def __init__(self) -> None:
+        self.e2e_latency = []
+        self.dt_tokens = []
```
McPatate: I'd spell it out, wasn't obvious to me at first glance

McPatate: I'd spell it out, wasn't obvious to me at first glance

benchmark_v2/framework/data_classes.py
```diff
@@ -0,0 +1,152 @@
...
+                decoded_output=data["decoded_output"][i],
+                gpu_metrics=gpu_metrics[i],
+            )
+        return new_instance
+
+    def get_measured_ttft(self) -> list[float]:
+        return [dt[0] for dt in self.dt_tokens if len(dt) > 0]
+
+    def get_measured_itl(self) -> list[float]:
+        return [(dt[-1] - dt[0]) / (len(dt) - 1) for dt in self.dt_tokens if len(dt) > 1]
```
McPatate: So each list of token deltas is cumulative, correct?

benchmark_v2/framework/hardware_metrics.py
McPatate: 🤢

benchmark_v2/framework/hardware_metrics.py
```diff
@@ -0,0 +1,172 @@
...
+
+class GPUStatsCollector:
+    """A class to get statistics about the GPU. It serves as a wrapper that holds the GPU total memory and its name,
+    which is used to call the right function to get the utilization and memory used."""
+
+    def __init__(self) -> None:
+        self.device_name, self.device_memory_total = get_device_name_and_memory_total()
+        # Monkey patch the get_utilization_and_memory_used method based on the GPU type
+        if "amd" in self.device_name.lower():
+            self.get_utilization_and_memory_used = get_amd_gpu_stats
```
McPatate, line 81: You forgot to rename here no?

benchmark_v2/framework/hardware_metrics.py
```diff
@@ -0,0 +1,172 @@
...
+            )
+            self.logger.debug(f"GPU monitoring completed: {len(self.gpu_utilization)} samples collected")
+        else:
+            metrics = GPURawMetrics(monitoring_status=GPUMonitoringStatus.NO_SAMPLES_COLLECTED)
+        return metrics
+
+    def _monitor_loop(self):
+        """Background monitoring loop using threading.Event for communication."""
+        while not self.stop_event.is_set():
+            utilization, memory_used = self.gpu_stats_getter.get_utilization_and_memory_used()
```
McPatate, line 167: Same here, rename needed

benchmark_v2/run_benchmarks.py
```diff
@@ -19,477 +19,93 @@
...
+                    kwargs["sequence_length"] = sequence_length
+                    kwargs["num_tokens_to_generate"] = num_tokens_to_generate
+                    benchmark_configs.append(BenchmarkConfig(**kwargs))
+
+    runner = BenchmarkRunner(logger, args.output_dir, args.commit_id)
+    results = runner.run_benchmarks(
+        args.model_id,
+        benchmark_configs,
+        args.num_tokens_to_profile,
+        pretty_print_summary=True,
```
McPatate, line 109: Is this what I was going to complain about?

benchmark_v2/framework/benchmark_runner.py
remi-or: works even if the repo is not trusted

remi-or: works even if the repo is not trusted

remi-or: Hum yes i think

remi-or: Idk, starting to have doubts

remi-or: You knwo what I am going to double check

benchmark_v2/framework/benchmark_runner.py
remi-or: Makes sense, changing it

benchmark_v2/framework/data_classes.py
remi-or: Exact, it records the time each token arrived. Will add a com.

benchmark_v2/framework/hardware_metrics.py
remi-or: Not sure what you mean

benchmark_v2/framework/hardware_metrics.py
remi-or: Same

benchmark_v2/run_benchmarks.py
remi-or: lol no, it was the two functions thing

benchmark_v2/framework/benchmark_config.py
remi-or: Sure, will do

benchmark_v2/framework/benchmark_config.py
remi-or: The `.get` method will return None if the key is not found. I would like to be more verbose but `ruff` would complain

benchmark_v2/framework/benchmark_runner.py
remi-or: I moved it to the `else` of pretty print, but then I though it might be usefull to always have this in the log if you want to retrieve total benchmark time rather than have it appear randomly or not. If it's ok with you we will keep it. Also if something screws up in the pretty print we know the benhmarks were finished.

benchmark_v2/framework/data_classes.py
remi-or: replaced with `token_generation_times`. let me know if that works for you.

benchmark_v2/framework/benchmark_runner.py
remi-or: I kinda dont agree here, so I asked claude which would be the best. So the separator will be `"=" * 80` and neither of us will be happy. Compromise!

remi-or: Updated to 100 so claude would not be happy as well.

benchmark_v2/framework/benchmark_config.py
McPatate: And None is a supported value?

benchmark_v2/framework/benchmark_runner.py
McPatate: I meant is this a stable way of fetching this info, we're prying on the internals of git. Was wondering if this could be subject to change and break in the future without warning

McPatate: 😂

McPatate: image

benchmark_v2/framework/benchmark_runner.py
McPatate: 😂

McPatate: I'll refactor this later and get someone else to approve my PR 🙄

benchmark_v2/framework/benchmark_runner.py
McPatate: It'd be odd for the pretty print to fail. I'm ok with not having a log for that if the pretty print is disabled, but I'll let you choose

benchmark_v2/framework/data_classes.py
McPatate: token_generation_deltas also worked 😄

benchmark_v2/framework/hardware_metrics.py
McPatate: Didn't you change the method'a name to get_measurements?

benchmark_v2/run_benchmarks.py
McPatate: Will need proof tmrw

benchmark_v2/framework/hardware_metrics.py
remi-or: No just the name of the retrieved variable

benchmark_v2/framework/benchmark_runner.py
remi-or: Ok ok will remove it as you promised to not touch the separator

PR #41409, opened 2025-10-07
Fix early CUDA initialisation

hmellor:
This import added in https://github.com/huggingface/transformers/pull/41370 was identified to be causing CUDA to initialise early.

It was identified in vLLM's Transformers nightly test https://buildkite.com/vllm/ci/builds/33777/steps/canvas?sid=0199bcd4-8cf7-442a-a94d-df2969d65bde


ArthurZucker (approved):
thanks and sorry about it 😢

PR #41411, opened 2025-10-07
Small Fix for imports 

MekkCyber:
# What does this PR do?

When using `importlib.util.find_spec(pkg_name)`, this might return some packages that are not importable but still exist. In this pr we simply check if they can be loaded

### DIFF
--- a/src/transformers/utils/import_utils.py
+++ b/src/transformers/utils/import_utils.py
@@ -45,7 +45,9 @@ PACKAGE_DISTRIBUTION_MAPPING = importlib.metadata.packages_distributions()
 
 def _is_package_available(pkg_name: str, return_version: bool = False) -> Union[tuple[bool, str], bool]:
     """Check if `pkg_name` exist, and optionally try to get its version"""
-    package_exists = importlib.util.find_spec(pkg_name) is not None
+    spec = importlib.util.find_spec(pkg_name)
+    # the spec might be not None but not importable
+    package_exists = spec is not None and spec.loader is not None
     package_version = "N/A"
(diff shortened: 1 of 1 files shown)


ArthurZucker (approved):
Not sure for which package you had that, a test is always welcome!

MekkCyber:
This happens when you have a folder with the same name as the package in your namespace, so `importlib` assumes it's the python package you are looking for, and marks the loader to be none when there is no `__init__` file. 

Example : 

```
/scripts_training$ mkdir test
/scripts_training$ python
Python 3.12.11 () [Clang 20.1.4 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import importlib.util
>>> importlib.util.find_spec("test")
ModuleSpec(name='test', loader=None, submodule_search_locations=_NamespacePath(['/scripts_training/test', '/scripts_training/test']))
>>> 
```
Not sure how i can add a test for this

ArthurZucker:
AH okay no worries feel free to merge!

BenjaminBossan:
@MekkCyber This commit is resulting in `optimum` not being considered installed, even though it is:

```python
from transformers.utils import is_optimum_available
assert is_optimum_available()
```

fails with this commit (`6850ba853f4a41c843d47e954064b5919a946585`) but works with the previous commit (`bf0bce8d5fcbabcde86e63c2634e87ef8c2614dc`).

MekkCyber:
@BenjaminBossan, thanks for reporting this, I reverted the changes here : https://github.com/huggingface/transformers/pull/41891, and will think about an other way to fix this issue

PR #41412, opened 2025-10-07
[Parakeet] unnecessary warning & auto mapping

eustlb:
# What does this PR do?

- fixes #41387
- and fixes warning 
```
🚨 Config not found for parakeet. You can manually add it to HARDCODED_CONFIG_FOR_MODELS in utils/auto_docstring.py
```
- and fixes auto tokenizer mapping that was not updated when moving from `ParakeetCTCTokenizer` to `ParakeetTokenizerFast`
---
What triggers the unexpected warning:

`@auto_doctring` on a class (not on a method) (here `ParakeetPreTrainedModel`, `ParakeetEncoder`, `ParakeetForCTC`) triggers `auto_method_docstring`  call **with parent_class kwargs setted**, that will trigger looking for a config class with the model name `parakeet` (extracted from `modeling_parakeet.py`). Yet we don't have a config class for `parakeet` alone, we have `ParakeetEncoderConfig` and `ParakeetCTCConfig`.

To solve, let's add `"parakeet": "ParakeetCTCConfig"` to `HARDCODED_CONFIG_FOR_MODELS`.

>[!NOTE]
> When other Parakeet variants will be integrated `ParakeetForTDT` and `ParakeetForRNTT`, then we might want to add a global ParakeetConfig that encapsulate them all


ArthurZucker (approved):
LGTM

PR #41415, opened 2025-10-07
Fix bnb fsdp loading for pre-quantized checkpoint

SunMarc:
# What does this PR do?

This PR fixes bnb loading when using FSDP for pre-quantized checkpoints. This happened because we changed how we load quantized checkpoints as we need to cache all the quantized stats before creating the quantized weight.

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -770,10 +770,10 @@ def _load_state_dict_into_meta_model(
                     if value.device.type == "meta":
                         continue
                     val_kwargs = value.__dict__
-                    if value.dtype in [torch.uint8, torch.int8]:
+                    if not value.is_floating_point():
                         val_kwargs["requires_grad"] = False
-                    param_to = "meta" if is_fsdp_enabled() and not is_local_dist_rank_0() else "cpu"
-                    value = type(value)(value.data.to(param_to), **val_kwargs)
+                    device = "meta" if is_fsdp_enabled() and not is_local_dist_rank_0() else "cpu"
+                    value = type(value)(value.data.to(device), **val_kwargs)
                     setattr(module, param_type, value)
 
         # Remove the param from the state dict if it was not loaded on the fly to avoid wasting memory
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -763,7 +763,7 @@ def _load_state_dict_into_meta_model(
                 # and then cast it to CPU to avoid excessive memory usage on each GPU
                 # in comparison to the sharded model across GPUs.
                 if is_fsdp_enabled() or is_deepspeed_zero3_enabled():
-                    param_name = hf_quantizer.update_param_name(param_name)
+                    param_name = hf_quantizer.get_param_name(param_name)
                     module, param_type = get_module_from_name(model, param_name)
                     value = getattr(module, param_type)
                     # We need to wait until the quantized value is created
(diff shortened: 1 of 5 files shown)


SunMarc:
cc @winglian

src/transformers/modeling_utils.py
```diff
@@ -766,18 +766,14 @@ def _load_state_dict_into_meta_model(
...
                         continue
-                    param_to = "cpu"
-                    if is_fsdp_enabled() and not is_local_dist_rank_0():
-                        param_to = "meta"
-                    val_kwargs = {}
-                    if (hasattr(module, "weight") and module.weight.__class__.__name__ == "Int8Params") or (
-                        value.dtype == torch.uint8 or value.dtype == torch.int8
-                    ):
+                    val_kwargs = value.__dict__
+                    if value.dtype in [torch.uint8, torch.int8]:
```
Cyrilvallez: Maybe just `value.is_floating_point()` if that works?

src/transformers/modeling_utils.py
```diff
@@ -766,18 +766,14 @@ def _load_state_dict_into_meta_model(
...
                         val_kwargs["requires_grad"] = False
-                    value = type(value)(value.data.to(param_to), **val_kwargs, **value.__dict__)
+                    param_to = "meta" if is_fsdp_enabled() and not is_local_dist_rank_0() else "cpu"
```
Cyrilvallez: Let's just call it device IMO, `param_to` is a bit weird

src/transformers/quantizers/quantizer_bnb_4bit.py
```diff
@@ -154,6 +154,19 @@ def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **
         module, name = get_module_from_name(model, param_name)
         return isinstance(module, bnb.nn.Linear4bit) and name != "bias"
 
+    def update_param_name(self, param_name: str) -> str:
+        """
```
Cyrilvallez: Let's maybe call it `get_param_name` instead as it does not update it

src/transformers/modeling_utils.py
```diff
@@ -766,18 +766,14 @@ def _load_state_dict_into_meta_model(
                     param_name = hf_quantizer.update_param_name(param_name)
                     module, param_type = get_module_from_name(model, param_name)
                     value = getattr(module, param_type)
-                    # special case for gpt_oss model, we wait for the param to be leave the meta device before casting it to cpu
-                    if model.config.model_type == "gpt_oss" and value.device.type == "meta":
+                    # We need to wait until the quantized value is created
+                    if value.device.type == "meta":
```
Cyrilvallez, line 770: Still a bit weird to me that we have to do this, but I wanted to investigate further anyway to remove the gpt-oss special exception - already happy to see it a bit more general and not gpt-oss-specific!

Cyrilvallez (approved):
Left a few comments about naming for clarity, otherwise LGTM!

src/transformers/quantizers/quantizer_bnb_4bit.py
SunMarc: done

src/transformers/modeling_utils.py
SunMarc: that should work ! I think that at some point it should be fine to even remove that if the modules are correctly initialized

src/transformers/modeling_utils.py
SunMarc: done

PR #41076, opened 2025-09-22
:rotating_light: [`v5`] Remove headmasking

vasqu:
As per title, time to deprecate this

This includes anything in the docs, tests, and modeling
- `xxx head mask`
- `xxx_headmasking`
- `xxx_head_mask`

PR #41417, opened 2025-10-07
🚨 [v5] Prune `prune_heads`

gante:
# What does this PR do?

Spiritual successor to #41076 -- removes `prune_heads` and related code/tests/docs.

`prune_heads` was a feature that we stopped including in models a while ago. For instance, the `test_head_pruning_save_load_from_config_init` test passes in 28 model classes, but is skipped in 444.


src/transformers/configuration_utils.py
```diff
@@ -123,11 +123,6 @@ class PreTrainedConfig(PushToHubMixin):
         tie_encoder_decoder (`bool`, *optional*, defaults to `False`):
             Whether all encoder weights should be tied to their equivalent decoder weights. This requires the encoder
             and decoder model to have the exact same parameter names.
-        prune_heads (`dict[int, list[int]]`, *optional*, defaults to `{}`):
-            Pruned heads of the model. The keys are the selected layer indices and the associated values, the list of
-            heads to prune in said layer.
-
-            For instance `{1: [0, 2], 2: [2, 3]}` will prune heads 0 and 2 on layer 1 and heads 2 and 3 on layer 2.
```
ydshieh, line 130: this won't break the loading if a saved config has this attribute?

ydshieh (approved):
🔥

src/transformers/configuration_utils.py
ydshieh: ok, we have `kwargs` so it won't break.

But this makes the things not transparent somehow, i.e. the saved config have it but not used at all during loading.

gante: We can load arbitrary serialized kwargs, so it shouldn't :)

Cyrilvallez (approved):
Very very nice! Good riddance! 🧹🤗

PR #41420, opened 2025-10-07
[Qwen3VLMoe] Fixed: Expected self.dtype to be equal to src.dtype - routing_weights casting

danielquintas8:
## What does this PR do?
Fixed: Expected self.dtype to be equal to src.dtype - routing_weights casting

## Related issue
Fixes [#41418](https://github.com/huggingface/transformers/issues/41418)


afonsosilva91:
+1

vasqu:
cc @ArthurZucker for Moe since this might affect more models (?)

i3hz:
I think you didn't run the code which generates the modeling files and that's why you're seeing the difference.

Rocketknight1:
cc @zucchini-nlp

ArthurZucker (approved):
Thanks, can you check this mistake was not propagated to other models please!

danielquintas8:
> Thanks, can you check this mistake was not propagated to other models please!

We found the same casting happening in a few more models (ernie4_5_moe, qwen2_moe, qwen3_moe, qwen3_next, qwen3_omni_moe) and updated them accordingly

The `hunyuan_v1_moe` implementation of `route_tokens_to_experts` has different arguments compared to the other models, so this falls outside the scope of this issue. -> I will inspect it further in a future issue

jaaabir:
@ArthurZucker can you merge this so that we can properly train these models?

danielquintas8:
Hey @ArthurZucker, all checks are green and the PR’s been referenced a couple times.
Just wanted to check if there’s anything else needed before merge. Thanks!

ArthurZucker (approved):
Thanks and sorry for coming back late

PR #41421, opened 2025-10-07
Restore cuda graphs to continuous batching

remi-or:
This PR restores cuda graphs in continuous batching. The main changes associated with this are:
1. the logic of how to generate tokens have been moved to the CB processor, which also handles the cuda graphs
2. the generation step automatically slices the tensors to remove all padding unless cuda graphs are activated
3. cuda graphs are captured on padded shapes, which is 25%, 50%, 75% or 100% of the queries axis and 1/8, ... 8/8 of the keys values axis, to strike a balance between the amount of padding and the quantity of cuda graphs 

Documentation is kind of lacking but will be added in next commits, I am opening the PR so @ArthurZucker can test stuff out

- [x] Add more documentation
- [x] Test it out and add performance numbers with / without on AMD / Nvidia with the three main attn implementations

### DIFF
--- a/src/transformers/generation/continuous_batching/continuous_api.py
+++ b/src/transformers/generation/continuous_batching/continuous_api.py
@@ -27,12 +27,17 @@ from tqdm import tqdm
 
 from ...configuration_utils import PreTrainedConfig
 from ...generation.configuration_utils import GenerationConfig
+from ...generation.logits_process import LogitsProcessor
 from ...integrations.hub_kernels import load_and_register_attn_kernel
 from ...utils.logging import logging
 from ...utils.metrics import ContinuousBatchProcessorMetrics, attach_tracer, traced
 from .cache import PagedAttentionCache
 from .requests import GenerationOutput, RequestState, RequestStatus, get_device_and_memory_breakdown, logger
 from .scheduler import SCHEDULER_MAPPING, FIFOScheduler, Scheduler
+
+
+PADDING = False
+PAD_TO = 10000
 
 
 def build_attention_mask(
@@ -178,8 +183,10 @@ class ContinuousBatchProcessor:
 
         # Retrieve the size of the sliding window if there is one
         self.sliding_window = 1 if getattr(config, "sliding_window", None) is None else config.sliding_window
-
+        # Accumulator for batch scheduling
         self.requests_in_batch: list[RequestState] = []
+        # Cuda graph for the generation step
+        self.graph = None
 
         # Set up metrics collector
         self.max_batch_tokens = cache.max_batch_tokens
@@ -272,14 +279,14 @@ class ContinuousBatchProcessor:
 
         # Reset the attributes that are lists of tensors
         for i in range(self.cache.num_groups):
-            self.write_index_storage[i][:q_len].fill_(-1)
-            self.read_index_storage[i][: q_len + k_len].fill_(-1)
+            self.write_index_storage[i][:q_len].fill_(-2)  # -1 is used to let the cache where new states go
+            self.read_index_storage[i][: q_len + k_len].fill_(-2)  # same
 
     def get_model_kwargs(self) -> PagedAttentionArgs:
         """Get model keyword arguments for the current batch."""
         # Compute the slice to return
-        q_len = self.actual_query_length
-        b_size = self.actual_batch_size
+        q_len = self.max_batch_tokens if PADDING else self.actual_query_length
+        b_size = self.max_batch_tokens if PADDING else self.actual_batch_size
 
         # Prepare the kwargs, the attributes that are either tensors or dict of tensors are initialized to empty dicts
         kwargs = {
@@ -299,7 +306,7 @@ class ContinuousBatchProcessor:
 
         # For the attributes that are lists of tensors, we construct list of tensor references
         for i in range(self.cache.num_groups):
-            r, w = self.actual_group_indices[i]
+            r, w = (PAD_TO + self.max_batch_tokens, self.max_batch_tokens) if PADDING else self.actual_group_indices[i]
             kwargs["read_index"].append(self.read_index_storage[i][:r])
             kwargs["write_index"].append(self.write_index_storage[i][:w])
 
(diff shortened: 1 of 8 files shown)


remi-or:
Some performances (with / without CG are for this PR) with `do_sample = False` and `FORCE_MAX_LENGTH = True` , all run with `python examples/pytorch/continuous_batching.py -mp none --samples 100`
| Model name | Attention | tok/sec on __main__ | tok/sec __without__ CG | tok/sec __with__ CG |
|:----------:|:---------:|:-------------------:|-----------------------:|--------------------:|
| Llama      | Eager     | 849.27tok/s  | 888.42tok/s  | 691.33tok/s  |
| Llama      | SDPA      | 1040.93tok/s | 1070.21tok/s | 889.91tok/s  |
| Llama      | Flash     | 2263.14tok/s | 2512.48tok/s | 3550.99tok/s |
| Qwen       | Eager     | 769.19tok/s  | 806.18tok/s  | 627.58tok/s  |
| Qwen       | SDPA      | 940.70tok/s  | 975.02tok/s  | 816.32tok/s  |
| Qwen       | Flash     | 1855.22tok/s | 1934.03tok/s | 3552.07tok/s |

I have not verified parity but outputs are not gibberish.

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -35,10 +36,45 @@
...
+To solve this, we pad those dimensions using fixed intervals. The size of the intervals is controlled by the variables
+below: NUM_X_CUDA_GRAPHS means that we create at most NUM_X_CUDA_GRAPHS graphs for the X dimension. So if the maximum
+number of queries tokens is 1000, and NUM_Q_CUDA_GRAPHS is 4, we will pad the number of queries token using intervals of
+1000 / 4 = 250 tokens, ie. to 250, 500, 750 or 1000 queries tokens.
+
+More intervals means more granularity and thus less padding. But since each graph takes up space on the GPU and time to
+create, we don't want to many graphs. And since the size of the KV dimension is the number of queries tokens plus the
+number of tokens cached, dimension of KV is usually much larger than the the dimension of Q. So we have more granularity
+for the KV dimension than the query dimension.
+"""
```
ArthurZucker, line 56: would add that we don't really pad, allocate once and slice

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -219,13 +259,13 @@ def setup_static_tensors(self, num_groups: int) -> None:
         layer_types = list(set(layer_types))
 
         self.cumulative_seqlens_k = {
-            layer_type: torch.empty((T + 1), **self.tensor_metadata) for layer_type in layer_types
+            l_type: torch.empty((self.max_batch_tokens + 1), **self.tensor_metadata) for l_type in layer_types
```
ArthurZucker, line 260: I perfer layer_type but its a super small nit

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -298,43 +336,55 @@ def get_model_kwargs(self) -> PagedAttentionArgs:
...
+            "write_index": [],
             "cache": self.cache,
             "use_cache": False,
         }
 
+        # If there is padding, we need to make sure all padding tokens are attended to, to avoid NaNs
+        if use_padding:
+            self.max_seqlen_q = max(self.max_seqlen_q, q_len - self.total_seqlen_q)
+            self.cumulative_seqlens_q[self.actual_batch_size+1:] = q_len
+            # TODO: can this be avoided with setting tokens to non-nan after the foward? might be faster.
```
ArthurZucker: that is very very counter intuitive 🤣 
We should mark this as a `# FIXME: ` because we might be wasting some compute there

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -35,10 +36,45 @@
...
+number of queries tokens is 1000, and NUM_Q_CUDA_GRAPHS is 4, we will pad the number of queries token using intervals of
+1000 / 4 = 250 tokens, ie. to 250, 500, 750 or 1000 queries tokens.
+
+More intervals means more granularity and thus less padding. But since each graph takes up space on the GPU and time to
+create, we don't want to many graphs. And since the size of the KV dimension is the number of queries tokens plus the
+number of tokens cached, dimension of KV is usually much larger than the the dimension of Q. So we have more granularity
+for the KV dimension than the query dimension.
+"""
+NUM_Q_CUDA_GRAPHS = 4
+NUM_KV_CUDA_GRAPHS = 8
```
ArthurZucker, line 58: not something we want here, let's pass this as args to the api

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -35,10 +36,45 @@
...
+
+
+def pad_by_intervals(size: int, max_value: int, nb_intervals: int) -> int:
```
ArthurZucker, line 61: there's probably a faster way to do this but np

ArthurZucker (approved):
Nice perf boosts!

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: Agreed, fixed

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: I think it's a good place to place the docstring though... I am going to add a parameter to the API, but let `NUM_Q_CUDA_GRAPHS` as the default value (and changing the name). Let me know if that sounds good to you or not

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: Yes this was suboptimal, making the change

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: Yes the bug is kinda weird anyway
I tried removing the block on LLama (which does not have the NaN issue anyway, only Qwen does) and the perf were:
```
with block: 3271, 3273, 3269 (tok/s)
without block: 3261, 3269, 3265 (tok/s)
```
so I think there is some perf impact but it's minimal.  Adding a fixme though, and a slightly longer comment.

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: its to avoid the 120 chars limit

ArthurZucker: not a good argument haha

src/transformers/generation/continuous_batching/continuous_api.py
ArthurZucker: doc yes, I am talking about the args!

PR #41423, opened 2025-10-07
[`JetMoe`] Fix KV head repetition and padding free

vasqu:
During the vllm refactor 2 small issues have been introduced
- Jetmoe doesn't repeat the kv heads with `repeat_interleave` but uses `repeat`, i.e. different patterns
- Position ids wasn't passed down leading to padding free not working

Should hopefully align the tests again in #41377 (at least checked locally on our dgx machines that the outputs are matching again) cc @ydshieh

### DIFF
--- a/src/transformers/models/jetmoe/modeling_jetmoe.py
+++ b/src/transformers/models/jetmoe/modeling_jetmoe.py
@@ -427,7 +427,8 @@ class JetMoeAttention(nn.Module):
                 "when creating this class."
             )
 
-        self.num_key_value_groups = config.num_experts_per_tok
+        self.num_key_value_groups = 1  # We ignore this by setting it to 1 as we have different repeat patterns
+        self.top_k = config.num_experts_per_tok
         self.attention_dropout = config.attention_dropout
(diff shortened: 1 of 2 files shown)


src/transformers/models/jetmoe/modeling_jetmoe.py
```diff
@@ -427,7 +427,8 @@ def __init__(self, config: JetMoeConfig, layer_idx: Optional[int] = None):
                 "when creating this class."
             )
 
-        self.num_key_value_groups = config.num_experts_per_tok
+        self.num_key_value_groups = 1  # We ignore this by setting it to 1 as we have different repeat patterns
+        self.top_k = config.num_experts_per_tok
```
ydshieh, line 431: Is this what we have done before #40132, or it's kind new stuff?

vasqu: That was the naming before the refactor + `num_key_value_groups` is used in our interface so I basically turn it off, i.e. no double repetitions. (it would need a different naming either way then)

ArthurZucker (approved):
Arf I did try to standardize too much by removing top_k here! thanks for following up!

PR #41426, opened 2025-10-07
fix gemma3n case failure

yao-matrix:
# Problem
`pytest -rA tests/models/gemma3n/test_modeling_gemma3n.py::Gemma3nIntegrationTest::test_generation_beyond_sliding_window_with_generation_config` generate different output even `do_sample=False` in `generation_config = GenerationConfig(max_new_tokens=20, do_sample=False)`, which makes UT fails

# Root cause
custom generation_config will be overode by model generation_config even it explicitly set, when the set value equals global default.

# Fix
Add a condition to make the overriding only happens when the custom_generation_config doesn't explicitly set the field.  

@SunMarc , pls help review, thx very much.


yao-matrix:
seems the ci failures are not related to my PR.

src/transformers/generation/utils.py
```diff
@@ -1750,9 +1750,11 @@ def _prepare_generation_config(
                     if key == "cache_implementation" and model_generation_config.cache_implementation == "hybrid":
                         continue
                     global_default_value = getattr(global_default_generation_config, key, None)
+                    has_custom_gen_config_value = hasattr(generation_config, key)
                     custom_gen_config_value = getattr(generation_config, key, None)
                     if (
-                        custom_gen_config_value == global_default_value
+                        (not has_custom_gen_config_value)
+                        and custom_gen_config_value == global_default_value
```
yao-matrix: Here is the case:
When user sets `generation_config = GenerationConfig(max_new_tokens=20, do_sample=False)`, here `do_sample` is False, while global_default_value is also `False`, and `model_gen_config_value` is True, the current condition cannot distinguish whether the value is explicitly set by user or just the global default, so it just modify it to model default. Add another condition to make sure it happens only when user didn't explicitly specify the value.

zucchini-nlp: Indeed this is an issue, and afair we discussed it with @gante after merging the PR with config value overwriting. cc @gante for review

gante (approved):
There are probably a few more cases where this inheritance is still broken, but the fix looks sensible in general :)

gante:
Actually, we can see in the CI that these changes break other cases :'(

@yao-matrix , let's instead add `use_model_defaults=False` in the `model.generate` call that is creating issues. We need to fully rewrite this attribute inheritance :(

yao-matrix:
> pytest -rA tests/models/gemma3n/test_modeling_gemma3n.py::Gemma3nIntegrationTest::test_generation_beyond_sliding_window_with_generation_config

done, pls review again, and do you need me create an issue to track the `generation_config` issue? It seems a HF recommended way to use `generate` in some docs and blogs.

yao-matrix:
@zucchini-nlp @gante , could you pls help review again? Thx

PR #41427, opened 2025-10-07
[`FA`] Cleanup loading logic

vasqu:
Multitude of miscellaneous fixes in regards to how we load flash attention
- Less important but it enables also flash attention via config setting, i.e. you can freely set `config._attn_implementation` and it will properly switch
- Fallback to kernels was broken and activated when other kernels were requested like the metals kernel
- Switch to new kernels naming for fa2 (even if it's redirected)
- CB specific fixes
  - FA3 is now included 
  - Centralized loading logic, should make it easier to directly include things for base fa and CB at the same time


src/transformers/integrations/hub_kernels.py
```diff
@@ -211,8 +210,8 @@ def load_and_register_kernel(attn_implementation: str) -> None:
     if hasattr(kernel, "flash_attn_varlen_func"):
         if attention_wrapper is None:
             attention_wrapper = flash_attention_forward
-        kernel_function = partial(attention_wrapper, implementation=kernel)
```
vasqu, line 265: This didn't have any effect currently and we only used the forced loading to get the correct implementation

src/transformers/integrations/hub_kernels.py
```diff
@@ -211,14 +210,16 @@ def load_and_register_kernel(attn_implementation: str) -> None:
...
-        lazy_import_flash_attention(kernel, force_import=True)
```
vasqu, line 266: Lazy import only happens at init or explicit setting time https://github.com/huggingface/transformers/blob/766ac18d898f15ba905934d22f7fd56042d12a25/src/transformers/modeling_utils.py#L2580-L2582
https://github.com/huggingface/transformers/blob/ced85ca7dacd1e828922415043ac1e3d248d3e89/src/transformers/generation/continuous_batching/continuous_api.py#L614-L618

Otherwise, it can happen when users only set the attention within the config, i.e. the interface detects a different implementation name than the currently loaded one https://github.com/huggingface/transformers/blob/766ac18d898f15ba905934d22f7fd56042d12a25/src/transformers/modeling_flash_attention_utils.py#L143-L144

src/transformers/integrations/hub_kernels.py
ArthurZucker: no this is super important for `paged_attention` wrappers 😓

ArthurZucker (commented):
make sure you test the `continious_batching_simple` script with fa2 !

src/transformers/integrations/hub_kernels.py
vasqu: Changed the logic so that CB is now able to use all fa versions with lazy loading, i.e. fa2, fa3, kernels fas

vasqu:
@ArthurZucker updated the logic a bit to include CB now as well. The loaded attention is now properly dependent on the config like in the base fa version + we can now use fa3 as well for CB.

ArthurZucker (commented):
If we do:
```
- partial(attention_wrapper, implementation=kernel)
```
in that case we should remove this arg from all xxx_paged

vasqu:
Seems like the `implementation` kwarg was only used in `flash_paged`, the other implementations don't have it:
- eager https://github.com/huggingface/transformers/blob/a99b1be3c73173509f71c261834d93d3d5e30ef1/src/transformers/integrations/eager_paged.py#L21-L29
- sdpa https://github.com/huggingface/transformers/blob/a99b1be3c73173509f71c261834d93d3d5e30ef1/src/transformers/integrations/sdpa_paged.py#L20-L29
- flash removed in this PR https://github.com/huggingface/transformers/blob/1ece1936f9d8293abfbbd6006d84980f344e504d/src/transformers/integrations/flash_paged.py#L9-L21
So we should be good @ArthurZucker?

ArthurZucker (approved):
thanks yeah makes it a bit simpler !

PR #41429, opened 2025-10-07
enhance patched_tearDown to support python 3.11+

yao-matrix:
@ydshieh , I tried your `PATCH_TESTING_METHODS_TO_COLLECT_OUTPUTS=yes`, it's cool. And my env is Python 3.12, so the `_Outcome.errors` has changed to `_Outcome.result.errors`, so I do a minor enhancement to make it work in my env. @ydshieh  Pls help review, thx very much.


ydshieh (approved):
Great, thank you a lot!

PR #41433, opened 2025-10-08
Fix: remove invalid PerceptionEncoder and ParakeetCTCTokenizer registry entries (#41387)

aijadugar:
This PR fixes issue #41387 by commenting out invalid auto registry entries:

- Removed non-existent `PerceptionEncoder` mapping in `modeling_auto.py`
- Removed non-existent `ParakeetCTCTokenizer` mapping in `tokenization_auto.py`

Local tests:

image

### DIFF
--- a/src/transformers/models/parakeet/tokenization_parakeet_fast.py
+++ b/src/transformers/models/parakeet/tokenization_parakeet_fast.py
@@ -17,6 +17,21 @@ import itertools
 from typing import Optional, Union
 
 from ...tokenization_utils_fast import PreTrainedTokenizerFast
+from ...tokenization_utils_base import PreTrainedTokenizerBase
+
+class ParakeetCTCTokenizer(PreTrainedTokenizerBase):
+    def __init__(self, vocab_file=None, **kwargs):
+        super().__init__()
+        self.vocab_file = vocab_file
+
+    def _tokenize(self, text):
+        return text.split()
+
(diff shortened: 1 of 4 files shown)


src/transformers/models/parakeet/tokenization_parakeet_fast.py
```diff
@@ -17,6 +17,21 @@
...
+        self.vocab_file = vocab_file
+
+    def _tokenize(self, text):
+        return text.split()
+
+    def _convert_token_to_id(self, token):
+        return 0
+
+    def _convert_id_to_token(self, index):
+        return ""
```
eustlb, line 34: Should not be added!

src/transformers/models/perception_lm/modeling_perception_lm.py
```diff
@@ -30,10 +30,21 @@
...
+class PerceptionEncoder(PreTrainedModel):
+    config_class = PretrainedConfig
+
+    def __init__(self, config):
+        super().__init__(config)
+        self.dummy_layer = None
+
+    def forward(self, x):
+        return x
+
```
eustlb, line 47: Should not be added!

eustlb (commented):
Thanks for the effort 🤗, nevertheless that is not the way this should be fixed.
#41412 will fix Parakeet

PR #41435, opened 2025-10-08
Allow passing `tp_plan` in `from_pretrained` directly

Cyrilvallez:
# What does this PR do?

As per the title. Also fixes the tp_plan internal for non-based models.

I checked with the following, and it works nicely:

```python
import torch
from torch.distributed.tensor.debug import visualize_sharding
from torch.distributed.tensor import DTensor
import os

rank = int(os.environ["RANK"])

model_id = "meta-llama/Llama-3.1-8B-Instruct"
# only shard attentions, but not mlps
tp_plan = {
    "model.layers.*.self_attn.q_proj": "colwise",
    "model.layers.*.self_attn.k_proj": "colwise",
    "model.layers.*.self_attn.v_proj": "colwise",
    "model.layers.*.self_attn.o_proj": "rowwise",
}

    model_id,
    dtype=torch.bfloat16,
    tp_plan=tp_plan,
)

inputs = tokenizer(["How are you?"], return_tensors="pt").to(rank)
out = model.generate(**inputs, max_new_tokens=10, do_sample=False)
text = tokenizer.batch_decode(out[:,inputs.input_ids.shape[-1]:])

if rank == 0:
    print(text)
    print(model.tp_plan)
    print("SHARDING OF Q_PROJ")
    visualize_sharding(model.model.layers[0].self_attn.q_proj.weight)
    print("SHARDING OF UP_PROJ")
    if isinstance(model.model.layers[0].mlp.up_proj.weight, DTensor):
        visualize_sharding(model.model.layers[0].mlp.up_proj.weight)
    else:
        print("NOT SHARDED")

```


src/transformers/modeling_utils.py
```diff
@@ -1883,10 +1883,12 @@ def post_init(self):
...
 
+        self._tp_plan, self._ep_plan, self._pp_plan = {}, {}, {}
         # If current model is a base model, attach `base_model_tp_plan` and `base_model_pp_plan` from config
-        self._pp_plan = self.config.base_model_pp_plan.copy() if self.config.base_model_pp_plan is not None else {}
-        self._tp_plan = self.config.base_model_tp_plan.copy() if self.config.base_model_tp_plan is not None else {}
-        self._ep_plan = self.config.base_model_ep_plan.copy() if self.config.base_model_ep_plan is not None else {}
+        if self.base_model is self:
+            self._pp_plan = self.config.base_model_pp_plan.copy() if self.config.base_model_pp_plan is not None else {}
+            self._tp_plan = self.config.base_model_tp_plan.copy() if self.config.base_model_tp_plan is not None else {}
+            self._ep_plan = self.config.base_model_ep_plan.copy() if self.config.base_model_ep_plan is not None else {}
```
Cyrilvallez, line 1892: This is unrelated, but currently the `_tp_plan` of ForCausalLM models would contain the base_model_plan 2 times, one with the `base_model_prefix` prefix, and one without - this fixes it

src/transformers/modeling_utils.py
```diff
@@ -1909,54 +1911,43 @@ def pp_plan(self) -> dict[str, tuple[str, str]]:
...
-                                        elif i < len(param_parts) and pattern_part == param_parts[i]:
-                                            match_count += 1
-                                    if match_count == len(pattern_parts):
-                                        flexible_matched = True
-                                        break
-                            if not flexible_matched:
-                                warnings.warn(
-                                    f"Layer pattern '{layer_pattern}' does not match any parameters in the model. "
-                                    f"This rule may not be applied during tensor parallelization."
-                                )
```
Cyrilvallez, line 1957: Simplify this check a lot: we only want direct matches, otherwise something is wrong

ArthurZucker: I think quantization required that, as I don't remmeber this code at all hahah but good to update!

ArthurZucker (approved):
Thanks, missing a test 😉

src/transformers/modeling_utils.py
Cyrilvallez: It was probably useful at the time when this function was used all the time anyway, i.e. we would all the time set `self.tp_plan = plan` which would run the check. Now, as we write in `_tp_plan` directly, instead of using the `setter`, the check is not run when not using custom tp. 
I checked quantization slow tests and they are fine!

PR #41439, opened 2025-10-08
Patch MistralCommonTokenizer

juliendenize:
# What does this PR do?

This PR patches the `MistralCommonTokenizer`:
1. [BUG FIX] spm now is correctly supported as previous usage would result in an error for `_piece_to_id` and `_is_control_token`.
2. [BUG FIX] previous implementation of `_piece_to_id` was incorrect for Tekkenizer.

a) the special tokens were not supported
b) the normal tokens were not shifted by adding the number of special tokens

4. [MAYBE BUG FIX] Changed `get_vocab` to a function that should better mimic what happens in `Transformers`: now the mapping is based on the real ids but some ids are missing due to conversion loss of some tokens in Tekken.
5. [FEATURE] `add_generation_prompt` has been added. This is to match signature of Transformers. In practice this value is ignored except if:

a) `continue_final_message` and `add_generation_prompt` are `True` an error is raised.
b) if `add_generation_prompt` is `True` and the last message is `assistant` then an error is raised as the user should have passed `continue_final_message`.

6. ~[FEATURE] Now the tokenizer is set to fast because:~

a) ~it is true~(Edit: not so sure about that after discussion)
b) it removes annoying message when initializing the tokenizer.

Edit 2: this has been reverted as it doesn't bring value and is misleading.

7. [OPTIMIZATION] now image tensors are converted without slowness warnings from torch.
8. [DOCS] updated some docs to remove unused args.

Also added minimal tests to ensure SPM also works in the future.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@patrickvonplaten @ArthurZucker @itazap

### DIFF
--- a/src/transformers/models/auto/tokenization_auto.py
+++ b/src/transformers/models/auto/tokenization_auto.py
@@ -436,7 +436,9 @@ TOKENIZER_MAPPING_NAMES = OrderedDict[str, tuple[Optional[str], Optional[str]]](
                 "MistralCommonTokenizer"
                 if is_mistral_common_available()
                 else ("LlamaTokenizer" if is_sentencepiece_available() else None),
-                "LlamaTokenizerFast" if is_tokenizers_available() and not is_mistral_common_available() else None,
+                "MistralCommonTokenizer"
+                if is_mistral_common_available()
+                else ("LlamaTokenizerFast" if is_tokenizers_available() else None),
             ),
         ),
         (
--- a/src/transformers/models/auto/tokenization_auto.py
+++ b/src/transformers/models/auto/tokenization_auto.py
@@ -436,9 +436,7 @@ TOKENIZER_MAPPING_NAMES = OrderedDict[str, tuple[Optional[str], Optional[str]]](
                 "MistralCommonTokenizer"
                 if is_mistral_common_available()
                 else ("LlamaTokenizer" if is_sentencepiece_available() else None),
-                "MistralCommonTokenizer"
-                if is_mistral_common_available()
-                else ("LlamaTokenizerFast" if is_tokenizers_available() else None),
+                "LlamaTokenizerFast" if is_tokenizers_available() and not is_mistral_common_available() else None,
             ),
         ),
         (
--- a/src/transformers/tokenization_mistral_common.py
+++ b/src/transformers/tokenization_mistral_common.py
@@ -1772,9 +1772,10 @@ class MistralCommonTokenizer(PushToHubMixin):
             raise ValueError("`init_inputs` are not supported by `MistralCommonTokenizer.from_pretrained`.")
 
         # Handle kwargs and AutoTokenizer case
-        if kwargs and not set(kwargs.keys()).issubset({"_from_auto", "trust_remote_code"}):
+        ignore_subset = {"_from_auto", "trust_remote_code"}
+        if kwargs and not (set_kwargs := set(kwargs.keys())).issubset(ignore_subset):
             raise ValueError(
-                f"Kwargs {list(kwargs.keys())} are not supported by `MistralCommonTokenizer.from_pretrained`."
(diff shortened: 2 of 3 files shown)


tests/test_tokenization_mistral_common.py
```diff
@@ -79,6 +79,10 @@ def setUpClass(cls):
         cls.ref_tokenizer: MistralTokenizer = MistralTokenizer.from_hf_hub(
             cls.repo_id, local_files_only=cls.local_files_only
         )
+
+        # Define SPM tokenizer to test the private methods that handle SPM and Tekken differencies.
+        cls.spm_repo_id = "mistralai/Mistral-7B-Instruct-v0.3"
```
juliendenize: Is it possible to add access to this gated model for the CI ?

src/transformers/models/auto/tokenization_auto.py
```diff
@@ -429,7 +429,9 @@
                 "MistralCommonTokenizer"
                 if is_mistral_common_available()
                 else ("LlamaTokenizer" if is_sentencepiece_available() else None),
-                "LlamaTokenizerFast" if is_tokenizers_available() and not is_mistral_common_available() else None,
```
patrickvonplaten, line 439: is this a good idea?

src/transformers/tokenization_mistral_common.py
```diff
@@ -277,6 +273,10 @@ def __init__(
 
         self._cache_get_vocab: Optional[dict[str, int]] = None
 
+    @property
+    def is_fast(self) -> bool:
+        return True
```
patrickvonplaten: Does tekken and llama fast tokenizer always match?

patrickvonplaten (approved):

src/transformers/models/auto/tokenization_auto.py
juliendenize: I think so yes, ideally we'd like mistral-common to supersede other tokenizers when available.

src/transformers/tokenization_mistral_common.py
juliendenize: AFAIK fast tokenizer is an implementation of tokenizers in rust no ?

Would love input from HF team here !

src/transformers/models/auto/tokenization_auto.py
ArthurZucker: We don't mind, this is going to be changed a bit in v5 anyway (the distinction between slow and fast)

src/transformers/tokenization_mistral_common.py
ArthurZucker: tekken == sentencepiece no? So yeah if you have the correct `Metaspace` pretokenizer attr (`prepend_scheme="first"`) then yes

ArthurZucker (approved):
Thanks for updating! 🤗

src/transformers/tokenization_mistral_common.py
juliendenize: tekken is based on tiktoken and is used for all our most recent models. Previous releases were based on sentencepiece.

Regarding Metaspace, we don't define a pre-tokenizer using mistral-common.

Not completly sure of the implications here and I think i lack context and what exactly a FastTokenizer does.

I don't want to mislead the users, it was mainly to avoid warnings from Transformers when loading the model as `slow` or `fast` don't really make sense for our tokenizer if it's simply a matter of speed. But after looking at this https://huggingface.co/learn/llm-course/en/chapter6/3#fast-tokenizers-special-powers maybe it's a bad idea.

@ArthurZucker do you believe I should revert that ?

juliendenize: I reverted following a discussion as it isn't needed but is misleading.

src/transformers/models/auto/tokenization_auto.py
juliendenize: Reverted as it's misleading and doesn't bring value.

PR #41440, opened 2025-10-08
Fix generate outputs and simplify cache tests

Cyrilvallez:
# What does this PR do?

Improve the cache tests.`test_past_key_values_format` should really rely on `_check_past_key_values_for_generate` as they do the same, and we otherwise have to overwrite both for models with non-standard cache format, which is really annoying!

This made me realize that `generate` does not correctly return the cache, if it's not called `past_key_values`. This fixes it as well.

cc @gante


gante (approved):
Thank you for updating these tests 🤗 🔥 (and for fixing the returned cache bug)

PR #41441, opened 2025-10-08
Enhance the handling of Union types in HfArgumentParser

cyyever:
# What does this PR do?

This PR enhance the handling of Union types containing `str` by two rules:
1. If `str` in `Union` and `Union` has more than one other types pass `str` to `parser.add_argument`. The caller is responsible to convert the string argument to a more proper type.
2. Otherwise, `Union` must be `X | str` and pass `X` to `parser.add_argument`

### DIFF
--- a/src/transformers/hf_argparser.py
+++ b/src/transformers/hf_argparser.py
@@ -176,9 +176,12 @@ class HfArgumentParser(ArgumentParser):
                     f" Problem encountered in field '{field.name}'."
                 )
             if type(None) not in field.type.__args__:
-                # filter `str` in Union
-                field.type = field.type.__args__[0] if field.type.__args__[1] is str else field.type.__args__[1]
-                origin_type = getattr(field.type, "__origin__", field.type)
+                if len(field.type.__args__) > 2:
+                    origin_type = str
+                else:
+                    # filter `str` in Union
+                    field.type = field.type.__args__[0] if field.type.__args__[1] is str else field.type.__args__[1]
+                    origin_type = getattr(field.type, "__origin__", field.type)
             elif bool not in field.type.__args__:
                 # filter `NoneType` in Union (except for `Union[bool, NoneType]`)
                 field.type = (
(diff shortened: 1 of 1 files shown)


cyyever:
@SunMarc This can resolve `fsdp` error.

Rocketknight1:
CI seems green so gentle ping @SunMarc

cyyever:
@Rocketknight1  @SunMarc  We also need it for moving to Python 3.10 `|` typing that unifying `Optional` and `Union` for config classes.

src/transformers/hf_argparser.py
```diff
@@ -176,9 +176,12 @@ def _parse_dataclass_field(parser: ArgumentParser, field: dataclasses.Field):
...
             if type(None) not in field.type.__args__:
-                # filter `str` in Union
-                field.type = field.type.__args__[0] if field.type.__args__[1] is str else field.type.__args__[1]
-                origin_type = getattr(field.type, "__origin__", field.type)
+                if len(field.type.__args__) > 2:
+                    origin_type = str
+                else:
+                    # filter `str` in Union
+                    field.type = field.type.__args__[0] if field.type.__args__[1] is str else field.type.__args__[1]
+                    origin_type = getattr(field.type, "__origin__", field.type)
```
SunMarc: This is a really special case when None is not in the field. In most cases, this don't happen at all. For example for fsdp, we have `dict[str, Any] | str | None.` So it is not going here but instead in the
` elif bool not in field.type.__args__:` condition. 

Could you try to clean a bit the mess here ? From what I see:
1. If we have bool + None, we don't touch 
2. if we don't have none, we only allow one class
3. if we have none, we can allow more:
  3.1 if we have str and other types, we should enforce the user to pass str. 
  3.2 if we don't have str, we use the first type passed

SunMarc (commented):
Thanks, this was always been a bit of a mess, I left a comment regarding how to fix this, would you like to fix fixing this ?

src/transformers/hf_argparser.py
SunMarc: If you can add some tests testing those, that would be huge also

PR #37878, opened 2025-04-29
PerceptionLM

shuminghu:
This PR implements PerceptionLM released by Meta:
https://github.com/facebookresearch/perception_models

PR #41442, opened 2025-10-08
Fix/perceptionlm auto alias

omsherikar:
### What does this PR do?

Removes references to a non-existent base model from the Auto mappings so enumeration and lazy imports work as expected.

- Remove `perception_encoder` entries from Auto mappings
  - `src/transformers/models/auto/modeling_auto.py`
  - `src/transformers/models/auto/configuration_auto.py`
- Keep existing `perception_lm` entries intact
- Regenerate modeling files (no runtime behavior change)

Fixes: [#41387](https://github.com/huggingface/transformers/issues/41387)

### Motivation

`MODEL_MAPPING.items()` failed because Auto mappings referenced `perception_encoder`, which doesn’t exist. Per maintainer guidance, we should remove non-existent models from auto-map rather than adding aliases.

### Dependencies

None.

### Before submitting

- [x] Minimal change; no behavior changes.
- [x] Linked the issue.
- [x] No docs changes required.
- [x] Local check with `PYTHONPATH=src`:
```python
from transformers.models.auto.modeling_auto import MODEL_MAPPING
list(MODEL_MAPPING.items())  # succeeds
```

### Who can review?

- Model loading / auto mappings: @CyrilVallez
- Multimodal: @zucchini-nlp


Rocketknight1:
@zucchini-nlp @Cyrilvallez this was added with the PerceptionLM PR: https://github.com/huggingface/transformers/pull/37878, I'm not sure why the mapping was broken, or why the tests didn't pick up a nonexistent class!

PR #41443, opened 2025-10-08
Qwen3: fix bool attention_mask in eager_attention_forward

ZhengKai91:
This PR fixes a silent numerical/logic bug that occurs when a boolean attention_mask is supplied to the eager attention path.
The current implementation simply adds the mask to the attention scores, which is correct for float masks that already contain −∞ values, but produces completely wrong logits when the mask is of type torch.bool.
The fix converts a boolean mask to float and fills the masked-out positions with −∞ before the addition, guaranteeing that softmax will assign them zero probability.
No breaking change.
 @vasqu @ArthurZucker  @Cyrilvallez

### DIFF
--- a/src/transformers/models/llama/modeling_llama.py
+++ b/src/transformers/models/llama/modeling_llama.py
@@ -182,9 +182,13 @@ def eager_attention_forward(
 
     attn_weights = torch.matmul(query, key_states.transpose(2, 3)) * scaling
     if attention_mask is not None:
-        causal_mask = attention_mask[:, :, :, : key_states.shape[-2]]
+        if attention_mask.dtype==torch.bool:
+            casual_mask = torch.zeros_like(attention_mask, dtype=torch.float)
+            causal_mask = casual_mask.masked_fill(~attention_mask, float("-inf"))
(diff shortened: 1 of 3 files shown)


vasqu:
Hey @ZhengKai91, do you have a repro where you encounter a boolean mask there? 

This shouldn't happen, i.e. we always create a float mask for eager, so I would be surprised if not https://github.com/huggingface/transformers/blob/8dfc8e8cfc80492e3af0e3d50dcc1803d3654634/src/transformers/masking_utils.py#L519-L521

ZhengKai91:
> Hey @ZhengKai91, do you have a repro where you encounter a boolean mask there?
> 
> This shouldn't happen, i.e. we always create a float mask for eager, so I would be surprised if not
> 
> https://github.com/huggingface/transformers/blob/8dfc8e8cfc80492e3af0e3d50dcc1803d3654634/src/transformers/masking_utils.py#L519-L521

Closing this PR—turns out the real issue was my incorrectly forcing a `sdpa_mask` into the eager interface, which produced the faulty logits. Thanks for the quick feedback!

PR #41444, opened 2025-10-08
Streaming should be handled at the request-level rather than at the istance level

LysandreJik:


ArthurZucker (commented):
Missing a test but marvelous if it works as is!

McPatate (approved):
lgtm!

ArthurZucker (approved):
Nice!

PR #41445, opened 2025-10-08
[`from_pretrained`] Small refactor `from_pretrained`: move around unrelated stuff

ArthurZucker:
# What does this PR do?
As we are moving towards having a dynamic weight loader, we need to have a simple `from_pretrained` call. 
This should help


src/transformers/generation/utils.py
```diff
@@ -363,6 +364,59 @@ class GenerationMixin(ContinuousMixin):
...
+        force_download,
+        proxies,
+        local_files_only,
+        token,
+        revision,
+        subfolder,
+        trust_remote_code,
+        **kwargs,
+    ):
+        if self.can_generate() and generation_config is not None:
```
Cyrilvallez, line 385: `self.can_generate()` is necessarily True if called from inside `GenerationMixin` so can be removed

src/transformers/generation/utils.py
```diff
@@ -363,6 +364,59 @@ class GenerationMixin(ContinuousMixin):
...
+            logger.info("The user-defined `generation_config` will be used to override the default generation config.")
```
Cyrilvallez: I think we can fully remove the info wdyt?

src/transformers/generation/utils.py
```diff
@@ -363,6 +364,59 @@ class GenerationMixin(ContinuousMixin):
...
+            self.generation_config = self.generation_config.from_dict(generation_config.to_dict())
+        elif self.can_generate() and pretrained_model_name_or_path is not None:
```
Cyrilvallez, line 387: same as before

src/transformers/modeling_utils.py
```diff
@@ -134,7 +138,7 @@
     from accelerate.utils.modeling import get_state_dict_from_offload
 
 if is_peft_available():
-    from .utils import find_adapter_config_file
+    pass
```
Cyrilvallez: Can be removed

src/transformers/modeling_utils.py
```diff
@@ -909,25 +859,36 @@ def update_key_name(keys):
...
     transformers_explicit_filename: Optional[str] = None,
 ) -> tuple[Optional[list[str]], Optional[dict]]:
     """Get all the checkpoint filenames based on `pretrained_model_name_or_path`, and optional metadata if the
     checkpoints are sharded.
     This function will download the data if necessary.
     """
+    cache_dir = download_kwargs.get("cache_dir")
+    force_download = bool(download_kwargs.get("force_download", False))
+    proxies = download_kwargs.get("proxies")
+    local_files_only = bool(download_kwargs.get("local_files_only", False))
```
Cyrilvallez: Why do we need additional `bool` casting here? Are they None? If so, maybe we should update defaults upstream

src/transformers/modeling_utils.py
```diff
@@ -4475,32 +4458,32 @@ def from_pretrained(
...
+            _ = kwargs.pop(name, None)
 
         # For BC on torch_dtype argument
         if torch_dtype is not None:
-            logger.warning_once("`torch_dtype` is deprecated! Use `dtype` instead!")
-            # If both kwargs are provided, use `dtype`
             dtype = dtype if dtype is not None else torch_dtype
 
+        if is_offline_mode() and not local_files_only:
+            logger.info("Offline mode: forcing local_files_only=True")
```
Cyrilvallez: Let's remove this logger.info as well no?

src/transformers/modeling_utils.py
```diff
@@ -4475,32 +4458,32 @@ def from_pretrained(
...
-        _ = kwargs.pop("from_tf", None)
-        _ = kwargs.pop("from_flax", None)
-        _ = kwargs.pop("offload_state_dict", None)
+        for name in ["mirror", "_fast_init", "low_cpu_mem_usage", "from_tf", "from_flax", "offload_state_dict"]:
+            _ = kwargs.pop(name, None)
 
         # For BC on torch_dtype argument
         if torch_dtype is not None:
-            logger.warning_once("`torch_dtype` is deprecated! Use `dtype` instead!")
-            # If both kwargs are provided, use `dtype`
```
Cyrilvallez, line 4480: Let's keep until we remove support no?

src/transformers/modeling_utils.py
```diff
@@ -4847,109 +4647,36 @@ def _assign_original_dtype(module):
...
-            repo_loading_kwargs = {
-                "cache_dir": cache_dir,
-                "force_download": force_download,
-                "proxies": proxies,
-                "local_files_only": local_files_only,
-                "token": token,
-                "revision": revision,
-                "subfolder": subfolder,
+        adjust_generation_fn = getattr(model, "adjust_generation_function", None)
+        if callable(adjust_generation_fn):
```
Cyrilvallez: Would still check `model.can_generate()` here, as I find this check a bit weird (also see previous comment about the check on `GenerationMixin` class)

src/transformers/modeling_utils.py
```diff
@@ -4509,240 +4492,90 @@ def from_pretrained(
...
-        # Just a helpful message in case we try to load safetensors files coming from old Transformers tf/flax classes
-        if is_from_file and checkpoint_files[0].endswith(".safetensors"):
-            with safe_open(checkpoint_files[0], framework="pt") as f:
-                metadata = f.metadata()
-            if metadata is not None and metadata.get("format") in ["tf", "flax"]:
-                logger.warning(
-                    "The safetensors checkpoint found has format `tf` or `flax`. This mean that the keys will very"
-                    "likely not match to the model you are trying to load, and will be newly initialized. If it's the case "
-                    "another warning will be raised later. Consider converting your checkpoint to the correct format."
-                )
```
Cyrilvallez, line 4718: I thought we could keep that for the first few releases of v5, but no strong opinions

src/transformers/modeling_utils.py
```diff
@@ -5177,45 +4904,18 @@ def _load_pretrained_model(
         # This offload index if for params explicitly on the "disk" in the device_map
         disk_offload_index = None
         disk_only_shard_files = []
-        # Prepare parameters offloading if needed
+        # Prepare parameters offloading if needed TODO: split this in a separate function
```
Cyrilvallez: the TODO has been done haha

src/transformers/quantizers/base.py
```diff
@@ -222,7 +222,29 @@ def preprocess_model(self, model: "PreTrainedModel", **kwargs):
...
+                if isinstance(child, PreTrainedModel):
+                    child.config._pre_quantization_dtype = original_dtype
+                _assign_original_dtype(child)
+
+        config._pre_quantization_dtype = original_dtype
+        _assign_original_dtype(model)
+
+        # Torchao needs access to all metadata later
+        if model.quantization_config.quant_method == QuantizationMethod.TORCHAO:
+            model.set_metadata(checkpoint_files)
```
Cyrilvallez: Very weird to do that here IMO -> let's remove and override the method in torchao quantizer (just have to call super() and add this)

src/transformers/integrations/accelerate.py
```diff
@@ -194,3 +203,149 @@ def find_tied_parameters(model: "nn.Module", **kwargs):
                 tied_param_groups[param_name].append(tied_param_name)
 
     return [sorted([weight] + list(set(tied))) for weight, tied in tied_param_groups.items()]
+
+
+def auto_set_device_map(device_map):
```
Cyrilvallez: Not a huge fan of the function name - maybe `check_and_set_device_map` or something similar? For maximum clarity

src/transformers/integrations/peft.py
```diff
@@ -614,3 +620,56 @@ def old_delete_adapter(model, adapter_name, prefix=None):
         if len(self.peft_config) == 0:
             del self.peft_config
             self._hf_peft_config_loaded = False
+
+
+def maybe_load_adapters(
```
Cyrilvallez, line 625: Maybe `find_potential_adapter_file` instead?

src/transformers/integrations/peft.py
```diff
@@ -614,3 +620,56 @@ def old_delete_adapter(model, adapter_name, prefix=None):
...
+            _commit_hash=download_kwargs.get("commit_hash"),
+            **adapter_kwargs,
+        )
+
+    if _adapter_model_path is not None and os.path.isfile(_adapter_model_path):
+        with open(_adapter_model_path, "r", encoding="utf-8") as f:
+            _adapter_model_path = pretrained_model_name_or_path
+            pretrained_model_name_or_path = json.load(f)["base_model_name_or_path"]
+
+    return _adapter_model_path
```
Cyrilvallez: We need to return the `pretrained_model_name_or_path` as well here, as it can change apparently!!
```suggestion
            _adapter_model_path = pretrained_model_name_or_path
            pretrained_model_name_or_path = json.load(f)["base_model_name_or_path"]

    return _adapter_model_path, pretrained_model_name_or_path
```

src/transformers/integrations/tensor_parallel.py
```diff
@@ -38,59 +38,81 @@
...
+    if tp_size is not None and tp_plan is None:
+        raise ValueError("tp_plan has to be set when tp_size is passed.")
+    if tp_plan is not None and tp_plan != "auto":
+        # TODO: we can relax this check when we support taking tp_plan from a json file, for example.
+        raise ValueError(f"tp_plan supports 'auto' only for now but got {tp_plan}.")
+    if tp_plan is not None and device_map is not None:
+        raise ValueError("`tp_plan` and `device_map` are mutually exclusive. Choose either one for parallelization.")
+    if device_mesh is None:
+        if tp_plan is None:
+            return None, None, None
```
Cyrilvallez: This can never be reached

src/transformers/modeling_utils.py
```diff
@@ -4509,240 +4492,90 @@ def from_pretrained(
...
-                tp_size = torch.distributed.get_world_size()
-
-        if use_auth_token is not None:
-            warnings.warn(
-                "The `use_auth_token` argument is deprecated and will be removed in v5 of Transformers. Please use `token` instead.",
-                FutureWarning,
+        if tp_plan is not None:  # TP warnings, and setup
+            tp_plan, device_map, device_mesh, tp_size = initialize_tensor_parallelism(
+                tp_plan, tp_size=tp_size, device_mesh=device_mesh, device_map=device_map
             )
```
Cyrilvallez: We need to enter in both cases!

```suggestion
        if tp_plan is not None or tp_size is not None:  # TP warnings, and setup
            tp_plan, device_map, device_mesh, tp_size = initialize_tensor_parallelism(
                tp_plan, tp_size=tp_size, device_mesh=device_mesh, device_map=device_map
            )
```

src/transformers/integrations/tensor_parallel.py
```diff
@@ -38,59 +38,81 @@
...
-def initialize_tensor_parallelism(tp_plan, tp_size=None):
+def initialize_tensor_parallelism(tp_plan, tp_size=None, device_mesh=None, device_map=None):
     r"""
     Sets up the device mesh and initialized the backend for tensor parallelism.
     This function is called when the model is loaded and the TP plan is set to 'auto'.
     """
-    if tp_plan is None:
-        return None, None, None
+    if tp_size is not None and tp_plan is None:
+        raise ValueError("tp_plan has to be set when tp_size is passed.")
```
Cyrilvallez, line 47: This check cannot be reached currently, also see my comment where the function is called

Cyrilvallez (commented):
Nice, will allow us to have much better visibility over what's happening!

src/transformers/generation/utils.py
ArthurZucker: No! IDK how but some udop models have it false becuase it's `model.....` and not `self....`

src/transformers/modeling_utils.py
ArthurZucker: No I think we always keep support, just don't warn

src/transformers/modeling_utils.py
ArthurZucker: ok

src/transformers/quantizers/base.py
ArthurZucker: yeah this was a hard move

src/transformers/integrations/tensor_parallel.py
ArthurZucker: it can it can! it was there before

src/transformers/integrations/tensor_parallel.py
ArthurZucker: mmmm but the thing is I just moved around haha 
Will check tho

src/transformers/modeling_utils.py
ArthurZucker: I feels like just for a warning don't really need it imo!

BlackSamorez:
Can you really delete `_process_model_after_weight_loading`?
I'm getting `TypeError: Can't instantiate abstract class FPQuantHfQuantizer with abstract method _process_model_after_weight_loading` on fp_quant and AQLM tests now.

BlackSamorez:
Oh, fixed by  #41613, I see.

PR #41446, opened 2025-10-08
Enable non-streaming mode in `transformers serve`

LysandreJik:
Needs this to be merged first: https://github.com/huggingface/transformers/pull/41444

Tests and docs need to be added before undraft

### DIFF
--- a/src/transformers/commands/serving.py
+++ b/src/transformers/commands/serving.py
@@ -35,6 +35,8 @@ from typing import Optional, TypedDict, Union
 
 from huggingface_hub import model_info
 from huggingface_hub.constants import HF_HUB_OFFLINE
+from openai.types.chat.chat_completion import Choice
+from starlette.responses import StreamingResponse
 from tokenizers.decoders import DecodeStream
 
 import transformers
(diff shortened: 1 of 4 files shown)


LysandreJik:
Test need a bit of a refactor, which will land once we merge @Wauplin's PR here: https://github.com/huggingface/transformers/pull/41487

src/transformers/commands/serving.py
```diff
@@ -35,6 +35,8 @@
 
 from huggingface_hub import model_info
 from huggingface_hub.constants import HF_HUB_OFFLINE
+from openai.types.chat.chat_completion import Choice
+from starlette.responses import StreamingResponse
```
ArthurZucker: IDK what starlette is, prob a new deps?

src/transformers/commands/serving.py
```diff
@@ -1152,7 +1220,51 @@ def generate_with_cache(**kwargs):
             finally:
                 thread.join()
 
-        return stream_chat_completion(generation_streamer, request_id)
+        if req.get("stream"):
+
+            def sse(_generator):
```
ArthurZucker: not a very helpful func name

src/transformers/modeling_utils.py
```diff
@@ -2509,14 +2509,14 @@ def _check_and_adjust_attn_implementation(
             try:
                 load_and_register_attn_kernel(applicable_attn_implementation)
                 # log that we used kernel fallback if successful
-                if attn_implementation.startswith("flash_attention"):
+                if attn_implementation.startswith("flash_"):
```
ArthurZucker: we should check `"flash_" in attn_implementation"` because if we fallback / use a kernel, it starts with `kernel_community/....`

src/transformers/commands/serving.py
ArthurZucker: (also do we really need it?)

ArthurZucker (approved):
I am missing a lot of context, but let's make sure funcs are not nested if possible, and func names are helpful + document why we need themn

src/transformers/commands/serving.py
LysandreJik: I moved the import from starlette to the one from FastAPI -> the FastAPI one is a re-export of the starlette one, but it's more coherent from an import shielding perspective

src/transformers/commands/serving.py
LysandreJik: Correct, replaced it by map so that it's cleaner

src/transformers/modeling_utils.py
LysandreJik: Good idea!

LysandreJik:
I cleaned it up a bit @ArthurZucker; IMO we might want to take a look at simplifying the stream/non-stream path in the future, depending on how the tool handling is done (still needs to be implemented for non-streaming generate and everything CB).

ArthurZucker (approved):
thanks!

PR #39956, opened 2025-08-06
Harmonize `past_key_value` to `past_key_valueS` everywhere

Cyrilvallez:
As per the title. I'm getting annoyed to see both, it was time to finally make it coherent everywhere. 
All the changes are made, so I'll only need to remove the decorators after next release (it's only internal modules, so no need for a long deprecation cycle)

Also reapplied modular to `examples/modular-transformer`

PR #41448, opened 2025-10-08
🚨 [v5] Rename left traces of `past_key_value` in BERT-like models

zucchini-nlp:
# What does this PR do?

As per title, finalizes the work from https://github.com/huggingface/transformers/pull/41425 and renames `past_key_value` as `past_key_values` in all files without deprecation cycle

It was a find and replace, so tests might fail 😅

### DIFF
--- a/src/transformers/models/autoformer/modeling_autoformer.py
+++ b/src/transformers/models/autoformer/modeling_autoformer.py
@@ -470,17 +470,17 @@ class AutoformerAttention(nn.Module):
                 is_updated = past_key_values.is_updated.get(self.layer_idx)
                 if is_cross_attention:
                     # after the first generated id, we can subsequently re-use all key/value_layer from cache
-                    curr_past_key_value = past_key_values.cross_attention_cache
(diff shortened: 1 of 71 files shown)


vasqu (approved):
Thanks a lot, all these big PRs made a mess :o lgtm

Just to be sure, can you check with run slow for the important models (bert, bart, gpt2, t5, roberta)

zucchini-nlp:
Ran slow tests on Bert, BART, Ernie, GPT2, T5 and MusicGen comparing to main branch, all green ✅

Cyrilvallez:
Not sure how they got back here because I had definitely made the switch in https://github.com/huggingface/transformers/pull/39956 already... We need to be very careful when touching a lot of files otherwise we shoot ourselves in the foot!

zucchini-nlp:
Yep, might be simply issues with rebasing main. Happens a lot when everyone is making huge PRs 😄

vasqu:
Yup, sorry that's likely on me with the vLLM refactor :cry: wasn't aware of that rename

PR #41449, opened 2025-10-08
Fix trainer simple tests

SunMarc:
# What does this PR do?

This PR should all simple trainer tests and some deepspeed tests. 
The only remaining tests to fix are deepspeed z2 grad acc tests but this is strange why it is failing ... cc @IlyasMoutawwakil maybe you have an idea as you worked on it for `HPU`


tests/trainer/test_trainer.py
```diff
@@ -4236,124 +4247,29 @@ def test_fp16_full_eval(self):
...
 
     @require_torch_gpu
-    @require_torch_non_multi_gpu
-    @require_torch_tensorrt_fx
-    def test_torchdynamo_full_eval(self):
-        from torch import _dynamo as torchdynamo
-
-        # torchdynamo at the moment doesn't support DP/DDP, therefore require a single gpu
-        n_gpus = get_gpu_count()
+    @pytest.mark.torch_compile_test
```
SunMarc, line 4146: move to compile instead

tests/trainer/test_trainer_seq2seq.py
```diff
@@ -38,8 +38,8 @@ def test_finetune_bert2bert(self):
         tokenizer = BertTokenizer.from_pretrained("google-bert/bert-base-uncased")
 
         bert2bert.config.vocab_size = bert2bert.config.encoder.vocab_size
-        bert2bert.config.eos_token_id = tokenizer.sep_token_id
-        bert2bert.config.decoder_start_token_id = tokenizer.cls_token_id
+        tokenizer.eos_token_id = tokenizer.sep_token_id
+        bert2bert.generation_config.decoder_start_token_id = tokenizer.cls_token_id
```
SunMarc, line 42: Is it the right fix @gante  ? I feel like we shouldn't udpate the config anymore and the safest is to update the generation_config for some special token and the tokenizer.

gante: for the moment it is 

(we should add some way to synchronize these attributes 👀 )

SunMarc:
hey @ydshieh is there a simple way to run trainer tests like the other tests ?

src/transformers/modeling_utils.py
```diff
@@ -5602,7 +5602,12 @@ def set_is_initialized_for_modules(module):
             import deepspeed
 
             not_initialized_parameters = list(
-                {v for v in self.state_dict().values() if not getattr(v, "_is_hf_initialized", False)}
+                {
+                    p
+                    for module in self.modules()
+                    if not getattr(module, "_is_hf_initialized", False)
+                    for p in module.parameters(recurse=False)
+                }
```
Cyrilvallez: I don't get why we have this change??

SunMarc: When launching this with deepspeed, the prior code returned empty tensors and this was causing issues as you can expect. So I checked what changed since then and I reverted the changes basically. Maybe deepspeed is messing up with the `state_dict` of the model.

Cyrilvallez: What I don't understand is that it's really the same thing... Like in the end we grab the same parameters, except the new way of doing can add already initialized as well, if a module has half the params initialized...

Cyrilvallez: I pushed another change and reverted this one, I believe it should fix it (wrong check on non-persistent buffers)

Cyrilvallez (approved):
LGTM, thanks a lot for digging deeper and uncovering the `keep_vars` issue!

Cyrilvallez:
Merging as the failing processor test is known and unrelated!

ydshieh:
> hey @ydshieh is there a simple way to run trainer tests like the other tests ?

Do you mean

run-slow: trainer

? I think it would work

PR #41450, opened 2025-10-08
Allow modifications of attributes in modular

molbap:
# What does this PR do?

Often, attributes are not derived from model names. Modular files based on a given model are then bloated because there is a tiny change in an attribute, needing the entire thing to be rewritten. See https://github.com/huggingface/transformers/pull/41224/files here for instance.

This introduces an attribute renamer, from an arbitrary attribute name to another.

```python
  class DINOv3ViTForImageClassification(Dinov2ForImageClassification):
      _attribute_renames = {"dinov2": "dinov3"}
```

On the attached PR, this expands the modeling file correctly.

### DIFF
--- a/utils/modular_model_converter.py
+++ b/utils/modular_model_converter.py
@@ -978,10 +978,14 @@ def replace_class_node(
             if m.matches(value_node, m.Dict()):
                 for element in value_node.elements:
                     if m.matches(element, m.DictElement()):
-                        old_key = element.key.evaluated_value if hasattr(element.key, 'evaluated_value') else None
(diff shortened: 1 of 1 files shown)


molbap:
After internal discussion: this would harm the overall modular philosophy, adding a mutable inside classes like this. For the moment, we accept rewriting modular files which differ just by an attribute naming. (cc @yonigozlan )

Instead, a util to convert on load with 0 cost would be preferrable, and is in the works. cc @Cyrilvallez @ArthurZucker

PR #41464, opened 2025-10-09
Fix auto model configuration for encoder of perceptionlm

fschlatt:
# What does this PR do?

Fixes #41387

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker @Cyrilvallez

### DIFF
--- a/src/transformers/models/auto/configuration_auto.py
+++ b/src/transformers/models/auto/configuration_auto.py
@@ -307,7 +307,6 @@ CONFIG_MAPPING_NAMES = OrderedDict[str, str](
         ("pegasus", "PegasusConfig"),
         ("pegasus_x", "PegasusXConfig"),
         ("perceiver", "PerceiverConfig"),
-        ("perception_encoder", "TimmWrapperConfig"),
         ("perception_lm", "PerceptionLMConfig"),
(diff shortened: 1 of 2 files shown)


eustlb (commented):
Looks like `perception_encoder` was first introduced in the PR that introduced `PerceptionLM` but later removed to use the timm model directly. IMO it should just simply be removed, but let me ping @zucchini-nlp here.

src/transformers/models/auto/configuration_auto.py
```diff
@@ -763,7 +763,7 @@
         ("pegasus", "Pegasus"),
         ("pegasus_x", "PEGASUS-X"),
         ("perceiver", "Perceiver"),
-        ("perception_encoder", "PerceptionEncoder"),
+        ("perception_encoder", "TimmWrapperConfig"),
```
zucchini-nlp: we can just delete it. When models re-use existing vision backbone, we dont usually add them as separate models in auto-map

fschlatt: Alright, I've deleted the entries. Thanks!

zucchini-nlp (approved):
thanks

PR #41466, opened 2025-10-09
Try to remove `pickle` - `BloomTokenizerFast`

ydshieh:
# What does this PR do?

I can see the effect as the value is changed as expected from

```
(Pdb) self.backend_tokenizer.pre_tokenizer
Sequence(pretokenizers=[Split(pattern=Regex(" ?[^(\s|[.,!?…。，、।۔،])]+"), behavior=Isolated, invert=False), ByteLevel(add_prefix_space=False, trim_offsets=True, use_regex=False)])
```
to
```
(Pdb) self.backend_tokenizer.pre_tokenizer
Sequence(pretokenizers=[Split(pattern=Regex(" ?[^(\s|[.,!?…。，、।۔،])]+"), behavior=Isolated, invert=False), ByteLevel(add_prefix_space=True, trim_offsets=True, use_regex=False)])
```

### DIFF
--- a/src/transformers/models/bloom/tokenization_bloom_fast.py
+++ b/src/transformers/models/bloom/tokenization_bloom_fast.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 """Tokenization classes for Bloom."""
 
-import pickle
 from typing import Optional
 
 from ...tokenization_utils_base import BatchEncoding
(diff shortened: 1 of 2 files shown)


ydshieh:
There are some similar places like

> src/transformers/models/cohere/tokenization_cohere_fast.py

But would like to have a confirmation first.

ArthurZucker (approved):
Yes and long due, this was not possible before, it is now because of the update to `tokenizers` ! Can be applied everywhere else!

ydshieh:
run-slow: bloom, cohere

ydshieh:
run-slow: bloom, cohere

PR #41468, opened 2025-10-09
Remove KERAS_NLP_IMPORT_ERROR

cyyever:
# What does this PR do?

Because `KERAS_NLP_IMPORT_ERROR` is not used.


Rocketknight1 (approved):
Yes, good catch, thank you!

PR #41470, opened 2025-10-09
[kernels] Cleanup deta kernel

MekkCyber:
# What does this PR do?

Cleanup the `deta` kernel since it's the same as `deformable_detr` kernel cleaned here : https://github.com/huggingface/transformers/pull/36853
And do the necessary modeling changes (even though the model is deprecated)


ArthurZucker (approved):
thanks 🧼

PR #41474, opened 2025-10-09
🚨 [v5] Toggle the serialization format in processors

### LINKED ISSUE #40447: Processor does not load optional attributes
### System Info

- `transformers` version: 4.55.0.dev0
- Platform: macOS-15.6.1-arm64-arm-64bit
- Python version: 3.12.2
- Safetensors version: 0.5.3
- Accelerate version: 1.9.0
- PyTorch version (accelerator?): 2.7.1 (NA)
- Jax version: not installed

### Who can help?

maybe @Rocketknight1 

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

Minimal reproduction:
if `image_processor` is in `attributes` it works, but if it is in `optional_attributes` it fails.

```python
import tempfile

from transformers import AutoImageProcessor, ProcessorMixin


class TextImageProcessor(ProcessorMixin):
    name = "text-image-processor"

    attributes = []
    image_processor_class = "AutoImageProcessor"
    optional_attributes = ["image_processor"]

    def __init__(self,
                 image_processor: AutoImageProcessor = None):
        super().__init__(image_processor=image_processor)

        self.image_processor = image_processor

        # Not used in this processor, but necessary for "from_pretrained" compatibility
        self.chat_template = None
        self.audio_tokenizer = None


with tempfile.TemporaryDirectory() as temp_dir:
    image_processor = AutoImageProcessor.from_pretrained("WinKawaks/vit-tiny-patch16-224")
    processor = TextImageProcessor(image_processor=image_processor)
    processor.save_pretrained(save_directory=temp_dir, push_to_hub=False)
    new_processor = TextImageProcessor.from_pretrained(temp_dir)

    assert new_processor.image_processor is not None
```

### Expected behavior

Both attributes AND optional attributes should be loaded and initialized.
Right now it seems like it would only load optional attributes if they are called `audio_tokenizer`

zucchini-nlp:
# What does this PR do?

As per title, in v5 we will start saving in the new nested format. The loading is still supported though for old-format configs

Also, we don't need an `optional_attributes` field that does not do what it is called. Since it is only `chat template`, let's just manually pop and set it. For audio tokenizers, it'd be nice to refactor a bit, I think we could try to move it in `self.attributes`

Fixes https://github.com/huggingface/transformers/issues/40447 by deleting ambiguously named `optional_attributes`


src/transformers/models/auto/processing_auto.py
```diff
@@ -300,7 +300,15 @@ def from_pretrained(cls, pretrained_model_name_or_path, **kwargs):
...
             config_dict, _ = ProcessorMixin.get_processor_dict(pretrained_model_name_or_path, **kwargs)
-            processor_class = config_dict.get("processor_class", None)
+            processor_classes = [
+                value.get("processor_class") for value in config_dict.values() if isinstance(value, dict)
+            ]
+            processor_classes += [config_dict.get("processor_class")]
+            for processor_class in processor_classes:
+                if processor_class is not None:
+                    processor_class = processor_class
+                    break
```
Cyrilvallez: Let's just iterate over another name or something instead of having a blank assignment and keeping the name we just iterated over for later

src/transformers/models/auto/processing_auto.py
```diff
@@ -300,7 +300,15 @@ def from_pretrained(cls, pretrained_model_name_or_path, **kwargs):
         processor_config_file = cached_file(pretrained_model_name_or_path, PROCESSOR_NAME, **cached_file_kwargs)
         if processor_config_file is not None:
             config_dict, _ = ProcessorMixin.get_processor_dict(pretrained_model_name_or_path, **kwargs)
-            processor_class = config_dict.get("processor_class", None)
+            processor_classes = [
+                value.get("processor_class") for value in config_dict.values() if isinstance(value, dict)
+            ]
```
Cyrilvallez: Why would the processor_class be set in an inner dict here?

src/transformers/processing_utils.py
```diff
@@ -652,7 +648,7 @@ def check_argument_for_proper_class(self, argument_name, argument):
 
         return proper_class
 
-    def to_dict(self, legacy_serialization=True) -> dict[str, Any]:
+    def to_dict(self, legacy_serialization=False) -> dict[str, Any]:
```
Cyrilvallez: IMO it does not make sense to keep legacy serialization option, if we can load both anyway -> let's force the usage of the new and better format by removing old one

src/transformers/processing_utils.py
```diff
@@ -693,6 +701,15 @@ def cast_array_to_list(dictionary):
...
             return dictionary
 
+        # Special case, add `audio_tokenizer` dict which points to model weights and path
+        if not legacy_serialization and "audio_tokenizer" in output:
+            audio_tokenizer_dict = {
+                "audio_tokenizer_class": self.audio_tokenizer.__class__.__name__,
+                "audio_tokenizer_name_or_path": self.audio_tokenizer.name_or_path,
+            }
+            # Update or overwrite, what do audio tokenizers expect when loading?
+            output["audio_tokenizer"] = audio_tokenizer_dict
```
Cyrilvallez: Don't really get the comment here haha, weird to have a question as a comment

src/transformers/processing_utils.py
```diff
@@ -747,7 +754,7 @@ def __repr__(self):
         attributes_repr = "\n".join(attributes_repr)
         return f"{self.__class__.__name__}:\n{attributes_repr}\n\n{self.to_json_string()}"
 
-    def save_pretrained(self, save_directory, push_to_hub: bool = False, legacy_serialization: bool = True, **kwargs):
+    def save_pretrained(self, save_directory, push_to_hub: bool = False, legacy_serialization: bool = False, **kwargs):
```
Cyrilvallez: Same in all the functions, let's remove legacy format altogether

Cyrilvallez (commented):
Nice! Let's remove the option to save in legacy format though, it does not really make sense to have it anymore as we want to save with better format starting with v5!

src/transformers/processing_utils.py
zucchini-nlp: forgot to delete. I was trying to move audio tokenizer to attributes, but it opened a can of worms

src/transformers/models/auto/processing_auto.py
zucchini-nlp: was kept for legacy. I removed it in last commit after deleting legacy flag completely.

zucchini-nlp:
Done, completely removed all legacy options for saving and in tests. CI is failing for unrelated reasons

src/transformers/processing_utils.py
```diff
@@ -832,9 +814,7 @@ def save_pretrained(self, save_directory, push_to_hub: bool = False, legacy_seri
         # plus we save chat_template in its own file
         output_processor_file = os.path.join(save_directory, PROCESSOR_NAME)
         output_chat_template_file_jinja = os.path.join(save_directory, CHAT_TEMPLATE_FILE)
-        output_chat_template_file_legacy = os.path.join(
-            save_directory, LEGACY_PROCESSOR_CHAT_TEMPLATE_FILE
-        )  # Legacy filename
+        output_chat_template_file_legacy = os.path.join(save_directory, LEGACY_PROCESSOR_CHAT_TEMPLATE_FILE)
```
Cyrilvallez, line 829: Do we need this one as well? I see we already save the jinja one

tests/test_processing_common.py
```diff
@@ -218,36 +218,6 @@ def test_processor_from_and_save_pretrained(self):
                     if "tokenizer" not in attribute:
                         self.assertEqual(repr(attribute_first), repr(attribute_second))
 
-    def test_processor_from_and_save_pretrained_as_nested_dict(self):
-        processor_first = self.get_processor()
```
Cyrilvallez, line 222: We should keep this one but just rename and change it a bit no? Or do we have another test checking that basic save/load work as expected?

Cyrilvallez (approved):
Nice, thanks a lot for removing all the legacy! Very happy to start saving with this new format!

src/transformers/processing_utils.py
zucchini-nlp: This is for @Rocketknight1 to decide if we can delete that totally. I think we're converging to one format for templates, though not sure when the legacy can be deleted

tests/test_processing_common.py
zucchini-nlp: This one works only when the feature extractor is saved as separate file. We can keep it to check BC with old-format files (if there're any in the hub)

I'll revert and save manually as legacy format

zucchini-nlp: Ah wait, you mean "this" test. Mostly it is already tested when we run common processing tests, unless the test is written in non-standard format. I will just revert and keep it, it's a fast test and does not hurt

src/transformers/processing_utils.py
Rocketknight1: I think legacy **saving** can be dropped in `v5` but legacy **loading** needs to stay, because a lot of checkpoints are still in the old format

PR #41476, opened 2025-10-09
Pickle - part 2

ydshieh:
# What does this PR do?

Require 
```
    if not strtobool(os.environ.get("TRUST_REMOTE_CODE", "False")):
        raise ValueError(
            "This part uses `pickle.load` which is insecure and will execute arbitrary code that is potentially "
            "malicious. It's recommended to never unpickle data that could have come from an untrusted source, or "
            "that could have been tampered with. If you already verified the pickle data and decided to use it, "
            "you can set the environment variable `TRUST_REMOTE_CODE` to `True` to allow it."
        )
```

I limit this PR to conversion scripts only. For other usage of pickle in our codebase, I will try to see if there is any alternative.


LysandreJik (approved):
Awesome, thank you!

PR #41477, opened 2025-10-09
Cleaning hub kernels 

MekkCyber:
# What does this PR do?

Cleaning some comments, and renaming all instances of `load_and_register_kernel` to `load_and_register_attn_kernel` since it's used only for attention kernels


ArthurZucker (commented):
Thanks

ArthurZucker (approved):

PR #41479, opened 2025-10-09
Remove SigOpt

SunMarc:
# What does this PR do?

This PR removes SigOpt integration as the library is not maintained anymore by intel. Also, users who used this library had to pass a API key, so since the service is down, the API won't work at all. I disabled the tests a few months ago since they started failing and now it is better to just remove the integration.


ydshieh (approved):
Goooo!

PR #41486, opened 2025-10-09
[`CI`] Fix copies on main

vasqu:
As per title


zucchini-nlp (approved):
Thanks, might be caused by me hehe

PR #41500, opened 2025-10-10
[QoL] modular conversion shows LoC saved

molbap:
# What does this PR do?

Tiny thing. Simply print the LoC saved when doing modular conversion.

image

## Why?

Because we sometimes hesitate too much (me first). Modular isn't magic, it's close,and it can't do everything. I sometimes want to optimize too much a conversion that's already leagues better than if I was using `modeling` directly.


ArthurZucker (approved):
💯

PR #41504, opened 2025-10-10
Revert `local_rank` deletion and some cleaning

SunMarc:
# What does this PR do?

This PR removes some bits that I forgot when removing `logging_dir`. Also, we need to keep `local_rank` as torch.distributed.launch inject `local_rank` in the script. I will deprecate at once torch removes it from their codebase + we don't support this version of pytorch which is in a super long time


MekkCyber (approved):
Nice ! more cleaning 🧼

PR #41505, opened 2025-10-10
🚨 [v5] `generate` delegates default cache initialization to the model

gante:
# What does this PR do?

See PR title. 

Now that all traces of legacy caches were removed, we can trust the model to initialize its own cache! This means we no longer need to set `cache_implementation="xxx"` defaults in new models, assuming the model's forward pass defaults to the right cache class.

Also fixes related bugs, uncovered by not feeding a cache to the model.


src/transformers/generation/utils.py
```diff
@@ -1999,17 +2001,15 @@ def _prepare_cache_for_generation(
...
-        # initialization should happen inside the model at prefill time.
-        else:
-            model_kwargs[cache_name] = DynamicCache(**dynamic_cache_kwargs)
-
         # TODO (joao): this logic is incomplete, e.g. `offloaded` should apply to both caches. Refactor this function
         # to correctly pass parameterization to both caches.
-        if requires_cross_attention_cache and not isinstance(model_kwargs[cache_name], EncoderDecoderCache):
-            model_kwargs[cache_name] = EncoderDecoderCache(
-                model_kwargs[cache_name],  # self-attention cache
+        if (
```
gante, line 2010: update to this branch -- we only want to convert the cache to `EncoderDecoderCache` here if:
1. the user has set custom cache args in `generate`
2. the model is encoder-decoder
3. (implicitly, all encoder-decoder models have `past_key_values` as their cache name)

In all other cases, we delegate cache init to the model itself

src/transformers/generation/utils.py
```diff
@@ -546,8 +546,10 @@ def prepare_inputs_for_generation(
         model_inputs["cache_position"] = cache_position
 
         # 2. Generic cache-dependent input preparation
```
gante, line 548: Changes in new L549-594 are related to recurrent_gemma: prior to the deletion of the old L2002-2007, generate was preparing a DynamicCache for recurrent_gemma. This cache was never used in forward, but it was inducing the correct behavior in prepare_inputs_for_generation (cache_position-based input slicing)

With the new logic, use_cache=True implies cache_position-based input slicing, even if the model is not using a standard cache.

src/transformers/models/bart/modeling_bart.py
```diff
@@ -802,7 +802,7 @@ def forward(
         if use_cache and past_key_values is None:
             past_key_values = (
                 EncoderDecoderCache(DynamicCache(config=self.config), DynamicCache(config=self.config))
-                if encoder_hidden_states is not None
+                if encoder_hidden_states is not None or self.config.is_encoder_decoder
```
gante, line 805: On encoder-decoder models that we may want to use as decoder-only, we want `EncoderDecoderCache` in two possible situations:
1. [missing] the model is encoder-decoder
2. the model is not encoder-decoder, but `encoder_hidden_states` passed (which means we will compute the cross-attention, and thus we should cache it)

tests/models/roformer/test_modeling_roformer.py
```diff
@@ -209,6 +210,8 @@ def create_and_check_for_generate_causal_lm(
         token_labels,
         choice_labels,
     ):
+        config = copy.deepcopy(config)
+        config.is_decoder = True
```
gante, line 214: `RoFormerForCausalLM` won't use the cache if `config.is_decoder!=True`, and this test tests cache usage 🙃

(A warning was being thrown)

src/transformers/generation/logits_process.py
```diff
@@ -2188,6 +2191,10 @@ def set_model(self, model):
         self.model = model
 
     def set_inputs(self, inputs):
+        # build `cache_position` on the fly
+        seq_length = inputs["input_ids"].shape[1]
+        inputs = self.model._get_initial_cache_position(seq_length, self.model.device, inputs)
+        # prepare other inputs
```
gante, line 2197: whisper has custom generation structure that doesn't follow our code patterns -> one of the issues is that it has stateful `LogitsProcessor` -> this state-related function uses `prepare_inputs_for_generation` out of the usual order -> changes related to this PR exposed that it was missing `cache_position` as an input here

(if we have bandwidth, we should revisit whisper generate to streamline its code)

src/transformers/generation/utils.py
```diff
@@ -1805,7 +1807,7 @@ def _get_initial_cache_position(self, seq_length, device, model_kwargs):
             cache_position = torch.ones(seq_length, dtype=torch.int64, device=device).cumsum(0) - 1
 
         past_length = 0
-        if model_kwargs.get("past_key_values") is not None:
+        if model_kwargs.get("past_key_values", None) is not None:
```
Cyrilvallez: Technically not needed, `get` has None as a default (and I thought ruff was now enforcing this one but apparently not 🤔 too many changes to the ruff rules recently)

src/transformers/models/fsmt/modeling_fsmt.py
```diff
@@ -616,10 +616,6 @@ def forward(
         else:
             raise ValueError("You have to specify either decoder_input_ids or decoder_inputs_embeds")
 
-        # initialize `past_key_values`
-        if use_cache and past_key_values is None:
-            past_key_values = EncoderDecoderCache(DynamicCache(config=self.config), DynamicCache(config=self.config))
-
```
Cyrilvallez, line 622: EncoderDecoderCache initialized in a Decoder only module... 🥲🥲🥲 good catch!

Cyrilvallez (approved):
LGTM, thanks!!

src/transformers/generation/logits_process.py
zucchini-nlp: > (if we have bandwidth, we should revisit whisper generate to streamline its code)


💯,  revisiting audio modality generation will be super helpful

src/transformers/generation/utils.py
```diff
@@ -546,8 +546,10 @@ def prepare_inputs_for_generation(
         model_inputs["cache_position"] = cache_position
 
         # 2. Generic cache-dependent input preparation
+        use_cache = kwargs.get("use_cache", False) or getattr(self.config, "use_cache", False)
```
zucchini-nlp: this will result in `True` even when the user-kwargs set caching as `False`. IIUC we just want to give priority to user-defined cache and not assume that caching is used whenever it is set to `True` in any of the places

src/transformers/generation/utils.py
```diff
@@ -546,8 +546,10 @@ def prepare_inputs_for_generation(
...
         if past_key_values is not None:
             model_inputs["past_key_values"] = past_key_values
+        if past_key_values is None or use_cache:
```
zucchini-nlp, line 556: maybe i am missing smth, do we apply cache slicing when the `past_key_values is None`? Looks not intuitive from first sight, so let's add a comment explaining why

zucchini-nlp (commented):
Thanks, left one question that seems to be a typo or just my bad understading

src/transformers/generation/utils.py
gante: @zucchini-nlp that's a good catch, user-defined `generate` kwargs >> config values! 

Will update accordingly

src/transformers/generation/utils.py
gante: stateful models like `recurrent_gemma` assume that slicing happens, but don't have a `Cache` cache -- will add a comment :)

zucchini-nlp: so they don't have a `Cache` object and also `if use_cache` does not catch those cases?

PR #41506, opened 2025-10-10
[SAM] Fix typing hints 

zucchini-nlp:
# What does this PR do?

As per title

### DIFF
--- a/src/transformers/models/sam/processing_sam.py
+++ b/src/transformers/models/sam/processing_sam.py
@@ -30,13 +30,15 @@ from ...utils import is_torch_available
 if is_torch_available():
     import torch
 
+NestedList = list[Union[Optional[float | int], "NestedList"]]
+
 
 class SamImagesKwargs(ImagesKwargs, total=False):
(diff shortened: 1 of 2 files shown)


src/transformers/models/sam/processing_sam.py
```diff
@@ -30,13 +30,15 @@
 if is_torch_available():
     import torch
 
+NestedList = list[Union[Optional[float | int], "NestedList"]]
```
molbap, line 33: seen it in Kosmos2, worth moving to a common util maybe?

zucchini-nlp: ideally we should just use `Sequence[int | float]` but hub validator does not yet support it. I think we just leave it here, to have more reasons to change it later

molbap: yeah that's fair!

molbap (approved):
looks good to me

zucchini-nlp:
Thanks

PR #41510, opened 2025-10-10
Fix detectron2 import

Cyrilvallez:
# What does this PR do?

Detectron2 does not uninstall correctly for some reason (probably the setup as the lib needs to be installed from source) so we need this cc @ydshieh


ydshieh:
Thanks!

PR #41522, opened 2025-10-11
Fix _init_weights to safely skip int8 quantized weights

KaparthyReddy:
This PR addresses a RuntimeError that occurs when loading W8A8 quantized Qwen2.5-VL models. 

Changes:
- Modified `_init_weights` in `Qwen2_5_VLPreTrainedModel` to safely skip non-floating tensors (e.g., int8 quantized weights) during initialization.
- Ensures float tensors are initialized normally, while int8 or other non-float tensors are ignored, preventing errors from `normal_()` on integer dtypes.
- Improves compatibility for loading quantized models without vLLM and allows safe custom hooks for research purposes.

Note: This PR does not include the tester file `test_init_weights_safe.py`; that file exists only in the fork for private testing.


Rocketknight1:
Can you link the issue this is fixing..?

PR #41524, opened 2025-10-11
Add max_eval_batches argument to TrainingArguments

### LINKED ISSUE #31561: Add argument to set number of eval steps in Trainer
### Feature request

I would like to add an argument to the [Trainer](https://huggingface.co/docs/transformers/main/en/main_classes/trainer#transformers.Trainer) class that allows for setting the number of eval steps (batches) for the evaluation procedure during training. 

The current behavior is to fully iterate through the entire dataset provided by the `eval_dataset` argument. However, when using a very large evaluation dataset, the evaluation process can take a long time. Especially for debugging, it would be helpful to be able to explicitly specify how many batches you would like to use as part of the evaluation procedure. 

### Motivation

Currently, I am using dataset streaming to train and evaluate models (motivated by large dataset size). My `eval_strategy` argument is set to `"steps"` since I'm using the streaming dataset, and so the Trainer runs the evaluation procedure every `eval_steps` number of steps. However, because there is no way to control how many steps are run during the evaluation procedure, the Trainer iterates through the entire eval dataset, which can be very time-consuming for a large dataset. Ideally, we could specify how many evaluation steps to run inside the evaluation procedure. 

Note that the argument I would like to add is different than the current `eval_steps` argument -- the current `eval_steps` argument specifies "Number of update steps _between_ two evaluations" but what I would like to specify is the number of steps _within_ the evaluations. 

### Your contribution

I would be happy to contribute code for this if this is deemed a relevant feature request and there isn't existing functionality that I'm unaware of.

KaparthyReddy:
# Add max_eval_batches argument to TrainingArguments

## Description
Adds a `max_eval_batches` parameter to `TrainingArguments` that allows users to limit the number of batches used during evaluation.

Fixes #31561

## Motivation
When working with large evaluation datasets, running evaluation on the entire dataset can be very slow. During development, hyperparameter tuning, or quick iteration, it's often sufficient to evaluate on a subset of the data.

This is similar to PyTorch Lightning's `limit_val_batches` parameter.

## Changes
- ✅ Added `max_eval_batches` parameter to `TrainingArguments`
- ✅ Implemented batch limiting in `Trainer.evaluation_loop`
- ✅ Added test coverage
- ✅ Added documentation in parameter metadata

## Usage Example
```python
from transformers import Trainer, TrainingArguments

training_args = TrainingArguments(
    output_dir="./output",
    evaluation_strategy="steps",
    eval_steps=100,
    max_eval_batches=50,  # Only evaluate on 50 batches
)

trainer = Trainer(
    model=model,
    args=training_args,
    eval_dataset=eval_dataset,
)

# Evaluation will stop after 50 batches instead of going through entire dataset
trainer.evaluate()

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -4448,6 +4448,8 @@ class Trainer:
 
         # Main evaluation loop
         for step, inputs in enumerate(dataloader):
+            if self.args.max_eval_batches is not None and step >= self.args.max_eval_batches:
+                break
             # Update the observed num examples
             observed_batch_size = find_batch_size(inputs)
             if observed_batch_size is not None:
--- a/src/transformers/training_args.py
+++ b/src/transformers/training_args.py
@@ -1080,6 +1080,20 @@ class TrainingArguments:
             )
         },
     )
+
+    max_eval_batches: Optional[int] = field(
+        default=None,
+        metadata={
+            "help": (
+                "Maximum number of batches to use for evaluation. "
+                "If None, all batches in the eval dataset will be used. "
+                "Useful for speeding up evaluation on large datasets."
+            )
+        },
+    )
+
+
+    
     dataloader_num_workers: int = field(
         default=0,
         metadata={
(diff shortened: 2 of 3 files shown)


Rocketknight1:
cc @sunmarc, but we might have another way to handle this already!

PR #41529, opened 2025-10-12
Add num_logits_to_keep to GPT-2 and improve Llama implementation

KaparthyReddy:
# Add num_logits_to_keep to GPT-2 and improve Llama implementation

## Description
Adds `num_logits_to_keep` parameter to GPT-2 for memory-efficient generation and simplifies the implementation in Llama.

### GPT-2
- ✅ Added `num_logits_to_keep` parameter to `GPT2LMHeadModel.forward()`
- ✅ Only computes logits for last N tokens when specified
- ✅ Default behavior unchanged (computes all logits when `num_logits_to_keep=0`)
- ✅ Added parameter documentation

### Llama
- ✅ Simplified `num_logits_to_keep` implementation
- ✅ Removed redundant slicing operation
- ✅ Fixed edge case where `logits_to_keep=0` would incorrectly slice all tokens
- ✅ Cleaner, more maintainable code

## Motivation
During generation, models typically compute logits for ALL tokens in the sequence. However, for next-token prediction, we only need the last token's logits. This parameter allows:
- 🚀 **Memory efficiency**: Reduce memory usage during generation
- ⚡ **Speed**: Less computation needed
- 🎯 **Flexibility**: Users can choose how many logits to keep

## Usage Example
```python
from transformers import GPT2LMHeadModel, GPT2Tokenizer

model = GPT2LMHeadModel.from_pretrained("gpt2")
tokenizer = GPT2Tokenizer.from_pretrained("gpt2")

inputs = tokenizer("Hello, I'm a language model", return_tensors="pt")

# Default: compute all logits
outputs = model(**inputs)
print(outputs.logits.shape)  # [1, sequence_length, vocab_size]

# Memory efficient: only compute last token's logits
outputs = model(**inputs, num_logits_to_keep=1)
print(outputs.logits.shape)  # [1, 1, vocab_size]

### DIFF
--- a/src/transformers/models/gpt2/modeling_gpt2.py
+++ b/src/transformers/models/gpt2/modeling_gpt2.py
@@ -778,7 +778,7 @@ class GPT2LMHeadModel(GPT2PreTrainedModel, GenerationMixin):
         output_attentions: Optional[bool] = None,
         output_hidden_states: Optional[bool] = None,
         return_dict: Optional[bool] = None,
-        logits_to_keep: Union[int, torch.Tensor] = 0,
+        num_logits_to_keep: int = 0,
         **kwargs,
     ) -> Union[tuple, CausalLMOutputWithCrossAttentions]:
         r"""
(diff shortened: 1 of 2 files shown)


Rocketknight1:
`logits_to_keep` is preferred over `num_logits_to_keep`! `num_logits_to_keep` is mostly only seen in older models and it's something we're trying to deprecate and update over time.

PR #40799, opened 2025-09-10
[Trainer] Fix DP loss

SunMarc:
# What does this PR do?

This PR fixes a couple of issues related to DP: 
- hight fp16 norm_grad -> we need to prepare the optimizer, otherwise, we don't unscale the optimizer when doing gradient clipping + grad scaler
- Fix logic when using `num_items_in_batch`. For DP, when `num_items_per_batch is not None`, the value is by default the sum of the tokens across all devices. This is not the case for DDP for example. So, we need to be careful of that. When using `num_items_in_batch`, we weren't applying DP at all. 
- Fix `num_input_tokens_seen` as it was always logged

Fixes https://github.com/huggingface/transformers/pull/40610, https://github.com/huggingface/transformers/issues/37474 and https://github.com/huggingface/transformers/pull/38938

In any case, users should use DDP instead of DP. I might start adding some warning to prompt users to switch

Fixes the following tests: 
`test_gradient_accumulation`, `test_gradient_accumulation_loss_alignment_with_loss_func`, `test_gradient_accumulation_loss_alignment_with_model_loss`

# Reproducer 

Combination of `fp16`, `average_tokens_across_devices` should all give the same loss approximately

```python
from datasets import load_dataset

from transformers import (
    AutoConfig,
    AutoTokenizer,
    DataCollatorForLanguageModeling,
    GPTNeoXForCausalLM,
    TrainingArguments,
    Trainer
)


config = AutoConfig.from_pretrained('EleutherAI/pythia-14m')
model = GPTNeoXForCausalLM(config=config).to('cuda')
tokenizer = AutoTokenizer.from_pretrained('EleutherAI/pythia-14m')
tokenizer.pad_token = tokenizer.eos_token
train_data = load_dataset("wiwu2390/minipile-100k", split="train")

def tokenize_function(sample):
    return tokenizer(sample["text"], truncation=True, max_length=512)

tokenized_dataset = train_data.map(tokenize_function, batched=True, remove_columns=["text"])

data_collator = DataCollatorForLanguageModeling(
    tokenizer=tokenizer, mlm=False
)

training_args = TrainingArguments(
    output_dir="../data/pythia-14m-minipile-100k",
    num_train_epochs=3,
    per_device_train_batch_size=16,
    per_device_eval_batch_size=16,
    gradient_accumulation_steps=4,
    logging_steps=1,
    save_steps=100,
    learning_rate=1e-3,
    weight_decay=0.01,
    warmup_steps=100,
    fp16=True,
)

[...]

PR #41530, opened 2025-10-12
Fix: Correct loss normalization in training_step for multi-GPU training

### LINKED ISSUE #37474: Trainer.training_step incorrectly normalizes mean token loss when n_gpu > 1
### System Info

```
- `transformers` version: 4.46.0
- Platform: Linux-5.15.0-136-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- Huggingface_hub version: 0.29.2
- Safetensors version: 0.5.3
- Accelerate version: 1.4.0
- PyTorch version (GPU?): 2.4.1+cu121 (True)
- Jax version: not installed
- Using GPU in script?: yes
- GPU type: NVIDIA RTX A5000
```

### Who can help?

@zach-huggingface @SunMarc @ArthurZucker

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

Full example setup:

```
config = AutoConfig.from_pretrained('EleutherAI/pythia-14m')
model = GPTNeoXForCausalLM(config=config).to('cuda')
tokenizer = AutoTokenizer.from_pretrained('EleutherAI/pythia-14m')
tokenizer.pad_token = tokenizer.eos_token
train_data = load_dataset("wiwu2390/minipile-100k", split="train")

def tokenize_function(sample):
    return tokenizer(sample["text"], truncation=True, max_length=512)

tokenized_dataset = train_data.map(tokenize_function, batched=True, remove_columns=["text"])

data_collator = DataCollatorForLanguageModeling(
    tokenizer=tokenizer, mlm=False
)

training_args = TrainingArguments(
    output_dir="../data/pythia-14m-minipile-100k",
    num_train_epochs=3,
    per_device_train_batch_size=16,
    per_device_eval_batch_size=16,
    evaluation_strategy="no",
    logging_steps=1,
    save_steps=100,
    learning_rate=1e-3,
    weight_decay=0.01,
    warmup_steps=100,
    fp16=True,
)

trainer = Trainer(
    model=model,
    args=training_args,
    train_dataset=tokenized_dataset,
    tokenizer=tokenizer,
    data_collator=data_collator,
)

trainer.train()
```

With 4 GPUs, the training loss at step 1 is ~2.7. However, the expected value is ~10.8. Indeed, this is what we get if we set CUDA_VISIBLE_DEVICES=0.

### Expected behavior

Since the model is being trained from initialization, the training loss at the first few steps should be around ~log(vocab_size)=10.8. However, when using 4 GPUs, the reported loss is 1/4 of that (2.7).

The reason that this is happening is that the DataParallel-wrapped model gets `num_items_in_batch` as an input kwarg in `Trainer.compute_loss`; this is equal to the number of tokens in the batch (combined across all devices). Each device gets a 1/4-size per-device batch and returns the sum of token losses divided by `num_items_in_batch` (see `transformers.loss.loss_utils.fixed_cross_entropy`). The correct way to aggregate these per-device losses is then to *sum* them. However, `Trainer.training_step` takes the mean:
https://github.com/huggingface/transformers/blob/953196a43dae6a3c474165fba7d215fcbc7b7730/src/transformers/trainer.py#L3759

A quick and dirty fix would be:
```
if self.args.n_gpu > 1:
    loss = loss.mean() if num_items_in_batch is None else loss.sum()
```
I'm not sure if this is compatible with other workflows though.

KaparthyReddy:
# Fix: Correct loss normalization in training_step for multi-GPU training

## Description
Fixes #37474 

This PR corrects the loss aggregation logic in `Trainer.training_step` when training with multiple GPUs.

## Problem
When `num_items_in_batch` is provided (e.g., for token-level loss normalization), each device computes loss as:
```python
per_device_loss = sum_of_losses / total_items_across_all_devices

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -3834,8 +3834,14 @@ class Trainer:
             if self.args.optim in [OptimizerNames.LOMO, OptimizerNames.ADALOMO]:
                 kwargs["learning_rate"] = self._get_learning_rate()
 
+            # Multi-GPU loss aggregation:
+            # - If num_items_in_batch is provided, each device returns loss normalized by
+            #   total items across ALL devices, so we sum the per-device losses
+            # - Otherwise, each device returns unnormalized loss, so we average
             if self.args.n_gpu > 1:
-                loss = loss.mean()  # mean() to average on multi-gpu parallel training
+                loss = loss.sum() if num_items_in_batch is not None else loss.mean()
+            else:
+                loss = loss.mean() if loss.ndim > 0 else loss
 
             # Finally we need to normalize the loss for reporting if GA loss bug is not fixed during compute loss
             if (not self.model_accepts_loss_kwargs or num_items_in_batch is None) and self.compute_loss_func is None:


Rocketknight1:
This should already be fixed in https://github.com/huggingface/transformers/pull/40799

@KaparthyReddy can you cool it down with these PRs, or at least read issues carefully and only make PRs when you understand the issue and you've checked they're not duplicates? You've made like 6 of them and they're all obviously written by Copilot, so it's very hard to tell if any of them actually do anything useful!

KaparthyReddy:
Noted. I’ll proceed as I see fit.

PR #41534, opened 2025-10-12
Add VideoMAE video processor 

Aki-07:
## What does this PR do?

- add a dedicated `VideoMAEVideoProcessor` that decodes/samples videos via TorchCodec and emits `pixel_values` ready for VideoMAE models
- document the new processor alongside the existing image processor so users can discover the GPU-friendly path
- cover the processor with torchvision-gated regression tests to ensure serialization, sampling, and output naming stay stable

Fixes #41520 

## Before submitting
      Pull Request section?
      to it if that's the case. (follow-up to the discussion with @zucchini-nlp on video processors)

### DIFF
--- a/src/transformers/models/videomae/video_processing_videomae.py
+++ b/src/transformers/models/videomae/video_processing_videomae.py
@@ -1,0 +1,44 @@
+# coding=utf-8
+# Copyright 2025 The HuggingFace Inc. team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Video processor class for VideoMAE."""
+
+from ...image_utils import IMAGENET_STANDARD_MEAN, IMAGENET_STANDARD_STD, PILImageResampling
+from ...video_processing_utils import BaseVideoProcessor
+
+
+class VideoMAEVideoProcessor(BaseVideoProcessor):
+    resample = PILImageResampling.BILINEAR
+    image_mean = IMAGENET_STANDARD_MEAN
+    image_std = IMAGENET_STANDARD_STD
+    size = {"shortest_edge": 224}
+    default_to_square = False
+    crop_size = {"height": 224, "width": 224}
+    do_resize = True
+    do_center_crop = True
+    do_rescale = True
+    rescale_factor = 1 / 255
+    do_normalize = True
+    do_convert_rgb = True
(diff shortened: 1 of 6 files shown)


Aki-07:
@zucchini-nlp Requesting for review

src/transformers/models/videomae/video_processing_videomae.py
```diff
@@ -0,0 +1,44 @@
...
+    rescale_factor = 1 / 255
+    do_normalize = True
+    do_convert_rgb = True
+    do_sample_frames = False  # Set to False for backward compatibility with image processor workflows.
+    model_input_names = ["pixel_values"]
+
+    def preprocess(self, videos, **kwargs):
+        batch = super().preprocess(videos, **kwargs)
+        if "pixel_values_videos" in batch:
+            batch["pixel_values"] = batch.pop("pixel_values_videos")
```
zucchini-nlp: the `pixel_values_videos` should always be in batch. If it is not, then smth clearly went wrong. So let's slice always and let python raise errors when the key is not found

zucchini-nlp (approved):
Looks good to me, thanks a lot for adding it! Can you also verify that the pixels obtained with video processor are identical to the pixels obtained with `VideoMAEImageProcessor` class?

Then let's change the docstring here to use video processor for decoding and sampling: https://github.com/Aki-07/transformers/blob/b3fb8511bec7cada4b91a47ef521f056ce8a11ac/src/transformers/models/videomae/modeling_videomae.py#L776-L838

src/transformers/models/videomae/video_processing_videomae.py
Aki-07: sure fixed it!

Aki-07:
Thanks for the review, @zucchini-nlp!

• Verified parity between VideoMAEVideoProcessor (TorchCodec) and VideoMAEImageProcessor — the outputs are numerically very close (mean diff ≈ 0.0008, max ≈ 0.0078 < 1%), which matches expected rounding differences from YUV→RGB conversion.
• Added the relaxed-tolerance check in the regression test to reflect this.
• Updated the model docstring example in modeling_videomae.py to use VideoMAEVideoProcessor for decoding and sampling as suggested.

tests/models/videomae/test_video_processing_videomae.py
```diff
@@ -0,0 +1,165 @@
...
+
+        video_frames_np = video[0]
+        video_frames_pil = [Image.fromarray(frame.astype("uint8")) for frame in video_frames_np]
+        video_out = video_processor(video_frames_pil, return_tensors="pt")
+        image_out = image_processor(video_frames_pil, return_tensors="pt")
+
+        diff = (video_out["pixel_values"] - image_out["pixel_values"]).abs()
+        max_diff = diff.max().item()
+        mean_diff = diff.mean().item()
+        print(f"max: {max_diff:.6f}  mean: {mean_diff:.6f}")
```
zucchini-nlp: print 😄  lets delete these 4 lines

zucchini-nlp (approved):
amazing, thanks for adding it! Let's just delete code used for debugging and we can merge

tests/models/videomae/test_video_processing_videomae.py
Aki-07: Ah apologies ! 😅 . Fixed in that commit

Issue #41208, opened 2025-09-30
Integrate mamba SSM kernels from the hub


romitjain:
### Feature request

Currently, mamba kernels are imported via the main source package ex, for [GraniteMoeHybrid](https://github.com/huggingface/transformers/blob/main/src/transformers/models/granitemoehybrid/modeling_granitemoehybrid.py#L44-L46)

Can we migrate this to use the kernels-hub (`kernels-community/mamba-ssm`) variation instead?

### Motivation

Removes the external dependency. Kernel hub is also integrated at several other places throughout the library.

### Your contribution

I can submit a PR for migrating from the PyPi `mamba_ssm` package to the `kernels` package for mamba ops.

vasqu:
Hey @romitjain, I think that would be very cool. We should keep the option to have the og package as well tho. We had a similar PR for causal-conv1d over here #40765 

cc @MekkCyber to confirm if that's ok

romitjain:
Thanks for the reference @vasqu 
If you are good with the proposal, I will raise a PR.

PS: I can't seem to open the [causal conv1d](https://huggingface.co/kernels-community/causal-conv1d) kernel package on the hub. Is that expected?

vasqu:
Me neither xD might be set to private atm, let's wait for @MekkCyber to respond as he's more involved with kernels

MekkCyber:
Yup sorry it's public now. Yes it's very cool to have `mamba-ssm` as well used in transformers from the hub. We already have [kernels-community/mamba-ssm](https://huggingface.co/kernels-community/mamba-ssm), You can definitely try to make it work and ping me when it's done 🤗 . If you have any issues with the kernel feel free to open an issue on the hub.

romitjain:
Thanks @MekkCyber 
Let me implement it and raise a PR

romitjain:
The `mamba_ssm` kernel is not compatible with the latest transformers repo. I have raised a PR to fix it, can someone here please take a look or tag an appropriate member?

https://huggingface.co/kernels-community/mamba-ssm/discussions/3

(Maybe @MekkCyber)

Thank you

romitjain:
I have also raised a draft PR (https://github.com/huggingface/transformers/pull/41540), which is contingent on the hub PR being merged

MekkCyber:
Thanks @romitjain ! will look into it and rebuild the kernel

romitjain:
[...]

PR #41540, opened 2025-10-13
Added kernels from kernel hub for Bamba model

romitjain:
# What does this PR do?

Adds support for `mamba_ssm` and `causal_conv1d` kernels from the kernel-hub in bamba models.

Fixes # (issue)

https://github.com/huggingface/transformers/issues/41208

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@vasqu @MekkCyber @drbh


romitjain:
Downstream changes have not been done (eg: `modeling_granitemoehybrid.py`)
Since it was mentioned I can't update those files, not sure how to go about them.

MekkCyber:
Hi @romitjain I'm working on a pr to make the kernel function mapping easier to do so we don't have to use new functions like `lazy_load_mamba_ssm` in the modeling files. Once merged we can refactor your PR to make it work with the new API, and then we can apply the changes to all other models using `make fix-copies`

romitjain:
@MekkCyber Sure, no worries. Let me know (or you can share your PR here) once done, and I will udpate my PR

MekkCyber:
sure here is the pr : https://github.com/huggingface/transformers/pull/41577

romitjain:
@MekkCyber I believe I can refactor my PR by using your new mapping function now?

MekkCyber:
Yes, i did that for falcon models here : https://github.com/huggingface/transformers/pull/41664, you can do that for bamba models then using the same API, however the mamba-ssm kernel needs to be fixed now before merging the PRs.

romitjain:
@MekkCyber @vasqu 

The `mamba_ssm` kernels were breaking for me because it was not able to import `mamba_chunk_scan_combined, mamba_split_conv1d_scan_combined`. This was working earlier but has broken in the latest revision. The default revision uses the incorrect output class (https://github.com/state-spaces/mamba/pull/807).

For my local testing, I made local changes to the mamba_ssm repo and tested the flow. However, we would need to fix it on kernels hub for this PR to work end to end.

Other than that, structurally, this is ready for review. Can you please have a look?

src/transformers/integrations/hub_kernels.py
```diff
@@ -286,7 +288,7 @@ def lazy_load_kernel(kernel_name: str, mapping: dict[str, Optional[ModuleType]]
         if callable(is_kernel_available) and is_kernel_available():
             # Try to import the module "{kernel_name}" from parent package level
             try:
-                module = importlib.import_module(f"{kernel_name}")
+                module = importlib.import_module(f"{new_kernel_name}")
```
romitjain, line 373: This is required to allow the function to fallback to the pypi package

src/transformers/integrations/hub_kernels.py
```diff
@@ -185,6 +185,7 @@ def register_kernel_mapping(*args, **kwargs):
 
 _HUB_KERNEL_MAPPING: dict[str, dict[str, str]] = {
     "causal-conv1d": {"repo_id": "kernels-community/causal-conv1d"},
+    "mamba-ssm": {"repo_id": "kernels-community/mamba-ssm", "revision": "clean-mamba-ssm"},
```
romitjain: Copied over from: https://github.com/huggingface/transformers/pull/41664

src/transformers/utils/import_utils.py
```diff
@@ -1176,6 +1176,11 @@ def is_mistral_common_available() -> bool:
     return _is_package_available("mistral_common")
 
 
+@lru_cache
+def is_einops_available() -> bool:
+    return _is_package_available("einops")
```
vasqu: Shouldnt be needed?

vasqu (commented):
I don't have many comments except we should align with #41664 on how to lazy load.

These kernels are essentially the same so we ought to standardize it. The mamba-ssm side should be fixed on their repo, thx for the PR there!

cc @MekkCyber if you can take look here since it's close to what you did for the other mamba related model (falcon)

romitjain:
@vasqu In the PR that you referenced, the `lazy_load_kernel` function will be called for every forward step [ref](https://github.com/huggingface/transformers/blob/9939e585c0c6d6bdc1eec90c5dfeab2e2e54db11/src/transformers/models/falcon_mamba/modular_falcon_mamba.py#L302)

Since it is not cached, IMO either the approach in this PR or adding `lru_cache` decorator to `lazy_load_kernel` would be a better solve.

WDYT?

src/transformers/utils/import_utils.py
romitjain: Now that I recall it, we would need to this check before trying to load mamba-ssm kernels from kernels hub according to README here: https://huggingface.co/kernels-community/mamba-ssm

Should I inject these requirements in `_HUB_KERNEL_MAPPING` in `src/transformers/integrations/hub_kernels.py`

vasqu: But we don't use this check anywhere, no? It doesn't hurt to have it either way, just a bit confused about the usage

vasqu:
> @vasqu Since it is not cached, IMO either the approach in this PR or adding lru_cache decorator to lazy_load_kernel would be a better solve.

SGTM. I will still leave it to @MekkCyber tho as he's the main guy for anything kernels. In essence, it just should be the same way everywhere to avoid unnecessary tech debt

src/transformers/models/bamba/modular_bamba.py
```diff
@@ -51,23 +52,37 @@
...
 from .configuration_bamba import BambaConfig
 
 
-if is_mamba_2_ssm_available():
-    from mamba_ssm.ops.triton.selective_state_update import selective_state_update
-    from mamba_ssm.ops.triton.ssd_combined import mamba_chunk_scan_combined, mamba_split_conv1d_scan_combined
-else:
-    selective_state_update = None
+def _lazy_load_kernels():
+    global causal_conv1d_update, causal_conv1d_fn
```
MekkCyber: It would be nicer to call it `load_bamba_kernels` for example

src/transformers/models/bamba/modular_bamba.py
```diff
@@ -275,11 +290,16 @@ def __init__(self, config: BambaConfig, layer_idx: int):
 
         self.out_proj = nn.Linear(self.intermediate_size, self.hidden_size, bias=self.use_bias)
 
+        _lazy_load_kernels()
+
+        global is_fast_path_available
+        is_fast_path_available = all((selective_state_update, causal_conv1d_fn, causal_conv1d_update))
+
```
MekkCyber: why do we need `is_fast_path_available` to be global ?

src/transformers/models/jamba/modular_jamba.py
```diff
@@ -24,34 +24,43 @@
...
 
 
-if is_mamba_ssm_available():
-    from mamba_ssm.ops.selective_scan_interface import mamba_inner_fn, selective_scan_fn
-    from mamba_ssm.ops.triton.selective_state_update import selective_state_update
-else:
-    selective_state_update, selective_scan_fn, mamba_inner_fn = None, None, None
+def _lazy_load_kernels():
+    global causal_conv1d_update, causal_conv1d_fn
+    causal_conv1d = lazy_load_kernel("causal-conv1d")
```
MekkCyber: same as above

src/transformers/utils/import_utils.py
MekkCyber: for the clean-mamba-ssm implementation I don't think we need this, let's just remove it for now

MekkCyber (commented):
Thanks a lot @romitjain !

src/transformers/models/bamba/modular_bamba.py
vasqu: Let's name it mamba kernels then tho. It will suggest that bamba kernels are different from mamba which is not the case, same for falcon

src/transformers/utils/import_utils.py
romitjain: Sure, will remove this

Also, re: @vasqu's comment, I had added this earlier but forgot to remove this in my latest commit. Will remove it since clean-mamba-ssm won't need it

src/transformers/models/bamba/modular_bamba.py
romitjain: @MekkCyber This variable is being used at a bunch of different places, for example: https://github.com/romitjain/transformers/blob/a8920655018ccfa1a9768cac0ee64a494541f7f6/src/transformers/models/bamba/modular_bamba.py#L693

MekkCyber: yes it's just a nit but I would prefer keeping it the way it was before, and only doing mandatory changes to add kernels

src/transformers/utils/import_utils.py
romitjain: @MekkCyber `clean-mamba-ssm` has a issue. It does not expose all the kernel functions: mamba_chunk_scan_combined, mamba_split_conv1d_scan_combined

see: https://github.com/huggingface/transformers/pull/41540#issuecomment-3512931063

MekkCyber: Hemmm, sure will export them in the branch

romitjain:
@MekkCyber I have addressed your comments, PTAL

PS: It would require resolution of this issue: https://github.com/huggingface/transformers/pull/41540#discussion_r2518310902

romitjain:
I did not run `make fix-copies` since I believe I should not edit the modeling files. But due to that the CI is failing, lmk what should be the steps?

SunMarc:
In your case, you indeed need to run make fix-copies to propagate the modification you did in the modular files to the real modeling files

romitjain:
@SunMarc Done, can you please review?
cc: @MekkCyber

ArthurZucker (commented):
Nice! As I discussed a bit offline with @MekkCyber we might update this later on but LGTM otherwise, BUT we need to keep BC (if kernel is not installed but ssm and causal conv1d are we gotta support that case)

romitjain:
@ArthurZucker That is working for me! If kernels is not installed, it falls back to look at default system packages

romitjain:
@MekkCyber Gentle reminder. What would be the next steps?

src/transformers/models/granitemoehybrid/modeling_granitemoehybrid.py
```diff
@@ -40,22 +41,9 @@
...
-if is_mamba_2_ssm_available():
-    from mamba_ssm.ops.triton.selective_state_update import selective_state_update
-    from mamba_ssm.ops.triton.ssd_combined import mamba_chunk_scan_combined, mamba_split_conv1d_scan_combined
-else:
-    selective_state_update = None
-
-if is_causal_conv1d_available():
-    from causal_conv1d import causal_conv1d_fn, causal_conv1d_update
-else:
-    causal_conv1d_update, causal_conv1d_fn = None, None
```
vasqu, line 56: Like Arthur said, we still need this as option - users may use the original libraries or kernels. Atm, only kernels would be supported

MekkCyber: this is handled in the lazy load

MekkCyber:
Hi @romitjain, nothing to do on your side for now. I need to refactor the kernel function calls to make them less verbose and less visually intrusive.

romitjain:
Thanks @MekkCyber 👍🏽

src/transformers/models/granitemoehybrid/modeling_granitemoehybrid.py
vasqu: Ah sorry about that, TIL

vasqu: Ok, no I'm confused: `lazy_load_kernel` only loads a kernel from the hub, no? I.e. we load the respective kernel but have no fallbacks to other implementations. 

What I'm referencing is the original packages, i.e. `mamba-ssm` and `causal-conv1d`. Users should be able to use both:
- The kernels from the hub
- The original packages

Maybe I'm overseeing something, but I don't see a reference to the original ones

Edit: I see this is possibly https://github.com/huggingface/transformers/blob/fa6ea6e95f2838067d2f64f7987f70c611bd66b1/src/transformers/integrations/hub_kernels.py#L302-L317

But aren't we ignoring version restrictions atp, i.e. we hope they have a good enough version of that package?

MekkCyber: yup, as we did before

vasqu: > But aren't we ignoring version restrictions atp, i.e. we hope they have a good enough version of that package?

Out of scope for this PR, but this would be nice to add for another PR, i.e. some version checking on fallback. Especially, mamba2 is susceptible to the specific version

MekkCyber: yup sounds good! I only rewrote what we had before in a more general way! we can add version checking later.

romitjain:
@MekkCyber Is there any ETA for this merge?
I can help out with other changes if required in `lazy_load_kernel`.

MekkCyber:
Thanks @romitjain ! We are currently working on adding a decorator to use function kernels from the hub. Once it's ready we can integrate it. Will let you know

romitjain:
@MekkCyber Would really appreciate an update on this. I am happy to help 

I see two functions `use_kernel_forward_from_hub` and `use_kernel_func_from_hub`, do you suggest that they should be used instead?

src/transformers/models/bamba/modular_bamba.py
```diff
@@ -52,20 +53,23 @@
...
 from .configuration_bamba import BambaConfig
 
 
-if is_mamba_2_ssm_available():
-    from mamba_ssm.ops.triton.selective_state_update import selective_state_update
-    from mamba_ssm.ops.triton.ssd_combined import mamba_chunk_scan_combined, mamba_split_conv1d_scan_combined
-else:
-    selective_state_update = None
+causal_conv1d = lazy_load_kernel("causal-conv1d")
+causal_conv1d_update = causal_conv1d.causal_conv1d_update if causal_conv1d is not None else None
```
MekkCyber: let's use the `getattr(causal_conv1d, "causal_conv1d_update", None)` syntax instead everywhere, it's cleaner

src/transformers/models/mamba2/modeling_mamba2.py
```diff
@@ -24,30 +24,37 @@
...
-    from mamba_ssm.ops.triton.selective_state_update import selective_state_update
-    from mamba_ssm.ops.triton.ssd_combined import mamba_chunk_scan_combined, mamba_split_conv1d_scan_combined
-else:
-    mamba_chunk_scan_combined, mamba_split_conv1d_scan_combined, selective_state_update = None, None, None
+causal_conv1d = lazy_load_kernel("causal-conv1d")
+causal_conv1d_update, causal_conv1d_fn = (
+    (causal_conv1d.causal_conv1d_update, causal_conv1d.causal_conv1d_fn) if causal_conv1d is not None else (None, None)
+)
+
+global selective_state_update, mamba_chunk_scan_combined, mamba_split_conv1d_scan_combined
```
MekkCyber: we don't need this I think

MekkCyber (commented):
Hi @romitjain ! After some thought, I think we will keep using the lazy loading for now, the other decorator is mainly going to be used for optimization purposes! left a few comments to address and good to go

src/transformers/models/bamba/modular_bamba.py
romitjain: @MekkCyber `lazy_load_kernel` can return None. In that case this would break

src/transformers/models/mamba2/modeling_mamba2.py
romitjain: I have removed this

romitjain:
@MekkCyber I have resolved your comments.
You might need to fix the upstream mamba-ssm package in kernel hub too.

https://github.com/huggingface/transformers/pull/41540#discussion_r2518310902

src/transformers/models/bamba/modular_bamba.py
MekkCyber: are you sure ? it should work even with a none

romitjain: Ah, my bad. You are right.
Let me update it

src/transformers/models/bamba/modular_bamba.py
```diff
@@ -57,19 +57,17 @@
...
-    mamba_ssm.ops.triton.selective_state_update.selective_state_update if mamba_ssm is not None else None
-)
-mamba_chunk_scan_combined = (
-    mamba_ssm.ops.triton.ssd_combined.mamba_chunk_scan_combined if mamba_ssm is not None else None
-)
-mamba_split_conv1d_scan_combined = (
-    mamba_ssm.ops.triton.ssd_combined.mamba_split_conv1d_scan_combined if mamba_ssm is not None else None
+mamba_ssm_triton = getattr(getattr(mamba_ssm, "ops", None), "triton", None)
+selective_state_update = getattr(
+    getattr(mamba_ssm_triton, "selective_state_update", None), "selective_state_update", None
```
MekkCyber: I think we need to rework how are these imports nested. It doesn't make sense

MekkCyber (commented):
Thanks for fixing this @romitjain, I will find some time this week to rework how imports are done inside the kernel because it's really not optimal to have such nested imports

src/transformers/models/bamba/modular_bamba.py
romitjain: I didn't get you @MekkCyber?
I am also not in favor of these nested imports, but that is what you had previously requested

MekkCyber: yes i mean it shouldn't be done this way in the kernel, like we should have an __init__ file with all necessary functions, and then we only use getattr once, and not in a nested way

romitjain: Sure, let me know once you update the kernels init file, I can make the changes here.

MekkCyber: Thanks a lot ! And sorry this is taking a lot of time 🙏, I’ll work on fixing things this week, at the latest.

romitjain: No worries at all. Thank you for your continued review on this!

MekkCyber:
Hi @romitjain ! It should be good now, you can use the kernels from this version : https://huggingface.co/kernels-community/mamba-ssm/tree/v0.0.4 instead of the `clean-mamba-ssm` one, let me know if you have any problems when you test it

romitjain:
@MekkCyber Thanks for making the upstream fix, all the imports have been improved now to avoid nested calls. PTAL

src/transformers/models/bamba/modeling_bamba.py
```diff
@@ -501,6 +489,15 @@ def apply_mask_to_padding_states(hidden_states, attention_mask):
...
 
 
+causal_conv1d = lazy_load_kernel("causal-conv1d")
+causal_conv1d_update = getattr(causal_conv1d, "causal_conv1d_update", None)
+causal_conv1d_fn = getattr(causal_conv1d, "causal_conv1d_fn", None)
+
+mamba_ssm = lazy_load_kernel("mamba-ssm")
+selective_state_update = getattr(mamba_ssm, "selective_state_update", None)
+mamba_chunk_scan_combined = getattr(mamba_ssm, "mamba_chunk_scan_combined", None)
+mamba_split_conv1d_scan_combined = getattr(mamba_ssm, "mamba_split_conv1d_scan_combined", None)
```
MekkCyber: let's move this inside the Mixer to avoid loading the kernels at import time, the same we did here : https://github.com/romitjain/transformers/blob/03590f18cf59be7a9e215b946bd7ade3d8b12a7a/src/transformers/models/mamba/modeling_mamba.py#L201:L214

MekkCyber (commented):
Thanks @romitjain ! lgtm, only one nit and good to go

src/transformers/models/bamba/modeling_bamba.py
romitjain: @MekkCyber Done for mamba2, jamba and bamba

MekkCyber:
Thanks @romitjain ! can you fix the styling issues with `make style` ?

romitjain:
Oops, I forgot to run that @MekkCyber 
Done now.

MekkCyber (approved):
lgtm! thanks for iterating on this PR

romitjain:
Thanks @MekkCyber 
What would be the next steps for the merge?

MekkCyber:
Failling tests seem unrelated! we just need to wait for a green CI to merge

romitjain:
Seems like it is failing for the latest commits: https://github.com/huggingface/transformers/commits/main/

ArthurZucker (approved):
ty

PR #41541, opened 2025-10-13
Untangle config inheritance

zucchini-nlp:
# What does this PR do?

As per title, deletes base config attributes that are not actually universal for all models (e.g. token ids used only for text models and tying can be done only if model has embedding layers)

After this PR, we need to clean up generation-related params from config classes (https://github.com/huggingface/transformers/pull/41695), and then we can easily turn config classes to a `dataclass`. Using `dataclass` is one of the requirements for hf hub type validation which I am working on currently


zucchini-nlp:
Oh no, I caused 100+ merge conflicts with another PR 🙃

src/transformers/models/aimv2/configuration_aimv2.py
```diff
@@ -188,14 +182,13 @@ def __init__(
         qkv_bias: bool = False,
         mlp_bias: bool = False,
         hidden_act: str = "silu",
-        pad_token_id: Optional[int] = None,
-        bos_token_id: Optional[int] = None,
```
zucchini-nlp: modular file shows that these were supposed to be deleted with `del bos_token_id`, but the conversion script could not handle it correctly

tests/test_modeling_common.py
```diff
@@ -1941,10 +1941,14 @@ def test_resize_embeddings_untied(self):
 
         original_config, inputs_dict = self.model_tester.prepare_config_and_inputs_for_common()
         original_config.tie_word_embeddings = False
+        original_config.get_text_config().tie_word_embeddings = False
```
zucchini-nlp: i'm so surprised this test was working fine, even if we didn't change the text config. Only two models fail if we stop changing `tie_word_embeddings` entirely, so this test prob isn't designed really well

src/transformers/configuration_utils.py
```diff
@@ -206,29 +168,14 @@ def __init__(
...
-        is_decoder: bool = False,
-        cross_attention_hidden_size: Optional[int] = None,
-        add_cross_attention: bool = False,
-        tie_encoder_decoder: bool = False,
         # Fine-tuning task arguments
         architectures: Optional[list[str]] = None,
-        finetuning_task: Optional[str] = None,
         id2label: Optional[dict[int, str]] = None,
         label2id: Optional[dict[str, int]] = None,
         num_labels: Optional[int] = None,
```
Cyrilvallez: Those 3 should disappear as well no? They have nothing to do here in general IMO

Cyrilvallez (commented):
Very nice in general! Happy to have less in the general one, and more in each separate model! looks like we could remove even more attrs from the general one though no? Also, CI not happy yet!

src/transformers/configuration_utils.py
zucchini-nlp: these I was not sure 100% because they are kind of general. Any model with a Sequence Classification head on top would need these and i think all models potentially can be classifiers

zucchini-nlp:
@bot /style

zucchini-nlp:
What happened while rebasing, why 3k test failures 😭

zucchini-nlp:
Oke, 400 failures after fixing tying embeddings in non-LLM models, better. I will wait for the weight tying to be fixed on main, otherwise there will be even more merge conflicts to fix

ArthurZucker (approved):
Very happy about how it looks now!

zucchini-nlp:
I will come back to this one, it will require some time to fix all the issues

src/transformers/models/align/configuration_align.py
```diff
@@ -99,11 +106,15 @@ def __init__(
...
         pad_token_id=0,
+        bos_token_id=None,
+        eos_token_id=None,
         use_cache=True,
+        is_decoder=False,
         **kwargs,
     ):
         super().__init__(**kwargs)
 
+        self.is_decoder = is_decoder
```
zucchini-nlp: this one should be also deleted because it isn't used for most models. Only some classes can be decoder and encoder, like BERT

I am keeping it because of the masking utility until all models start using `create_causal/bidirectional_mask`

src/transformers/configuration_utils.py
```diff
@@ -277,25 +228,15 @@ def __init__(
         self._output_attentions = output_attentions  # has public property
 
         # Less common kwargs, only used by some models
-        if "tie_encoder_decoder" in kwargs:
-            tie_encoder_decoder = kwargs.pop("tie_encoder_decoder")
-            tie_word_embeddings = tie_encoder_decoder or tie_word_embeddings
-
```
zucchini-nlp, line 282: I am deleting `tie_encoder_decoder`. Can't find any model on the hub that uses it and the attr was added only for the custom `EncoderDecoderModel` class. However we can't tie encoder to decoder in `EncoderDecoderModel` anymore, no matter what is the value of this attribute

src/transformers/modeling_utils.py
```diff
@@ -2350,7 +2350,10 @@ def get_expanded_tied_weights_keys(self, all_submodels: bool = False) -> dict:
 
         tied_mapping = self._tied_weights_keys
         # If the config does not specify any tying, return empty dict
-        if not self.config.tie_word_embeddings:
+        # NOTE: not all modules have `tie_word_embeddings` attr, for example vision-only
+        # modules do not have any word embeddings!
+        tie_word_embeddings = getattr(self.config.get_text_config(decoder=True), "tie_word_embeddings", False)
+        if not tie_word_embeddings:
```
zucchini-nlp, line 2398: important! Please review here. There are two big changes:
1. we don't assume that `tie_word_embeddings` on all configs. Some models have no input-output word embeddings and thus nothing to tie (e.g. ViT)
2. We check for the text config here. In multimodals we usually have `tie_word_embeddings` under text config but the text module will not have a class attr `_tied_weights_keys`. For ex: Paligemma might be tying weights via its text config but `GemmaModel` is a base model with no lm head. So we will not execute the rest of code and early exit

zucchini-nlp:
run-slow: afmoe, aimv2, albert, align, altclip, apertus, arcee, aria, audioflamingo3, auto, bamba, bark, bart, bert, bert_generation, big_bird, llama, gemma, gemma3, gemma2, paligemma, llava, llava_next, mistral, qwen2_vl, t5gemma, florence

zucchini-nlp:
Wow, slow tests passing! 😮  In any case, I want to do full CI and I will ask Yih-Dar after the vacations

src/transformers/modeling_utils.py
```diff
@@ -2350,7 +2350,10 @@ def get_expanded_tied_weights_keys(self, all_submodels: bool = False) -> dict:
 
         tied_mapping = self._tied_weights_keys
         # If the config does not specify any tying, return empty dict
-        if not self.config.tie_word_embeddings:
+        # NOTE: not all modules have `tie_word_embeddings` attr, for example vision-only
+        # modules do not have any word embeddings!
+        tie_word_embeddings = getattr(self.config.get_text_config(decoder=True), "tie_word_embeddings", False)
```
Cyrilvallez: No, this should not check the subconfigs! This is very important as this one is already applied on each submodel, so each one should check its own config, not subconfigs!!

tests/test_modeling_common.py
```diff
@@ -1214,9 +1214,18 @@ def test_all_tensors_are_parameter_or_buffer(self):
...
             model.init_weights()
 
+            # Prepare inputs to correct device
+            inputs = self._prepare_for_class(inputs_dict, model_class)
+            final_inputs = {}
+            for k, v in inputs.items():
+                if isinstance(v, torch.Tensor):
+                    final_inputs[k] = v.to(device="cpu")
+                else:
+                    final_inputs[k] = v
```
Cyrilvallez: To remove, this has changed already

tests/test_modeling_common.py
```diff
@@ -2083,7 +2092,10 @@ def test_resize_embeddings_untied(self):
         inputs_dict.pop("labels", None)
 
         # if model cannot untied embeddings -> leave test
-        if original_config.tie_word_embeddings:
+        if (
+            getattr(original_config, "tie_word_embeddings", False)
+            or original_config.get_text_config().tie_word_embeddings
+        ):
```
Cyrilvallez: See my first comment, but this should not need to change I believe

tests/test_modeling_common.py
```diff
@@ -2328,13 +2340,14 @@ def test_tied_weights_keys(self):
...
             model_tied = model_class(copied_config)
 
             tied_weight_keys = _get_tied_weight_keys(model_tied)
             # If we don't find any tied weights keys, and by default we don't tie the embeddings, it's because the model
-            # does not tie them
-            if len(tied_weight_keys) == 0 and not original_config.tie_word_embeddings:
+            # does not tie them or does not have embedding layer (non-text model)
+            if len(tied_weight_keys) == 0 and not getattr(
+                original_config.get_text_config(), "tie_word_embeddings", False
+            ):
```
Cyrilvallez: Same, without the change to `modleing_utils`, this probably does not need to change either

tests/utils/test_configuration_utils.py
```diff
@@ -38,23 +38,10 @@
...
     "id2label": {0: "label"},
     "label2id": {"label": "0"},
-    "tokenizer_class": "BertTokenizerFast",
-    "prefix": "prefix",
-    "bos_token_id": 6,
-    "pad_token_id": 7,
-    "eos_token_id": 8,
-    "sep_token_id": 9,
-    "decoder_start_token_id": 10,
-    "task_specific_params": {"translation": "some_params"},
```
Cyrilvallez, line 57: Are we sure all those atgs that disappeared from PreTrainedConfig will still be inside all the model tester configs??

Cyrilvallez (commented):
Nice! Left a few comments, most notably we really don't want/need to change the logic in `modeling_utils` to check subconfigs!!

src/transformers/modeling_utils.py
zucchini-nlp: yeah, though in VLMs the attribute is stored in text config but the tying is done in `TaskHeadModel`. So when trying to tie weights, we end up checking `self.config` and not `self.config.text_config`
Do we have any models with many text-submodules where one module ties wieghts and the other doesn't? If not, we can default to always checking text config imo

tests/test_modeling_common.py
zucchini-nlp: yep, just needed for CI

tests/utils/test_configuration_utils.py
zucchini-nlp: they aren't used for model tester right now. They were needed only for config tests to check defaults and yeah, the tests are passing now

src/transformers/modeling_utils.py
Cyrilvallez: No no we sorted that issue before. It should NOT check subconfig. Each model (along with its config) is responsible for its own `_tied_weights_keys` set. Most of the time, the submodel do not tie anything, as they don't have lm_head and it's the most usual tying.
If they have `_tied_weights_keys`, then they should adapt their config as needed.
If a ForConditional model needs to tie to a weight of one of its submodels, then he is the one responsible for it once again
Each model is responsible for itself, otherwise the whole logic collapses

Cyrilvallez: If you know any model that should NOT tie and had `tie_word_embeddings=False` in their subconfig instead of themselves, then we should hardcode something in their config, otherwise it's all good as it's True by default

zucchini-nlp:
run-slow: afmoe, aimv2, albert, align, altclip, apertus, arcee, aria, audioflamingo3, auto, bamba, bark, bart, bert, bert_generation, big_bird, llama, gemma, gemma3, gemma2, paligemma, llava, llava_next, llava_next_video, mistral, qwen2_vl, t5gemma, florence

zucchini-nlp:
Full CI didn't show any test errors, so I will be wrapping this up today-tmrw

zucchini-nlp:
Thanks, now I will go back to the hub validation and making configs a `dataclass` 🎉

PR #41545, opened 2025-10-13
TDT for HF

hainan-xv:
# What does this PR do?

Parakeet TDT model integration.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/transformers/models/auto/configuration_auto.py
+++ b/src/transformers/models/auto/configuration_auto.py
@@ -301,7 +301,10 @@ CONFIG_MAPPING_NAMES = OrderedDict[str, str](
         ("owlvit", "OwlViTConfig"),
         ("paligemma", "PaliGemmaConfig"),
         ("parakeet_ctc", "ParakeetCTCConfig"),
+        ("parakeet_tdt", "ParakeetTDTConfig"),
         ("parakeet_encoder", "ParakeetEncoderConfig"),
+        ("parakeet_tdt_decoder", "ParakeetTDTDecoderConfig"),
+        ("parakeet_tdt_joint", "ParakeetTDTJointConfig"),
         ("patchtsmixer", "PatchTSMixerConfig"),
         ("patchtst", "PatchTSTConfig"),
         ("pegasus", "PegasusConfig"),
@@ -759,7 +762,10 @@ MODEL_NAMES_MAPPING = OrderedDict[str, str](
         ("paligemma", "PaliGemma"),
         ("parakeet", "Parakeet"),
         ("parakeet_ctc", "Parakeet"),
+        ("parakeet_tdt", "ParakeetTDT"),
         ("parakeet_encoder", "ParakeetEncoder"),
+        ("parakeet_tdt_decoder", "ParakeetTDTDecoder"),
+        ("parakeet_tdt_joint", "ParakeetTDTJoint"),
         ("patchtsmixer", "PatchTSMixer"),
         ("patchtst", "PatchTST"),
         ("pegasus", "Pegasus"),
@@ -1002,7 +1008,10 @@ SPECIAL_MODEL_TYPE_TO_MODULE_NAME = OrderedDict[str, str](
         ("fastspeech2_conformer_with_hifigan", "fastspeech2_conformer"),
         ("video_llama_3_vision", "video_llama_3"),
         ("parakeet_encoder", "parakeet"),
+        ("parakeet_tdt_decoder", "parakeet"),
+        ("parakeet_tdt_joint", "parakeet"),
         ("parakeet_ctc", "parakeet"),
+        ("parakeet_tdt", "parakeet"),
     ]
 )
 
--- a/src/transformers/models/auto/modeling_auto.py
+++ b/src/transformers/models/auto/modeling_auto.py
@@ -301,7 +301,10 @@ MODEL_MAPPING_NAMES = OrderedDict(
         ("owlvit", "OwlViTModel"),
         ("paligemma", "PaliGemmaModel"),
         ("parakeet_ctc", "ParakeetForCTC"),
+        ("parakeet_tdt", "ParakeetForTDT"),
         ("parakeet_encoder", "ParakeetEncoder"),
+        ("parakeet_tdt_decoder", "ParakeetTDTDecoder"),
+        ("parakeet_tdt_joint", "ParakeetTDTJoint"),
         ("patchtsmixer", "PatchTSMixerModel"),
         ("patchtst", "PatchTSTModel"),
         ("pegasus", "PegasusModel"),
@@ -1624,6 +1627,14 @@ MODEL_FOR_CTC_MAPPING_NAMES = OrderedDict(
     ]
 )
 
+MODEL_FOR_TDT_MAPPING_NAMES = OrderedDict(
+    [
+        # Model for Token-and-Duration Transducer (TDT) mapping.
+        ("parakeet_tdt", "ParakeetForTDT"),
+    ]
+)
+
+
 MODEL_FOR_AUDIO_FRAME_CLASSIFICATION_MAPPING_NAMES = OrderedDict(
     [
         # Model for Audio Classification mapping
@@ -1883,6 +1894,7 @@ MODEL_FOR_AUDIO_CLASSIFICATION_MAPPING = _LazyAutoMapping(
     CONFIG_MAPPING_NAMES, MODEL_FOR_AUDIO_CLASSIFICATION_MAPPING_NAMES
 )
 MODEL_FOR_CTC_MAPPING = _LazyAutoMapping(CONFIG_MAPPING_NAMES, MODEL_FOR_CTC_MAPPING_NAMES)
+MODEL_FOR_TDT_MAPPING = _LazyAutoMapping(CONFIG_MAPPING_NAMES, MODEL_FOR_TDT_MAPPING_NAMES)
 MODEL_FOR_SPEECH_SEQ_2_SEQ_MAPPING = _LazyAutoMapping(CONFIG_MAPPING_NAMES, MODEL_FOR_SPEECH_SEQ_2_SEQ_MAPPING_NAMES)
 MODEL_FOR_AUDIO_FRAME_CLASSIFICATION_MAPPING = _LazyAutoMapping(
     CONFIG_MAPPING_NAMES, MODEL_FOR_AUDIO_FRAME_CLASSIFICATION_MAPPING_NAMES
@@ -2200,6 +2212,11 @@ class AutoModelForCTC(_BaseAutoModelClass):
 
 AutoModelForCTC = auto_class_update(AutoModelForCTC, head_doc="connectionist temporal classification")
 
+class AutoModelForTDT(_BaseAutoModelClass):
+    _model_mapping = MODEL_FOR_TDT_MAPPING
+
+
+AutoModelForTDT = auto_class_update(AutoModelForTDT, head_doc="token-and-duration transducer")
 
 class AutoModelForSpeechSeq2Seq(_BaseAutoModelClass):
     _model_mapping = MODEL_FOR_SPEECH_SEQ_2_SEQ_MAPPING
@@ -2305,6 +2322,7 @@ __all__ = [
     "MODEL_FOR_CAUSAL_IMAGE_MODELING_MAPPING",
     "MODEL_FOR_CAUSAL_LM_MAPPING",
     "MODEL_FOR_CTC_MAPPING",
+    "MODEL_FOR_TDT_MAPPING",
     "MODEL_FOR_DOCUMENT_QUESTION_ANSWERING_MAPPING",
     "MODEL_FOR_DEPTH_ESTIMATION_MAPPING",
     "MODEL_FOR_IMAGE_CLASSIFICATION_MAPPING",
@@ -2352,6 +2370,7 @@ __all__ = [
     "AutoModelForAudioXVector",
     "AutoModelForCausalLM",
     "AutoModelForCTC",
+    "AutoModelForTDT",
     "AutoModelForDepthEstimation",
     "AutoModelForImageClassification",
     "AutoModelForImageSegmentation",
--- a/src/transformers/models/fastspeech2_conformer/modeling_fastspeech2_conformer.py
+++ b/src/transformers/models/fastspeech2_conformer/modeling_fastspeech2_conformer.py
@@ -490,12 +490,12 @@ class FastSpeech2ConformerConvolutionModule(nn.Module):
             kernel_size = module_config["kernel_size"]
             self.activation = ACT2FN[module_config.get("activation", "silu")]
         self.padding = (kernel_size - 1) // 2
-        self.pointwise_conv1 = nn.Conv1d(channels, 2 * channels, kernel_size=1, stride=1, padding=0, bias=True)
+        self.pointwise_conv1 = nn.Conv1d(channels, 2 * channels, kernel_size=1, stride=1, padding=0, bias=config.attention_bias)
         self.depthwise_conv = nn.Conv1d(
-            channels, channels, kernel_size, stride=1, padding=self.padding, groups=channels, bias=True
+            channels, channels, kernel_size, stride=1, padding=self.padding, groups=channels, bias=config.attention_bias
         )
         self.norm = nn.BatchNorm1d(channels)
-        self.pointwise_conv2 = nn.Conv1d(channels, channels, kernel_size=1, stride=1, padding=0, bias=True)
+        self.pointwise_conv2 = nn.Conv1d(channels, channels, kernel_size=1, stride=1, padding=0, bias=config.attention_bias)
 
     def forward(self, hidden_states, attention_mask=None):
         """
--- a/src/transformers/models/parakeet/configuration_parakeet.py
+++ b/src/transformers/models/parakeet/configuration_parakeet.py
@@ -150,6 +150,65 @@ class ParakeetEncoderConfig(PreTrainedConfig):
         self.initializer_range = initializer_range
 
 
+
+class ParakeetTDTDecoderConfig(PreTrainedConfig):
+    model_type = "parakeet_tdt_decoder"
+    keys_to_ignore_at_inference = ["past_key_values"]
+    output_hidden_states = False
+
+    def __init__(
+        self,
+        hidden_size=640,
+        num_hidden_layers=1,
+        dropout=0,
+        vocab_size=1024,
+        forget_gate_bias=1.0,
+        t_max=None,
+        weights_init_scale=1.0,
+        hidden_hidden_bias_scale=0,
+        **kwargs,
+    ):
+        super().__init__(
+            **kwargs,
+        )
+        self.hidden_size = hidden_size
+        self.num_hidden_layers = num_hidden_layers
+        self.dropout = dropout
+        self.vocab_size = vocab_size
+        self.forget_gate_bias=forget_gate_bias
+        self.t_max=t_max
+        self.weights_init_scale=weights_init_scale
+        self.hidden_hidden_bias_scale=hidden_hidden_bias_scale
+
+
+class ParakeetTDTJointConfig(PreTrainedConfig):
+    model_type = "parakeet_tdt_joint"
+    keys_to_ignore_at_inference = ["past_key_values"]
+
+    def __init__(
+        self,
+        enc_hidden_size=1024,
+        pred_hidden_size=640,
+        hidden_size=640,
+        vocab_size=1024,
+        durations=[0,1,2,3,4],
+        norm=None,
+        dropout=0.0,
+        activation='relu',
+        **kwargs,
+    ):
+        super().__init__(
+            **kwargs,
+        )
+        self.enc_hidden_size = enc_hidden_size
+        self.pred_hidden_size = pred_hidden_size
+        self.hidden_size = hidden_size
+        self.vocab_size = vocab_size
+        self.durations = durations
+        self.dropout = dropout
+        self.activation = activation
+
+
 class ParakeetCTCConfig(PreTrainedConfig):
     r"""
     This is the configuration class to store the configuration of a [`ParakeetForCTC`]. It is used to instantiate a
@@ -232,4 +291,83 @@ class ParakeetCTCConfig(PreTrainedConfig):
         return cls(encoder_config=encoder_config.to_dict(), **kwargs)
 
 
-__all__ = ["ParakeetCTCConfig", "ParakeetEncoderConfig"]
+class ParakeetTDTConfig(PreTrainedConfig):
+
+    model_type = "parakeet_tdt"
+    sub_configs = {"encoder_config": ParakeetEncoderConfig, "decoder_config": ParakeetTDTDecoderConfig, "joint_config": ParakeetTDTJointConfig}
+
+    def __init__(
+        self,
+#        bos_token_id=1,
+#        eos_token_id=2,
+#        pad_token_id=1024,
+        tdt_loss_reduction="mean",
+        encoder_config: Union[dict, ParakeetEncoderConfig] = None,
+        decoder_config: Union[dict, ParakeetTDTDecoderConfig] = None,
+        joint_config: Union[dict, ParakeetTDTJointConfig] = None,
+        **kwargs,
+    ):
+
+        if encoder_config is None:
+            self.encoder_config = ParakeetEncoderConfig()
+        elif isinstance(encoder_config, dict):
+            self.encoder_config = ParakeetEncoderConfig(**encoder_config)
+        elif isinstance(encoder_config, ParakeetEncoderConfig):
+            self.encoder_config = encoder_config
+        else:
+            raise ValueError(
+                f"`encoder_config` must be a dictionary or an instance of `ParakeetEncoderConfig`, got {type(encoder_config)}"
+            )
+
+        if decoder_config is None:
+            self.decoder_config = ParakeetTDTDecoderConfig()
+        elif isinstance(decoder_config, dict):
+            self.decoder_config = ParakeetTDTDecoderConfig(**decoder_config)
+        elif isinstance(decoder_config, ParakeetTDTDecoderConfig):
+            self.decoder_config = decoder_config
+        else:
+            raise ValueError(
+                f"`decoder_config` must be a dictionary or an instance of `ParakeetEncoderConfig`, got {type(encoder_config)}"
+            )
+
+        if joint_config is None:
+            self.joint_config = ParakeetTDTJointConfig()
+        elif isinstance(joint_config, dict):
+            self.joint_config = ParakeetTDTJointConfig(**joint_config)
+        elif isinstance(joint_config, ParakeetTDTJointConfig):
+            self.joint_config = joint_config
+        else:
+            raise ValueError(
+                f"`decoder_config` must be a dictionary or an instance of `ParakeetEncoderConfig`, got {type(encoder_config)}"
+            )
+
+        vocab_size = self.joint_config.vocab_size
+        self.vocab_size = vocab_size
+
+        self.blank_token_id = vocab_size
+        super().__init__(
+#            pad_token_id=self.blank_token_id,
+            **kwargs,
+        )
+
+    @classmethod
+    def from_configs(
+        cls,
+        encoder_config: ParakeetEncoderConfig,
+        decoder_config: ParakeetTDTDecoderConfig,
+        joint_config: ParakeetTDTJointConfig,
+        **kwargs):
+        r"""
+        Instantiate a [`ParakeetConfig`] (or a derived class) from parakeet encoder model configuration.
+
+        Returns:
+            [`ParakeetConfig`]: An instance of a configuration object
+        """
+
+        return cls(
+            encoder_config=encoder_config.to_dict(),
+            decoder_config=decoder_config.to_dict(),
+            joint_config=joint_config.to_dict(),
+            **kwargs)
+
+__all__ = ["ParakeetCTCConfig", "ParakeetTDTConfig", "ParakeetEncoderConfig", "ParakeetTDTDecoderConfig", "ParakeetTDTJointConfig"]
--- a/src/transformers/models/parakeet/modular_parakeet.py
+++ b/src/transformers/models/parakeet/modular_parakeet.py
@@ -17,21 +17,21 @@
 import math
 from collections.abc import Callable
 from dataclasses import dataclass
-from typing import Optional, Union
+from typing import Optional, Union, Tuple
 
 import torch
 from torch import nn
 
 from ...activations import ACT2FN
 from ...modeling_layers import GradientCheckpointingLayer
-from ...modeling_outputs import BaseModelOutput, CausalLMOutput
+from ...modeling_outputs import BaseModelOutput, CausalLMOutput, BaseModelOutputWithNoAttention
 from ...modeling_utils import ALL_ATTENTION_FUNCTIONS, PreTrainedModel
 from ...processing_utils import Unpack
 from ...utils import ModelOutput, TransformersKwargs, auto_docstring, can_return_tuple
 from ...utils.generic import check_model_inputs
 from ..fastspeech2_conformer.modeling_fastspeech2_conformer import FastSpeech2ConformerConvolutionModule
 from ..llama.modeling_llama import LlamaAttention, eager_attention_forward
-from .configuration_parakeet import ParakeetCTCConfig, ParakeetEncoderConfig
+from .configuration_parakeet import PreTrainedConfig, ParakeetCTCConfig, ParakeetTDTConfig, ParakeetEncoderConfig, ParakeetTDTDecoderConfig, ParakeetTDTJointConfig
 
 
 class ParakeetEncoderRelPositionalEncoding(nn.Module):
@@ -111,7 +111,7 @@ class ParakeetEncoderAttention(LlamaAttention):
         super().__init__(config, layer_idx=layer_idx)
         self.is_causal = False
         # W_{k,R} projection
-        self.relative_k_proj = nn.Linear(config.hidden_size, config.num_attention_heads * self.head_dim, bias=False)
+        self.relative_k_proj = nn.Linear(config.hidden_size, config.num_attention_heads * self.head_dim, bias=config.attention_bias)
         # global content bias
         self.bias_u = nn.Parameter(torch.zeros(config.num_attention_heads, self.head_dim))
         # global positional bias
@@ -301,7 +301,7 @@ class ParakeetEncoderBlock(GradientCheckpointingLayer):
 
 @auto_docstring
 class ParakeetPreTrainedModel(PreTrainedModel):
-    config: ParakeetCTCConfig
+    config: PreTrainedConfig
     base_model_prefix = "model"
     main_input_name = "input_features"
     input_modalities = "audio"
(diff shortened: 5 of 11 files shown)


Rocketknight1:
cc @eustlb @ebezzam for audio

src/transformers/models/parakeet/configuration_parakeet.py
```diff
@@ -232,4 +291,83 @@ def from_encoder_config(cls, encoder_config: ParakeetEncoderConfig, **kwargs):
...
+class ParakeetTDTConfig(PreTrainedConfig):
+
+    model_type = "parakeet_tdt"
+    sub_configs = {"encoder_config": ParakeetEncoderConfig, "decoder_config": ParakeetTDTDecoderConfig, "joint_config": ParakeetTDTJointConfig}
+
+    def __init__(
+        self,
+#        bos_token_id=1,
+#        eos_token_id=2,
+#        pad_token_id=1024,
```
ebezzam: We can remove commented lines aren't used

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -301,7 +301,7 @@ def forward(
 
 @auto_docstring
 class ParakeetPreTrainedModel(PreTrainedModel):
-    config: ParakeetCTCConfig
+    config: PreTrainedConfig
```
ebezzam, line 304: Makes sense to change this. Just wondering if you've tried removing the line altogether? Or did you have to specify `config` to something?

src/transformers/models/auto/modeling_auto.py
```diff
@@ -2305,6 +2322,7 @@ def from_pretrained(cls, pretrained_model_name_or_path, *model_args, **kwargs):
     "MODEL_FOR_CAUSAL_IMAGE_MODELING_MAPPING",
     "MODEL_FOR_CAUSAL_LM_MAPPING",
     "MODEL_FOR_CTC_MAPPING",
+    "MODEL_FOR_TDT_MAPPING",
```
ebezzam, line 2325: Can you place in alphabetic order?

src/transformers/models/fastspeech2_conformer/modeling_fastspeech2_conformer.py
```diff
@@ -490,12 +490,12 @@ def __init__(self, config: FastSpeech2ConformerConfig, module_config=None):
...
         self.padding = (kernel_size - 1) // 2
-        self.pointwise_conv1 = nn.Conv1d(channels, 2 * channels, kernel_size=1, stride=1, padding=0, bias=True)
+        self.pointwise_conv1 = nn.Conv1d(channels, 2 * channels, kernel_size=1, stride=1, padding=0, bias=config.attention_bias)
         self.depthwise_conv = nn.Conv1d(
-            channels, channels, kernel_size, stride=1, padding=self.padding, groups=channels, bias=True
+            channels, channels, kernel_size, stride=1, padding=self.padding, groups=channels, bias=config.attention_bias
         )
         self.norm = nn.BatchNorm1d(channels)
-        self.pointwise_conv2 = nn.Conv1d(channels, channels, kernel_size=1, stride=1, padding=0, bias=True)
+        self.pointwise_conv2 = nn.Conv1d(channels, channels, kernel_size=1, stride=1, padding=0, bias=config.attention_bias)
```
ebezzam, line 498: [This PR](https://github.com/huggingface/transformers/pull/41969/files#diff-7c70128195a0a5a9a6b522cf2387d5704b9698fea138a7f30b5c010921fa1054) seems to have already made this change and merged into main with the name `config.convolution_bias`. Can you sync with main?

src/transformers/models/auto/modeling_auto.py
```diff
@@ -2352,6 +2370,7 @@ def from_pretrained(cls, pretrained_model_name_or_path, *model_args, **kwargs):
     "AutoModelForAudioXVector",
     "AutoModelForCausalLM",
     "AutoModelForCTC",
+    "AutoModelForTDT",
```
ebezzam, line 2373: Can you also place in alphabetic order?

src/transformers/models/parakeet/configuration_parakeet.py
```diff
@@ -150,6 +150,65 @@ def __init__(
         self.initializer_range = initializer_range
 
 
+
+class ParakeetTDTDecoderConfig(PreTrainedConfig):
+    model_type = "parakeet_tdt_decoder"
+    keys_to_ignore_at_inference = ["past_key_values"]
+    output_hidden_states = False
```
ebezzam: It already defaults to False ([here](https://github.com/huggingface/transformers/blob/a8f32a0e9c2fec0f196fab5d8316a03f35d0c528/src/transformers/configuration_utils.py#L204)) so no need to set.

src/transformers/models/parakeet/configuration_parakeet.py
```diff
@@ -150,6 +150,65 @@ def __init__(
...
+
+    def __init__(
+        self,
+        hidden_size=640,
+        num_hidden_layers=1,
+        dropout=0,
+        vocab_size=1024,
+        forget_gate_bias=1.0,
+        t_max=None,
```
ebezzam, line 166: (Transformers convention) is this parameter used? I see it is only used [here](https://github.com/hainan-xv/transformers/blob/8a3e1cdb762f2b0334fec00193f4d740c04a3ae7/src/transformers/models/parakeet/modular_parakeet.py#L532). If the final model checkpoint does not use it, the Transformer convention is to remove such variables and unused code paths.

_If it is used_, could you use a more explicit name than `t_max`? Namely being more verbose on what `t` is.

src/transformers/models/parakeet/configuration_parakeet.py
```diff
@@ -150,6 +150,65 @@ def __init__(
...
+        weights_init_scale=1.0,
+        hidden_hidden_bias_scale=0,
```
ebezzam, line 168: just `hidden_bias_scale` here and elsewhere? (instead of double `hidden_hidden_bias_scale`)

src/transformers/models/parakeet/configuration_parakeet.py
```diff
@@ -150,6 +150,65 @@ def __init__(
...
+    keys_to_ignore_at_inference = ["past_key_values"]
+
+    def __init__(
+        self,
+        enc_hidden_size=1024,
+        pred_hidden_size=640,
+        hidden_size=640,
+        vocab_size=1024,
+        durations=[0,1,2,3,4],
+        norm=None,
```
ebezzam, line 195: can we remove `norm` as it seems unused?

src/transformers/models/parakeet/configuration_parakeet.py
```diff
@@ -232,4 +291,83 @@ def from_encoder_config(cls, encoder_config: ParakeetEncoderConfig, **kwargs):
...
+        if encoder_config is None:
+            self.encoder_config = ParakeetEncoderConfig()
+        elif isinstance(encoder_config, dict):
+            self.encoder_config = ParakeetEncoderConfig(**encoder_config)
+        elif isinstance(encoder_config, ParakeetEncoderConfig):
+            self.encoder_config = encoder_config
+        else:
+            raise ValueError(
+                f"`encoder_config` must be a dictionary or an instance of `ParakeetEncoderConfig`, got {type(encoder_config)}"
+            )
```
ebezzam, line 320: No need to raise ValueError. You can do like in the CTC config [here](https://github.com/huggingface/transformers/blob/a8f32a0e9c2fec0f196fab5d8316a03f35d0c528/src/transformers/models/parakeet/configuration_parakeet.py#L212-L215)

src/transformers/models/parakeet/configuration_parakeet.py
```diff
@@ -150,6 +150,65 @@ def __init__(
...
+    ):
+        super().__init__(
+            **kwargs,
+        )
+        self.hidden_size = hidden_size
+        self.num_hidden_layers = num_hidden_layers
+        self.dropout = dropout
+        self.vocab_size = vocab_size
+        self.forget_gate_bias=forget_gate_bias
+        self.t_max=t_max
```
ebezzam, line 179: If used, also renaming `self.t_max` to something more verbose

src/transformers/models/parakeet/configuration_parakeet.py
```diff
@@ -232,4 +291,83 @@ def from_encoder_config(cls, encoder_config: ParakeetEncoderConfig, **kwargs):
...
+
+    model_type = "parakeet_tdt"
+    sub_configs = {"encoder_config": ParakeetEncoderConfig, "decoder_config": ParakeetTDTDecoderConfig, "joint_config": ParakeetTDTJointConfig}
+
+    def __init__(
+        self,
+#        bos_token_id=1,
+#        eos_token_id=2,
+#        pad_token_id=1024,
+        tdt_loss_reduction="mean",
```
ebezzam, line 304: seems to be unused and can be removed?

src/transformers/models/parakeet/configuration_parakeet.py
ebezzam: Similarly if `t_max` is always used, from what I understand `forget_gate_bias` would not be used [here](https://github.com/hainan-xv/transformers/blob/8a3e1cdb762f2b0334fec00193f4d740c04a3ae7/src/transformers/models/parakeet/modular_parakeet.py#L547C14-L554) so that code path could be removed instead

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+                    hidden_size = n // 4
+                    p.data.fill_(0)
+                    p.data[hidden_size : 2 * hidden_size] = torch.log(
+                        torch.nn.init.uniform_(p.data[0:hidden_size], 1, t_max - 1)
+                    )
+                    # forget gate biases = log(uniform(1, Tmax-1))
+                    p.data[0:hidden_size] = -p.data[hidden_size : 2 * hidden_size]
+                    # input gate biases = -(forget gate biases)
+
+        elif forget_gate_bias is not None:
```
ebezzam, line 547: As mentioned above, are there final model checkpoints that will use the two code paths of `t_max` and `forget_gate_bias`?

In general (Transformers convention) if a code path is not used by a final model checkpoint, it should be removed.

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -111,7 +111,7 @@ def __init__(self, config: ParakeetEncoderConfig, layer_idx: int):
         super().__init__(config, layer_idx=layer_idx)
         self.is_causal = False
         # W_{k,R} projection
-        self.relative_k_proj = nn.Linear(config.hidden_size, config.num_attention_heads * self.head_dim, bias=False)
+        self.relative_k_proj = nn.Linear(config.hidden_size, config.num_attention_heads * self.head_dim, bias=config.attention_bias)
```
ebezzam, line 114: Interesting that `bias` was hardcoded to False...

Can you check if this breaks Parakeet CTC? It has `attention_bias=True` (see [here](https://huggingface.co/nvidia/parakeet-ctc-1.1b/blob/main/config.json#L10)) but would have been hardcoding to False before.

src/transformers/models/parakeet/configuration_parakeet.py
```diff
@@ -150,6 +150,65 @@ def __init__(
...
+            **kwargs,
+        )
+        self.hidden_size = hidden_size
+        self.num_hidden_layers = num_hidden_layers
+        self.dropout = dropout
+        self.vocab_size = vocab_size
+        self.forget_gate_bias=forget_gate_bias
+        self.t_max=t_max
+        self.weights_init_scale=weights_init_scale
+        self.hidden_hidden_bias_scale=hidden_hidden_bias_scale
```
ebezzam, line 181: As above, renaming to `self.hidden_bias_scale`

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+            hidden_hidden_bias_scale: Float scale for the hidden-to-hidden bias scale. Set to 0.0 for
+                the default behaviour.
+
+        Returns:
+            A `torch.nn.LSTM`.
+        """
+        super(ParakeetLSTM, self).__init__()
+
+        self.lstm = torch.nn.LSTM(
+            input_size=input_size, hidden_size=hidden_size, num_layers=num_layers, dropout=dropout, proj_size=proj_size
```
ebezzam, line 529: For using `dropout`, see Llama ([here](https://github.com/huggingface/transformers/blob/f8e5ae6a5072cfea78c5d052ebe61b63c9cb10cc/src/transformers/models/llama/modeling_llama.py#L288)) for setting it according to `self.training`.

Namely something like:
```python
self.lstm = torch.nn.LSTM(
    input_size=input_size, hidden_size=hidden_size, num_layers=num_layers, dropout=0.0 if not self.training else dropout, proj_size=proj_size
)
```

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+                v.data *= float(weights_init_scale)
+
+    def forward(
+        self, x: torch.Tensor, h: Optional[Tuple[torch.Tensor, torch.Tensor]] = None
+    ) -> Tuple[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
+
+        x, h = self.lstm(x, h)
+
+        if self.dropout:
+            x = self.dropout(x)
```
ebezzam, line 569: Let's use `nn.functional.dropout` instead, like in `ParakeetEncoderFeedForward` [here](https://github.com/huggingface/transformers/blob/f8e5ae6a5072cfea78c5d052ebe61b63c9cb10cc/src/transformers/models/parakeet/modular_parakeet.py#L108)

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+
+        self.dropout = torch.nn.Dropout(dropout) if dropout else None
+
+        for name, v in self.named_parameters():
+            if 'weight' in name or 'bias' in name:
+                v.data *= float(weights_init_scale)
+
+    def forward(
+        self, x: torch.Tensor, h: Optional[Tuple[torch.Tensor, torch.Tensor]] = None
+    ) -> Tuple[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
```
ebezzam, line 564: (Transformers convention) can you change single letter variables for the inputs and within the function to more explicit names, e.g. `hidden_states` for `h` and `input_features` for `x`?

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+        elif forget_gate_bias is not None:
+            for name, v in self.lstm.named_parameters():
+                if "bias_ih" in name:
+                    bias = getattr(self.lstm, name)
+                    bias.data[hidden_size : 2 * hidden_size].fill_(forget_gate_bias)
+                if "bias_hh" in name:
+                    bias = getattr(self.lstm, name)
+                    bias.data[hidden_size : 2 * hidden_size] *= float(hidden_hidden_bias_scale)
+
+        self.dropout = torch.nn.Dropout(dropout) if dropout else None
```
ebezzam, line 556: Can be removed after switching to functional dropout below

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+
+class ParakeetTDTPredictor(ParakeetPreTrainedModel):
+
+    def __init__(self, config: ParakeetTDTDecoderConfig):
+        super().__init__(config)
+        self.gradient_checkpointing = False
+        self.config = config
+
+        self.embed = torch.nn.Embedding(config.vocab_size + 1, config.hidden_size) # +1 for blank
+        self.dec_rnn = self.rnn(
```
ebezzam, line 631: (Transformers convention) let's be more explicit here and directly call `ParakeetLSTM` instead of this functional wrapper (which can be removed)

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+    @check_model_inputs()
+    @can_return_tuple
+    def forward(
+        self,
+        input_token,
+        states,
+        hidden_state = None,
+        **kwargs: Unpack[TransformersKwargs],
+    ):
+        assert input_token is not None
```
ebezzam, line 679: (Transformers convention) we don't use assert statements but rather `raise ValueError("msg")`. Though in this case I think we don't need either?

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+
+        device = self.embed.weight.device
+        if input_token.device != device:
+            input_token = input_token.to(device)
+        return self.predict(input_token, state=states)
```
ebezzam, line 684: (Transformers convention) let's be more explicit here and write out what `self.predict` does instead of calling a separate function (which can be removed after the change)

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -627,4 +898,106 @@ def generate(
...
+        self,
+        input_features: torch.Tensor,
+        **kwargs: Unpack[TransformersKwargs],
+    ):
+        encoder_outputs = self.encoder(
+            input_features=input_features,
+            **kwargs,
+        )
+
+        logits = self.joint.joint_net(self.joint.enc(encoder_outputs.last_hidden_state)) #[:,:,:self.joint.vocab_size]
```
ebezzam, line 930: Let's break this up into multiple lines for better readability (see below for an idea)

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -627,4 +898,106 @@ def generate(
...
+
+        return CausalLMOutput(
+            loss=torch.sum(encoder_outputs.last_hidden_state), # a fake loss here.
```
ebezzam, line 933: is it possible to use a more meaningful loss?

And also shifting loss computation above as in `ParakeetForCTC.forward` [here](https://github.com/huggingface/transformers/blob/f8e5ae6a5072cfea78c5d052ebe61b63c9cb10cc/src/transformers/models/parakeet/modular_parakeet.py#L522).

For example:
```python
    @auto_docstring
    @can_return_tuple
    def forward(
        self,
        input_features: torch.Tensor,
        labels: Optional[torch.Tensor] = None,
    ) -> CausalLMOutput:
    """ TODO example in the docstring"""

        encoder_outputs = self.encoder(
            input_features=input_features,
            **kwargs,
        )

        hidden_states = encoder_outputs.last_hidden_state
        joint_encoder_output = self.joint.enc(hidden_states)
        logits = self.joint.joint_net(joint_encoder_output)

        loss = None
        if labels is not None:
            raise ValueError("Loss not implemented yet.")

        return CausalLMOutput(
            loss=loss,
            logits=logits,
            hidden_states=encoder_outputs.hidden_states,
            attentions=encoder_outputs.attentions,
        )

```

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -627,4 +898,106 @@ def generate(
...
+    @can_return_tuple
+    def forward(
+        self,
+        input_features: torch.Tensor,
+        **kwargs: Unpack[TransformersKwargs],
+    ):
+        encoder_outputs = self.encoder(
+            input_features=input_features,
+            **kwargs,
+        )
```
ebezzam, line 928: Is it intended that the encoder doesn't take as input an `attention_mask` like in the CTC variant?

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -627,4 +898,106 @@ def generate(
...
+class ParakeetForTDT(ParakeetPreTrainedModel):
+    config: ParakeetTDTConfig
+
+    def __init__(self, config: ParakeetTDTConfig):
+        super().__init__(config)
+        self.encoder = ParakeetEncoder(config.encoder_config)
+        self.decoder = ParakeetTDTDecoder(config.decoder_config)
+        self.joint = ParakeetTDTJoint(config.joint_config)
+        self.blank_token_id = config.blank_token_id
+        self.max_token_per_frame = 2
```
ebezzam, line 915: can this be added to the configuration?

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -627,4 +898,106 @@ def generate(
...
+        self,
+        input_features: torch.Tensor,
+        **kwargs: Unpack[TransformersKwargs],
+    ):
+
+        encoder_outputs = self.encoder(
+            input_features=input_features,
+            **kwargs,
+        )
+        output = self.greedy_decode(encoder_outputs.last_hidden_state)
```
ebezzam, line 950: (Transformers convention) explicitly write out operations than calling another function (`greedy_decode` can be removed after the change)

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -627,4 +898,106 @@ def generate(
...
+            hidden_states=encoder_outputs.hidden_states,
+            attentions=encoder_outputs.attentions,
+        )
+
+    @torch.no_grad()
+    def generate(
+        self,
+        input_features: torch.Tensor,
+        **kwargs: Unpack[TransformersKwargs],
+    ):
```
ebezzam, line 944: (Transformers convention) can you imitate `ParakeetForCTC` [here](https://github.com/huggingface/transformers/blob/f8e5ae6a5072cfea78c5d052ebe61b63c9cb10cc/src/transformers/models/parakeet/modular_parakeet.py#L594), namely:
- example in the docstring
- add `return_dict_in_generate`
- `attention_mask` not needed for the encoder?

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -627,4 +898,106 @@ def generate(
...
+
+        encoder_outputs = self.encoder(
+            input_features=input_features,
+            **kwargs,
+        )
+        output = self.greedy_decode(encoder_outputs.last_hidden_state)
+
+        return output
+
+    def greedy_decode(self, encoder_output):
```
ebezzam, line 954: (Transformers convention) as mentioned above, can you use more verbose variable names instead of `T`, `t`, `hyp`, `dec_out`, `v`, `g`. Thanks!

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+    ) -> Tuple[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
+
+        x, h = self.lstm(x, h)
+
+        if self.dropout:
+            x = self.dropout(x)
+
+        return x, h
+
+class ParakeetTDTJoint(ParakeetPreTrainedModel):
```
ebezzam, line 573: (Transformers convention) Since `ParakeetTDTJoint` is not expected to be a self-sufficient model, but rather as part of `ParakeetForTDT` let's simply have it inherit from `nn.Module`

src/transformers/models/parakeet/configuration_parakeet.py
```diff
@@ -150,6 +150,65 @@ def __init__(
...
+
+class ParakeetTDTJointConfig(PreTrainedConfig):
+    model_type = "parakeet_tdt_joint"
+    keys_to_ignore_at_inference = ["past_key_values"]
+
+    def __init__(
+        self,
+        enc_hidden_size=1024,
+        pred_hidden_size=640,
+        hidden_size=640,
```
ebezzam, line 192: Is there a case where `pred_hidden_size` would be different than `hidden_size`?

If so, can we rename to `decoder_hidden_size` and use the terminology "decoder" instead of "predictor" throughout?

src/transformers/models/parakeet/configuration_parakeet.py
```diff
@@ -232,4 +291,83 @@ def from_encoder_config(cls, encoder_config: ParakeetEncoderConfig, **kwargs):
...
+        self,
+#        bos_token_id=1,
+#        eos_token_id=2,
+#        pad_token_id=1024,
+        tdt_loss_reduction="mean",
+        encoder_config: Union[dict, ParakeetEncoderConfig] = None,
+        decoder_config: Union[dict, ParakeetTDTDecoderConfig] = None,
+        joint_config: Union[dict, ParakeetTDTJointConfig] = None,
+        **kwargs,
+    ):
```
ebezzam, line 309: (Transformers convention) Let's do something similar to `ParakeetCTCConfig` where we only define a subconfig for the encoder. 

I don't expect `ParakeetTDTDecoderConfig` and `ParakeetTDTJointConfig` to be used by models other than `ParakeetForTDT`? (But correct me if I'm wrong). The reason a subconfig was made for `ParakeetEncoderConfig` was because it is being used by CTC, TDT, and maybe more.

For example:

```python
class ParakeetTDTConfig(PreTrainedConfig):
    """ TODO docstring"

    model_type = "parakeet_tdt"
    sub_configs = {"encoder_config": ParakeetEncoderConfig}

    def __init__(
        self,
        encoder_config: Union[dict, ParakeetEncoderConfig] = None,
        hidden_size=640,
        decoder_hidden_size=640,     # only if there is a case where different from `hidden_size`
        hidden_act="relu",
        num_hidden_layers=1,
        dropout=0.0,
        [forget_gate_bias and/or t_max as discussed above,
        weights_init_scale=1.0,
        durations=[0, 1, 2, 3, 4],   
        hidden_bias_scale=0,
        vocab_size=1024,
        blank_token_id=1024,      # can pad_token_id naming be used instead and passed to `super().__init__`?
        **kwargs,
    ):
        # TODO set parameters like in ParakeetCTCConfig
        ...

        super().__init__(**kwargs)

    @property
    def encoder_hidden_size(self) -> int:
        return self.encoder_config.hidden_size

```

Note that we can define properties for attributes like `encoder_hidden_size` above that are meant to be equivalent, and I've suggested more verbose variable names and naming as per Transformers convention.

Moreover, in your conversion script, I notice that you are setting the `pad_token_id`, `vocab_size`, and `blank_token_id` [here](https://github.com/hainan-xv/transformers/blob/8a3e1cdb762f2b0334fec00193f4d740c04a3ae7/src/transformers/models/parakeet/convert_nemo_to_hf.py#L340-L342), but in your config [this will](https://github.com/hainan-xv/transformers/blob/8a3e1cdb762f2b0334fec00193f4d740c04a3ae7/src/transformers/models/parakeet/configuration_parakeet.py#L344-L347) take priority. So we should be sure that they are defaulted correctly above.  

I also notice that you don't use `pad_token_id` in the TDT model. I'm wondering if we should use that naming instead of `blank_token_id` (as this variable isn't common in Transformers). We can come back to that after a first iteration of changes.

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+        enc = enc.view([-1, self.config.enc_hidden_size])  # making it B, D
+        enc = self.enc(enc)
+        pred = self.pred(pred)
+
+        assert enc.shape[0] == pred.shape[0]
+        output = self.joint_net(enc + pred)
+        return BaseModelOutput(last_hidden_state=output)
+
+
+class ParakeetTDTPredictor(ParakeetPreTrainedModel):
```
ebezzam: Let's call `ParakeetTDTDecoder` to be consistent with the configuration name?

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+    _can_record_outputs = {}
+    _no_split_modules = None
+
+    def __init__(self, config: ParakeetTDTJointConfig):
+        super().__init__(config)
+        self.config = config
+        self.gradient_checkpointing = False
+
+        self.enc = torch.nn.Linear(config.enc_hidden_size, config.hidden_size)
+        self.pred = torch.nn.Linear(config.pred_hidden_size, config.hidden_size)
```
ebezzam, line 590: (Transformers convention) When defining these layers within `ParakeetForTDT`, let's use more explicit naming:
```python
self.encoder_projector = torch.nn.Linear(config.encoder_hidden_size, config.hidden_size)
self.decoder_projector = torch.nn.Linear(config.decoder_hidden_size, config.hidden_size)
```

Note I'm making the assumption that we rename "predictor" to "decoder" in relevant places for more consistent naming.

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+
+        num_classes = config.vocab_size + 1 + len(config.durations)
+
+        layers = (
+            [torch.nn.ReLU(inplace=True)]
+            + ([torch.nn.Dropout(p=self.config.dropout)])
+            + [torch.nn.Linear(config.hidden_size, num_classes)]
+        )
+        self.joint_net = torch.nn.Sequential(*layers)
```
ebezzam, line 599: (Transformers convention) let's switch to [ModuleList](https://docs.pytorch.org/docs/stable/generated/torch.nn.ModuleList.html) instead of Sequential. You will have to update the conversion script as this will change the state dict

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+        self.gradient_checkpointing = False
+
+        self.enc = torch.nn.Linear(config.enc_hidden_size, config.hidden_size)
+        self.pred = torch.nn.Linear(config.pred_hidden_size, config.hidden_size)
+
+        num_classes = config.vocab_size + 1 + len(config.durations)
+
+        layers = (
+            [torch.nn.ReLU(inplace=True)]
+            + ([torch.nn.Dropout(p=self.config.dropout)])
```
ebezzam, line 596: Can we use `ACT2FN` with the configuration parameters, as in the encoder [here](https://github.com/huggingface/transformers/blob/f8e5ae6a5072cfea78c5d052ebe61b63c9cb10cc/src/transformers/models/parakeet/modular_parakeet.py#L102)

And also use `self.training` for the dropout layer

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+        )
+        self.joint_net = torch.nn.Sequential(*layers)
+        self.post_init()
+
+    @auto_docstring
+    @check_model_inputs()
+    def forward(
+        self,
+        enc: torch.Tensor,
+        pred: torch.Tensor,
```
ebezzam, line 607: can we use more verbose variable names? e.g. `encoder_output` and `decoder_output`

src/transformers/models/auto/configuration_auto.py
```diff
@@ -301,7 +301,10 @@
         ("owlvit", "OwlViTConfig"),
         ("paligemma", "PaliGemmaConfig"),
         ("parakeet_ctc", "ParakeetCTCConfig"),
+        ("parakeet_tdt", "ParakeetTDTConfig"),
         ("parakeet_encoder", "ParakeetEncoderConfig"),
+        ("parakeet_tdt_decoder", "ParakeetTDTDecoderConfig"),
+        ("parakeet_tdt_joint", "ParakeetTDTJointConfig"),
```
ebezzam, line 307: Such mappings can be removed after removing `ParakeetTDTDecoderConfig` and `ParakeetTDTJointConfig`

tests/models/parakeet/test_modeling_parakeet.py
```diff
@@ -375,3 +608,195 @@ def test_1b_model_integration_batched(self):
...
         self.assertListEqual(predicted_transcripts, EXPECTED_TRANSCRIPTIONS)
+
+
+
+class ParakeetForTDTModelTester:
+    def __init__(self,
+                 parent,
+                 encoder_kwargs=None,
+                 decoder_kwargs=None,
+                 joint_kwargs=None,
```
ebezzam, line 619: Adapting to new configuration structure

tests/models/parakeet/test_modeling_parakeet.py
```diff
@@ -184,6 +191,232 @@ def test_model_get_set_embeddings(self):
         pass
 
 
+class ParakeetTDTDecoderModelTester:
```
ebezzam, line 194: Such tests won't be needed anymore for the decoder and joint "models" after integrating into the TDT model!

tests/models/parakeet/test_modeling_parakeet.py
```diff
@@ -375,3 +608,195 @@ def test_1b_model_integration_batched(self):
...
+            )
+
+    def _load_datasamples(self, num_samples):
+        self._load_dataset()
+        ds = self._dataset
+        speech_samples = ds.sort("id")[:num_samples]["audio"]
+        return [x["array"] for x in speech_samples]
+
+    @slow
+    def test_1b_model_integration(self):
```
ebezzam, line 780: I suppose the reproducers and integration tests still need to be done?

The ones for [CTC](https://gist.github.com/ebezzam/6382bdabfc64bb2541ca9f77deb7678d#file-reproducer_single-py) are a good example

ebezzam (commented):
@hainan-xv thanks for the PR to add the TDT variant!

It may seem like a lot of comments at the first glance, but they are mainly about Transformers conventions which aren't obvious. So I've tried to be explicit to help you with the changes (but let me know if something is unclear). It's already a great start that you've implemented the changes through modular 👏 

A couple other points that come to mind:
- I suppose [this](https://huggingface.co/hainanx/parakeet-tdt-0.6b-v3) is the Transformers-compatible checkpoint you've created with conversion script?
- We should also update the [documentation file](https://github.com/huggingface/transformers/blob/main/docs/source/en/model_doc/parakeet.md) to mention the TDT variant

Thanks 🤗

src/transformers/models/parakeet/modular_parakeet.py
ebezzam, line 623: (Transformers convention) Similar to `ParakeetTDTJoint` this can inherit from nn.Module instead

src/transformers/models/parakeet/modular_parakeet.py
ebezzam: Eq 4 of [the paper](https://arxiv.org/pdf/2304.06795) for the loss.

Also I just noticed that the `forward` method isn't being called by `generate`? This should be the case (see [CTC](https://github.com/huggingface/transformers/blob/6217adc6c8f0be7b5374e6a46129ad2214e4c6ed/src/transformers/models/parakeet/modular_parakeet.py#L623)), so we'll have to rethink how the components are called within `forward`. We can see after a first iteration of changes.

ebezzam (commented):
A couple more thoughts after going through the paper

src/transformers/models/parakeet/modular_parakeet.py
```diff
@@ -483,6 +483,277 @@ class ParakeetGenerateOutput(ModelOutput):
...
+
+    @auto_docstring
+    @check_model_inputs()
+    @can_return_tuple
+    def forward(
+        self,
+        input_token,
+        states,
+        hidden_state = None,
+        **kwargs: Unpack[TransformersKwargs],
```
ebezzam, line 677: Can be dropped as `hidden_state` unused and `**kwargs` should no longer be necessary when `ParakeetTDTPredictor` inherits from `nn.Module`

ebezzam:
#44171 is the current PR for adding TDT

PR #41548, opened 2025-10-13
[device_map] Accelerate loading by computing device_map much faster

Cyrilvallez:
# What does this PR do?

If a model has a lot of different parameters (which is the case for models with a lot of experts, such as Qwen3Next), the computation of the device_map itself (yes, just the mapping of modules to devices) is EXTREMELY slow when using `accelerate`.
This is because they traverse the model graph in a very non-optimized manner, resulting in quadratic complexity in the number of parameters and modules, when we could do it in linear time instead. For example, `Qwen/Qwen3-Next-80B-A3B-Instruct` has a total of `173529` parameters and intermediate modules, which is a a lot. So doing the following:

```python
from transformers import AutoConfig, Qwen3NextForCausalLM
from transformers.modeling_utils import _get_device_map
import torch
import time

model_name = "Qwen/Qwen3-Next-80B-A3B-Instruct"
config = AutoConfig.from_pretrained(model_name)

with torch.device("meta"):
    model = Qwen3NextForCausalLM(config)

t0 = time.time()
device_map = _get_device_map(model, device_map="auto", dtype=torch.bfloat16, max_memory=None, hf_quantizer=None, keep_in_fp32_regex=None)
dt = time.time() - t0
print(f"Time needed: {dt:.2f} s")

>>> Time needed: 60.02 s
```

takes un UNBEARABLE `60.02` s on my Macbook Pro M3. And this is only for the device_map computation, the loading has not even started yet...

This PR fixes the issue by rewriting the needed functions from `accelerate` with much better complexity and operations directly in Transformers, leading to linear time and the following:

```python
>>> Time needed: 1.33 s
```

So effectively going 60x faster from 1min to 1s

The device_map created are the exact same.

### DIFF
--- a/src/transformers/integrations/accelerate.py
+++ b/src/transformers/integrations/accelerate.py
@@ -16,11 +16,12 @@ Some of the functions here are derived from the `accelerate` library, with some 
 and simplicity/ease of use.
 """
 
+import copy
 import inspect
 import os
 from collections import defaultdict
 from contextlib import contextmanager
-from typing import TYPE_CHECKING, Optional, Union
+from typing import TYPE_CHECKING
 
 from ..utils import (
     is_accelerate_available,
@@ -42,8 +43,6 @@ if is_accelerate_available():
     from accelerate.utils import check_tied_parameters_on_same_device, get_max_memory
 
 if TYPE_CHECKING:
-    import re
-
     from ..modeling_utils import PreTrainedModel
     from ..quantizers import HfQuantizer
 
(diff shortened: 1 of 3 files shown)


src/transformers/modeling_utils.py
```diff
@@ -4590,8 +4500,15 @@ def from_pretrained(
...
 
-        # Regex to keep a fixed dtype
+        # Regex to keep a fixed fp32 dtype
         keep_in_fp32_regex = get_keep_in_fp32_regex(model, hf_quantizer, dtype)
+        # Set some modules to fp32 if needed
+        if keep_in_fp32_regex is not None and dtype != torch.float32:
+            for name, param in model.named_parameters():
+                if keep_in_fp32_regex.search(name):
+                    # param = param.to(torch.float32) does not work here as only in the local scope.
+                    param.data = param.data.to(torch.float32)
```
Cyrilvallez: By doing this here, earlier than before, we can avoid passing `keep_in_fp32_regex` to every functions in the loading pipeline, simply relying on the dtype of the meta-device model.
This is much simpler and avoids checking the regex several times

src/transformers/integrations/accelerate.py
```diff
@@ -249,6 +257,203 @@ def check_and_set_device_map(device_map):
     return device_map
 
 
+def compute_module_sizes(
```
ArthurZucker, line 260: we can prob push this further btw: 
1. hash every nn.Modulist's element's print (to know if there are different variable names etc) 
2. compute only for the different hashes (worst case all layers are different, best case (like classic experts) we only compute this once per DECODER layer -> linearly faster

src/transformers/modeling_utils.py
ArthurZucker: can we do this inside get_keep_in_fp32 (returning the model) ?

ArthurZucker (commented):
Turbo nice

src/transformers/modeling_utils.py
Cyrilvallez: Yes indeed, 100%!

Cyrilvallez: Made it a method of PreTrainedModel, as it modifies it in-place

src/transformers/integrations/accelerate.py
Cyrilvallez: I checked and 75-80% of the time taken in `_get_device_map` is consumed by the remaining call to accelerate function `infer_auto_device_map`. For now, I did not replace it as it was not the bottleneck, but I'm planning to do it in the future (but it's not such a big issue for now, this PR already fixes the big one).
So it's not really worth the dive for now, as we really would not gain much. I'll do it when replacing `infer_auto_device_map`, because then we would see true gains!

SunMarc:
FYI, we have this small improvement in accelerate https://github.com/huggingface/accelerate/pull/3776 but nice that you manage to optimize further

PR #41552, opened 2025-10-13
[Bugfix] Fix quantizer import

kylesayrs:
## Purpose ##
* Fix bad import in `src/transformers/quantizers/base.py`

## Changes ##
* Import `PreTrainedModel` dynamically, rather than only importing when type checking
  * The class is needed for an `isinstance` check

## Testing ##
* Code which calls `_assign_original_dtype` which was previously failing is now passing
  * Can provide a minimum example if required

### DIFF
--- a/src/transformers/quantizers/base.py
+++ b/src/transformers/quantizers/base.py
@@ -18,7 +18,6 @@ from ..utils import is_torch_available, logging
 from ..utils.quantization_config import QuantizationConfigMixin, QuantizationMethod
 from .quantizers_utils import get_module_from_name
 
-
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


deepNoah:
The code that was working fine until yesterday generates an error with the message below and the model cannot be loaded.
`NameError: name 'PreTrainedModel' is not defined`

image

please edit

vasqu:
cc @ArthurZucker @Cyrilvallez

kylesayrs:
https://github.com/huggingface/transformers/pull/41571

vasqu:
Too many PRs for the same thing 😢 thx for the PR nonetheless

PR #41559, opened 2025-10-14
Fix executorch export with dynamic shapes

justinchuby:
# What does this PR do?

This PR addd shape compatibility check for attention mask to ensure torch.export can reason about the logic without failing _when exporting with dynamic shapes_.

It additionally simplified the sdpa forward function for export only usage.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@CyrilVallez @jackzhxng @guangy10


justinchuby:
Is fx tracing still supported? It doesn’t seem to be compatible with torch._check here?

Cyrilvallez:
Hey @justinchuby! Sorry but I don't get why this is needed. `torch._check` are very internal dynamo checks, and I don't see why we would have this here. Moreover, `torch.export` is only supported with a mask through the `executorch` integration, as otherwise the mask creation breaks due to using `vmap`

justinchuby:
@Cyrilvallez thanks for the suggestion - I can test with the executorch integration. For context, we are looking to enable onnx export via torch.export. 

Could you share guidance on a potential integration « onnx » directory next to the current executorch directory, for correct usages of onnx export calls? I assume we can follow a similar model with executorch where we have an integration and a separate optimum-onnx project. 

torch.export recently resolved the vmap issues, and the torch._check was useful for the dynamic shapes engine of the tracer to understand equivalence. I agree that it should be part of the torch.export integration.

justinchuby:
I updated the PR to fix the executorch integration. Please take another look. Thanks!

Cyrilvallez:
>torch.export recently resolved the vmap issues

Ouuhhh that's very very nice, I wasn't aware of it!

src/transformers/integrations/executorch.py
```diff
@@ -1200,3 +1203,54 @@ def sdpa_mask_without_vmap(
...
+        value,
+        attn_mask=attention_mask,
+        dropout_p=dropout,
+        scale=scaling,
+        is_causal=is_causal,
+        **sdpa_kwargs,
+    )
+    attn_output = attn_output.transpose(1, 2).contiguous()
+
+    return attn_output, None
```
Cyrilvallez, line 1256: Can we instead simply do the `torch._check` and then call `sdpa_attention_forward`? Would be much simpler!

Cyrilvallez (commented):
Nice! I like this solution much more, as it's contained to `export`-specific functionalities only! However, I believe that we can simplify a lot!

Cyrilvallez:
BTW, we will probably add a non-vmap path to the sdpa mask creation due to vmap being much slower, and the impact being noticeable for small models (https://github.com/huggingface/transformers/issues/41639). This PR is still welcome though, as its focus is on the sdpa attention, not sdpa mask, but I thought I'd mention it just in case!

Cyrilvallez:
>Is fx tracing still supported? It doesn’t seem to be compatible with torch._check here?

Concerning this, the answer is no: https://github.com/huggingface/transformers/pull/41683!

src/transformers/integrations/executorch.py
justinchuby: The check is needed because of this line `attention_mask = attention_mask[:, :, :, : key.shape[-2]]` Do you have suggestion on how it can be avoided or moved out maybe? The check needs to follow this slice operation. The reason is that the when slicing and when key.shape[-2] is dynamic, torch.export wouldn't know if it is actually getting the exact full slice.

I also simplified the gqa logic to always use enable_gqa instead of repeat_interleave because the option is now supported.

Cyrilvallez: Ohhh, I see - actually this slicing ops should not be here at all, i.e. the mask should be correctly prepared upstream, which is the case for all recent models using the nice mask primitives.
I wanted to check if we still have older models for which it's still necessary at some point, here and in the `eager_attention_forward` attentions. 
And if you start by slicing the mask, then _check, and then call `sdpa_attention_forward`, would it work? It would get re-sliced in the call, but to the exact same length, not sure if the _check would be lost then?

justinchuby: > And if you start by slicing the mask, then _check, and then call sdpa_attention_forward, would it work? 

I can check that - but it is going to create a duplicated slice op which is not ideal. I also want to make sure `enable_gqa` is used. Maybe for that we can update sdpa_attention_forward to always use enable_gqa instead?

justinchuby: If we can get slicing out from the forward function, and ensure `enable_gqa` when exporting, then this patch can be simplified to `_check` then call sdpa_attention_forward

justinchuby: An attempt: https://github.com/huggingface/transformers/pull/41699

justinchuby: Actually, export does not care about the math kernel. So we still need to specify it here or improve the check

jiqing-feng:
I've tried this PR and found it didn't solve the performance regression in #41639 . Will it be fixed in the next PR? 

cc @justinchuby  @Cyrilvallez

src/transformers/integrations/executorch.py
vasqu: I thought recent versions had support for xformers backend but seems like I was wrong :cry: so yea we will need to change the conditions somehow.

For the slices, as Cyril said it's due to some old models using a different mask creation where we sometime encountered this. I think it's time to check which are still affected :eyes:

justinchuby:
> I've tried this PR and found it didn't solve the performance regression in #41639 . Will it be fixed in the next PR?
> 
> cc @justinchuby @Cyrilvallez

I believe this PR is unrelated to the issue you linked

src/transformers/integrations/executorch.py
justinchuby: What condition would you suggest I change to? Thanks

Maybe I can instead remove the slice and see what breaks?

justinchuby: https://github.com/huggingface/transformers/pull/41900

vasqu: > What condition would you suggest I change to? Thanks

I don't have a good suggestion tbh, is there a way to tell we are tracing with export only? The issue is that we really can't expect people to only use export and falling back to the math kernel is likely more expensive than the manual repeats. 

> Maybe I can instead remove the slice and see what breaks? 

Responding here but saw the other PR. I think that's the right way but let's wait on @Cyrilvallez to come back (next week).

justinchuby:
I think we can merge https://github.com/huggingface/transformers/pull/41900, then I will create a separate PR to guard torch.export on dynamic shapes.

tests/test_executorch.py
```diff
@@ -127,3 +127,37 @@ def test_decoder_only_lm_export(self):
             inputs_embeds=self.inputs_embeds, cache_position=self.cache_position
         )
         torch.testing.assert_close(eager_output_embeds, exported_output_embeds, atol=1e-4, rtol=1e-4)
+
+    def test_decoder_only_lm_export_with_dynamic_shapes(self):
+        """Test TorchExportableModuleForDecoderOnlyLM export with dynamic sequence lengths."""
```
justinchuby, line 132: Need to guard dynamic shapes on the mask

src/transformers/integrations/executorch.py
justinchuby: Yes, we can use `torch.compiler.is_exporting()` to check that.

vasqu: Then let's add a condition on the gqa function (i.e. here https://github.com/huggingface/transformers/blob/02c324f43fe0ef5d484e846417e5f3bf4484524c/src/transformers/integrations/sdpa_attention.py#L30) to allow GQA in any case when we detect exporting. Is it limited to a specific version of export? Might need double check the torch version. 

Side note and semi relevant: the masks are refactored a bit in #41852 so we won't need a workaround for export in the future.

justinchuby: Great! I can create a separate update for use_gqa_in_sdpa. The most important fix has been removing the slice on masks.

PR #41567, opened 2025-10-14
Fully deprecate AutoGPTQ and AutoAWQ for GPT-QModel

Qubitium:
Remove autogptq clutter and autogptq related configs that are not worth adding backward compat. 

GPTQModel has a slight project name change (pypi package and import name stays the same) to GPT-QModel with `-` as we now have added `awq`/AutoAWQ into our repo and will be making pr soon to address awq loading using GPT-QModel. 

`GPTQConfig` has the most important changes in this PR:

```python

# New GPTQConfig Property. Applicable for sister Peft/Optimum PRs
act_group_aware (`bool`, *optional*, defaults to `True`):
    Use GAR (group aware activation order) during quantization. Has measurable positive impact on quantization
    quality. Only applicable when `desc_act = False`. Will forced to be `False` when `desc_act = True`.
    
    
# Removed GPTQConfig Properties:
use_cuda_fp16
use_exllama
exllama_config
```

The 3 removed properties are all related `kernel` selection. These 3 are a hot potatoe mess and legacy from autogptq. GPT-QModel uses unified `backend` (existing) property to select kernels.  There were compat codes written to `convert`  these 3 properties to `backend` behind the scenes in 2024 but no longer relevant for 2025. 

Note: 

* Transformers/Optimum/Peft  CI tests should never check for `kernel.QUANT_TYPE` (str). GPTQ-QModel will return best performing kernel for the relevant module and it may be different per module due to in/out features and other gptq/module properties in relation to device type + dtype + many factors. 
* CI tests should only assert check for `kernel.QUANT_TYPE` if the test specifies a specific kernel via `backend` selection.


Rocketknight1:

Qubitium:
We have begun AutoAWQ deprecation as well. 

* Fused module codes have all been removed. AutoAWQ used to do quant linear level fusing but I do not believe that this is  maintainable or good since if SGLang/vLLM adopts Transformers v5 for model loading, they will do their own auto fusing and the quant module should not interfere with this. 

* IPEX is deprecated by Intel and we have a new AwqTorchFused kernel (based on same Intel TorchFused kernel for GPTQ) so any code/unit tests for IPEX now will point to AwqTorchFused kernel.

MekkCyber:
Hi @Qubitium ! Thanks a lot for working on this! Quick question, what do you mean by AutoAWQ being part of GPT-QModel now? Did you integrate the entire library (including the transformers dependency, like AutoAWQ does), or did you just port over the linear layers, kernels, and related components?

Qubitium:
> Hi @Qubitium ! Thanks a lot for working on this! Quick question, what do you mean by AutoAWQ being part of GPT-QModel now? Did you integrate the entire library (including the transformers dependency, like AutoAWQ does), or did you just port over the linear layers, kernels, and related components?

Long story short. We folded AutoAWQ into GPT-QModel in multiple stage over the past few weeks. Stage 1. Simple/Directly port/copy  the AutoAWQ code over.  Stage 2. Refractor. Stage 3. Fixed bugs, add new kernels. Major refractor to align with new internal life cycle in GPT-QModel v5.0+. So we are current post Stage 3 where GPT-QModel base retains minimal original AutoAWQ code. Most AutoAWQ code have been refractored away. 

Major Changes vs AutoAWQ:

1. New kernels. We have 2 new kernels for AWQ (`AwqTorch` pure torch based, and `AwqTorchFused` which is cpu optimized based on work by Intel @jiqing-feng. 
2. Plan to add 3rd new AWQ kernel based on Bitblas as most gptq kernels are compatbile with AWQ with some small changes. Marlin kernel also sycned with gptq Marlin kernel for updated Marlin fixes/otpimizations via vllm port. 
3. QuantLinear code have been rewritten/refractored. 
4. Quant logic is new due to GPT-QModel 5.0+ life cycle which is not compatible with AutoAWQ. 

HF eco system compat:

Work on `Peft` integration is happening in a parallel pr by @LRL2-ModelCloud https://github.com/huggingface/peft/pull/2917 in coordination from @BenjaminBossan  https://github.com/huggingface/peft/issues/2342#issuecomment-3547697772

The `Peft` pr will need to co-exist concurrently with this PR due to interdependency. 

We will hold off `Optimum` change last if we  can help it, or may have to parallel a 3rd Pr to `Optimum` as well if inter-dependency causes trouble there as well. 

Final goal of the 2 Prs is to remove dead AutoGPTQ code (no one uses it or should use it frankly) and almost dead AutoAWQ (repo in readonly and no longer accepting bug fixes or new model support).  Compat of model loading of old models that use these two packages will be maintained.

SunMarc:
Thanks for working on this @Qubitium . We are still debating if this is something that should offload to GPTQ-Model or we should start upstreaming some of the inference code directly into transformers + `kernels`. Here is a proposal from a contributor https://github.com/huggingface/transformers/pull/42256. 
The goal would be to only upstream the GEMM path but we can potentially leave the other kernels to GPTQ-Model WDYT ? 

About `GPTQ-Model`, will it be possible to create awq quants for newer models that are compatible with other frameworks (e.g vllm) just like `autoawq` did ?

Qubitium:
> WDYT

I think badly of this proposal. 

> Thanks for working on this @Qubitium . We are still debating if this is something that should offload to GPTQ-Model or we should start upstreaming some of the inference code directly into transformers + `kernels`. Here is a proposal from a contributor #42256. The goal would be to only upstream the GEMM path but we can potentially leave the other kernels to GPTQ-Model WDYT ?
> 
> About `GPTQ-Model`, will it be possible to create awq quants for newer models that are compatible with other frameworks (e.g vllm) just like `autoawq` did ?

I just checked the PR which has no code. I am not going to waste time arguing over vaporware vs what I have done with AWQ in GPT-QModel over the past 2 months that is awq inference and quantization full stack complete with new kernels, new model support with full ci kernel and modeling validation. GPT-Qmodel can be viewed as not an AutoAWQ port and a full point release upgrade in every regard. 

Edit: I have outlined in prior post on why `Fusing` is a bad idea. It is not AWQ's job to fuse in 2025. Leave it to model makers and higher level engines such as SGLang, vlLLM which HF v5.0 is targeting from my understading.

Qubitium:
> About GPTQ-Model, will it be possible to create awq quants for newer models that are compatible with other frameworks (e.g vllm) just like autoawq did ?

Our quantized models are more compatible with vllm/SGLang than ones quantized with Optimum or AutoAWQ

SGLang/vLLM compat is a number 1 target/design from day one so 100% yes.

jiqing-feng:
It's a good chance to deprecate autoawq as it's archived. I suppose the best way is to go upstream to the transformer's main codes, just like we did in the Autogptq replacement. For example, the [IPEX linear in AutoAWQ](https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/awq.py#L212) is out-of-date, we need a new implementation for it. The new linear implementation is TorchFusedLinear in gptqmodel.

Qubitium:
> It's a good chance to deprecate autoawq as it's archived. I suppose the best way is to go upstream to the transformer's main codes, just like we did in the Autogptq replacement. For example, the [IPEX linear in AutoAWQ](https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/awq.py#L212) is out-of-date, we need a new implementation for it. The new linear implementation is TorchFusedLinear in gptqmodel.

@jiqing-feng  The AWQ version of GPTQ `TorchFusedKernel` has been added to gpt-qmodel as `AwqTorchFusedKernel`. Same underlying code but memory layout tweaks to get it to work. AWQ Kernel output tests passing. 

@SunMarc For the most part the kernels for AWQ and GPTQ are shared. For example, we do not compile an extra awq only Marlin kernel for AWQ. The previous gptq only Marlin kernel is synced from vLLM to run AWQ weights as well.

Qubitium:
CI Passing status using GPT-QModel `main` branch:

```
transformers/tests/quantization/autoawq/test_awq.py:
test_awq.py::AwqTest::test_quantized_model PASSED
test_awq.py::AwqTest::test_quantized_model_bf16 PASSED
test_awq.py::AwqTest::test_quantized_model_conversion PASSED
test_awq.py::AwqTest::test_quantized_model_exllama FAILED <-- Needs fixing. 
test_awq.py::AwqTest::test_quantized_model_multi_accelerator SKIPPED
test_awq.py::AwqTest::test_quantized_model_no_device_map PASSED
test_awq.py::AwqTest::test_save_pretrained PASSED
test_awq.py::AwqTest::test_raise_if_non_quantized PASSED
test_awq.py::AwqTest::test_quantized_model_no_k_proj_quantized PASSED
test_awq.py::AwqScaleTest::test_load_quantized_model PASSED
test_awq.py::AwqIPEXTest::test_quantized_model_ipex PASSED <-- test needs to be renamed to AwqTorchFused (ipex removed)
 
peft/tests/test_gpu_examples.py:
PeftAwqGPUTests PASSED
PeftGPTQGPUTests PASSED
 ```
 
 @SunMarc  PR is in working state and ready for prelim review.  Look at the code diffs, we are eliminating 5x more crud for every line of code we add for the new awq integration.

File: src/transformers/quantizers/quantizer_awq.py [hunk 1]
```diff
@@ -129,12 +129,6 @@ def _process_model_before_weight_loading(
             )
 
     def _process_model_after_weight_loading(self, model, **kwargs):
-        if self.quantization_config.do_fuse:
-            from ..integrations import fuse_awq_modules
-
-            model = fuse_awq_modules(model, self.quantization_config)
-            model._awq_is_fused = True  # TODO: consider storing this flag in model.config instead
-
```
SunMarc, line 135: > Fused module codes have all been removed. AutoAWQ used to do quant linear level fusing but I do not believe that this is maintainable or good since if SGLang/vLLM adopts Transformers v5 for model loading, they will do their own auto fusing and the quant module should not interfere with this.

Fusing only happens when the user specify `do_fuse` in the config when loading the awq model using `from_pretained`, so it shouldn't impact at all sglang or vllm at all. Also you can't serialize the model if we fuse the modules. So I think we should still try to maintain that path if possible

src/transformers/integrations/awq.py
```diff
@@ -124,35 +124,35 @@ def replace_with_awq_linear(
 
     backend = quantization_config.backend
 
-    if not is_auto_awq_available():
+    if not is_gptqmodel_available():
         raise ValueError(
-            "AWQ (either `autoawq` or `llmawq`) is not available. Please install it with `pip install autoawq` or check out the installation guide in https://github.com/mit-han-lab/llm-awq"
+            "AWQ (either `llmawq`) is not available. Please install it with `pip install gptqmodel` or check out the installation guide in https://github.com/mit-han-lab/llm-awq"
```
SunMarc: Just a note but if this doesn't make into v5, we will have to slowly deprecate autoawq

src/transformers/quantizers/quantizer_gptq.py
```diff
@@ -22,7 +22,7 @@
 if TYPE_CHECKING:
     from ..modeling_utils import PreTrainedModel
 
-from ..utils import is_auto_gptq_available, is_gptqmodel_available, is_optimum_available, is_torch_available, logging
+from ..utils import is_gptqmodel_available, is_optimum_available, is_torch_available, logging
```
SunMarc, line 25: can do two seperate pr for gptq and awq ? For gptq one, I will be able to quickly merge it

src/transformers/integrations/awq.py
```diff
@@ -124,35 +124,35 @@ def replace_with_awq_linear(
...
         )
 
     if backend == AwqBackendPackingMethod.AUTOAWQ:
         if quantization_config.version == AWQLinearVersion.GEMM:
-            from awq.modules.linear.gemm import WQLinear_GEMM
+            from gptqmodel.nn_modules.qlinear.awq_gemm import AwqGEMMQuantLinear
 
-            target_cls = WQLinear_GEMM
+            target_cls = AwqGEMMQuantLinear
```
SunMarc: As I said, we might have replace this path by one handled by `kernels` at some point

src/transformers/integrations/awq.py
```diff
@@ -124,35 +124,35 @@ def replace_with_awq_linear(
...
-            target_cls = WQLinear_GEMV
+            target_cls = AwqGEMVQuantLinear
         elif quantization_config.version == AWQLinearVersion.EXLLAMA:
             if quantization_config.exllama_config["version"] == ExllamaVersion.ONE:
-                from awq.modules.linear.exllama import WQLinear_Exllama
+                from gptqmodel.nn_modules.qlinear.awq_exllama import AwqExllamaQuantLinear
 
-                target_cls = WQLinear_Exllama
+                target_cls = AwqExllamaQuantLinear
             elif quantization_config.exllama_config["version"] == ExllamaVersion.TWO:
```
SunMarc: unlike gptq, the version selection is not automatic, is that right ?

SunMarc (commented):
Thanks, left a couple of comments. As I said, I'm happy to see that you are willing to fill the hole left by AutoAWQ and eager to see this PR merged. However, note that maybe in the future, we will add a default working path for GEMM if gptq-model is not installed. As those libraries depends on kernels that requires to deal with building + distribution for each new version of torch, we never know when this will suddenly stop. 
Left a couple of comments. Also maybe it will be better to split this PR into 2: one for gptq and one for awq ?

src/transformers/integrations/awq.py
Qubitium: Just double checked this code and it is incomplete. Need to changed to do auto-kernel selection just like gptq kernel selection. The reason is same `format` (`version`) has multiple compatible kernels (GEMM can be mapped to [AwqTorch, AwqTorchFued, AwqGEMM, AwqMarlin]) and in the same reason unreasonable to expect users to manually pass `backend` to select kernels. 

After udpate, this entire block of manual kernel seection will replaced by one line of `hf_select_awq_kernel` or something similar. 

Note that we will be removing old awq flawed terminalogy of `version` (actually `format`), and `backend` (no need for this unique for llm-awq as we will auto-compat during config loading for llm-awq where there is no `quant_method` and only `version` attribute). Backward compat will be maintained via config load/save mapping.

src/transformers/integrations/awq.py
Qubitium: That's fine. At that point in the future, HF staff can override the auto kernel selection code from gpt-qmodel and return specific `AwqGEMM` from `kernels`. It will be a clean override and requires no changes from gpt-qmodel. My point is that it is unreasonable to burden our task further by imposing a future optional requirement that that does not resolve the issue now.

src/transformers/quantizers/quantizer_awq.py
Qubitium: Fusing has no place in AWQ. QKV, MLP fusing should be done at an upper level just like what vLLM and SGLang is doing. This is wrong code/logic for 2025. AutoAWQ did it to squeeze some inference perpformance but if you look at the code, it is a hot mess of static model class mapping and unmtainable. The number of new models coming out and emergence of SGLang, vLLM make this obsolete. It is my understanding that v5 HF is going to be used for SGLang/vLLM loading model foudnation so that makes it even more so. 

For those users that depend on `AutoAWQ` fusing, they need to choose. We are not gonna spend vaulable energy supporting dead code.

src/transformers/integrations/awq.py
Qubitium: I will try to make this as clean as possible to reach that v5 goal. We are removig 75% of the code, not adding. Other than `fusing`, the features are not deprecated and only improved with zero compt issues. New kernels, bettter hw compat, faster kernels, and even bug fixes. Current awq kernels are `fp16` only and failed all our `bf16` kernel output quality tests. We will make sure users do not execute in bf16 or at least warn when this happens (loading model and executing in bf16 when awq kernels are validated for fp16 only).

src/transformers/quantizers/quantizer_gptq.py
Qubitium: https://github.com/huggingface/peft/pull/2917#discussion_r2551674217

GPTQ changes are minimal and mostly cosmetic. But this PR is required for https://github.com/huggingface/peft/pull/2917#discussion_r2551674217 due to interdependency.

Qubitium:
@SunMarc @MekkCyber  Hold off on review. I will ping once ready. I need to remove more code related to fuse and kernel selection.

Qubitium:
@SunMarc  @MekkCyber  Update. This PR will updated once we finish small refractor and add sync auto kernel selection just like what we did with gptq in https://github.com/ModelCloud/GPTQModel/pull/2214/. Both gptq and awq kernel selection will be folded into single `hf_select_quant_linear_v2` interface for stability and single entry point. 

in addition, the original AwqGEMM kernel will be split into effectively 3 distinct kernels, TorchGEMM, CudaGEMM, TritonGEMM. The autoawq gemm kernel was actually 3 kernels in one monolithic one. Sound nice but terrible for ci/kernel output regression/comparison tests with zero performance benefit. GPT-QModel will auto select the kernels based on system env, device_map, kernel qualifications (method, format, etc).  This will knock off another layer of complexity in existing HF code.

```py
# public/stable api exposed to transformer/optimum
def hf_select_quant_linear_v2(
        bits: int,
        group_size: int,
        desc_act: bool,
        sym: bool,
        format: Union[str, FORMAT], # awq `version` should be pre-mapped to format
        quant_method: Union[str, METHOD], # awq llm-awq `version` should be pre-mapped to method
        zero_point: Optional[bool] = True, # awq only
        dtype: Optional[Union[str, torch.dtype]] = None,
        meta: Optional[Dict[str, any]] = None,
        pack: Optional[bool] = True,
        device_map: Optional[Union[str, dict]] = None,
        backend: Optional[Union[str, BACKEND]] = None,
) -> Type[BaseQuantLinear]:
```

jiqing-feng:
> Thanks, left a couple of comments. As I said, I'm happy to see that you are willing to fill the hole left by AutoAWQ and eager to see this PR merged. However, note that maybe in the future, we will add a default working path for GEMM if gptq-model is not installed. As those libraries depends on kernels that requires to deal with building + distribution for each new version of torch, we never know when this will suddenly stop. Left a couple of comments. Also maybe it will be better to split this PR into 2: one for gptq and one for awq ?

Does it mean we can upstream some specific ops for awq or gptq in the kernel-community? In that case, gptqmodel can pull kernels from the community at runtime?

Qubitium:
@SunMarc @MekkCyber PR is now synced to Peft/Optimum pending Prs. Ready for code review for this portion.  All tests passing with pending gpt-qmodel 5.4.4 release (later today). 

Notable changes:

1. `hf_select_quant_linear_v2` will now auto select kernel for both gptq/autoawq. No more kernel selection crud in transformers and gptq/awq kernel selection merged into single api strictly used for `hf` for future api stability. Let gpt-qmodel decide as it has the best view to return the best/latest kernel.

2. AutoAWQ `fusing` codes have been removed. This code is not maintainable (static map based, model arch specific) and is not relevant for vllm/sglang as they do their own fusing. Tranformer v5 I believe is also introducing more generic fusing so any manual, per model arch, fusing done by previous autoawq code should be eliminated. 

3. AwqConfig now inherits from GPTQConfig due to shared properties. For GPTQ, legacy `checkpoint_format` is remapped to `format` internally but for backward compat, until future deprecation, we also write to `checkpoint_format` on save via `to_dict`.  For AWQ, `version` is now mapped to `format` internally, and likewise for compat, we write to `version` using `format` value in `to_dict`.  This is consistent with what gpt-qmodel does for code clarity while maintaining backward compat.

src/transformers/quantizers/quantizer_awq.py
```diff
@@ -129,36 +87,23 @@ def _process_model_before_weight_loading(
...
+            model = hf_gptqmodel_post_init(model, use_act_order=self.quantization_config.desc_act)
 
-        if self.quantization_config.version == AWQLinearVersion.IPEX:
-            from ..integrations import post_init_awq_ipex_modules
-
-            model = post_init_awq_ipex_modules(model)
+        # if self.quantization_config.version == AWQLinearVersion.IPEX:
+        #     from ..integrations import post_init_awq_ipex_modules
+        #
+        #     model = post_init_awq_ipex_modules(model)
```
SunMarc: should this be removed ?

src/transformers/utils/quantization_config.py
```diff
@@ -864,80 +807,57 @@ class AwqConfig(QuantizationConfigMixin):
...
             that quantize their own models using `llm-awq` library.
         do_fuse (`bool`, *optional*, defaults to `False`):
-            Whether to fuse attention and mlp layers together for faster inference
+            Deprecated, Whether to fuse attention and mlp layers together for faster inference
         fuse_max_seq_len (`int`, *optional*):
-            The Maximum sequence length to generate when using fusing.
+            Deprecated, The Maximum sequence length to generate when using fusing.
         modules_to_fuse (`dict`, *optional*, default to `None`):
-            Overwrite the natively supported fusing scheme with the one specified by the users.
+            Deprecated, Overwrite the natively supported fusing scheme with the one specified by the users.
```
SunMarc: remove it directly since those are not used

src/transformers/utils/quantization_config.py
```diff
@@ -864,80 +807,57 @@ class AwqConfig(QuantizationConfigMixin):
         version (`AWQLinearVersion`, *optional*, defaults to `AWQLinearVersion.GEMM`):
             The version of the quantization algorithm to use. GEMM is better for big batch_size (e.g. >= 8) otherwise,
             GEMV is better (e.g. < 8 ). GEMM models are compatible with Exllama kernels.
-        backend (`AwqBackendPackingMethod`, *optional*, defaults to `AwqBackendPackingMethod.AUTOAWQ`):
+        backend (`AwqBackendPackingMethod`, *optional*, defaults to `AwqBackendPackingMethod.GPTQMODEL`):
```
SunMarc: update docstring

src/transformers/utils/quantization_config.py
```diff
@@ -864,80 +807,57 @@ class AwqConfig(QuantizationConfigMixin):
         version (`AWQLinearVersion`, *optional*, defaults to `AWQLinearVersion.GEMM`):
             The version of the quantization algorithm to use. GEMM is better for big batch_size (e.g. >= 8) otherwise,
             GEMV is better (e.g. < 8 ). GEMM models are compatible with Exllama kernels.
```
SunMarc: this is not used anymore so put in below in legacy param section for example.

tests/quantization/autoawq/test_awq.py
```diff
@@ -45,38 +41,38 @@
...
+    #     elif device_type == "xpu":
+    #         support_llm_awq = True
+    #
+    #     if support_llm_awq:
+    #         # LLMAWQ should work on an A100
+    #         AwqConfig(bits=4, backend="llm-awq")
+    #     else:
+    #         # LLMAWQ does not work on a T4
+    #         with self.assertRaises(ValueError):
+    #             AwqConfig(bits=4, backend="llm-awq")
```
SunMarc: update

tests/quantization/autoawq/test_awq.py
```diff
@@ -564,29 +305,28 @@ def test_load_quantized_model(self):
...
+#         tokenizer = AutoTokenizer.from_pretrained("TheBloke/TinyLlama-1.1B-Chat-v0.3-AWQ")
+#         input_ids = tokenizer.encode("How to make a cake", return_tensors="pt")
+#         pad_token_id = tokenizer.eos_token_id
+#         output = model.generate(input_ids, do_sample=False, max_length=20, pad_token_id=pad_token_id)
+#         print(tokenizer.decode(output[0], skip_special_tokens=True))
+#
+#         expected_output = (
+#             "How to make a cake with a round tin?\nHow to make a cake with a round tin?\n1. Preheat the oven to 180°"
+#         )
+#         self.assertIn(tokenizer.decode(output[0], skip_special_tokens=True), expected_output)
```
SunMarc: if this doesn't work yet, we can just skip for now

tests/quantization/gptq/test_gptq.py
```diff
@@ -177,36 +176,31 @@ def test_quantized_layers_class(self):
         Simple test to check if the model conversion has been done correctly by checking on
         the class type of the linear layers of the converted models
         """
-        if is_gptqmodel_available():
-            from gptqmodel.utils.importer import hf_select_quant_linear
+        if not is_gptqmodel_available():
+            self.skipTest("gptqmodel not available")
```
SunMarc: do not skip test here,  add a decorator instead

tests/quantization/gptq/test_gptq.py
```diff
@@ -243,29 +237,16 @@ def test_serialization(self):
...
-                    quant_type = "ipex" if is_ipex_available() else "torch"
-                else:
-                    # We expect tritonv2 to be used here, because exllama backend doesn't support packing https://github.com/ModelCloud/GPTQModel/issues/1354
-                    # TODO: Remove this once GPTQModel exllama kernels supports packing
-                    quant_type = "tritonv2"
-                quantized_model_from_saved = AutoModelForCausalLM.from_pretrained(
-                    tmpdirname, device_map=self.device_map
-                )
+                quant_type = "exllamav2"
+            # if self.quantized_model.config["quantization_config"]["format"] == ""
```
SunMarc: remove ?

SunMarc (commented):
Thanks, left some minor comments !

src/transformers/integrations/__init__.py
```diff
@@ -165,7 +164,6 @@
 if TYPE_CHECKING:
     from .aqlm import replace_with_aqlm_linear
     from .awq import (
-        fuse_awq_modules,
         post_init_awq_exllama_modules,
         post_init_awq_ipex_modules,
         replace_quantization_scales,
```
SunMarc, line 171: remove imports if we deleted them

SunMarc:
@bot /style

tests/quantization/autoawq/test_awq.py
```diff
@@ -144,7 +144,11 @@ def setUpClass(cls):
         Setup quantized model
         """
         cls.tokenizer = AutoTokenizer.from_pretrained(cls.model_name)
-        cls.quantized_model = AutoModelForCausalLM.from_pretrained(cls.model_name, device_map=cls.device_map)
+        # Use GEMM so that test_save_pretrained() writes out the quantized weights.
+        quantization_config = AwqConfig(backend=AwqBackend.GEMM)
+        cls.quantized_model = AutoModelForCausalLM.from_pretrained(
+            cls.model_name, device_map=cls.device_map, quantization_config=quantization_config
+        )
```
Qubitium, line 154: @SunMarc This part of the awq test was modified since previous test assumes kernel loading and weight save is 1-to-1 compatible. It is not as exllama/marlin kernels mutates the weights on load so they are in-effect, not packable/re-savable. To make the test pass, which expect quantized model to load, and then save, and thenreload/inference again, we need to specify the `gemm` kernel which does not mutates weights.

tests/quantization/autoawq/test_awq.py
Qubitium: Fixed.

Qubitium:
@SunMarc Since last review:

1. Unused awq properties (fuse related) removed.
2. Fixed commented out code related to IPEX and change it to test TorchFused kernel instead (ipex replacement)
3. Fixed hf awq kernel selection not passing `device_map` to `hf_select_quant_linear_v2`. Without device_map, it was selecting the wrong kernel since gpt-qmodel needs device info to return best kernel for hw. 

The PR currently depends on GPT-QModel 5.4.4 which is not yet released as we are working to resolve asap an internal regression related to gptq packing code: https://github.com/ModelCloud/GPTQModel/issues/2234

tests/quantization/autoawq/test_awq.py
SunMarc: Makes sense. Feel free to re-write / add some of the tests if you feel like they are way to outdated / can be simplified. For saving for example, I would expect that that we raise an error with one tries to save with exllama or marlin but it passes with `gemm`.

SunMarc:
LMK when we can merge !

Qubitium:
> LMK when we can merge !

We are performing final CI run in gpt-qmodel and once that passes. I will push 5.6.0 release asap to pypi so this PR can be ready for merge.

Qubitium:
@SunMarc  Ready. GPT-QModel v5.6.0 has been released with wheels currently building slowly for all the python/torch versions (may take a few hours): https://github.com/ModelCloud/GPTQModel/releases/tag/v5.6.0

docs/source/zh/llm_tutorial.md
```diff
@@ -261,7 +261,7 @@ LLMs是[仅解码器](https://huggingface.co/learn/nlp-course/chapter1/6?fw=pt)
 ### 延迟、吞吐量和内存利用率
 
 1. [指南](llm_tutorial_optimization),如何优化LLMs以提高速度和内存利用；
-2. [指南](main_classes/quantization), 关于`quantization`，如bitsandbytes和autogptq的指南，教您如何大幅降低内存需求。
+2. [指南](main_classes/quantization), 关于`quantization`，如bitsandbytes和GPT-QModeldel的指南，教您如何大幅降低内存需求。
```
MekkCyber:
```suggestion
2. [指南](main_classes/quantization), 关于`quantization`，如bitsandbytes和GPT-QModel的指南，教您如何大幅降低内存需求。
```

src/transformers/models/auto/tokenization_auto.py
```diff
@@ -389,6 +389,8 @@ def load_merges(merges_file):
 
 
 def tokenizer_class_from_name(class_name: str) -> Union[type[Any], None]:
+    if class_name == "BloomTokenizer":
+        return TokenizersBackend
```
MekkCyber: Not sure about this, the tokenizer should be fixed in a separate pr, and team members are already working on an other fix

MekkCyber (commented):
Thank you all for your work! I only have 2 nits

src/transformers/models/auto/tokenization_auto.py
ZX-ModelCloud: Yes, I see other PRs are fixing this issue. I've already reverted it.

This is a temporary fix; otherwise, I can't load the Bloom Tokenizer.

Qubitium: @MekkCyber  the CI test will fail without this `hack` (now reverted) so after CI tests run for this PR, the bloom model issue will fail unless the other (`Bloom` tokenizer prs by other hf team members) are merged first.

MekkCyber (approved):
Thank you for iterating lgtm

SunMarc (approved):
Thanks ! About gptq-model installation, would it be possible at some point to fetch the correct wheels depending on the user setup like torch using pip ? This is actually a quite a barrier for new users, or try to give more instructions ?

BenjaminBossan:
Just to be sure, this will be part of transformers v5?

Qubitium:
> Thanks ! About gptq-model installation, would it be possible at some point to fetch the correct wheels depending on the user setup like torch using pip ? This is actually a quite a barrier for new users, or try to give more instructions ? 

Right now the pip install script already auto download the precompiled gpt-qmodel whl (125mb) from gitbub releases, if we have the env python,torch,cuda version match. But yeah, it is painful to install if you have to compile from src ranging from 10-20 minutes for compile.

SunMarc:
> Just to be sure, this will be part of transformers v5?

Yes !

SunMarc:
> Right now the pip install script already auto download the precompiled gpt-qmodel whl (125mb) from gitbub releases, if we have the env python,torch,cuda version match. But yeah, it is painful to install if you have to compile from src ranging from 10-20 minutes for compile.

Okay, I must have a mismatched values hence it wasn't downloading the pre-compiled wheels. Is there a way to return a warning if this is not the case ?

Qubitium:
> 
> Okay, I must have a mismatched values hence it wasn't downloading the pre-compiled wheels. Is there a way to return a warning if this is not the case ? 

We missed building torch 2.9.1 whl for 5.6.0 so that is likely the cause. Will push 5.6.2 to resolve this other misc setup issues other users have reported plus logwarning when setup cannt match downloaded whl.

PR #41568, opened 2025-10-14
Use `yaml.safe_load`

ydshieh:
# What does this PR do?

Use `yaml.safe_load` instead of `yaml.load` to avoid potential security issue, and to close the security report we received.

https://pyyaml.org/wiki/PyYAMLDocumentation

> Warning: It is not safe to call yaml.load with any data received from an untrusted source! yaml.load is as powerful as pickle.load and so may call any Python function. Check the yaml.safe_load function though.

### DIFF
--- a/src/transformers/models/marian/convert_marian_to_pytorch.py
+++ b/src/transformers/models/marian/convert_marian_to_pytorch.py
@@ -105,7 +105,7 @@ def load_config_from_state_dict(opus_dict):
     import yaml
 
     cfg_str = "".join([chr(x) for x in opus_dict[CONFIG_KEY]])
-    yaml_cfg = yaml.load(cfg_str[:-1], Loader=yaml.BaseLoader)
+    yaml_cfg = yaml.safe_load(cfg_str[:-1], Loader=yaml.BaseLoader)
     return cast_marian_config(yaml_cfg)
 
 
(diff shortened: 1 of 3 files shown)


src/transformers/models/marian/convert_marian_to_pytorch.py
```diff
@@ -105,7 +105,7 @@ def load_config_from_state_dict(opus_dict):
     import yaml
 
     cfg_str = "".join([chr(x) for x in opus_dict[CONFIG_KEY]])
-    yaml_cfg = yaml.load(cfg_str[:-1], Loader=yaml.BaseLoader)
+    yaml_cfg = yaml.safe_load(cfg_str[:-1], Loader=yaml.BaseLoader)
```
Cyrilvallez, line 108: `safe_load` does not accept the `Loader` argument

Cyrilvallez (commented):
We cannot only do that as it's not equivalent (safe_load does not accept Loader arg). Also, according to yaml documentation, the BaseLoader and FullLoader that are passed here are not unsafe (https://github.com/yaml/pyyaml/issues/292), so I don't think we need to update anything!

ydshieh:
sorry for not checking the function arguments. I will ask the security reporter about the safeness. It seems depending on the versions.

ydshieh:
close as not a real issue

PR #41572, opened 2025-10-14
Gemma3 fixes

remi-or:
This PR fixes three things in `gemma3`:
- a multiple-device error where `torch.where` takes some of its coefficients from a tensor that is not on the right device and is a full_like, so we just replace it with the filling element
- an error in the `flash_attn_inference_equivalence` which is due to the model needing more parameters than are generated by defualt. To avoid this, we add a flag that specifies if we need to check the forward pass with training or not, and make this check default for both and left padding (cc. @vasqu )
- the test `flash_attn_from_config` was failing for the same reasons (`token_type_ids` is required as a model input when training) so I added a `.eval()` to avoid this. It does not seem the model needs to be in train mode for this test, but I can also add an option to the test to only call `.eval()` if a flag is passed


tests/test_modeling_common.py
```diff
@@ -3113,12 +3118,14 @@ def flash_attn_inference_equivalence(
...
-                    # Check it can run in training mode
-                    model.train()
-                    _ = model(**second_inputs)
                 else:
                     torch.testing.assert_close(logits_2_eager[:-1], logits_2_fa[:-1], atol=atol, rtol=rtol)
 
+                # Check it can run in training mode
+                if check_forward_in_train:
+                    model.train()
+                    _ = model(**second_inputs)
```
vasqu: Would it make more sense on the `flash_attn_from_config` test (with the new kwarg and making it a default to check for training)? This is still a bit weird to be in this test after another thought --> we only want to check inference equivalence tbh, the config test is more general and checks whether things break.

vasqu (approved):
Overall happy with the changes, imo we should move the training mode check to the config test instead + check with run-slow on our CIs just to be sure

tests/test_modeling_common.py
remi-or: Seems like a good idea -- that test is already doing a fwd in train mode. Changing it.

tests/test_modeling_common.py
```diff
@@ -3667,6 +3664,10 @@ def flash_attn_from_config(self, attn_implementation: str):
             fa_model = model_class._from_config(
                 config, attn_implementation=attn_implementation, dtype=torch.bfloat16
             ).to(torch_device)
+            if test_fwd_in_train:
+                fa_model = fa_model.train()
```
vasqu, line 3673: Can we add a small comment here to clarify that it is indeed different, e.g. dropout? Otherwise, lgtm

vasqu:
run-slow: gemma3

vasqu:
Even better than main CI :heart: feel free to merge after adding a small comment to why train vs eval

tests/test_modeling_common.py
```diff
@@ -3114,9 +3114,6 @@ def flash_attn_inference_equivalence(
                 torch.testing.assert_close(logits_1_eager, logits_1_fa, atol=atol, rtol=rtol)
                 if padding_side == "left":
                     torch.testing.assert_close(logits_2_eager[1:], logits_2_fa[1:], atol=atol, rtol=rtol)
-                    # Check it can run in training mode
-                    model.train()
-                    _ = model(**second_inputs)
```
Cyrilvallez, line 3119: Indeed from the name of the test it does not seem necesary

tests/test_modeling_common.py
Cyrilvallez: Actually, can we add another explaining that we set train mode because it's strictly harder than eval? I.e. if it works in train, it works in eval but not necessarily the other way around. Just because it's not obvious why we would set train mode here by default otherwise
Sorry for being annoying but did not get it from first glance

Cyrilvallez (approved):

PR #41575, opened 2025-10-14
Fixed : Bad error message for AutoTokenizer loading Voxtral #41553 

i3hz:
Fixes # 41553
Added a small check in both src/transformers/models/llama/tokenization_llama_fast.py and src/transformers/models/llama/tokenization_llama.py.

Who can review?

@ArthurZucker

### DIFF
--- a/src/transformers/masking_utils.py
+++ b/src/transformers/masking_utils.py
@@ -409,9 +409,13 @@ def sdpa_mask_recent_torch(
     # but without data-dependent slicing (i.e. torch.compile friendly)
     kv_arange = torch.arange(kv_length, device=cache_position.device)
     kv_arange += kv_offset
+    original_mask_function = mask_function
+    
     # Potentially add the padding 2D mask
     if padding_mask is not None:
         mask_function = and_masks(mask_function, padding_mask_function(padding_mask))
+        if original_mask_function is bidirectional_mask_function:
+            return padding_mask[:,None,None,:].expand(batch_size, 1, q_length, kv_length)
 
     batch_arange = torch.arange(batch_size, device=cache_position.device)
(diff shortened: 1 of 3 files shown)


Rocketknight1:
This PR should probably not be deleting those CUDA kernels :sweat_smile:

i3hz:
Yeah that's weird I noticed that too ,but I didn't really do anything that would cause that .
https://github.com/huggingface/transformers/pull/41575/commits/90af495b8c33360f92b3fd5f79aba50f6e34a82a
This seems to be the commit that removed the kernels

Rocketknight1:
Let me try a rebase, one sec.

Rocketknight1:
Okay, now that I can actually review the code, I'm even more confused! Why should the Llama tokenizers require `mistral-common`? They don't use it.

i3hz:
Yeah that's very confusing . But I assumed since they have added model_type : llama in the [config](https://huggingface.co/mistralai/Voxtral-Mini-3B-2507/blob/main/config.json) they probably use the same tokenizer .

jackzhxng (commented):
The model type is llama, but the tokenizer used is a specific format called tekken (tekken.json in the repo), which mistral-common provides a tokenizer for. The error message is confusing because the stack trace is also misleading when `mistral-common` isn't installed, since I don't think it should be going through the llama tokenizer at all

i3hz:
Soo what exactly should I look into rn? I'm just confused lol

Rocketknight1:
I'll stop trying to resolve this and just get the people who actually know about tokenizers :sweat_smile:

cc @arthurzucker @itazap

vasqu:
Copy pasting a bit but let's coordinate https://github.com/huggingface/transformers/issues/41553#issuecomment-3406606156

PR #41577, opened 2025-10-14
[kernels] refactor function kernel calling

MekkCyber:
# What does this PR do?

This should simplify lazy kernel loading in Transformers.  
We simply define a mapping between each kernel name and the repository it should be pulled from, then load it using the `lazy_load_kernel` function. This function adds the kernel to a global cache shared across all models.  
If the kernel isn’t available, we check whether it’s installed as a module for backward compatibility; otherwise, we return `None`.

### DIFF
--- a/src/transformers/integrations/hub_kernels.py
+++ b/src/transformers/integrations/hub_kernels.py
@@ -14,11 +14,16 @@
 import re
 from collections.abc import Callable
 from functools import partial
+from types import ModuleType
 from typing import Optional, Union
 
 from ..modeling_flash_attention_utils import lazy_import_flash_attention
+from ..utils import logging
+from ..utils.import_utils import is_kernels_available
 from .flash_attention import flash_attention_forward
 
+
+logger = logging.get_logger(__name__)
 
 try:
     from kernels import (
@@ -158,6 +163,13 @@ except ImportError:
         raise RuntimeError("register_kernel_mapping requires `kernels` to be installed. Run `pip install kernels`.")
 
 
+_KERNEL_SIMPLE_MAPPING: dict[str, str] = {
+    "causal-conv1d": "kernels-community/causal-conv1d",
+}
+
+_KERNEL_MAPPING_ACROSS_MODELS: dict[str, Optional[ModuleType]] = {}
+
+
 def is_kernel(attn_implementation: Optional[str]) -> bool:
     """Check whether `attn_implementation` matches a kernel pattern from the hub."""
     return (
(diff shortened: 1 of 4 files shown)


src/transformers/integrations/hub_kernels.py
```diff
@@ -220,9 +232,54 @@ def load_and_register_attn_kernel(attn_implementation: str, attention_wrapper: O
     ALL_MASK_ATTENTION_FUNCTIONS.register(attn_implementation, ALL_MASK_ATTENTION_FUNCTIONS["flash_attention_2"])
 
 
+def lazy_load_kernel(kernel_name: str, mapping: dict[str, Optional[ModuleType]]):
+    if kernel_name in mapping and isinstance(mapping[kernel_name], ModuleType):
```
MekkCyber: the main utility function applied to the case of `causal-conv1d`

MekkCyber:
run-slow: falcon_mamba, mamba

src/transformers/integrations/hub_kernels.py
```diff
@@ -158,6 +163,13 @@ def register_kernel_mapping(*args, **kwargs):
         raise RuntimeError("register_kernel_mapping requires `kernels` to be installed. Run `pip install kernels`.")
 
 
+_KERNEL_SIMPLE_MAPPING: dict[str, str] = {
```
ArthurZucker: why "simple" ?

src/transformers/models/falcon_mamba/modular_falcon_mamba.py
```diff
@@ -258,7 +256,12 @@ def rms_forward(hidden_states, variance_epsilon=1e-6):
 
 class FalconMambaMixer(MambaMixer):
     def warn_slow_implementation(self):
-        causal_conv1d_update, causal_conv1d_fn = _lazy_load_causal_conv1d()
+        causal_conv1d = lazy_load_kernel("causal-conv1d", _KERNEL_MAPPING_ACROSS_MODELS)
+        causal_conv1d_update, causal_conv1d_fn = (
+            (causal_conv1d.causal_conv1d_update, causal_conv1d.causal_conv1d_fn)
+            if causal_conv1d is not None
+            else (None, None)
+        )
```
ArthurZucker: we should not need to pass the kernel mapping, lazy load kernel can handle it!

ArthurZucker (commented):
ncie

src/transformers/models/falcon_mamba/modular_falcon_mamba.py
MekkCyber: Done

src/transformers/integrations/hub_kernels.py
```diff
@@ -220,9 +231,53 @@ def load_and_register_attn_kernel(attn_implementation: str, attention_wrapper: O
     ALL_MASK_ATTENTION_FUNCTIONS.register(attn_implementation, ALL_MASK_ATTENTION_FUNCTIONS["flash_attention_2"])
 
 
+def lazy_load_kernel(kernel_name: str, mapping: dict[str, Optional[ModuleType]] = _KERNEL_MODULE_MAPPING):
```
ArthurZucker, line 234: should be None, if none default to kernel mapping python will cry otherwise!

src/transformers/models/falcon_mamba/modeling_falcon_mamba.py
```diff
@@ -268,7 +240,12 @@ def __init__(self, config: FalconMambaConfig, layer_idx: int):
         self.rms_eps = config.mixer_rms_eps
 
     def warn_slow_implementation(self):
-        causal_conv1d_update, causal_conv1d_fn = _lazy_load_causal_conv1d()
+        causal_conv1d = lazy_load_kernel("causal-conv1d")
+        causal_conv1d_update, causal_conv1d_fn = (
+            (causal_conv1d.causal_conv1d_update, causal_conv1d.causal_conv1d_fn)
+            if causal_conv1d is not None
+            else (None, None)
+        )
```
ArthurZucker, line 248: nice and simple

ArthurZucker (approved):
Thanks 🤗

PR #41578, opened 2025-10-14
Fix import in quantization 

SunMarc:
# What does this PR do?

This PR fixes a small import that was introduced in https://github.com/huggingface/transformers/issues/41445.


SunMarc:
fixed by https://github.com/huggingface/transformers/pull/41571

PR #41580, opened 2025-10-14
Refactor weight loading

ArthurZucker:
# CORE REFACTORING, loading, converting, logging
More helpful debugging report when loading weights
image

If you just want to fuse qkv: 
image

It can. You just need to make sure you change the model code and pouf! 
```python
            WeightConverter(
                ["self_attn.q_proj", "self_attn.k_proj", "self_attn.v_proj"],
                "self_attn.qkv_proj",
                operations=[Concatenate(dim=0)],  # more like stack?
            ),
```

For deepseek we will embed the rope permute:
```python
            WeightConverter(
                ["self_attn.q_proj", "self_attn.k_proj", "self_attn.v_proj"],
                operations=[RopePermute()],  # more like stack?
            ),
```
# `WeightConverter` API:

The API allows you to define a mapping using `WeightConverter`. You can define many to one source/target keys, quantization opérations and distributed opérations along with normal opérations. For now `MergeModuleLIst` and `Concatenate`, will add the `RopePermute` one soon. 

```python 
_checkpoint_conversion_mapping = {
    "mixtral": [
        WeightConverter(
            source_keys=[
                "mlp.experts.*.w1.weight",
                "mlp.experts.*.w3.weight",
            ],
            target_keys="mlp.experts.gate_up_proj",
            operations=[MergeModulelist(dim=0), Concatenate(dim=1)],
        ),
        WeightConverter(
            source_keys=["mlp.experts.*.w2.weight"],
            target_keys="mlp.experts.down_proj",
            operations=[MergeModulelist(dim=0)],
        ),
    ],
}
```

We use to have this:

https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L4545-L4568

But now its just explicit:

```python
        "legacy": [
            WeightConverter(
                source_keys="LayerNorm.gamma",
                target_keys="LayerNorm.weight",
            ),
            WeightConverter(
                source_keys="LayerNorm.beta",
                target_keys="LayerNorm.bias",
            ),
        ],
    }
    if hasattr(torch.nn.utils.parametrizations, "weight_norm"):
        mapping["legacy"] += [
            WeightConverter(
                source_keys="weight_g",
                target_keys="parametrizations.weight.original0",
            ),
            WeightConverter(
                source_keys="weight_v",
                target_keys="parametrizations.weight.original1",
            ),
        ]
    else:
        mapping["legacy"] += [
            WeightConverter(
                source_keys="parametrizations.weight.original0",
                target_keys="weight_g",
            ),
            WeightConverter(
                source_keys="parametrizations.weight.original1",
                target_keys="weight_v",
            ),
        ]
```
and its faster cuz we don't iterate over the whole checkpoint

The `core` logic is:
Iterate over all of the dict keys:
1. collect the keys that match the glob patterns from all source keys (you pipe the ones that are from the same weight converter): `(mlp.experts.*.gate_proj.weight|mlp.experts.*.up_proj.weight)` into a dict with key target key

This produces:
```
{ 
"mlp.experts.gate_up_proj" : 
    {"mlp.experts.*.w1.weight":
        { "mlp.experts.0.w1.weight": [t0, t1, t2, etc], "mlp.experts.1.w1.weight": [t0, t1, t2, etc]},
     "mlp.experts.*.w3.weight":
        { "mlp.experts.0.w3.weight": [t0, t1, t2, etc], "mlp.experts.1.w3.weight": [t0, t1, t2, etc]},
    }
  ....
}
```
We need to keep track of which layers were collected, and from which source pattern. 

1bis. Schedule tensor materialization, without blocking the GIL (as this takes the most amount of time). We distribute the tensor at this stage, before any operations. This IS the trickiest. We do this during collection to not waste time.

2. We collect the results of materialization, and we apply the `operations` on all the collected values (at this point ` { "mlp.experts.0.w1.weight": [t0, t1, t2, etc], "mlp.experts.1.w1.weight": [t0, t1, t2, etc]}.values()` gives a list of lists. 
3. We create a `dict` with the `target_key` and the output `values`. We pass this to the quantizer
4. We quantize the input tensors, outputting the final dict. 
5. We set the param into the model. 

Keys are handled a lot better!

## Enable MoE quantization for FP8 

This script does not work on main
```python
import torch
from transformers import MixtralForCausalLM, AutoTokenizer, FineGrainedFP8Config
import time 
quantization_config = FineGrainedFP8Config(modules_to_not_convert=["model.layers.*.mlp.gate"])
model = MixtralForCausalLM.from_pretrained("mistralai/Mixtral-8x7B-v0.1", quantization_config=quantization_config, tp_plan="auto")
```

## Enable TP + MoE without OOM
This script does not work on main
```python
model = MixtralForCausalLM.from_pretrained("mistralai/Mixtral-8x7B-v0.1", tp_plan="auto")
```

## Enable `device_map="auto"` + MoE + FP8

This script does not work on main
```python
quantization_config = FineGrainedFP8Config(modules_to_not_convert=["model.layers.*.mlp.gate"])
model = MixtralForCausalLM.from_pretrained("mistralai/Mixtral-8x7B-v0.1", quantization_config=quantization_config, device_map="auto")
```

### Uses staging buffers per conversion op

- [x] 4x speedup with `device_map="auto"`
- [x] Full `MoE` quantization with FP8

TODOS:
- [x] Test with TP / EP
- [x] Add TQDM! 
- [ ] Test with deepspeedd
- [ ] Test with loras and peft
- [ ] Test with vllm backend
- [ ] Test with fsdp
- [ ] Add saving

Script: 
```python
import torch
from torch import nn
from transformers import MixtralForCausalLM, AutoTokenizer

import time 
start = time.time()
model = MixtralForCausalLM.from_pretrained("mistralai/Mixtral-8x7B-v0.1", device_map="auto")
end = time.time() 
print("loading took ", end-start)
tokenizer = AutoTokenizer.from_pretrained("mistralai/Mixtral-8x7B-v0.1")
inputs = tokenizer("hey how are you?", return_tensors="pt").to(model.device)
out = model.generate(**inputs, max_new_tokens=16)
print(tokenizer.batch_decode(out))
```
```bash
loading took  14.271092891693115
['<s> hey how are you?\n\nI am a 20 year old male and I have been having']
```
⬆️  is with: merge modulelist, concat gate_up 
⬇️ is naive loading.
```bash
loading took  54.271092891693115
['<s> hey how are you?\n\nI am a 20 year old male and I have been having']
```


src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    Base class with a reusable buffer to avoid repeated allocations.
+    Subclasses implement `convert(collected_tensors) -> torch.Tensor` and
+    write results into a view of `self._buffer`.
+    """
+
+    target_tensor_shape: torch.Tensor
+    can_be_quantized: bool = True
+    can_be_distributed: bool = False
+
```
ArthurZucker, line 115: revers_op : ConversionOps

src/transformers/conversion_mapping.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# Either we keep it here, or we move it to the config, but for newcomers, seeing this is kinda weird no?
+
+from ...core_model_loading import Fuse, MergeModuleList, WeightConversion, ConversionType
+
+_checkpoint_conversion_mapping = { "mixtral": {
+    "experts.*.(w1|w2).weight$": WeightConversion(
+        "experts.gate_up_proj.weight", [ConversionType.MERGE_MODULE_LIST, ConversionType.FUSE]
+    ),
+    "self_attn.(q|k|v)_proj": WeightConversion("self_attn.qkv_proj", ConversionType.FUSE),
```
ArthurZucker: TODO: generate a module replacement scheme out of this?

src/transformers/conversion_mapping.py
```diff
@@ -0,0 +1,17 @@
...
+#
+#
+#
+#
+# Either we keep it here, or we move it to the config, but for newcomers, seeing this is kinda weird no?
+
+from ...core_model_loading import Fuse, MergeModuleList, WeightConversion, ConversionType
+
+_checkpoint_conversion_mapping = { "mixtral": {
+    "experts.*.(w1|w2).weight$": WeightConversion(
```
ArthurZucker:
```suggestion
    "experts.*.(w1|w3).weight$": WeightConversion(
```

src/transformers/conversion_mapping.py
```diff
@@ -0,0 +1,77 @@
...
+from .core_model_loading import Concatenate, MergeModulelist, WeightConverter
+
+
+_checkpoint_conversion_mapping = {
+    "mixtral": [
+        WeightConverter(
+            source_keys=[
+                "block_sparse_moe.experts.*.w1.weight",
+                "block_sparse_moe.experts.*.w3.weight",
+            ],  # you give me a list of 2 keys, I collect a list of tensors
```
LysandreJik:
```suggestion
            ],  # you give me a list of 2 keys, I collect a list of a list of tensors
```

src/transformers/conversion_mapping.py
```diff
@@ -0,0 +1,77 @@
...
+_checkpoint_conversion_mapping["glm4v_moe"] = _checkpoint_conversion_mapping["qwen2_moe"].copy()
+_checkpoint_conversion_mapping["jamba"] = _checkpoint_conversion_mapping["qwen2_moe"].copy()
+_checkpoint_conversion_mapping["lfm2_moe"] = _checkpoint_conversion_mapping["mixtral"].copy()
+_checkpoint_conversion_mapping["long_cat_flash"] = _checkpoint_conversion_mapping["qwen2_moe"].copy()
+_checkpoint_conversion_mapping["qwen3_moe"] = _checkpoint_conversion_mapping["qwen2_moe"].copy()
+_checkpoint_conversion_mapping["qwen3_omni_moe"] = _checkpoint_conversion_mapping["qwen2_moe"].copy()
+_checkpoint_conversion_mapping["qwen3_next"] = _checkpoint_conversion_mapping["qwen2_moe"].copy()
+_checkpoint_conversion_mapping["qwen3_vl_moe"] = _checkpoint_conversion_mapping["qwen2_moe"].copy()
+_checkpoint_conversion_mapping["hunyuan_v1_moe"] = _checkpoint_conversion_mapping["qwen2_moe"].copy()
+_checkpoint_conversion_mapping["minimax"] = _checkpoint_conversion_mapping["mixtral"].copy()
```
LysandreJik: there's maybe a cleaner way to handle the above rather than all the `.copy()`, but for a future PR

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
+# coding=utf-8
+# Copyright 2024 The HuggingFace Inc. team. All rights reserved.
```
LysandreJik:
```suggestion
# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
```

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+    _FP8_DTYPE = torch.int8
+    _FP8_MIN, _FP8_MAX = -127, 127
+    _FP8_IS_INT = True
+    logger.warning_once(
+        "torch.float8_e4m3fn not available; falling back to int8 emulation for Fp8Quantize operations."
+    )
+
+try:
+    from torch.profiler import ProfilerActivity
+    from torch.profiler import profile as torch_profile
```
LysandreJik: AFAIK, available in torch 2.3 onwards. We have a lower bound on torch 2.2, should we update the lower bound or is it not mandatory to have it?

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+    *,
+    digits_only: bool = True,
+    allow_prefix: bool = True,
+) -> tuple[re.Pattern, dict[str, str]]:
+    """
+    Build one compiled regex alternation with a named group per glob.
+    - digits_only: '*' => digits only (\\d+) if True, else any chars (.+)
+    - allow_prefix: if True, allow arbitrary prefix before the pattern
+                    (keeps '$' so we still require a full suffix match)
+    Returns (compiled_regex, name->glob map).
```
LysandreJik: Possible to put an example of what it does?

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+    Replace each '*' in 'pattern' with the next value from 'star_values' (in order).
+    """
+    it = iter(star_values)
+    out = []
+    for ch in pattern:
+        if ch == "*":
+            out.append(str(next(it)))
+        else:
+            out.append(ch)
+    return "".join(out)
```
LysandreJik: If you want to maximize understandability, I'd put a single example for each of those methods so that they may be reused by others in different contexts down the road

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+        profile: bool = False,
+    ) -> Any:
+        """
+        Execute the conversion while measuring runtime and optionally profiling the call.
+        """
+        start = time.perf_counter()
+        result = self.convert(value)
+        elapsed = time.perf_counter() - start
+        if profile:
+            print(elapsed)
```
LysandreJik: possible to use the centralized logger instead?

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+        return result
+
+
+class Chunk(ConversionOps):
```
LysandreJik, line 149: Love those atomic ops
Probably simple to add a `revert_convert` if you want to have a bijection

LysandreJik: (ah, just saw the `_inverse_op`, cool too)

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+
+    def __post_init__(self):
+        if not isinstance(self.source_keys, list):
+            self.source_keys = [self.source_keys]
+        if not isinstance(self.target_keys, list):
+            if self.target_keys is None:
+                self.target_keys = self.source_keys
+            else:
+                self.target_keys = [self.target_keys]
+        self._regex_pat = build_glob_alt(self.source_keys)
```
LysandreJik: TODO: add some sanity checks:
- Make sure that if multiple source keys are sent, the target key cannot be a list and has to be solo
- The reverse for multiple target keys
- The target keys and source keys must all be globs

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+
+
+def set_param_for_module(
+    model, k, v, meta_model_state_dict, empty_tensor, mismatch_keys, missing_keys, misc, distributed_operation
+):
```
LysandreJik, line 507: TODO: please document
Please make sure the user is aware of the in-place dict/object updates that are happening here (invisible rn)

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+
+
+def spawn_materialize(EXEC, _file_sems, file_id, t) -> Future:
+    sem = _file_sems[file_id]
+
+    def _job():
+        with sem:
+            return _materialize_copy(t)
+
+    return EXEC.submit(_job)
```
LysandreJik: black magic 🤯

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+    unexpected_keys = set()
+    # Global executor + per-file semaphores
+    EXEC = ThreadPoolExecutor(max_workers=GLOBAL_WORKERS)
+    _file_sems = defaultdict(lambda: threading.Semaphore(PER_FILE_LIMIT))
+
+    _patterns = list(itertools.chain.from_iterable([k.source_keys for k in weight_mapping]))
+    source_to_target = {sk: k for k in weight_mapping for sk in k.source_keys}
+    weight_pattern_alt, weight_pattern_by_group_name = build_glob_alt(_patterns)
+    tp_plan_alt, tp_plan_by_group_name = build_glob_alt(list(tp_plan.keys()))
+    dtype_policy_alt, dtype_policy_by_group_name = build_glob_alt(list(keep_in_dtype.keys()))
```
LysandreJik: maybe some quick inline comments of what the inside looks like could help understand what is happening

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+
+        entry.collected_tensors[target_key].setdefault(converter_key, []).append(fut)
+        for t in target_key.split("|"):
+            empty_tensor = meta_model_state_dict.get(t)
+            if empty_tensor is None:
+                unexpected_keys.add(t)
+                continue
+            if quantizer is not None and quantizer.param_needs_quantization(model, t):
+                # converter.quantization_operation[target_key] = quantizer.quantize_tensor
+                converter.quantization_operation[t] = Fp8Quantize()
```
LysandreJik: If the key doesn't exist, you can skip materialization earlier

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+                    values = [[k.result() for k in inner] for inner in tensors_for_this_layer.values()]
+
+                    if op := converter.distributed_operation:
+                        try:
+                            values = op(values)
+                        except Exception as e:
+                            misc[layer_name] = (
+                                f"Failed to apply {converter.distributed_operation.__class__.__name__}: {e}"
+                            )
+                            continue
```
LysandreJik:
```suggestion
```

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,853 @@
...
+                    for op in operations:
+                        try:
+                            values = op(values)
+                        except Exception as e:
+                            misc[layer_name] = (
+                                f"{e}\nError: {op.__class__.__name__} on tensors collected from {converter.source_keys}. Ckpt contains: {values}"
+                            )
+
+                    values = [values] if not isinstance(values, list) else values
+                    realized_value = {k: t for k, t in zip(concrete_target_keys, values) if k not in unexpected_keys}
```
LysandreJik: Arthur's self-review: might need a try-except here

src/transformers/modeling_utils.py
```diff
@@ -3829,7 +3721,8 @@ def save_pretrained(
 
             if safe_serialization:
                 # At some point we will need to deal better with save_function (used for TPU and other distributed
-                # joyfulness), but for now this enough.
+                # joyfulness), but for now this enough. # TODO: we should def parallelize this we are otherwise just waiting
+                # too much before scheduling the next write when its on a different
```
LysandreJik:
```suggestion
                # too much before scheduling the next write when its in a different file
```

src/transformers/modeling_utils.py
```diff
@@ -4772,46 +4592,79 @@ def _load_pretrained_model(
...
+                        device = device_map[key]
+                    else:
+                        device = device_map[""]
+                        if isinstance(device, torch.device):
+                            device = device.index  # safetensors only
+                    file_pointer = safe_open(
+                        os.path.join(checkpoint_files[0].rsplit("/", 1)[0], v), framework="pt", device=device
+                    )
+                    all_pointer.add(file_pointer)
+                    merged_state_dict[k] = (v, file_pointer.get_slice(k))  # don't meterialize yet
```
LysandreJik:
```suggestion
                    merged_state_dict[k] = (v, file_pointer.get_slice(k))  # don't materialize yet
```

LysandreJik (approved):
Impressive effort

src/transformers/modeling_utils.py
```diff
@@ -467,11 +467,39 @@ def _end_ptr(tensor: torch.Tensor) -> int:
...
+        return result
+    if isinstance(value, Iterable):
+        return list(value)
+    return [value]
+
+
+def _extract_tied_value_names(tied_weight_keys) -> list[str]:
+    if tied_weight_keys is None:
+        return []
+    if isinstance(tied_weight_keys, dict):
```
BenjaminBossan: This bit is confusing. AFAICT, `_extract_tied_value_names` is only used in one place, namely below in `_get_tied_weight_keys`:

`value_names = _extract_tied_value_names(list(module._tied_weights_keys.keys()))`

Therefore, the input will always be a list of strings (keys). So why check for `dict` here? Was this supposed to also extract the values?

Moreover, `_as_list` seems to be unnecessary if we only ever pass a list of strings.

ArthurZucker: Will remove it don't worry

BenjaminBossan: Okay. So the intent is to return only the keys (i.e. the targets), is that right?

src/transformers/modeling_utils.py
```diff
@@ -3505,6 +3300,10 @@ def save_pretrained(
                 For backward compatibility with PEFT library, in case adapter weights are attached to the model, all
                 keys of the state dict of adapters needs to be prepended with `base_model.model`. Advanced users can
                 disable this behaviours by setting `save_peft_format` to `False`.
+            save_original_format (`bool`, *optional*, defaults to `True`):
```
kylesayrs, line 3141: Docstring says default is True, kwargs says default is False

src/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -71,10 +71,10 @@ def __init__(self, config):
         self.num_experts = config.num_local_experts
         self.hidden_size = config.hidden_size
         self.expert_dim = self.intermediate_size
-        self.gate_up_proj = nn.Parameter(torch.empty(self.num_experts, self.hidden_size, 2 * self.expert_dim))
-        self.gate_up_proj_bias = nn.Parameter(torch.empty(self.num_experts, 2 * self.expert_dim))
+        self.gate_up_proj = nn.Parameter(torch.zeros(self.num_experts, self.hidden_size, 2 * self.expert_dim))
```
kylesayrs, line 74: Curious why these are zeros and not empty?

kylesayrs (commented):
This is really awesome, thanks for the amazing work here.

src/transformers/modeling_utils.py
kylesayrs, line 3141: Would you consider writing this field to the checkpoint config? What's the plan for loading models in the converted format?

ArthurZucker: Yeah it can make sense!

src/transformers/models/gpt_oss/modeling_gpt_oss.py
ArthurZucker: This will be updated I think, zeros is not good it allocates memory

src/transformers/modeling_utils.py
ArthurZucker: Will be true!

docs/source/en/perf_infer_gpu_multi.md
```diff
@@ -149,7 +149,7 @@ The example below packs `up_proj` and `gate_proj` into a single `gate_up_proj` m
 ```python
 class Llama4TextExperts(nn.Module):
     ...
-    self.gate_up_proj = nn.Parameter(torch.empty(self.num_experts, self.hidden_size, 2 * self.expert_dim))
+    self.gate_up_proj = nn.Parameter(torch.zeros(self.num_experts, self.hidden_size, 2 * self.expert_dim))
```
vasqu, line 152: Is there a reason we do zeros? Might make sense to have 1s instead? 

Ig this is tied to not using init weights

report.json
vasqu, line 1: The last bastion :joy:

src/transformers/conversion_mapping.py
```diff
@@ -0,0 +1,141 @@
...
+                    MergeModulelist(
+                        dim=0
+                    ),  # each process has two lists of tensors, we cat each list. -> we end up with 2 tensors
+                ],  # we want the loading to add this shard operation here. Though we can't shard after concats and merge, needs to be first
+            ),
+            # WeightConverter(
+            #     ["self_attn.q_proj", "self_attn.k_proj", "self_attn.v_proj"],
+            #     "self_attn.qkv_proj",
+            #     operations=[Concatenate(dim=0)],  # more like stack?
+            # ),
```
vasqu, line 57: Fused qkv? Imo, this might be a cool feature for another PR to allow fused qkv in general - it does have perf boosts

src/transformers/conversion_mapping.py
```diff
@@ -0,0 +1,141 @@
...
+    mapping["glm4v_moe"] = mapping["qwen2_moe"].copy()
+    mapping["jamba"] = mapping["qwen2_moe"].copy()
+    mapping["lfm2_moe"] = mapping["mixtral"].copy()
+    mapping["long_cat_flash"] = mapping["qwen2_moe"].copy()
+    mapping["qwen3_moe"] = mapping["qwen2_moe"].copy()
+    mapping["qwen3_omni_moe"] = mapping["qwen2_moe"].copy()
+    mapping["qwen3_next"] = mapping["qwen2_moe"].copy()
+    mapping["qwen3_vl_moe"] = mapping["qwen2_moe"].copy()
+    mapping["hunyuan_v1_moe"] = mapping["qwen2_moe"].copy()
+    mapping["minimax"] = mapping["mixtral"].copy()
```
vasqu, line 124: nit: maybe make this a private constant outside of the function? meaning the affected models list basically

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,735 @@
...
+        merged = []
+        for group in value:
+            if not isinstance(group, Sequence) or len(group) == 0:
+                raise ValueError("MergeModulelist requires non-empty sub-sequences.")
+            group = [k for k in group if k.ndim]
+            merged.append(torch.stack(group, dim=self.dim))
+        return merged
+
+
+class SplitModulelist(ConversionOps):
```
vasqu, line 213: Ig this is what I can use to split a list of experts into two separate ones, i.e. what I need for ernie vl :smile:

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,735 @@
...
+class Concatenate(ConversionOps):
+    """Concatenate tensors along `dim` using a reusable buffer."""
+
+    reverse_op: type[ConversionOps]
+
+    def __init__(self, dim: int = 0):
+        self.dim = dim
+        self.reverse_op = Chunk
+
+    @torch.no_grad
```
vasqu, line 180: Seeing this here and there, should it be used for all ops?

And it should be `@torch.no_grad()` with the (). Unsure if the decorator will work without

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,735 @@
...
+        result: list[list[torch.Tensor]] = []
+        for tensor, split_sizes in zip(value, self.sizes):
+            if not isinstance(tensor, torch.Tensor):
+                raise TypeError("SplitModulelist can only split torch.Tensor instances.")
+            splits = torch.split(tensor, split_sizes, dim=self.dim)
+            result.append(list(splits))
+        return result
+
+
+class PermuteForRope(ConversionOps):
```
vasqu, line 239: Yes, yes, yes

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,735 @@
...
+        return self
+
+    def __irshift__(self, *args, **kwargs):
+        return self
+
+    return LoadedParam
+
+
+def _materialize_copy(tensor, dtype=None):
+    tensor = tensor[...]
```
vasqu, line 431: I feel like a small comment is warranted here why we do this feels a bit magicy

src/transformers/core_model_loading.py
```diff
@@ -0,0 +1,735 @@
...
+        else:
+            missing_keys.discard(layer_name)
+            param_value._is_hf_initialized = True  # super important otherwise _init_weight re-initi if bias is missing
+            setattr(module_obj, param_name, param_value)
+
+
+class SkipLayer(Exception):
+    """Control-flow sentinel: abort processing of the current layer only."""
+
+    pass
```
vasqu, line 545: Maybe adding the layer name to the exception? Or do we handle this elsewhere?

src/transformers/core_model_loading.py
vasqu: Seeing this elsewhere, so maybe I'm out of the loop

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -424,3 +568,133 @@ def replace_with_fp8_linear(
         )
 
     return model
+
+
+class QuantizationOp(ConversionOps):
+    """Base class for quantization operations."""
+
+    pass
```
vasqu, line 576: would move this elsewhere but I think you are aware as the other quants are a TODO

vasqu (commented):
Just a few comments from the first look. Too much to go through at once but it looks super clean especially the conversion ops concept is neat!

xenova:
Very cool! 🔥 After pulling latest changes from main and trying to load gpt-oss-20b, I get this error:
```py
>>> from transformers import AutoModelForCausalLM
>>> model = AutoModelForCausalLM.from_pretrained("openai/gpt-oss-20b")
Unrecognized keys in `rope_parameters` for 'rope_type'='yarn': {'truncate'}
Unrecognized keys in `rope_parameters` for 'rope_type'='yarn': {'truncate'}
Loading weights: 100%|████████████████████████████████████████████████████████████████████████████████████████| 459/459 [00:00<00:00, 574.18it/s, Materializing param=lm_head.weight]
GptOssForCausalLM LOAD REPORT from: openai/gpt-oss-20b
Key                                                   | Status     | 
------------------------------------------------------+------------+-
model.layers.{0...23}.mlp.experts.down_proj_blocks    | UNEXPECTED | 
model.layers.{0...23}.mlp.experts.gate_up_proj_scales | UNEXPECTED | 
model.layers.{0...23}.mlp.experts.gate_up_proj_blocks | UNEXPECTED | 
model.layers.{0...23}.mlp.experts.down_proj_scales    | UNEXPECTED | 
model.layers.{0...23}.mlp.experts.gate_up_proj        | MISSING    | 
model.layers.{0...23}.mlp.experts.down_proj           | MISSING    | 

Notes:
- MISSING       :those params were newly initialized because missing form the checkpoint. Consider training on your downstream task.
```

Just flagging as it seems to break backwards compatibility. I can also confirm that checking out the 2nd last commit (i.e., without this change) does not result in the error.

ArthurZucker:
It won't break, @MekkCyber and @SunMarc are working on MXFp4 support!

MekkCyber:
Yes @xenova we are taking care of that here : https://github.com/huggingface/transformers/pull/42070, we just need to fix some issues and it will be good to go

src/transformers/models/qwen2_moe/modeling_qwen2_moe.py
```diff
@@ -289,66 +289,80 @@ def forward(
...
-    ModuleList of experts.
-    """
+class Qwen2MoeExperts(nn.Module):
+    """Collection of expert weights stored as 3D tensors."""
 
     def __init__(self, config):
         super().__init__()
         self.num_experts = config.num_experts
-        for _ in range(config.num_experts):
-            self.append(Qwen2MoeMLP(config, intermediate_size=config.moe_intermediate_size))
```
fxmarty-amd, line 301: This change is not straightforward and breaks downstream libraries expecting `Qwen2MoeExperts` experts to be `nn.Linear`. Is there an easy workaround?

src/transformers/models/qwen3_moe/modeling_qwen3_moe.py
```diff
@@ -209,61 +209,77 @@ def forward(self, x):
...
-    """
+class Qwen3MoeExperts(nn.Module):
+    """Collection of expert weights stored as 3D tensors."""
 
-    def __init__(self, config: Qwen3MoeConfig):
+    def __init__(self, config):
         super().__init__()
         self.num_experts = config.num_experts
-        for _ in range(self.num_experts):
-            self.append(Qwen3MoeMLP(config, intermediate_size=config.moe_intermediate_size))
```
fxmarty-amd, line 221: same comment

fxmarty-amd:
image

🫠

fxmarty-amd:
Just for my understanding - is this expected to land in 4.58?

ArthurZucker:
@fxmarty-amd this is v5!

rkazants:
Hi @ArthurZucker,

Can you please tell - why did you combine weights (like down_proj, up_proj) into one tensor for Qwen3-next (I think it relates to other models)? Before these changes they were in unsplit form - down_proj per each expert?

Best regards,
Roman

ArthurZucker:
@rkazants hey! the modules list were indeed split per experts, preventing the usage of optimized kernels

IlyasMoutawwakil:
This is actually super cool and could allow for batched experts inference which is traceable/exportable and faster in some cases when memory is not an issue ! Great work @ArthurZucker ! is there a plan for enabling pure-pytorch batched experts ? I can imagine something like `moe_implementation`  which can be `sequential`/`batched` ?

ArthurZucker:
Yes! cc @3outeille if he has time, but if you want to tackle it you should! (torch._bmm they have a native op now)

3outeille:
@IlyasMoutawwakil yes definitely, how time sensitive is it ?

IlyasMoutawwakil:
@3outeille great ! nothing time sensitive (since we are patching the MoEs in optimum-onnx/optimum-intel for now) but would make life much easier to control this behavior with a single argument. lmk if you would like to tackle this !

PR #41585, opened 2025-10-14
[Trainer] [Breaking change] `use_cache` default to `False`

SunMarc:
# What does this PR do?

This PR adds a new arg that allows users to change the value of `use_cache` in the model config. For training, usually we don't need to use a cache and this created quite a lot of issues in the past (incompatibility, regression, lower perfs). Now that v5 is coming, let's do this change. 

cc @gante


BenjaminBossan (commented):
Just for my understanding by checking some modeling files, the `use_cache` argument serves to instantiate a cache in case that `past_key_values is None`. If that is right, I think we don't need the exception for PEFT. Those PEFT methods that need to use the cache will create it and pass it explicitly as the `past_key_values` argument. Thus, it will be used even if `use_cache = False`, right?

SunMarc:
>  If that is right, I think we don't need the exception for PEFT. Those PEFT methods that need to use the cache will create it and pass it explicitly as the past_key_values argument. Thus, it will be used even if use_cache = False, right?

If those methods create the `past_key_values` argument without checking `use_cache` argument, then probably yes. Do you have tests in peft for those methods ? cc @gante for visibility

src/transformers/training_args.py
```diff
@@ -752,6 +752,11 @@ class TrainingArguments:
             Whether or not to average tokens across devices. If enabled, will use all_reduce to synchronize
             num_tokens_in_batch for precise loss calculation. Reference:
             https://github.com/huggingface/transformers/issues/34242
+
+        use_model_cache (`Optional[bool]`, *optional*, defaults to `None`):
```
Cyrilvallez: Just a nit, but I feel like `use_model_cache` is a bit hard to understand, why not the simple `use_cache`?

BenjaminBossan:
> If those methods create the `past_key_values` argument without checking `use_cache` argument, then probably yes. Do you have tests in peft for those methods ?

I ran a couple of tests in PEFT on this branch*:

`pytest tests/test_decoder_models.py tests/test_encoder_decoder_models.py -k "prefix or prompt"`

Even if I hard-code `self.model.config.use_cache = False`, they pass, so I think we're good even without the PEFT exception.

*I had to also fix the `DynamicCache` init issue discussed on the other PR.

src/transformers/training_args.py
SunMarc: I thought that it might be easier to understand but happy to switch to `use_cache`

SunMarc:
> I ran a couple of tests in PEFT on this branch*:
> 
> pytest tests/test_decoder_models.py tests/test_encoder_decoder_models.py -k "prefix or prompt"
> 
> Even if I hard-code self.model.config.use_cache = False, they pass, so I think we're good even without the PEFT exception.
> 
> *I had to also fix the DynamicCache init issue discussed on the other PR.

Thanks for checking ! I reverted the changes about peft and just set the `default` to False.

ArthurZucker (approved):
SG, we also have by default in the `check_model_inputs` something that sets use_cache to false if training

BenjaminBossan (approved):
I checked the current state of the branch again with PEFT and the relevant tests pass, so this looks good from PEFT's point of view. Regarding the more general context of this change, I can't comment on that.

SunMarc:
> SG, we also have by default in the check_model_inputs something that sets use_cache to false if training
```python
    def wrapped_fn(func):
        @wraps(func)
        def wrapper(self, *args, **kwargs):
            use_cache_arg_index = None
            if "use_cache" in func.__code__.co_varnames:
                use_cache_arg_index = func.__code__.co_varnames.index("use_cache") - 1  # -1 for self

            if (
                use_cache_arg_index is not None
                and len(args) > use_cache_arg_index
                and args[use_cache_arg_index] is not None
            ):
                use_cache = args[use_cache_arg_index]
            elif kwargs.get("use_cache") is not None:
                use_cache = kwargs["use_cache"]
            else:
                use_cache = getattr(self.config, "use_cache", None)

            if use_cache is not None:
                if getattr(self, "gradient_checkpointing", False) and self.training and use_cache:
                    logger.warning_once(
                        "`use_cache=True` is incompatible with gradient checkpointing. Setting `use_cache=False`."
                    )
                    use_cache = False
```

Only when the users tries to use gradient accumulation also !

PR #41586, opened 2025-10-14
Add fast path for bidirectional mask creation to fix regression

i3hz:
# What does this PR do?
This PR fixes the performance regression due to bidirectional masks.

Fixes # (issue): 41566

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@vasqu @Cyrilvallez


src/transformers/masking_utils.py
```diff
@@ -409,11 +409,13 @@ def sdpa_mask_recent_torch(
     # but without data-dependent slicing (i.e. torch.compile friendly)
     kv_arange = torch.arange(kv_length, device=cache_position.device)
     kv_arange += kv_offset
+    original_mask_function = mask_function
 
     # Potentially add the padding 2D mask
     if padding_mask is not None:
         mask_function = and_masks(mask_function, padding_mask_function(padding_mask))
-
+        if original_mask_function is bidirectional_mask_function:
```
vasqu: We shouldn't need to declare `original_mask_function = mask_function`, but just check directly before applying the and mask:
```suggestion
        if mask_function is bidirectional_mask_function:
```
Can you also check whether the "old" torch version is also affected, i.e. `sdpa_mask_older_torch`

vasqu: Also let's add a comment with the issue linked. E.g. we want to avoid vmapping over the masks and allow opting out early given a bidirectional mask (which is all true)

vasqu (approved):
It feels like a rubber band patch but it's a good solution without breaking too much of the existing code. Checked locally for perf and we are like ~ -5% slower so imo a good balance. Thanks a lot!  

Want to have a final opinion from @Cyrilvallez tho

src/transformers/masking_utils.py
i3hz: Yeah the old function also had the same issue . Can you look into the fix please? I've made the changes . Thanks for the feedback :)

src/transformers/masking_utils.py
```diff
@@ -489,15 +491,24 @@ def sdpa_mask_older_torch(
...
+        
+        
+        # This creates the 4D mask easily. Note that we do not include vmap over the batch_idx dimension as well,
+        # as vmap cannot handle slicing a tensor from scalar tensor (it internally calls `.item()` which vmap does not allow
+        # However, in more recent version of Pytorch, a trick was introduced to handle it - which is the reason we have
+        # `sdpa_mask_recent_torch`, as it allows more general `mask_function`
+        causal_mask = _vmap_for_bhqkv(mask_function, bh_indices=False)(None, None, cache_position, kv_arange)
+        causal_mask = causal_mask[None, None, :, :].expand(batch_size, -1, -1, -1)
+        if padding_mask is not None:
+            causal_mask = causal_mask * padding_mask[:, None, None, :]
```
vasqu:
```suggestion
    if mask_function is bidirectional_mask_function and padding_mask is not None:
        return padding_mask[:, None, None, :].expand(batch_size, 1, q_length, kv_length)

    # This creates the 4D mask easily. Note that we do not include vmap over the batch_idx dimension as well,
    # as vmap cannot handle slicing a tensor from scalar tensor (it internally calls `.item()` which vmap does not allow
    # However, in more recent version of Pytorch, a trick was introduced to handle it - which is the reason we have
    # `sdpa_mask_recent_torch`, as it allows more general `mask_function`
    causal_mask = _vmap_for_bhqkv(mask_function, bh_indices=False)(None, None, cache_position, kv_arange)
    causal_mask = causal_mask[None, None, :, :].expand(batch_size, -1, -1, -1)
    if padding_mask is not None:
        causal_mask = causal_mask * padding_mask[:, None, None, :]
```
Let's simplify this by early exiting here directly no need for an if else here

src/transformers/masking_utils.py
```diff
@@ -489,15 +491,24 @@ def sdpa_mask_older_torch(
...
-    # This creates the 4D mask easily. Note that we do not include vmap over the batch_idx dimension as well,
-    # as vmap cannot handle slicing a tensor from scalar tensor (it internally calls `.item()` which vmap does not allow
-    # However, in more recent version of Pytorch, a trick was introduced to handle it - which is the reason we have
-    # `sdpa_mask_recent_torch`, as it allows more general `mask_function`
-    causal_mask = _vmap_for_bhqkv(mask_function, bh_indices=False)(None, None, cache_position, kv_arange)
-    causal_mask = causal_mask[None, None, :, :].expand(batch_size, -1, -1, -1)
-    if padding_mask is not None:
-        causal_mask = causal_mask * padding_mask[:, None, None, :]
+    
+    # Fix for issue #41566. We want to avoid vmapping over the masks and allow opting out early given a bidirectional mask (which is all true).
```
vasqu:
```suggestion
    # Vmapping over masks can incur performance issues as reported in #41566. 
    # Thus, given the base case for bidirectional masks, we allow early exiting if we do not detect any modification 
    # to the base mask function (where the mask is all `True`).
```
Let's make the comment more explicit that it's for performance reasons (above as well)

vasqu (commented):
LGTM, this mostly nits to just simplify things. Cyril will take a look tomorrow!

Cyrilvallez:
I think you performed a bad rebase or merge, so I just force pushed to restore the branch in correct state (I was pushing on it as well)

Cyrilvallez:
Thanks a lot for the PR! I will commit a bit more to make it slightly more robust then we can merge!

i3hz:
Thanks a lot idk why the qualify test keeps failing , even tho there is no blank space D: , but i'm also blind so i won't be surprised if i missed it

Cyrilvallez:
@vasqu I cleaned the PR a bit and made it more general, could you have a last look before we merge? Also, we'll need to revert the changes you made in `executorch` integration, as it won't be needed anymore!

src/transformers/masking_utils.py
```diff
@@ -409,7 +409,9 @@ def sdpa_mask_recent_torch(
     # padding mask. Thus, we allow early exit here if we do not detect any modification to the base mask function
     if mask_function is bidirectional_mask_function:
         if padding_mask is not None:
-            return padding_mask[:, None, None, kv_offset : kv_offset + kv_length].expand(-1, -1, q_length, -1)
+            # used for slicing without data-dependent slicing
+            mask_indices = torch.arange(kv_length, device=cache_position.device) + kv_offset
+            return padding_mask[:, None, None, mask_indices].expand(-1, -1, q_length, -1)
```
vasqu, line 414: Is there a reason that we need additional slicing here? Shouldn't `prepare_padding_mask(attention_mask, kv_length, kv_offset, _slice=False)` already handle this?

vasqu:
LGTM, just a bit confused why we need another slicing :thinking: let me open a follow up PR for executorch, keeping things a bit cleaner

src/transformers/masking_utils.py
Cyrilvallez: No, because it has `_slicing=False` in this version (the recent one)!

vasqu: Gotcha, let's merge then

Cyrilvallez:
Merged, thanks a lot for the PR @i3hz!

PR #41587, opened 2025-10-14
remove ray_scope and check_quantized_param

SunMarc:
# What does this PR do?

This PR removes a few things that we don't need in v5.


MekkCyber (approved):
lgtm

PR #41591, opened 2025-10-14
[docs] Duplicate entry

stevhliu:
Fixes duplicate entry in the `toctree` that may be breaking the `custom-doc-build` job for translations (https://github.com/huggingface/transformers/actions/runs/18381182160/job/52725830409)


ydshieh (approved):
For sure!

PR #41603, opened 2025-10-15
Fix video processing channel format

zucchini-nlp:
# What does this PR do?

FIxes failing VJEPA test. The test was failing because the video was not permuted to the correct format that torchvision expects `(T, C, H, W)`


zucchini-nlp:
run-slow: vjepa

ydshieh (approved):
Works, thank you. Do you know what it worked before you PR? (or your PR change something regarding this channel stuff?)

zucchini-nlp:
> Do you know what it worked before you PR? (or your PR change something regarding this channel stuff?)

My PR stopped casting all inputs to a numpy format, so when a video is already a `torch.Tensor` it started failing. It's an assumption that channel format should not be adjuested if input is in torch

src/transformers/video_processing_utils.py
```diff
@@ -338,10 +338,16 @@ def _prepare_input_videos(
         for video in videos:
             # `make_batched_videos` always returns a 4D array per video
             if isinstance(video, np.ndarray):
-                video = to_channel_dimension_format(video, ChannelDimension.FIRST, input_data_format)
```
vasqu, line 341: Why do we not need this anymore?

src/transformers/video_processing_utils.py
```diff
@@ -338,10 +338,16 @@ def _prepare_input_videos(
...
                 # not using F.to_tensor as it doesn't handle (C, H, W) numpy arrays
                 video = torch.from_numpy(video).contiguous()
 
+            # Infer the channel dimension format if not provided
+            if input_data_format is None:
+                input_data_format = infer_channel_dimension_format(video)
```
vasqu, line 346: Could be moved out of the loop no? Or do we allow to load videos differently?

vasqu (commented):
Just some thoughts since I'm also exploring the vid processor myself in ernie vl :D

src/transformers/video_processing_utils.py
zucchini-nlp: it's moved down so we can adjust channel format always, not only when the input is in numpy format

src/transformers/video_processing_utils.py
zucchini-nlp: we can, though it won't reduce number of ops. After first loop the `input_data_format` is guaranteed to be not `None`

zucchini-nlp:
run-slow: vjepa

src/transformers/video_processing_utils.py
vasqu: Mhm, yea I missed that we override with the arg directly.

vasqu (approved):
Thx, makes sense

zucchini-nlp:
slow CI is not running anymore 🥲  Locally it's green, so I believe it's really fixed

vasqu:
Did my comment break up things? :eyes:

zucchini-nlp:
Lol, no, looks like slow CI isn't being triggered since yesterday night. I've had similar issues in other PRs

PR #41604, opened 2025-10-15
Fix TypeError: find_adapter_config_file() got an unexpected keyword argument '_adapter_model_path'

albertvillanova:
Fix TypeError: find_adapter_config_file() got an unexpected keyword argument '_adapter_model_path'
- Pass original dict instead of copy to `maybe_load_adapters`

The `adapter_kwargs` is passed to `maybe_load_adapters` as unpacked keyword arguments: Python does not pass the original dictionary, but instead creates a new dictionary inside the function; so the `pop` inside the function does not have the intended behavior.
https://github.com/huggingface/transformers/blob/5db730786d87bde8397a207bd84ae0014054fc37/src/transformers/modeling_utils.py#L4383-L4387
```python
def maybe_load_adapters(
    pretrained_model_name_or_path,
    download_kwargs: DownloadKwargs,
    **adapter_kwargs,
):
    ...
    _adapter_model_path = adapter_kwargs.pop("_adapter_model_path", None)
```

This issue was introduced by PR:
- #41445

Related downstream issue:
- https://github.com/huggingface/trl/issues/4273


albertvillanova:
CC: @ArthurZucker (who opened the PR), @Cyrilvallez (who reviewed it).

ArthurZucker (approved):
LGTM, we are probably missing a test as the CI was full green on our side!

BenjaminBossan:
I encountered this error in the PEFT CI, not sure if we need a separate test for this in transformers. Here is a minimal reproducer:

```python
model_id = "peft-internal-testing/opt-125m-dummy-lora"
AutoModelForCausalLM.from_pretrained(model_id)
```

I can confirm that this PR fixes the issue, but relying on the side effect of calling `.pop` on the original dict is maybe not the best solution IMO.

albertvillanova:
Let me know if you would prefer that `maybe_load_adapters` returns a cleaned version of `adapter_kwargs`, besides current return values.

BenjaminBossan:
> Let me know if you would prefer that `maybe_load_adapters` returns a cleaned version of `adapter_kwargs`, besides current return values.

I think that's the cleaner solution but I'll leave the final decision to you / @ArthurZucker.

albertvillanova:
@ArthurZucker is it OK if I merge this PR as it is? Or should I wait for a transformers maintainer to merge it?
Thanks! :hugs:

tomaarsen:
I'm also encountering this issue in my latest `sentence-transformers` tests, and I can confirm that this PR resolves the issue.

- Tom Aarsen

ArthurZucker:
@albertvillanova I did approve the PR to make sure you merge it as its important :)

albertvillanova:
Thanks @ArthurZucker , I was not sure about the merging policy in transformers. I know now! 🤗

PR #39182, opened 2025-07-02
Mllama fixes

remi-or:
This PR adds the `is_causal` attribute to some Attention modules in mllama and disables FA2 for the `MllamaVisionModel` . 
Some tests used to fail when `is_causal` was missing: `MllamaForCausalLMModelTest::test_flash_attn_2_fp32_ln`,  `MllamaForConditionalGenerationModelTest::test_eager_matches_fa2_generate`, ...and when it was added FA2 failed for these tests, on both Mi355 and A100. This is because the vision model uses a 4D attn mask which is not supported by FA2. 
Not an expert in VLM so can you chek `is_causal` is right @zucchini-nlp please? 

I also added Expectations for AMD MI355. After these changes, we go from `15 failed, 185 passed, 94 skipped` to `196 passed, 98 skipped` on AMD MI355.

PR #41605, opened 2025-10-15
Fix fp32_ln for various models

remi-or:
This PR fixes the test `test_flash_attn_2_fp32_ln` for several models:
- `bark` was failing the test because it call `_flash_attention_forward` directly without checking the `queries` dtype, and so the test could fail if the dtype was `torch.float32`. To fix this we re-factored out a code block into a function  `get_target_dtype`  that takes care of infering whether to cast the fp32 tesnor to fp16 or bf16, and added a called to it before the call to FA
- same for `stablelm`
- `mllama` was failing the test because `MllamaTextSelfAttention` lacks the `is_causal`attribute, which was added and set to True (it's a text attention so it's causal, as discussed in #39182)
- same for `kosmos2` but the test still fails for many many other reasons

The list of fixed test is here:
```
FAILED tests/models/bark/test_modeling_bark.py::BarkSemanticModelTest::test_flash_attn_2_fp32_ln - RuntimeError: FlashAttention only support fp16 and bf16 data type
FAILED tests/models/bark/test_modeling_bark.py::BarkCoarseModelTest::test_flash_attn_2_fp32_ln - RuntimeError: FlashAttention only support fp16 and bf16 data type
FAILED tests/models/mllama/test_modeling_mllama.py::MllamaForCausalLMModelTest::test_flash_attn_2_fp32_ln - AttributeError: 'MllamaTextSelfAttention' object has no attribute 'is_causal'
FAILED tests/models/mllama/test_modeling_mllama.py::MllamaForConditionalGenerationModelTest::test_flash_attn_2_fp32_ln - AttributeError: 'MllamaTextSelfAttention' object has no attribute 'is_causal'
FAILED tests/models/stablelm/test_modeling_stablelm.py::StableLmModelTest::test_flash_attn_2_fp32_ln - RuntimeError: FlashAttention only support fp16 and bf16 data type
```


ArthurZucker (approved):
LGTM thanks

PR #41606, opened 2025-10-15
fix(processing): Filter kwargs in ProcessorMixin call to prevent Type…

### LINKED ISSUE #41598: EncodecFeatureExtractor.__call__() got an unexpected keyword argument 'pad_to_multiple_of'
### System Info

- `transformers` version: 4.57.0
- Platform: Windows-10-10.0.26200-SP0
- Python version: 3.10.18
- Safetensors version: 0.5.3
- Accelerate version: 1.10.1
- PyTorch version (accelerator?): 2.4.0+cpu (NA)
- Jax version: not installed

### Who can help?

Hello  @eustlb @ebezzam @vasqu

Just want to report a suspected bug. I keep getting an error when i try to inference the CSM-1B model using transformers with context. The error that i got is `Error with generated segments: EncodecFeatureExtractor.__call__() got an unexpected keyword argument 'pad_to_multiple_of'`. Perhaps the issue is particularly with how the audio is stored (doesn't give error when i apply_chat_template using one text input). I tried many things to store the audio such as storing it in Torch.tensor format and referencing it from a save wav file, but it still failed.

Here is how i saved the generated_segments:
```
audio_file = f"temp_vad_recording_{int(time.time())}.wav"
audio_tensor = torch.tensor(audio_data, dtype=torch.float32).unsqueeze(0)
torchaudio.save(audio_file, audio_tensor, sample_rate)

user_input = self.transcribe_audio(audio_file)

self.generated_segments.append(
    {
        "role": "1",
        "content": [{"type": "text", "text": user_input}, {"type": "audio", "path": audio_file}],
    }
)
```

Here is how i applied chat template:
```
inputs = self.tts_processor.apply_chat_template(
                    self.generated_segments,
                    tokenize=True,
                    return_dict=True,
                ).to(self.tts_generator.device)
```

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

1. run the code using a vad and whisper for transcribing into text:
```
audio_file = f"temp_vad_recording_{int(time.time())}.wav"
audio_tensor = torch.tensor(audio_data, dtype=torch.float32).unsqueeze(0)
torchaudio.save(audio_file, audio_tensor, sample_rate)

user_input = self.transcribe_audio(audio_file)

self.generated_segments.append(
    {
        "role": "1",
        "content": [{"type": "text", "text": user_input}, {"type": "audio", "path": audio_file}],
    }
)
```

### Expected behavior

apply_chat_template should work when referencing local audio files in the path, like mentioned in the the error `Incorrect format used for `audio`. Should be an url linking to an audio, a local path, or numpy array.`

CodersAcademy006:
This PR fixes an issue where multimodal processors could raise a `TypeError` when called with arguments specific to one modality. The `ProcessorMixin.__call__` method was passing all keyword arguments to each sub-processor (tokenizer, feature extractor, etc.), causing a crash if a processor received an argument it didn't recognize (e.g., `EncodecFeatureExtractor` receiving `pad_to_multiple_of` from the tokenizer).

This fix resolves the problem by filtering the keyword arguments before they are passed to each sub-processor. It uses Python's `inspect` module to get the valid parameters for each processor's `__call__` method and only passes the arguments that match. This prevents `TypeError`s and makes the processor logic more robust.

**Fixes #41598**

### DIFF
--- a/src/transformers/processing_utils.py
+++ b/src/transformers/processing_utils.py
@@ -686,7 +686,21 @@ class ProcessorMixin(PushToHubMixin):
             attribute = getattr(self, attribute_name, None)
             input_data, input_kwargs = attribute_to_kwargs[attribute_name]
             if input_data is not None and attribute is not None:
-                attribute_output = attribute(input_data, **kwargs[input_kwargs])
+                modality_kwargs = kwargs[input_kwargs]
+                params = inspect.signature(attribute.__call__).parameters
+                if any(p.kind == inspect.Parameter.VAR_KEYWORD for p in params.values()):
+                    # Sub-processor accepts **kwargs, so it handles unknown keys itself.
+                    attribute_output = attribute(input_data, **modality_kwargs)
+                else:
+                    # Sub-processor has an explicit signature. Some kwargs may be declared
+                    # in multiple modality TypedDicts (e.g. `pad_to_multiple_of` appears in
+                    # both TextKwargs and AudioKwargs), so _merge_kwargs routes them to every
+                    # matching modality bucket. Filter to only the params this processor
+                    # explicitly accepts to avoid a TypeError.
+                    valid_keys = set(params.keys()) - {"self"}
+                    attribute_output = attribute(
+                        input_data, **{k: v for k, v in modality_kwargs.items() if k in valid_keys}
+                    )
                 outputs.update(attribute_output)
 
         return BatchFeature(outputs)
--- a/tests/test_processing_common.py
+++ b/tests/test_processing_common.py
@@ -20,6 +20,7 @@ import random
 import shutil
 import sys
 import tempfile
+import unittest
 from pathlib import Path
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


Rocketknight1:
cc @zucchini-nlp! I'm not sure about this because some of the submodules may just have `**kwargs`

src/transformers/processing_utils.py
```diff
@@ -626,7 +627,15 @@ def __call__(
             attribute = getattr(self, attribute_name, None)
             input_data, input_kwargs = attribute_to_kwargs[attribute_name]
             if input_data is not None and attribute is not None:
-                attribute_output = attribute(input_data, **kwargs[input_kwargs])
```
zucchini-nlp, line 689: in this line `kwargs` are filtered already and we are obtaining `kwargs[input_kwargs]`. That should work for all models if if there's a TypeError anywhere, I'd suggest to see what has gone wrong with specific model

auli-aziz: The model where this issue is occurring is CSM-1B, so it might be related to how that particular architecture handles its input kwargs.

Qodo-Free-For-OSS:
Hi, `ProcessorMixin.__call__` unconditionally calls `inspect.signature(attribute.__call__)`, which can raise `TypeError`/`ValueError` for some callable implementations and would crash processing before invoking the sub-processor.

**Severity:** remediation recommended | **Category:** reliability

**How to fix:** Add signature() fallback path

**Agent prompt to fix** - you can give this to your LLM of choice:

> ### Issue description
> `ProcessorMixin.__call__` now relies on `inspect.signature(attribute.__call__)` to decide whether to filter kwargs. If `inspect.signature()` raises for a particular callable implementation, processing crashes before the sub-processor is called.
> 
> ### Issue Context
> Some sub-processors may be dynamically loaded or implemented in ways that are not inspectable.
> 
> ### Fix Focus Areas
> - src/transformers/processing_utils.py[689-703]
> 
> ### Suggested approach
> - Wrap `inspect.signature(attribute.__call__)` in `try/except (TypeError, ValueError)`.
> - On exception, fall back to a safe behavior. Recommended fallback:
>   - Attempt calling the sub-processor with full `modality_kwargs` (preserves current behavior for processors that tolerate extras via internal handling).
>   - If you want to preserve the original goal of avoiding TypeError, alternatively fall back to filtering using a conservative allowlist (e.g., empty filter) only if that won’t break required args.
> - Add/adjust a unit test that simulates an uninspectable `__call__` (e.g., a callable object whose `__call__` is a built-in or raises when inspected) and assert the processor call does not crash.

---
*Found by [Qodo](https://qodo.ai) code review*

PR #41607, opened 2025-10-15
[v5] Delete `videos` from image processing classes 

zucchini-nlp:
# What does this PR do?

As per title, it was deprecated for v5


yonigozlan (approved):
Nice! Definitely needed

Cyrilvallez (commented):
I assumed the removed files were fully redundant with the corresponding video processors? 
Also, you'll need to update the modular for vidoe llama 3, apparently it did not like this change!

zucchini-nlp:
> I assumed the removed files were fully redundant with the corresponding video processors?

Yeah, they are same in terms of functionality. But the deleted ones inherit from slow image processors while the video processor cannot be "slow". So the only diff is that we're making `torchvision` the only option

zucchini-nlp:
@Cyrilvallez do you have any comments to address?

Cyrilvallez (approved):
Nop sorry! Did not approve before as we still had conflicts with modulars! Thanks!

PR #41609, opened 2025-10-15
Fix gemma gguf tokenizer

CodersAcademy006:
This PR fixes an issue where `AutoTokenizer.from_pretrained` would instantiate an incorrect `Unigram` tokenizer when loading from a Gemma GGUF file, instead of the correct `BPE`-based `GemmaTokenizer`.

The root cause was that the GGUF loading logic did not have an explicit check to handle the Gemma architecture specifically. This fix introduces a check that:
1. Detects if a `gguf_file` is being loaded.
2. Reads the GGUF metadata to confirm the architecture is `gemma` and the tokenizer model is `bpe`.
3. Overrides the `tokenizer_class` in the loaded configuration to `GemmaTokenizer`, forcing the rest of the function to instantiate the correct class.

This ensures that loading a tokenizer from a Gemma GGUF file produces the same tokenization as loading from the original Hugging Face repository.

**Fixes #41494**


src/transformers/models/auto/tokenization_auto.py
```diff
@@ -1072,6 +1072,27 @@ def from_pretrained(
...
+                    architecture = reader.get_architecture()
+                    tokenizer_model = reader.get_tokenizer_model()
+
+                    # This is the key condition
+                    if architecture == "gemma" and tokenizer_model == "bpe":
+                        logger.info("Gemma GGUF with BPE tokenizer detected. Forcing tokenizer class to 'GemmaTokenizer'.")
+                        tokenizer_config["tokenizer_class"] = "GemmaTokenizer"
+            except Exception as e:
+                logger.warning(f"Could not read GGUF metadata to determine tokenizer class: {e}")
+        # --- END OF PROPOSED FIX ---
```
Isotr0py: I think here is not a good place to add such a fix for gemma special case. Can we propose a fix at ggml integration? https://github.com/huggingface/transformers/blob/e2122c4bcb74d942bb93c11dcb55aafc4c7fdf23/src/transformers/integrations/ggml.py#L675-L737

Rocketknight1:
cc @sunmarc @mekkcyber for GGUF

PR #41610, opened 2025-10-15
torch 2.9 don't ❤️ torchcodec 💔 

ydshieh:
# What does this PR do?

Let's pin torch 2.8 for now. 

See https://github.com/meta-pytorch/torchcodec/issues/912

It is causing problems on our CircleCI for now.


vasqu (approved):
Pre approving as we need this for other PRs to not crash :pray:

PR #41612, opened 2025-10-15
Fix EncoderDecoder cache

remi-or:
In #41569 we restored thr `__iter__` method to `DynamicCache` but I missed the fact that it was also removed from `EncoderDecoderCache`. This PR fixes that and modifies the `__init__` of `EncoderDecoderCache` in the case of DDP, so it is compatible with the new system.

### DIFF
--- a/tests/utils/test_modeling_utils.py
+++ b/tests/utils/test_modeling_utils.py
@@ -1807,8 +1807,8 @@ class ModelUtilsTest(TestCasePlus):
         # simulate injecting virtual tokens like in prefix tuning
         num_virtual_tokens = 3
         past_key_values = [
-            (torch.randn(1, 2, num_virtual_tokens, 8), torch.randn(1, 2, num_virtual_tokens, 8), None),
-            (torch.randn(1, 2, num_virtual_tokens, 8), torch.randn(1, 2, num_virtual_tokens, 8), None),
+            (torch.randn(1, 2, num_virtual_tokens, 8), torch.randn(1, 2, num_virtual_tokens, 8)),
+            (torch.randn(1, 2, num_virtual_tokens, 8), torch.randn(1, 2, num_virtual_tokens, 8)),
         ]
         past_key_values = DynamicCache(past_key_values)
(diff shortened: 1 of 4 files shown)


Cyrilvallez (commented):
Whisper and a few others used to rely on the same `__init__` as dp/ddp, so they are not happy with the change unfortunately 🥲 Should not be too much trouble to fix it by adding a None though as the other tests on the other PR!

BenjaminBossan:
@remi-or I tested this branch on PEFT (`pytest tests/test_decoder_models.py tests/test_encoder_decoder_models.py -k "prefix or prompt"`) and the tests are passing, thanks!

remi-or:
Nice! I think the failing tests have nothing to do with the PR, so this is ready to be merged imo :)

ydshieh:
run-slow: vit

remi-or:
run-slow: vit

tests/utils/test_modeling_utils.py
```diff
@@ -1807,8 +1807,8 @@ def test_cache_when_needed_at_train_time(self):
         # simulate injecting virtual tokens like in prefix tuning
         num_virtual_tokens = 3
         past_key_values = [
-            (None, torch.randn(1, 2, num_virtual_tokens, 8), torch.randn(1, 2, num_virtual_tokens, 8)),
-            (None, torch.randn(1, 2, num_virtual_tokens, 8), torch.randn(1, 2, num_virtual_tokens, 8)),
+            (torch.randn(1, 2, num_virtual_tokens, 8), torch.randn(1, 2, num_virtual_tokens, 8), None),
+            (torch.randn(1, 2, num_virtual_tokens, 8), torch.randn(1, 2, num_virtual_tokens, 8), None),
```
Cyrilvallez: We could even remove the Nones completely now

Cyrilvallez (approved):
Yes, LGTM! Thanks for the fix!

tests/utils/test_modeling_utils.py
remi-or: Done!

PR #41613, opened 2025-10-15
Fix quantization base class 

SunMarc:
# What does this PR do?

This PR fixes issues that was created from this [PR](https://github.com/huggingface/transformers/pull/41445) regarding quantization.


src/transformers/quantizers/base.py
```diff
@@ -350,12 +353,6 @@ def update_state_dict_with_metadata(self, state_dict, metadata):
         """Update state dict with metadata. Default behaviour returns state_dict"""
         return state_dict
 
-    @abstractmethod
-    def _process_model_before_weight_loading(self, model, **kwargs): ...
-
-    @abstractmethod
-    def _process_model_after_weight_loading(self, model, **kwargs): ...
-
```
MekkCyber, line 366: we no longer need this ?

MekkCyber (approved):
Thanks for fixing

src/transformers/quantizers/base.py
SunMarc: it is already defined above and it is not an abstract class anymore

PR #41614, opened 2025-10-15
[`Docs`] Fix changed references

vasqu:
#41445 moved `load_sharded_checkpoint` to the trainer utils instead of the modeling utils. This PR changes the remaining references and (hopefully) fixes the CI on main with the doc builder


Cyrilvallez (approved):
Nice catch, thsanks!

ArthurZucker (approved):
Thanks for the catch indeed!

PR #41617, opened 2025-10-15
Reinstate early CUDA init fix

hmellor:
vLLM found that this import was causing CUDA to be initialised early


MekkCyber (approved):
Sounds good ! thanks for fixing

ArthurZucker (approved):
sorry for breaking again

ydshieh:
merge as the failing tests is irrelevant and fixed 

[41618](https://github.com/huggingface/transformers/pull/41618)

PR #41620, opened 2025-10-15
Remove the head masking block in some vision models

ydshieh:
# What does this PR do?

#36545 introduced a mistake regarding the headmask

before that PR, we had some

```
        # Mask heads if we want to
        if head_mask is not None:
            attention_probs = attention_probs * head_mask
```
but after the PR, it became

```
    # Mask heads if we want to
        attn_weights = attn_weights * attention_mask
```
which is apparently wrong.

Furthermore, we already remove the support of head masking.

This PRs simply remove the blocks.


vasqu (approved):
Thx, what nasty bug :sweat:

PR #41625, opened 2025-10-15
[`Masks`] Fix mask handling in eager for vision models

vasqu:
As per title, some vision models use masks, let's sync with bert this time and reduce errors that could be introduced

### DIFF
--- a/src/transformers/models/audio_spectrogram_transformer/modeling_audio_spectrogram_transformer.py
+++ b/src/transformers/models/audio_spectrogram_transformer/modeling_audio_spectrogram_transformer.py
@@ -96,30 +96,29 @@ class ASTPatchEmbeddings(nn.Module):
         return embeddings
 
 
-# Copied from transformers.models.vit.modeling_vit.eager_attention_forward
+# Copied from transformers.models.bert.modeling_bert.eager_attention_forward
 def eager_attention_forward(
     module: nn.Module,
     query: torch.Tensor,
     key: torch.Tensor,
     value: torch.Tensor,
     attention_mask: Optional[torch.Tensor],
-    scaling: float,
+    scaling: Optional[float] = None,
     dropout: float = 0.0,
-    **kwargs,
+    **kwargs: Unpack[TransformersKwargs],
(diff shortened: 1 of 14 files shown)


ydshieh:
il faut run-slow svp, Monsieur, au moins une parti 🙏

vasqu:
run-slow: audio_spectrogram_transformer, deit, dinov2, dinov2_with_registers, dinov3_vit, dpt, ijepa, videomae, vit, vit_mae, vit_msn, vitpose_backbone, vivit, yolos

vasqu:
Checked against main and the failing tests are the same, videomae in general seems to be broken tho (integration tests)

src/transformers/models/audio_spectrogram_transformer/modeling_audio_spectrogram_transformer.py
```diff
@@ -96,25 +96,28 @@ def forward(self, input_values: torch.Tensor) -> torch.Tensor:
...
 
-    # Normalize the attention scores to probabilities.
-    attn_weights = nn.functional.softmax(attn_weights, dim=-1, dtype=torch.float32).to(query.dtype)
+    if attention_mask is not None:
+        attention_mask = attention_mask[:, :, :, : key.shape[-2]]
+        attn_weights = attn_weights + attention_mask
 
-    # This is actually dropping out entire tokens to attend to, which might
-    # seem a bit unusual, but is taken from the original Transformer paper.
+    attn_weights = nn.functional.softmax(attn_weights, dim=-1)
```
vasqu, line 120: Only real diff is here where we don't cast in fp32, however this was originally also not the case and was only introduced in the attn refactor (which shouldn't have happened), see https://github.com/huggingface/transformers/pull/36545/files#diff-e3964696c93e4ad2f40a0ffd9f1079e8fc58645d328e8f7ad27f9f9c5dba47b5L225-L226

NielsRogge (approved):
Thanks! I hope the slow tests pass

ArthurZucker (approved):

PR #41627, opened 2025-10-15
[`Executorch`] Simplify for encoder models

vasqu:
Now that we include a fast path using no vmapping, we can revert the extra treatment. Followup to #41586


vasqu:
run-slow: albert, bert, distilbert, mobilebert, roberta

vasqu:
Flaky fa test as usual :laughing:

Cyrilvallez (approved):
Thanks a lot for reverting!

PR #41631, opened 2025-10-15
Incorrect access of dataset field fixed

vignesh1507:
In `test_tokenizer`:

`for text in dataset[i]["premise"].values():
    test_string(slow, fast, text)

# hypothesis, all languages
for text in dataset[i]["hypothesis"]["translation"]:
    test_string(slow, fast, text)`

The "premise" field in the XNLI dataset is a string, not a dictionary**. Calling .values() on a string will raise an error. Similarly, "hypothesis" is also a string (not a dictionary), so "hypothesis"]["translation" will fail.


Rocketknight1:
cc @ydshieh do you know if this script is still in use?

ydshieh:
@Rocketknight1 

Can't find any usage in the codebase. The recent changes in the history are more about type hints. Let's remove it.

PR #41632, opened 2025-10-15
[CI] Build translated docs

stevhliu:
Actually, this might be the reason why the translated docs aren't building when commenting `build-doc`. The `tr` and `te` language codes don't exist 😅


ydshieh (approved):
Thank you. I swear it was not me adding those. Back to Aug 4, I just copied them from `.github/workflows/build_documentation.yml` which has those langs.

ydshieh:
build-doc

PR #41634, opened 2025-10-15
[MODEL] Nanochat implementation

burtenshaw:
This PR introduces Andrej Karpathy's `nanochat` to transformers.

The transformers compatible weights are currently available in two places:

- A [pr](https://huggingface.co/karpathy/nanochat-d32/discussions/1) on Karpathy's repo (large model)
- A [repo](https://huggingface.co/nanochat-students/nanochat-d20) on the community org  (small model)

# ToDo

- CI is failing dues to tests for `SwiftFormerForQuestionAnswering` etc.


src/transformers/models/nanochat/tokenizer_nanochat.py
xenova, line 1: I know the PR is still in draft, but we might not need this file. Most of the chat template logic should be able to be moved to the chat_template.jinja file.

Also, if we do need this file, it should be named `tokenization_nanochat.py`

burtenshaw: Thanks. Deleted it and used a jinja. https://huggingface.co/nanochat-students/nanochat-d20/blob/main/chat_template.jinja

src/transformers/models/nanochat/configuration_nanochat.py
```diff
@@ -0,0 +1,100 @@
+from transformers.configuration_utils import PretrainedConfig
+from transformers.utils import logging
```
xenova:
```suggestion
from ...configuration_utils import PretrainedConfig
from ...utils import logging
```

src/transformers/models/nanochat/configuration_nanochat.py
```diff
@@ -0,0 +1,100 @@
...
+        legacy_num_kv_heads = kwargs.pop("n_kv_head", None)
+
+        if legacy_hidden_size is not None:
+            hidden_size = legacy_hidden_size
+        if legacy_num_layers is not None:
+            num_hidden_layers = legacy_num_layers
+        if legacy_num_heads is not None:
+            num_attention_heads = legacy_num_heads
+        if legacy_num_kv_heads is not None:
+            num_key_value_heads = legacy_num_kv_heads
```
xenova: maybe not necessary if we just update the 1 or 2 checkpoints on the hub?

docs/source/en/model_doc/nanochat.md
```diff
@@ -0,0 +1,119 @@
...
+
+# NanoChat
+
+<div style="float: right;">
+    <div class="flex flex-wrap space-x-1">
+        <img alt="PyTorch" src="https://img.shields.io/badge/PyTorch-DE3412?style=flat&logo=pytorch&logoColor=white">
+    </div>
+</div>
+
+[NanoChat](https://huggingface.co/karpathy/nanochat-d32) is a compact decoder-only transformer model designed for educational purposes and efficient training. The model features several fundamental architectural innovations which are common in modern transformer models. Therefore, it is a good model to use as a starting point to understand the principles of modern transformer models.
```
vasqu: Nit: we could mention that it is based on llama

docs/source/en/model_doc/nanochat.md
```diff
@@ -0,0 +1,119 @@
...
+
+The architecture is based on [nanochat](https://github.com/karpathy/nanochat) by [Andrej Karpathy](https://huggingface.co/karpathy), adapted for the Hugging Face Transformers library.
```
vasqu: You can also mention yourself as contributor if you want to ;)

src/transformers/models/nanochat/convert_nanochat_checkpoints.py
vasqu, line 1: Conversion scripts usually also go under the model, so `models/nanochat/convert_....`

src/transformers/models/nanochat/convert_nanochat_checkpoints.py
```diff
@@ -0,0 +1,151 @@
...
+    rows = key_weight.shape[0]
+    head_dim = config.hidden_size // config.num_attention_heads
+    if rows % head_dim != 0:
+        return config.num_key_value_heads
+    inferred = rows // head_dim
+    LOGGER.info("Inferred %s key_value heads from checkpoint", inferred)
+    return max(inferred, 1)
+
+
+def convert_layer(old_prefix: str, new_prefix: str) -> dict[str, str]:
```
vasqu, line 40: You can check out csm to standardize this a bit https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/csm/convert_csm.py#L37

src/transformers/models/nanochat/configuration_nanochat.py
vasqu, line 1: Dont forget the licences

src/transformers/models/nanochat/configuration_nanochat.py
```diff
@@ -0,0 +1,100 @@
...
+class NanoChatConfig(PretrainedConfig):
+    r"""Configuration for the NanoChat model."""
+
+    model_type = "nanochat"
+    keys_to_ignore_at_inference = ["past_key_values"]
+    attribute_map = {
+        "hidden_size": "n_embd",
+        "num_hidden_layers": "n_layer",
+        "num_attention_heads": "n_head",
+    }
```
vasqu: If we have a conversion, we should already update these there already! This should be more of a last measure kind of thing here with the mapping

src/transformers/models/nanochat/configuration_nanochat.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+logger = logging.get_logger(__name__)
+
+
+__all__ = ["NanoChatConfig"]
+
+
+class NanoChatConfig(PretrainedConfig):
+    r"""Configuration for the NanoChat model."""
```
vasqu: Probably aware but docs please

src/transformers/models/nanochat/configuration_nanochat.py
```diff
@@ -0,0 +1,100 @@
...
+            num_attention_heads = legacy_num_heads
+        if legacy_num_kv_heads is not None:
+            num_key_value_heads = legacy_num_kv_heads
+
+        super().__init__(
+            bos_token_id=bos_token_id,
+            eos_token_id=eos_token_id,
+            pad_token_id=pad_token_id,
+            **kwargs,
+        )
```
vasqu: Should be at the end

src/transformers/models/nanochat/configuration_nanochat.py
```diff
@@ -0,0 +1,100 @@
...
+        initializer_range: float = 0.02,
+        rope_theta: float = 10000.0,
+        rope_scaling: dict | None = None,
+        use_cache: bool = True,
+        logits_soft_cap: float | None = 15.0,
+        qkv_bias: bool = False,
+        bos_token_id: int = 0,
+        eos_token_id: int = 1,
+        pad_token_id: int = 1,
+        **kwargs,
```
vasqu, line 126: We can also add the rope stuff here already, I will explain why we can use the Llama RoPE in the modeling file

tests/models/nanochat/test_modeling_nanochat.py
```diff
@@ -0,0 +1,218 @@
...
+class NanoChatModelTester(CausalLMModelTester):
+    config_class = NanoChatConfig
+    if is_torch_available():
+        base_model_class = NanoChatModel
+        causal_lm_class = NanoChatForCausalLM
+
+
+@require_torch
+class NanoChatModelTest(CausalLMModelTest, unittest.TestCase):
+    model_tester_class = NanoChatModelTester
```
vasqu, line 48: Nice!

tests/models/nanochat/test_modeling_nanochat.py
```diff
@@ -0,0 +1,218 @@
...
+
+        # Check first 10 logits at position [0,0,:10]
+        EXPECTED_SLICE = torch.tensor(
+            [-12.875, -13.0625, -13.1875, -13.1875, -13.1875, -13.1875, -13.1875, -13.1875, -12.625, -4.21875]
+        )
+        torch.testing.assert_close(logits[0, 0, :10], EXPECTED_SLICE, rtol=1e-3, atol=1e-2)
+
+        del model
+        backend_empty_cache(torch_device)
+        gc.collect()
```
vasqu: See https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/tests/models/llama/test_modeling_llama.py#L68-L76 instead

vasqu: Same for the other integration tests

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+        value_states = self.v_proj(hidden_states).view(batch, seq_len, self.num_kv_heads, self.head_dim)
+
+        # Apply Rotary Embeddings to queries and keys to get relative positional encoding
+        cos, sin = position_embeddings
+        query_states, key_states = _apply_rotary_pos_emb(query_states, key_states, cos, sin)
+
+        # Apply QK normalization (RMSNorm) - a key feature of NanoChat architecture
+        # This helps stabilize training and is applied AFTER RoPE
+        query_states = F.rms_norm(query_states, (query_states.size(-1),), eps=self.config.rms_norm_eps)
+        key_states = F.rms_norm(key_states, (key_states.size(-1),), eps=self.config.rms_norm_eps)
```
vasqu: We have multiple usages of torch functional. Let's move this to a module instead, first glance llama4 has a non weight version here https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama4/modeling_llama4.py#L98-L110

We want this to be attributes so we can use modular for the forward here. See Gemma3 for attention with qk norm (there might be a better alternative) https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/gemma3/modeling_gemma3.py#L306-L311

We will then need to only overwrite the __init__

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+from ...modeling_utils import ALL_ATTENTION_FUNCTIONS, PreTrainedModel
+from ...processing_utils import Unpack
+from ...utils import TransformersKwargs, auto_docstring, logging
+from .configuration_nanochat import NanoChatConfig
+
+
+logger = logging.get_logger(__name__)
+
+
+def _apply_rotary_pos_emb(
```
vasqu: I will add comments step by step explaining why this is essentially Llama RoPE

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+        cos: Cosine part of the rotary embedding of shape [1, seq_len, 1, head_dim//2]
+        sin: Sine part of the rotary embedding of shape [1, seq_len, 1, head_dim//2]
+
+    Returns:
+        Tuple of rotated query and key tensors of shape [batch, seq_len, num_heads, head_dim] and [batch, seq_len, num_kv_heads, head_dim]
+    """
+    # Expects query/key as [B, T, H, D] and cos/sin as [1, T, 1, D//2]
+    d = query.shape[3] // 2
+    q1, q2 = query[..., :d], query[..., d:]
+    k1, k2 = key[..., :d], key[..., d:]
```
vasqu: The same as https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L111-L112

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+
+    query_rot = torch.cat([q1 * cos + q2 * sin, q1 * (-sin) + q2 * cos], dim=-1)
+    key_rot = torch.cat([k1 * cos + k2 * sin, k1 * (-sin) + k2 * cos], dim=-1)
```
vasqu: A bit more complicated and we need a slight adjustment in our `rotate_half`. In Llama we rotate with https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L113
and add with https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L138
Mind that our prepared sin/cos are `head_dim`, not `head_dim//2`, by duplicating https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L102

Mathematically, with Llama we get:
```
[q1 * cos - q2 * sin, q1 * sin + q2 * cos]
```
Not that we rotate correctly but the minus/plus signs don't add up. For this we need to change the rotation slightly from https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L113
to `torch.cat((x2, -x1), dim=-1)` so that we go to 
```
[q1 * cos + q2 * sin, -(q1 * sin) + q2 * cos]
```
which is the same as here

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+        # Return cached embeddings if they exist and match device/dtype
+        if self._rotary_cos is not None and self._rotary_cos.device == device and self._rotary_cos.dtype == dtype:
+            return self._rotary_cos, self._rotary_sin
+
+        head_dim = self.config.hidden_size // self.config.num_attention_heads
+        # Stride the time steps (positions)
+        positions = torch.arange(self.config.max_position_embeddings, device=device, dtype=torch.float32)
+        # Stride the channels (only even channels, head_dim//2 frequencies)
+        freqs = torch.arange(0, head_dim, 2, dtype=torch.float32, device=device)
+        inv_freq = 1.0 / (self.config.rope_theta ** (freqs / head_dim))
```
vasqu: The same as default rope (in llama) https://github.com/huggingface/transformers/blob/bdbc2d037b1751c71b30a488f0a168deabef9a90/src/transformers/modeling_rope_utils.py#L125-L131

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+        # Calculate the rotation frequencies at each (time, channel) pair
+        angles = torch.einsum("i,j->ij", positions, inv_freq)
+        # Compute cos and sin, add batch and head dims for broadcasting
+        cos = angles.cos()[None, :, None, :]  # [1, seq_len, 1, head_dim//2]
+        sin = angles.sin()[None, :, None, :]  # [1, seq_len, 1, head_dim//2]
```
vasqu: We have `[bsz, seq_len, head_dim]` instead by duplicating https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L102
and the aforementioned extension of the outer product to 2D (positions)

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+    q1, q2 = query[..., :d], query[..., d:]
+    k1, k2 = key[..., :d], key[..., d:]
+
+    query_rot = torch.cat([q1 * cos + q2 * sin, q1 * (-sin) + q2 * cos], dim=-1)
+    key_rot = torch.cat([k1 * cos + k2 * sin, k1 * (-sin) + k2 * cos], dim=-1)
+
+    return query_rot.to(query.dtype), key_rot.to(key.dtype)
+
+
+def _repeat_kv(hidden_states: torch.Tensor, n_rep: int) -> torch.Tensor:
```
vasqu: https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L159

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+        Repeated key and value tensors of shape [batch, seq_len, num_kv_heads * n_rep, head_dim]
+    """
+    batch, num_key_value_heads, seq_len, head_dim = hidden_states.shape
+    if n_rep == 1:
+        return hidden_states
+    hidden_states = hidden_states[:, :, None, :, :].expand(batch, num_key_value_heads, n_rep, seq_len, head_dim)
+    return hidden_states.reshape(batch, num_key_value_heads * n_rep, seq_len, head_dim)
+
+
+def eager_attention_forward(
```
vasqu, line 163: https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L171

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+
+
+class NanoChatAttention(nn.Module):
+    """
+    Multi-headed attention from NanoChat with custom RoPE and QK normalization.
+
+    Based on: https://github.com/karpathy/nanochat/blob/main/nanochat/gpt.py#L64
+    """
+
+    def __init__(self, config: NanoChatConfig, layer_idx: int):
```
vasqu, line 199: We need the `is_causal` flag https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L208

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+    """MLP module for NanoChat with ReLU^2 activation."""
+
+    def __init__(self, config: NanoChatConfig):
+        super().__init__()
+        self.fc = nn.Linear(config.hidden_size, config.intermediate_size, bias=False)
+        self.proj = nn.Linear(config.intermediate_size, config.hidden_size, bias=False)
+
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        hidden_states = self.fc(hidden_states)
+        hidden_states = F.relu(hidden_states).square()
```
vasqu: We have a relu² act here https://github.com/huggingface/transformers/blob/bdbc2d037b1751c71b30a488f0a168deabef9a90/src/transformers/activations.py#L206

We should be able to infer this with something like https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L152

And there is probably a model which does this then from which we can possibly inherit

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+        attention_mask: Optional[torch.Tensor] = None,
+        position_ids: Optional[torch.LongTensor] = None,
+        past_key_values: Optional[Cache] = None,
+        use_cache: Optional[bool] = False,
+        cache_position: Optional[torch.LongTensor] = None,
+        position_embeddings: Optional[tuple[torch.Tensor, torch.Tensor]] = None,
+        **kwargs: Unpack[TransformersKwargs],
+    ) -> tuple[torch.Tensor, Optional[torch.Tensor]]:
+        residual = hidden_states
+        hidden_states = F.rms_norm(hidden_states, (hidden_states.size(-1),), eps=self.config.rms_norm_eps)
```
vasqu: See my comments in the attention module, we should avoid using functional here, and make this a module instead, should also allow for easier inheritance

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+        hidden_states, self_attn_weights = self.self_attn(
```
vasqu:
```suggestion
        hidden_states, _ = self.self_attn(
```
Llama style with https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L325-L328

this will handle everything for you, kwargs, collecting the tensors etc

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+        hidden_states = residual + hidden_states
+        return hidden_states, self_attn_weights
+
+
+@auto_docstring
+class NanoChatPreTrainedModel(PreTrainedModel):
+    config_class = NanoChatConfig
+    base_model_prefix = "model"
+    _no_split_modules = ["NanoChatDecoderLayer"]
+    _supports_attention_backend = True
```
vasqu, line 346: We should have all flags enabled here https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L316-L324

Currently, it seems only eager attention is run then + the record attribute as a reminder (handles output_xxx)

https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/models/llama/modeling_llama.py#L325-L328

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+
+    def _init_weights(self, module: nn.Module) -> None:
```
vasqu, line 352: Can you check with https://github.com/huggingface/transformers/blob/bdbc2d037b1751c71b30a488f0a168deabef9a90/src/transformers/modeling_utils.py#L2629 if there is some stuff already handled? E.g. linear looks sus

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+        attention_mask: Optional[torch.Tensor] = None,
+        position_ids: Optional[torch.LongTensor] = None,
+        past_key_values: Optional[Cache] = None,
+        inputs_embeds: Optional[torch.FloatTensor] = None,
+        use_cache: Optional[bool] = None,
+        cache_position: Optional[torch.LongTensor] = None,
+        **kwargs: Unpack[TransformersKwargs],
+    ) -> BaseModelOutputWithPast:
+        output_attentions = kwargs.get("output_attentions", False)
+        output_hidden_states = kwargs.get("output_hidden_states", False)
```
vasqu: Don't need this with the recorder

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -0,0 +1,497 @@
...
+        position_ids_to_use = cache_position.clamp(0, self.config.max_position_embeddings - 1)
+        cos = cos[:, position_ids_to_use]
+        sin = sin[:, position_ids_to_use]
+
+        hidden_states = inputs_embeds
+        hidden_states = F.rms_norm(hidden_states, (hidden_states.size(-1),), eps=self.config.rms_norm_eps)
+
+        # Collect hidden states and attentions if requested
+        all_hidden_states = () if output_hidden_states else None
+        all_self_attns = () if output_attentions else None
```
vasqu: Same for this and similar logic

src/transformers/models/nanochat/configuration_nanochat.py
```diff
@@ -0,0 +1,100 @@
...
+    r"""Configuration for the NanoChat model."""
+
+    model_type = "nanochat"
+    keys_to_ignore_at_inference = ["past_key_values"]
+    attribute_map = {
+        "hidden_size": "n_embd",
+        "num_hidden_layers": "n_layer",
+        "num_attention_heads": "n_head",
+    }
+
```
vasqu, line 104: Just a random point but it might be nice to have a tp/pp plan if the model is big https://github.com/huggingface/transformers/blob/bdbc2d037b1751c71b30a488f0a168deabef9a90/src/transformers/models/llama/configuration_llama.py#L146-L159

vasqu (commented):
An initial review, I went a bit into the deep end with the RoPE but I hope I could explain what's going on :sweat_smile: 

The biggest tasks are 
- modular
- RoPE
- `_can_record` to properly support some features

src/transformers/models/auto/modeling_auto.py
```diff
@@ -594,7 +596,6 @@ class _BaseModelWithGenerate(PreTrainedModel, GenerationMixin):
         ("mvp", "MvpForConditionalGeneration"),
         ("nezha", "NezhaForMaskedLM"),
         ("nllb-moe", "NllbMoeForConditionalGeneration"),
-        ("nystromformer", "NystromformerForMaskedLM"),
```
vasqu, line 582: Can you check what happened here and below, multiple models are changed etc :eyes:

src/transformers/models/nanochat/modeling_nanochat.py
burtenshaw: > Currently, it seems only eager attention is run then + the record attribute as a reminder (handles output_xxx)

Sorry. I don't fully understand this. Can you check [16a0207d9e1fedf0a9f0445a60bddcfbe722d02b](https://github.com/huggingface/transformers/pull/41634/commits/16a0207d9e1fedf0a9f0445a60bddcfbe722d02b) please.

vasqu: I think you still need to convert to the modeling file, i.e. something along `python utils/modular_model_converter.py nanochat`. But yes, that looks correct. 

- Each flag indicates that another attn flavor can be run --> no flags == only eager attention can be run, no other more efficient variations such as SDPA or Flash Attn.
- The record flag allows avoiding collecting all these tensors manually, i.e. no more tuples/lists to collect manually + the output_xxx kwargs also handled automatically then.

burtenshaw: Thanks for explanation.

burtenshaw:
> An initial review, I went a bit into the deep end with the RoPE but I hope I could explain what's going on 😅
> 
> The biggest tasks are
> 
> * modular
> * RoPE
> * `_can_record` to properly support some features

@vasqu Thanks for the thorough review. It's really helpful to me. I've now responded to all the comments in line, reran the conversion, and updated the model repos. 

edit: I ended up having to revert some of last changes and (re)use less llama modules because the model performance regressed. https://github.com/huggingface/transformers/pull/41634/commits/99801d04cd9cae2c9b614708f11444e290fb6ac1

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -343,18 +371,13 @@ def _init_weights(self, module: nn.Module) -> None:
...
-    """
-
     def __init__(self, config: NanoChatConfig):
-        # Call PreTrainedModel.__init__ directly to avoid LlamaModel's __init__
         super().__init__(config)
         self.padding_idx = config.pad_token_id
         self.vocab_size = config.vocab_size
 
         self.embed_tokens = nn.Embedding(config.vocab_size, config.hidden_size, self.padding_idx)
+        self.initial_norm = NanoChatRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
```
vasqu: Could it be that this was the regression, i.e. a norm was missing that was there before?

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -373,22 +402,25 @@ def forward(
         position_ids: Optional[torch.LongTensor] = None,
         past_key_values: Optional[Cache] = None,
         inputs_embeds: Optional[torch.FloatTensor] = None,
-        cache_position: Optional[torch.LongTensor] = None,
         use_cache: Optional[bool] = None,
+        cache_position: Optional[torch.LongTensor] = None,
         **kwargs: Unpack[TransformersKwargs],
     ) -> BaseModelOutputWithPast:
+        output_attentions = kwargs.get("output_attentions", False)
+        output_hidden_states = kwargs.get("output_hidden_states", False)
```
vasqu: With record outputs and the `check_mdoel_inputs` decorator, we should not need to do this + the manual collection below.

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -406,45 +438,63 @@ def forward(
 
         hidden_states = inputs_embeds
         position_embeddings = self.rotary_emb(hidden_states, position_ids)
+        hidden_states = self.initial_norm(hidden_states)
```
vasqu: This seems like the only functional real change

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -65,38 +90,31 @@ def __init__(self, config: NanoChatConfig, device=None):
...
+            cos = freqs.cos() * self.attention_scaling
+            sin = freqs.sin() * self.attention_scaling
 
-class NanoChatRMSNorm(torch.nn.Module):
-    """
-    NanoChatRMSNorm inherits from Llama4TextL2Norm (weight-less RMS normalization).
-    Overrides __init__ to match NanoChat's API with hidden_size parameter.
-    """
+        # Add extra dimension for NanoChat's broadcasting: [batch, seq_len] -> [batch, seq_len, 1, head_dim//2]
+        return cos.unsqueeze(2).to(dtype=x.dtype), sin.unsqueeze(2).to(dtype=x.dtype)
```
vasqu: We later squeeze and unsqueeze again. Imo, the changes to the RoPE seems unnecessary to me :thinking:

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -206,28 +220,46 @@ def forward(
...
+            .contiguous()
+        )
 
-        cos, sin = position_embeddings
+        # Apply Rotary Embeddings to queries and keys to get relative positional encoding
+        cos, sin = position_embeddings  # [batch, seq_len, 1, head_dim//2]
+        cos = cos.squeeze(2)
+        sin = sin.squeeze(2)
+        cos = torch.cat([cos, cos], dim=-1)
+        sin = torch.cat([-sin, -sin], dim=-1)
```
vasqu: This is the additional squeeze (and unsqueeze in apply rot) I mean. These cats are more expensive than duplicating the sin/cos from the getgo as we do this torch cat for all layers while the other way, we only do it once per forward

vasqu (commented):
Just some quick comments because I see your comment about a regression. We should really not have to change the RoPE by that much. It rather seems to me that a rms normalization that was missing?

I will take a proper look on monday

src/transformers/models/nanochat/modeling_nanochat.py
burtenshaw: 100%

burtenshaw:
@vasqu Thanks again. It's now running on a llama rope with just a sine inversion. https://github.com/huggingface/transformers/pull/41634/commits/57e0e3573630200c551af59fe7bdb0c27ffe56a2

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+from ..llama.modeling_llama import LlamaPreTrainedModel, LlamaRotaryEmbedding, apply_rotary_pos_emb, eager_attention_forward
+from ..llama4.modeling_llama4 import Llama4TextL2Norm
+from .configuration_nanochat import NanoChatConfig
+
+
+class NanoChatRMSNorm(Llama4TextL2Norm):
+    """
+    NanoChatRMSNorm inherits from Llama4TextL2Norm (weight-less RMS normalization).
+    Overrides __init__ to match NanoChat's API with hidden_size parameter.
+    """
```
vasqu: Let's not make comments like, it inherits from xxx or modular specific comments (override). We could either drop this completely or just say that it is a weightless RMSNorm.

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+
+    def __init__(self, hidden_size, eps=1e-6):
+        super().__init__(eps=eps)
+        self.hidden_size = hidden_size
```
vasqu: Can be removed, no?

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+            self.v_proj(hidden_states)
+            .view(batch, seq_len, self.num_kv_heads, self.head_dim)
+            .transpose(1, 2)
+            .contiguous()
+        )
+
+        # Apply Rotary Embeddings to queries and keys to get relative positional encoding
+        cos, sin = position_embeddings
+        # NanoChat uses a negative sine for the rotary embedding
+        sin = -sin
```
vasqu: I'm really annoying on this, but let's change this back with the rotate half again, i.e. we change the signs on rotate half and don't need this then. This allows us to properly use modular on the attention module; providing an example in a second.

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+    def __init__(self, hidden_size, eps=1e-6):
+        super().__init__(eps=eps)
+        self.hidden_size = hidden_size
+
+
+class NanoChatRotaryEmbedding(LlamaRotaryEmbedding):
+    pass
+
+
+class NanoChatAttention(nn.Module):
```
vasqu: We can inherit from Qwen3 https://github.com/huggingface/transformers/blob/ce4ffeeb6c39a9e7e2e62e23606d0b83b9a49b96/src/transformers/models/qwen3/modeling_qwen3.py#L223

We can at least inherit from the init while changing the following
- We do not use `layer_type` in nanochat, so we can remove it after super init with `del layer_type`
- We set `self.sliding_window=None`
- `qkv_bias` <-> attention_bias

On the forward, I can see one difference then (when the signs are changed in rotate half instead) --> the order of norm an RoPE might be different, i.e. rope -> norm (nanochat) vs norm -> rope (qwen3). Would need to check if this has any effect; if yes, then we have to overwrite the forward and mark this as special with a small comment

src/transformers/models/nanochat/modular_nanochat.py
vasqu: Let's change this in the inits of the other modules instead. It's weird to have hidden_size when it's unused

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+        cache_position: Optional[torch.LongTensor] = None,
+        **kwargs: Unpack[TransformersKwargs],
+    ) -> tuple[torch.Tensor, Optional[torch.Tensor]]:
+        batch, seq_len, _ = hidden_states.shape
+        input_shape = hidden_states.shape[:-1]
+
+        # Project the input to get queries, keys, and values [batch, num_heads, seq_len, head_dim]
+        query_states = (
+            self.q_proj(hidden_states).view(batch, seq_len, self.num_heads, self.head_dim).transpose(1, 2).contiguous()
+        )
```
vasqu: General tips, in case we have to override it even with qwen3 inheritance. Don't explicitly cast the shape, use things like https://github.com/huggingface/transformers/blob/ce4ffeeb6c39a9e7e2e62e23606d0b83b9a49b96/src/transformers/models/qwen3/modeling_qwen3.py#L263-L265

Super important for tensor parallelism to properly shard across heads.

vasqu: Also the contiguous is unnecessary, we call it if it's necessary in our attention versions

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+        sin = -sin
+        query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)
+
+        # Apply QK normalization (RMSNorm)
+        query_states = self.query_norm(query_states)
+        key_states = self.key_norm(key_states)
+
+        # Apply KV cache: insert current k,v into cache, get the full view so far
+        if past_key_values is not None:
+            cache_kwargs = {"cache_position": cache_position}
```
vasqu: https://github.com/huggingface/transformers/blob/ce4ffeeb6c39a9e7e2e62e23606d0b83b9a49b96/src/transformers/models/qwen3/modeling_qwen3.py#L274

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+            query_states,
+            key_states,
+            value_states,
+            attention_mask,
+            dropout=0.0 if not self.training else self.attention_dropout,
+            scaling=self.scaling,
+            **kwargs,
+        )
+
+        # Reshape and project output
```
vasqu: We have a lot of these small comments which we should drop imo

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+        attn_output = attn_output.reshape(*input_shape, -1).contiguous()
+        attn_output = self.o_proj(attn_output)
+        return attn_output, attn_weights
+
+
+class NanoChatMLP(nn.Module):
```
vasqu: Inheritable from https://github.com/huggingface/transformers/blob/818f7f10e40cbceb3ce8a8b9a1f174f6b9181da2/src/transformers/models/clip/modeling_clip.py#L354

Just needs renames of the projections

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+        self.act_fn = ACT2FN[config.hidden_act]
+
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        hidden_states = self.fc(hidden_states)
+        hidden_states = self.act_fn(hidden_states)
+        hidden_states = self.proj(hidden_states)
+        return hidden_states
+
+
+class NanoChatDecoderLayer(GradientCheckpointingLayer):
```
vasqu: Fully inheritable from https://github.com/huggingface/transformers/blob/818f7f10e40cbceb3ce8a8b9a1f174f6b9181da2/src/transformers/models/llama/modeling_llama.py#L296

Only needs to override the norms without hidden size (if we do it)

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+
+        residual = hidden_states
+        hidden_states = self.post_attention_layernorm(hidden_states)
+        hidden_states = self.mlp(hidden_states)
+        hidden_states = residual + hidden_states
+        return hidden_states, self_attn_weights
+
+
+@auto_docstring
+class NanoChatPreTrainedModel(LlamaPreTrainedModel):
```
vasqu, line 133: Mostly inheritable from https://github.com/huggingface/transformers/blob/818f7f10e40cbceb3ce8a8b9a1f174f6b9181da2/src/transformers/models/llama/modeling_llama.py#L342

Just override the weight init

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+            if name == "o_proj.weight":
+                nn.init.normal_(
+                    param,
+                    mean=0.0,
+                    std=self.config.initializer_range / math.sqrt(2 * self.config.num_hidden_layers),
+                )
+
+
+@auto_docstring
+class NanoChatModel(NanoChatPreTrainedModel):
```
vasqu: We can make use of https://github.com/huggingface/transformers/blob/818f7f10e40cbceb3ce8a8b9a1f174f6b9181da2/src/transformers/models/llama/modeling_llama.py#L361

Add `initial_norm` to the init (before the super call iirc)

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+        attention_mask: Optional[torch.Tensor] = None,
+        position_ids: Optional[torch.LongTensor] = None,
+        past_key_values: Optional[Cache] = None,
+        inputs_embeds: Optional[torch.FloatTensor] = None,
+        use_cache: Optional[bool] = None,
+        cache_position: Optional[torch.LongTensor] = None,
+        **kwargs: Unpack[TransformersKwargs],
+    ) -> BaseModelOutputWithPast:
+        output_attentions = kwargs.get("output_attentions", False)
+        output_hidden_states = kwargs.get("output_hidden_states", False)
```
vasqu: We don't need this, please copy https://github.com/huggingface/transformers/blob/818f7f10e40cbceb3ce8a8b9a1f174f6b9181da2/src/transformers/models/llama/modeling_llama.py#L378-L436

and only add the initial norm before the layers start

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,431 @@
...
+        return BaseModelOutputWithPast(
+            last_hidden_state=hidden_states,
+            past_key_values=past_key_values if use_cache else None,
+            hidden_states=all_hidden_states,
+            attentions=all_self_attns,
+        )
+
+
+@auto_docstring
+class NanoChatForCausalLM(NanoChatPreTrainedModel, GenerationMixin):
```
vasqu: Inherit from https://github.com/huggingface/transformers/blob/818f7f10e40cbceb3ce8a8b9a1f174f6b9181da2/src/transformers/models/llama/modeling_llama.py#L440

Only the forward needs a change due to the soft cap.

vasqu (commented):
We are not using modular to its fullest, I added some comments where we could apply modular:
- RoPE change rotate half again to align with our attention modules' implementation
- Attention -> Qwen3 Attention; forward might need one adjustment = order of ops (rope, rms norm) in the forward
- MLP -> Clip MLP
- Decoder Layer -> Llama Decoder Layer
- PreTrainedModel -> Llama PreTrainedModel (with different init weights)
- Model -> Llama (at least for the init)
- CausalLM -> Llama (mostly the same except the softcaping in the forward)

I tried to find the differences, but maybe one or two changes are additionally needed, e.g. some attribute has a different name, a norm needs a diffferent init etc

src/transformers/models/nanochat/modular_nanochat.py
vasqu: (only init, forward needs to change see below)

src/transformers/models/nanochat/modular_nanochat.py
burtenshaw: > We do not use layer_type in nanochat, so we can remove it after super init with del layer_type

I've implemented most of this, but I can't find a self.layer_type in `Qwen3Attention`

burtenshaw:
@vasqu Thanks again. I've integrated all these changes and there are no regression. Just one [nit](https://github.com/huggingface/transformers/pull/41634#discussion_r2445801535) I couldn't understand.

src/transformers/models/nanochat/configuration_nanochat.py
```diff
@@ -0,0 +1,163 @@
...
+        "layers.*.self_attn.o_proj": "rowwise",
+        "layers.*.mlp.fc1": "colwise",
+        "layers.*.mlp.fc2": "rowwise",
+    }
+
+    def __init__(
+        self,
+        vocab_size: int = 50304,
+        hidden_size: int = 768,
+        intermediate_size: int | None = None,  # TODO: default value
```
vasqu: I've added a few TODOs and FIXMEs in modular/modeling/config/conversion

src/transformers/models/nanochat/configuration_nanochat.py
```diff
@@ -0,0 +1,163 @@
...
+        self.initializer_range = initializer_range
+        self.use_cache = use_cache
+        self.logits_soft_cap = logits_soft_cap
+        self.qkv_bias = qkv_bias
+
+        # Validate the correctness of rotary position embeddings parameters
+        self.rope_parameters = rope_parameters
+        rope_theta = kwargs.get("rope_theta", 10000.0)
+        standardize_rope_params(self, rope_theta=rope_theta)
+        rope_config_validation(self)
```
vasqu: This is new, so I also expect the conversion script needing to be changed slightly :sweat:

src/transformers/models/nanochat/convert_nanochat_checkpoints.py
```diff
@@ -0,0 +1,223 @@
+# coding=utf-8
```
vasqu, line 1: Let's check if the script still work and apply some changes, e.g.:
- qkv_bias --> attention_bias
- RoPE

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,298 @@
...
+
+        attn_output = attn_output.reshape(*input_shape, -1).contiguous()
+        attn_output = self.o_proj(attn_output)
+        return attn_output, attn_weights
+
+
+class NanoChatMLP(CLIPMLP):
+    def __init__(self, config):
+        super().__init__(config)
+        # TODO: check if the module names are respected in the conversion script
```
vasqu: Not sure if this is already handled in the conversion script

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,298 @@
...
+        )
+
+        hidden_states = outputs.last_hidden_state
+        # Only compute necessary logits, and do not upcast them to float if we are not computing the loss
+        slice_indices = slice(-logits_to_keep, None) if isinstance(logits_to_keep, int) else logits_to_keep
+        logits = self.lm_head(hidden_states[:, slice_indices, :])
+
+        # Soft-cap the logits. The main difference to LlamaForCausalLM
+        cap = self.config.logits_soft_cap
+        logits = cap * torch.tanh(logits / cap)
```
vasqu: I changed this assuming it is used by all models so far (we don't like to add branches)

vasqu (commented):
Thanks for bearing with me @burtenshaw, I know I'm petty on a lot things here :sweat_smile: I added a few changes to make a bit more use out of modular. The nit was confusing because main diverged with the current main of this branch, you can now see what I do in the modular class.

I added a few TODOs and FIXMEs which will likely need to be addressed. These are mostly related to check for the conversion script and the config. Lmk if something is unclear

src/transformers/models/nanochat/modular_nanochat.py
vasqu: This was due to a divergince in the main branch, it existed in main but not your local branch :D no worries, see my review at https://github.com/huggingface/transformers/pull/41634#pullrequestreview-3360598177

burtenshaw:
Thanks again @vasqu ! I've responded to the inline TODOs. Two points:

- I definitely think we should stick with `karpathy/nanochat-d32` because it's his model and has more visibility. He's agreed to merge the weights via a hub PR.
- I changed the conversion script to work with the output of the `nanochat` repos configuration. This should be more compatible.

vasqu:
run-slow: gemma2, gemma3, gemma3n, vaultgemma

src/transformers/models/nanochat/convert_nanochat_checkpoints.py
```diff
@@ -0,0 +1,311 @@
...
+                "eos_token_id",
+                "pad_token_id",
+            ]:
+                if key in extra_config:
+                    config_kwargs[key] = extra_config[key]
+                # Handle legacy qkv_bias -> attention_bias conversion
+                elif key == "attention_bias" and "qkv_bias" in extra_config:
+                    config_kwargs[key] = extra_config["qkv_bias"]
+                elif key == "final_logit_softcapping" and "logits_soft_cap":
+                    config_kwargs[key] = extra_config["logits_soft_cap"]
```
vasqu: This is to align with gemma2 so we can use modular.

src/transformers/models/nanochat/convert_nanochat_checkpoints.py
```diff
@@ -0,0 +1,311 @@
...
+    if config.num_attention_heads % config.num_key_value_heads != 0:
+        print(f"Adjusting num_attention_heads from {config.num_attention_heads} to {config.num_key_value_heads}")
+        config.num_attention_heads = config.num_key_value_heads
+
+    print("Converting model...")
+    state_dict = {}
+    rename_map = {}
+
+    def assign(old_key: str, new_key: str) -> None:
+        global old_state, state_dict, rename_map
```
vasqu: Should work, haven't checked - ruff complained before as usual lol

Would be good to sanity check

tests/models/nanochat/test_modeling_nanochat.py
```diff
@@ -0,0 +1,207 @@
...
+        generated_tokens = generated_ids[0, inputs["input_ids"].shape[1] :]
+        generated_text = tokenizer.decode(generated_tokens, skip_special_tokens=True)
+
+        # Check that text was generated
+        self.assertGreater(len(generated_text), 0)
+        self.assertGreater(len(generated_tokens), 0)
+
+        # The model should generate a reasonable response (with greedy decoding this is deterministic)
+        # Expected: "The capital of France is Paris."
+        self.assertIn("Paris", generated_text)
```
vasqu: Let's generate a number of tokens and compare against the sequence instead. 1 Token is a bit wonky. 

You can look at Llama for example, imo something along 32 new tokens would suffice. Same goes for the other generation tests

vasqu (commented):
Finish line! I noticed that the soft capping is also done in gemma models, so I changed a few things there; if it doesnt work out I'll revert - no need to do anything on your side.

Remaining things to check:
- I didnt break the conversion script
- Expand the generation tests to have a longer sequence to check against (e.g. 32 tokens)
- Fix integration tests (?), see https://github.com/huggingface/transformers/actions/runs/18712904195/job/53365510466

For the integrations test, we use a10 gpus. If you confirm that everything works on your side, I can just update the values if needed

vasqu:
run-slow: vaultgemma

vasqu:
run-slow: gemma3

vasqu:
run-slow: gemma3

vasqu:
run-slow: nanochat

tests/models/nanochat/test_modeling_nanochat.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # Check expected mean logits (with tolerance for numerical variation)
+        EXPECTED_MEAN = torch.tensor([[-6.6695, -7.8112]])
+        torch.testing.assert_close(logits.mean(-1), EXPECTED_MEAN, rtol=1e-3, atol=0.02)
+
+        # Check first 10 logits at position [0,0,:10]
+        EXPECTED_SLICE = torch.tensor(
+            [-12.875, -13.125, -13.1875, -13.1875, -13.1875, -13.1875, -13.1875, -13.1875, -12.625, -4.40625]
+        )
+        torch.testing.assert_close(logits[0, 0, :10], EXPECTED_SLICE, rtol=1e-3, atol=0.2)
```
vasqu: Let's keep 1e-2 notation. Btw how big are the models in parameter size?

tests/models/nanochat/test_modeling_nanochat.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # Check expected mean logits (with tolerance for numerical variation)
+        EXPECTED_MEAN = torch.tensor([[-5.6836, -8.3400]])
+        torch.testing.assert_close(logits.mean(-1), EXPECTED_MEAN, rtol=1e-3, atol=0.1)
+
+        # Check first 10 logits at position [0,0,:10]
+        EXPECTED_SLICE = torch.tensor(
+            [-12.4375, -13.1875, -12.875, -13.1875, -13.1875, -13.1875, -13.1875, -13.1875, -12.0, -1.6172]
+        )
+        torch.testing.assert_close(logits[0, 0, :10], EXPECTED_SLICE, rtol=1e-3, atol=0.1)
```
vasqu: Let's keep the atols as low as possible; 1e-1 is too high. I can update on our runners if everything on your side is good

tests/models/nanochat/test_modeling_nanochat.py
burtenshaw: The d20 is 600m and the d32 is 1.8b.

tests/models/nanochat/test_modeling_nanochat.py
burtenshaw: Ok. I've updated the tolerances. Can you define the `EXPECTED_SLICE` from the runners because this runs on my hardware...

src/transformers/models/auto/tokenization_auto.py
```diff
@@ -468,6 +468,7 @@
         ("musicgen_melody", ("T5Tokenizer", "T5TokenizerFast" if is_tokenizers_available() else None)),
         ("mvp", ("MvpTokenizer", "MvpTokenizerFast" if is_tokenizers_available() else None)),
         ("myt5", ("MyT5Tokenizer", None)),
+        ("nanochat", (None, "PreTrainedTokenizerFast" if is_tokenizers_available() else None)),
```
vasqu, line 499: Added to auto, just to be sure can you cross check if that's correct

vasqu:
run-slow: nanochat

tests/models/nanochat/test_modeling_nanochat.py
```diff
@@ -0,0 +1,233 @@
...
+            )
+
+        # Decode only the generated tokens
+        generated_text = [
+            tokenizer.decode(generated_ids[0, inputs["input_ids"].shape[1] :], skip_special_tokens=True),
+            tokenizer.decode(generated_ids[1, inputs["input_ids"].shape[1] :], skip_special_tokens=True),
+        ]
+
+        EXPECTED_TEXT_COMPLETION = [
+            "The capital of France is Paris.PatrickPatrickPatrickPatrickPatrickPatrickPatrickPatrickPatrickPatrickPatrickPatrick",
```
vasqu: This is weird, might need to be checked if something went wrong

vasqu: It's due to the generation config having wrong token ids https://huggingface.co/nanochat-students/nanochat-d20/blob/main/generation_config.json

TODO: update the config

vasqu:
@ArthurZucker for core maintainer review when you are back :D

burtenshaw:
looking forward to the review. 

@vasqu I fixed the nits in configs. Can you rerun `slow` to be sure.

vasqu:
run-slow: nanochat

src/transformers/models/gemma2/modular_gemma2.py
vasqu, line 1: For visibility, these changes here and in the other gemma models are intentional:
- It wasn't properly using our features or introduced manual stuff we did before which is no longer necessary
- Enables nanochat to inherit from gemma2 properly

vasqu:
run-slow: nanochat

ydshieh:
run-slow: nanochat

ydshieh:
run-slow: nanochat

vasqu:
cc @ArthurZucker @Cyrilvallez for core maintainer when you have time

ydshieh:
Have to resolve the conflicts :-)

vasqu:
Not anymore @ydshieh :p

Cyrilvallez:
run-slow: nanochat

docs/source/en/model_doc/nanochat.md
```diff
@@ -0,0 +1,119 @@
...
+model_id = "karpathy/nanochat-d32"
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+tokenizer = AutoTokenizer.from_pretrained(model_id)
+model = AutoModelForCausalLM.from_pretrained(
+    model_id,
+    dtype=torch.bfloat16,
+    device_map="auto",
+)
+model.eval()
```
Cyrilvallez: Nit: model is already in `eval` mode as output of `from_pretrained`, so it can be safely removed!

src/transformers/models/auto/configuration_auto.py
```diff
@@ -724,6 +725,8 @@
         ("musicgen_melody", "MusicGen Melody"),
         ("mvp", "MVP"),
         ("myt5", "myt5"),
+        ("nanochat", "NanoChat"),
+        ("nanochat", "NanoChatForCausalLM"),
```
Cyrilvallez: We need only the name here, it's not an AutoMapping!
```suggestion
        ("nanochat", "NanoChat"),
```

src/transformers/models/gemma2/modular_gemma2.py
```diff
@@ -439,30 +430,16 @@ def forward(
...
-        output_hidden_states: Optional[bool] = None,
         cache_position: Optional[torch.LongTensor] = None,
         **kwargs: Unpack[TransformersKwargs],
     ) -> BaseModelOutputWithPast:
-        output_attentions = output_attentions if output_attentions is not None else self.config.output_attentions
-        output_hidden_states = (
-            output_hidden_states if output_hidden_states is not None else self.config.output_hidden_states
-        )
-        use_cache = use_cache if use_cache is not None else self.config.use_cache
-
```
Cyrilvallez, line 452: Very nice cleanup of Gemma2 for the integration! 🚀❤️

src/transformers/models/nanochat/configuration_nanochat.py
```diff
@@ -0,0 +1,164 @@
...
+        intermediate_size: int | None = 8192,
+        num_hidden_layers: int = 12,
+        num_attention_heads: int = 6,
+        num_key_value_heads: int | None = None,
+        max_position_embeddings: int = 2048,
+        hidden_act: str = "relu2",
+        attention_dropout: float = 0.0,
+        rms_norm_eps: float = 1e-6,
+        initializer_range: float = 0.02,
+        rope_parameters: RopeParameters | dict[RopeParameters] | None = None,
```
Cyrilvallez: Probably you meant this? I.e. a simple dict? Our other type hints are wrong, just noticed it (`RopeParameters` is itself a `dict` already)
```suggestion
        initializer_range: float = 0.02,
        rope_parameters: RopeParameters | dict | None = None,
```

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,247 @@
...
+
+class NanoChatRotaryEmbedding(LlamaRotaryEmbedding):
+    pass
+
+
+def rotate_half(x):
+    """Rotates half the hidden dims of the input with flipped signs for NanoChat."""
+    x1 = x[..., : x.shape[-1] // 2]
+    x2 = x[..., x.shape[-1] // 2 :]
+    return torch.cat((x2, -x1), dim=-1)
```
Cyrilvallez, line 57: Not needed, modular should pick it up automatically!

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,247 @@
...
+        query_states = self.q_proj(hidden_states).view(hidden_shape).transpose(1, 2)
+        key_states = self.k_proj(hidden_states).view(hidden_shape).transpose(1, 2)
+        value_states = self.v_proj(hidden_states).view(hidden_shape).transpose(1, 2)
+
+        cos, sin = position_embeddings
+        query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)
+
+        # RoPE -> Norm (instead of usual Norm -> RoPE)
+        query_states = self.q_norm(query_states)
+        key_states = self.k_norm(key_states)
```
Cyrilvallez, line 90: Hummm, here due to the added norms, I think we should replicate the layers in the `tp_plan` unfortunately... Otherwise it does not look like the math is equivalent! Did you try running it with tp by any chance?

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,247 @@
...
+    def _init_weights(self, module: nn.Module) -> None:
+        PreTrainedModel._init_weights(module)
+
+        for name, param in module.named_parameters():
+            if name == "o_proj.weight":
+                nn.init.normal_(
+                    param,
+                    mean=0.0,
+                    std=self.config.initializer_range / math.sqrt(2 * self.config.num_hidden_layers),
+                )
```
Cyrilvallez: We absolutely cannot do that unfortunately, as we will iterate over all params for every module. If we need to change the default scheme, we need to rewrite _init_weights completely. Here, you could add a check isinstance(module, NanoChatAttention)`, and init correctly the o_proj from this

src/transformers/models/nanochat/modular_nanochat.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+@auto_docstring
+class NanoChatModel(LlamaModel):
+    def __init__(self, config: NanoChatConfig):
+        super().__init__(config)
+
+        self.initial_norm = NanoChatRMSNorm(eps=config.rms_norm_eps)
+        self.norm = NanoChatRMSNorm(eps=config.rms_norm_eps)
```
Cyrilvallez: We actually don't need 2 different norms here, as they don't have weights anyway!

Cyrilvallez (commented):
Amazing work! Super clean, and you also refactored gemmas super nicely 🤗🤗🚀
It's really almost ready to be merged, only 2 frictions points are the `_init_weights`, and the tp_plan currently! They need to be changed a bit for correctness/efficiency! Let me know if something is unclear about that!

Cyrilvallez:
And super sorry for the delay on this! Past 2 weeks have been really crazy! 😬

src/transformers/models/nanochat/modular_nanochat.py
vasqu: Notice the flipped signs (also as comment), it's different from other models doing `torch.cat((-x2, x1), dim=-1)` - so it can't be inherited

Cyrilvallez: Oh indeed, was a bit too fast on this one! Looks so similar to what I'm used to hahaha! Feel free to disregard!

src/transformers/models/nanochat/modular_nanochat.py
burtenshaw: Ok. Just so I follow, do you mean check before iterating `named_parameters` or just check before `nn.init.normal_` like so:

```suggestion
    def _init_weights(self, module: nn.Module) -> None:
        PreTrainedModel._init_weights(module)

        for name, param in module.named_parameters():
            if isinstance(module, NanoChatAttention) and name == "o_proj.weight":
                nn.init.normal_(
                    param,
                    mean=0.0,
                    std=self.config.initializer_range / math.sqrt(2 * self.config.num_hidden_layers),
                )
```

Cyrilvallez: No, I meant that we cannot iterate on params at all as we already iterate on modules haha! We can either do the following:
```python
def _init_weights(self, module: nn.Module) -> None:
        PreTrainedModel._init_weights(module)
        if isinstance(module, NanoChatAttention):
             init.normal_(module.o_proj, ...)
```
but it's a bit inefficient as every o_proj will be first initialized in `PreTrainedModel._init_weights(module)`, then again in the 2nd call, or we can do
```python
def _init_weights(self, module: nn.Module) -> None:
        if isinstance(module, NanoChatAttention):
             init.normal_(module.o_proj, ...)
             init.normal_(module.k_proj, ...)
             init.normal_(module.q_proj, ...)
             ....
        elif isinstance(module, NanoChatMLP):
               init.normal_(module.fc1, ....)
               ....
        elif ...
```
which will avoid the double init issue, but needs to be much more verbose because we cannot rely on `isinstance(module, Linear)` anymore as not all Linear have the same

TBH, as the `o_proj` are expected to be small enough in this case for this model, I would go with option 1 even if it's technically less efficient

burtenshaw:
> Amazing work! Super clean, and you also refactored gemmas super nicely 🤗🤗🚀 It's really almost ready to be merged, only 2 frictions points are the `_init_weights`, and the tp_plan currently! They need to be changed a bit for correctness/efficiency! Let me know if something is unclear about that!

@Cyrilvallez Thanks for the review! I've responded to all those changes and added suggestions.

vasqu:
run-slow: nanochat

burtenshaw:
@Cyrilvallez @vasqu Thanks both. Looks like tests are good and I've merged main. When do you think this can be merged?

src/transformers/models/nanochat/modular_nanochat.py
vasqu: Wouldn't gemma3 also need a replicate then (even if rope -> norm)? I think the plan before was correct, we should revert 35570e3 IMO

Cyrilvallez: Humm no, I think it's just Gemma3 plan being broken indeed haha 🥲 Since we take the mean over last dim in the norm, it cannot really be equivalent with or without TP.... Let me know if you try it empirically on main! In the meantime I just commited here to fix the tp_plan! Worst case, it's just less efficient but this one cannot be wrong!

Cyrilvallez (approved):
Alright, amazing! Thanks a lot! I pushed the last details and merging it now!! Congrats! 🤗🤗

src/transformers/models/nanochat/modular_nanochat.py
vasqu: Yea makes sense, I'll try to check it out if I find time. Definitely cant be wrong that way 😄

burtenshaw:
Thanks again @Cyrilvallez @vasqu . It was a pleasure working with you.

vasqu:
Likewise @burtenshaw :)

PR #41636, opened 2025-10-15
Adjust device logging level and add minor fixes

mario-koddenbrock:
This commit addresses a noisy warning and improves the robustness of the base pipeline implementation.

- The device placement message in the pipeline base class has been changed from a `warning` to a `debug` log. This reduces log noise for users who are aware of their device setup, while still providing the information for debugging purposes.

- Additionally, potential `UnboundLocalError` exceptions in the `_pad` and `check_model_type` functions have been prevented by initializing variables before their conditional assignment.


Rocketknight1 (approved):
All good fixes! I need to refactor this to simplify the separate logic for each `dim` value, though - that shouldn't be necessary.

PR #41642, opened 2025-10-16
Fix confusing cls assignment

cyyever:
# What does this PR do?

Don't overwrite `cls` incidentally in `__new__`.


Rocketknight1 (approved):
Yeah, this is quite confusing in the original code, and the new code should be equivalent

PR #41644, opened 2025-10-16
Fix Pylint warnings

cyyever:
# What does this PR do?

This PR fixes some Pylint warnings.

### DIFF
--- a/tests/models/bart/test_modeling_bart.py
+++ b/tests/models/bart/test_modeling_bart.py
@@ -514,6 +514,7 @@ def assert_tensors_close(a, b, atol=1e-12, prefix=""):
     try:
         if torch.allclose(a, b, atol=atol):
             return True
+        raise Exception
     except Exception:
         pct_different = (torch.gt((a - b).abs(), atol)).float().mean().item()
         if a.numel() > 100:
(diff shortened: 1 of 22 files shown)


tests/models/bart/test_modeling_bart.py
```diff
@@ -514,7 +514,6 @@ def assert_tensors_close(a, b, atol=1e-12, prefix=""):
     try:
         if torch.allclose(a, b, atol=atol):
             return True
-        raise
```
Rocketknight1, line 517: This doesn't look like it's equivalent! The `raise` can get called when `a` and `b` are not `allclose`

tests/models/blenderbot/test_modeling_blenderbot.py
```diff
@@ -270,7 +270,6 @@ def assert_tensors_close(a, b, atol=1e-12, prefix=""):
     try:
         if torch.allclose(a, b, atol=atol):
             return True
-        raise
```
Rocketknight1, line 273: Same here (and in several other instances of this pattern)

Rocketknight1 (commented):
Mostly looks good, but I think removing the `raise` line is not equivalent, right?

cyyever:
@Rocketknight1 
Indeed raise should follow an Exception, otherwise the [documentation](https://docs.python.org/3.14/reference/simple_stmts.html#raise) says
```
If no expressions are present, raise re-raises the exception 
that is currently being handled, 
which is also known as the active exception.
```
IMO we should add the missing Exception.

Rocketknight1 (approved):
Yes, looks good now!

cyyever:
@Rocketknight1 CI is now green.

Rocketknight1 (approved):

PR #41645, opened 2025-10-16
Fix bark after #41445

ydshieh:
# What does this PR do?

Some Bark model classes are not `GenerationMixin` and have their own `generation` method.

#41445 makes change to `GenerationMixin` and those changes are not applied to those Bark model classes, then causes the bark tests to fail. One example is

> tests/models/bark/test_modeling_bark.py::BarkModelIntegrationTests::test_generate_batching

This PR adds `GenerationMixin` to those models and the tests pass now.


src/transformers/models/bark/modeling_bark.py
```diff
@@ -1315,7 +1315,7 @@ def generate(
     output sound according to specific predefined voice.
     """
 )
-class BarkModel(BarkPreTrainedModel):
+class BarkModel(BarkPreTrainedModel, GenerationMixin):
```
ydshieh, line 1321: I didn't change `class BarkFineModel(BarkPreTrainedModel):` however. Too many tests fail when I enable it.

ArthurZucker (commented):
its good that it fixes is, tho a bit weird as the bark model is not semantically a gernerate model. IDT that's the best solution 👀

ydshieh:
> bark model is not semantically a gernerate model

I agree, but it has a `generate` method which makes it even confused ...

(and it definitely try to use generation config )

ydshieh:
As @gante wrote in the codebase,

```
    def can_generate(cls) -> bool:
        # Bark has a unique model structure, where the external class (`BarkModel`) doesn't need to inherit from
        # `GenerationMixin` (it has a non-standard generation method), but one of the internal models do
        # (`BarkSemanticModel`). This means that the base `can_generate()` will return `False`, but we need to
        # override it so as to do `GenerationConfig` handling in multiple parts of the codebase.
        return True
```

we see `BarkModel` is: 

  - has `generate` method (although previously not inherits from `GenerationMixin`)
  - set `can_generate = True` in order to handle `GenerationConfig`

So after the change #41445, the fix of the failing tests in this PR seems to be natural IMO.

ydshieh:
Merged as approved offline on slack

PR #41646, opened 2025-10-16
Use | for Optional and Union typing

cyyever:
# What does this PR do?

This PR replace `Optional` and `Union` typing with new `|` notation in files except the src ones.


Rocketknight1 (approved):
Yep, happy to use the new syntax now we're dropping 3.9 support!

PR #41647, opened 2025-10-16
[Fix] Deepseek V3 expert bias routing

fjosw:
# What does this PR do?

By chance we noticed that https://github.com/huggingface/transformers/pull/40132 seems to have introduced a bug in the Deepseek V3 routing implementation: The Deepseek-V3 technical report explicitly states
> Note that the bias term is only used for routing. The gating value, which will be multiplied with
the FFN output, is still derived from the original affinity score

This was the case in transformers until https://github.com/huggingface/transformers/pull/40132 which changed the routing code such that the gating values are now derived from the tensor with the added bias term. I wrote a quick fix for the Deepseek-V3 model in this PR, not sure if other models are also affected. Can you please have a look @ArthurZucker and confirm that this is indeed a bug?

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/transformers/models/deepseek_v3/modeling_deepseek_v3.py
+++ b/src/transformers/models/deepseek_v3/modeling_deepseek_v3.py
@@ -178,7 +178,9 @@ class DeepseekV3MoE(nn.Module):
         router_logits = router_logits.sigmoid()
         router_logits_for_choice = router_logits + self.gate.e_score_correction_bias
         group_scores = (
-            router_logits_for_choice.view(-1, self.n_group, self.n_routed_experts // self.n_group).topk(2, dim=-1)[0].sum(dim=-1)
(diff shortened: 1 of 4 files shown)


ArthurZucker (approved):
Thanks for catching! Can confirm, we previously gathered on the `scores` with the index.

ArthurZucker (commented):
can you just run `make fix-copies` that will fix dependant models

fjosw:
Thanks @ArthurZucker, I just pushed the `fix-copies` changes. Are you aware of any not directly dependent models that use Deepseek style expert bias routing and could have been affected by the refactor from a few weeks ago?

ArthurZucker:
They are the ones that inherit from deepseek! 
Can you just run `make style` and you should be good to go!

PR #41648, opened 2025-10-16
[v5] Delete legacy chat template saving

zucchini-nlp:
# What does this PR do?

As per title


Rocketknight1 (approved):
Yes, LGTM! Cleanly removes all the legacy code paths.

PR #41650, opened 2025-10-16
Remove randomly added script

Cyrilvallez:
# What does this PR do?

cc @zucchini-nlp you added it in https://github.com/huggingface/transformers/pull/41474. Please be very careful about adding commits after we approve a PR, and always review the diffs of the PR before merging to avoid such issues!


zucchini-nlp:
Sorry about that

PR #41653, opened 2025-10-16
Double router compute?

molbap:
# What does this PR do?

Not sure what the intended fix is (force f32 or not) but currently the first is ignored

### DIFF
--- a/src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
+++ b/src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
@@ -116,7 +116,6 @@ class DeepseekV2Moe(nn.Module):
     def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
         residuals = hidden_states
         orig_shape = hidden_states.shape
-        router_logits = nn.functional.linear(hidden_states.type(torch.float32), self.gate.weight.type(torch.float32))
         router_logits = self.gate(hidden_states)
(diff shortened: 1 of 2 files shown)


src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
```diff
@@ -116,7 +116,6 @@ def route_tokens_to_experts(self, router_logits):
     def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
         residuals = hidden_states
         orig_shape = hidden_states.shape
-        router_logits = nn.functional.linear(hidden_states.type(torch.float32), self.gate.weight.type(torch.float32))
         router_logits = self.gate(hidden_states)
```
vasqu: Looks like it's the other way around, see https://github.com/huggingface/transformers/blob/a543095c9967ddd3dd0915ffa19df412a4788a44/src/transformers/models/deepseek_v2/modeling_deepseek_v2.py#L67-L68

molbap: alright, so intended force of f32? updating

vasqu: Yup, exactly

vasqu (approved):
LGTM, thank you :)

ArthurZucker (approved):
ty

PR #41655, opened 2025-10-16
Switch to CB if cache_implementation == paged

remi-or:
This PR fixes one of the issues raised in #41501

This PR adds `paged` as a static cache implementation and switches from `generate` to `generate_batch` if it's the chosen implementation. It also raises errors if an argument for `generate` is passed to `generate_batch` but is not supported, and warn the user that `generate_batch` is used. 
The interface for the user stays the same, as the sample script below shows:
```
import os
import torch

from transformers import AutoModelForCausalLM, AutoTokenizer, GenerationConfig


if __name__ == '__main__':
    model_path = "Qwen/Qwen3-0.6B"

    tokenizer = AutoTokenizer.from_pretrained(
        model_path,
        trust_remote_code=True,
        add_bos_token=False,
        add_eos_token=False,
    )

        model_path,
        dtype=torch.bfloat16,
        trust_remote_code=True,
        device_map="cuda",
    ).eval()

    prompt="How are you today?"
    chat_prompt=tokenizer.apply_chat_template(
        [{"role" : "user", "content" : prompt}],
        tokenize=False,
        add_generation_prompt=True,
        enable_thinking=False,
    )

    generation_config=GenerationConfig(
        do_sample=True,
        temperature=1.0,
        top_p=1.0,
        # num_return_sequences=8,
        max_new_tokens=100,
        eos_token_id=tokenizer.eos_token_id,
        pad_token_id=tokenizer.pad_token_id,
        use_cache=True,
        return_dict_in_generate=True,
        output_logits=True,
    )

    chat_prompt_ids = tokenizer(chat_prompt, return_tensors="pt")["input_ids"].to(model.device)

    print("Classic generate")
    output = model.generate(input_ids=chat_prompt_ids, generation_config=generation_config)[0]
    output_text = tokenizer.batch_decode(output[:, chat_prompt_ids.shape[1]:], skip_special_tokens=True)
    print(output_text)

    print("\nWith paged cache")
    output = model.generate(input_ids=chat_prompt_ids, generation_config=generation_config, cache_implementation="paged")[0]
    output_text = tokenizer.batch_decode(output[:, chat_prompt_ids.shape[1]:], skip_special_tokens=True)
    print(output_text)

```
which outputs
```
Classic generate
`generation_config` default values have been modified to match model-specific defaults: {'temperature': 0.6, 'top_k': 20, 'top_p': 0.95, 'bos_token_id': 151643}. If this is not desired, please set these values explicitly.
["Hello! I'm here to help you with anything you need. How can I assist you today?"]

With paged cache
Detected cache_implementation=paged: switching to continuous batching. You should consider using generate_batch directly instead.
No behavior specified for use_cuda_graph, defaulting to self.use_cuda_graph = False because self.model.config._attn_implementation = 'paged|sdpa'. If you want to save memory, turn off cuda graphs, but they can improve performances.
Solving 1 requests: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  2.16request/s]
["Hello! I'm here to help you with anything you need. How can I assist you today? 😊"]
```

### DIFF
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -2393,8 +2393,7 @@ class GenerationMixin(ContinuousMixin):
             return custom_generate_function(model=self, **generate_arguments)
 
         # 0.b. If requested, switched to continuous batching generation
-        if kwargs.get("cache_implementation") == "paged_cache":
-
+        if kwargs.get("cache_implementation") == "paged":
             # generate_batch expects a list of lists of ints, so we create it from the inputs or input_ids
             inputs = inputs if inputs is not None else kwargs.get("input_ids")
             if inputs is None:
@@ -2436,12 +2435,23 @@ class GenerationMixin(ContinuousMixin):
                 logger.warning(f"synced_gpus is not ignored for continuous batching. Got {synced_gpus = }")
 
             # switch to CB and signal it
-            logger.warning("Detected cache_implementation=paged_cache, switching to continuous batching.")
-            return self.generate_batch(
+            logger.warning(
+                "Detected cache_implementation=paged: switching to continuous batching. You should consider using "
+                "generate_batch directly instead."
+            )
+            outputs = self.generate_batch(
                 inputs=inputs,
                 generation_config=self._prepare_generation_config(generation_config, use_model_defaults, **kwargs)[0],
                 **kwargs,
             )
+            sequences = [
+                outputs[f"req_{i}"].prompt_ids + outputs[f"req_{i}"].generated_tokens for i in range(len(outputs))
+            ]
+
+            # To use the same indexing (outputs[0]) as the regular generate method, we unsqueeze the tensor
+            sequences_as_tensor = torch.tensor(sequences, dtype=torch.long, device=self.device)
+            sequences_as_tensor = sequences_as_tensor.unsqueeze(0)
+            return sequences_as_tensor
 
         # 1. Handle kwargs, `generation_config`, validate them and obtain generation mode
         generation_mode_kwargs = self._extract_generation_mode_kwargs(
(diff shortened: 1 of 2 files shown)


src/transformers/generation/utils.py
```diff
@@ -2392,6 +2392,67 @@ def generate(
...
+                raise NotImplementedError(
+                    f"negative_prompt_attention_mask is not supported for continuous batching. Got {negative_prompt_attention_mask = }"
+                )
+
+            # others are ignored
+            if synced_gpus is not None:
+                logger.warning(f"synced_gpus is not ignored for continuous batching. Got {synced_gpus = }")
+
+            # switch to CB and signal it
+            logger.warning(
```
McPatate: perhaps we should move this log higher for the error msgs to make sense, e.g. "not supported for CB" isn't clear when calling `generate` if we don't show "switchong to CB" log before

McPatate (approved):
not the best suited to review, CB code calling looks ok!

src/transformers/generation/utils.py
remi-or: makes sense, doing that

src/transformers/generation/utils.py
```diff
@@ -2392,6 +2392,67 @@ def generate(
             )
             return custom_generate_function(model=self, **generate_arguments)
 
+        # 0.b. If requested, switched to continuous batching generation
```
ArthurZucker, line 2395: nice!

src/transformers/generation/utils.py
ArthurZucker: missing beam search and etc as well

ArthurZucker: and num return sequences

ArthurZucker (approved):

src/transformers/generation/utils.py
remi-or: Added num_return_sequences and num_beams, that should cover it

PR #41657, opened 2025-10-16
Erroring when KernelConfig is passed without use_kernels = True

MekkCyber:
# What does this PR do?

as per title


src/transformers/modeling_utils.py
```diff
@@ -4426,6 +4426,11 @@ def from_pretrained(
                     "loaded from GGUF files."
                 )
 
+        if kernel_config is not None and not use_kernels:
+            raise ValueError(
+                "kernel_config is only supported when use_kernels is set to True, please set use_kernels to True or remove kernel_config."
+            )
```
SunMarc, line 4436: maybe we can set `use_kernels` to `True` automatically with one passes `kernel_config`

SunMarc (approved):
Thanks ! Just a nit suggestion

src/transformers/modeling_utils.py
MekkCyber: why not! done

PR #41658, opened 2025-10-16
improve `utils/check_bad_commit.py`

ydshieh:
# What does this PR do?

This script (used to find which commits fail certain failing tests on our CI) fails recently, due to the fact that the warning `sys:1: DeprecationWarning: builtin type swigvarlink has no __module__ attribute` was written to `stderr` and we use `stderr` to check if there is error.

This is not a good approach, we should instead check `result.returncode`. Also a few improvements added along this here


utils/check_bad_commit.py
```diff
@@ -103,22 +100,11 @@ def find_bad_commit(target_test, start_commit, end_commit):
...
-            pytest_error = pytest_errors[0]
-            index = pytest_error.find("pytest failed to run: ")
-            index += len("pytest failed to run: ")
-            pytest_error = pytest_error[index:]
-            error_msg += f"pytest error: {pytest_error}"
-
-        raise ValueError(error_msg)
+        error_msg = f"Error when running git bisect:\nbash error: {result.stderr}\nbash output:\n{result.stdout}\nset `bad_commit` to `None`."
+        print(error_msg)
+        return None
```
ydshieh, line 108: Let's set it to `None` but not fail the whole script. We still want to continue to check the next failing test and find out which commit cause it. 

Otherwise, we would not receive the reprot (the github action job using this script will fail)

utils/check_bad_commit.py
```diff
@@ -45,22 +45,19 @@ def create_script(target_test):
...
         exit(0)
     else:
-        print(f"pytest failed to run: {{result.stderr}}")
+        print(f"pytest gets unknown error: {{result.stderr}}")
         exit(-1)
-elif f"FAILED {target_test}" in result.stdout:
-    print("test failed")
-    exit(2)
 
 exit(0)
```
vasqu, line 63: Nit: would we add `print("test passed")` before here now

utils/check_bad_commit.py
vasqu: Fair enough, let's see how many of the error logs above we get :smile:

vasqu (approved):
Thx, lgtm overall

Do we want this to be a permanent solution or is it a workaround for now? Let's see how it fares

utils/check_bad_commit.py
ydshieh: you mean before `exit(0)`? Good point, it's not always `passed`, it could be `SKIPPED` or anything that is not `FAILED` and `ERROR`.

I will add a print.

utils/check_bad_commit.py
ydshieh: It's not much I believe, except those

DeprecationWarning: builtin type swigvarlink has no __module__ attribute

being thought as `error` (before this PR), which is not.

ydshieh:
> Do we want this to be a permanent solution or is it a workaround for now? Let's see how it fares

Kind of, it is improved along the time: it's use `git bisect` (with subprocess) which calls a  script (created on the fly) (a nested subprocess), so a bit complicated to handle the scenario. 

Later, I will need to handle the flaky tests too to make the results more reliable.

PR #41660, opened 2025-10-16
Add optional logits slicing support to GPT2LMHeadModel

abdulrehhhhman:
This PR adds support for an optional `logits_to_keep` argument in GPT2LMHeadModel.
It allows users to slice logits output for memory-efficient inference or analysis.

Example:
```python
outputs = model(**inputs, logits_to_keep=2)

### DIFF
--- a/src/transformers/models/gpt2/modeling_gpt2.py
+++ b/src/transformers/models/gpt2/modeling_gpt2.py
@@ -820,6 +820,11 @@ class GPT2LMHeadModel(GPT2PreTrainedModel, GenerationMixin):
 
         slice_indices = slice(-logits_to_keep, None) if isinstance(logits_to_keep, int) else logits_to_keep
         logits = self.lm_head(hidden_states[:, slice_indices, :])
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by abdulrehhhhman** [comment]
Hi @huggingface-team 👋
This PR introduces an optional logits_to_keep argument for GPT2LMHeadModel.forward, enabling selective slicing of output logits for memory-efficient inference and analysis.
It is backward-compatible and passes the existing GPT-2 test suite locally.
Thanks for reviewing!

**Comment by Rocketknight1** [comment]
The two lines above already do this!

```python
slice_indices = slice(-logits_to_keep, None) if isinstance(logits_to_keep, int) else logits_to_keep
logits = self.lm_head(hidden_states[:, slice_indices, :])
```

abdulrehhhhman:
Thanks for the clarification! That makes sense — I see it’s already handled by those lines.
I learned a lot from digging into GPT2LMHeadModel’s internals

PR #41662, opened 2025-10-16
Small changes to benchmarking script

remi-or:
This PR:
- adds throughput to the final pretty print at the end of the benchmark runs
- adds the information of the output shape to the decoded text of the output (usefull to double check if throughput is right)
- changes the behavior of the `run_benchmarks.py` script so that it will only run 3 configs unless instructed to do otherwise

This way, anyone can run a "quick"
```
python benchmark_v2/run_benchmarks.py --model-id "meta-llama/Meta-Llama-3-8B" -b 32 -s 128 -n 256
```
after changing the `generate` code or some model specific code to check perf did not take a big hit. 
Thanks @SunMarc for the suggestion!


SunMarc (approved):
Thanks a lot !

PR #41663, opened 2025-10-16
[Fix] dots1 expert bias routing

fjosw:
# What does this PR do?

This is a follow up to https://github.com/huggingface/transformers/pull/41647. I went over the model code again and noticed that the dots1 routing apparently also broke in https://github.com/huggingface/transformers/pull/40132. dots1 seems to be very close in architecture to Deepseek V3 but not explcitly dependent. Not sure if there is a more elegant way to fix this?

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker


fjosw:
Reminder: I think the dots1 model is still broken on main @ArthurZucker, not sure if this PR is the best way to address this, though.

vasqu (approved):
LGTM, I double checked and this looks correct to me. The selection is affected by the e_score_bias but not the final logits that are then selected. 

Imo, we could refactor this later on to be within the router instead! (including deepseek). cc @3outeille since you worked a lot on MoEs recently it might be interesting

PR #41665, opened 2025-10-16
Fix dtype casting with quantization

Cyrilvallez:
# What does this PR do?

We should always trust the dtype of the params in the meta model, except when the parameter is going to be quantized!
Got broken in https://github.com/huggingface/transformers/pull/41548, where we stopped looking at the `keep_in_fp_32_regex`, to instead trust the model dtypes directly!


ydshieh (approved):
LGTM

PR #41667, opened 2025-10-16
Simplify pipeline padding logic

Rocketknight1:
Follow-up to #41636 

When reviewing that PR I noticed the padding logic was cluttered and much more complex than it needed to be. We can eliminate a lot of code by just constructing shapes as lists and trusting in implicit slice dimensions


gante (approved):
LGTM 🤗

PR #41671, opened 2025-10-16
Fix graphormer model compilation with Cython 3.1.4

alexmalyshev:
Hitting this kind of error when running:

```
cython src/transformers/models/deprecated/graphormer/algos_graphormer.pyx
```

```
Error compiling Cython file:
------------------------------------------------------------
...
    (nrows, ncols) = path.shape
    assert nrows == ncols
    cdef unsigned int n = nrows
    cdef unsigned int max_dist_copy = max_dist

    path_copy = path.astype(long, order='C', casting='safe', copy=True)
                            ^
------------------------------------------------------------
src/transformers/models/deprecated/graphormer/algos_graphormer.pyx:88:28: undeclared name not builtin: long
```

This appears to have changed between cython==3.0 and cython==3.1.  AFAICT the correct type to use here would be `int`.  Switching to it makes the command succeed and generate an algos_graphormer.c file.

### DIFF
--- a/src/transformers/models/deprecated/graphormer/algos_graphormer.pyx
+++ b/src/transformers/models/deprecated/graphormer/algos_graphormer.pyx
@@ -85,8 +85,8 @@ def gen_edge_input(max_dist, path, edge_feat):
     cdef unsigned int n = nrows
     cdef unsigned int max_dist_copy = max_dist
 
-    path_copy = path.astype(long, order='C', casting='safe', copy=True)
-    edge_feat_copy = edge_feat.astype(long, order='C', casting='safe', copy=True)
+    path_copy = path.astype(int, order='C', casting='safe', copy=True)
+    edge_feat_copy = edge_feat.astype(int, order='C', casting='safe', copy=True)
     assert path_copy.flags['C_CONTIGUOUS']
     assert edge_feat_copy.flags['C_CONTIGUOUS']
 


Rocketknight1:
I don't think we can diagnose `cython` bugs at all! In general there is very little point in running Transformers code with Cython because almost all of the performance-intensive work happens in `torch` anyway, which is already heavily optimized C++/CUDA code.

alexmalyshev:
@Rocketknight1 Relevant to performance or not, it'd be unfortunate to leave this Cython file hanging around in a non-buildable state with the current Cython compiler.  I get that it's in the deprecated directory, but the fix seems simple enough.

From our end, we're not actually using this file but we do have an aggressive CI system that tries to build all Cython sources to validate they're in a good state.  I'm working around that internally but I figure it's worth fixing upstream.

Rocketknight1:
Ah, I misread, I'm sorry! I didn't realize we actually had Cython files hiding in the `deprecated` folder of our codebase. My fault for skimming too quickly.

If it's causing you build trouble then a PR to replace `long` with `int` is welcome.

alexmalyshev:
All this PR needs is an accept and a merge then :)

Rocketknight1 (approved):
Yep, happy to merge if there's nothing else to add!

PR #41673, opened 2025-10-16
[FIX]: CLIP support for flash-attention-3

### LINKED ISSUE #41668: CLIP incompatible with Flash Attention 3
### System Info

- `transformers` version: 4.57.0
- Platform: Linux-6.14.0-33-generic-x86_64-with-glibc2.39
- Python version: 3.12.3
- Huggingface_hub version: 1.0.0.rc4

### Who can help?

@ArthurZucker @Cyrilvallez

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

```
from diffusers import EulerDiscreteScheduler, StableDiffusionXLPipeline
import torch
import os

model_path =os.getenv('MODEL_PATH')
print(model_path)
pipe = StableDiffusionXLPipeline.from_pretrained(
    model_path,
    variant="fp16",
    torch_dtype=torch.float16
).to("cuda")
pipe.fuse_qkv_projections()

pipe.text_encoder.config._attn_implementation = "flash_attention_3"
pipe.text_encoder_2.config._attn_implementation = "flash_attention_3"

prompt = ["A city at night with people walking around."] * 8
image = pipe(prompt, num_inference_steps=1).images[0]
```

Input into `_upad_input`
```
Starting with query_length: 77
query_layer shape: torch.Size([8, 77, 12, 64])
key_layer shape: torch.Size([8, 77, 12, 64])
value_layer shape: torch.Size([8, 77, 12, 64])
attention_mask shape: torch.Size([8, 1, 77, 77])
```

output from `_upad_input`

```
q shape: torch.Size([23408, 12, 64])
k shape: torch.Size([23408, 12, 64])
v shape: torch.Size([23408, 12, 64])
cu_seq_lens_q: torch.Size([8, 1, 78])
cu_seq_lens_k: torch.Size([8, 1, 78])
max_length_q: 0
max_length_k: 0

```

### Expected behavior

I'm trying to enable Flash Attention 3 for the CLIP encoder in SDXL. 

Is the line in the forward pass supposed to include FA3 as well?

https://github.com/huggingface/transformers/blob/19df66dcba7690fab0b82105c154685c63016a21/src/transformers/models/clip/modeling_clip.py#L322

Currently when the attention_implementation is set to FlashAttention3 the attention_mask is set to the the causal_attention_mask which is 4D and when passed into [_upad_input](https://github.com/huggingface/transformers/blob/19df66dcba7690fab0b82105c154685c63016a21/src/transformers/modeling_flash_attention_utils.py#L237) which only supports a 2D attention_mask leads to garbage output https://github.com/huggingface/transformers/blob/19df66dcba7690fab0b82105c154685c63016a21/src/transformers/modeling_flash_attention_utils.py#L258

akashpalla:
# What does this PR do?

Fixes #41668, Enabling CLIP to work with Flash attention 3

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker @Cyrilvallez

### DIFF
--- a/src/transformers/models/clip/modeling_clip.py
+++ b/src/transformers/models/clip/modeling_clip.py
@@ -22,7 +22,7 @@ import torch
 from torch import nn
 
 from ...activations import ACT2FN
-from ...modeling_attn_mask_utils import _create_4d_causal_attention_mask, _prepare_4d_attention_mask
+from ...masking_utils import create_causal_mask, create_bidirectional_mask
 from ...modeling_layers import GradientCheckpointingLayer
 from ...modeling_outputs import BaseModelOutput, BaseModelOutputWithPooling, ImageClassifierOutput
 from ...modeling_utils import ALL_ATTENTION_FUNCTIONS, PreTrainedModel
@@ -319,9 +319,10 @@ class CLIPAttention(nn.Module):
         values = values.view(batch_size, seq_length, -1, self.head_dim).transpose(1, 2)
         # CLIP text model uses both `causal_attention_mask` and `attention_mask`
         # in case FA2 kernel is called, `is_causal` should be inferred from `causal_attention_mask`
-        if "flash" in self.config._attn_implementation:
-            self.is_causal = causal_attention_mask is not None
+        if "flash" in self.config._attn_implementation or "sdpa" in self.config._attn_implementation:
+            is_causal = True
         else:
+            is_causal = self.is_causal
             if attention_mask is not None and causal_attention_mask is not None:
                 attention_mask = attention_mask + causal_attention_mask
             elif causal_attention_mask is not None:
@@ -337,7 +338,7 @@ class CLIPAttention(nn.Module):
             keys,
             values,
             attention_mask,
-            is_causal=self.is_causal,
+            is_causal=is_causal,
             scaling=self.scale,
             dropout=0.0 if not self.training else self.dropout,
             output_attentions=output_attentions,
@@ -605,15 +606,26 @@ class CLIPTextTransformer(nn.Module):
         hidden_states = self.embeddings(input_ids=input_ids, position_ids=position_ids)
 
         # CLIP's text model uses causal mask, prepare it here.
-        # https://github.com/openai/CLIP/blob/cfcffb90e69f37bf2ff1e988237a0fbe41f33c04/clip/model.py#L324
-        causal_attention_mask = _create_4d_causal_attention_mask(
-            input_shape, hidden_states.dtype, device=hidden_states.device
+
+        cache_position = torch.arange(
+            hidden_states.shape[1], device=hidden_states.device
         )
 
-        # expand attention_mask
-        if attention_mask is not None and "flash" not in self.config._attn_implementation:
-            # [batch_size, seq_len] -> [batch_size, 1, tgt_seq_len, src_seq_len]
-            attention_mask = _prepare_4d_attention_mask(attention_mask, hidden_states.dtype)
+        causal_attention_mask = create_causal_mask(
+            config=self.config,
+            input_embeds=hidden_states,
+            attention_mask=attention_mask,
+            cache_position=cache_position,
+            past_key_values=None,
+            position_ids=position_ids,
+        )
+        
+        if attention_mask is not None:
+            attention_mask = create_bidirectional_mask(
+                config=self.config,
+                input_embeds=hidden_states,
+                attention_mask=attention_mask,
+            )
 
         encoder_outputs: BaseModelOutput = self.encoder(
             inputs_embeds=hidden_states,
--- a/src/transformers/models/clip/modeling_clip.py
+++ b/src/transformers/models/clip/modeling_clip.py
@@ -22,7 +22,7 @@ import torch
 from torch import nn
 
 from ...activations import ACT2FN
-from ...masking_utils import create_causal_mask, create_bidirectional_mask
+from ...modeling_attn_mask_utils import _create_4d_causal_attention_mask, _prepare_4d_attention_mask
 from ...modeling_layers import GradientCheckpointingLayer
 from ...modeling_outputs import BaseModelOutput, BaseModelOutputWithPooling, ImageClassifierOutput
 from ...modeling_utils import ALL_ATTENTION_FUNCTIONS, PreTrainedModel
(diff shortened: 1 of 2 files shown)


src/transformers/models/clip/modeling_clip.py
```diff
@@ -319,7 +319,7 @@ def forward(
         values = values.view(batch_size, seq_length, -1, self.head_dim).transpose(1, 2)
         # CLIP text model uses both `causal_attention_mask` and `attention_mask`
         # in case FA2 kernel is called, `is_causal` should be inferred from `causal_attention_mask`
-        if self.config._attn_implementation == "flash_attention_2":
+        if "flash" in self.config._attn_implementation:
             self.is_causal = causal_attention_mask is not None
```
zucchini-nlp: wow, first time seeing it. In general it's not a good idea to dynamically change `self` attributes in forward call. This can have unwanted side-effects especially if the attention is changed dynamically after loading with `model.set_attn_implementation("eager")`

cc @vasqu for attention interfaces, I remember there was a discussion on `self.is_causal` somewhere. Do we have a suggested way to define `is_causal` based on inputs. For CLIP we can have causal mask if the attention is used with text and otherwise it's bidirectional

src/transformers/models/clip/modeling_clip.py
```diff
@@ -611,7 +611,7 @@ def forward(
         )
 
         # expand attention_mask
-        if attention_mask is not None and self.config._attn_implementation != "flash_attention_2":
+        if attention_mask is not None and "flash" not in self.config._attn_implementation:
             # [batch_size, seq_len] -> [batch_size, 1, tgt_seq_len, src_seq_len]
             attention_mask = _prepare_4d_attention_mask(attention_mask, hidden_states.dtype)
```
zucchini-nlp: I'd love to see the masking utils used here, for ex `from ...masking_utils import create_causal_mask, create_bidirectional_mask`. They'll handle attention implementations internally

src/transformers/models/clip/modeling_clip.py
vasqu: On the models I was involved with, have never seen dynamic changes to the attribute :eyes: we need this for fa and sdpa attention (sdpa is less important when a mask is always provided)

I think there is no good alternative for now. I thought about allowing `is_causal` as kwarg to take precedence over the attribute - already possible in flash attn, would need adjustments in sdpa tho. Could already be done here ig, the sdpa case is already ignored either way it seems :sweat:

vasqu: Fa logic: https://github.com/huggingface/transformers/blob/38fdad1bf8dbbfc1ef4671a2f13c39895c675adf/src/transformers/integrations/flash_attention.py#L66-L69

zucchini-nlp: > allowing is_causal as kwarg to take precedence over the attribute 

Personally I prefer this more to give us freedom and re-use single attention layer for multimodal models like CLIP

vasqu: Yea, totally fair. I'll open a PR for SDPA to handle this. In the meantime, we can already adjust this here as well, setting it to a local attribute and pass it then.

zucchini-nlp: @akashpalla can you fix this part and instead try to pass an `is_causal=True/False` arg to the attention function?

src/transformers/models/clip/modeling_clip.py
akashpalla: I’m not super familiar with attention/masking, so I tried hacking together a solution based this on patterns I saw in other models but this resulted in a null  `causal_attention_mask` for SDPA + FA and broke some tests 

```
cache_position = torch.arange(
    hidden_states.shape[1], device=hidden_states.device
)

causal_attention_mask = create_causal_mask(
    config=self.config,
    input_embeds=hidden_states,
    past_key_values=None,
    position_ids=position_ids,
)

    attention_mask = create_bidirectional_mask(
        config=self.config,
        input_embeds=hidden_states,
    )
    
```
If there’s a better way to handle this, I’d love some guidance — or if you prefer, this could also be addressed in a separate PR.

zucchini-nlp: > resulted in a null causal_attention_mask for SDPA + FA and broke some tests

yeah, the null is actually intended and in that case we fallback to the FA kernel and the masking is controlled by `is_causal` attr. So imo the issue is that `is_causal` is not set to correct value. Can you check in the tests what is the value now and what did the mask look like before this PR?

Same for FA2 tests, must be the mask

src/transformers/models/clip/modeling_clip.py
```diff
@@ -337,7 +338,7 @@ def forward(
             keys,
             values,
             attention_mask,
-            is_causal=self.is_causal,
+            is_causal=is_causal,
```
zucchini-nlp, line 341: we can do `is_causal=causal_attention_mask is not None or self.is_text_attention` for all cases of attentions. Rn prob it is not differentiating correctly between vision or text attention because the presence of causal mask is not a marker with new masking utility

src/transformers/models/clip/modeling_clip.py
vasqu: This is a bit weird, if you look closer `_create_4d_causal_attention_mask` is a very old mask API. If I see it correctly, then what happened here is that only a normal causal attention mask with padding is created atm (by adding them up) - `_create_4d_causal_attention_mask` creates a causal mask, the other mask creates the padding mask.

Since it relies on a 4D mask, FA is not supported there (also marked within CLIP accordingly for anything text-related).

What should happen instead --> use new causal mask API `create_causal_mask` and only generate one mask --> pass `is_causal=True` for the layers below --> remove the overcomplicated FA logic and masking logic with additions. This should also enable FA support in text-based situations. I'm not sure if this is out-of-scope / too hard here; I would try looking into it myself.

src/transformers/models/clip/modeling_clip.py
vasqu: Would be careful with inline conditionals, they have caused troubles with torch compile in the past :sweat:

src/transformers/models/clip/modeling_clip.py
vasqu: Also we need to merge with main, for SDPA to work with `is_causal` kwarg

vasqu:
Sorry about this but #41750 will likely supersede this PR. Things got a bit more complicated than initially anticipated :sweat:

akashpalla:
Closing in favor of: https://github.com/huggingface/transformers/pull/41750. Thanks for the quick fix

PR #41675, opened 2025-10-17
Use | for Optional and Union typing 

cyyever:
# What does this PR do?
This PR replaces `Optional` and `Union` typing with `|` in some src files.


Rocketknight1 (approved):
Yep, LGTM. Thank you!

PR #41678, opened 2025-10-17
Enable faiss-cpu on Windows

cyyever:
# What does this PR do?

The current faiss-cpu supports Windows. Therefore the PR enables that test case.


Rocketknight1 (approved):
Yep, LGTM!

PR #41679, opened 2025-10-17
Use the process_group from device_mesh to all_reduce in GatherParallel

y-sq:
# What does this PR do?

In`GatherParallel`, where the `device_mesh` is passed, but not used by the `all_reduce` call. 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@3outeille @ArthurZucker 

<!-- Your PR will be replied to more quickly if you can figure out the right person to tag with @

 If you know how to use git blame, that is the easiest way, otherwise, here is a rough guide of **who to tag**.
 Please tag fewer than 3 people.

### DIFF
--- a/src/transformers/integrations/tensor_parallel.py
+++ b/src/transformers/integrations/tensor_parallel.py
@@ -459,10 +459,12 @@ class GatherParallel(TensorParallelLayer):
 
     @staticmethod
     def _prepare_output_fn(output_layouts, use_local_output, mod, outputs, device_mesh):
+        assert device_mesh.ndim == 1, f"Expected 1D device mesh but got {device_mesh.ndim}-D device mesh"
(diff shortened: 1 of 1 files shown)


3outeille:
Since we handle only 1D parallelism for now, the `all_reduce` will use by default the default process group. So we dont need to pass it explicitly

However, as we move on later to support ND parallelism, the `pg = device_mesh.get_group(mesh_dim=0)` is not correct because we don't know if `mesh_dim=0` means TP axis or DP axis for example. Plus, you added an `assert` to enforce 1D parallelism, so this is the same as letting torch uses the default process group. So I dont think the patch is necessary for now

Rocketknight1:
cc @arthurzucker for tensor parallel

PR #41681, opened 2025-10-17
🚨 Delete deprecations with end-cycle in v4.xx and v5.0

zucchini-nlp:
# What does this PR do?

Deletes deprecation for the v5 release. cc @yonigozlan and @eustlb to verify that we can really delete them, since it is mostly deprecated stuff from processors

For core maintainers:
1. The max-size and pad-pixels args can't be deleted due to many models on the hub still using it. Instead I decided to stop raising warning and silently load max size if present. However users cannot anymore use these args when calling the processor
2. For SDPA attention weights, the warning is reworded to push us to update those models. Usually if model has the new attention API, the warning is raised internally in `sdpa_attention`, but for old models we have to manually duplicate the waning message


src/transformers/models/aimv2/modeling_aimv2.py
```diff
@@ -445,13 +443,11 @@ def __init__(self, config: Aimv2VisionConfig):
     def get_input_embeddings(self) -> nn.Module:
         return self.embeddings.patch_embed
 
-    @deprecate_kwarg("attention_mask", version="v4.58.0")
     @check_model_inputs(tie_last_hidden_states=False)
     @auto_docstring
     def forward(
         self,
         pixel_values,
```
yonigozlan, line 453: As part of vision and vlms refactor, we'll be trying to add/standardize support for attention masks, so we might want to keep that? Cc @molbap @vasqu

src/transformers/models/beit/modeling_beit.py
```diff
@@ -323,21 +315,6 @@ def forward(
...
-                'This warning can be removed using the argument `attn_implementation="eager"` when loading the model.'
-            )
-            return super().forward(
-                hidden_states=hidden_states,
-                output_attentions=output_attentions,
-                relative_position_bias=relative_position_bias,
-                interpolate_pos_encoding=interpolate_pos_encoding,
-                resolution=resolution,
-            )
-
```
yonigozlan, line 340: Not sure what the behavior will be here then, in any case I'll update these soon!

src/transformers/models/conditional_detr/image_processing_conditional_detr.py
```diff
@@ -829,15 +829,7 @@ def __init__(
...
 
-        if "max_size" in kwargs:
-            logger.warning_once(
-                "The `max_size` parameter is deprecated and will be removed in v4.26. "
-                "Please specify in `size['longest_edge'] instead`.",
-            )
-            max_size = kwargs.pop("max_size")
-        else:
-            max_size = None if size is None else 1333
-
```
yonigozlan, line 840: Agreed that this needs to be done at some point, but it might also break a lot of models on the hub. Just checking if we are fine with this.
There have been a few PRs trying to deprecate this (https://github.com/huggingface/transformers/issues/37939 https://github.com/huggingface/transformers/pull/35090 https://github.com/huggingface/transformers/pull/34998), scanning the hub to see which models would be impacted, but it looks like it was never fully deprecated, not sure why.
I can try to make another scan on the hub, and have a script open PRs to modify the configs.

src/transformers/models/deformable_detr/image_processing_deformable_detr.py
```diff
@@ -834,15 +834,7 @@ def __init__(
         if "pad_and_return_pixel_mask" in kwargs:
             do_pad = kwargs.pop("pad_and_return_pixel_mask")
 
-        if "max_size" in kwargs:
```
yonigozlan, line 837: same as conditional detr

src/transformers/models/maskformer/image_processing_maskformer.py
```diff
@@ -59,7 +58,7 @@
 
 
 if TYPE_CHECKING:
-    from transformers import MaskFormerForInstanceSegmentationOutput
+    pass
```
yonigozlan:
```suggestion
```

src/transformers/models/sam/modeling_sam.py
```diff
@@ -846,18 +846,6 @@ def __init__(self, config, window_size):
         super().__init__(config, window_size)
 
     def forward(self, hidden_states: torch.Tensor, output_attentions=False) -> torch.Tensor:
-        if output_attentions:
```
yonigozlan, line 849: Might crash but will be refactored anyway as well :)

yonigozlan (approved):
Love this so much 🥲, thanks for taking the time to do this @zucchini-nlp !
My only reservation is for max_size, I'll try to scan the hub to see how bad the damage would be right now, and open PRs where necessary

src/transformers/models/aimv2/modeling_aimv2.py
zucchini-nlp: Not sure how useful it will be for vision models where the pixels are not padded, therefore nothing to mask out usually

src/transformers/models/beit/modeling_beit.py
zucchini-nlp: oh, this one must have been deleted months ago. In text models we've deleted it in favor of the below code 

https://github.com/huggingface/transformers/blob/1eb45cd61dd9068e9ce53a216bd0bfa6f62d0548/src/transformers/configuration_utils.py#L359-L369

src/transformers/models/aimv2/modeling_aimv2.py
yonigozlan: I know some people prefer to use padding rather than resize with changing aspect ratios, even for model with image processors that don't pad by default, so I think some users would still appreciate it

src/transformers/models/beit/modeling_beit.py
yonigozlan: Ok nice!

src/transformers/models/aimv2/modeling_aimv2.py
zucchini-nlp: ah, didn't know about that

src/transformers/models/wav2vec2/modeling_wav2vec2.py
```diff
@@ -1301,18 +1277,6 @@ def __init__(self, config: Wav2Vec2Config):
...
-        """
-        Calling this function will disable the gradient computation for the feature encoder so that its parameters will
-        not be updated during training.
-        """
-        warnings.warn(
-            "The method `freeze_feature_extractor` is deprecated and will be removed in Transformers v5. "
-            "Please use the equivalent `freeze_feature_encoder` method instead.",
-            FutureWarning,
-        )
-        self.freeze_feature_encoder()
```
eustlb, line 1303: a 4 years deprecation cycle ahah

src/transformers/models/aimv2/modeling_aimv2.py
vasqu: Weird that it was marked as deprecation then but don't have anything against it if it has a proper use-case :+1:

src/transformers/models/mimi/modeling_mimi.py
```diff
@@ -839,22 +839,6 @@ def forward(
...
-            )
-            return super().forward(
-                hidden_states=hidden_states,
-                attention_mask=attention_mask,
-                position_ids=position_ids,
-                past_key_values=past_key_values,
-                output_attentions=output_attentions,
-                use_cache=use_cache,
-                cache_position=cache_position,
-            )
```
eustlb, line 886: saw your above answer to @yonigozlan about that, but don't you think it would be better to keep a warning like [here](https://github.com/huggingface/transformers/blob/c01ceffeb48157d3b7fd3f44f5a58a553b9f172d/src/transformers/integrations/sdpa_attention.py#L55-L59)?
Ideally, everything that was before copied from gemma (that is where this comes from) should have been also refactoed to use 
```
```
so that we ensure a user is correctly logged when setting `output_attentions=True` in the forward, which is not the case if we simply remove this, we will silently not return attentions

eustlb (commented):
Thanks a lot for taking care of that!
General comment:

For every models that had something like
```python
logger.warning_once(
                "MimiModel is using MimiSdpaAttention, but `torch.nn.functional.scaled_dot_product_attention` does not support `output_attentions=True`. Falling back to the manual attention implementation, "
                'but specifying the manual implementation will be required from Transformers version v5.0.0 onwards. This warning can be removed using the argument `attn_implementation="eager"` when loading the model.'
            )
```
because of being copied from Gemma, I think we should keep a warning. I've seen your answer above about that, but this is only when output_attentions is setted in the config, not in the forward, no? Therefore, until those implementations rely on `ALL_ATTENTION_FUNCTIONS`, we still have to handle the warning here to ensure we don't silently skip outputting attentions.

Review from my side:
core models (meaning ones that do not rely on modular)
- moshi
- mimi
- clap
- seamless_m4t
- unispeech
- unispeech_sat
- wav2vec2
- wav2vec2_bert
- whisper

the rest are impacted by modular directly:
- hubert (wav2vec2)
- kyutai_speech_to_text (moshi)
- sew (wav2vec2)
- sew_d (wav2vec2)
- wavlm (wav2vec2)
- data2vec_audio (wav2vec2)

zucchini-nlp:
> Therefore, until those implementations rely on ALL_ATTENTION_FUNCTIONS

Ah, after looking I see that it's not going to raise any warning unless we use  `ALL_ATTENTION_FUNCTIONS`.  In that case I think we can reword the warning that "Attention weights will not be returned" and remove the workaround. Users have been warned about SDPA so I think it will not be a big problem

Also it will nudge us to incorporate `ALL_ATTENTION_FUNCTIONS`. WDYT?

eustlb:
Yep, totally agree!

src/transformers/models/conditional_detr/image_processing_conditional_detr.py
zucchini-nlp: I think in this case it's better to convert warning to a `ValueError`, so users know what is the reason if it crashes and how to fix it

Hub-related deprecations are so annoying sometimes 🙃

src/transformers/models/bloom/modeling_bloom.py
```diff
@@ -1149,7 +1103,6 @@ def forward(
         self,
         input_ids: Optional[torch.LongTensor] = None,
         attention_mask: Optional[torch.FloatTensor] = None,
-        position_ids: Optional[torch.LongTensor] = None,
```
zucchini-nlp, line 1145: model doesn't use positions, I believe it was simply copied from other models without checking

src/transformers/models/conditional_detr/image_processing_conditional_detr.py
zucchini-nlp: Oh wait, I see a lot of models with high usage (including official ckpt) that still have `max_size`. The PR has been open on the hub for a while, so seems like it won't get merged :(

I think, the best solution in that case is to support it forever but stop raising warnings, because it is not actionable for users

src/transformers/models/aimv2/modeling_aimv2.py
zucchini-nlp: A few tests will fail if we start propagating the mask, so I will remove it as it's not actually used currently. It can be re-added in subsequent PRs with adjusting tests

zucchini-nlp:
Hub down? 😢

zucchini-nlp:
Oke, now it is done I think. @yonigozlan @eustlb if you want to take another look, I will also request core mantainer's review

src/transformers/models/rt_detr/image_processing_rt_detr_fast.py
```diff
@@ -520,5 +512,12 @@ def post_process_object_detection(
 
         return results
 
+    def preprocess(
+        self,
+        images: ImageInput,
+        **kwargs: Unpack[RTDetrImageProcessorKwargs],
+    ) -> BatchFeature:
+        return super().preprocess(images, **kwargs)
```
Cyrilvallez: Let's remove that, it's not needed (we only call the parent)

Cyrilvallez (approved):
Amazing cleanup! Thanks a lot! Just a few conflicts to fix and let's go!

PR #41683, opened 2025-10-17
🚨 Remove torch.fx support

Cyrilvallez:
# What does this PR do?

As per the title. The API is old and very badly supported. It's also superseded by dynamo


ArthurZucker (approved):
Thanks for removing!

PR #41685, opened 2025-10-17
[kernels] Add version to function mapping

MekkCyber:
# What does this PR do?

We simply add a version attribute to the function kernel mapping to avoid breaking changes for kernels in the future


ArthurZucker (approved):
Can you add a small test? 🤗

MekkCyber:
Yes added the test here : https://github.com/huggingface/transformers/pull/41765 with all the other tests, and kernels CI. I will merge this and rebase to make sure evtn passes

PR #41687, opened 2025-10-17
fix(data): Handle integer labels in DataCollatorWithFlattening

### LINKED ISSUE #41652: DataCollatorWithFlattening incompatible with sequence classification
### System Info

-  OS: Ubuntu 24.04
- python version 3.11.9
- transformers version 4.56.2

### Who can help?

@ArthurZucker 

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

```python
from datasets import Dataset
from transformers import DataCollatorWithFlattening

ds = Dataset.from_list([
    {"input_ids": [0, 1, 2, 3], "labels": 1}, {"input_ids": [4, 5, 6], "labels": 0}
])

collator = DataCollatorWithFlattening()
batch = collator(ds)
print(batch)
```

## Observed Error

```
   2026 batch["input_ids"] += input_ids
   2027 if is_labels_provided:
-> 2028     batch["labels"] += [separator_id] + sample["labels"][1:]
   2029 else:
   2030     batch["labels"] += [separator_id] + input_ids[1:]

TypeError: 'int' object is not subscriptable
```

## Expected Behavior
Batch the labels in a `[batch_size,]` int64 tensor

## Additional Context

This [line](https://github.com/huggingface/transformers/blob/8cb5963cc22174954e7dca2c0a3320b7dc2f4edc/src/transformers/data/data_collator.py#L2146C17-L2146C73) in the `DataCollatorWithFlattening` class raises an exception if the "labels" column in a dataset is a literal instead of a sequence. This happens in sequence classification where there is a label for the whole sequence instead of individual tokens.

yashisthebatman:
# What does this PR do?

This PR fixes a `TypeError` in `DataCollatorWithFlattening` that occurs when processing datasets with integer labels, which is a common format for sequence classification tasks.

The collator was written with the assumption that the `labels` field would always be a sequence, and it attempted to slice it (`sample["labels"][1:]`). When `sample["labels"]` is an integer, this causes the application to crash.

The fix introduces a check to determine if the `labels` are sequence-like.
- If the labels are a sequence (the original expected format), the collator preserves the exact original logic.
- If the labels are integers, the collator now correctly broadcasts the integer label across all tokens of the corresponding `input_ids`.

This change makes the collator more robust to common data formats without altering its behavior for existing use cases, thereby preventing unexpected crashes. A new test case has been added to verify this fix.

Fixes #41652

## Before submitting
      Pull Request section?
      to it if that's the case.
- [X] Did you write any new necessary tests?

## Who can review?

This change relates to data collators, which are often used with the Trainer. Tagging from the original issue and the trainer section.
@ArthurZucker @SunMarc

### DIFF
--- a/src/transformers/data/data_collator.py
+++ b/src/transformers/data/data_collator.py
@@ -1370,6 +1370,7 @@ class DataCollatorWithFlattening(DefaultDataCollator):
     - no padding will be added, returns `input_ids`, `labels` and `position_ids` by default
     - optionally returns the kwargs contained in FlashAttentionKwargs
     - optionally returns seq_idx indicating which sequence each token belongs to
+    - `pack_sequence_labels`: if True, will pack integer labels for sequence classification into a `(batch_size,)` tensor instead of broadcasting them to match `input_ids`.
 
     <Tip warning={true}>
 
@@ -1386,6 +1387,7 @@ class DataCollatorWithFlattening(DefaultDataCollator):
         separator_id=-100,
         return_flash_attn_kwargs=False,
         return_seq_idx=False,
+        pack_sequence_labels=False,
         **kwargs,
     ):
         super().__init__(*args, **kwargs)
@@ -1393,6 +1395,7 @@ class DataCollatorWithFlattening(DefaultDataCollator):
         self.separator_id = separator_id
         self.return_flash_attn_kwargs = return_flash_attn_kwargs
         self.return_seq_idx = return_seq_idx
+        self.pack_sequence_labels = pack_sequence_labels
         self._int_64_keys = {"labels", "position_ids", "input_ids"}
         self._batch_dim_keys = {"labels", "position_ids", "input_ids", "seq_idx"}
         self._py_int_keys = {"max_length_q", "max_length_k"}
@@ -1403,6 +1406,9 @@ class DataCollatorWithFlattening(DefaultDataCollator):
         if separator_id is None:
             separator_id = self.separator_id
         is_labels_provided = "labels" in features[0]
+
+        is_labels_sequence = is_labels_provided and isinstance(features[0].get("labels"), (list, tuple, np.ndarray))
+
         batch = {"input_ids": [], "labels": []}
         if self.return_position_ids:
             batch.update({"position_ids": []})
@@ -1411,13 +1417,19 @@ class DataCollatorWithFlattening(DefaultDataCollator):
         if self.return_flash_attn_kwargs:
             cu_seq_lens = [0]
             max_length = 0
+
         for seq_idx, sample in enumerate(features):
             input_ids = sample["input_ids"]
             batch["input_ids"] += input_ids
             if is_labels_provided:
-                batch["labels"] += [separator_id] + sample["labels"][1:]
+                if is_labels_sequence:
+                    # Original logic for token-level labels.
+                    batch["labels"] += [self.separator_id] + sample["labels"][1:]
+                else:
+                    # Default "safe" behavior: broadcast the integer label to all tokens.
+                    batch["labels"] += [sample["labels"]] * len(input_ids)
             else:
-                batch["labels"] += [separator_id] + input_ids[1:]
+                batch["labels"] += [self.separator_id] + input_ids[1:]
             if self.return_position_ids:
                 batch["position_ids"] += list(range(len(input_ids)))
             if self.return_seq_idx:
@@ -1426,11 +1438,14 @@ class DataCollatorWithFlattening(DefaultDataCollator):
                 cu_seq_lens.append(cu_seq_lens[-1] + len(input_ids))
                 max_length = max(max_length, len(input_ids))
 
+        # If packing is enabled for sequence classification, overwrite the broadcasted labels.
+        if is_labels_provided and not is_labels_sequence and self.pack_sequence_labels:
+            batch["labels"] = [feature["labels"] for feature in features]
+
         if self.return_flash_attn_kwargs:
             batch["cu_seq_lens_q"] = batch["cu_seq_lens_k"] = cu_seq_lens
             batch["max_length_q"] = batch["max_length_k"] = max_length
 
-        # FlashAttentionKwargs and seq_idx are expected to be int32s.
         if return_tensors == "pt":
             import torch
 
@@ -1445,9 +1460,12 @@ class DataCollatorWithFlattening(DefaultDataCollator):
             raise ValueError(f'return_tensors must be one of ("pt", "np"), {return_tensors=} not supported')
 
         for k, v in batch.items():
-            if k in self._batch_dim_keys:
+            # For packed sequence labels, we want a 1D tensor, not a 2D tensor of shape (1, batch_size).
+            if k == "labels" and is_labels_provided and not is_labels_sequence and self.pack_sequence_labels:
+                pass
+            elif k in self._batch_dim_keys:
                 v = [v]
-            # Flash attention max_len_{q,k} are python ints
+
             if k not in self._py_int_keys:
                 batch[k] = data_cls(v, dtype=dtype_64 if k in self._int_64_keys else dtype_32)
 
--- a/tests/trainer/test_data_collator.py
+++ b/tests/trainer/test_data_collator.py
@@ -18,6 +18,8 @@ import tempfile
 import unittest
 
 import numpy as np
+import pytest
+import torch
 
 from transformers import (
     BertTokenizer,
@@ -1965,3 +1967,36 @@ class DataCollatorForLanguageModelingUnitTest(unittest.TestCase):
         ).astype(bool)
 
         np.testing.assert_array_equal(output_mask, expected_mask)
+
+
+@pytest.mark.parametrize("pack_sequence_labels", [True, False])
+def test_data_collator_with_flattening_for_sequence_classification(pack_sequence_labels):
+    """
+    Tests that DataCollatorWithFlattening can handle integer labels for sequence classification,
+    both with broadcasting (default) and simple packing (for advanced use cases).
+    """
+    from transformers import DataCollatorWithFlattening
+
+    features = [
+        {"input_ids": [0, 1, 2, 3], "labels": 1},
+        {"input_ids": [4, 5, 6], "labels": 0},
+    ]
+
+    collator = DataCollatorWithFlattening(pack_sequence_labels=pack_sequence_labels, return_tensors="pt")
+    batch = collator(features)
+
+    # The input_ids are always concatenated.
+    expected_input_ids = torch.tensor([[0, 1, 2, 3, 4, 5, 6]])
+    assert torch.equal(batch["input_ids"], expected_input_ids)
+
+    # The labels tensor shape and content depend on the packing flag.
+    if pack_sequence_labels:
+        # The reviewer's requested behavior: a 1D tensor of shape (batch_size,).
+        expected_labels = torch.tensor([1, 0])
+        assert batch["labels"].shape == (2,)
+    else:
+        # The default, safe behavior: broadcast the label to all tokens, resulting in a 2D tensor.
+        expected_labels = torch.tensor([[1, 1, 1, 1, 0, 0, 0]])
+        assert batch["labels"].shape == (1, 7)
+
+    assert torch.equal(batch["labels"], expected_labels)


src/transformers/data/data_collator.py
```diff
@@ -1403,6 +1403,12 @@ def __call__(self, features, return_tensors=None, separator_id=None):
         if separator_id is None:
             separator_id = self.separator_id
         is_labels_provided = "labels" in features[0]
+
+        # --- Start of our fix ---
+        # Determine if labels are sequence-like (token-level) or single values (sequence-level)
+        are_labels_sequence = is_labels_provided and isinstance(features[0].get("labels"), (list, tuple, np.ndarray))
```
LPirch: minor suggestion: `is_labels_sequence` would be more a consistent naming with the rest of the function

src/transformers/data/data_collator.py
```diff
@@ -1411,13 +1417,27 @@ def __call__(self, features, return_tensors=None, separator_id=None):
...
-                batch["labels"] += [separator_id] + sample["labels"][1:]
+                if are_labels_sequence:
+                    # This is the original logic for token-level labels. It's preserved.
+                    batch["labels"] += [self.separator_id] + sample["labels"][1:]
+                else:
+                    # New logic for sequence-level integer labels.
+                    # We must create a label for each token in the input_ids.
+                    # The label for the sequence is applied to all its tokens.
+                    # We do NOT add a separator, as that concept doesn't apply to this kind of label.
+                    batch["labels"] += [sample["labels"]] * len(input_ids)
```
LPirch, line 1430: Why not `batch["labels"].append(sample["labels"])`?

What I had in mind was a sequence classification task in which a model receives a sequence (batch_size, seq_length) and outputs scores for each class (batch_size, num_classes). The collator would now return a (batch_size, seq_length) tensor which needs further processing on the user side.

Is there a use case where expanding the class to the number of tokens is required?

LPirch (changes requested):
Thanks for your work! There is one thing I didn't fully understand (see comments) and which needs further clarification.

src/transformers/data/data_collator.py
```diff
@@ -1411,13 +1417,27 @@ def __call__(self, features, return_tensors=None, separator_id=None):
...
             if is_labels_provided:
-                batch["labels"] += [separator_id] + sample["labels"][1:]
+                if are_labels_sequence:
+                    # This is the original logic for token-level labels. It's preserved.
+                    batch["labels"] += [self.separator_id] + sample["labels"][1:]
+                else:
+                    # New logic for sequence-level integer labels.
+                    # We must create a label for each token in the input_ids.
+                    # The label for the sequence is applied to all its tokens.
+                    # We do NOT add a separator, as that concept doesn't apply to this kind of label.
```
Rocketknight1: All of these comment lines should probably be removed, particularly `# --- Start of our fix ---`! The ones in the test are okay, but descriptions of "new logic" and "original logic" will confuse people, since they won't know which bits were added in a PR and which weren't.

yashisthebatman:
@LPirch Thanks for the review i have removed the unecessary comments and also have also followed the more consistent naming scheme but regarding the Why not batch["labels"].append(sample["labels"])? :
The entire purpose of DataCollatorWithFlattening is to create one single long sequence, so the labels tensor must have the same length as the final input_ids tensor. Your suggestion (.append()) would create mismatched lengths (e.g., input_ids of length 7 but labels of length 2), which would immediately crash any model's loss function.

LPirch:
Let me give a bit more context to prevent misunderstandings: In sequence classification, you don't have a label per token but a label per sample. Now, when sequences get very long, you might want to use flash attention (as in my case). The only way to process a batch of long sequences is to pack them using the flattening collator and run them through the language model. The needs to be unpacked using the `position_ids` tensor (see example below) to recover the original samples.

In my case, you then need a pooling that creates an embedding per sequence and a classification head (linear layers) producing an output score. The loss function is then just a regular classification loss.

unpacking example:
```python
def pool_last_token_flat(last_hidden_states: torch.Tensor, position_ids: torch.Tensor) -> torch.Tensor:
    end_idx = torch.argwhere(position_ids == 0)[:, 1][1:]
    end_idx = torch.cat([end_idx, torch.tensor([last_hidden_states.shape[1] - 1], device=last_hidden_states.device)])
    return last_hidden_states[0, end_idx]
```

This pooling takes the last layers hidden vector of the last token of each (sub-)sequence in the flattened sequence. 

TLDR: We don't create flat sequences for the sake of flattening but because flash attention 2 can't handle padding. Recovering the batch structure is required anyways in the case of sequence classification.

yashisthebatman:
Hi @LPirch ,
Thank you for that context! That makes total sense. I didn't consider the flash attention -> pack -> unpack -> pool workflow.
You're right, my original fix is useful for preventing a crash but actively unhelpful for your use case, since you'd just have to undo it. What do you think of this?
We add a boolean flag to the collator: pack_sequence_labels=False.
Default (False): It does my original fix (broadcasts the label). This is the "safe" option for anyone not doing what you're doing, and it guarantees input_ids and labels have the same length.
Your case (True): You'd set pack_sequence_labels=True. The collator would then just collect the integer labels into a simple list [label1, label2, ...], giving you the (batch_size,) tensor you need for your pooling logic.
This feels like the cleanest way to support both scenarios without breaking things for anyone.
If that sounds good to you, I'll go ahead and implement it.

LPirch:
Sounds good! Thanks for your support :raised_hands:

yashisthebatman:
@LPirch I have done all the changes that we discussed please do check it

LPirch (approved):
Looks perfect. Thanks for your work!

yashisthebatman:
@LPirch thank you for accepting my pr i am new to open source could you tell me where i could improve myself or any suggestions that could help me become a better contributor

LPirch:
@yashisthebatman I think some steps in the CI pipeline failed (code quality). Can you fix those?

About your question: You're doing well! Keep the positivity in communication and stay curious / be open to learn from others and keep a focus on understanding the bigger picture before finalizing a PR - that are the key ingredients if you ask me. The rest is experience.

yashisthebatman:
@LPirch fixed those formatting errors this time it must work thank you for the patience

yashisthebatman:
@LPirch all the tests have passed successfully

yashisthebatman:
@LPirch could you please merge this problem I have had successfully solved all the formatting issues

LPirch:
Just to clarify: I'm not a maintainer of this project. I can't assign, resolve or merge anything. I'm just a user who reported an issue.

From now on, @Rocketknight1 would need to take the lead. I suggest we wait patiently until he finds the time and see if he has additional feedback. Also, let's not bloat the conversation too much.

Rocketknight1:
Hey all! I'm sorry for the delays, but I finally got around to doing a proper review! I have one main question: I realize that packing is used as a way of improving performance with some FlashAttention backends, but do our `SequenceClassification` models actually support packed inputs?

Packing works for CLM training because the objective for those models is token-level, so you can just pack sequences together and concatenate `input_ids` with a separator, then do the same with their labels. However, most `SequenceClassification` models emit one logit distribution per sequence, not per token. If you pack inputs together, it's not clear to me how sequence classification training is supposed to work, and I think the user would have to unpack the outputs before passing them to the model anyway. What exactly is the workflow supposed to look like when a `Flattening` collator is combined with a `SequenceClassification` model?

yashisthebatman:
@Rocketknight1 A `SequenceClassification` model won't know what to do with a packed input straight out of the box. This feature is really for an advanced, manual workflow that some users need for performance, especially with things like Flash Attention 2 that don't like padding.

So the idea is, a user would grab this collator with `pack_sequence_labels=True`. This gives them:
1.  The packed `input_ids` and `position_ids` they need to feed the base model.
2.  And crucially, a clean `labels` tensor of shape `(batch_size,)`. **This is the missing piece my PR provides.**

From there, the workflow looks like this on the user's end:

First, they'd get the raw hidden states by calling just the base model, bypassing the classification head:
```python
# User's code
outputs = model.base_model(packed_input_ids)
last_hidden_state = outputs.last_hidden_state # A single giant tensor
```

Next, they have to do the manual part: use the `position_ids` to find the last token of each original sequence in that giant tensor and pool them. This turns the flattened output back into a standard `(batch_size, hidden_size)` tensor.

And now they have everything they need. They can feed their pooled outputs into `model.classifier`, and the `labels` tensor from the collator will match up perfectly for the loss calculation.

So basically this PR doesn't make the models  understand packed inputs rather provides the correctly shaped `labels` tensor that makes this whole manual packing/unpacking dance possible for sequence classification.

Rocketknight1:
Hmmn. I see what you're going for, but there's a distribution shift there: If a model is fine-tuned for sequence classification, it will generally not see concatenated documents with separators during training. Packing works for CLM training because documents are often concatenated across document boundaries anyway, so the model is familiar with that, but I expect the performance of a sequence classification model to sharply drop if you use the hidden states from a later sequence that follows multiple earlier sequences and separator tokens.

In other words, packed sequences aren't actually independent and the later sequences can still attend to the earlier ones, they just learn not to do that if they see enough concatenated docs in training!

As such, I'm not sure we want to add this feature - I think it's very niche and likely to be confusing to users who expect it to work, when really it requires a lot of manual effort to separate the outputs for the classification head, and probably also degrades performance in most cases as well.

LPirch:
Oh, I was not aware that there are cross-sequence information leaks in a packed sequence. I didn't think of this as an option because (as you said) this would also contaminate gradients on regular token predictions, i.e. attending to tokens of previous sequences in a batch. From a user perspective, I find it unintuitive that flash attention implicitly sacrifices soundness for efficiency but that maybe belongs to a different issue.

Generally, I can understand your argument though and I would be fine if you close it. I have already monkey-patched the collator to my needs in the meantime and only thought it would be a useful feature for others as well.

yashisthebatman:
Thanks for clarifying your position, @LPirch. Given the valid concerns about attention leakage and the potential for user confusion, I agree that the best path forward is to close this PR. This was a really insightful discussion, and I appreciate the detailed feedback from both of you. I'll go ahead and close this now.

Rocketknight1:
Hang on, my mistake! I dug into the FlashAttention docs and I was wrong; as long as `position_ids` and `cu_seq_lens` are set correctly then there shouldn't be information leakage and performance shouldn't degrade. I didn't realize that `cu_seq_lens` controlled attention masking like that.

I'm debating whether it's worth adding this after all, in that case

yashisthebatman:
@Rocketknight1 Hey, wow, thank you so much for digging into the FlashAttention docs and for that update! That's a super important clarification and really great to know.
No problem at all, I'll keep the PR open. Please take your time to think it over. If you do decide it's a worthwhile addition, just let me know if there's anything else you'd like me to do. No rush at all!

yashisthebatman:
@Rocketknight1 any update ?

Rocketknight1:
Will try to review it tomorrow!

src/transformers/data/data_collator.py
```diff
@@ -1411,13 +1417,19 @@ def __call__(self, features, return_tensors=None, separator_id=None):
...
-                batch["labels"] += [separator_id] + sample["labels"][1:]
+                if is_labels_sequence:
+                    # Original logic for token-level labels.
+                    batch["labels"] += [self.separator_id] + sample["labels"][1:]
+                else:
+                    # Default "safe" behavior: broadcast the integer label to all tokens.
+                    batch["labels"] += [sample["labels"]] * len(input_ids)
             else:
-                batch["labels"] += [separator_id] + input_ids[1:]
+                batch["labels"] += [self.separator_id] + input_ids[1:]
```
Rocketknight1, line 1432: I think we should revert this change, since it overrides any passed `separator_id`
```suggestion
                batch["labels"] += [separator_id] + input_ids[1:]
```

src/transformers/data/data_collator.py
```diff
@@ -1403,6 +1406,9 @@ def __call__(self, features, return_tensors=None, separator_id=None):
         if separator_id is None:
             separator_id = self.separator_id
         is_labels_provided = "labels" in features[0]
+
+        is_labels_sequence = is_labels_provided and isinstance(features[0].get("labels"), (list, tuple, np.ndarray))
```
Rocketknight1, line 1410: Is this case exhaustive? I think `torch.tensor` may also be possible.

Rocketknight1 (approved):
With apologies for the extremely long delay, I finally did a proper review! I think it's good, with two comments. The old behaviour for sequence labels is preserved, and the behaviour in the case of non-sequence labels was definitely unwanted before, so this seems like a straightforward improvement. 

Take a look at the two highlighted areas and let me know when you're ready for final approval + merging

PR #41690, opened 2025-10-17
further improve `utils/check_bad_commit.py` (#41658)

ydshieh:
# What does this PR do?

### DIFF
--- a/.github/workflows/check_failed_tests.yml
+++ b/.github/workflows/check_failed_tests.yml
@@ -41,7 +41,10 @@ env:
 
 jobs:
   check_new_failures:
-    name: " "
+    name: "Find commits for new failing tests"
+    strategy:
+      matrix:
+        run_idx: [1]
     runs-on:
       group: aws-g5-4xlarge-cache
     container:
@@ -118,6 +121,10 @@ jobs:
         run: |
           python3 utils/print_env.py
 
+      - name: Install pytest-flakefinder
+        if: ${{ env.process == 'true' }}
+        run: python3 -m pip install pytest-flakefinder
+
       - name: Show installed libraries and their versions
         working-directory: /transformers
         if: ${{ env.process == 'true' }}
@@ -126,20 +133,50 @@ jobs:
       - name: Check failed tests
         working-directory: /transformers
         if: ${{ env.process == 'true' }}
-        run: python3 utils/check_bad_commit.py --start_commit ${{ inputs.start_sha }} --end_commit ${{ env.END_SHA }} --file ci_results_${{ inputs.job }}/new_failures.json --output_file new_failures_with_bad_commit.json
+        run: python3 utils/check_bad_commit.py --start_commit ${{ inputs.start_sha }} --end_commit ${{ env.END_SHA }} --file ci_results_${{ inputs.job }}/new_failures.json --output_file new_failures_with_bad_commit_${{ inputs.job }}_${{ matrix.run_idx }}.json
 
       - name: Show results
         working-directory: /transformers
         if: ${{ env.process == 'true' }}
         run: |
-          ls -l new_failures_with_bad_commit.json
-          cat new_failures_with_bad_commit.json
+          ls -l new_failures_with_bad_commit_${{ inputs.job }}_${{ matrix.run_idx }}.json
+          cat new_failures_with_bad_commit_${{ inputs.job }}_${{ matrix.run_idx }}.json
 
-      - name: Checkout back
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: new_failures_with_bad_commit_${{ inputs.job }}_${{ matrix.run_idx }}
+          path: /transformers/new_failures_with_bad_commit_${{ inputs.job }}_${{ matrix.run_idx }}.json
+
+  process_new_failures_with_commit_info:
+    name: "process bad commit reports"
+    needs: [check_new_failures]
+    runs-on:
+      group: aws-g5-4xlarge-cache
+    container:
+      image: ${{ inputs.docker }}
+      options: --gpus all --shm-size "16gb" --ipc host -v /mnt/cache/.cache/huggingface:/mnt/cache/
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          pattern: new_failures_with_bad_commit_${{ inputs.job }}*
+          path: /transformers/new_failures_with_bad_commit_${{ inputs.job }}
+          merge-multiple: true
+
+      - name: Check files
+        working-directory: /transformers
+        run: |
+          ls -la /transformers
+          ls -la /transformers/new_failures_with_bad_commit_${{ inputs.job }}
+
+      # Currently, we only run with a single runner by using `run_idx: [1]`. We might try to run with multiple runners
+      # to further reduce the false positive caused by flaky tests, which requires further processing to merge reports.
+      - name: Merge files
+        shell: bash
         working-directory: /transformers
         if: ${{ env.process == 'true' }}
         run: |
-          git checkout ${{ inputs.start_sha }}
+          cp /transformers/new_failures_with_bad_commit_${{ inputs.job }}/new_failures_with_bad_commit_${{ inputs.job }}_1.json new_failures_with_bad_commit.json
 
       - name: Process report
         shell: bash
--- a/.github/workflows/check_failed_tests.yml
+++ b/.github/workflows/check_failed_tests.yml
@@ -188,6 +188,10 @@ jobs:
         run: |
           cp /transformers/new_failures_with_bad_commit_${{ inputs.job }}/new_failures_with_bad_commit_${{ inputs.job }}_1.json new_failures_with_bad_commit.json
 
+      - name: Update clone
+        working-directory: /transformers
+        run: git fetch && git checkout ${{ inputs.commit_sha || github.sha }}
+
       - name: Process report
         shell: bash
         working-directory: /transformers
--- a/.github/workflows/check_failed_tests.yml
+++ b/.github/workflows/check_failed_tests.yml
@@ -165,6 +165,11 @@ jobs:
     steps:
       - uses: actions/download-artifact@v4
         with:
+          name: ci_results_${{ inputs.job }}
+          path: /transformers/ci_results_${{ inputs.job }}
+
+      - uses: actions/download-artifact@v4
+        with:
           pattern: new_failures_with_bad_commit_${{ inputs.job }}*
           path: /transformers/new_failures_with_bad_commit_${{ inputs.job }}
           merge-multiple: true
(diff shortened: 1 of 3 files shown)


.github/workflows/check_failed_tests.yml
```diff
@@ -41,7 +41,10 @@ env:
 
 jobs:
   check_new_failures:
-    name: " "
+    name: "Find commits for new failing tests"
+    strategy:
+      matrix:
+        run_idx: [1]
```
ydshieh, line 47: In the future, we might use multiple runners to further reduce the chance of reporting flaky tests

.github/workflows/check_failed_tests.yml
```diff
@@ -118,6 +121,10 @@ jobs:
         run: |
           python3 utils/print_env.py
 
+      - name: Install pytest-flakefinder
+        if: ${{ env.process == 'true' }}
+        run: python3 -m pip install pytest-flakefinder
```
ydshieh, line 131: But so far, it's easier and more reliable to use this pytest-plugin `pytest-flakefinder` to run the same test multiple times in a single pytest launch

.github/workflows/check_failed_tests.yml
```diff
@@ -126,20 +133,50 @@ jobs:
       - name: Check failed tests
         working-directory: /transformers
         if: ${{ env.process == 'true' }}
-        run: python3 utils/check_bad_commit.py --start_commit ${{ inputs.start_sha }} --end_commit ${{ env.END_SHA }} --file ci_results_${{ inputs.job }}/new_failures.json --output_file new_failures_with_bad_commit.json
+        run: python3 utils/check_bad_commit.py --start_commit ${{ inputs.start_sha }} --end_commit ${{ env.END_SHA }} --file ci_results_${{ inputs.job }}/new_failures.json --output_file new_failures_with_bad_commit_${{ inputs.job }}_${{ matrix.run_idx }}.json
```
ydshieh, line 141: just to avoid name collapsing. not so important so far:

- `${{ inputs.job }}`: only the job will run this workflow
- `${{ matrix.run_idx }}`: currently only 1 runner

but no harm to add them here

.github/workflows/check_failed_tests.yml
```diff
@@ -126,20 +133,50 @@ jobs:
...
-          ls -l new_failures_with_bad_commit.json
-          cat new_failures_with_bad_commit.json
+          ls -l new_failures_with_bad_commit_${{ inputs.job }}_${{ matrix.run_idx }}.json
+          cat new_failures_with_bad_commit_${{ inputs.job }}_${{ matrix.run_idx }}.json
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: new_failures_with_bad_commit_${{ inputs.job }}_${{ matrix.run_idx }}
+          path: /transformers/new_failures_with_bad_commit_${{ inputs.job }}_${{ matrix.run_idx }}.json
```
ydshieh, line 154: need this as we now separate the 1 job into 2 jobs:

- `check_new_failures`: using `matrix`
- `process_new_failures_with_commit_info` (belwo)

and the 2nd job need the file from the 1st job

.github/workflows/check_failed_tests.yml
```diff
@@ -126,20 +133,50 @@ jobs:
...
+      group: aws-g5-4xlarge-cache
+    container:
+      image: ${{ inputs.docker }}
+      options: --gpus all --shm-size "16gb" --ipc host -v /mnt/cache/.cache/huggingface:/mnt/cache/
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          pattern: new_failures_with_bad_commit_${{ inputs.job }}*
+          path: /transformers/new_failures_with_bad_commit_${{ inputs.job }}
+          merge-multiple: true
```
ydshieh, line 175: download the artifact generated from the 1st job

.github/workflows/check_failed_tests.yml
```diff
@@ -126,20 +133,50 @@ jobs:
...
-      - name: Checkout back
+      # Currently, we only run with a single runner by using `run_idx: [1]`. We might try to run with multiple runners
+      # to further reduce the false positive caused by flaky tests, which requires further processing to merge reports.
+      - name: Merge files
+        shell: bash
         working-directory: /transformers
         if: ${{ env.process == 'true' }}
         run: |
-          git checkout ${{ inputs.start_sha }}
+          cp /transformers/new_failures_with_bad_commit_${{ inputs.job }}/new_failures_with_bad_commit_${{ inputs.job }}_1.json new_failures_with_bad_commit.json
```
ydshieh, line 189: at this moment, this is dummy operation, not really merge

utils/process_bad_commit_report.py
```diff
@@ -26,6 +26,33 @@
...
+
+    report_repo_id = os.getenv("REPORT_REPO_ID")
+
+    commit_info = api.upload_file(
+        path_or_fileobj="new_failures_with_bad_commit.json",
+        path_in_repo=f"{report_repo_folder}/ci_results_{job_name}/new_failures_with_bad_commit.json",
+        repo_id=report_repo_id,
+        repo_type="dataset",
+        token=os.environ.get("TRANSFORMERS_CI_RESULTS_UPLOAD_TOKEN", None),
+    )
```
ydshieh, line 54: just upload to hub for `new_failures_with_bad_commit.json`

not really necessary, but sometimes handy to deal with this format

utils/check_bad_commit.py
```diff
@@ -38,26 +39,35 @@ def create_script(target_test):
...
+    ["python3", "-m", "pytest", "-v", "--flake-finder", "--flake-runs=4", "-rfEp", f"{target_test}"],
     capture_output = True,
     text=True,
 )
 print(result.stdout)
 
 if f"FAILED {target_test}" in result.stdout:
     print("test failed")
-    exit(2)
+    exit(1)
```
ydshieh, line 57: no particular reason. I don't remember why I use this `2` previously

utils/check_bad_commit.py
```diff
@@ -38,26 +39,35 @@ def create_script(target_test):
...
+        # if a test is added between the `good commit` (exclusive) and `bad commit` (inclusive) (in git bisect terminology).
+        # So we return 0 here in order to allow the process being able to identify the first commit that fails the test.
         exit(0)
     elif "ERROR: not found: " in result.stderr:
         print("test not found in this commit")
         exit(0)
     else:
         print(f"pytest gets unknown error: {{result.stderr}}")
-        exit(-1)
+        exit(1)
```
ydshieh, line 70: Let's not fail git bisect (i.e. return -1) if pytest have any issue.

If `pytest` failed for whatever reason, we treat it as a bad commit.

utils/check_bad_commit.py
```diff
@@ -67,20 +77,63 @@ def create_script(target_test):
...
+    result = subprocess.run(
+        ["python3", "target_script.py"],
+        capture_output=True,
+        text=True,
+    )
+
+    # Restore to original commit
+    repo.git.checkout(original_head)
+
+    return result.returncode != 0
```
ydshieh, line 100: used to determine the the commits (start, end) are passing, failing a test or if the test is flaky.

This is important to not give confusing false alert of failing test on a wrong commit.

utils/check_bad_commit.py
```diff
@@ -67,20 +77,63 @@ def create_script(target_test):
...
+            None,
+            f"flaky: test fails on the current CI run but passed in the previous run which is running on the same commit {end_commit}.",
+        )
+
+    # Now, we are (almost) sure `target_test` is not failing at `end_commit`
+    # check if `start_commit` fail the test
+    failed_now = is_bad_commit(target_test, start_commit)
+    if not failed_now:
+        # failed on CI run, but not reproducible here --> don't report
+        return None, f"flaky: test fails on the current CI run (commit: {start_commit}) but passes during the check."
```
ydshieh, line 136: the logic implemented to not producing confusing alerts at the 2 boundries.

We need this because `git bisect` won't run on those 2 commits, as it assume one is the good commit and the other is the bad commit as you pass to it.

Cyrilvallez (approved):
You're the boss for this!

.github/workflows/check_failed_tests.yml
vasqu: Iirc, it is also a quite old pkg at this point. Might be worth to check out if there are alternatives, not a prio

utils/check_bad_commit.py
```diff
@@ -38,26 +39,35 @@ def create_script(target_test):
...
 if f"FAILED {target_test}" in result.stdout:
     print("test failed")
-    exit(2)
+    exit(1)
 elif result.returncode != 0:
     if "ERROR: file or directory not found: " in result.stderr:
         print("test file or directory not found in this commit")
+        # git bisect treats exit code 125 as `test not found`. But this causes it not be able to make the conclusion
+        # if a test is added between the `good commit` (exclusive) and `bad commit` (inclusive) (in git bisect terminology).
+        # So we return 0 here in order to allow the process being able to identify the first commit that fails the test.
```
vasqu, line 63: Nice comment :pray:

utils/check_bad_commit.py
```diff
@@ -117,7 +170,7 @@ def find_bad_commit(target_test, start_commit, end_commit):
     print(f"Between `start_commit` {start_commit} and `end_commit` {end_commit}")
     print(f"bad_commit: {bad_commit}\n")
 
-    return bad_commit
+    return bad_commit, "git bisect find the bad commit."
```
vasqu:
```suggestion
    return bad_commit, "git bisect found the bad commit."
```
nit

utils/check_bad_commit.py
```diff
@@ -38,26 +39,35 @@ def create_script(target_test):
...
 
+_ = subprocess.run(
+    ["python3", "-m", "pip", "install", "-e", "."],
+    capture_output = True,
+    text=True,
+)
+
 result = subprocess.run(
-    ["python3", "-m", "pytest", "-v", "-rfEp", f"{target_test}"],
+    ["python3", "-m", "pytest", "-v", "--flake-finder", "--flake-runs=4", "-rfEp", f"{target_test}"],
```
vasqu, line 49: Why 4? Any particular reason?

vasqu (approved):
LGTM, moreso nits and one question on flake finder usage

.github/workflows/check_failed_tests.yml
ydshieh: yeah, but might be also that is mature and don't need many new feature (as it does one thing and only that thing)

utils/check_bad_commit.py
ydshieh: No, I feel `2` won't be enough,  `3` might be good enough to detect flaky,  so I choose `4` to feel more comfortable 😆

PR #41692, opened 2025-10-17
[`Attn`] Allow dynamic causality in SDPA via Kwargs

vasqu:
As per title, it's
- 1. Nice for power users
- 2. Apparently some models do need to switch like CLIP (depending on their input, see https://github.com/huggingface/transformers/pull/41673#discussion_r2438734215)

This allows us to rely on the set module's attribute per default but overwrite with kwarg if given. cc @zucchini-nlp

### DIFF
--- a/src/transformers/integrations/flash_attention.py
+++ b/src/transformers/integrations/flash_attention.py
@@ -34,6 +34,7 @@ def flash_attention_forward(
     scaling: Optional[float] = None,
     sliding_window: Optional[int] = None,
     softcap: Optional[float] = None,
+    is_causal: Optional[bool] = None,
     **kwargs,
 ) -> tuple[torch.Tensor, None]:
     if kwargs.get("output_attentions", False):
@@ -64,9 +65,7 @@ def flash_attention_forward(
     target_dtype = get_target_dtype(query, module)
 
     # Instead of relying on the value set in the module directly, we use the is_causal passed in kwargs if it is presented
-    is_causal = kwargs.pop("is_causal", None)
-    if is_causal is None:
-        is_causal = module.is_causal
+    is_causal = is_causal if is_causal is not None else module.is_causal
 
     attn_output = _flash_attention_forward(
         query,
--- a/src/transformers/integrations/sdpa_attention.py
+++ b/src/transformers/integrations/sdpa_attention.py
@@ -68,13 +68,22 @@ def sdpa_attention_forward(
     if attention_mask is not None and attention_mask.ndim == 4:
         attention_mask = attention_mask[:, :, :, : key.shape[-2]]
 
-    # We dispatch to SDPA's Flash Attention or Efficient kernels via this `is_causal` if statement instead of an inline conditional assignment
-    # in SDPA to support both torch.compile's dynamic shapes and full graph options. An inline conditional prevents dynamic shapes from compiling.
-    # Note that it is important to check first for the shape, otherwise compile will fail with `argument 'is_causal' must be bool, not SymBool`
-    if is_causal is None:
-        # The last condition is for encoder (decoder) models which specify this by passing their own `is_causal` flag
-        # This is mainly due to those models having mixed implementations for encoder, decoder, and encoder-decoder attns
-        is_causal = query.shape[2] > 1 and attention_mask is None and getattr(module, "is_causal", True)
+    # Kwarg takes precedence over the defined module's attribute
+    # - Allows dynamic switching, e.g. when model's switch based on the model input type (CLIP)
+    # - Defaults to "normal" behavior for all attention types (encoder, decoder, cross)
+    is_causal = is_causal if is_causal is not None else getattr(module, "is_causal", True)
+
+    # SDPA's Flash Attention (and cuDNN) kernels rely on the `is_causal` flag. However, there are certain conditions:
+    # - Not in decoding phase (== full attention)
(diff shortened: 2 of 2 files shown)


src/transformers/integrations/flash_attention.py
```diff
@@ -34,6 +34,7 @@ def flash_attention_forward(
     scaling: Optional[float] = None,
     sliding_window: Optional[int] = None,
     softcap: Optional[float] = None,
+    is_causal: Optional[bool] = None,
```
vasqu, line 37: Imo, it's nicer to define this as explicit kwarg here instead of doing `kwarg.get...`

src/transformers/integrations/sdpa_attention.py
```diff
@@ -68,13 +68,22 @@ def sdpa_attention_forward(
...
-    if is_causal is None:
-        # The last condition is for encoder (decoder) models which specify this by passing their own `is_causal` flag
-        # This is mainly due to those models having mixed implementations for encoder, decoder, and encoder-decoder attns
-        is_causal = query.shape[2] > 1 and attention_mask is None and getattr(module, "is_causal", True)
+    # Kwarg takes precedence over the defined module's attribute
+    # - Allows dynamic switching, e.g. when model's switch based on the model input type (CLIP)
+    # - Defaults to "normal" behavior for all attention types (encoder, decoder, cross)
+    is_causal = is_causal if is_causal is not None else getattr(module, "is_causal", True)
+
+    # SDPA's Flash Attention (and cuDNN) kernels rely on the `is_causal` flag. However, there are certain conditions:
```
vasqu, line 74: Reworked the comments to explain this in more detail

src/transformers/integrations/sdpa_attention.py
```diff
@@ -68,13 +68,22 @@ def sdpa_attention_forward(
...
-    # We dispatch to SDPA's Flash Attention or Efficient kernels via this `is_causal` if statement instead of an inline conditional assignment
-    # in SDPA to support both torch.compile's dynamic shapes and full graph options. An inline conditional prevents dynamic shapes from compiling.
-    # Note that it is important to check first for the shape, otherwise compile will fail with `argument 'is_causal' must be bool, not SymBool`
-    if is_causal is None:
-        # The last condition is for encoder (decoder) models which specify this by passing their own `is_causal` flag
-        # This is mainly due to those models having mixed implementations for encoder, decoder, and encoder-decoder attns
-        is_causal = query.shape[2] > 1 and attention_mask is None and getattr(module, "is_causal", True)
+    # Kwarg takes precedence over the defined module's attribute
+    # - Allows dynamic switching, e.g. when model's switch based on the model input type (CLIP)
+    # - Defaults to "normal" behavior for all attention types (encoder, decoder, cross)
```
Cyrilvallez: nit: this comment looks a bit complicated to me, a simple "we give precedence to kwarg, then module if not present" as for fa would be clearer IMO, but no strong opinion - feel free to disregard if you think otherwise!

zucchini-nlp (approved):
Functionality-wise looks good to me. I think we can delete now passing it explicitly as in 

https://github.com/huggingface/transformers/blob/12a50f294d50e3d0e124511f2b6f43625f73ffce/src/transformers/models/idefics/vision.py#L254 and fix models like CLIP that change self attributes at run-time. Searching shows a few models with similar pattern

src/transformers/integrations/sdpa_attention.py
```diff
@@ -68,13 +68,22 @@ def sdpa_attention_forward(
...
+    is_causal = is_causal if is_causal is not None else getattr(module, "is_causal", True)
+
+    # SDPA's Flash Attention (and cuDNN) kernels rely on the `is_causal` flag. However, there are certain conditions:
+    # - Not in decoding phase (== full attention)
+    # - Attention mask is not to be provided (even if it is a causal pattern)
+    # - Internally, we marked this as compatible with causal, i.e. it is a decoder attention type
```
Cyrilvallez: Actually, isn't the decoding case valid as well? Looking [here](https://github.com/huggingface/transformers/blob/main/src/transformers/masking_utils.py#L253-L254), it seems that using `is_causal = True/False` would be the same with a single-row matrix, i.e. we take the whole row anyway

Or the comment on the masking is wrong, and it's simply that we can skip the mask as well, but then we need full attention on the single query token (I think it's this one, do you mind updating comment on the mask?)

```suggestion
    # - Not in decoding phase (otherwise we want full attention on the single query token)
    # - Attention mask is not to be provided (even if it is a causal pattern)
    # - Internally, we marked this as compatible with causal, i.e. it is a decoder attention type
```

Cyrilvallez (approved):
Nice, thanks for improving! Just 2 small comments!

src/transformers/integrations/sdpa_attention.py
vasqu: Yea, you're correct it's the case with only one query token. Much better comment lemme update this with yours

src/transformers/integrations/sdpa_attention.py
vasqu: Lol, yea fair enough updating with the fa version in a second

vasqu:
https://github.com/huggingface/transformers/pull/41692#pullrequestreview-3350812991 cc @yonigozlan @molbap for vision models refactors to keep in mind!

PR #41694, opened 2025-10-17
[Parakeet] add output_attention_mask

eustlb:
# What does this PR do?

Parakeet encoder modified the sequence length, making that the input attention_mask cannot be directly used to know which embeddings positions correspond to padding. 

Here we do it differently than the actual external approach like we have on mimi that consists to retreive the correct mask:
```python
attention_mask = mimi_model.get_audio_codes_mask(inputs.attention_mask)
```
and that recomputes as mask that is already known in the forward and that can simply be returned along with the outputs.

### DIFF
--- a/src/transformers/models/parakeet/modeling_parakeet.py
+++ b/src/transformers/models/parakeet/modeling_parakeet.py
@@ -37,6 +37,16 @@ from ...utils.generic import check_model_inputs
 from .configuration_parakeet import ParakeetCTCConfig, ParakeetEncoderConfig
 
 
+@dataclass
+@auto_docstring(
+    custom_intro="""
+    Extends [~modeling_outputs.BaseModelOutput] to include the output attention mask since sequence length is not preserved in the model's forward.
+    """
+)
+class ParakeetEncoderModelOutput(BaseModelOutput):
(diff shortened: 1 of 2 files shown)


src/transformers/models/parakeet/modeling_parakeet.py
```diff
@@ -567,7 +581,9 @@ def forward(
                     **kwargs,
                 )
 
-        return BaseModelOutput(last_hidden_state=hidden_states)
+        return ParakeetEncoderModelOutput(
+            last_hidden_state=hidden_states, attention_mask=output_mask.int() if output_attention_mask else None
```
nithinraok, line 585: can we return lengths directly instead or along with attention_mask

eustlb: I'd rather keep an explicit approach here for potiential future usage of the same with left padding, you can retreive lengths by doing `attention_mask.sum(-1)`

nithinraok: sounds good.

ArthurZucker (approved):
Okay sounds good !

PR #41695, opened 2025-10-17
🚨 Delete generation params from model config

zucchini-nlp:
# What does this PR do?

Disentangles config and generation config. Now the model's config will throw an error if generation params are found. To keep BC, we'll still be loading models from the hub where generation params are in config, but we'll manually drop these params from being set as an attribute.


docs/source/en/model_doc/qwen2_audio.md
```diff
@@ -54,7 +54,7 @@ processor = AutoProcessor.from_pretrained("Qwen/Qwen2-Audio-7B", trust_remote_co
 prompt = "<|audio_bos|><|AUDIO|><|audio_eos|>Generate the caption in English:"
 url = "https://qianwen-res.oss-cn-beijing.aliyuncs.com/Qwen-Audio/glass-breaking-151256.mp3"
 audio, sr = librosa.load(BytesIO(urlopen(url).read()), sr=processor.feature_extractor.sampling_rate)
-inputs = processor(text=prompt, audios=audio, return_tensors="pt").to(model.device)
+inputs = processor(text=prompt, audio=audio, return_tensors="pt").to(model.device)
```
zucchini-nlp, line 57: Audio renaming is not related, accidentally slipped from another branch. I'll rebase main after https://github.com/huggingface/transformers/pull/41681 is merged

zucchini-nlp:
cc @vasqu maybe since Joao left and I need someone to help merge generation PRs. This one particularly is blocking me. I'd love to clean-up and break BC for v5, since generation config is too complicated at this point

This PR particularly splits generation config away from model config, a dependency we had for a few ages due to legacy reasons. It's time to move on I think and raise a `ValueError` if users add manually generation params in a model config

NOTE: bad formatted configs from hub can still be loaded for BC

src/transformers/generation/configuration_utils.py
```diff
@@ -918,7 +918,10 @@ def from_pretrained(
         else:
             logger.info(f"loading configuration file {configuration_file} from cache at {resolved_config_file}")
 
-        if kwargs.get("return_unused_kwargs") is True:
+        if kwargs.get("_from_model_config", False):
+            config = cls.from_model_config(config_dict)
+            return config
```
```suggestion
            return cls.from_model_config(config_dict)
```
nit

src/transformers/generation/utils.py
```diff
@@ -410,6 +410,14 @@ def adjust_generation_fn(
...
                     "Generation config file not found, using a generation config created from the model config."
                 )
+                self.generation_config = GenerationConfig.from_pretrained(
+                    pretrained_model_name_or_path,
+                    config_file_name="config.json",
+                    _from_auto=from_auto_class,
+                    _from_pipeline=from_pipeline,
+                    _from_model_config=True,
+                    **repo_loading_kwargs,
+                )
```
vasqu, line 420: This is the BC loading from the saved config, just to check iiuc

src/transformers/generation/utils.py
```diff
@@ -1773,14 +1781,12 @@ def _prepare_generation_config(
...
-                        "You have modified the pretrained model configuration to control generation. This is a"
-                        " deprecated strategy to control generation and will be removed in v5."
+                    raise ValueError(
+                        "You have modified the pretrained model configuration to control generation."
+                        " This strategy to control generation is not supported anymore. "
                         " Please use and modify the model generation configuration (see"
                         " https://huggingface.co/docs/transformers/generation_strategies#default-text-generation-configuration )",
-                        UserWarning,
                     )
-                    self.generation_config = new_generation_config
```
vasqu, line 1788: Will this be BC? Our workaround uses `_from_model_config` and may have non-default values. Might be wrong, but would double check

src/transformers/modeling_utils.py
```diff
@@ -1826,6 +1826,8 @@ def __init__(self, config: PreTrainedConfig, *inputs, **kwargs):
         self.config._attn_implementation_internal = self._check_and_adjust_attn_implementation(
             self.config._attn_implementation, is_init_check=True
         )
+        if self.can_generate():
+            self.generation_config = GenerationConfig.from_model_config(config)
```
vasqu, line 1460: Why did we need to change this

src/transformers/models/bart/configuration_bart.py
```diff
@@ -172,11 +172,10 @@ def __init__(
         )
 
         # ensure backward compatibility for BART CNN models
-        if self.forced_bos_token_id is None and kwargs.get("force_bos_token_to_be_generated", False):
```
vasqu, line 169: Update comment?

tests/utils/test_modeling_utils.py
```diff
@@ -1742,35 +1731,32 @@ def prepare_inputs_for_generation(self):
 
     def test_save_and_load_config_with_custom_generation(self):
         """
-        Regression test for the ability to save and load a config with a custom generation kwarg (i.e. a parameter
-        that gets moved to the generation config and reset on the model config)
+        Tests that saving and loading a config with a custom generation kwarg iss not possible
```
vasqu:
```suggestion
        Tests that saving and loading a config with a custom generation kwarg is not possible
```

vasqu (commented):
I'm in general aligned with this, just have a few nits/questions just to be sure.

Don't forget
- the bad audio(s) 
- v5 tag and add it to the issue

src/transformers/generation/utils.py
zucchini-nlp: yep, some very old models used to save all params in `config.json`, so we have to support it

src/transformers/generation/utils.py
zucchini-nlp: This error is supposed to be raised only if users change the config (not generation config) manually after loading the model. For ex: 

```python
# First load any old model where config.json has generation params
model = load_model()
model.config.do_sample = True
model.generate(input_ids) # raise error that you should not modify the config

```

src/transformers/modeling_utils.py
zucchini-nlp: It the same line which I deleted below and tbh no reason to change the order. Most prob there was a reason at first, but then I changed my mind 😄

src/transformers/models/bart/configuration_bart.py
zucchini-nlp: it is still for BC, we just won't have `force_bos_token_to_be_generated` by default. It will be available only if config has the value saved

zucchini-nlp: In general. "model.config" used to contain generation params until this PR, i.e. we could do `model.config.do_sample` and get a default value. From now on, it will throw attribute error unless the model's config actually had a "do_sample" saved in json file

src/transformers/generation/utils.py
vasqu: Gotcha, thx for clarifying

src/transformers/modeling_utils.py
vasqu: It happens :laughing: let's keep it, doesn't have any effect either way

src/transformers/models/bart/configuration_bart.py
vasqu: Gotcha, makes sense

vasqu (approved):
Thx, lgtm

Time to merge with main and check conflicts :eyes: just the same things as I said before re audios + v5

zucchini-nlp:
Rebased main, now the diff has no `audio`-related changes. Added the v5 tag and a rotation light as it's a bit breaking. Merging

PR #38301, opened 2025-05-22
🔴[`Attention`] Bert-based Models Attention Refactor

vasqu:
Keeping track of the models that are done:
- [x] Bert
- [x] Roberta
- [x] Albert
- [x] Data2VecText
- [x] Xmod 
- [x] Electra 
- [x] XLM Roberta 
- [x] Roberta Prelayernorm
- [x] Ernie 
- [x] Camembert 
- [x] Bert Generation 
- [x] XLM Robert XL 
- [x] RoCBert
- [x] Mobile Bert

Up to discussion:
- Flash attention is flaky; I suspect the norm layers to be responsible (encountered similar things with gemma3)

Would need another round; questionable if worth it (ordered by prio):
- [ ] Tapas
- [ ] Vision x Text models, e.g.
  - [ ] Bridgetower (text would be good to go, would need image counterpart)
  - [ ] Altclip 
  - [ ] ...
- [ ] RemBert
- [ ] Megatron Bert

PR #41696, opened 2025-10-17
[`PEFT`] Fix prefix tuning

vasqu:
Prefix tuning is a bit more of a special case where we have different q and kv lengths despite having the encoder attention. We now rely on the mask for kv length now (if provided). The assumption was holding before because causal masks always have q == kv length. Fixes https://github.com/huggingface/transformers/pull/38301#issuecomment-3415672560

cc @BenjaminBossan


vasqu:
run-slow: bert,bart

vasqu:
run-slow: bert,bart

BenjaminBossan (approved):
Thanks a lot for taking quick care of this regression.

I also tested this PR with the [original PEFT unit tests that were failing](https://github.com/huggingface/peft/actions/runs/18595690268/job/53021214843#step:5:2288) and they pass now!

BenjaminBossan:
Ping @ArthurZucker could we please get this in before the next release?

src/transformers/masking_utils.py
```diff
@@ -784,9 +784,14 @@ def _preprocess_mask_arguments(
...
-    # Otherwise, the sizes are simply the input sizes
+    # Otherwise, we infer based on our input
     else:
-        kv_length, kv_offset = input_embeds.shape[1], 0
+        # 1. Rely on input directly
+        if attention_mask is None:
+            kv_length, kv_offset = input_embeds.shape[1], 0
+        # 2. Rely on the mask instead - needed for special cases like prefix tuning in PEFT
+        else:
+            kv_length, kv_offset = attention_mask.shape[-1], 0
```
Cyrilvallez, line 713: A bit sad to see the default case becoming the mask here, as the `input_embeds` is much less error prone IMO 🥲
Could you quickly explain why we need it to be the 2d mask size? If the created 4d mask does not match the input size, how do we avoid a dimension crash later on when entering attention computation?

vasqu: Hmm, not sure what you mean by the 2D issue? We index with `-1` which works with any mask shape from 2-4D as the last dim is always the kv index. 

Do you have an example? Maybe you're concerned about SWA, chunked, etc? Will take another look

Cyrilvallez (dismissed):
All right, let's go for it even if it's quite a weird pattern, as both should always be the same in general 🥲

Cyrilvallez (approved):
Alright, after all the discussions it's a bit too hard to fix in peft directly, so let's go for it!

PR #39483, opened 2025-07-17
Bye bye env vars, keep everything as configs

SunMarc:
# What does this PR do?

Supersedes https://github.com/huggingface/transformers/pull/37259

As @BenjaminBossan found, (https://github.com/huggingface/accelerate/pull/3252), the TrainingArguments will set environmental variables automatically when using Accelerate because before it wouldn't work otherwise. Nowadays the only env variable required for things to run smoothly is the ones for model init (fsdp cpu eff ram).

This PR does a few things:
- We completely remove the need for environmental variables, creating the proper configs (dynamo relies on https://github.com/huggingface/accelerate/pull/3251)
- I've refactored how mixed_precision gets set, to simplify the training arguments and combine 7 args into 2.
- Removes old references/updates the logic in Trainer to reflect the choices

PR #41697, opened 2025-10-17
[Trainer] remove env vars 

SunMarc:
# What does this PR do?

This PR removes env var as much as possible from trainer.
Similar to https://github.com/huggingface/transformers/pull/39483/files but with less changes.


SunMarc:
run_slow: trainer

SunMarc:
run_slow: fsdp,deepspeed

src/transformers/training_args.py
```diff
@@ -2762,6 +2645,126 @@ def set_dataloader(
...
+            sync_module_states = str(self.fsdp_config.get("sync_module_states", "true")).lower()
+            cpu_ram_efficient_loading = str(self.fsdp_config.get("cpu_ram_efficient_loading", "false")).lower()
+            if sync_module_states == "false" and cpu_ram_efficient_loading == "true":
+                # In this case, all the processes except the main process would have random weights leading
+                # to unexpected behaviour during training, thus throwing error here to prevent it.
+                raise ValueError('`sync_module_states` must be `"True"` if `cpu_ram_efficient_loading` is `"True"`')
+
+            # we need to set the env here as otherwise we get a warning in accelerate + we need to set it for transformers
+            fsdp_plugin_args["cpu_ram_efficient_loading"] = cpu_ram_efficient_loading
+            os.environ["FSDP_CPU_RAM_EFFICIENT_LOADING"] = cpu_ram_efficient_loading
```
BenjaminBossan, line 2760: Is this env var needed or could it be set directly on `fsdp_plugin`?

BenjaminBossan (approved):
Overall, nice improvement. I don't have a lot to add, just one question.

src/transformers/training_args.py
SunMarc: We are setting it in`fsdp_plugin` but we are also sending a warning in that case. So it is better to make sure it is set to not trigger any warning. Also, we only use this env in transformers actually for the following: 

```python
def is_fsdp_enabled():
    if is_torch_available():
        import torch

        return (
            torch.distributed.is_available()
            and torch.distributed.is_initialized()
            and strtobool(os.environ.get("ACCELERATE_USE_FSDP", "False")) == 1
            and strtobool(os.environ.get("FSDP_CPU_RAM_EFFICIENT_LOADING", "False")) == 1
        )

    return False
```

SunMarc:
run_slow: fsdp

SunMarc:
run_slow: fsdp

SunMarc:
run_slow: fsdp

SunMarc:
run_slow: fsdp

SunMarc:
run_slow: fsdp

SunMarc:
run_slow: fsdp

SunMarc:
run_slow: fsdp

SunMarc:
run_slow: fsdp

SunMarc:
run_slow: fsdp

PR #41699, opened 2025-10-17
Simplify GQA conditions in sdpa_attention.py

justinchuby:
Removed unnecessary checks for key being a torch.fx.Proxy in GQA conditions because fx tracing is no longer supported, and torch.export supports enable_gqa.

### DIFF
--- a/src/transformers/integrations/sdpa_attention.py
+++ b/src/transformers/integrations/sdpa_attention.py
@@ -32,13 +32,11 @@ def use_gqa_in_sdpa(attention_mask: Optional[torch.Tensor], key: torch.Tensor) -
     # 1.cuda or Ascend NPU
     #   - torch version >= 2.5
     #   - attention_mask is None (otherwise it will fall back to the math kernel)
-    #   - key is not a torch.fx.Proxy (otherwise it will fail with a tracing error)
     # 2.xpu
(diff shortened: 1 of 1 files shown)


vasqu (approved):
Perfect, thank you! Makes more sense esp since we removed fx support explicitly in #41683

src/transformers/integrations/sdpa_attention.py
```diff
@@ -32,13 +32,11 @@ def use_gqa_in_sdpa(attention_mask: Optional[torch.Tensor], key: torch.Tensor) -
     # 1.cuda or Ascend NPU
     #   - torch version >= 2.5
     #   - attention_mask is None (otherwise it will fall back to the math kernel)
```
justinchuby, line 34: Is it required that attention_mask is None? Just wanting to check it the situation has changed

vasqu: Yup, at least on older versions it is! Was only recently relaxed

vasqu: See the description here https://docs.pytorch.org/docs/2.5/generated/torch.nn.functional.scaled_dot_product_attention.html (under gqa --> supports only flash and math --> flash only works without a mask)

We want to avoid the math kernel

PR #41701, opened 2025-10-17
Fix qwen3_vl mix precision dtype

theruiwang:
`fast_pos_embed_interpolate` returns `pos_embeds` in the same dtype as the master weights.

Therefore, when the master weights are in FP32 but the forward pass runs in BF16, `hidden_states` will be upcast to FP32, causing dtype mismatches with other activations.

CC @yonigozlan @molbap @zucchini-nlp

### DIFF
--- a/src/transformers/models/qwen3_vl/modular_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/modular_qwen3_vl.py
@@ -640,10 +640,11 @@ class Qwen3VLVisionModel(Qwen3VLPreTrainedModel):
         Returns:
             `torch.Tensor`: hidden_states.
         """
+        input_dtype = hidden_states.dtype
         hidden_states = self.patch_embed(hidden_states)
 
         pos_embeds = self.fast_pos_embed_interpolate(grid_thw)
-        hidden_states = hidden_states + pos_embeds
+        hidden_states = (hidden_states + pos_embeds).to(input_dtype)
 
         rotary_pos_emb = self.rot_pos_emb(grid_thw)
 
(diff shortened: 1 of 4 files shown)


src/transformers/models/qwen3_vl/modular_qwen3_vl.py
```diff
@@ -640,10 +640,11 @@ def forward(self, hidden_states: torch.Tensor, grid_thw: torch.Tensor, **kwargs)
         Returns:
             `torch.Tensor`: hidden_states.
         """
+        input_dtype = hidden_states.dtype
         hidden_states = self.patch_embed(hidden_states)
 
         pos_embeds = self.fast_pos_embed_interpolate(grid_thw)
-        hidden_states = hidden_states + pos_embeds
+        hidden_states = (hidden_states + pos_embeds).to(input_dtype)
```
zucchini-nlp, line 647: i think we should cast only `pos_embeds` to the input dtype here

molbap: Agreed, further couldn't we fix in `fast_pos_embed_interpolate` instead of recasting? To avoid too many conversions - could for instance pass the `input_dtype`. 
In ` h_idxs = torch.linspace(0, self.num_grid_per_side - 1, h)` passing the wanted dtype should be enough no?

zucchini-nlp: Thought the same thing, but I am not sure if positional embedding was intentionally done in full precision for better performance 🤔

theruiwang: Yeah, this change intentionally keeps it running in FP32/same dtype as master weights for now, without changing the numerical dynamics.

Casting only `pos_embeds` or fixing inside `fast_pos_embed_interpolate` would have numerical implications, which requires ablations with training results if we want to be careful.

Happy to discuss — I'm leaning towards not changing model behaviors for now.

theruiwang: Friendly reminder.

Let me know what you think.

CC @molbap @zucchini-nlp

PR #41592, opened 2025-10-14
Improve AutoTokenizer error message for Voxtral models missing mistral-common

Khansa435:
## What does this PR do?
Adds a clearer ImportError message when users try to load a Voxtral tokenizer without having `mistral-common` installed.

## Issue
Fixes #41553
Fixes misleading `TypeError: not a string` when loading Voxtral models.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?
@ArthurZucker @Rocketknight1 
Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #41718, opened 2025-10-18
AutoTokenizer: clear ImportError when loading Voxtral without mistral-common + unit test

Elon7069:
Summary: Fix confusing TypeError from SentencePiece when loading Voxtral tokenizer without mistral-common by raising a clear ImportError instead. Fixes #41553.
Rationale: Without mistral-common, loading via AutoTokenizer ends in a low-level TypeError inside sentencepiece. Users should see a direct, actionable message.
What’s changed
AutoTokenizer guard:
In AutoTokenizer.from_pretrained, after resolving config and before class selection, raise a clear error if config.model_type == "voxtral" and mistral-common is missing.
Message:
"The Voxtral tokenizer requires the 'mistral-common' package. Please install it using pip install mistral-common."
Tests:
Added tests/models/voxtral/test_tokenization_voxtral.py
Mocks is_mistral_common_available to False and get_tokenizer_config to avoid network, then asserts ImportError mentioning "mistral-common".
Why this approach
Keeps mapping logic intact; avoids unexpected fallbacks.
Ensures the user sees a clear, actionable message as early as possible in the loading path.
Testing
Targeted test: pytest tests/models/voxtral/test_tokenization_voxtral.py -q → passes.
No network calls thanks to test monkeypatching.
Backward compatibility
No behavior change when mistral-common is installed.
Only affects Voxtral when the dependency is missing.

### DIFF
--- a/src/transformers/models/auto/tokenization_auto.py
+++ b/src/transformers/models/auto/tokenization_auto.py
@@ -1096,6 +1096,12 @@ class AutoTokenizer:
                 trust_remote_code, pretrained_model_name_or_path, has_local_code, has_remote_code, upstream_repo
             )
 
+        # Detect missing dependency for Voxtral early and provide a clear error message
+        if getattr(config, "model_type", None) == "voxtral" and not is_mistral_common_available():
+            raise ImportError(
+                "The Voxtral tokenizer requires the 'mistral-common' package. Please install it using `pip install mistral-common`."
+            )
+
         if has_remote_code and trust_remote_code:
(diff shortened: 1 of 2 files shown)


Elon7069:
Maintainer please check my PR lemme know what i have to improve more

Rocketknight1:
I think this is a duplicate of #41592!

Elon7069:
I think both are different ki lemme check> I think this is a duplicate of #41592!

PR #41725, opened 2025-10-19
Add GLPNImageProcessorFast 

Aravind-11:
# What does this PR do?

This PR adds a **fast image processor for the GLPN model**, implemented as `GLPNImageProcessorFast`.  

Fixes # (issue)

## Before submitting
- Implements `GLPNImageProcessorFast` using `BaseImageProcessorFast`.
- Adds tests and documentation updates.

### 🧪 Testing
- All tests pass except for the (`test_slow_fast_equivalence_batched`). I would like some help here. 

### 📄 Files updated
- `src/transformers/models/glpn/image_processing_glpn_fast.py`
- `src/transformers/models/glpn/__init__.py`
- `src/transformers/models/auto/image_processing_auto.py`
- `tests/models/glpn/test_image_processing_glpn.py`
- `docs/source/en/model_doc/glpn.md`

## Before submitting
- [x] Read the [contributor guidelines](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request).
- [x] Updated documentation and tests.
- [x] Verified style and quality with `make style` and `make quality`.

## Who can review?

@yonigozlan @molbap


tests/models/glpn/test_image_processing_glpn.py
```diff
@@ -109,55 +125,70 @@ def test_image_processor_properties(self):
...
-
-        # Test not batched input (GLPNImageProcessor doesn't support batching)
         encoded_images = image_processing(image_inputs[0], return_tensors="pt").pixel_values
         expected_output_image_shape = self.image_processor_tester.expected_output_image_shape(image_inputs)
         self.assertTrue(tuple(encoded_images.shape) == (1, *expected_output_image_shape))
 
     def test_call_numpy_4_channels(self):
-        # Initialize image_processing
         image_processing = self.image_processing_class(**self.image_processor_dict)
-        # create random numpy tensors
```
molbap, line 153: Agreed comments are not very informative but no need to remove them all imo

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,230 @@
...
+        for k, v in d.items():
+            if k in keep_always or k in keep_values:
+                out[k] = v
+            elif k in null_out:
+                out[k] = None
+            else:
+                # For any other unexpected fast-only keys, set None to be safe
+                out[k] = None
+
+        return out
```
molbap: This is a bit too convoluted IMO - we can just pop out unwanted keys like we do in other serializing methods in fast image processors (also, no need for emojis in the comments)

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,230 @@
...
+                tensor_type = return_tensors
+            else:
+                # mimic slow processor: leave as list so BatchFeature won't tensorize
+                processed = [img.cpu().numpy() for img in reordered]
+                tensor_type = None
+        else:
+            processed = reordered
+            tensor_type = None
+
+        return BatchFeature(data={"pixel_values": processed}, tensor_type=tensor_type)
```
molbap: this parts isn't "fast": it converts to numpy when shapes differ, it's why the test `test_slow_fast_equivalence_batched` fails, when shapes differ tensor_type is set to `None`

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,230 @@
...
+        if predicted_depth.ndim == 4 and predicted_depth.shape[1] == 1:
+            predicted_depth = predicted_depth.squeeze(1)
+        elif predicted_depth.ndim == 3:
+            pass
+        else:
+            # fallback: ensure (B, H, W)
+            if predicted_depth.ndim == 4:
+                predicted_depth = predicted_depth[:, 0, ...]
+            else:
+                raise ValueError("Unexpected depth prediction shape")
```
molbap: I don't entirely understand why we need to guard against shape mismatches like this

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,230 @@
...
+        results = []
+        target_sizes = target_sizes or [None] * predicted_depth.shape[0]
+        for depth, tgt in zip(predicted_depth, target_sizes):
+            if tgt is not None:
+                # slow adds [None, None, ...], interpolates, then squeezes
+                d = depth[None, None, ...]
+                d = torch.nn.functional.interpolate(d, size=tgt, mode="bicubic", align_corners=False)
+                depth = d.squeeze(0).squeeze(0)
+            results.append({"predicted_depth": depth})
+        return results
```
molbap: let's avoid single-letter variables and abbreviations

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,230 @@
...
+
+from ...image_processing_utils import BatchFeature
+from ...image_processing_utils_fast import BaseImageProcessorFast, group_images_by_shape, reorder_images
+from ...image_utils import (
+    IMAGENET_STANDARD_MEAN,
+    IMAGENET_STANDARD_STD,
+    PILImageResampling,
+)
+
+# optional typing container (similar to ZoeDepthImageProcessorKwargs)
```
molbap:
```suggestion
```

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,230 @@
...
+)
+
+
+class GLPNImageProcessorKwargs(ImagesKwargs, total=False):
+    # Public (persisted) key — must match slow processor:
+    size_divisor: int
+    # Back-compat alias (NOT persisted):
+    ensure_multiple_of: int
+    # Allow overriding resample (persisted like slow):
+    resample: PILImageResampling
```
molbap: Why isn't this imported?

molbap (commented):
Hey, thanks for starting this! Left some initial comments :)

src/transformers/models/glpn/image_processing_glpn_fast.py
Aravind-11: My bad. We don't need that. It's not defined in the slow processor. I've removed it.

src/transformers/models/glpn/image_processing_glpn_fast.py
Aravind-11: Got it.

tests/models/glpn/test_image_processing_glpn.py
Aravind-11: Added them back.

Aravind-11:
> Hey, thanks for starting this! Left some initial comments :)

Thanks a lot for reviewing Pablo! I've made the changes.

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,199 @@
...
+        processed_groups = {}
+        sd = size_divisor if size_divisor is not None else self.size_divisor
+
+        for shape, stacked_images in grouped_images.items():
+            if do_resize:
+                stacked_images = self._crop_to_multiple(stacked_images, sd)
+            if do_rescale:
+                stacked_images = self.rescale(stacked_images, rescale_factor)
+            if do_normalize:
+                stacked_images = self.normalize(stacked_images, image_mean, image_std)
```
molbap: We can fuse the rescale and normalize ops with `rescale_and_normalize`

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,199 @@
...
+    ) -> BatchFeature:
+        """
+        GLPN fast preprocessing:
+        - crop to floored multiple of size_divisor
+        - rescale [0,1]
+        - normalize (off by default)
+        """
+        # avoid validation error: inject dummy size/resample for validate_preprocess_arguments
+        if size is None:
+            size = {"height": 480, "width": 640}
```
molbap: that should not be needed, let's define defaults in the `__init__` rather

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,199 @@
...
+    - (No normalization by default)
+    """
+
+    # Persist ONLY the same keys as the slow processor
+    do_resize = True
+    do_rescale = True
+    do_normalize = False
+    resample = PILImageResampling.BILINEAR
+    size_divisor = 32
+    # Don't persist an explicit `size` for GLPN (slow doesn't)
```
molbap: it's fine to persist here

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,199 @@
...
+    image_mean = IMAGENET_STANDARD_MEAN
+    image_std = IMAGENET_STANDARD_STD
+    size = {"height": 480, "width": 640}  # only for validation; we still crop, not resize
+    interpolation = F.InterpolationMode.BILINEAR
+    # valid_kwargs = GLPNImageProcessorKwargs
```
molbap:
```suggestion
    valid_kwargs = GLPNImageProcessorKwargs
```

molbap: + import that from slow

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,199 @@
...
+    # Persist ONLY the same keys as the slow processor
+    do_resize = True
+    do_rescale = True
+    do_normalize = False
+    resample = PILImageResampling.BILINEAR
+    size_divisor = 32
+    # Don't persist an explicit `size` for GLPN (slow doesn't)
+    image_mean = IMAGENET_STANDARD_MEAN
+    image_std = IMAGENET_STANDARD_STD
+    size = {"height": 480, "width": 640}  # only for validation; we still crop, not resize
```
molbap: ah but size is actually defined here - no need to re-define it after!

src/transformers/models/glpn/image_processing_glpn_fast.py
molbap: hey, I'm pretty confident `test_slow_fast_equivalence_batched` will fail with this setup currently - also looking at the `slow` test, what would cause the shapes to become heterogeneous, not resizing? In that case let's pad the batch and return it as a tensor IMO

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,199 @@
...
+            "default_to_square",
+            "data_format",  # Fast processor params
+        }
+
+        # Set all other keys to None (don't persist their values)
+        for key in list(d.keys()):
+            if key not in keys_to_keep:
+                d[key] = None
+
+        return d
```
molbap: no single-letter variables, please

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,199 @@
...
+
+    @torch.no_grad()
```
molbap:
```suggestion
```

tests/models/glpn/test_image_processing_glpn.py
```diff
@@ -161,3 +176,39 @@ def test_call_numpy_4_channels(self):
         expected_output_image_shape = self.image_processor_tester.expected_output_image_shape(image_inputs)
         self.assertTrue(tuple(encoded_images.shape) == (1, *expected_output_image_shape))
         self.image_processing_class.num_channels = 3
+
+    def test_equivalence_slow_fast(self):
```
molbap: Naming should align with the rest of the lib:

```suggestion
    def test_slow_fast_equivalence(self):
```

and another test should be added `test_slow_fast_equivalence_batched`

molbap (commented):
Thanks for iterating! Did a second review 🤗

src/transformers/models/glpn/image_processing_glpn_fast.py
Aravind-11: Ahh my bad! Sorry.

src/transformers/models/glpn/image_processing_glpn_fast.py
Aravind-11: I defined the kwargs in the slow processor and imported it.

src/transformers/models/glpn/image_processing_glpn_fast.py
Aravind-11: Got it.

Aravind-11:
> Thanks for iterating! Did a second review 🤗

Thank you! I've made the changes.

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,208 @@
...
+    image_mean = IMAGENET_STANDARD_MEAN
+    image_std = IMAGENET_STANDARD_STD
+    interpolation = F.InterpolationMode.BILINEAR
+    valid_kwargs = GLPNImageProcessorKwargs
+
+    # If BaseImageProcessorFast supports it, this makes persistence explicit:
+    try:
+        config_keys = {"do_resize", "size_divisor", "resample", "do_rescale"}
+    except Exception:
+        pass
```
molbap: I'm not sure why we want to persist these keys? Might be a misunderstanding on my end

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    Fast image processor for GLPN using the Torch/TorchVision backend.
+
+    Performs:
+    - Crop H,W down to the nearest multiple of `size_divisor` (default 32)
+    - Rescale [0,255] → [0,1]
+    - (No normalization by default)
+    """
+
+    # Persist ONLY the same keys as the slow processor
```
molbap: same question, why do we need this?

Aravind-11:
> > Thanks for iterating! Did a second review 🤗
> 
> Thank you! I've made the changes.

Hi! Is there further review required and anything I should change in the implementation? Please let me know. Thank you!

src/transformers/models/glpn/image_processing_glpn.py
```diff
@@ -223,6 +236,26 @@ def preprocess(
...
+                for img in images:
+                    h, w = img.shape[-2:]
+                    if h < max_height or w < max_width:
+                        # Create padded array with zeros
+                        padded = np.zeros((*img.shape[:-2], max_height, max_width), dtype=img.dtype)
+                        padded[..., :h, :w] = img
+                        padded_images.append(padded)
+                    else:
+                        padded_images.append(img)
+                images = padded_images
```
molbap: Let's use `np.pad` in the slow path

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,208 @@
...
+                stacked_images = self._crop_to_multiple(stacked_images, sd, interpolation)
+            stacked_images = self.rescale_and_normalize(
+                stacked_images, do_rescale, rescale_factor, do_normalize, image_mean, image_std
+            )
+            processed_groups[shape] = stacked_images
+
+        reordered = reorder_images(processed_groups, grouped_index)
+
+        if return_tensors:
+            # Detect heterogeneous shapes
```
molbap: are there heterogeneous shapes or not? else a pattern like 

```python
        processed_images = torch.stack(processed_images, dim=0) if return_tensors else processed_images

        return BatchFeature(data={"pixel_values": processed_images}, tensor_type=return_tensors)
```
would be much preferred. Else let's at least extract the padding logic to a function, look in image processing utils fast, there's a padding method already. Why not use it?

src/transformers/models/glpn/image_processing_glpn_fast.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        Resize images (B,C,H,W) by flooring H and W to nearest multiple of `size_divisor`.
+        Uses interpolation to match slow GLPN behavior.
+        """
+        _, _, h, w = images.shape
+        new_h = (h // size_divisor) * size_divisor
+        new_w = (w // size_divisor) * size_divisor
+        if (new_h, new_w) == (h, w):
+            return images
+        # Resize (not crop) to match slow processor behavior
```
molbap:
```suggestion
```

molbap (commented):
I left additional comments because I'm not 100% convinced by the padding logic, let's make sure it's needed, and if it is let's use existing methods!

src/transformers/models/glpn/image_processing_glpn_fast.py
Aravind-11: Yes, its producing heterogenous shapes. I used the pad function from utils.

src/transformers/models/glpn/image_processing_glpn_fast.py
Aravind-11: Removed them.

Aravind-11:
> I left additional comments because I'm not 100% convinced by the padding logic, let's make sure it's needed, and if it is let's use existing methods!

Thanks a lot for reviewing! Appreciate your help.

yonigozlan (approved):
Hey @Aravind-11, thanks a lot for working on this! I made some final changes to get this merged. Mostly removed the padding logic so as not to break BC as it wasn't in the original image processor.
I'll merge when the CI passes!

Aravind-11:
> Hey @Aravind-11, thanks a lot for working on this! I made some final changes to get this merged. Mostly removed the padding logic so as not to break BC as it wasn't in the original image processor. I'll merge when the CI passes!

Thank you so much @yonigozlan! for the necessary commits and review! Does the failing 'tests_non_model' arise from the pr?

yonigozlan:
> Thank you so much @yonigozlan! for the necessary commits and review! Does the failing 'tests_non_model' arise from the pr?

I don't think so, I'm seeing it in other PRs...

PR #41742, opened 2025-10-20
Fix typo in LFM-VL

zucchini-nlp:
# What does this PR do?

As per title, the typo was reported by the authors, as they are planning another release soon. The configs one the hub were saved with an "underscore"


yonigozlan (approved):
Sg I hate "-" in model types anyway 😁

PR #41754, opened 2025-10-21
Add pytree registration for static cache

angelayi:
* Refactored the existing pytree registration for DynamicCache to be a more generalized form
* Also fixed the existing pytree registration which triggers [lazy_initialization](https://github.com/huggingface/transformers/blob/9aab965b1e61d92d402809bd467c317ec464e560/src/transformers/cache_utils.py#L94-L95). This shouldn't be needed because we already have all the contents of the dynamic cache created?
* Added pytree support for StaticCache

cc @tugsbayasgalan

### DIFF
--- a/src/transformers/integrations/executorch.py
+++ b/src/transformers/integrations/executorch.py
@@ -15,13 +15,18 @@ from collections.abc import Callable
 from typing import Optional
 
 import torch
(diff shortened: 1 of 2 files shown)


ArthurZucker (commented):
Happy to merge at your convenience! Sorry missed the notif

PR #41755, opened 2025-10-21
Swap columns and rows of the grid layout in LFM2-VL

ankke:
Fixing swapped rows and columns that were resulting in incorrect positional special tokens. Fixed integration tests on the modeling file.

@zucchini-nlp


zucchini-nlp (approved):
Thanks! And for the batched test, I think we will have to fix it in separate PR. I will trigger slow CI and it might complain that the test is failing

zucchini-nlp:
run-slow: lfm2_vl

zucchini-nlp:
Interesting, in the runners the batched test gives better generation such as `"In this image, we see a cat and a dog lying on a pink blanket. They are both sleeping peacefully. They are"` . Could then prob be smth with the versions of installed packaged or hardware

PR #39396, opened 2025-07-14
[gemma3] fix bidirectional image mask

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/39389. There were two issues. First, using `getattr` on dict will never look for a dict key thus we were skipping on `token_type_ids_mask`. 

Second, `kv_idx` can go beyond `token_type_ids` length because it includes the new tokens to be generated. Using a simple check on `ids.shape` fails in vmap due to dynamic control flow. One option would be to create dummy padded token type ids using `kv_idx` but I think a simple try-except is also fine

PR #41757, opened 2025-10-21
Fix CUDA index out of bounds for q_idx in VLM token type masking for Gemma3, PaliGemma, and example modular

### LINKED ISSUE #4281: CI fails with dev dependencies: torch.AcceleratorError: CUDA error: device-side assert triggered
CI fails with dev dependencies: https://github.com/huggingface/trl/actions/runs/18522980056/job/52787019404
> torch.AcceleratorError: CUDA error: device-side assert triggered
```python
FAILED tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_vlm_0_trl_internal_testing_tiny_Gemma3ForConditionalGeneration - torch.AcceleratorError: CUDA error: device-side assert triggered
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

ERROR tests/test_rloo_trainer.py::TestRLOOTrainer::test_training_vlm_0_trl_internal_testing_tiny_Gemma3ForConditionalGeneration - torch.AcceleratorError: CUDA error: device-side assert triggered
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```

albertvillanova:
Fix CUDA index out of bounds error that occurs during generation with static caches when using token type IDs for bidirectional image attention masking.

### Background

After PR
- #41505 

changed cache initialization behavior in `generate()`, a **latent bug in the VLM masking code was exposed**. The error manifests as:

```python
CUDA error: device-side assert triggered
/pytorch/aten/src/ATen/native/cuda/IndexKernel.cu:113: operator(): block: [0,0,0], thread: [0,0,0]
Assertion -sizes[i] <= index && index < sizes[i] && "index out of bounds" failed.
```

### Bug

In the `token_type_ids_mask_function` inner mask, the code correctly handles out-of-bounds `kv_idx` values but **fails to handle out-of-bounds `q_idx` values**.

The PR
- #39396 

originally fixed the bidirectional image masking by adding bounds checking for `kv_idx`, but overlooked that `q_idx` needed the same protection.

During generation with static caches:
- Cache shapes can exceed the actual input sequence length (e.g., static cache of 2048 positions with 512 token input)
- The masking function receives both `q_idx` and `kv_idx` that can exceed `token_type_ids.shape[1]`
- Direct indexing like `token_type_ids[batch_idx, q_idx]` causes CUDA index out of bounds errors when `q_idx >= token_type_ids.shape[1]`

The code comment on line 740 already acknowledged this issue:
> "NOTE: static cache shape goes beyond input seq length, while token_type_ids.shape[1] == input seq length"

Bounds checking was implemented for `kv_idx`, but `q_idx` was overlooked.

### Fix

This PR adds the same `torch.where` bounds-checking pattern for `q_idx` that already existed for `kv_idx`:
1. Create `safe_q_idx` to clamp indices within valid range
2. Use safe indices for tensor access
3. Apply `torch.where` to mask out-of-bounds values with appropriate defaults (0 for `token_type_ids`, -1 for `image_group_ids`)

### Affected Models
- `Gemma3ForConditionalGeneration`
- `PaliGemmaForConditionalGeneration`
- Example modular transformer template (`modeling_new_task_model.py`)

### Testing

This PR fixes the downstream failing test in TRL:
> tests/test_grpo_trainer.py::TestGRPOTrainer::test_training_vlm_0_trl_internal_testing_tiny_Gemma3ForConditionalGeneration

See associated issue:
- https://github.com/huggingface/trl/issues/4281

### Related Issues

- Regression exposed by: cache initialization refactor
  - #41505

- This PR completes the fix started in (which fixed `kv_idx` but missed `q_idx`):
  - #39396

- This PR will fix https://github.com/huggingface/trl/issues/4281

CC:
- @gante, who made the PR #41505
- @zucchini-nlp , who made the PR #39396

### DIFF
--- a/examples/modular-transformers/modeling_new_task_model.py
+++ b/examples/modular-transformers/modeling_new_task_model.py
@@ -125,15 +125,23 @@ def token_type_ids_mask_function(
         # If it's 1 for both query and key/value, we are in an image block
         # NOTE: static cache shape goes beyond input seq length, while token_type_ids.shape[1] == input seq length
         # Since vmap doesn't support `if statement` we workaround it with `torch.where`
-        safe_idx = torch.where(kv_idx < token_type_ids.shape[1], kv_idx, 0)
-        token_type_ids_at_kv_idx = token_type_ids[batch_idx, safe_idx]
+        safe_q_idx = torch.where(q_idx < token_type_ids.shape[1], q_idx, 0)
+        safe_kv_idx = torch.where(kv_idx < token_type_ids.shape[1], kv_idx, 0)
+
+        token_type_ids_at_q_idx = token_type_ids[batch_idx, safe_q_idx]
+        token_type_ids_at_q_idx = torch.where(q_idx < token_type_ids.shape[1], token_type_ids_at_q_idx, 0)
+
+        token_type_ids_at_kv_idx = token_type_ids[batch_idx, safe_kv_idx]
         token_type_ids_at_kv_idx = torch.where(kv_idx < token_type_ids.shape[1], token_type_ids_at_kv_idx, 0)
 
-        image_group_ids_at_kv_idx = image_group_ids[batch_idx, safe_idx]
+        image_group_ids_at_q_idx = image_group_ids[batch_idx, safe_q_idx]
+        image_group_ids_at_q_idx = torch.where(q_idx < image_group_ids.shape[1], image_group_ids_at_q_idx, -1)
+
+        image_group_ids_at_kv_idx = image_group_ids[batch_idx, safe_kv_idx]
         image_group_ids_at_kv_idx = torch.where(kv_idx < image_group_ids.shape[1], image_group_ids_at_kv_idx, -1)
 
-        is_image_block = (token_type_ids[batch_idx, q_idx] == 1) & (token_type_ids_at_kv_idx == 1)
-        same_image_block = image_group_ids[batch_idx, q_idx] == image_group_ids_at_kv_idx
+        is_image_block = (token_type_ids_at_q_idx == 1) & (token_type_ids_at_kv_idx == 1)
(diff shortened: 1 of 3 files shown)


mellowpraful:
unsubscribe

zucchini-nlp (commented):
Makes sense to me! Can you also check `pytest -k test_generate_with_static_cache tests/models/gemma3/test_modeling_gemma3.py`, since it was supposed to fails for gemma3 in that case?

Prob the test doesn't pass token type ids, or is already failing on main and we didn't notice it?

albertvillanova:
Thanks for your review, @zucchini-nlp.

I have run the tests as requested and everything is OK:
```python
pytest -k test_generate_with_static_cache tests/models/gemma3/test_modeling_gemma3.py

tests/models/gemma3/test_modeling_gemma3.py::Gemma3TextModelTest::test_generate_with_static_cache PASSED                                                                                            [ 50%]
tests/models/gemma3/test_modeling_gemma3.py::Gemma3Vision2TextModelTest::test_generate_with_static_cache PASSED                                                                                     [100%]

============================================================================== 2 passed, 373 deselected, 2 warnings in 9.62s ==============================================================================
```

zucchini-nlp:
Thanks

zucchini-nlp (approved):
Oh right, have to approve to merge

albertvillanova:
Thank YOU, @zucchini-nlp for your fast and efficient review!

PR #41763, opened 2025-10-21
Timesfm 2.5

kashif:
# What does this PR do?

Add TimesFM 2.5 model

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


kashif:
@rajatsen91 @siriuz42 as this port is more or less ready can I kindly request a review and also could you kindly run the conversion script and perhaps upload the resulting checkpoints in the transformers format to the hub to say the `http://huggingface.co/google/timesfm-2.5-200m-transformers` or any appropriate place so I can then test it? thanks

kashif:
thanks @rajatsen91 all is working with support for flash atten and flex attention backends as well

rajatsen91 (approved):

docs/source/en/model_doc/timesfm2_5.md
```diff
@@ -0,0 +1,80 @@
...
+
+The abstract from the paper is the following:
+
+*Motivated by recent advances in large language models for Natural Language Processing (NLP), we design a time-series foundation model for forecasting whose out-of-the-box zero-shot performance on a variety of public datasets comes close to the accuracy of state-of-the-art supervised forecasting models for each individual dataset. Our model is based on pretraining a decoder style attention model with input patching, using a large time-series corpus comprising both real-world and synthetic datasets. Experiments on a diverse set of previously unseen forecasting datasets suggests that the model can yield accurate zero-shot forecasts across different domains, forecasting horizons and temporal granularities.*
+
+This model was contributed by [kashif](https://huggingface.co/kashif). The original code can be found [here](https://github.com/google-research/timesfm).
+
+You can find the checkpoint at [`google/timesfm-2.5-200m-transformers`](https://huggingface.co/google/timesfm-2.5-200m-transformers).
```
vasqu, line 38: Taking a look at the conversion in a second but any reason we could not utilize the weight converter instead? I think that would make it easier for users if there were only one repo but more of a pointer, not a must if another repo is preferred

src/transformers/models/auto/configuration_auto.py
```diff
@@ -442,6 +442,7 @@
         ("textnet", "TextNetConfig"),
         ("time_series_transformer", "TimeSeriesTransformerConfig"),
         ("timesfm", "TimesFmConfig"),
+        ("timesfm_2p5", "Timesfm2P5Config"),
```
vasqu: Just a convention thing: Imo it would make more sense to use something along `timesfm2_5`, similar to qwen3_5

src/transformers/models/timesfm2_5/convert_timesfm2_5_original_to_hf.py
```diff
@@ -0,0 +1,355 @@
...
+        use_qk_norm=True,
+        use_per_dim_scale=True,
+        use_bias=False,
+        activation="swish",
+        query_pre_attn_scalar=256.0,
+        quantiles=[0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9],
+        max_position_embeddings=16384,
+        rope_theta=10000.0,
+    )
+    timesfm_config.save_pretrained(tmp_model_path)
```
vasqu, line 106: I only see the config as what could be problematic? The conversions seem fairly basic

src/transformers/models/timesfm2_5/convert_timesfm2_5_original_to_hf.py
vasqu, line 1: Licence missing

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
+# Copyright 2025 the HuggingFace Team. All rights reserved.
```
vasqu:
```suggestion
# Copyright 2026 the HuggingFace Team. All rights reserved.
```

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+    TimesFmOutput,
+    TimesFmOutputForPrediction,
+    TimesFmPreTrainedModel,
+)
+
+
+logger = logging.get_logger(__name__)
+
+
+def timesfm_2p5_eager_attention_forward(
```
vasqu: Not sure why we need a separate implementation here, llama is pretty much the same as here (minus the `repeat_kv` which would be no ops). 

You can just import `eager_attention_forward` from llama and modular handles the deps itself --> llama eager instead of gemma one

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+
+    return attn_output, attn_weights
+
+
+class Timesfm2P5Config(TimesFmConfig):
+    """
+    Configuration class for TimesFM 2.5 model.
+
+    Example checkpoint: [google/timesfm-2.5-200m-transformers](https://huggingface.co/google/timesfm-2.5-200m-transformers).
+    """
```
vasqu: Let's make a proper docstring for the attributes etc similar to https://github.com/huggingface/transformers/blob/65935e09212bcd5c3ceecbd6e12ace3ec26a804f/src/transformers/models/timesfm/configuration_timesfm.py#L24-L73

utils/check_config_attributes.py
```diff
@@ -33,6 +33,14 @@
...
 SPECIAL_CASES_TO_ALLOW = {
     "ExaoneMoeConfig": ["first_k_dense_replace"],  # BC for other frameworks
+    "Timesfm2P5Config": [
+        "freq_size",
+        "normalize_inputs",
+        "pad_val",
+        "tolerance",
+        "use_positional_embedding",
+        "use_rotary_embeddings",
+    ],  # Inherited from parent TimesFmConfig
```
vasqu: We can instead delete them if they are really not used, after the super init, e.g. `del self.attr`

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        initializer_range: float = 0.02,
+        min_timescale: float = 1.0,
+        max_timescale: float = 10000.0,
+        # Override defaults for 2.5
+        num_hidden_layers: int = 20,
+        output_quantile_len: int = 1024,  # From original TimesFM 2.5 config
+        decode_index: int = 5,
+        use_rotary_embeddings: bool = True,
+        use_qk_norm: bool = True,
+        use_per_dim_scale: bool = True,
```
vasqu: Are there models with different values here? Otherwise, I'd like to avoid these if possible especially the rotary one

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        use_bias: bool = False,
+        activation: str = "swish",
+        use_positional_embedding: bool = False,  # TimesFM 2.5 uses rotary embeddings instead
```
vasqu: Same here, we should not have attributes that are not used

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        normalize_inputs: bool = True,
+        force_flip_invariance: bool = True,
+        infer_is_positive: bool = True,
+        # Gemma2-compatible parameters for query scaling
+        query_pre_attn_scalar: float = 256.0,  # This provides the per-dim scaling
+        attn_logit_softcapping: float | None = None,
+        layer_types: list | None = None,  # All layers are the same type
+        sliding_window: int | None = None,  # No sliding window
+        max_position_embeddings: int = 16384,  # Should match context_length
+        rope_theta: float = 10000.0,  # RoPE theta parameter
```
vasqu: Not needed, we provide this ourselves https://github.com/huggingface/transformers/blob/65935e09212bcd5c3ceecbd6e12ace3ec26a804f/src/transformers/modeling_rope_utils.py#L630

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        activation: str = "swish",
+        use_positional_embedding: bool = False,  # TimesFM 2.5 uses rotary embeddings instead
+        use_continuous_quantile_head: bool = True,
+        normalize_inputs: bool = True,
+        force_flip_invariance: bool = True,
+        infer_is_positive: bool = True,
+        # Gemma2-compatible parameters for query scaling
+        query_pre_attn_scalar: float = 256.0,  # This provides the per-dim scaling
+        attn_logit_softcapping: float | None = None,
+        layer_types: list | None = None,  # All layers are the same type
```
vasqu: Would also like to avoid if not needed, rather rewrite than force inheritance of unused stuff

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        sliding_window: int | None = None,  # No sliding window
+        max_position_embeddings: int = 16384,  # Should match context_length
+        rope_theta: float = 10000.0,  # RoPE theta parameter
+        rope_parameters: dict | None = None,
+        **kwargs,
+    ):
+        self.rope_parameters = rope_parameters
```
vasqu, line 161: If the original model had the init on top, you can also move it to the bottom so that we can apply our rope conversion logic https://github.com/huggingface/transformers/blob/65935e09212bcd5c3ceecbd6e12ace3ec26a804f/src/transformers/modeling_rope_utils.py#L632

You won't need `self.rope_parameters = self.rope_parameters or {"rope_type": "default", "rope_theta": rope_theta}` at the end then

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+@dataclass
+@auto_docstring
+class Timesfm2P5Output(TimesFmOutput):
+    r"""
+    context_mu (`torch.Tensor` of shape `(batch_size, num_patches)`):
+        Running means computed per input patch during normalization.
+    context_sigma (`torch.Tensor` of shape `(batch_size, num_patches)`):
+        Running standard deviations computed per input patch during normalization.
+    patch_padding (`torch.Tensor` of shape `(batch_size, num_patches)`):
+        Boolean mask indicating fully padded patches.
```
vasqu: Any reason this is needed in the output?

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+    full_predictions (`torch.Tensor` of shape `(batch_size, horizon_length, quantiles)`):
+        Quantile forecasts including the median after denormalization.
+    loss (`torch.Tensor` of shape `(1,)`, *optional*, returned when `future_values` is provided):
+        Training loss combining MSE and quantile losses when targets are supplied.
+    """
+
+    pass
+
+
+class Timesfm2P5MLP(nn.Module):
```
vasqu: Can be inherited from clip https://github.com/huggingface/transformers/blob/65935e09212bcd5c3ceecbd6e12ace3ec26a804f/src/transformers/models/clip/modeling_clip.py#L347

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+
+        Args:
+            x: Input tensor of shape (batch_size, ..., input_dims)
+
+        Returns:
+            Output tensor of shape (batch_size, ..., output_dims)
+        """
+        hidden = self.hidden_layer(x)
+        hidden = self.activation(hidden)
+        output = self.output_layer(hidden)
```
vasqu: Is this not essentially the same as in `Timesfm2P5MLP`. It could not if the dimensionalities are a bit different ig but then I'd rather make it possible in the MLP to change dims respectively

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        residual = self.residual_layer(x)
+        return output + residual
```
vasqu: The residual style can be adapted in the module above, e.g. https://github.com/huggingface/transformers/blob/65935e09212bcd5c3ceecbd6e12ace3ec26a804f/src/transformers/models/llama/modeling_llama.py#L332-L336

Ok, now seeing that it likely stems from the timesfm model so we should inherit from that one!

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+    """
+
+    def __init__(self, config: Timesfm2P5Config, layer_idx: int):
+        super().__init__(config, layer_idx)
+
+        # Add QK normalization specific to TimesFM 2.5
+        self.use_qk_norm = getattr(config, "use_qk_norm", True)
+        if self.use_qk_norm:
+            self.query_ln = Timesfm2P5RMSNorm(self.head_dim, eps=config.rms_norm_eps)
+            self.key_ln = Timesfm2P5RMSNorm(self.head_dim, eps=config.rms_norm_eps)
```
vasqu:
```suggestion
        # Add QK normalization specific to TimesFM 2.5
        self.use_qk_norm = config.use_qk_norm
        if self.use_qk_norm:
            self.q_norm = Timesfm2P5RMSNorm(self.head_dim, eps=config.rms_norm_eps)
            self.k_norm = Timesfm2P5RMSNorm(self.head_dim, eps=config.rms_norm_eps)
```
If we have different models with different values here, it should be defined in the config

Imo, we could use afmoe which uses q/k norm

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+
+
+class Timesfm2P5Config(TimesFmConfig):
+    """
+    Configuration class for TimesFM 2.5 model.
+
+    Example checkpoint: [google/timesfm-2.5-200m-transformers](https://huggingface.co/google/timesfm-2.5-200m-transformers).
+    """
+
+    model_type = "timesfm_2p5"
```
vasqu: Noticing no TP plan, but the model is small so doubt it would be valuable to add

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        # Add QK normalization specific to TimesFM 2.5
+        self.use_qk_norm = getattr(config, "use_qk_norm", True)
+        if self.use_qk_norm:
+            self.query_ln = Timesfm2P5RMSNorm(self.head_dim, eps=config.rms_norm_eps)
+            self.key_ln = Timesfm2P5RMSNorm(self.head_dim, eps=config.rms_norm_eps)
+
+        # Add per-dimension scaling parameter (same as TimesFmAttention)
+        self.use_per_dim_scale = getattr(config, "use_per_dim_scale", True)
+        if self.use_per_dim_scale:
+            self.scaling = nn.Parameter(torch.empty((self.head_dim,)))
```
vasqu: Same here, are there different models each time?
```suggestion
        # Add per-dimension scaling parameter (same as TimesFmAttention)
        self.use_per_dim_scale = config.use_per_dim_scale
        if self.use_per_dim_scale:
            self.scaling = nn.Parameter(torch.empty((self.head_dim,)))
```

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        cos, sin = position_embeddings
+        query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)
+
+        # Apply QK normalization (TimesFM 2.5 specific)
+        if self.use_qk_norm:
+            query_states = self.query_ln(query_states)
+            key_states = self.key_ln(key_states)
+
+        # Apply per-dimension scaling to query (TimesFM 2.5 specific)
+        query_states = self._scale_query(query_states)
```
vasqu:
```suggestion
        # Apply per-dimension scaling to query (TimesFM 2.5 specific)
        query_states = self._scale_query(query_states)
        scaling = self.scaling
        if self.use_per_dim_scale:
            scaling = F.softplus(self.scaling).mul(1.442695041 / math.sqrt(self.head_dim))[None, None, None, :]
```
Rather not move it to a separate function and specific scaling. My only concern is that sdpa/flash etc may not handle tensors as scaling then it won't work

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
```diff
@@ -0,0 +1,930 @@
...
+        )
+
+        attn_output, attn_weights = attention_interface(
+            self,
+            query_states,
+            key_states,
+            value_states,
+            attention_mask,
+            dropout=self.attention_dropout if self.training else 0.0,
+            scaling=1.0,
```
vasqu, line 295: So essentially we only scale when we have `_scale_query` --> this leads me to believe that it is always used

vasqu: I.e. we can directly multiply etc with no branches and extra function

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+            scaling=1.0,
+            **kwargs,
+        )
+
+        attn_output = attn_output.reshape(*input_shape, -1).contiguous()
+        attn_output = self.o_proj(attn_output)
+        return attn_output, attn_weights
+
+
+class Timesfm2P5DecoderLayer(nn.Module):
```
vasqu: Can be inherited from gemma2 https://github.com/huggingface/transformers/blob/9140fbee216c40fb18141e5a2ed24d5a96782dcc/src/transformers/models/gemma2/modeling_gemma2.py#L305

2 norms each time before and after

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        hidden_states = self.mlp(hidden_states)
+        hidden_states = self.post_ff_ln(hidden_states) + residual
+
+        return scores, hidden_states
+
+
+@auto_docstring
+class Timesfm2P5PreTrainedModel(TimesFmPreTrainedModel):
+    config_class = Timesfm2P5Config
+    base_model_prefix = "timesfm_2p5"
```
vasqu:
```suggestion
    base_model_prefix = "model"
```
if no strong opinion, would like to avoid these specific ones

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+    _supports_flash_attn = True
+    _supports_flex_attn = True
+
+
+@auto_docstring
+class Timesfm2P5Model(Timesfm2P5PreTrainedModel):
+    def __init__(self, config: Timesfm2P5Config):
+        super().__init__(config)
+        self.config = config
+        self._tolerance = 1e-6
```
vasqu:
```suggestion
        self.tolerance = 1e-6
```
any particular this is private?

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+
+        # Input projection with TimesFM 2.5 ResidualBlock
+        # Note: tokenizer uses bias=True (different from transformer layers)
+        self.input_ff_layer = Timesfm2P5ResidualBlock(
+            input_dims=2 * config.patch_length,  # 64 (32*2)
+            hidden_dims=config.hidden_size,  # 1280 (not intermediate_size)
+            output_dims=config.hidden_size,  # 1280
+            use_bias=True,  # tokenizer uses bias=True
```
vasqu: Would remove this and adjust on init instead tbh, it's not configurable either way

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+
+        # Transformer layers with TimesFM 2.5 specific components
+        self.layers = nn.ModuleList(
+            [Timesfm2P5DecoderLayer(config, layer_idx) for layer_idx in range(config.num_hidden_layers)]
+        )
+
+        # TimesFM 2.5 uses rotary embeddings - add rotary embedding component
+        self.rotary_emb = Timesfm2P5RotaryEmbedding(config)
+
+        # Initialize weights and apply final processing
```
vasqu: We can also enable gradient ckpting imo, `self.gradient_checkpointing = False`. Would need the gradient ckpting layer on the layers which should be doable especially after using gemma2 decoder layer

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        self.post_init()
+
+    def _revin(self, x: torch.Tensor, loc: torch.Tensor, scale: torch.Tensor, reverse: bool = False) -> torch.Tensor:
```
vasqu:
```suggestion
    def _revin(self, hidden_states: torch.Tensor, loc: torch.Tensor, scale: torch.Tensor, reverse: bool = False) -> torch.Tensor:
```
nit

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+
+        return (x - loc) / safe_scale
+
+    @staticmethod
+    def _update_running_stats(
+        n: torch.Tensor,
+        mu: torch.Tensor,
+        sigma: torch.Tensor,
+        x: torch.Tensor,
+        mask: torch.Tensor,
```
vasqu: This is a bit hard for me to parse but I'm not sure if this is simply me being unfamiliar with the methodology. Would it be possible to use non-single letter args?

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        is_valid = (~mask).to(x.dtype)
+        inc_n = is_valid.sum(dim=-1)
```
vasqu: Same here with abbreviations, ig it's "incrementing_n" e.g.

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
vasqu: Would be nice to just have some comment/docstring for any reference or similar

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+
+    @can_return_tuple
+    @auto_docstring
+    def forward(
+        self,
+        past_values: torch.Tensor,
+        past_values_padding: torch.LongTensor | None = None,
+        output_attentions: bool = False,
+        output_hidden_states: bool = False,
+        **kwargs: Unpack[FlashAttentionKwargs],
```
vasqu:
```suggestion
    @merge_with_config_defaults
    @capture_outputs
    @auto_docstring
    def forward(
        self,
        past_values: torch.Tensor,
        past_values_padding: torch.LongTensor | None = None,
        **kwargs: Unpack[TransformersKwargs],
```
Would need to properly set `_can_record_outputs`, see https://github.com/huggingface/transformers/blob/65935e09212bcd5c3ceecbd6e12ace3ec26a804f/src/transformers/models/llama/modeling_llama.py#L353-L356
but also simplifies the forward by taking care of all the unnecessary stuff re collecting attns, hidden states, etc

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+
+        if mu_history:
+            context_mu = torch.stack(mu_history, dim=1)
+            context_sigma = torch.stack(sigma_history, dim=1)
+        else:
+            context_mu = mu.unsqueeze(1)
+            context_sigma = sigma.unsqueeze(1)
+
+        normed_inputs = self._revin(patched_inputs, context_mu, context_sigma, reverse=False)
+        normed_inputs = torch.where(patched_masks_bool, torch.zeros_like(normed_inputs), normed_inputs)
```
vasqu: Any reason this is not directly within `revin`

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+
+        tokenizer_inputs = torch.cat(
+            [normed_inputs, patched_masks_bool.to(dtype=normed_inputs.dtype)],
+            dim=-1,
+        ).to(dtype=self.dtype)
```
vasqu: Not a fan of using self.dtype, is it really needed?

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        padding_mask = (~patch_padding).to(torch.int64)
+        cache_position = torch.arange(sequence_length, device=input_embeddings.device)
+        attention_mask = create_causal_mask(
+            self.config, input_embeddings, padding_mask, cache_position, past_key_values=None
+        )
+        position_embeddings = self.rotary_emb(input_embeddings, position_ids)
+
+        hidden_states = input_embeddings
+        all_hidden_states = () if output_hidden_states else None
+        all_attentions = () if output_attentions else None
```
vasqu: Not needed when decorator x `_can_record_outputs`

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+            context_mu=context_mu,
+            context_sigma=context_sigma,
+            patch_padding=patch_padding,
+        )
+
+
+@auto_docstring
+class Timesfm2P5ModelForPrediction(TimesFmModelForPrediction):
+    def __init__(self, config: Timesfm2P5Config):
+        # Call the parent's __init__ first to get the basic structure
```
vasqu:
```suggestion
```
a lot of these comments are a bit too much imo, could be fairly reduced

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
```diff
@@ -0,0 +1,930 @@
...
+
+
+@auto_docstring
+class Timesfm2P5ModelForPrediction(TimesFmModelForPrediction):
+    def __init__(self, config: Timesfm2P5Config):
+        # Call the parent's __init__ first to get the basic structure
+        super().__init__(config)
+
+        # Now override with TimesFM 2.5 specific components
+        self.config = config
```
vasqu, line 540:
```suggestion
```
modular should handle it no?

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        super().__init__(config)
+
+        # Now override with TimesFM 2.5 specific components
+        self.config = config
+        self.context_len = config.context_length
+        self.horizon_len = config.horizon_length
+
+        # Override decoder with TimesFM 2.5 model
+        self.decoder = Timesfm2P5Model(config)
```
vasqu: Same here?

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        # Keep the parent's horizon_ff_layer for compatibility but we'll use separate projections
+        # Note: parent's horizon_ff_layer will exist but we won't use it in our _postprocess_output
+
+        # Initialize weights and apply final processing
+        self.post_init()
+
+    def _preprocess(
+        self, inputs: Sequence[torch.Tensor], freq: Sequence[int], context_len: int | None = None
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        """Override parent's _preprocess to support custom context_len for forecasting.
```
vasqu: Would specify to timesfm specifically tbh, we can be concrete

vasqu: Also, would it make sense to override the parent to support this instead? seems small enough

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        past_values: Sequence[torch.Tensor],
+        window_size: int | None = None,
+        future_values: torch.Tensor | None = None,
+        forecast_context_len: int | None = None,
+        return_forecast_on_context: bool = False,
+        truncate_negative: bool | None = None,
+        force_flip_invariance: bool | None = None,
+        output_attentions: bool | None = None,
+        output_hidden_states: bool | None = None,
+        **kwargs,
```
vasqu:
```suggestion
        **kwargs: [TransformersKwargs],
```

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        return_forecast_on_context (`bool`, *optional*):
+            Whether to return forecasts on context positions when available.
+        truncate_negative (`bool`, *optional*):
+            Whether to clamp outputs to non-negative values. If `None`, defaults to `config.infer_is_positive`.
+        force_flip_invariance (`bool`, *optional*):
+            Whether to apply the flip-invariance combination. If `None`, defaults to
+            `config.force_flip_invariance`.
+        """
+        if forecast_context_len is None:
+            fcontext_len = self.context_len
```
vasqu:
```suggestion
            forecast_context_len = self.context_len
```
would not shorten and directly use instead

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        else:
+            fcontext_len = forecast_context_len
+
+        device = past_values[0].device
+
+        if return_forecast_on_context:
+            raise NotImplementedError("`return_forecast_on_context` is not supported for TimesFM 2.5 conversion yet.")
```
vasqu:
```suggestion
            raise NotImplementedError("`return_forecast_on_context` is not supported for TimesFM 2.5 yet.")
```

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+
+        inputs = [ts[-fcontext_len:] for ts in past_values]
+        inp_min = torch.min(torch.stack([torch.min(ts) for ts in inputs]))
```
vasqu:
```suggestion
        input_min = torch.min(torch.stack([torch.min(time_series) for time_series in inputs]))
```
similar here, let's be more direct

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        freq = [0] * len(inputs)
+
+        if output_attentions is None:
+            output_attentions = self.config.output_attentions
+        if output_hidden_states is None:
+            output_hidden_states = self.config.output_hidden_states
+        if truncate_negative is None:
+            truncate_negative = getattr(self.config, "infer_is_positive", False)
+        if force_flip_invariance is None:
+            force_flip_invariance = getattr(self.config, "force_flip_invariance", True)
```
vasqu: shouldnt need gettattr, should be directly in the config

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        mu_global = input_ts.mean(dim=1, keepdim=True)
+        sigma_global = input_ts.std(dim=1, keepdim=True)
+        sigma_safe = torch.where(
+            sigma_global < self.decoder._tolerance,
+            torch.ones_like(sigma_global),
+            sigma_global,
+        )
+
+        normalized_ts = (input_ts - mu_global) / sigma_safe
+        normalized_ts = normalized_ts.to(self.decoder.dtype)
```
vasqu: Can be handled with the revin method no?

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+        quantile_output = self.decoder._revin(quantile_output_norm, context_mu, context_sigma, reverse=True)
+
+        # Reshape point output: [batch, num_patches, horizon_len * num_quantiles] -> [batch, num_patches, horizon_len, num_quantiles]
+        batch_size, num_patches, _ = point_output.shape
+        num_quantiles = len(self.config.quantiles) + 1
+        point_output = point_output.view(batch_size, num_patches, self.config.horizon_length, num_quantiles)
+        pf_outputs = point_output[:, -1, :, :]  # Take last patch: [batch, horizon_len, num_quantiles]
+
+        # Reshape quantile output: [batch, num_patches, output_quantile_len * num_quantiles] -> [batch, num_patches, output_quantile_len, num_quantiles]
+        output_quantile_len = getattr(self.config, "output_quantile_len", 1024)
```
vasqu:
```suggestion
        output_quantile_len = self.config.output_quantile_len
```

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
```diff
@@ -0,0 +1,930 @@
...
+        num_quantiles = len(self.config.quantiles) + 1
+        point_output = point_output.view(batch_size, num_patches, self.config.horizon_length, num_quantiles)
+        pf_outputs = point_output[:, -1, :, :]  # Take last patch: [batch, horizon_len, num_quantiles]
+
+        # Reshape quantile output: [batch, num_patches, output_quantile_len * num_quantiles] -> [batch, num_patches, output_quantile_len, num_quantiles]
+        output_quantile_len = getattr(self.config, "output_quantile_len", 1024)
+        quantile_output = quantile_output.view(batch_size, num_patches, output_quantile_len, num_quantiles)
+        quantile_spreads = quantile_output[:, -1, :, :]  # Take last patch: [batch, output_quantile_len, num_quantiles]
+
+        if force_flip_invariance:
```
vasqu, line 664: This whole block seems like a repetition of the above expect for the beginning input with the minus and the flip at the end. Imo makes sense to create a method to cover this whole block so we can have a more simplified forward

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,930 @@
...
+            last_hidden_state=model_outputs.last_hidden_state,
+            attentions=model_outputs.attentions if output_attentions else None,
+            hidden_states=model_outputs.hidden_states if output_hidden_states else None,
+            mean_predictions=mean_predictions,
+            full_predictions=full_predictions,
+            loss=loss,
+        )
+
+    @staticmethod
+    def _timesfm_moving_average(arr: torch.Tensor, window_size: int) -> list[torch.Tensor]:
```
vasqu: Modular should inherit, no?

tests/models/timesfm_2p5/test_modeling_timesfm_2p5.py
```diff
@@ -0,0 +1,338 @@
+# Copyright 2025 the HuggingFace Team. All rights reserved.
```
vasqu:
```suggestion
# Copyright 2026 the HuggingFace Team. All rights reserved.
```

tests/models/timesfm_2p5/test_modeling_timesfm_2p5.py
```diff
@@ -0,0 +1,338 @@
...
+        return config, inputs_dict
+
+
+@require_torch
+class Timesfm2P5ModelTest(ModelTesterMixin, unittest.TestCase):
+    all_model_classes = (Timesfm2P5ModelForPrediction,) if is_torch_available() else ()
+    all_generative_model_classes = ()
+    all_parallelizable_model_classes = ()
+    fx_compatible = False
+    test_pruning = False
```
vasqu: Looks outdated / not needed to me, a lot of flags got deprecated

tests/models/timesfm_2p5/test_modeling_timesfm_2p5.py
```diff
@@ -0,0 +1,338 @@
...
+    test_resize_embeddings = False
+    test_model_parallel = False
```
vasqu: Same here but unsure

vasqu (commented):
Some first initial comments, I went a bit longer but I think we can benefit a lot more from modular and simplify

vasqu (commented):
~Some first initial comments, I went a bit longer but I think we can benefit a lot more from modular and simplify~ Github was messing with me, ignore this :D

kashif:
thanks! having a look and fixing

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
kashif: about `context_mu`/`context_sigma`, they are needed — `ForPrediction._decode_and_project` to denormalize

kashif:
@vasqu thanks so apart from renaming, I beleive i have addressed the rest of the issues

docs/source/en/model_doc/timesfm2_5.md
kashif: so we wanted seperate metrics for the transformers part vs their own timesfm library downloads, so we thought it would be easier to have an own model for transformers

vasqu: gotcha, makes sense 👍

docs/source/en/model_doc/timesfm_2p5.md
```diff
@@ -0,0 +1,80 @@
...
+```python
+import numpy as np
+import torch
+from transformers import Timesfm2P5ModelForPrediction
+
+
+model = Timesfm2P5ModelForPrediction.from_pretrained(
+    "google/timesfm-2.5-200m-transformers",
+    attn_implementation="sdpa",
+    dtype=torch.float32,
```
vasqu:
```suggestion
```
these are defaults already so shouldn't need it (I guess the model is stored in fp32)

src/transformers/models/auto/configuration_auto.py
vasqu: Ah this is what you meant with the renaming?

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
```diff
@@ -0,0 +1,803 @@
...
+        context_length: int = 16384,
+        horizon_length: int = 128,
+        quantiles: list = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9],
+        hidden_size: int = 1280,
+        intermediate_size: int = 1280,
+        head_dim: int = 80,
+        num_attention_heads: int = 16,
+        num_key_value_heads: int = 16,
+        num_hidden_layers: int = 20,
+        rms_norm_eps: float = 1e-6,
```
vasqu, line 136:
```suggestion
        rms_norm_eps: float = 1e-5,
```
just wanna make sure that this is correct and intended 👀

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+        decode_index: int = 5,
+        use_qk_norm: bool = True,
+        use_per_dim_scale: bool = True,
+        use_bias: bool = False,
+        activation: str = "swish",
+        use_continuous_quantile_head: bool = True,
+        force_flip_invariance: bool = True,
+        infer_is_positive: bool = True,
+        max_position_embeddings: int = 16384,
+        rope_theta: float = 10000.0,
```
vasqu:
```suggestion
    rope_parameters: RopeParameters | dict[str, RopeParameters] | None = None,
```
we changed rope a bit and have a mixin https://github.com/huggingface/transformers/blob/94e51ee367bae004230270f6612f2b13dfb1d11f/src/transformers/modeling_rope_utils.py#L625

I think you use default rope and also the default theta https://github.com/huggingface/transformers/blob/94e51ee367bae004230270f6612f2b13dfb1d11f/src/transformers/modeling_rope_utils.py#L630 so you will only need to make sure that the super call comes at last (wihin the configuration file)

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+        **kwargs,
+    ):
+        # Set before super().__init__() so PreTrainedConfig detects it and populates rope_parameters
+        self.rope_parameters = None
```
vasqu: Imo we can also change the parent config if that makes it easier; the position of the super call shouldnt matter anymore and it's generally better to be at the end

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+            num_attention_heads=num_attention_heads,
+            num_key_value_heads=num_key_value_heads,
+            rms_norm_eps=rms_norm_eps,
+            attention_dropout=attention_dropout,
+            attention_bias=attention_bias,
+            initializer_range=initializer_range,
+            num_hidden_layers=num_hidden_layers,
+            use_positional_embedding=False,
+            layer_types=["attention"] * num_hidden_layers,
+            sliding_window=None,
```
vasqu: I'd rather we changeup on the module's itself than to populate unused attributes, i.e. we always use full attention and no sliding windows so we shouldn't need these

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+        self.use_bias = use_bias
+        self.activation = activation
+        self.use_continuous_quantile_head = use_continuous_quantile_head
+        self.force_flip_invariance = force_flip_invariance
+        self.infer_is_positive = infer_is_positive
+        # Gemma2-compatible: needed by Gemma2Attention.__init__
+        self.query_pre_attn_scalar = 256.0
+        self.attn_logit_softcapping = None
+        self.max_position_embeddings = max_position_embeddings
+        self.rope_theta = rope_theta
```
vasqu: Same here I'd rather we find a different module to copy from than to forcefully add attributes just to make it work
```suggestion
        self.query_pre_attn_scalar = 256.0
        self.max_position_embeddings = max_position_embeddings
```

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
```diff
@@ -0,0 +1,803 @@
...
+@auto_docstring
+class Timesfm2P5OutputForPrediction(TimesFmOutputForPrediction):
+    r"""
+    mean_predictions (`torch.Tensor` of shape `(batch_size, horizon_length)`):
+        Deterministic forecasts after denormalization.
+    full_predictions (`torch.Tensor` of shape `(batch_size, horizon_length, quantiles)`):
+        Quantile forecasts including the median after denormalization.
+    loss (`torch.Tensor` of shape `(1,)`, *optional*, returned when `future_values` is provided):
+        Training loss combining MSE and quantile losses when targets are supplied.
+    """
```
vasqu, line 216: I think autodocstring picks these up, not 100% sure

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
vasqu: Bumping this, just needs renames but it is essentially the same

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+        hidden_states = self.ff0(hidden_states)
+        hidden_states = self.activation(hidden_states)
+        return self.ff1(hidden_states)
+
+
+class Timesfm2P5ResidualBlock(nn.Module):
+    """[`TimesFmResidualBlock`] variant with configurable `use_bias` and `activation`."""
+
+    def __init__(
+        self, input_dims: int, hidden_dims: int, output_dims: int, use_bias: bool = True, activation: str = "swish"
```
vasqu: Would it be possible to have these as config attributes instead? I.e. use bias and activation

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+        self.ff1 = nn.Linear(config.intermediate_size, config.hidden_size, bias=config.use_bias)
+        self.activation = ACT2FN[config.activation]
+
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        hidden_states = self.ff0(hidden_states)
+        hidden_states = self.activation(hidden_states)
+        return self.ff1(hidden_states)
+
+
+class Timesfm2P5ResidualBlock(nn.Module):
```
vasqu: We can inherit from timesfm no? The forward is identical

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+
+class Timesfm2P5Attention(Gemma2Attention):
+    """TimesFM 2.5 attention with QK normalization and learnable per-dimension query scaling."""
+
+    def __init__(self, config: Timesfm2P5Config, layer_idx: int):
+        super().__init__(config, layer_idx)
+
+        if config.use_qk_norm:
+            self.q_norm = Timesfm2P5RMSNorm(self.head_dim, eps=config.rms_norm_eps)
+            self.k_norm = Timesfm2P5RMSNorm(self.head_dim, eps=config.rms_norm_eps)
```
vasqu: Are there variations with qk norm or not? We could simply remove it from the config

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+
+        if config.use_per_dim_scale:
```
vasqu: Same here, if this is always used then we can remove from config and always use

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
```diff
@@ -0,0 +1,803 @@
...
+        )
+
+        attn_output, attn_weights = attention_interface(
+            self,
+            query_states,
+            key_states,
+            value_states,
+            attention_mask,
+            dropout=self.attention_dropout if self.training else 0.0,
+            scaling=1.0,
```
vasqu, line 295: Let's add a comment because this is a tad special from other models --> learnable scaling instead of static vars

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+        attn_output = self.o_proj(attn_output)
+        return attn_output, attn_weights
+
+
+class Timesfm2P5DecoderLayer(Gemma2DecoderLayer):
+    """TimesFM 2.5 Transformer decoder layer with pre/post RMS normalization.
+
+    Inherits from Gemma2DecoderLayer (which provides GradientCheckpointingLayer support) and
+    overrides forward to return `(hidden_states, attn_weights)` tuple for output capturing.
+    """
```
vasqu: Isnt the key difference that we don't have a cache and that's it? Maybe we can mention that solely. The second paragraph doesnt seem really useful

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+
+        new_var = (term1 + term2 + term3 + term4) / new_count_safe
+        new_var = torch.where(new_count == 0, torch.zeros_like(new_var), new_var)
+        new_std = torch.sqrt(torch.clamp(new_var, min=0.0))
+
+        return new_count, new_mean, new_std
+
+    @merge_with_config_defaults
+    @capture_outputs
+    @can_return_tuple
```
vasqu:
```suggestion
```
iirc the decorators above should already handle it

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+    _no_split_modules = ["Timesfm2P5DecoderLayer"]
+    _supports_flash_attn = True
+    _supports_flex_attn = True
+    _can_record_outputs = {
+        "hidden_states": Timesfm2P5DecoderLayer,
+        "attentions": Timesfm2P5Attention,
+    }
+
+
+@auto_docstring
```
vasqu:
```suggestion
```
not needed when we have it on the forward already

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+            **kwargs,
+        )
+        hidden_states = self.post_attention_layernorm(hidden_states) + residual
+
+        residual = hidden_states
+        hidden_states = self.pre_feedforward_layernorm(hidden_states)
+        hidden_states = self.mlp(hidden_states)
+        hidden_states = self.post_feedforward_layernorm(hidden_states) + residual
+
+        return hidden_states, attn_weights
```
vasqu:
```suggestion
        return hidden_states
```
shouldnt be needed the capture hooks already handle it

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+        self,
+        hidden_states: torch.Tensor,
+        position_embeddings: tuple[torch.Tensor, torch.Tensor] | None = None,
+        attention_mask: torch.Tensor | None = None,
+        position_ids: torch.Tensor | None = None,
+        **kwargs: Unpack[TransformersKwargs],
+    ) -> tuple[torch.Tensor, torch.Tensor | None]:
+        residual = hidden_states
+        hidden_states = self.input_layernorm(hidden_states)
+        hidden_states, attn_weights = self.self_attn(
```
vasqu:
```suggestion
        hidden_states, _ = self.self_attn(
```

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+        cache_position = torch.arange(sequence_length, device=input_embeddings.device)
+        attention_mask = create_causal_mask(
+            self.config, input_embeddings, padding_mask, cache_position, past_key_values=None
+        )
+        position_embeddings = self.rotary_emb(input_embeddings, position_ids)
+
+        hidden_states = input_embeddings
+
+        for layer in self.layers:
+            hidden_states, _ = layer(
```
vasqu:
```suggestion
            hidden_states = layer(
```
see my previous comments, no need to return the weights up till then

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
vasqu: bumping

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+            input_dims=config.hidden_size,
+            hidden_dims=config.hidden_size,
+            output_dims=config.output_quantile_len * num_quantiles,
+            use_bias=config.use_bias,
+            activation=config.activation,
+        )
+
+        self.post_init()
+
+    def _preprocess(
```
vasqu: Rechecking if we could instead change it a bit in the parent model and inherit it from there; I don't see much diff except for the context_len being an extra kwarg

vasqu: Or at least highlight the key difference with a small comment :D

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+            input_ts.append(ts)
+            input_padding.append(padding)
+
+        return torch.stack(input_ts, dim=0), torch.stack(input_padding, dim=0)
+
+    def _decode_and_project(
+        self,
+        normalized_ts: torch.Tensor,
+        input_padding: torch.Tensor,
+        **kwargs,
```
vasqu:
```suggestion
        **kwargs: Unpack[TransformersKwargs],
```

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+            Whether to apply the flip-invariance combination. If `None`, defaults to
+            `config.force_flip_invariance`.
+        """
+        if forecast_context_len is None:
+            forecast_context_len = self.context_len
+
+        device = past_values[0].device
+
+        if return_forecast_on_context:
+            raise NotImplementedError("`return_forecast_on_context` is not supported for TimesFM 2.5 yet.")
```
vasqu: Just to be sure but it is planned?

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
```diff
@@ -0,0 +1,803 @@
...
+            new_inputs: list[torch.Tensor] = []
+            for ts in inputs:
+                new_inputs.extend(self._timesfm_moving_average(ts, window_size))
+            inputs = new_inputs
+
+        if truncate_negative is None:
+            truncate_negative = self.config.infer_is_positive
+        if force_flip_invariance is None:
+            force_flip_invariance = self.config.force_flip_invariance
+
```
vasqu, line 652: Might be good candidates for the `merge_with_config_defaults` decorator can you check?

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,803 @@
...
+        return_forecast_on_context (`bool`, *optional*):
+            Whether to return forecasts on context positions when available.
+        truncate_negative (`bool`, *optional*):
+            Whether to clamp outputs to non-negative values. If `None`, defaults to `config.infer_is_positive`.
+        force_flip_invariance (`bool`, *optional*):
+            Whether to apply the flip-invariance combination. If `None`, defaults to
+            `config.force_flip_invariance`.
+        """
+        if forecast_context_len is None:
+            forecast_context_len = self.context_len
```
vasqu: Do we really need this? If we just straight pass it, we fallback either way, no?

tests/models/timesfm_2p5/test_modeling_timesfm_2p5.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def test_create_and_run_model(self):
+        config, inputs_dict = self.model_tester.prepare_config_and_inputs_for_common()
+        model = Timesfm2P5ModelForPrediction(config)
+        model.to(torch_device)
+        model.eval()
+        results = model(**inputs_dict)
+        assert results.mean_predictions is not None
+
+    @unittest.skip(reason="Compile not yet supported because of masks")
```
vasqu:
```suggestion
    @unittest.skip(reason="FA backend not yet supported because of forced masks")
```
nit

tests/models/timesfm_2p5/test_modeling_timesfm_2p5.py
```diff
@@ -0,0 +1,320 @@
...
+        pass
+
+    @unittest.skip(reason="Model does not have input embeddings")
+    def test_model_get_set_embeddings(self):
+        pass
+
+    @unittest.skip(reason="Model does not have head mask")
+    def test_headmasking(self):
+        pass
+
```
vasqu: Shouldnt be a test anymore

tests/models/timesfm_2p5/test_modeling_timesfm_2p5.py
```diff
@@ -0,0 +1,320 @@
...
+    # the main input name is `inputs`
+    def test_model_main_input_name(self):
```
vasqu: Dont see why we overwrote that one?

vasqu (commented):
A lot of nitpicking but the core thing I noticed is that we use attribtues just to inherit from something, that shouldn't be the case: They are unused in a sense and should not be part of the model if they themselves have not used these attributes in different models.

I think there are 1-2 legacy things in the tests we can change/remove 👀

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
kashif: yes their official code had 1e-6

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
kashif: yes for some we can use config for sure

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
kashif: the current pretrained weights are with it set to true

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
kashif: `use_per_dim_scale` is also always true

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
kashif: removed

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
kashif: simplified

tests/models/timesfm_2p5/test_modeling_timesfm_2p5.py
kashif: removed

tests/models/timesfm_2p5/test_modeling_timesfm_2p5.py
kashif: removed as well

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
kashif: doesnt seem like it will work as it has a hard-coded list of `args_with_config_defaults`

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
kashif: the CI was complaining without this... i also thought it would pick it up...

docs/source/en/model_doc/timesfm2_5.md
vasqu, line 1: I guess now is too late for the renaming with `timesfm_2p5 <-> timesfm2_5` - more of a nit but follows our conventions a bit more, e.g. qwen

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,739 @@
...
+    TimesFmOutputForPrediction,
+    TimesFmPreTrainedModel,
+    TimesFmResidualBlock,
+)
+
+
+logger = logging.get_logger(__name__)
+
+
+class Timesfm2P5Config(TimesFmConfig):
```
vasqu:
```suggestion
class Timesfm_2P5Config(TimesFmConfig):
```
Nit wouldnt it be this to follow the current convention?

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,739 @@
...
+        activation: str = "swish",
+        use_continuous_quantile_head: bool = True,
+        force_flip_invariance: bool = True,
+        infer_is_positive: bool = True,
+        max_position_embeddings: int = 16384,
+        rope_parameters: RopeParameters | dict[str, RopeParameters] | None = None,
+        **kwargs,
+    ):
+        # Set all attributes before super().__init__() so RotaryEmbeddingConfigMixin
+        # can find rope_parameters and max_position_embeddings during initialization.
```
vasqu:
```suggestion
```
good as interal header but not really needed for the codebase

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,739 @@
...
+    full_predictions (`torch.Tensor` of shape `(batch_size, horizon_length, quantiles)`):
+        Quantile forecasts including the median after denormalization.
+    loss (`torch.Tensor` of shape `(1,)`, *optional*, returned when `future_values` is provided):
+        Training loss combining MSE and quantile losses when targets are supplied.
+    """
+
+    pass
+
+
+class Timesfm2P5MLP(nn.Module):
```
vasqu: Reopening, can be inherited from Clip, just needs to renames
- `ff1<->fc1`
- `ff2<->fc2`
- `activation<->activation_fn`

vasqu: Which we can do with the weight convert if that's more convenient than the manual conversion + on the hub update

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,739 @@
...
+
+class Timesfm2P5RMSNorm(LlamaRMSNorm):
+    pass
+
+
+class Timesfm2P5RotaryEmbedding(LlamaRotaryEmbedding):
+    pass
+
+
+class Timesfm2P5Attention(LlamaAttention):
```
vasqu: Super nit but we can even use another one to save on the norms, see https://github.com/huggingface/transformers/blob/af5dfb6c5cb98091aa76c2c768bd4b7a38891b01/src/transformers/models/apertus/modeling_apertus.py#L216

src/transformers/models/timesfm2_5/modular_timesfm2_5.py
```diff
@@ -0,0 +1,739 @@
...
+        )
+
+        attn_output, attn_weights = attention_interface(
+            self,
+            query_states,
+            key_states,
+            value_states,
+            attention_mask,
+            dropout=self.attention_dropout if self.training else 0.0,
+            # scaling=1.0 because per-dimension learnable scaling is already applied to query_states above
```
vasqu, line 294: Nice!

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
```diff
@@ -0,0 +1,739 @@
...
+            scaling=1.0,
+            **kwargs,
+        )
+
+        attn_output = attn_output.reshape(*input_shape, -1).contiguous()
+        attn_output = self.o_proj(attn_output)
+        return attn_output, attn_weights
+
+
+class Timesfm2P5DecoderLayer(LlamaDecoderLayer):
```
vasqu: I think we can still use the gemma2 one, no? It would at least make it easier on the forward path as it's the same then, the init would need to delete a few attributes, e.g. `del self.attention_type`

tests/models/timesfm/test_modeling_timesfm.py
```diff
@@ -144,6 +145,59 @@ def test_create_and_run_model(self):
     def test_sdpa_can_dispatch_on_flash(self):
         pass
 
+    @parameterized.expand(TEST_EAGER_MATCHES_SDPA_INFERENCE_PARAMETERIZATION)
```
vasqu, line 148: They were already failing I guess? Just for my interest, not super relevant to this PR

tests/models/timesfm_2p5/test_modeling_timesfm_2p5.py
```diff
@@ -0,0 +1,308 @@
...
+from transformers.testing_utils import require_flash_attn, require_torch, require_torch_accelerator, slow, torch_device
+
+from ...test_configuration_common import ConfigTester
+from ...test_modeling_common import TEST_EAGER_MATCHES_SDPA_INFERENCE_PARAMETERIZATION, ModelTesterMixin
+
+
+if is_torch_available():
+    from transformers import Timesfm2P5ModelForPrediction
+
+TOLERANCE = 1e-4
```
vasqu:
```suggestion
```
A bit useless imo, we can just do it directly in that one test

vasqu (approved):
Nice, only nits left to improve things a tad more otherwise looks super solid!

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
kashif: CLIP doesn't support bias which what i need

vasqu: We can use it for the forward and overwrite the init, just goes to show the connection a little bit better

docs/source/en/model_doc/timesfm2_5.md
kashif: renamed

tests/models/timesfm/test_modeling_timesfm.py
kashif: yes was failing so i fixed it

src/transformers/models/timesfm_2p5/modular_timesfm_2p5.py
kashif: config would then need the `layer_types` too...

vasqu:
run-slow: timesfm2_5

vasqu:
run-slow: timesfm2_5

kashif:
@rajatsen91 can you kindly merge my hub PR due to changes in the config and renamings etc. https://huggingface.co/google/timesfm-2.5-200m-transformers/discussions/3

vasqu:
run-slow: timesfm2_5

PR #41764, opened 2025-10-21
Fix logic error in `prepare_inputs_for_generation` cache slicing condition

### LINKED ISSUE #4272: CI fails with dev dependencies: TypeError: 'NoneType' object is not subscriptable
CI fails with dev dependencies: https://github.com/huggingface/trl/actions/runs/18493152127/job/52691262212
> TypeError: 'NoneType' object is not subscriptable
```python
FAILED tests/test_modeling_geometric_mixture_wrapper.py::TestGeometricMixtureWrapper::test_prepare_inputs_for_generation - TypeError: 'NoneType' object is not subscriptable
```

Stacktrace:
```python
>       inputs = self.wrapper.prepare_inputs_for_generation(input_ids, attention_mask=attention_mask, use_cache=True)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/test_modeling_geometric_mixture_wrapper.py:65: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
trl/models/modeling_base.py:717: in prepare_inputs_for_generation
    model_inputs = self.model.prepare_inputs_for_generation(*args, **kwargs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/transformers/generation/utils.py:613: in prepare_inputs_for_generation
    inputs_embeds, input_ids = self._cache_dependant_input_preparation(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Qwen2ForCausalLM(
  (model): Qwen2Model(
    (embed_tokens): Embedding(151665, 8)
    (layers): ModuleList(
      (0-1...-06)
    (rotary_emb): Qwen2RotaryEmbedding()
  )
  (lm_head): Linear(in_features=8, out_features=151665, bias=False)
)
input_ids = tensor([[1, 2, 3, 4, 5]], device='cuda:0'), inputs_embeds = None
cache_position = None

    def _cache_dependant_input_preparation(
        self,
        input_ids: torch.LongTensor,
        inputs_embeds: Optional[torch.FloatTensor],
        cache_position: Optional[torch.LongTensor],
    ) -> tuple[torch.FloatTensor, torch.LongTensor]:
        """
        Generic cache-dependent input preparation
        The code is put in a separate function to allow granular unit testing
        as it needs a different implementation to be exportable.
    
        If we have cache: let's slice `input_ids` through `cache_position`, to keep only the unprocessed tokens
        - Exception 1: when passing input_embeds, input_ids may be missing entries
        - Exception 2: some generation methods do special slicing of input_ids, so we don't need to do it here
        - Exception 3: with synced GPUs cache_position may go out of bounds, but we only want dummy token in that case.
        - Exception 4: If input_embeds are passed then slice it through `cache_position`, to keep only the unprocessed tokens and
          generate the first token for each sequence. Later use the generated Input ids for continuation.
    
        The current implementation does not rely on ``self`` and could be
        a class method. It is left as a standard method to be easily rewritten.
        """
        if is_torchdynamo_exporting():
            return self._cache_dependant_input_preparation_exporting(input_ids, inputs_embeds, cache_position)
        if inputs_embeds is not None and input_ids.shape[1] == 0:  # Exception 4
            inputs_embeds = inputs_embeds[:, -cache_position.shape[0] :]
        elif (
            inputs_embeds is not None  # Exception 1
>           or (cache_position[-1] >= input_ids.shape[1])  # Exception 3
                ^^^^^^^^^^^^^^^^^^
        ):
E       TypeError: 'NoneType' object is not subscriptable

.venv/lib/python3.12/site-packages/transformers/generation/utils.py:509: TypeError
```

albertvillanova:
Fix logic error in `prepare_inputs_for_generation` cache slicing condition:
> TypeError: 'NoneType' object is not subscriptable

### Background

I think the PR
- #41505

introduced a logic error where the condition for calling `_cache_dependant_input_preparation` uses `is None` instead of `is not None`, causing crashes when `prepare_inputs_for_generation` is called with `past_key_values=None` and `use_cache=False`.

### Bug

PR #41505 introduced this condition:
```python
if past_key_values is None or use_cache:
    inputs_embeds, input_ids = self._cache_dependant_input_preparation(...)
```

The condition `past_key_values is None or use_cache` means:
- Do cache-dependent slicing if we DON'T have a cache OR if caching is enabled

This triggers the function even when:
- past_key_values=None (no cache)
- use_cache=False (caching disabled)
- cache_position=None (no position information)

This combination is invalid for cache-dependent preparation and causes a crash when accessing `cache_position[-1]` (line 456).

Note that during normal generation, it works fine because `use_cache=True`, making the buggy `past_key_values is None` part irrelevant.

### Fix

This PR changes the condition to:
```diff
- if past_key_values is None or use_cache:
+ if past_key_values is not None or use_cache:
```

The condition `past_key_values is not None or use_cache` means:
- Do cache-dependent slicing if we DO have a cache OR if caching is enabled

This is semantically correct and matches the intent described in the PR #41505 comment: https://github.com/huggingface/transformers/pull/41505#discussion_r2426080715
 > stateful models like `recurrent_gemma` assume that slicing happens, but don't have a Cache

The `use_cache` part handles stateful models, while `past_key_values is not None` handles normal cached models.

### Testing

This PR fixes the downstream failing test in TRL:
> tests/test_modeling_geometric_mixture_wrapper.py::TestGeometricMixtureWrapper::test_prepare_inputs_for_generation

See the associated issue:
- https://github.com/huggingface/trl/issues/4272

### Related

- This PR addresses a logic error introduced by:
  - #41505

- This PR will fix https://github.com/huggingface/trl/issues/4272

CC:
- @gante, who made the PR #41505
- @zucchini-nlp , who reviewed the PR #41505; see https://github.com/huggingface/transformers/pull/41505#discussion_r2425632292
  > maybe i am missing smth, do we apply cache slicing when the `past_key_values is None`?


zucchini-nlp:
I had same question, so I think it's better to ping @gante who has more understanding of the situation. IIRC it had smth to do with Reformer model

gante:
Hi @albertvillanova @zucchini-nlp  👋 

Have a look at this slack thread, I think it contains the full discussion related to this bug (assuming it is the same issue, I think it is) and how it interacts with TRL: https://huggingface.slack.com/archives/C01N44FJDHT/p1760725474636279

TL;DR I think there is a fix we can implement in TRL right now (see thread) to work around it, the transformers-side fix is trickier

zucchini-nlp (approved):
Actually, I just looked at the special RecurrentGemma model which has no `past_key_values`, and it should not have an edge case if we check values of `use_cache`. For other special models like Mamba, they have their own input preparation method

So I don't think we needed to check `past_key_values is None`, unless the edge case mentioned by Joao isn't tested. So let's merge it

zucchini-nlp:
I will just rebase main to check no new test failures :)

PR #41765, opened 2025-10-21
[kernels] Add Tests & CI for kernels

MekkCyber:
# What does this PR do?

Adds tests for kernels, and proper daily CI, and slack notifications

run example : https://github.com/huggingface/transformers/actions/runs/18688016017/job/53285883834

### DIFF
--- a/.github/workflows/self-scheduled.yml
+++ b/.github/workflows/self-scheduled.yml
@@ -463,6 +463,69 @@ jobs:
           name: ${{ env.machine_type }}_run_quantization_torch_gpu_${{ env.matrix_folders }}_test_reports
           path: /transformers/reports/${{ env.machine_type }}_run_quantization_torch_gpu_${{ matrix.folders }}_test_reports
 
+  run_kernels_gpu:
+    if: ${{ inputs.job == 'run_kernels_gpu' }}
+    name: Kernel tests
+    strategy:
+      fail-fast: false
+      matrix:
+        machine_type: [aws-g5-4xlarge-cache]
+    runs-on:
+      group: '${{ matrix.machine_type }}'
+    container:
+      image: ${{ inputs.docker }}
+      options: --gpus all --shm-size "16gb" --ipc host -v /mnt/cache/.cache/huggingface:/mnt/cache/
+    steps:
+      - name: Update clone
+        working-directory: /transformers
+        run: git fetch && git checkout ${{ inputs.commit_sha || github.sha }}
+
+      - name: Reinstall transformers in edit mode
+        working-directory: /transformers
+        run: python3 -m pip uninstall -y transformers && python3 -m pip install -e .[testing]
+
+      - name: NVIDIA-SMI
+        run: nvidia-smi
+
+      - name: Environment
+        working-directory: /transformers
+        run: python3 utils/print_env.py
+
+      - name: Show installed libraries and their versions
+        working-directory: /transformers
+        run: pip freeze
+
+      - name: Set `machine_type` for report and artifact names
+        working-directory: /transformers
+        shell: bash
+        run: |
+          if [ "${{ matrix.machine_type }}" = "aws-g5-4xlarge-cache" ]; then
+            machine_type=single-gpu
+          else
+            machine_type=${{ matrix.machine_type }}
+          fi
+          echo "machine_type=$machine_type" >> $GITHUB_ENV
+
+      - name: Install kernels
+        working-directory: /transformers
+        run: python3 -m pip install -U kernels
+
+      - name: Run kernel tests on GPU
+        working-directory: /transformers
+        run: |
+          python3 -m pytest -v --make-reports=${{ env.machine_type }}_run_kernels_gpu_test_reports tests/kernels/test_kernels.py
+
+      - name: Failure short reports
+        if: ${{ failure() }}
+        continue-on-error: true
+        run: cat /transformers/reports/${{ env.machine_type }}_run_kernels_gpu_test_reports/failures_short.txt
+
+      - name: "Test suite reports artifacts: ${{ env.machine_type }}_run_kernels_gpu_test_reports"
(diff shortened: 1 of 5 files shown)


.github/workflows/self-scheduled.yml
```diff
@@ -463,6 +463,70 @@ jobs:
...
+          if [ "${{ matrix.machine_type }}" = "aws-g5-4xlarge-cache" ]; then
+            machine_type=single-gpu
+          else
+            machine_type=${{ matrix.machine_type }}
+          fi
+          echo "machine_type=$machine_type" >> $GITHUB_ENV
+
+      - name: Install kernels
+        working-directory: /transformers
+        run: python3 -m pip install -U kernels
```
ydshieh: we should move this before the step 

> - name: Show installed libraries and their versions

.github/workflows/self-scheduled.yml
```diff
@@ -463,6 +463,70 @@ jobs:
...
+
+      - name: Run kernel tests on GPU
+        working-directory: /transformers
+        run: |
+          python3 -m pytest -v --make-reports=${{ env.machine_type }}_run_kernels_gpu_test_reports tests/kernels/test_kernels.py
```
ydshieh, line 516: we don't need any flag and env. variable to enable kernels? 

Just simply install it would enable?

tests/kernels/test_kernels.py
```diff
@@ -0,0 +1,400 @@
...
+            ]
+            for k in keys_to_remove:
+                _KERNEL_MODULE_MAPPING.pop(k, None)
+        except Exception:
+            pass
+
+        # Free accelerator memory/cache and trigger GC
+        gc.collect()
+        backend_empty_cache(torch_device)
+        gc.collect()
```
ydshieh: We could simply use the `cleanup` 

> from transformers.testing_utils

like

> cleanup(torch_device, gc_collect=True)

tests/kernels/test_kernels.py
```diff
@@ -0,0 +1,400 @@
...
+@require_kernels
+class TestHubKernels(TestCasePlus):
+    def setUp(self):
+        self.model_id = "unsloth/Llama-3.2-1B-Instruct"
+        self.tokenizer = AutoTokenizer.from_pretrained(self.model_id)
+        self.model_kernelized = AutoModelForCausalLM.from_pretrained(
+            self.model_id, use_kernels=True, device_map=torch_device
+        )
+        self.model_not_kernelized = AutoModelForCausalLM.from_pretrained(
+            self.model_id, use_kernels=False, device_map=torch_device
```
ydshieh: is this loading fast enough? We can probably do it once in `setupClass`

tests/kernels/test_kernels.py
ydshieh, line 1: test class or method that use real model checkpoint should use `@slow`.

If all test methods in a class quality `slow` in this sense, you can add `@slow` to the class and not each method

ydshieh, line 1: Maybe also ask Daniel to have a look here too.

ydshieh (commented):
A few comments to address. I saw you triggered a few runs. Is this working well?

.github/workflows/self-scheduled.yml
MekkCyber: yes we don't need any flags, we do evtn in `test_kernels.py`

tests/kernels/test_kernels.py
MekkCyber: Yes it's really fast, since we use a 1B model, and it's cached for all runs, but setupClass seems like a good way to do it too

ydshieh: then it's up to you, fine for me :)

ydshieh: but we still need to add `@slow`

.github/workflows/self-scheduled.yml
MekkCyber: good catch! thanks

tests/kernels/test_kernels.py
MekkCyber: yes done, I added `@slow` and used `setUpClass`

tests/kernels/test_kernels.py
```diff
@@ -0,0 +1,398 @@
...
+        cls.tokenizer = AutoTokenizer.from_pretrained(cls.model_id)
+        cls.model_kernelized = AutoModelForCausalLM.from_pretrained(
+            cls.model_id, use_kernels=True, device_map=torch_device
+        )
+        cls.model_not_kernelized = AutoModelForCausalLM.from_pretrained(
+            cls.model_id, use_kernels=False, device_map=torch_device
+        )
+        cls.input = "Hello"
+
+    def tearDown(self):
```
ydshieh: this should be  `tearDownClass` now.

we should still keep `tearDown` with only `cleanup(torch_device, gc_collect=True)`.

tests/kernels/test_kernels.py
```diff
@@ -0,0 +1,398 @@
...
+
+        # Clear any temporary kernel module cache entries populated by tests
+        try:
+            keys_to_remove = [
+                k for k, v in list(_KERNEL_MODULE_MAPPING.items()) if v is None or isinstance(v, types.ModuleType)
+            ]
+            for k in keys_to_remove:
+                _KERNEL_MODULE_MAPPING.pop(k, None)
+        except Exception:
+            pass
```
ydshieh, line 84: Could you explain why we needs this?

MekkCyber: when using kernels `_KERNEL_MODULE_MAPPING` gets populated by kernels, we are just cleaning it here for other tests

ydshieh (approved):
Still prefer @danieldk to take a look on the new test file (the logic of the testing `kernels`)

But LGTM from my side for testing

PR #41778, opened 2025-10-22
Fix Qwen3-Omni RoPE

zucchini-nlp:
# What does this PR do?

As per title, after the last PR with RoPE refactoring, Qwen3-Omni model has issues when loading the model. One of the many sub-configs doesn't call standardization on RoPE which causes issues

I also updated slow tests with correct checkpoint, right now they use Omni-2 checkpoints and thus do not test anything

cc @BakerBunker

### DIFF
--- a/src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py
+++ b/src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py
@@ -217,7 +217,7 @@ class Qwen3OmniMoeTextConfig(Qwen3MoeConfig):
         # Validate the correctness of rotary position embeddings parameters
         rope_theta = kwargs.get("rope_theta", 1000000.0)
         standardize_rope_params(self, rope_theta=rope_theta)
-        rope_config_validation(self)
+        rope_config_validation(self, ignore_keys={"mrope_section", "interleaved", "mrope_interleaved"})
 
 
 class Qwen3OmniMoeThinkerConfig(Qwen2_5OmniThinkerConfig):
@@ -1691,7 +1691,7 @@ class Qwen3OmniMoeTalkerOutputWithPast(MoeCausalLMOutputWithPast):
     generation_step: Optional[int] = None
 
 
-class Qwen3OmniMoeTalkerRotaryEmbedding(Qwen3RotaryEmbedding):
+class Qwen3OmniMoeTalkerRotaryEmbedding(Qwen3OmniMoeThinkerTextRotaryEmbedding):
     pass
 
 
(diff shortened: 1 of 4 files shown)


zucchini-nlp:
run-slow: qwen3_omni_moe

zucchini-nlp:
run-slow: qwen3_omni_moe

zucchini-nlp:
run-slow: qwen3_omni_moe

zucchini-nlp:
run-slow: qwen3_omni_moe

src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py
```diff
@@ -1681,7 +1691,7 @@ class Qwen3OmniMoeTalkerOutputWithPast(MoeCausalLMOutputWithPast):
     generation_step: Optional[int] = None
 
 
-class Qwen3OmniMoeTalkerRotaryEmbedding(Qwen3RotaryEmbedding):
+class Qwen3OmniMoeTalkerRotaryEmbedding(Qwen3OmniMoeThinkerTextRotaryEmbedding):
```
zucchini-nlp, line 1694: Simply reverting as it was before RoPE changes, talker has same type of RoPE as thinker

src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py
```diff
@@ -2480,7 +2490,7 @@ def generate(
         thinker_result = self.thinker.generate(input_ids=input_ids, **thinker_kwargs)
 
         if not generate_audio:
-            return thinker_result, None
+            return thinker_result
```
zucchini-nlp, line 2493: same as in qwen2-omni

src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py
```diff
@@ -2492,19 +2502,19 @@ def generate(
                 for hidden_states in thinker_result.hidden_states
             ],
             dim=1,
-        ).to(self.talker.device)  # [1 t d]
+        ).to(input_ids.device)  # [1 t d]
```
zucchini-nlp, line 2505: if the model is offloaded, the weights are in meta device. Instead let's rely on the actual device from input ids

tests/models/qwen3_omni_moe/test_modeling_qwen3_omni_moe.py
```diff
@@ -619,7 +619,9 @@ def test_get_rope_index_video_with_audio(self):
 @require_torch
 class Qwen2_5OmniModelIntegrationTest(unittest.TestCase):
     def setUp(self):
-        self.processor = AutoProcessor.from_pretrained("Qwen/Qwen2.5-Omni-7B")
+        self.processor = AutoProcessor.from_pretrained(
+            "Qwen/Qwen3-Omni-30B-A3B-Instruct", min_pixels=28 * 28, max_pixels=56 * 56
+        )
         self.audio_url = "https://qianwen-res.oss-cn-beijing.aliyuncs.com/Qwen2-Audio/audio/glass-breaking-151256.mp3"
```
zucchini-nlp, line 625: The diffs here are caused by updating the checkpoint after release

ArthurZucker (approved):
a small fast test would also be nice but LGTM

zucchini-nlp:
The slow tests should have triggered before merging but it was not using the correct checkpoint. So after I adjusted them, they will be failing without fix from this PR

I don't think we need a dummy test, because it fails for certain checkpoint from the hub

PR #41781, opened 2025-10-22
flash attn pytest marker

ydshieh:
# What does this PR do?

A first step toward making flash attn test again 🔥 

@vasqu Do we really need to have both

    @pytest.mark.flash_attn_test

    @pytest.mark.flash_attn_3_test

or we can have only `@pytest.mark.flash_attn_test` which cover both?


vasqu (approved):
Imo, yes as FA3 has more restrictions than FA2 so would be nice to differentiate them in tests --> can quickly check what is quirky about FA3 :D

If it's too annoying, I don't mind fusing more of a preference tbh

PR #39847, opened 2025-08-01
🚨 [v5] Refactor RoPE for layer types

zucchini-nlp:
# What does this PR do?

This PR enables rope layers to compute different frequencies for different layer types, which will help us to support models like ModernBert without monkey patching config on-the-fly

Main changes:
- In config classes the `rope_parameters` is a required attribute if model has RoPE layers. The attr must be a dict containing `rope_theta` and optionally other parameters to configure rope. In case we want different params per layer type, it should be a nested dict of format `{"full_attn": {**rope_params}, "sliding_attn": {**different_rope_params}}`
- The config attr `rope_scaling` is deprecated in favor of `rope_parameters` and raises warning. The latter name is more descriptive
- Default rope freq computation is moved to the model definition similar to `eager_attention_forward`, and copied with modular in each file
- RoPE layer now looks for layer types in the config and computes `inv_freq` for each type. If the given layer types has no rope parameters saved in config (e.g. `config.rope_scaling` has no key=="sliding_window") we raise an error
- All models copy from rope layers llama when possible, so that changing one file will update it everywhere. Models with layer types copy from gemma2
- Config classes now have typing hint in all language models and the rope scaling attribute is typed with  `TypedDict`. It will make our lives easier when we decide to enforce strict type validation on configs

The changes are BC and we will support old-format config files, and standardize it when initializing the config class. The best way to review is to start from `modeling_rope_utils.py` -> `all llama model files` -> `gemma2 and gemma33 model files` -> `tests`

PR #41789, opened 2025-10-22
Use indices as position_ids in modernebert

remi-or:
Currently there is an issue in `modernbert` stemming from #39847 : the `rotary_emb` always passes `position_ids` as if it were a tensor, but when the implementation is flash it can be `None`. To avoid the error `None has no attributes .shape` we use `indices` as the `positions_ids` in the rotary embedding.

### DIFF
--- a/src/transformers/models/modernbert/modeling_modernbert.py
+++ b/src/transformers/models/modernbert/modeling_modernbert.py
@@ -916,7 +916,11 @@ class ModernBertModel(ModernBertPreTrainedModel):
         hidden_states = self.embeddings(input_ids=input_ids, inputs_embeds=inputs_embeds)
         position_embeddings = {}
         for layer_type in self.config.layer_types:
-            position_embeddings[layer_type] = self.rotary_emb(hidden_states, position_ids, layer_type)
+            position_embeddings[layer_type] = self.rotary_emb(
+                x=hidden_states,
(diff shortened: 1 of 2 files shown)


src/transformers/models/modernbert/modeling_modernbert.py
```diff
@@ -916,7 +916,11 @@ def forward(
         hidden_states = self.embeddings(input_ids=input_ids, inputs_embeds=inputs_embeds)
         position_embeddings = {}
         for layer_type in self.config.layer_types:
-            position_embeddings[layer_type] = self.rotary_emb(hidden_states, position_ids, layer_type)
+            position_embeddings[layer_type] = self.rotary_emb(
+                x=hidden_states,
+                position_ids=(indices.unsqueeze(0) if position_ids is None else position_ids),
+                layer_type=layer_type,
+            )
```
zucchini-nlp: Nice catch, modern bert has special FA-RoPE layer for packed inputs. So if position embeddings are not used for FA2 models, maybe we can instead create them `if attn_implementationo="flash_attention"`

remi-or: Ok, updating this!

remi-or:
run-slow: modernbert

ArthurZucker (approved):

PR #41790, opened 2025-10-22
Fix attention mask in mamba layers

zucchini-nlp:
# What does this PR do?

As per title. It was reported by LFM-VL team that the batched generation is outputting garbage with one of the checkpoints. I found that the masking is not being applied at all for mamba layers

Firstly, mamba layer do not have a 4D attention weight and thus need a normal attention in 2D. Also we do not need to check if the attention has a certain shape, instead we only make sure it is applied in prefill stage

This fixes LFM-VL but Ig all mamba models are affected. I'm still surprised we didn't get issues before and that bigger checkpoints of LFM-VL generated normal text even without proper masking. I'm going to fix other mamba models and tag for review when ready


src/transformers/models/lfm2/modeling_lfm2.py
```diff
@@ -561,6 +562,7 @@ def forward(
         position_ids: Optional[torch.LongTensor] = None,
         past_key_values: Optional[Lfm2HybridConvCache] = None,
         cache_position: Optional[torch.LongTensor] = None,
+        attention_mask_conv: Optional[torch.LongTensor] = None,
```
vasqu: We should not introduce another mask, rather change the update logic per layer type above. See bamba as an example over here https://github.com/huggingface/transformers/blob/93671b4444414b01ea034bd64614856644297a66/src/transformers/models/bamba/modeling_bamba.py#L1202-L1213

src/transformers/models/lfm2/modeling_lfm2.py
```diff
@@ -422,7 +422,8 @@ def apply_mask_to_padding_states(hidden_states, attention_mask):
     """
     Tunes out the hidden states for padding tokens, see https://github.com/state-spaces/mamba/issues/66
     """
-    if attention_mask is not None and attention_mask.shape[1] > 1 and attention_mask.shape[0] > 1:
+    # Apply if seq length > 1, i.e. prefill stage only
+    if attention_mask is not None and hidden_states.shape[1] > 1:
```
vasqu: This condition shouldnt be needed to change or do we have sequences with bsz == 1 and padding? :thinking: 

We could relax this in other models as well tbh, no biggie, just a bit weird

vasqu (commented):
Wanted to get ahead before you invest a lot of time on other models.

src/transformers/models/lfm2/modeling_lfm2.py
vasqu: Oops, seeing it changed to hidden states... yea could be a tad more efficient for sure

zucchini-nlp: no, we don't usually get such inputs. I am updating still, so we'll check if input contains any masked tokens :)

src/transformers/models/lfm2/modeling_lfm2.py
zucchini-nlp: yep, PR is WIP still, forgot to mark it in title

src/transformers/models/bamba/modeling_bamba.py
```diff
@@ -1368,17 +1368,6 @@ def _prepare_4d_causal_attention_mask_with_cache_position(
...
-        """
-        No need for zeroing states when
-            1. Cached forward
-            2. Attending to all inputs
-        """
-        mamba_mask = attention_mask
-        if cache_position[0] > 0 or (attention_mask is not None and torch.all(attention_mask == 1)):
-            mamba_mask = None
-        return mamba_mask
-
```
zucchini-nlp, line 1381: not super needed to have a separate utility and "prepare" mask, since we aren't doing any modification. Instead made sure we check for prefill and padding in the `apply_mask_to_padding` function

Less code -> less chances to get smth wrong as it happened with LFM-VL

vasqu: But now we call `torch.all` on all layers, no? Not really sure if this is really (more) performant tbh.

Did the batched generation tests pass before? The core issue would still happen if we dont properly pass the correct mask per layer type so it doesnt really have much to do with the issue there.

Edit: And we do modify the mask, it's similar to flash attention preparation

zucchini-nlp: still WIP 😅 , I will tag when tests are fixed and all is ready for review

vasqu: Sorry 😔

zucchini-nlp:
run-slow: lfm2, lfm2_moe, lfm2_vl

zucchini-nlp:
run-slow: lfm2, lfm2_moe, lfm2_vl

zucchini-nlp:
run-slow: lfm2, lfm2_moe, lfm2_vl

tests/models/lfm2_vl/test_modeling_lfm2_vl.py
```diff
@@ -300,7 +300,7 @@ def test_integration_test_batched(self):
         )
 
         # Create inputs
-        text = ["<image>In this image, we see", "<image>In this image, we see"]
+        text = ["<image>In this image, we see", "<image>In this image, we see a cat"]
```
zucchini-nlp, line 303: changed just to make a one text longer than the other

src/transformers/models/lfm2_moe/modular_lfm2_moe.py
```diff
@@ -180,15 +180,18 @@ def forward(
             past_key_values=past_key_values,
             position_ids=position_ids,
         )
+        # Skip masking for decoding stage. We check shape here to be compile-friendly
+        mamba_attention = attention_mask if inputs_embeds.shape[1] != 1 else None
```
vasqu: Just to be sure, 2 questions just to be sure:
- 1. We use convolution here, not mamba right? Maybe make the naming a bit more explicit so that's not too confusing for external ppl
- 2. We can not reuse the same update logic as in mamba models with update_xxx? I think it's also fine more about a unified approach

zucchini-nlp: Nope, existing logic causes compile failures in LFM. Looks like other models are not compile compatible so it's fine for them

Note that we don't have a perfect way to check "if prefill" rn, and it caused us many problems in VLMs already. This check is neither perfect, e.g. it doesn't work with assisted decoding which is very niche in any case

zucchini-nlp: Joao had a plan to split out prefill from decoding, so we can better prepare inputs for generation. It hasn't made its way to a PR yet

vasqu: Gotcha, makes sense then let's rename the mask and then I'm fine with

src/transformers/models/lfm2/modeling_lfm2.py
```diff
@@ -665,15 +666,18 @@ def forward(
             past_key_values=past_key_values,
             position_ids=position_ids,
         )
+        # Skip masking for decoding stage. We check shape here to be compile-friendly
+        linear_attention = attention_mask if inputs_embeds.shape[1] != 1 else None
```
vasqu, line 670: Maybe convolution_mask?

vasqu (approved):
Thx a lot, trusting you on the slow tests :D

src/transformers/models/lfm2/modeling_lfm2.py
zucchini-nlp: "linear" was inspired by qwen-next, but it's just a matter of taste. Merging then, thanks for review

PR #41791, opened 2025-10-22
[`Onnx docs`] Remove some traces

vasqu:
Other langs docs are failing on main. This looked like the root cause


.github/scripts/codeowners_for_review_action
```diff
@@ -22,7 +22,6 @@ tests/generation/ @gante
 /src/transformers/models/auto/ @ArthurZucker
 /src/transformers/utils/ @ArthurZucker @Rocketknight1
 /src/transformers/loss/ @ArthurZucker
-/src/transformers/onnx/ @michaelbenayoun
```
vasqu, line 25: Since we have removed this folder its no longer needed to be maintained

ydshieh (approved):
Sorry about the breaking news!

PR #41792, opened 2025-10-22
Bump AMD docker

remi-or:
This PR bumps the base version of the AMD docker to ROCm 7.0, which means it is compatible with MI355 which was not the case before. 
To avoid issues with detectron, the install is split from the rest of the pip install, and to support MI355 for flash attn, the gfx950 architecture was added to FA compilation.
I tested the docker for Mi325 and Mi355 on important models and there was no major new failures.


ydshieh (approved):
AMD 🆙

PR #40894, opened 2025-09-15
Chat response parsing

Rocketknight1:
This PR is a replacement for #39609. The idea is that models can include a message schema, allowing model output to be parsed into a structured form. The original plan was to allow parsing of the entire chat history, essentially the inverse operation of `apply_chat_template`, but the schemas involved were too complex and there was no realistic hope that users would be able to write them!

This PR simplifies things - we focus only on parsing the output generated by the model. This is mainly relevant for **tool calling** and **chain of thought** models, both of which emit structured output that often needs manual handling before it can be appended to the chat.

The output schema is stored as a key on the tokenizer. It consists of a JSON schema, representing the structure of messages emitted by the model, with additional `x-` keys that indicate how parsing should be performed. Parsing is mostly done through regexes, but there is also support for common tool call formats like `json` to be directly parsed without you having to write an entire JSON regex parser :sweat_smile: 

Work to do:
- [x] Actually move parsing onto the tokenizer
- [x] Add support to TextGenerationPipeline
- [x] Add load/saving of output schemas
- [x] Document, document, document (50% done)
- [x] Figure out what we're calling it ("chat parsing"?)
- [x] Make sure extra fields don't break older versions
- [x] ~Support parsing in `Processor` classes too~  (Will move to separate PR)
- [x] ~Support parsing in `ImageTextToText` pipeline~ (Will move to separate PR)
- [x] Write schemas for some popular models
   - [x] GPT-OSS
   - [x] Cohere
   - [x] ERNIE
   - [x] ~Deepseek~ (tool calling isn't working in their template, will fix after)
   - [x] SmolLM3
   - [x] Qwen3
   - [x] Qwen3-coder (requires `xml` support)

Documentation to do:
- [x] Expand `parse_response` explanation and show how it works with `TextGenerationPipeline`
- [x] Move the writing guide to a separate doc, and expand it?
- [x] Write the reference for the allowed `x-` fields
- [ ] Add method docstring and make sure it shows up in the docs

Open questions:
- [ ] Fold `chat_parsing_utils.py` into `chat_template_utils.py`?

PR #41793, opened 2025-10-22
Fix chat schema tests

Rocketknight1:
My bad - the [chat schema PR](https://github.com/huggingface/transformers/pull/40894) added some tests that were failing after my final commits, and I didn't realize because they weren't running in the CI for the PR.

The main problem was that I dropped support for `"type": "any"`, which is no longer part of the JSON schema standard, but left it in some tests, which caused errors. This PR changes the code to read a missing `type` annotation as implicitly allowing any type, which is compatible with the standard. It also removes a test for Processor schema save/loading, which was dropped from the PR (but will be added soon!)

### DIFF
--- a/src/transformers/utils/chat_parsing_utils.py
+++ b/src/transformers/utils/chat_parsing_utils.py
@@ -55,7 +55,7 @@ def recursive_parse(
         return None
 
     # If not, we have to do a little parsing. First, set some vars and do basic validation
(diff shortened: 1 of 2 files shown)


ydshieh (approved):
🙏 

(I trust you that you run it locally , don't let me down 🙃 please)

Rocketknight1:
I did!

PR #41801, opened 2025-10-22
Closed prototype PR – superseded by future implementation

### LINKED ISSUE #35467: Support Flash Attention 2 for LayoutLMv3
### Feature request

Support Flash Attention 2 for LayoutLMv3.

### Motivation

Speed up document transformer training and inference.

### Your contribution

I can add it.

jackiehimel:
# What does this PR do?

Adds SDPA and FlashAttention-2 support to LayoutLMv3 following the same pattern as other models. Fully backward compatible.

SDPA converts masks to boolean format. FA2 uses `_upad_input` for variable-length sequences and avoids redundant unpads. Both fall back gracefully when needed. FA2 is O(N) memory vs O(N²).

Fixes #35467

# Changes

- Added `LayoutLMv3SdpaAttention` using `torch.nn.functional.scaled_dot_product_attention`
- Added `LayoutLMv3FlashAttention2` with `flash_attn_func` / `flash_attn_varlen_func`
- Both inherit from `LayoutLMv3Attention`
- Fallback to standard attention when backends unavailable or `output_attentions=True` / relative position bias is used

# Testing

- 121 tests passed in `test_modeling_layoutlmv3.py`
- Manually verified forward passes with/without attention masks

## Before submitting

- [ ] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).

## Who can review?

### DIFF
--- a/src/transformers/models/layoutlmv3/configuration_layoutlmv3.py
+++ b/src/transformers/models/layoutlmv3/configuration_layoutlmv3.py
@@ -105,7 +105,7 @@ class LayoutLMv3Config(PreTrainedConfig):
     ```"""
 
     model_type = "layoutlmv3"
-    
+
     # Support flags for attention implementations
     _supports_sdpa = True
     _supports_flash_attn_2 = True
--- a/src/transformers/models/layoutlmv3/configuration_layoutlmv3.py
+++ b/src/transformers/models/layoutlmv3/configuration_layoutlmv3.py
@@ -105,6 +105,10 @@ class LayoutLMv3Config(PreTrainedConfig):
     ```"""
 
     model_type = "layoutlmv3"
+    
+    # Support flags for attention implementations
+    _supports_sdpa = True
+    _supports_flash_attn_2 = True
 
     def __init__(
         self,
@@ -138,6 +142,7 @@ class LayoutLMv3Config(PreTrainedConfig):
         num_channels=3,
         patch_size=16,
         classifier_dropout=None,
+        _attn_implementation="eager",
         **kwargs,
     ):
         super().__init__(
@@ -173,6 +178,7 @@ class LayoutLMv3Config(PreTrainedConfig):
         self.num_channels = num_channels
         self.patch_size = patch_size
         self.classifier_dropout = classifier_dropout
+        self._attn_implementation = _attn_implementation
 
 
 __all__ = ["LayoutLMv3Config"]
--- a/src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
+++ b/src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
@@ -423,7 +423,7 @@ class LayoutLMv3SdpaAttention(LayoutLMv3Attention):
         else:
             attn_mask = None
             if attention_mask is not None:
-                attn_mask = attention_mask > 0
+                attn_mask = attention_mask >= 0
             
             attn_output = torch.nn.functional.scaled_dot_product_attention(
                 query_layer,
@@ -454,13 +454,13 @@ class LayoutLMv3FlashAttention2(LayoutLMv3Attention):
     def _upad_input(self, query_states, key_states, value_states, attention_mask, query_length):
         batch_size, seq_len, num_heads, head_dim = query_states.shape
         
-        indices_q, cu_seqlens_q, max_seqlen_q = unpad_input(query_states, attention_mask)
-        indices_k, cu_seqlens_k, max_seqlen_k = unpad_input(key_states, attention_mask)
-        indices_v, cu_seqlens_v, max_seqlen_v = unpad_input(value_states, attention_mask)
-        
         query_states = query_states.view(-1, num_heads, head_dim)
         key_states = key_states.view(-1, num_heads, head_dim)
         value_states = value_states.view(-1, num_heads, head_dim)
+        
+        query_states, indices_q, cu_seqlens_q, max_seqlen_q = unpad_input(query_states, attention_mask)
+        key_states, indices_k, cu_seqlens_k, max_seqlen_k = unpad_input(key_states, attention_mask)
+        value_states, indices_v, cu_seqlens_v, max_seqlen_v = unpad_input(value_states, attention_mask)
         
         return (
             query_states,
--- a/src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
+++ b/src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
@@ -380,7 +380,6 @@ class LayoutLMv3SdpaAttention(LayoutLMv3Attention):
             self,
             hidden_states,
             attention_mask=None,
-            head_mask=None,
             output_attentions=False,
             rel_pos=None,
             rel_2d_pos=None,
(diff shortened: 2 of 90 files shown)


src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -355,6 +364,173 @@ def forward(
         return outputs
 
 
+class LayoutLMv3SdpaAttention(LayoutLMv3Attention):
```
vasqu: Sorry but this approach is outdated and we changed it to a unified class instead. An older Bert version should give a good idea over here https://github.com/huggingface/transformers/blob/9db58abd6e21c34dbb44068432f2e4b146ab472b/src/transformers/models/bert/modeling_bert.py#L121-L258
the relative positions are no longer in main (deprecated) but since they are used here, it's a better pointer

We will also need to update the mask creations to something along https://github.com/huggingface/transformers/blob/91b5a680c050d56a395e4dd96eb1b48520bf752f/src/transformers/models/bert/modeling_bert.py#L749-L753

And you will need `_supports_xxx` flags in order to indicate that the model really supports these attention flavors, e.g. https://github.com/huggingface/transformers/blob/91b5a680c050d56a395e4dd96eb1b48520bf752f/src/transformers/models/bert/modeling_bert.py#L562-L564

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -355,6 +364,173 @@ def forward(
...
+            batch_size, seq_len, self.self.num_attention_heads, self.self.attention_head_size
+        ).transpose(1, 2)
+        key_layer = key_layer.view(
+            batch_size, seq_len, self.self.num_attention_heads, self.self.attention_head_size
+        ).transpose(1, 2)
+        value_layer = value_layer.view(
+            batch_size, seq_len, self.self.num_attention_heads, self.self.attention_head_size
+        ).transpose(1, 2)
+
+        if self.self.has_relative_attention_bias and rel_pos is not None:
```
vasqu: This is a fallback to eager attention, no?

vasqu (commented):
This approach is sadly outdated, I made a comment on what to do instead. I suspect that the tests also passed because the model didn't update their `_supports_xxx` flag. Without them these tests for FA, SDPA won't be run.

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
jackiehimel: Thanks for the feedback @vasqu! You're right - I've refactored to the unified class approach following the BERT pattern. I added the _supports_sdpa and _supports_flash_attn_2 flags to the config and updated the mask creation to use create_bidirectional_mask. 

Just pushed the changes - let me know if there's anything else that needs adjusting. Thanks again!

src/transformers/models/layoutlmv3/configuration_layoutlmv3.py
```diff
@@ -106,6 +106,10 @@ class LayoutLMv3Config(PreTrainedConfig):
 
     model_type = "layoutlmv3"
 
+    # Support flags for attention implementations
+    _supports_sdpa = True
+    _supports_flash_attn_2 = True
```
vasqu, line 111: These flags go under the pretrained model class, e.g. in llama https://github.com/huggingface/transformers/blob/77e8b9f8dfc8e736ad2f603a5b2ae2b1076ed271/src/transformers/models/llama/modeling_llama.py#L348-L353

src/transformers/models/layoutlmv3/configuration_layoutlmv3.py
```diff
@@ -173,6 +178,7 @@ def __init__(
         self.num_channels = num_channels
         self.patch_size = patch_size
         self.classifier_dropout = classifier_dropout
+        self._attn_implementation = _attn_implementation
```
vasqu, line 181: We handle this already, no need to modify anything in the config

src/transformers/models/layoutlmv3/my_changes.txt
vasqu, line 1: To be removed?

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -328,32 +344,174 @@ def forward(self, hidden_states: torch.Tensor, input_tensor: torch.Tensor) -> to
         return hidden_states
 
 
-# Copied from transformers.models.layoutlmv2.modeling_layoutlmv2.LayoutLMv2Attention with LayoutLMv2->LayoutLMv3
+# Adapted from transformers.models.layoutlmv2.modeling_layoutlmv2.LayoutLMv2Attention with LayoutLMv2->LayoutLMv3
+# Enhanced with unified attention implementation supporting eager, SDPA, and FlashAttention-2
```
vasqu, line 348: It would be better if we could change the upstream model instead. Or only copy relevant parts.

BUT, we also shouldn't modify this class by much. You should follow the structure of Bert for example: 
- Wrapper before the actual attention classes (without the additional cross attn logic for this model here) https://github.com/huggingface/transformers/blob/77e8b9f8dfc8e736ad2f603a5b2ae2b1076ed271/src/transformers/models/bert/modeling_bert.py#L309-L337
- The actual attention class https://github.com/huggingface/transformers/blob/9db58abd6e21c34dbb44068432f2e4b146ab472b/src/transformers/models/bert/modeling_bert.py#L121-L258

Currently, the wrapper class is modified and a lot of custom paths are introduced. We use our own integrations with the interface here (linking the files as examples below as well) https://github.com/huggingface/transformers/blob/9db58abd6e21c34dbb44068432f2e4b146ab472b/src/transformers/models/bert/modeling_bert.py#L236-L243
What needs to be handled:
- the mask creation
- adapting the new attention class (as per Bert for example)

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -328,32 +344,174 @@ def forward(self, hidden_states: torch.Tensor, input_tensor: torch.Tensor) -> to
...
 
+        # Store attention implementation config
+        self.is_causal = False
+        self._attn_implementation = config._attn_implementation if hasattr(config, "_attn_implementation") else "eager"
+
     def forward(
         self,
         hidden_states,
         attention_mask=None,
+        head_mask=None,
```
vasqu, line 367: Head mask is deprecated

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -328,32 +344,174 @@ def forward(self, hidden_states: torch.Tensor, input_tensor: torch.Tensor) -> to
...
+                self_outputs = self._sdpa_attention_forward(hidden_states, attention_mask, head_mask)
+
+        else:  # eager
+            self_outputs = self.self(hidden_states, attention_mask, output_attentions, rel_pos, rel_2d_pos)
+
         attention_output = self.output(self_outputs[0], hidden_states)
         outputs = (attention_output,) + self_outputs[1:]  # add attentions if we output them
         return outputs
 
+    def _sdpa_attention_forward(self, hidden_states, attention_mask=None, head_mask=None):
```
vasqu, line 406: See https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/sdpa_attention.py

which will be invoked when used with the interface

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -328,32 +344,174 @@ def forward(self, hidden_states: torch.Tensor, input_tensor: torch.Tensor) -> to
...
+            dropout_p=self.self.dropout.p if self.training else 0.0,
+            scale=1.0 / math.sqrt(self.self.attention_head_size),
+        )
+
+        attn_output = attn_output.transpose(1, 2).contiguous()
+        attn_output = attn_output.view(batch_size, seq_len, self.self.all_head_size)
+
+        return (attn_output,)
+
+    def _flash_attention_forward(self, hidden_states, attention_mask=None):
```
vasqu, line 459: See https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/flash_attention.py

which will be invoked when used with the interface

vasqu (commented):
Please check that you follow what other model implementations do for integrating other attentions variations. Bert is a good model to follow in this case, I also referenced an older version of Bert (with relative position support) which should help here. We like to avoid using custom paths and use our interface to handle specifics.

Currently, we likely still dont support the other attention variations as the flags are set incorrectly.

jackiehimel:
Thank you so much @vasqu for your detailed review and guidance! It is greatly appreciated. After looking more closely at the integration patterns and the BERT implementation, I realize this needs a more substantial refactor to match the current approach. Closing this for now but will revisit with a cleaner implementation later. Appreciate it!

PR #39754, opened 2025-07-29
Fix GPT2 with cross attention

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/39746

Issue #41639, opened 2025-10-16
MarianMTModel performance regression due to Bidirectional masks


jiqing-feng:
### System Info

torch 2.10.0.dev20251008+cpu

### Who can help?

@vasqu 

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

`numactl -C 0-3 --membind 0 python test.py`

```python
import time
import torch
from transformers import pipeline, AutoTokenizer

model_id = "Helsinki-NLP/opus-mt-mul-en"
tokenizer = AutoTokenizer.from_pretrained(model_id)
generator = pipeline("translation_fr_to_en", model=model_id, dtype=torch.float16, tokenizer=tokenizer)
generation_config = generator.model.generation_config
generation_config.max_new_tokens = 128
generation_config.min_new_tokens = 128
generation_config.do_sample = False
generation_config.temperature = 1.0
generation_config.num_beams = 1
generation_config.cache_implementation="static"
input_sentence = "C'est fait et soumis. Vous pouvez jouer à « Survival of the Tastiest » sur Android, et sur le web. Le jeu sur le web fonctionne, mais vous devez simuler plusieurs touches pour déplacer les tables, ce qui peut être un peu déroutant. Il y a beaucoup de choses dont j'aimerais parler. Je vais passer en revue chaque sujet, au lieu de faire la liste habituelle de ce qui s'est bien passé et de ce qui s'est mal passé. Concept Travailler sur le thème a probablement été l'une des tâches les plus difficiles auxquelles j'ai dû faire face. A l'origine, j'avais une idée du type de jeu que je voulais développer, du point de vue du gameplay - quelque chose avec beaucoup d'ennemis/acteurs"
for _ in range(5):
    output = generator(input_sentence, generation_config=generation_config)
for _ in range(5):
    start = time.time()
    output = generator(input_sentence, generation_config=generation_config)
    end = time.time()
    print(f"pipeline latency: {(end-start)*1000} ms")
```

### Expected behavior

The performance has 40% regression after PR [41265](https://github.com/huggingface/transformers/pull/41265).

Same issue as #41566 . But the fix PR #41586 only fixes the Bert model regression, and didn't fix MarianMTModel performance regression.

i3hz:
Oops that's my bad . I'll look into it and try to fix it .

[...]

PR #41806, opened 2025-10-23
revert `_prepare_4d_causal_attention_mask_with_cache_position` for gpt2

jiqing-feng:
Hi @zucchini-nlp 

The PR #39754 deleted `_prepare_4d_causal_attention_mask_with_cache_position` on gpt2, which caused 40% performance regression on CPU. You can reproduce it by 
`numactl -C 0-7 --membind 0 python test.py`

```python
import time
import torch
from transformers import pipeline, set_seed, AutoTokenizer

set_seed(42)

model_id = "openai-community/gpt2"
tokenizer.padding_side = 'left'
if tokenizer.pad_token_id is None:
    tokenizer.pad_token_id = tokenizer.eos_token_id

pipe = pipeline("text-generation", model=model_id, tokenizer=tokenizer, torch_dtype=torch.float16, device_map="cpu")

generation_config = pipe.model.generation_config
generation_config.do_sample = False
generation_config.use_cache = True
generation_config.max_new_tokens = 128
generation_config.min_new_tokens = 128
generation_config.cache_implementation="static"
generation_config.temperature = 1.0
generation_config.top_p = 1.0
generation_config.num_beams = 1
pipe.model.config._attn_implementation="sdpa"

inputs = "It is done, and submitted. You can play 'Survival of the Tastiest' on Android, and on the web. Playing on the web works, but you have to simulate multiple touch for table moving and that can be a bit confusing. There is a lot I'd like to talk about. I will go through every topic, insted of making the typical what went right/wrong list. Concept Working over the theme was probably one of the hardest tasks which I had to face. Originally, I had an idea of what kind of game I wanted to develop, gameplay wise - something with a lot of enemies/actors"

for _ in range(5):
    set_seed(42)
    pipe(inputs, generation_config=generation_config)

for _ in range(5):
    set_seed(42)
    start = time.time()
    pipe(inputs, generation_config=generation_config)
    end = time.time()
    print(f"{pipe.model.dtype} time costs {(end-start)*1000} ms")
```

Revert `_prepare_4d_causal_attention_mask_with_cache_position` can fix the regression.

### DIFF
--- a/src/transformers/models/gpt2/modeling_gpt2.py
+++ b/src/transformers/models/gpt2/modeling_gpt2.py
@@ -743,6 +743,59 @@ class GPT2Model(GPT2PreTrainedModel):
             cross_attentions=all_cross_attentions,
         )
 
+    @staticmethod
+    def _prepare_4d_causal_attention_mask_with_cache_position(
+        attention_mask: torch.Tensor,
+        sequence_length: int,
+        target_length: int,
+        dtype: torch.dtype,
+        cache_position: torch.Tensor,
+        batch_size: int,
+        **kwargs,
+    ):
+        """
+        Creates a causal 4D mask of shape `(batch_size, 1, query_length, key_value_length)` from a 2D mask of shape
+        `(batch_size, key_value_length)`, or if the input `attention_mask` is already 4D, do nothing.
+
+        Args:
+            attention_mask (`torch.Tensor`):
+                A 2D attention mask of shape `(batch_size, key_value_length)` or a 4D attention mask of shape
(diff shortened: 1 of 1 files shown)


jiqing-feng:
run-slow: gpt2

vasqu (commented):
It's the same issue as in #41639 (vmap in causal masking), I already discussed with @Cyrilvallez that we will add a non-vmap path to the mask creations. This will revert the perf regressions, so I'd like you to wait for the PR instead as we don't want to introduce old functions (we want to deprecate) back into the code.

jiqing-feng:
OK, please let me know when the PR is ready. Thanks!

PR #41808, opened 2025-10-23
QwenVL: add skipped keys in `setattr` as well

zucchini-nlp:
# What does this PR do?

As per title, getattr and setattr skip different set of keys now


molbap (approved):
Thanks! Sounds good

PR #41811, opened 2025-10-23
Add a safeguard around a flaky test in gemma2

remi-or:
The test `tests/models/gemma2/test_modeling_gemma2.py::Gemma2IntegrationTest::test_model_2b_pipeline_bf16_flex_attention` is flaky because of the `_compile` flag at https://github.com/huggingface/transformers/blob/fe11cbb808b4301399240e40c5cb6cca9bb00d4d/src/transformers/masking_utils.py#L675
The issue is solved when upgrading torch to nightly 2.10, so it makes little sense to change the code.
But when the test causes an error, it causes all subsequent tests to fail. So this PR adds the `@run_test_using_subprocess` decorator around the test so this does not happen anymore. And a TODO to remove this once the fix is in the stable version of torch.


ydshieh (approved):
Coolllllllll!

PR #41815, opened 2025-10-23
further reducing flakiness in `utils/check_bad_commit.py` (#41658) 

ydshieh:
# What does this PR do?

In #41690, we run each test (on each git bisect commit) 4 times to avoid the false positive being reported to specific team member.

However, when we check if a test passes on the commit of the current CI run (in the last job of checking bad commit) that failed in the same current run in the usual job for that model, we only needs 1 pass among 4 runs to conclude it's flaky.

Prior to this PR, it requires the 4 runs all passing to say "it pass on the check time" and conclude it's flaky which is incorrect, and this leads the false positive being still reported (even with less chance).

### DIFF
--- a/utils/check_bad_commit.py
+++ b/utils/check_bad_commit.py
@@ -142,8 +142,8 @@ def find_bad_commit(target_test, start_commit, end_commit):
     # Now, we are (almost) sure `target_test` is not failing at `end_commit`
     # check if `start_commit` fail the test
     # **IMPORTANT** we only need one pass to conclude the test is flaky on the current run with `start_commit`!
-    failed_now, n_failed, n_passed = is_bad_commit(target_test, start_commit)
-    if not failed_now or n_passed > 0:
+    _, n_failed, n_passed = is_bad_commit(target_test, start_commit)
+    if n_passed > 0:
         # failed on CI run, but not reproducible here --> don't report
         return None, f"flaky: test fails on the current CI run (commit: {start_commit}) but passes during the check."
 
--- a/utils/check_bad_commit.py
+++ b/utils/check_bad_commit.py
@@ -97,7 +97,17 @@ def is_bad_commit(target_test, commit):
     # Restore to original commit
     repo.git.checkout(original_head)
 
-    return result.returncode != 0
+    n_passed = 0
+    o = re.findall(r"====.* (\d+) passed", result.stdout)
+    if len(o) > 0:
+        n_passed = int(o[0])
+
+    n_failed = 0
+    o = re.findall(r"====.* (\d+) failed", result.stdout)
+    if len(o) > 0:
+        n_failed = int(o[0])
+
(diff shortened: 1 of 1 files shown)


utils/check_bad_commit.py
```diff
@@ -130,8 +141,9 @@ def find_bad_commit(target_test, start_commit, end_commit):
 
     # Now, we are (almost) sure `target_test` is not failing at `end_commit`
     # check if `start_commit` fail the test
-    failed_now = is_bad_commit(target_test, start_commit)
-    if not failed_now:
+    # **IMPORTANT** we only need one pass to conclude the test is flaky on the current run with `start_commit`!
+    failed_now, n_failed, n_passed = is_bad_commit(target_test, start_commit)
+    if not failed_now or n_passed > 0:
```
ydshieh: ok, I can probably change to `if n_passed > 0` here.

The point is: we only need 1 pass (among the multiple runs) to conclude it's flaky and not to report it.

utils/check_bad_commit.py
```diff
@@ -194,12 +206,13 @@ def get_commit_info(commit):
...
+    url = f"https://api.github.com/repos/huggingface/transformers/commits/{commit}"
+    commit_info = requests.get(url).json()
+    parent = commit_info["parents"][0]["sha"]
     if author is None:
-        url = f"https://api.github.com/repos/huggingface/transformers/commits/{commit}"
-        commit_info = requests.get(url).json()
         author = commit_info["author"]["login"]
 
-    return {"commit": commit, "pr_number": pr_number, "author": author, "merged_by": merged_author}
+    return {"commit": commit, "pr_number": pr_number, "author": author, "merged_by": merged_author, "parent": parent}
```
ydshieh, line 215: just add the parent commit , so easier to copy-paste (or automation) when a team member want to check after being pinged by me.

utils/check_bad_commit.py
```diff
@@ -97,7 +97,17 @@ def is_bad_commit(target_test, commit):
...
 
-    return result.returncode != 0
+    n_passed = 0
+    o = re.findall(r"====.* (\d+) passed", result.stdout)
+    if len(o) > 0:
+        n_passed = int(o[0])
+
+    n_failed = 0
+    o = re.findall(r"====.* (\d+) failed", result.stdout)
+    if len(o) > 0:
```
vasqu, line 107: ~Nit: but could we not just directly call len on the regex find all? We don't really need this extra if imo 🤔~

Nvm, interpreted the code wrong here

utils/check_bad_commit.py
vasqu: Yea fair enough, I'd be pro simplifying to only `n_passed` except there is a good reason to include the `failed_now`.

vasqu (approved):
Lgtm, just small nits 

Way better now 😄 curse flaky tests

PR #38635, opened 2025-06-06
[cache] make all classes cache compatible finally

zucchini-nlp:
# What does this PR do?

As per title, and let's get rid of `_supports_cache/_supports_quantized_cache` flags. We will assume all models from now on support cache and initialize a `DynamicCache` (model-specific cache in case of mamba) by default

For the static cache, we can't yet assume all models support it because even if the model technically can use `StaticCache`, it can't always compile fullgraph. We have an auto-compilation for static cache enabled, so maybe the compilation should check for smth like `_can_compile_fullgraph` and not `_supports_static_cache`?

I checked all models are updated and the generation tests are passing. Note, the PR depends on #38751 which cleans up non-generative models from `past_key_values`

PR #41818, opened 2025-10-23
:rotating_light: Fix gradient checkpointing for several models and improve test robustness  

githubnemo:
Support for gradient checkpointing was lost in the major refactoring in PR #38635 and this is the attempt to re-add it.

I extended the tests to
- test `use_reentrant=True` and `False`
- make sure `model.train` is called so that gradient checkpointing works; this is a limiation of the tests currently used by GPTBigCode
- make sure that one (the first) gradient checkpointing layer is called
- make sure that the same non-zero grads are there for normal and checkpointing runs - this is something we tripped over before in PEFT due to the possibly incompletely stored runtime environment in the checkpointed forward step, see also peft#2826

Note that the invocation of `GPTBigCodeBlock.forward` has changed:

- `layer_past` is now passed as a keyword argument so that `GradientCheckpointingLayer.__call__` can see and filter this parameter (`use_reentrant=False` fails otherwise)
- `{encoder_}hidden_states` are still passed as positional arguments so that `torch.utils.checkpoint.checkpoint` receives them as pos. args and computes gradients for these (kwargs would be filtered by `GradientCheckpointingLayer`).

:rotating_light: Note that this is breaking compatibility by changing the forward signature in `GPTBigCodeBlock.forward`!


src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
```diff
@@ -291,9 +292,9 @@ def __init__(self, config, layer_idx=None):
     def forward(
         self,
         hidden_states: Optional[tuple[torch.Tensor]],
+        encoder_hidden_states: Optional[torch.Tensor] = None,
         layer_past: Optional[Cache] = None,
         attention_mask: Optional[torch.Tensor] = None,
-        encoder_hidden_states: Optional[torch.Tensor] = None,
```
vasqu, line 296: Let's not change the order here, we could break things for users here. Rather change the args, kwargs positions if necessary on the module call

tests/models/gpt_bigcode/test_modeling_gpt_bigcode.py
```diff
@@ -307,18 +308,43 @@ def create_and_check_lm_head_model(self, config, input_ids, input_mask, token_ty
         self.parent.assertEqual(result.logits.shape, (self.batch_size, self.seq_length, self.vocab_size))
 
     def create_and_check_forward_and_backwards(
-        self, config, input_ids, input_mask, token_type_ids, *args, gradient_checkpointing=False
```
vasqu, line 310: I'm a bit surprised that it was overriden here. It would be nicer if we could move this into `test_modeling_common` instead.

tests/models/gpt_bigcode/test_modeling_gpt_bigcode.py
```diff
@@ -307,18 +308,43 @@ def create_and_check_lm_head_model(self, config, input_ids, input_mask, token_ty
...
+        if gradient_checkpointing_kwargs:
+            # Make sure that gradient checkpointing is executed (>1 forward call) and that the
+            # non-zero gradients of both inference runs match (since torch.util.checkpoint doesn't save
+            # the whole environment, the forward calls might differ).
+            model.gradient_checkpointing_enable(gradient_checkpointing_kwargs)
+
+            for _, p in trainable_params:
+                p.grad = None
+
+            checkpointing_layer = next(m for m in model.modules() if isinstance(m, GradientCheckpointingLayer))
```
vasqu: Ah ok so we are bound to the new gradient ckpting then. Guess there will be a need to check for all models to use this properly.

tests/models/gpt_bigcode/test_modeling_gpt_bigcode.py
```diff
@@ -307,18 +308,43 @@ def create_and_check_lm_head_model(self, config, input_ids, input_mask, token_ty
...
-            model.gradient_checkpointing_enable()
 
+        trainable_params = [(n, p) for (n, p) in model.named_parameters() if p.requires_grad]
         result = model(input_ids, token_type_ids=token_type_ids, labels=input_ids)
         self.parent.assertEqual(result.loss.shape, ())
         self.parent.assertEqual(result.logits.shape, (self.batch_size, self.seq_length, self.vocab_size))
         result.loss.backward()
 
+        non_zero_grads_normal = {n for n, p in trainable_params if p.grad.abs().sum() > 0}
+        assert non_zero_grads_normal
```
vasqu: Let's use https://github.com/huggingface/transformers/blob/81b4f9882c8a46c8274084503d297874bb372260/tests/models/bart/test_modeling_bart.py#L182
instead of normal asserts. Depends on if we move the test too ig.

vasqu (commented):
The tests are neat, I think we should move them to common tests tho. Not exactly sure why it was specially treated here. 

And ig there will be a need for another round to check similar models that may have been accidentally overriden with the ckpting layer :sweat: not necessarily this PR tho

src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
githubnemo: I'm not sure that this is possible. It is *mandatory* that we pass `layer_past` as keyword argument, otherwise `GradientCheckpointingLayer` will not be able to remove it from the kwargs in case of gradient checkpointing. On the other hand every input that may require gradients (`hidden_states`, `encoder_hidden_states`) must be passed as positional argument for `checkpoint()` to work. Maybe I'm missing something but I don't think we can bring those together without moving `encoder_hidden_states` up in the list.

tests/models/gpt_bigcode/test_modeling_gpt_bigcode.py
githubnemo: Agreed but I'm not sure how to deal with the fact that not all models use a `GradientCheckpointingLayer` and sometimes call the function by themselves. Do you have a suggestion how to deal with that?

src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
vasqu: I mean that the signature should stay the same, e.g. see https://github.com/huggingface/transformers/blob/84d19be41e0131e6f2a660fe6af8b77094906af7/src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py#L586-L596

It will need to adjust the calls from the module above like https://github.com/huggingface/transformers/blob/84d19be41e0131e6f2a660fe6af8b77094906af7/src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py#L901-L910

Changing the signature is breaking a bit too much!

tests/models/gpt_bigcode/test_modeling_gpt_bigcode.py
vasqu: We can check if the layer exists somewhere, no? If we do not detect that it exists, raise an error and check which models fails --> all models should already have this but bigger PRs apparently clashed so it isnt the case anymore

vasqu:
cc @ArthurZucker since this might become a bit more breaking than initially thought, and it likely affects more models

githubnemo:
I've updated the general tests. From the commit message:

    - Compare that the non-zero gradients in a reference run are present in the checkpointing run
    - Make sure that the forward of at least one gradient checkpointing layer is actually called
      more than once (as expected during gradient checkpointing backward)
    
    Currently there are some problems with Bert-derived MultipleChoice models, when dropout is
    enabled there are scenarios during gradient checkpointing where `classifier.bias.grad` is None.
    I don't yet have a good explanation for this, disabling dropout resolves this. I would have
    understood, if it is dropout on the classification layer but enabling attention dropout is
    also leading to this behavior.
    
    MoE models have selective sparsity depending on the selected experts, for this reason we
    only compare gradients on parameters collected on the reference backward run.

Currently these models are expected to fail since they're not implementing `GradientCheckpointingLayers`:

- swiftformer
- xlstm
- zamba
- zamba2

most likely these as well:
- janus (no training testing?)
- clvp

As I explained in the commit message, there's a strange bug with Bert-derived models when testing the `BertForMultipleChoice` case. When attention(!) dropout is active, sometimes the classification bias receives a `None` gradient. I don't have a good explanation for this right now but that seems fishy. Happy about any input.

I didn't revert the GPTBigCode test changes yet since I first wanted to get an opinion if we want to proceed with these more general tests or not.

tests/test_modeling_common.py
```diff
@@ -827,6 +829,12 @@ def test_gradient_checkpointing_enable_disable(self):
             model = model_class(copy.deepcopy(config))
             self.assertFalse(model.is_gradient_checkpointing)
 
+            # Gradient checkpointing is implemented via GradientCheckpointingLayer, if none is present this is likely
+            # an implementation issue. Note we exclude xlstm and zamba* for now since they are still not using
+            # GradientCheckpointingLayer.
+            if config.model_type not in ["xlstm", "zamba", "zamba2"]:
+                self.assertTrue([m for m in model.modules() if isinstance(m, GradientCheckpointingLayer)])
```
vasqu, line 836: So the same as below (with more models), no?

tests/test_modeling_common.py
```diff
@@ -1151,22 +1159,64 @@ def check_training_gradient_checkpointing(self, gradient_checkpointing_kwargs=No
...
+
+                # make sure that test runs are consistent by disabling dropout
+                #
+                # TODO I don't understand why attention_probs_dropout_prob influences classifier.bias in
+                # BertForMultipleChoice (and other Bert derived models). Sometimes classifier.bias is None
+                # when attention_probs_dropout_prob > 0. This might indicate a bug somewhere.
+                if hasattr(config, "hidden_dropout_prob"):
+                    config.hidden_dropout_prob = 0.0
+                if hasattr(config, "attention_probs_dropout_prob"):
+                    config.attention_probs_dropout_prob = 0.0
```
vasqu: This is only for the multiple choice class? Or are other model types also affected?

I don't think these have high usage either way so it's fine when we leave an explanation here

src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
vasqu: For viz, as discussed internally, we need this to be breaking

vasqu (commented):
I think this is fine, we will likely need to break other model's signature as well? I.e. not only got bigcode. This PR will get bigger than initially thought but let's fix these models

We can allow this for v5 but let's also mention this PR in the v5 thread (#40822) when we merge this.

tests/test_modeling_common.py
githubnemo: > This is only for the multiple choice class? Or are other model types also affected?

It didn't seem to make a difference for other models with the limited test runs I made. If you want I can limit it to the problematic model class.

tests/test_modeling_common.py
githubnemo: Yep, made it consistent.

tests/test_modeling_common.py
vasqu: I think it's fine with the comment but would be nice if you could double check if it may effect other model (classes)

tests/test_modeling_common.py
```diff
@@ -827,6 +829,20 @@ def test_gradient_checkpointing_enable_disable(self):
...
+            if config.model_type not in [
+                "xlstm",
+                "zamba",
+                "zamba2",
+                "swiftformer",
+                "janus_vqgan",
+                "clvp_encoder",
+                "clvp_decoder",
+            ]:
+                self.assertTrue([m for m in model.modules() if isinstance(m, GradientCheckpointingLayer)])
```
vasqu, line 836: Like discussed, let's fix these within this PR directly. Seems like these were more like unintentional regressions due to other big PRs

src/transformers/models/xlstm/modeling_xlstm.py
```diff
@@ -22,17 +22,21 @@
...
     from xlstm.xlstm_large.model import RMSNorm as xLSTMRMSNorm
-    from xlstm.xlstm_large.model import mLSTMBlock as xLSTMBlock
-    from xlstm.xlstm_large.model import mLSTMStateType, soft_cap
+    from xlstm.xlstm_large.model import mLSTMBlock, mLSTMStateType, soft_cap
 
     external_xlstm = True
+
+    class xLSTMBlock(GradientCheckpointingLayer, mLSTMBlock):
+        pass
+
```
vasqu, line 39:
```suggestion
```
?

vasqu (commented):
Just noticed this small thing in xlstm

Re: Clvp, let's isolate it for now. We can come back later except you have a good idea how to refactor/handle this properly

src/transformers/models/xlstm/modeling_xlstm.py
githubnemo: No, `mLSTMBlock` doesn't implement `GradientCheckpointingLayer`: https://github.com/NX-AI/xlstm/blob/main/xlstm/xlstm_large/model.py#L455

vasqu: Gotcha, it's an external dep didnt notice at first

src/transformers/models/gpt_bigcode/modeling_gpt_bigcode.py
vasqu: Just to be sure this model is the only one that needed a break in the signature?

tests/test_modeling_common.py
```diff
@@ -827,6 +829,17 @@ def test_gradient_checkpointing_enable_disable(self):
...
             self.assertFalse(model.is_gradient_checkpointing)
 
+            # Gradient checkpointing is implemented via GradientCheckpointingLayer, if none is present this is likely
+            # an implementation issue. Note we exclude janus and clvp for now since they are still not using
+            # GradientCheckpointingLayer.
+            if config.model_type not in [
+                "janus_vqgan",
+                "clvp",
+                "clvp_encoder",
+                "clvp_decoder",
```
vasqu: Would it be possible to override this test in `test_modeling_clvp` instead? No biggie if not

vasqu (approved):
Let's also change the title here a bit since the scope changed

cc @ArthurZucker @Cyrilvallez if you can take a last look --> fixes a last few gradient ckpting models and makes the test more robust towards actually using it properly

tests/test_modeling_common.py
githubnemo: I've removed some exceptions (janus, clvp in training test) but for clvp/clvp_decoder I think it is better to have the one single exception visible instead of duplicating the test code without the single assertion. It also didn't make sense to me to refactor the assertion into an abstract base method for the tests since it is only CLVP and nothing else.

vasqu: Gotcha, thanks for iterating here and bearing with me.

vasqu (approved):
LGTM, cc @Cyrilvallez for core maintainer

vasqu:
Can we also change the title of the PR tho. The PR also makes the test more robust and properly checks for gradient ckpting capabilities

Cyrilvallez (approved):
Not much to complain about, thanks a lot for this!

PR #41821, opened 2025-10-23
Share embedding modules in BART, not only weights

githubnemo:
Embedding modules are now shared between encoder, decoder and shared - it is the same module, like in the T5 implementation.

This has the benefit that it does not matter which module is returned in `get_input_embeddings`, the caller of the latter can be sure that modifications done to that (e.g., hooks) apply to the embeddings.

Background: While revamping the gradient checkpointing tests in PEFT via peft#2860 we found that the gradient enable step
(`modeling_utils.enable_input_require_grads`) does not work for BART. This leads to gradient checkpointing with `use_reentrant=True` to fail as it will not detect any gradients. The reason for this is that the returned value by `get_input_embeddings` (`self.shared`) is not the module that is called in the encoder, therefore any hooks added to `self.shared` are not run - in this case the hook set by `enable_input_require_grads`.

Since the background is a missing hook I've added a test that tests directly the ability to define hooks and their ability to being called.

### DIFF
--- a/src/transformers/models/bart/modeling_bart.py
+++ b/src/transformers/models/bart/modeling_bart.py
@@ -538,12 +538,12 @@ class BartEncoder(BartPreTrainedModel):
         self.max_source_positions = config.max_position_embeddings
         embed_scale = math.sqrt(embed_dim) if config.scale_embedding else 1.0
 
-        self.embed_tokens = BartScaledWordEmbedding(
-            config.vocab_size, embed_dim, self.padding_idx, embed_scale=embed_scale
-        )
-
         if embed_tokens is not None:
             self.embed_tokens = embed_tokens
+        else:
+            self.embed_tokens = BartScaledWordEmbedding(
+                config.vocab_size, embed_dim, self.padding_idx, embed_scale=embed_scale
+            )
 
         self.embed_positions = BartLearnedPositionalEmbedding(
             config.max_position_embeddings,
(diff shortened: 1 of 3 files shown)


tests/models/bart/test_modeling_bart.py
```diff
@@ -477,6 +478,21 @@ def test_inputs_embeds(self):
             with torch.no_grad():
                 model(**inputs)[0]
 
+    def test_input_embeddings_support_forward_hook(self):
```
vasqu, line 481: Can you add a comment here:
- Why we have this test
- Reference this PR

vasqu:
run-slow: bart

src/transformers/models/bart/modeling_bart.py
```diff
@@ -687,7 +687,7 @@ def __init__(self, config: BartConfig, embed_tokens: Optional[nn.Embedding] = No
         )
 
         if embed_tokens is not None:
-            self.embed_tokens.weight = embed_tokens.weight
+            self.embed_tokens = embed_tokens
```
vasqu, line 686: It might need to be directly set, not after creation. E.g. https://github.com/huggingface/transformers/blob/81b4f9882c8a46c8274084503d297874bb372260/src/transformers/models/t5/modeling_t5.py#L653

vasqu (commented):
Just a bit unsure about the failing multi gpu test: https://github.com/huggingface/transformers/actions/runs/18757351833/job/53512753724; I can check afterwards if this is due to the changes here that the tests failed

Just have some nits + make sure to check other models as well (with `make fix-copies`).

vasqu:
Checked in #41828 for the multi gpu tests that failed, seems like even without changes it fails in the runners...

Edit: seems like a general issue, talked internally in slack so nothing to do on your side

src/transformers/models/bart/modeling_bart.py
githubnemo: Can you elaborate?

vasqu: It's a big nit either way but as far as I know, the weights are always tied and we pass the shared embedding. So we could do the same as in T5: https://github.com/huggingface/transformers/blob/4faf675232d01b99329aa8dfb32ab1aa2fba980b/src/transformers/models/t5/modeling_t5.py#L653

(the embeddings are just sometimes unnecessary construced when we pass the tokens directly)

vasqu (approved):
Thx a lot for iterating!

PR #41824, opened 2025-10-23
Fix const parsing for dict inputs in chat schemas

Rocketknight1:
Found one more bug in chat schemas and updated the tests to cover it! `const` nodes were not being handled correctly when the parent node had structured content.

This PR also moves `test_chat_schema_utils.py` to `test_chat_parsing_utils.py` to match the name of the actual file `utils/chat_parsing_utils.py`.

Tests may not be running in the CI yet but I ran them all locally and they're passing!


gante (approved):
LGTM, thank you for fixing :)

PR #41830, opened 2025-10-23
fix continuous batching issues, extend ut cases to xpu

yao-matrix:
@SunMarc , pls help review, thx very much.


yao-matrix:
@SunMarc , could you pls take a review? Thx very much.

tests/generation/test_continuous_batching.py
```diff
@@ -20,7 +20,14 @@
 from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer, LogitsProcessorList
 from transformers.generation.continuous_batching.cache import group_layers_by_attn_type
 from transformers.generation.continuous_batching.continuous_api import build_attention_mask
-from transformers.testing_utils import Expectations, require_kernels, require_torch_gpu, slow
+from transformers.testing_utils import (
+    Expectations,
+    require_kernels,
+    require_torch_accelerator,
+    require_torch_gpu,
```
remi-or, line 28: Is @require_torch_gpu still used after those changes?

src/transformers/integrations/mxfp4.py
```diff
@@ -395,7 +400,7 @@ def load_and_swizzle_mxfp4(module, param_name, param_value, target_device, trito
         else:
             blocks = blocks.reshape(local_experts, -1, module.intermediate_size // 2)
         if getattr(target_device, "type", target_device) == "cpu":
-            target_device = "cuda"
+            target_device = torch.accelerator.current_accelerator().type if hasattr(torch, "accelerator") else "cuda"
```
remi-or, line 403: if @SunMarc can OK this, not familiar w/ different accelerators

remi-or (approved):
LGTM! Thanks for adding, I just have 2 nits :)

tests/generation/test_continuous_batching.py
yao-matrix: @remi-or paged attention enabling on XPU in kernels is still ongoing, suppose be ready soon, then we will transfer the left cases from CUDA-only to XPU, at this time, we still have some paged attention cases which are still CUDA only.

yao-matrix:
@SunMarc , I think we are OK to go now, thx very much.

src/transformers/integrations/mxfp4.py
SunMarc: should be fine !

PR #41832, opened 2025-10-23
HF Trainer: ALST/Ulysses sequence parallelism integration via HF Accelerate

stas00:
Integrates HF Accelerate's support for ALST/Ulysses sequence parallelism https://github.com/huggingface/accelerate/pull/3817 into HF Trainer

TODO:

- [x] docs - no idea where? the FSDP/CP is not documented, or any parallelism for that matter.
- [x] tests
- [x] need to merge https://github.com/huggingface/accelerate/pull/3817
- [x] need to wait for a new HF Accelerate release and to use its version in compatibility checks in the code 1.11.1 most likely
- [x] need to wait for a merge of https://github.com/deepspeedai/DeepSpeed/pull/7649
- [x] need to wait for a new Deepspeed release after above is merged


stas00:
I have a hard time finding where I can add the documentation for this new backend. Context parallelism isn't documented anywhere in HF Trainer - how can users discover it? If I'm missing it, could you please point me to where I should extend the documentation? Thank you.

Same story with CP tests - there are none :( so need to figure out how to write some.

stas00:
OK, the next issue with the existing integration of CP/FSDP. What does the following mean?
```
 $ sometrainerscript.py --help
 [...]
 --parallelism_config PARALLELISM_CONFIG, --parallelism-config PARALLELISM_CONFIG
```
this arg tells users absolutely nothing about what value(s) to pass to `--parallelism_config`. If it's not meant to be in CLI args and can only be used explicitly by writing code perhaps it shouldn't be listed in `--help` or at least say that it has to be coded?

Rocketknight1:
cc @sunmarc

src/transformers/testing_utils.py
```diff
@@ -2005,14 +2006,12 @@ def get_env(self):
         paths = [self.repo_root_dir_str, self.src_dir_str]
         if "/examples" in self.test_file_dir_str:
             paths.append(self.examples_dir_str)
-        else:
-            paths.append(self.tests_dir_str)
```
stas00, line 2009: To reviewers I added this path 5 years ago when I created this library, I don't remember now why, but I see the deepspeed tests break with recent deepspeed versions if installed in `devel` style, because if we included `tests` in `sys.path` the `deepspeed` dir is attempted to be loaded as the real `deepspeed` module and things break massively. So removing this. 

If for some reason this breaks something else I can of course have a look, but the other recourse is to rename `tests/deepspeed` to something else.

src/transformers/testing_utils.py
```diff
@@ -2005,14 +2006,12 @@ def get_env(self):
...
         paths.append(env.get("PYTHONPATH", ""))
 
         env["PYTHONPATH"] = ":".join(paths)
         return env
 
-    def get_auto_remove_tmp_dir(self, tmp_dir=None, before=None, after=None):
+    def get_auto_remove_tmp_dir(self, tmp_dir=None, before=None, after=None, return_pathlib_obj=False):
```
stas00, line 2014: this is a really old version. In the latest incarnation it always return a `Path` object. But to keep BC, I added a new flag here instead. The tests are less clunkier then.

The latest version is here: https://github.com/stas00/ml-engineering/blob/master/testing/testing_utils.py

If wanted you could switch to the latest version instead and adapt tests to simplify.

stas00:
@SunMarc, this is ready for a review. Tests fail because they need the accelerate PR https://github.com/huggingface/accelerate/pull/3817

I just didn't know where to update docs since parallelism doesn't seem to be documented here at all. Please correct me if I'm wrong.

Thanks to @kashif with the test.

src/transformers/testing_utils.py
SunMarc: cc @ydshieh

src/transformers/trainer.py
```diff
@@ -2159,6 +2159,17 @@ def train(
...
             )
 
+    def get_cp_size(self) -> int:
+        """Get the context parallel size"""
+        if getattr(self.accelerator, "parallelism_config", None) is None:
+            return 1
+        pc = self.accelerator.parallelism_config
+        if pc.cp_backend == "deepspeed":
+            return pc.cp_size
+        # XXX: not sure if backend=="torch" needs to return cp_size here as well it wasn't originally (a bug?)
```
SunMarc: let's return cp_size in all cases. This was probably an oversight from matej

src/transformers/trainer.py
```diff
@@ -3820,6 +3845,33 @@ def compute_loss(
...
+                sp_group = self.accelerator.torch_device_mesh["cp"].get_group()
+                sp_world_size = pc.cp_size
+                # differentiable weighted per-shard-loss aggregation across ranks
+                losses_per_rank = torch.distributed.nn.functional.all_gather(loss, group=sp_group)
+                # special dealing with SFT that has prompt tokens that aren't used in loss computation
+                good_tokens = (shift_labels != -100).view(-1).sum()
+                good_tokens_per_rank = torch.distributed.nn.functional.all_gather(good_tokens, group=sp_group)
+                total_loss = sum(losses_per_rank[rank] * good_tokens_per_rank[rank] for rank in range(sp_world_size))
+                total_good_tokens = sum(good_tokens_per_rank)
+                loss = total_loss / max(total_good_tokens, 1)
```
SunMarc: We probably don't need to do this if `num_items_in_batch` is computed and passed in in `unwrapped_model.loss_function`.  `num_items_in_batch` was introduced to fix the gradient accumulation https://unsloth.ai/blog/gradient. `num_items_in_batch` is basically `total_good_tokens` if grad_acc = 1.

src/transformers/trainer.py
```diff
@@ -5136,7 +5194,9 @@ def _get_num_items_in_batch(self, batch_samples: list, device: torch.device) ->
                     # In the DataParallel case, convert the scalar tensor into a 2-dim tensor with the same value repeated
                     num_items_in_batch = num_items_in_batch.unsqueeze(0).expand(self.args.n_gpu, -1)
                 # Divide by number of devices with the same batch
-                if pc := getattr(self.accelerator, "parallelism_config", None):
+                # XXX: double check if this is only for fsdp
+                pc = getattr(self.accelerator, "parallelism_config", None)
+                if pc is not None and pc.cp_backend == "torch":
```
SunMarc: you didn't pass `num_items_in_batch` in the loss computation or in the model, so this is why it has no effect in your tests. One thing to test would be to change the bs and grad_acc value to check if the loss if the same.

src/transformers/trainer.py
```diff
@@ -3820,6 +3845,33 @@ def compute_loss(
...
+
+            outputs = model(**inputs)
+            shift_labels = inputs["shift_labels"]
+            loss = unwrapped_model.loss_function(
+                logits=outputs.logits,
+                labels=None,
+                shift_labels=shift_labels,
+                vocab_size=unwrapped_model.config.vocab_size,
+            )
+
```
SunMarc: Can we try to merge this code into the existing one instead of doing everything here again ? Or we should write a helper function for `deepspeed_cp_loss` so that it is easier to maintain. Also in TRL, they usually overwrite `compute_loss`. cc @kashif @qgallouedec 
 if you have any comment on how to upstream this feature more easily

SunMarc (commented):
Thanks for this clean integration ! Left a couple of comments. It would be great @kashif @qgallouedec if you can have a look at this PR so that we can also make it compatible with TRL.

kashif:
Thanks @SunMarc I did an initial run with default settings in TRL and these branches, and it all worked nicely (apart from model saving due to not updating deepspeed i think). I will check the config that uses a bespoke `compute_loss` in the sft trainer, thanks for the heads up!

SunMarc:
> @SunMarc, this is ready for a review. Tests fail because they need the accelerate PR https://github.com/huggingface/accelerate/pull/3817
> 
> I just didn't know where to update docs since parallelism doesn't seem to be documented here at all. Please correct me if I'm wrong.
> 
> Thanks to @kashif with the test.

We indeed do not have docs related to that. @kashif as you added cp support in trainer, would you be willing to add some docs around that.
@stas00, I think we can either update deepspeed docs https://huggingface.co/docs/transformers/main/en/deepspeed and/or create a new docs called contextparallel like in accelerate.

kashif:
@SunMarc docs on TRL side: https://github.com/huggingface/trl/pull/4420

kashif:
i can look at the docs on the accelerate side as well...

src/transformers/trainer.py
stas00: sorry, what is not needed?

This code block is because we need to compute the correct loss across SP ranks. If you just average those it'll be incorrect in the case of -100 masked tokens (SFT), since each rank is likely to process a different number of unmasked tokens (this is not DP averaging).

Unless what you mean is that we don't need to calculate `total_good_tokens` since  `num_items_in_batch` is already that, but the rest of the code remains - did I understand you correctly?

src/transformers/trainer.py
stas00: I initially tried to have it integrated and it was a mess as it'd require many complex conditional. Given it's a simple self-contained code I thought it'd be much easier to maintain and avoid bugs if it remains self-contained.

I can move it into `deepspeed_cp_loss` no problem - is that what you prefer?

stas00:
> @stas00, I think we can either update deepspeed docs https://huggingface.co/docs/transformers/main/en/deepspeed and/or create a new docs called contextparallel like in accelerate.

Ideally we would have a dedicated doc like you suggested, which could then link into deepspeed for nuances as one way to do that. The key is for the user to quickly understand what's possible, thus a single context parallel entry point doc would be very useful to users.

src/transformers/trainer.py
SunMarc: if you pass `num_items_in_batch` in `loss_function`, it will sum the loss then divide it by `num_items_in_batch` directly. This way I think we don't need to actually to recalculate the `total_loss` from the averaged losses and the good_tokens_per_rank. Maybe I'm wrong so please correct me ! But I think this might solve the grad acc issue. In any case, we will keep the current code as not all models accepts `num_items_in_batch` when calculating the loss. 

`total_loss = sum(losses_per_rank[rank] * good_tokens_per_rank[rank] for rank in range(sp_world_size))`

```python
def ForCausalLMLoss(
    logits,
    labels,
    vocab_size: int,
    num_items_in_batch: Optional[int] = None,
    ignore_index: int = -100,
    shift_labels: Optional[torch.Tensor] = None,
    **kwargs,
) -> torch.Tensor:
    # Upcast to float if we need to compute the loss to avoid potential precision issues
    logits = logits.float()

    if shift_labels is None:
        # Shift so that tokens < n predict n
        labels = nn.functional.pad(labels, (0, 1), value=ignore_index)
        shift_labels = labels[..., 1:].contiguous()

    # Flatten the tokens
    logits = logits.view(-1, vocab_size)
    shift_labels = shift_labels.view(-1)
    # Enable model parallelism
    shift_labels = shift_labels.to(logits.device)
    loss = fixed_cross_entropy(logits, shift_labels, num_items_in_batch, ignore_index, **kwargs)
    return loss


def fixed_cross_entropy(
    source: torch.Tensor,
    target: torch.Tensor,
    num_items_in_batch: Optional[int] = None,
    ignore_index: int = -100,
    **kwargs,
) -> torch.Tensor:
    reduction = "sum" if num_items_in_batch is not None else "mean"
    loss = nn.functional.cross_entropy(source, target, ignore_index=ignore_index, reduction=reduction)
    if reduction == "sum":
        loss = loss / num_items_in_batch
    return loss

```

stas00: If you pass `num_items_in_batch` you indeed don't need to do local loss calculation since it'll do that already. But we need to calculate a distributed across ranks loss.

Here is an example: Let's take a 2k tokens sample SP-split across 2 ranks using SFT:

1. SP rank0 - 900 masked and 100 non-masked tokens (a long initial prompt that is -100 masked out)
2. SP rank1 - 100 masked and 900 non-masked tokens

So each rank produces the correct loss if we use `num_items_in_batch` - but how do you combine the losses of 2 ranks. straight average will give a very skewed result, because the rank0's loss contributes 9x less non-masked tokens.

Let's take it to a more telling example:

1. SP rank0 - 1000 masked and 0 non-masked tokens (a long initial prompt that is masked out)
2. SP rank1 - 0 masked and 1000 non-masked tokens

here rank0 can't even contribute anything to the total loss -  a normal averaging of 2 losses would be completely broken, since you'd average with an undefined behavior, since the loss function will return a `NaN` or `None`.

src/transformers/trainer.py
stas00: done.

src/transformers/testing_utils.py
stas00: it's much better for it to always return a `pathlib.Path` object but you'd need to tweak a few tests which use this API.

src/transformers/trainer.py
stas00: Thank you for validating! Comment question removed.

src/transformers/trainer.py
stas00: Marc, please feel free to tweak it in any way you'd like. This looked like the most maintainer-friendly solution to me (and I wrote many parts of the original HF Trainer back in the day). Deepspeed often handles many things on its own, so it sort of fits with its behavior almost everywhere else. 

Deepspeed preceeds many of the modern techs so back then it had to do everything internally as there was no external support for much.

SunMarc: Yeah then let's move it into _deepspeed_cp_compute_loss() for now !

src/transformers/trainer.py
SunMarc: > So each rank produces the correct loss if we use num_items_in_batch - but how do you combine the losses of 2 ranks. straight average will give a very skewed result, because the rank0's loss contributes 9x less non-masked tokens.

The denominator of the losses is both `num_items_in_batch`, the value of each loss already takes into account the number of non-masked tokens as we do `reduction = "sum"`. So we just sum them to get the final loss. In your first examples, `num_items_in_batch` will be equal to 1000. For rank0, the loss will be equal to (L1+...L100)/1000 and for rank1, it will be (l1+..+l900)/1000

stas00: I have a feeling we are missing each other. I'm talking about differentiable loss combination across ranks and I think you're talking about the local rank's loss.

Could you please point me to the code in HF Trainer that performs a differentiable loss combination across multiple ranks? I couldn't find any.

src/transformers/trainer.py
stas00: pushed.

kashif:
@stas00 @SunMarc added initial docs https://github.com/stas00/transformers/pull/3

stas00:
Thanks a lot, Kashif - merged your docs.

kashif:
deepspeed  1.18.2 is out BTW

sfc-gh-sbekman:
Thanks Kashif, yes, I already updated the OP. I need to find time to do a massive adaptation to `s/cp/sp/` since it's wrong to call SP CP and it will prevent future combination of SP+CP, since the 2 are unrelated to each other, while seemingly doing the same thing. Marc has already given a green light, I was just busy with other work.

sfc-gh-sbekman:
@SunMarc, `s/cp/sp/` is done - this is ready for your review. Also made some corresponding tweaks on the Accelerate size.

We would need to merge the Accelerate PR first, then make a new release there, then anchor this PR on that new release.

docs/source/en/deepspeed.md
```diff
@@ -368,6 +368,104 @@ The example ZeRO-3 and ZeRO-Infinity config below sets most of the parameter val
...
+)
+```
+
+As mentioned earlier SP and DP by default use the same ranks, in which case you only need to configure `sp_size`. But sometimes you may want to make SP-replicas. For example, let's say we have 8 GPUs, and you want 2 replicas of SP. Then set `sp_size=4` and `dp_replicate_size=2`:
+
+```py
+parallelism_config = ParallelismConfig(
+    sp_backend="deepspeed",
+    sp_size=4,
+    dp_replicate_size: 2,
```
kashif:
```suggestion
    dp_replicate_size=2,
```

zhangwj618:
Can I quickly mention 2 issues I discovered while trying out the new sequence-parallelism:
1. Trainer.get_cp_size() currently doesn't take pc.sp_size into account, and hence total_train_batch_size is not correct;
2. After accelerator.deepspeed_ulysses_dl_adapter() gets called, len_dataloader is not updated and hence steps_in_epoch is not correct.
@stas00 Hope this helps and thanks for the great work!

docs/source/en/deepspeed.md
```diff
@@ -411,44 +411,44 @@ parallelism_config:
...
 
 Sequence parallelism can be combined with data parallelism for even better scaling. The example below uses 8 GPUs with 4-way sequence parallelism and 2-way data parallelism.
 
 ```py
 parallelism_config = ParallelismConfig(
-    cp_backend="deepspeed",
-    cp_size=4,
+    sp_backend="deepspeed",
+    sp_size=4,
     dp_shard_size=2,
```
SunMarc: from the comment in accelerate, it should be dp_replicate_size here no ?

src/transformers/trainer.py
```diff
@@ -2176,8 +2184,8 @@ def get_tp_size(self) -> int:
     def get_total_train_batch_size(self, args) -> int:
         """Calculates total batch size (micro_batch * grad_accum * dp_world_size).
 
-        Note: Only considers DP and TP (dp_world_size = world_size // tp_size)."""
-        dp_world_size = args.world_size // self.get_tp_size()
+        Note: Only considers DP and TP and SP/CP (dp_world_size = world_size // tp_size // cp_size)."""
+        dp_world_size = args.world_size // self.get_tp_size() // self.get_cp_size()
```
SunMarc: need to update for SP

src/transformers/trainer.py
```diff
@@ -2301,6 +2309,11 @@ def _inner_training_loop(
         else:
             self.optimizer = self.accelerator.prepare(self.optimizer)
 
+        # since DataLoader was Accelerate prepared w/o a model arg in the same call, we now have to complete the DL wrapping for ALST/UlyssesSP, after model has been prepared
+        pc = getattr(self.accelerator, "parallelism_config", None)
+        if pc is not None and pc.sp_backend == "deepspeed":
+            train_dataloader = self.accelerator.deepspeed_ulysses_dl_adapter(train_dataloader, model)
+
```
SunMarc, line 2344: I prefer to be careful so maybe also check for `pc.sp_enabled` here ?

src/transformers/trainer.py
```diff
@@ -3635,23 +3648,32 @@ def _prepare_context_parallel_inputs(self, model, inputs: dict[str, Union[torch.
...
+                        _ignore_index = -100
+                        labels = nn.functional.pad(inputs["labels"], (0, 1), value=_ignore_index)
+                        inputs["shift_labels"] = labels[:, 1:].contiguous()
+
+            # carve out space to make it clear there are other backends with different requirements, even though no code needs to be run at the moment
+            elif self.accelerator.parallelism_config.sp_backend == "deepspeed":
+                # - accelerator.parallelism_config performs the `model.config._attn_implementation` checks already and it supports more than `dspa`
+                # - UlyssesSPDataLoaderAdapter called from Accelerate performs the `shift_label` creation - must not interfere
+                # - position_ids generation should be done by HF Trainer if it wasn't done by the user
+                pass
```
SunMarc: let's remove this since this is better to not put sp related code here

src/transformers/trainer.py
```diff
@@ -3820,6 +3842,10 @@ def compute_loss(
         Subclass and override for custom behavior. If you are not using `num_items_in_batch` when computing your loss,
         make sure to overwrite `self.model_accepts_loss_kwargs` to `False`. Otherwise, the loss calculating might be slightly inaccurate when performing gradient accumulation.
         """
+        pc = getattr(self.accelerator, "parallelism_config", None)
+        if pc is not None and pc.sp_backend == "deepspeed":
```
SunMarc: check for `pc.sp_enabled`, otherwise, this will most likely be True

src/transformers/trainer.py
```diff
@@ -3873,6 +3899,49 @@ def compute_loss(
...
+        outputs = model(**inputs)
+        shift_labels = inputs["shift_labels"]
+        loss = unwrapped_model.loss_function(
+            logits=outputs.logits,
+            labels=None,
+            shift_labels=shift_labels,
+            vocab_size=unwrapped_model.config.vocab_size,
+        )
+
+        if pc.sp_size > 1:
```
SunMarc: this check should always be True here so let's remove that and check for it before.

src/transformers/trainer.py
```diff
@@ -3913,7 +3982,11 @@ def save_model(self, output_dir: Optional[str] = None, _internal_call: bool = Fa
             Path(os.path.join(output_dir, "user_content.pt")).touch()
         # We are in N-D parallelism if we have parallelism_config set, so we check accelerate if we're on a to_save rank
         elif getattr(self.accelerator, "parallelism_config", None) is not None:
-            if self.accelerator.should_save_model:
+            # deepspeed already takes care of saving the checkpoint below, so we need this only for the torch cp backend
+            if (
+                self.accelerator.should_save_model
+                and getattr(self.accelerator, "parallelism_config").cp_backend == "torch"
+            ):
```
SunMarc: update this for sp deepspeed

stas00:
Thanks a lot for trying it out and spotting problems, @zhangwj618. I should have time to look at it at the end of the day, unless you already have the fixes and would like to propose them.

src/transformers/trainer.py
```diff
@@ -5136,7 +5211,8 @@ def _get_num_items_in_batch(self, batch_samples: list, device: torch.device) ->
                     # In the DataParallel case, convert the scalar tensor into a 2-dim tensor with the same value repeated
                     num_items_in_batch = num_items_in_batch.unsqueeze(0).expand(self.args.n_gpu, -1)
                 # Divide by number of devices with the same batch
-                if pc := getattr(self.accelerator, "parallelism_config", None):
+                pc = getattr(self.accelerator, "parallelism_config", None)
+                if pc is not None and pc.cp_backend == "torch":
```
SunMarc: let's revert this and just set `self.model_accepts_loss_kwargs` to False when sp is enabled. This way we will skip all the logic related to `num_items_in_batch`

SunMarc (commented):
Thanks for updating ! Just a few nits

src/transformers/trainer.py
stas00: no. the dp world size doesn't change with sp>1. The magic happens in the DL Adapter which gets a unique data stream from all ranks, and then uses sp ranks to process one sample at a time - this is sort of gradient accumulation.

docs/source/en/deepspeed.md
sfc-gh-sbekman: you must have had an old version opened when you did the comments, it's already so.

src/transformers/trainer.py
sfc-gh-sbekman: remove the code but to leave the comments in place? Breadcrumbs help the maintainers to know why some parts are skipped. But it's up to you

docs/source/en/deepspeed.md
SunMarc: indeed !

src/transformers/trainer.py
SunMarc: yeah we can do that

src/transformers/trainer.py
```diff
@@ -2159,6 +2159,14 @@ def train(
...
             )
 
+    def get_cp_size(self) -> int:
+        """Get the context parallel size"""
+        if getattr(self.accelerator, "parallelism_config", None) is None:
+            return 1
+        else:
+            pc = self.accelerator.parallelism_config
+            return pc.cp_size
+
```
kashif, line 2186: Account for both context parallelism (torch/FSDP2) and sequence parallelism (DeepSpeed)

```suggestion
    def get_cp_size(self) -> int:
        """Get the context/sequence parallel size"""
        if getattr(self.accelerator, "parallelism_config", None) is None:
            return 1
        else:
            pc = self.accelerator.parallelism_config
            return pc.cp_size * pc.sp_size

```

src/transformers/trainer.py
sfc-gh-sbekman: I don't see anything to update here, what am I missing? This code is to be run for torch.

src/transformers/trainer.py
kashif, line 2344:
```suggestion
            # Update len_dataloader since UlyssesSPDataLoaderAdapter changes the dataloader length (multiplies by sp_world_size)
            len_dataloader = len(train_dataloader) if has_length(train_dataloader) else None

```

src/transformers/trainer.py
stas00: It looks like you're proposing UPS (SP+CP), let's do one step at a time and just add SP first, check that it works well and then see if SP+CP can be made to work?

stas00: This PR only proposes DP+SP combo, which has been well tested.

kashif:
@stas00  I have fixes for the issue from @zhangwj618 as suggestions above for your consideration

src/transformers/trainer.py
stas00: done

src/transformers/trainer.py
stas00: done

stas00: thank you so much, Kashif! That was one of the bugs @zhangwj618 spotted!

src/transformers/trainer.py
stas00: thank you - done.

src/transformers/trainer.py
stas00: unless you propose to revert the code part? I thought more explicit will make it easier down the road.

src/transformers/trainer.py
stas00: done

src/transformers/trainer.py
stas00: done

src/transformers/trainer.py
stas00: oh, wait, is this to fix this reported issue?

> Trainer.get_cp_size() currently doesn't take pc.sp_size into account, and hence total_train_batch_size is not correct;

stas00: in which case we should rework how it's calculated, otherwise we are confusing sp with cp

stas00: Please see the proposed fix https://github.com/huggingface/transformers/pull/41832/commits/f55427760eab4d79748addcd28c0a0310e5d51fd

stas00: the variable name becomes wrong then, since `DP != world_size / SP` but it's `DP == SP == world_size`. The magic happens on the DL level where it instead extends the DL length by SP times.

src/transformers/trainer.py
```diff
@@ -2176,8 +2192,12 @@ def get_tp_size(self) -> int:
...
         """Calculates total batch size (micro_batch * grad_accum * dp_world_size).
 
-        Note: Only considers DP and TP (dp_world_size = world_size // tp_size)."""
-        dp_world_size = args.world_size // self.get_tp_size()
+        Note: Only considers DP and TP and SP/CP (dp_world_size = world_size // tp_size // cp_size // sp_size).
+
+        Though do note that DP_world_size isn't args.world_size // self.get_sp_size, since SP ranks reuse DP ranks, so the variable below is misleading.
+        """
+
+        dp_world_size = args.world_size // self.get_tp_size() // self.get_cp_size() // self.get_sp_size
```
kashif:
```suggestion
        dp_world_size = args.world_size // self.get_tp_size() // self.get_cp_size() // self.get_sp_size()
```

src/transformers/trainer.py
stas00: the confusion is in the variable name - let's continue this discussion here - where the same fix is being discussed:
https://github.com/huggingface/transformers/pull/41832#discussion_r2539092844

stas00:
@zhangwj618, could you please re-test with the current HEAD? Kashif fixed both issues.

src/transformers/trainer.py
```diff
@@ -2176,8 +2192,12 @@ def get_tp_size(self) -> int:
     def get_total_train_batch_size(self, args) -> int:
         """Calculates total batch size (micro_batch * grad_accum * dp_world_size).
 
-        Note: Only considers DP and TP (dp_world_size = world_size // tp_size)."""
-        dp_world_size = args.world_size // self.get_tp_size()
+        Note: Only considers DP and TP and SP/CP (dp_world_size = world_size // tp_size // cp_size // sp_size).
+
+        Though do note that DP_world_size isn't args.world_size // self.get_sp_size, since SP ranks reuse DP ranks, so the variable below is misleading.
+        """
```
kashif: Fix comment
```suggestion
        Accounts for all parallelism dimensions: TP, CP, and SP.
        
        Formula: dp_world_size = world_size // (tp_size * cp_size * sp_size)
        
        Where:
        - TP (Tensor Parallelism): Model layers split across GPUs
        - CP (Context Parallelism): Sequences split using Ring Attention (FSDP2)
        - SP (Sequence Parallelism): Sequences split using ALST/Ulysses (DeepSpeed)
        
        All dimensions are separate and multiplicative: world_size = dp_size * tp_size * cp_size * sp_size
        """
```

docs/source/en/deepspeed.md
```diff
@@ -368,6 +368,104 @@ The example ZeRO-3 and ZeRO-Infinity config below sets most of the parameter val
...
+  parallelism_config_sp_size: 4  # Sequence parallel size
+  parallelism_config_sp_backend: deepspeed
+  parallelism_config_sp_seq_length_is_variable: true
+  parallelism_config_sp_attn_implementation: sdpa
+```
+
+Important configuration parameters include the following.
+
+* `sp_backend` must be set to `"deepspeed"` to use ALST/Ulysses sequence parallelism.
+* `sp_size` is the degree of sequence parallelism. For example, `sp_size=4` means 4 GPUs will process a single sequence in parallel. You need at least 2 GPUs to enable sequence parallelism. Please note that SP and DP reuse the same GPU ranks, so with 4 GPUs by default you would have SP=4 and DP=4.
```
kashif:
```suggestion
* `sp_size` is the degree of sequence parallelism. For example, `sp_size=4` means 4 GPUs will process a single sequence in parallel. You need at least 2 GPUs to enable sequence parallelism. Note that SP and DP dimensions multiply: with 4 GPUs and `sp_size=4`, you have SP=4 and DP=1 (since `dp_world_size = world_size / sp_size = 4 / 4 = 1`)
```

docs/source/en/deepspeed.md
```diff
@@ -368,6 +368,104 @@ The example ZeRO-3 and ZeRO-Infinity config below sets most of the parameter val
...
+from transformers import DataCollatorForLanguageModeling
+
+data_collator = DataCollatorForLanguageModeling(
+    tokenizer=tokenizer,
+    mlm=False,
+    pad_to_multiple_of=4,  # Ensure sequences are divisible by sp_size
+)
+```
+
+As mentioned earlier SP and DP by default use the same ranks, in which case you only need to configure `sp_size`. But sometimes you may want to make SP-replicas. For example, let's say we have 8 GPUs, and you want 2 replicas of SP. Then set `sp_size=4` and `dp_replicate_size=2`:
```
kashif:
```suggestion
By default, when you only configure `sp_size`, DP is automatically calculated as `dp_size = world_size / sp_size`. For example, with 8 GPUs and `sp_size=4`, you get `dp_size=2`. But sometimes you may want to explicitly configure data parallelism replicas. For example, let's say we have 8 GPUs and you want 2 replicas of SP across 4 GPUs each. Then set `sp_size=4` and `dp_replicate_size=2`:
```

stas00:
@kashif, how do we communicate that the DL should still feed unique data streams on each rank. That's why I was trying to say with `DP=SP=world_size`, since the sample sharding happens in the DL Adapter, unlike TP where each rank has to feed the same sample - which would be very wrong here.

kashif:
@stas00 good point! Let me see where to add the important difference.

docs/source/en/deepspeed.md
```diff
@@ -368,6 +368,104 @@ The example ZeRO-3 and ZeRO-Infinity config below sets most of the parameter val
...
+  parallelism_config_sp_size: 4  # Sequence parallel size
+  parallelism_config_sp_backend: deepspeed
+  parallelism_config_sp_seq_length_is_variable: true
+  parallelism_config_sp_attn_implementation: sdpa
+```
+
+Important configuration parameters include the following.
+
+* `sp_backend` must be set to `"deepspeed"` to use ALST/Ulysses sequence parallelism.
+* `sp_size` is the degree of sequence parallelism. For example, `sp_size=4` means 4 GPUs will process a single sequence in parallel. You need at least 2 GPUs to enable sequence parallelism. Note that SP and DP dimensions multiply: with 4 GPUs and `sp_size=4`, you have SP=4 and DP=1 (since `dp_world_size = world_size / sp_size = 4 / 4 = 1`)
```
kashif:
```suggestion
* `sp_size` is the degree of sequence parallelism. For example, `sp_size=4` means 4 GPUs will process a single sequence in parallel. You need at least 2 GPUs to enable sequence parallelism. **Data feeding**: Each rank receives a unique data stream from the DataLoader (like DP). **Batch size calculation**: The effective `dp_world_size = world_size / sp_size`. So with 4 GPUs and `sp_size=4`, each of the 4 ranks gets different samples from the DataLoader, but `dp_world_size=1` for total batch size calculations
```

docs/source/en/deepspeed.md
```diff
@@ -368,6 +368,104 @@ The example ZeRO-3 and ZeRO-Infinity config below sets most of the parameter val
...
+    sp_handler=DeepSpeedSequenceParallelConfig(
+        sp_seq_length_is_variable=True,
+        sp_attn_implementation="flash_attention_2",
+    ),
+)
+```
+
+[`Trainer`] automatically handles the special requirements for sequence parallelism including:
+
+* Adapting the data loader via DeepSpeed's [`UlyssesSPDataLoaderAdapter`](https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/runtime/sequence_parallel/ulysses_sp.py) to shard sequences across GPUs
```
kashif:
```suggestion
* Adapting the data loader via DeepSpeed's [`UlyssesSPDataLoaderAdapter`](https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/runtime/sequence_parallel/ulysses_sp.py) to shard sequences across GPUs. **Important**: Unlike Tensor Parallelism (where all ranks must receive identical data), each rank with SP receives a unique data stream from the DataLoader (similar to DP). The adapter handles distributing sequence chunks across SP ranks internally, so your DataLoader should continue feeding different samples to each rank.
```

zhangwj618:
> @zhangwj618, could you please re-test with the current HEAD? Kashif fixed both issues.

I tested the current code, and realized that max_steps (and state.max_steps) also has the same problem of not taking sp_size into account, affecting ProgressCallback.training_bar and potentially a few other things. So maybe the best place to update len_dataloader with sp_size is inside set_initial_training_values(), not after?

kashif:
@zhangwj618, just thinking, should the SP affect the number of training/optimizer steps?  The SP adapter multiplies dataloader length because each rank needs `sp_world_size` iterations to process all sequence chunks. But does this mean we should take `sp_world_size` times more optimizer steps? What do you think?

zhangwj618:
@kashif imo yes, specifically in the epoch-based max_steps calculation. Because with sp the dp_world_size and the global batch_size become 1 / sp_size times smaller (c.f. get_total_train_batch_size()), hence max_steps needs to be sp_size times bigger to go through the same number of epochs. Think of decreasing per_device_train_batch_size in the pure dp scenario, which would result in more training steps, increasing sp_size alone (keeping per_device_train_batch_size fixed) has the same effect. To keep the training steps constant, the per_device_train_batch_size need to be manually adjusted to sp_times bigger (which restores the original global batch_size when multiplied by 1 / sp_size).

zhangwj618:
Another way to look at it is that [UlyssesSPDataLoaderAdapter](https://github.com/deepspeedai/DeepSpeed/blob/64c0052fa08438b4ecf4cae30af15091a92d2108/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L442) yields sp_size times more data batches than the original DL, thus requires more steps to iterate through.

kashif:
@zhangwj618 ok I added a fix in this PR: https://github.com/stas00/transformers/pull/4 if you want to double check on your end you can manually add the change

src/transformers/trainer.py
SunMarc: the thing is that `cp_backend` default value is "torch" even if you decide to use sp with `sp_backend="deepspeed"`. So it will always run. maybe change to something like this ?

```python
pc = getattr(self.accelerator, "parallelism_config")
if self.accelerator.should_save_model and not (pc.sp_enabled and  pc.sp_backend == "deepspeed"):
```

src/transformers/trainer.py
```diff
@@ -3820,6 +3861,11 @@ def compute_loss(
         Subclass and override for custom behavior. If you are not using `num_items_in_batch` when computing your loss,
         make sure to overwrite `self.model_accepts_loss_kwargs` to `False`. Otherwise, the loss calculating might be slightly inaccurate when performing gradient accumulation.
         """
+        pc = getattr(self.accelerator, "parallelism_config", None)
+        if pc is not None and pc.sp_backend == "deepspeed" and pc.sp_enabled:
+            self.model_accepts_loss_kwargs = False  # does its own treatment of num_items_in_batch
```
SunMarc: It will be better to do that as early as possible in the `__init__`. For example, you can maybe put it that after this: 

```python
        # Check if the model has explicit setup for loss kwargs,
        # if not, check if `**kwargs` are in model.forward
        if hasattr(unwrapped_model, "accepts_loss_kwargs"):
            self.model_accepts_loss_kwargs = unwrapped_model.accepts_loss_kwargs
        else:
            forward_params = inspect.signature(unwrapped_model.forward).parameters
            self.model_accepts_loss_kwargs = any(
                k.kind == inspect.Parameter.VAR_KEYWORD for k in forward_params.values()
            )
```

zhangwj618:
> @zhangwj618 ok I added a fix in this PR: [stas00#4](https://github.com/stas00/transformers/pull/4) if you want to double check on your end you can manually add the change

Yep, from my side, with this PR, the number of steps looks correct.

zhangwj618:
@stas00 There is a potential issue inside _deepspeed_sp_compute_loss(), that on some rank(s) good_tokens could be 0 and loss would be nan. Skipping these ranks when suming the total_loss can fix this problem. I did some tests and can confirm that, after fixing this, the loss convergence is aligned between sp+dp and dp-only setups (using the same global batch_size).

src/transformers/trainer.py
```diff
@@ -3873,6 +3919,48 @@ def compute_loss(
...
+        )
+
+        sp_group = self.accelerator.torch_device_mesh["sp"].get_group()
+        sp_world_size = pc.sp_size
+        # differentiable weighted per-shard-loss aggregation across ranks
+        losses_per_rank = torch.distributed.nn.functional.all_gather(loss, group=sp_group)
+        # special dealing with SFT that has prompt tokens that aren't used in loss computation
+        good_tokens = (shift_labels != -100).view(-1).sum()
+        good_tokens_per_rank = torch.distributed.nn.functional.all_gather(good_tokens, group=sp_group)
+        total_loss = sum(losses_per_rank[rank] * good_tokens_per_rank[rank] for rank in range(sp_world_size))
```
kashif: Skip rank with zero valid tokens

```suggestion
        # Skip ranks with zero valid tokens
        total_loss = sum(
            losses_per_rank[rank] * good_tokens_per_rank[rank]
            for rank in range(sp_world_size)
            if good_tokens_per_rank[rank] > 0
        )
```

kashif: something like this @zhangwj618 ?

stas00:
Indeed, in our reference implementation the nan's were already avoided earlier so weighted sum was safe:
https://github.com/snowflakedb/ArcticTraining/blob/d9328917d77fd888e80ff368f273447aa7c0ff09/arctic_training/trainer/sft_trainer.py#L129-L139

Looks like Kashif has already fixed it here:
https://github.com/huggingface/transformers/pull/41832/commits/6ae2bb0b2caf0e2ebe261605604db4daab75c535

src/transformers/trainer.py
stas00: perhaps @kashif can help here?

kashif: sure I can have a look

kashif: resolved by https://github.com/stas00/transformers/pull/5

stas00: @kashif, please feel free to push directly ;)

src/transformers/trainer.py
```diff
@@ -3913,7 +4008,11 @@ def save_model(self, output_dir: Optional[str] = None, _internal_call: bool = Fa
             Path(os.path.join(output_dir, "user_content.pt")).touch()
         # We are in N-D parallelism if we have parallelism_config set, so we check accelerate if we're on a to_save rank
         elif getattr(self.accelerator, "parallelism_config", None) is not None:
-            if self.accelerator.should_save_model:
+            # deepspeed already takes care of saving the checkpoint below, so we need this only for the torch cp backend
+            if (
+                self.accelerator.should_save_model
+                and getattr(self.accelerator, "parallelism_config").cp_backend == "torch"
+            ):
```
kashif:
```suggestion
            # DeepSpeed SP already handles checkpoint saving below, so skip manual save in that case
            pc = getattr(self.accelerator, "parallelism_config")
            if self.accelerator.should_save_model and not (pc.sp_enabled and pc.sp_backend == "deepspeed"):
```

src/transformers/trainer.py
```diff
@@ -3873,6 +3921,53 @@ def compute_loss(
 
         return (loss, outputs) if return_outputs else loss
 
+    def _deepspeed_sp_compute_loss(self, model, inputs, return_outputs, pc):
+        """
+        How the loss is computed by Trainer under context parallelism scenario with sp_backend==deepspeed and sp_size>1. By default, all models return the loss in the first element.
```
kashif:
```suggestion
        How the loss is computed by Trainer under sequence parallelism with sp_backend=="deepspeed" and sp_size>1.
        Performs weighted loss aggregation across SP ranks, accounting for varying numbers of valid tokens per rank
        (e.g., when some ranks receive only padding or prompt tokens that are masked with -100).
```

src/transformers/trainer.py
```diff
@@ -3873,6 +3921,55 @@ def compute_loss(
...
+        """
+        How the loss is computed by Trainer under sequence parallelism with sp_backend=="deepspeed" and sp_size>1.
+        Performs weighted loss aggregation across SP ranks, accounting for varying numbers of valid tokens per rank
+        (e.g., when some ranks receive only padding or prompt tokens that are masked with -100).
+
+        Args:
+            model (`nn.Module`):
+                The model to compute the loss for.
+            inputs (`dict[str, Union[torch.Tensor, Any]]`):
+                The input data for the model.
```
kashif:
```suggestion
                The input data for the model. Must include "shift_labels" key.
```

kashif (approved):

zhangwj618:
Is there a mismatch between the docs and the code? `docs/source/en/deepspeed.md` says:
"By default, when you only configure `sp_size`, DP is automatically calculated as `dp_size = world_size / sp_size`."
However, when I run the code with sp_size != world_size, I get [this](https://github.com/stas00/accelerate/blob/a4005e7265575a31280e6db7ae909ded9ea83328/src/accelerate/parallelism_config.py#L353) error, unless I specify dp_replicate_size manully.

kashif:
@zhangwj618 the doc is wrong... my bad let me fix it!

docs/source/en/deepspeed.md
```diff
@@ -368,6 +368,104 @@ The example ZeRO-3 and ZeRO-Infinity config below sets most of the parameter val
...
+> For more detailed information about sequence parallelism, see the Accelerate [Sequence Parallelism](https://huggingface.co/docs/accelerate/concept_guides/sequence_parallelism) guide.
+
+To enable ALST/Ulysses sequence parallelism with [`Trainer`], configure `parallelism_config` in [`TrainingArguments`]. Sequence parallelism is configured via Accelerate's `ParallelismConfig` and requires Accelerate version higher than 1.11.0.
+
+```py
+from accelerate.utils import ParallelismConfig, DeepSpeedSequenceParallelConfig
+
+parallelism_config = ParallelismConfig(
+    sp_backend="deepspeed",
+    sp_size=4,  # Number of GPUs to split sequence across
```
kashif, line 389:
```suggestion
# Example: 4 GPUs with sp_size=4, dp_replicate_size=1 (no data parallelism)
# Ensure total_size = dp_replicate_size * dp_shard_size * sp_size = 1 * 1 * 4 = 4 GPUs
parallelism_config = ParallelismConfig(
    sp_backend="deepspeed",
    sp_size=4,  # Number of GPUs to split sequence across
    dp_replicate_size=1,  # Explicit: no data parallelism
```

docs/source/en/deepspeed.md
```diff
@@ -368,6 +368,107 @@ The example ZeRO-3 and ZeRO-Infinity config below sets most of the parameter val
...
+
+```yaml
+distributed_type: DEEPSPEED
+deepspeed_config:
+  deepspeed_config_file: path/to/ds_config.json
+machine_rank: 0
+num_machines: 1
+num_processes: 4  # Total number of processes
+parallelism_config:
+  parallelism_config_sp_size: 4  # Sequence parallel size
```
kashif, line 414:
```suggestion
  parallelism_config_sp_size: 4  # Sequence parallel size
  parallelism_config_dp_replicate_size: 1  # Must be: dp_replicate_size * dp_shard_size * sp_size = num_processes
```

docs/source/en/deepspeed.md
```diff
@@ -368,6 +368,108 @@ The example ZeRO-3 and ZeRO-Infinity config below sets most of the parameter val
...
+from transformers import DataCollatorForLanguageModeling
+
+data_collator = DataCollatorForLanguageModeling(
+    tokenizer=tokenizer,
+    mlm=False,
+    pad_to_multiple_of=4,  # Ensure sequences are divisible by sp_size
+)
+```
+
+By default, when you only configure `sp_size`, DP is automatically calculated as `dp_size = world_size / sp_size`. For example, with 8 GPUs and `sp_size=4`, you get `dp_size=2`. But sometimes you may want to explicitly configure data parallelism replicas. For example, let's say we have 8 GPUs and you want 2 replicas of SP across 4 GPUs each. Then set `sp_size=4` and `dp_replicate_size=2`:
```
kashif:
```suggestion
When using `sp_size` with multiple GPUs, you **must** explicitly set `dp_replicate_size` or `dp_shard_size` to ensure `total_size = dp_replicate_size * dp_shard_size * sp_size` equals your total number of GPUs. For example, with 8 GPUs and `sp_size=4`, you must set `dp_replicate_size=2` (since 2 × 1 × 4 = 8):
```

src/transformers/training_args.py
```diff
@@ -1139,7 +1139,7 @@ class TrainingArguments:
     )
     parallelism_config: Optional[ParallelismConfig] = field(
         default=None,
-        metadata={"help": ("Parallelism configuration for the training run. Requires Accelerate `1.10.1`")},
+        metadata={"help": ("Parallelism configuration for the training run. Requires Accelerate `1.11.1`")},
```
SunMarc: don't forget to update those 
```suggestion
        metadata={"help": ("Parallelism configuration for the training run. Requires Accelerate `1.12.0`")},
```

src/transformers/trainer.py
```diff
@@ -4981,9 +5080,11 @@ def create_accelerator_and_postprocess(self):
 
         # We defer compatibility checks to accelerator
         if self.args.parallelism_config is not None:
-            if not is_accelerate_available("1.10.1"):
+            # XXX: this will need to change once https://github.com/huggingface/accelerate/pull/3817 is merged and 1.11.1 is out
+            min_accelerate_version = "1.10.1"
```
SunMarc: I will release a new version of accelerate when the deepspeed pr is merged
```suggestion
            min_accelerate_version = "1.12.0"
```

src/transformers/trainer.py
```diff
@@ -4986,9 +5085,11 @@ def create_accelerator_and_postprocess(self):
 
         # We defer compatibility checks to accelerator
         if self.args.parallelism_config is not None:
-            if not is_accelerate_available("1.10.1"):
+            # XXX: this will need to change once https://github.com/huggingface/accelerate/pull/3817 is merged and 1.11.1 is out
```
kashif:
```suggestion
```

docs/source/en/deepspeed.md
```diff
@@ -368,6 +368,108 @@ The example ZeRO-3 and ZeRO-Infinity config below sets most of the parameter val
...
+
+Arctic Long Sequence Training (ALST) uses a combination of sharding inputs along the sequence dimension and attention head parallelism. With this approach, you can train models with sequence lengths up to 500K tokens on a single H100 GPU, 3.7M on a single node, or 15M tokens on just four nodes with Llama-8B. The implementation described here enables one component of the full ALST system. For additional optimizations like TiledMLP and activation checkpoint offloading, refer to the [DeepSpeed ALST tutorial](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/).
+
+> [!TIP]
+> For more detailed information about sequence parallelism, see the Accelerate [Sequence Parallelism](https://huggingface.co/docs/accelerate/concept_guides/sequence_parallelism) guide.
+
+To enable ALST/Ulysses sequence parallelism with [`Trainer`], configure `parallelism_config` in [`TrainingArguments`]. Sequence parallelism is configured via Accelerate's `ParallelismConfig` and requires Accelerate version higher than 1.11.0.
```
kashif:
```suggestion
To enable ALST/Ulysses sequence parallelism with [`Trainer`], configure `parallelism_config` in [`TrainingArguments`]. Sequence parallelism is configured via Accelerate's `ParallelismConfig` and requires an Accelerate version higher than 1.12.0.
```

ArthurZucker (approved):

SunMarc (approved):

ArthurZucker:
Thanks a lot @stas00 for your work 🤗

stas00:
super! Thanks a lot to Marc and Kashif for help with integration and Weijie Zhang for being the first early adopter!

PR #41837, opened 2025-10-24
multiple tokenizers with different filenames can save now

### LINKED ISSUE #41816: Processor saving does not work when multiple tokenizers
### System Info

- Platform: macOS-26.0.1-arm64-arm-64bit
- Python version: 3.12.2
- Safetensors version: 0.5.3
- Accelerate version: 1.10.1
- PyTorch version (accelerator?): 2.8.0 (NA)
- Jax version: not installed

### Who can help?

@Cyrilvallez 

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

Currently, processors are saved with fixed names:
https://github.com/huggingface/transformers/blob/ca01fe4d133a498a81bdfd335024d14801e106a2/src/transformers/utils/__init__.py#L272-L278

That means, that if you create a processor that uses two of the same kind of subprocessors (for example, a byte level tokenizer and a BPE tokenizer, or two image processors, etc), they override eachother, because they use the same file name.

```py
import tempfile

from transformers import ProcessorMixin, AutoTokenizer, PreTrainedTokenizer


class OtherProcessor(ProcessorMixin):
    name = "other-processor"

    attributes = [
        "tokenizer1",
        "tokenizer2",
    ]
    tokenizer1_class = "AutoTokenizer"
    tokenizer2_class = "AutoTokenizer"

    def __init__(self,
                 tokenizer1: PreTrainedTokenizer,
                 tokenizer2: PreTrainedTokenizer
                 ):
        super().__init__(tokenizer1=tokenizer1,
                         tokenizer2=tokenizer2)


tokenizer1 = AutoTokenizer.from_pretrained("google/gemma-3-270m")
tokenizer2 = AutoTokenizer.from_pretrained("HuggingFaceTB/SmolLM2-1.7B")

processor = OtherProcessor(tokenizer1=tokenizer1,
                           tokenizer2=tokenizer2)

with tempfile.TemporaryDirectory() as temp_dir:
    # Save processor
    processor.save_pretrained(save_directory=temp_dir, push_to_hub=False)
    # Load processor
    new_processor = OtherProcessor.from_pretrained(temp_dir)

assert processor.tokenizer1.__class__ != processor.tokenizer2.__class__ # passes
assert new_processor.tokenizer1.__class__ != new_processor.tokenizer2.__class__ # fails
```

### Expected behavior

You should be able to use multiple processors within a processor.

aijadugar:
# What does this PR do?

This PR fixes an issue where saving a custom Processor that includes multiple sub-tokenizers of the same type caused them to overwrite each other during serialization.

The root cause was that all sub-components were being saved using the same default filenames, leading to collisions.

This update introduces unique naming and loading logic in the ProcessorMixin save/load methods, allowing processors with multiple tokenizers to be safely saved and reloaded without data loss.

Fixes #41816

## Before submitting
I have read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request).

 The change was discussed in [issue #41816](https://github.com/huggingface/transformers/issues/41816).

 I’ve tested the processor save/load logic locally with multiple tokenizers.

 No documentation changes were required.

 Added/verified tests for multiple sub-tokenizers loading correctly.

## Who can review?

Tagging maintainers familiar with processor and tokenizer internals:

@CyrilVallez

@ArthurZucker


AmitMY:
Running the example from the issue, I get:
```py
  File "/Users/amitmoryossef/dev/sign/visual-text-decoder/example.py", line 32, in <module>
    processor.save_pretrained(save_directory=temp_dir, push_to_hub=False)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/visual_text_decoder/lib/python3.12/site-packages/transformers/processing_utils.py", line 804, in save_pretrained
    attribute.save_pretrained(attribute_save_dir, save_jinja_files=save_jinja_files)
                                                                   ^^^^^^^^^^^^^^^^
NameError: name 'save_jinja_files' is not defined
```

aijadugar:
hi, @AmitMY , can you try it once again!

AmitMY:
Strangely, now the error is
> AttributeError: GemmaTokenizerFast has no attribute to_dict

It would be good if you add a test for your code.

Create `tests/utils/test_processor_utils.py`

```py
import tempfile

from transformers.testing_utils import TestCasePlus

from transformers import ProcessorMixin, AutoTokenizer, PreTrainedTokenizer


class ProcessorSavePretrainedMultipleAttributes(TestCasePlus):
    def test_processor_loads_separate_attributes(self):
        class OtherProcessor(ProcessorMixin):
            name = "other-processor"

            attributes = [
                "tokenizer1",
                "tokenizer2",
            ]
            tokenizer1_class = "AutoTokenizer"
            tokenizer2_class = "AutoTokenizer"

            def __init__(self,
                         tokenizer1: PreTrainedTokenizer,
                         tokenizer2: PreTrainedTokenizer
                         ):
                super().__init__(tokenizer1=tokenizer1,
                                 tokenizer2=tokenizer2)

        tokenizer1 = AutoTokenizer.from_pretrained("google/gemma-3-270m")
        tokenizer2 = AutoTokenizer.from_pretrained("HuggingFaceTB/SmolLM2-1.7B")

        processor = OtherProcessor(tokenizer1=tokenizer1,
                                   tokenizer2=tokenizer2)
        assert processor.tokenizer1.__class__ != processor.tokenizer2.__class__

        with tempfile.TemporaryDirectory() as temp_dir:
            # Save processor
            processor.save_pretrained(save_directory=temp_dir, push_to_hub=False)
            # Load processor
            new_processor = OtherProcessor.from_pretrained(temp_dir)

        assert new_processor.tokenizer1.__class__ != new_processor.tokenizer2.__class__
```

src/transformers/processing_utils.py
```diff
@@ -654,30 +654,16 @@ def to_dict(self) -> dict[str, Any]:
...
-        if "chat_template" in output:
-            del output["chat_template"]
+        attrs_to_save = list(sig.parameters) + self.__class__.attributes + ["auto_map"]
+
+        for key in ["tokenizer", "qformer_tokenizer", "protein_tokenizer", "char_tokenizer", "chat_template"]:
+            output.pop(key, None)
 
         def save_public_processor_class(dictionary):
-            # make sure private name "_processor_class" is correctly
-            # saved as "processor_class"
```
AmitMY, line 680: i guess you should probably make the PR minimal - keep comments

src/transformers/processing_utils.py
```diff
@@ -687,33 +673,24 @@ def save_public_processor_class(dictionary):
             return dictionary
 
         def cast_array_to_list(dictionary):
-            """
-            Numpy arrays are not serialiazable but can be in pre-processing dicts.
-            This function casts arrays to list, recusring through the nested configs as well.
-            """
```
AmitMY, line 693: keep these too

src/transformers/processing_utils.py
```diff
@@ -687,33 +673,24 @@ def save_public_processor_class(dictionary):
...
             for key, value in dictionary.items():
                 if isinstance(value, np.ndarray):
                     dictionary[key] = value.tolist()
                 elif isinstance(value, dict):
                     dictionary[key] = cast_array_to_list(value)
             return dictionary
 
-        # Special case, add `audio_tokenizer` dict which points to model weights and path
```
AmitMY, line 701: and this

src/transformers/processing_utils.py
```diff
@@ -687,33 +673,24 @@ def save_public_processor_class(dictionary):
...
         if "audio_tokenizer" in output:
             audio_tokenizer_dict = {
                 "audio_tokenizer_class": self.audio_tokenizer.__class__.__name__,
                 "audio_tokenizer_name_or_path": self.audio_tokenizer.name_or_path,
             }
             output["audio_tokenizer"] = audio_tokenizer_dict
 
-        # Serialize attributes as a dict
```
AmitMY, line 709: and this

src/transformers/processing_utils.py
```diff
@@ -687,33 +673,24 @@ def save_public_processor_class(dictionary):
...
         output = {
             k: v.to_dict() if isinstance(v, PushToHubMixin) else v
             for k, v in output.items()
-            if (
-                k in attrs_to_save  # keep all attributes that have to be serialized
```
AmitMY, line 714: and revert this

AmitMY (changes requested):
Cool! now i guess need to make all the other tests pass...

src/transformers/processing_utils.py
AmitMY: you added the comment in the wrong place

AmitMY (changes requested):
look at the diff. ideally, don't change things that are not relevant

AmitMY:
```
W291 [*] Trailing whitespace
   --> src/transformers/processing_utils.py:658:38
    |
656 |         """
657 |         # shallow copy to avoid deepcopy errors
658 |         output = self.__dict__.copy()  
    |                                      ^^
659 |
660 |         # Get the kwargs in `__init__`.
    |
help: Remove trailing whitespace

F821 Undefined name `save_jinja_files`
   --> src/transformers/processing_utils.py:792:80
    |
790 |                 attribute_save_dir = os.path.join(save_directory, attribute_name)
791 |                 os.makedirs(attribute_save_dir, exist_ok=True)
792 |                 attribute.save_pretrained(attribute_save_dir, save_jinja_files=save_jinja_files)
    |                                                                                ^^^^^^^^^^^^^^^^
793 |             elif attribute._auto_class is not None:
794 |                 custom_object_save(attribute, save_directory, config=attribute)
    |

W291 [*] Trailing whitespace
    --> src/transformers/processing_utils.py:1423:68
     |
1421 |                 attribute_class = cls.get_possibly_dynamic_module(class_name)
1422 |
1423 |             # updated loading path for handling multiple tokenizers 
     |                                                                    ^
1424 |             attribute_path = os.path.join(pretrained_model_name_or_path, attribute_name)
1425 |             if os.path.isdir(attribute_path):
     |
help: Remove trailing whitespace

I001 [*] Import block is un-sorted or un-formatted
 --> tests/test_processor_utils.py:1:1
  |
1 | / import tempfile
2 | |
3 | | from transformers.testing_utils import TestCasePlus
4 | | from transformers import ProcessorMixin, AutoTokenizer, PreTrainedTokenizer
  | |___________________________________________________________________________^
  |
help: Organize imports

Found 4 errors.
[*] 3 fixable with the `--fix` option.
```

aijadugar:
hellow @AmitMY , thanks for your continuous feedback and detailed reports!...

AmitMY:
Tests still fail

src/transformers/processing_utils.py
```diff
@@ -748,7 +740,7 @@ def __repr__(self):
         attributes_repr = "\n".join(attributes_repr)
         return f"{self.__class__.__name__}:\n{attributes_repr}\n\n{self.to_json_string()}"
 
-    def save_pretrained(self, save_directory, push_to_hub: bool = False, **kwargs):
+    def save_pretrained(self, save_directory, save_jinja_files=False, push_to_hub: bool = False, **kwargs):
```
molbap, line 743: let's avoid adding an arg here for API consistency, and I don't think it's needed too

tests/test_processor_utils.py
molbap, line 1: does not need to be in an entirely new file, can be in `test_processing_common` and actually be part of the processor testing suite

tests/test_processor_utils.py
```diff
@@ -0,0 +1,37 @@
+import tempfile
+
+from transformers import AutoTokenizer, PreTrainedTokenizer, ProcessorMixin
+from transformers.testing_utils import TestCasePlus
+
+
+class ProcessorSavePretrainedMultipleAttributes(TestCasePlus):
```
molbap, line 7: The test case `test_processor_from_and_save_pretrained` should be modified to handle the multi-tokenizers case

src/transformers/processing_utils.py
```diff
@@ -792,9 +784,12 @@ def save_pretrained(self, save_directory, push_to_hub: bool = False, **kwargs):
...
 
-            # Save the tokenizer in its own vocab file. The other attributes are saved as part of `processor_config.json`
-            if attribute_name == "tokenizer":
-                attribute.save_pretrained(save_directory)
+            # if attribute is tokenizer, then save it in its own file for avoid overwriting
+            if hasattr(attribute, "save_pretrained"):
+                # use the attribute_name as prefix to create a unique file
+                attribute_save_dir = os.path.join(save_directory, attribute_name)
+                os.makedirs(attribute_save_dir, exist_ok=True)
+                attribute.save_pretrained(attribute_save_dir, save_jinja_files=save_jinja_files)
```
molbap, line 792: Indeed evolla already uses the subdir approach. however I'd reserve that for multi-tokenizers models only, checking with `len(tokenizer_attributes) > 1` and keeping the normal save path in other cases.

molbap (commented):
Took a look, thanks for the contrib! agree with @yonigozlan on the issue description that we are currently revamping a lot for v5 and need to be careful about all API changes so exercising caution.

yonigozlan (commented):
Hey @aijadugar , thanks for working on this. Indeed better support is needed for multiple tokenizers, however this PR radically changes how we save processors and subprocessors config files, as it would create subfolders for all sub processors.

In order to avoid having to save too many config files or subfolders, and centralize the information, we decided that all sub processors configs (except tokenizers) should be saved as nested dicts in the preprocessor_config.json files. This is not incompatible with having multiple subprocessors of the same types, although this is not yet supported.
As for tokenizers, the solution in this PR seems ok to me, except for the fact that the base "tokenizer" would also be saved in a separate folder, which would break backward compatibility with existing models on the hub, + given that the very large majority of models use only one tokenizer, it makes more sense to me to only create subfolders for additional tokenizers, and keep the base tokenizer files at the root of the checkpoint. 

The [refactoring PR](https://github.com/huggingface/transformers/pull/41633) adds support for multiple tokenizers by default ([logic is here](https://github.com/yonigozlan/transformers/blob/cc45a7ed5459369b90b10d9e719447784486b041/src/transformers/processing_utils.py#L825-L833)), and should be merged shortly. I'll try to also add support for multiple sub processors other than tokenizers in this PR or in a subsequent one.

aijadugar:
Thanks for the clarification @yonigozlan , I’ll wait for the v5 refactoring PR!

ArthurZucker:
v5 refactoring is done should be easier for you to update!

yonigozlan:
Closing this as completed with https://github.com/huggingface/transformers/pull/42630 and https://github.com/huggingface/transformers/pull/42667

PR #41843, opened 2025-10-24
Fix Qwen2Audio flash attention mask format for generation

Abdennacer-Badaoui:
## What does this PR fix?

This PR fixes the `test_eager_matches_fa2_generate` test failure for Qwen2Audio by using the `create_bidirectional_mask` utility function to properly handle attention masks across different attention implementations.

The Qwen2Audio model was manually creating a 4D attention mask with `-inf` values for the audio encoder, regardless of the attention implementation being used. This caused issues with Flash Attention 2/3, which requires a 2D boolean mask (shape `(batch_size, seq_len)`) with `1` for valid tokens and `0` for padding.

### DIFF
--- a/src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
+++ b/src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
@@ -775,13 +775,22 @@ class Qwen2AudioForConditionalGeneration(Qwen2AudioPreTrainedModel, GenerationMi
                 # Create mask
                 padding_mask = seq_range >= lengths_expand
 
-                audio_attention_mask_ = padding_mask.view(batch_size, 1, 1, max_seq_len).expand(
-                    batch_size, 1, max_seq_len, max_seq_len
-                )
-                audio_attention_mask = audio_attention_mask_.to(
-                    dtype=self.audio_tower.conv1.weight.dtype, device=self.audio_tower.conv1.weight.device
-                )
-                audio_attention_mask[audio_attention_mask_] = float("-inf")
+                # Flash attention requires a 2D boolean mask (batch_size, seq_len) where 1 means valid and 0 means padding
+                # Eager/SDPA attention requires a 4D mask with -inf for padding
+                if self.audio_tower.config._attn_implementation in ["flash_attention_2", "flash_attention_3"]:
+                    # For flash attention: 2D mask with 1 for valid tokens, 0 for padding
+                    audio_attention_mask = (~padding_mask).to(
+                        dtype=torch.int32, device=self.audio_tower.conv1.weight.device
+                    )
+                else:
+                    # For eager/SDPA: 4D mask with -inf for padding
+                    audio_attention_mask_ = padding_mask.view(batch_size, 1, 1, max_seq_len).expand(
+                        batch_size, 1, max_seq_len, max_seq_len
+                    )
+                    audio_attention_mask = audio_attention_mask_.to(
+                        dtype=self.audio_tower.conv1.weight.dtype, device=self.audio_tower.conv1.weight.device
+                    )
+                    audio_attention_mask[audio_attention_mask_] = float("-inf")
 
                 audio_outputs = self.audio_tower(input_features, attention_mask=audio_attention_mask)
                 selected_audio_feature = audio_outputs.last_hidden_state
--- a/src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
+++ b/src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
@@ -777,8 +777,9 @@ class Qwen2AudioForConditionalGeneration(Qwen2AudioPreTrainedModel, GenerationMi
                 padding_mask = seq_range >= lengths_expand
                 audio_attention_mask_2d = (~padding_mask).to(dtype=torch.long, device=audio_feat_lengths.device)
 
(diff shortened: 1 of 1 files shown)


src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
```diff
@@ -775,13 +775,22 @@ def forward(
...
-                audio_attention_mask_ = padding_mask.view(batch_size, 1, 1, max_seq_len).expand(
-                    batch_size, 1, max_seq_len, max_seq_len
-                )
-                audio_attention_mask = audio_attention_mask_.to(
-                    dtype=self.audio_tower.conv1.weight.dtype, device=self.audio_tower.conv1.weight.device
-                )
-                audio_attention_mask[audio_attention_mask_] = float("-inf")
+                # Flash attention requires a 2D boolean mask (batch_size, seq_len) where 1 means valid and 0 means padding
+                # Eager/SDPA attention requires a 4D mask with -inf for padding
+                if self.audio_tower.config._attn_implementation in ["flash_attention_2", "flash_attention_3"]:
```
vasqu: We have also kernels implementations for flash attention so rather do `"flash" in ...`

Abdennacer-Badaoui:
@remi-or ready for review 🙂

src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
```diff
@@ -775,13 +775,22 @@ def forward(
...
 
-                audio_attention_mask_ = padding_mask.view(batch_size, 1, 1, max_seq_len).expand(
-                    batch_size, 1, max_seq_len, max_seq_len
-                )
-                audio_attention_mask = audio_attention_mask_.to(
-                    dtype=self.audio_tower.conv1.weight.dtype, device=self.audio_tower.conv1.weight.device
-                )
-                audio_attention_mask[audio_attention_mask_] = float("-inf")
+                # Flash attention requires a 2D boolean mask (batch_size, seq_len) where 1 means valid and 0 means padding
+                # Eager/SDPA attention requires a 4D mask with -inf for padding
```
vasqu: Am I seeing it right that we have the correct (inverted) 2D mask?

Imo, it would be nicer if we could move away from manually creating these masks and use https://github.com/huggingface/transformers/blob/9c20660138830ca362533551ca978c27b48283a1/src/transformers/masking_utils.py#L896-L899 instead. It will need some respective (dummy) embedding tho!

vasqu: It's also used in Bert when you need an example.

vasqu (commented):
We should move away from this way of creating masks. It will make our lives are harder to maintain

src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
Abdennacer-Badaoui: Thanks, @vasqu. I’ll use `create_bidirectional_mask` instead. Indeed, it’s cleaner.

src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
```diff
@@ -774,14 +775,19 @@ def forward(
                 lengths_expand = audio_feat_lengths.unsqueeze(1).expand(batch_size, max_seq_len)
                 # Create mask
                 padding_mask = seq_range >= lengths_expand
+                audio_attention_mask_2d = (~padding_mask).to(dtype=torch.long, device=audio_feat_lengths.device)
 
-                audio_attention_mask_ = padding_mask.view(batch_size, 1, 1, max_seq_len).expand(
-                    batch_size, 1, max_seq_len, max_seq_len
+                dummy_embeds = torch.zeros(
+                    (batch_size, max_seq_len, self.audio_tower.config.d_model),
```
vasqu: We can make a hidden dim of 1 iirc. It's only used for it's shape + device. Can you check?

Abdennacer-Badaoui: That's true. good catch!

vasqu:
run-slow: qwen2_audio

vasqu (approved):
The nccl error is unrelated, known to fail atm 

LGTM

vasqu:
Can you push an empty commit? Cant merge with red CI

src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
```diff
@@ -774,14 +775,19 @@ def forward(
...
                 # Create mask
                 padding_mask = seq_range >= lengths_expand
+                audio_attention_mask_2d = (~padding_mask).to(dtype=torch.long, device=audio_feat_lengths.device)
 
-                audio_attention_mask_ = padding_mask.view(batch_size, 1, 1, max_seq_len).expand(
-                    batch_size, 1, max_seq_len, max_seq_len
+                dummy_embeds = torch.zeros(
+                    (batch_size, max_seq_len, 1),
+                    dtype=self.audio_tower.conv1.weight.dtype,
+                    device=self.audio_tower.conv1.weight.device,
```
vasqu: Ah sorry maybe one last nit: can we change the device/dtype here? Inputs_embeds should suffice?

vasqu:
Thx a lot :hugs:

PR #41848, opened 2025-10-24
More data in benchmarking

remi-or:
This PR changes the benchmarking workflow to gather more data per benchmark. cc. @McPatate I only changed one workflow file, not sure I did not miss others.
It also adds a check on the benchmark config to disable FA when it is not installed to avoid an exception.

### DIFF
--- a/benchmark_v2/framework/benchmark_config.py
+++ b/benchmark_v2/framework/benchmark_config.py
@@ -3,6 +3,8 @@ import json
 import logging
 from typing import Any
 
+from transformers.utils.import_utils import is_flash_attn_2_available
+
 
 KERNELIZATION_AVAILABLE = False
 try:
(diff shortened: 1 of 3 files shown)


benchmark_v2/framework/benchmark_config.py
```diff
@@ -18,6 +20,16 @@
...
 
+    all_attn_implementations = [
+        ("flash_attention_2", None),
+        ("eager", None),
+        ("sdpa", "math"),
+        ("sdpa", "flash_attention"),
+        ("flex_attention", None),
+    ]
+
+    all_compiled_modes = [None, "default", "reduce-overhead", "max-autotune", "max-autotune-no-cudagraphs"]
```
McPatate, line 31: This could live outside the code in a configuration file:
```yml
attn_impl:
  - flash_attention_2:
  - eager:
  - sdpa:
    - "math"
    - "flash_attention"
  - flex_attention:

compiled_modes:
  - null
  - "default"
  - "reduce-overhead"
  - "max-autotune"
  - "max-autotune-no-cudagraphs"
```

remi-or: It could, but I would be reluctant to this for a few reasons:
- `transformers` has a "one model, one file" policy which if we extrapolate means having things self contained, so I dont see the reason to create an additional config file with only this
- we would have to open and read from the file when instantiating the class, while here the list is already an attribute, which is practical
- makes the code less readable / more fragmented

McPatate: was just an idea, though I don't necessarily agree with your points, we can ignore this

for now

McPatate (approved):

PR #41849, opened 2025-10-24
Allow parse_response to accept token IDs

Rocketknight1:
While I was working on `parse_response` I noticed that the UX was annoying in places because I had to keep decoding and then calling `parse_response`, which added a lot of boilerplate. Making `parse_response` optionally able to handle token IDs directly cut down on that a lot, so I figured I'd fix it now while there's still time before release!


gante (approved):
LGTM 👍

PR #41850, opened 2025-10-24
speed up loading checkpoints for zero stage 3

ri938:
Loading checkpoints for model Qwen/Qwen3-Next-80B-A3B-Instruct was very slow

this change brought down checkpoint loading times from 10mins+ to about 1.5mins

the load method is called over 4 million times (123,799 per layer and there are 41 layers). Looping through the state dict (size ~1500) was therefore very slow. This change speed the loading method by avoiding looping through the state dict.

Flagged for review
- model loading (from pretrained, etc): @CyrilVallez
- distributed: @3outeille @ArthurZucker
- Big Model Inference: @SunMarc

### DIFF
--- a/src/transformers/integrations/peft.py
+++ b/src/transformers/integrations/peft.py
@@ -218,13 +218,10 @@ class PeftAdapterMixin:
             token = adapter_kwargs.pop("token")
 
         if peft_config is None:
-            new_args = adapter_kwargs.copy()
-            if 'token' in new_args:
-                new_args.pop('token')
             adapter_config_file = find_adapter_config_file(
                 peft_model_id,
                 token=token,
-                **new_args,
+                **adapter_kwargs,
             )
 
             if adapter_config_file is None:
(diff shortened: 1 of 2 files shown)


3outeille:
taking a look

src/transformers/integrations/peft.py
```diff
@@ -218,10 +218,13 @@ def load_adapter(
             token = adapter_kwargs.pop("token")
 
         if peft_config is None:
+            new_args = adapter_kwargs.copy()
+            if 'token' in new_args:
+                new_args.pop('token')
```
3outeille: why do we need to pop again? isnt it handled already by lines 217 & 218 ?

ri938: oh sorry this code is not meant to be in this merge request (this was just a hack to fix model loading failing on master branch)

ri938:
Removed accidenal addition of some other temporary code. Merge request should be good to go now.

3outeille:
lgtm, need approval from core maintainers @ArthurZucker @Cyrilvallez

ArthurZucker (approved):
GG

PR #41852, opened 2025-10-24
[`Attn Masks`] Non-vmap default for attention masks

### LINKED ISSUE #41639: MarianMTModel performance regression due to Bidirectional masks

vasqu:
Non-vmap creation of masks. These work with all our base masks and we only default back to vmap when using patterns we cannot guarantee (i.e. additional and/or masks).

Note:
- Non-vmap works with every mask that has anything index based
- Merged old/new sdpa under one function --> easier maintenance imo
- Executorch does not need an additional masking fn anymore
- Lifts some restrictions on older torch versions, e.g. chunked attn with padding, packed attn masks etc

Fixes #41639

cc @jiqing-feng @IlyasMoutawwakil

### DIFF
--- a/src/transformers/integrations/executorch.py
+++ b/src/transformers/integrations/executorch.py
@@ -229,10 +229,10 @@ class TorchExportableModuleForDecoderOnlyLM(torch.nn.Module):
                 "Using `StaticCache` for export as `layer_types` is not specified or `sliding_window` is `null` in the config."
             )
             self.model = TorchExportableModuleWithStaticCache(model, batch_size, max_cache_len, device)
-        # This is the same as sdpa, but mask creation does not use `vmap` which is not exportable
+        """# This is the same as sdpa, but mask creation does not use `vmap` which is not exportable
         ALL_MASK_ATTENTION_FUNCTIONS.register("sdpa_without_vmap", sdpa_mask_without_vmap)
         ALL_ATTENTION_FUNCTIONS.register("sdpa_without_vmap", ALL_ATTENTION_FUNCTIONS["sdpa"])
-        self.model.model.config._attn_implementation = "sdpa_without_vmap"
+        self.model.model.config._attn_implementation = "sdpa_without_vmap"""
 
     def forward(
         self,
@@ -768,10 +768,10 @@ def convert_and_export_with_cache(
 
     import torch.export._trace
 
-    # This is the same as sdpa, but mask creation does not use `vmap` which is not exportable
-    ALL_MASK_ATTENTION_FUNCTIONS.register("sdpa_without_vmap", sdpa_mask_without_vmap)
-    ALL_ATTENTION_FUNCTIONS.register("sdpa_without_vmap", ALL_ATTENTION_FUNCTIONS["sdpa"])
-    model.config._attn_implementation = "sdpa_without_vmap"
+    """# This is the same as sdpa, but mask creation does not use `vmap` which is not exportable
+        ALL_MASK_ATTENTION_FUNCTIONS.register("sdpa_without_vmap", sdpa_mask_without_vmap)
+        ALL_ATTENTION_FUNCTIONS.register("sdpa_without_vmap", ALL_ATTENTION_FUNCTIONS["sdpa"])
+        self.model.model.config._attn_implementation = "sdpa_without_vmap"""
 
     with torch.no_grad():
         # TODO: The default inputs only work for text models. We need to add support for vision/audio models.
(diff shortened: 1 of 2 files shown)


src/transformers/integrations/executorch.py
```diff
@@ -1109,92 +1088,3 @@ def _unflatten_dynamic_cache(values, context: torch.utils._pytree.Context):
         value = value_list[idx] if idx < len(value_list) else None
         cache.update(key, value, idx)
     return cache
-
-
-def sdpa_mask_without_vmap(
```
vasqu, line 1114: No longer needed as vmap was the reason we needed this workaround in the first place

src/transformers/masking_utils.py
```diff
@@ -82,8 +82,10 @@ def causal_mask_function(batch_idx: int, head_idx: int, q_idx: int, kv_idx: int)
 def bidirectional_mask_function(batch_idx: int, head_idx: int, q_idx: int, kv_idx: int) -> bool:
     """
     This creates a full bidirectional mask.
+
+    NOTE: It is important to keep an index-based version for non-vmap expansion.
     """
-    return q_idx.new_ones((), dtype=torch.bool)
+    return q_idx >= 0
```
vasqu, line 88: As noted above, for non-vmap we need this as index based version

src/transformers/masking_utils.py
```diff
@@ -402,128 +428,58 @@ def sdpa_mask_recent_torch(
...
+        # We don't need to add an offset to the mask_function either, as we vmap directly the correct indices for k and kv indices
+        with TransformGetItemToIndex():
+            attention_mask = _vmap_for_bhqkv(mask_function)(batch_arange, head_arange, cache_position, kv_arange)
+
+    # Option 3: Limited vmap mask creation (torch<2.6 and custom patterns)
+    else:
+        logger.warning_once(
+            "Using vmap mask creation (custom patterns with and/or masks) has limited capabilities under "
+            "`torch<2.6.0`. We cannot guarantee that the correct mask is constructed."
+        )
```
vasqu: We shouldn't use vmaps on older torch versions, we still attempt it but warn the users as we cannot guarantee it to work as expected. It all depends on the custom fn they have passed along

src/transformers/masking_utils.py
```diff
@@ -402,128 +428,58 @@ def sdpa_mask_recent_torch(
...
-        causal_mask = causal_mask * padding_mask[:, None, None, :]
+            attention_mask = attention_mask * padding_mask
 
     # Due to a bug in versions of torch<2.5, we need to update the mask in case a query is not attending to any
     # tokens (due to padding). See details in https://github.com/pytorch/pytorch/issues/110213
     if not _is_torch_greater_or_equal_than_2_5 and allow_torch_fix:
-        causal_mask |= torch.all(~causal_mask, dim=-1, keepdim=True)
-    return causal_mask
-
+        attention_mask = attention_mask | torch.all(~attention_mask, dim=-1, keepdim=True)
```
vasqu, line 435: I encountered issues with the inplace version where we'd need a clone (e.g. when using swa). This is safer

ArthurZucker (approved):
can we add a test to default test no graph break on this?

jiqing-feng:
Hi @vasqu . Anything blocks merge?

Cyrilvallez (approved):
Very very nice, this solves quite a lot of different issues at the same time! I'm very happy to avoid special handling for export! Very clever use of broadcasting from the optimum team, I did not know we could simply do such things! Thanks a lot for upstreaming directly to us.

Do you mind expanding a bit more on what are the limitations of the broadcasting approach here for posterity? Is it only index-based operations as you mention on the comments, or are there more subtle things?

IlyasMoutawwakil:
> Do you mind expanding a bit more on what are the limitations of the broadcasting approach here for posterity? Is it only index-based operations as you mention on the comments, or are there more subtle things?

Nothing I'm aware of, the only condition is to write the mask_function as a comparison between the indexes (and constants).
One example of this is the bidirectional_mask_function which was written as `q_idx.new_ones((), dtype=torch.bool)` simply returns a scalar 1 (true) and @vasqu rewrote it as `q_idx >= 0` (always true). I think any mask function can be written as f(b, h, q, kv) but I can't prove it 😂

vasqu:
Merging this then! Let's see what crazy masks come up in the future; for now the "mask hypothesis" holds :laughing:

vasqu:
@jiqing-feng was mainly blocked by me being out last week ;)

PR #41871, opened 2025-10-26
Fix default image_rows and image_cols initialization in Idefics3 and SmolVLM processors

MilkClouds:
# What does this PR do?

This PR fixes a bug in the default initialization of `image_rows` and `image_cols` in the Idefics3 and SmolVLM processors.

## Problem

The original code incorrectly used `len(text)` to determine the size of the default `image_rows` and `image_cols` lists:
```python
image_rows = inputs.pop("rows", [[0] * len(text)])
image_cols = inputs.pop("cols", [[0] * len(text)])
```

This creates a single list with length equal to the number of text samples in the batch, which is incorrect. The correct behavior should create a list for each sample, where each inner list has length equal to the number of image tokens in that specific sample.

## Solution

Changed the default initialization to count image tokens per sample:
```python
image_rows = inputs.pop("rows", [[0] * sample.count(self.image_token) for sample in text])
image_cols = inputs.pop("cols", [[0] * sample.count(self.image_token) for sample in text])
```

This ensures that:
1. Each sample in the batch gets its own list
2. Each list has the correct length matching the number of image tokens in that sample

## Impact

This fix affects:
- `Idefics3Processor` in `src/transformers/models/idefics3/processing_idefics3.py`
- `SmolVLMProcessor` in `src/transformers/models/smolvlm/processing_smolvlm.py`

The bug would have caused issues when processing batches with multiple samples or samples with multiple images, as the default values would have incorrect dimensions.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@zucchini-nlp (multimodal models)

### DIFF
--- a/src/transformers/models/idefics3/processing_idefics3.py
+++ b/src/transformers/models/idefics3/processing_idefics3.py
@@ -286,8 +286,8 @@ class Idefics3Processor(ProcessorMixin):
                         f"The number of images in the text {n_images_in_text} and images {n_images_in_images} should be the same."
                     )
 
-                image_rows = inputs.pop("rows", [[0] * len(text)])
-                image_cols = inputs.pop("cols", [[0] * len(text)])
+                image_rows = inputs.pop("rows", [[0] * sample.count(self.image_token) for sample in text])
+                image_cols = inputs.pop("cols", [[0] * sample.count(self.image_token) for sample in text])
 
                 fake_image_token = self.fake_image_token
                 image_token = self.image_token
--- a/src/transformers/models/idefics3/processing_idefics3.py
+++ b/src/transformers/models/idefics3/processing_idefics3.py
@@ -286,8 +286,8 @@ class Idefics3Processor(ProcessorMixin):
                         f"The number of images in the text {n_images_in_text} and images {n_images_in_images} should be the same."
                     )
 
-                image_rows = inputs.pop("rows", [[0] * sample.count(self.image_token) for sample in text])
-                image_cols = inputs.pop("cols", [[0] * sample.count(self.image_token) for sample in text])
+                image_rows = inputs.pop("rows", [[0] * n_images for n_images in n_images_in_text])
+                image_cols = inputs.pop("cols", [[0] * n_images for n_images in n_images_in_text])
 
                 fake_image_token = self.fake_image_token
                 image_token = self.image_token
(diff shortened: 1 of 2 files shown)


MilkClouds:
p.s. this issue seems to be ignored before because `return_row_col_info` was always given to True to ImageProcessor! I've met error raised when I used non-`SmolVLMImageProcessor` for image processor of SmolVLMProcesor.

MilkClouds:
For reviewer's information: this PR is ready for review!

src/transformers/models/idefics3/processing_idefics3.py
```diff
@@ -286,8 +286,8 @@ def __call__(
                         f"The number of images in the text {n_images_in_text} and images {n_images_in_images} should be the same."
                     )
 
-                image_rows = inputs.pop("rows", [[0] * len(text)])
-                image_cols = inputs.pop("cols", [[0] * len(text)])
+                image_rows = inputs.pop("rows", [[0] * sample.count(self.image_token) for sample in text])
+                image_cols = inputs.pop("cols", [[0] * sample.count(self.image_token) for sample in text])
```
yonigozlan: You can use `n_images_in_text` defined above here

src/transformers/models/smolvlm/processing_smolvlm.py
```diff
@@ -172,8 +172,12 @@ def __init__(
...
     def expand_text_with_image_tokens(self, text, image_rows, image_cols):
         prompt_strings = []
-        image_rows = image_rows if image_rows is not None else [[0] * len(text)]
-        image_cols = image_cols if image_cols is not None else [[0] * len(text)]
+        image_rows = (
+            image_rows if image_rows is not None else [[0] * sample.count(self.image_token) for sample in text]
+        )
+        image_cols = (
+            image_cols if image_cols is not None else [[0] * sample.count(self.image_token) for sample in text]
+        )
```
yonigozlan: Same here, you can define image_rows and image_cols before the call to expand_text_with_image_tokens using n_images_in_text

yonigozlan (approved):
Good catch! Thanks for fixing. Suggested a minor improvement, but overall LGTM!

src/transformers/models/idefics3/processing_idefics3.py
MilkClouds: I changed the lines as you said!

src/transformers/models/smolvlm/processing_smolvlm.py
MilkClouds: Same here, I changed the lines as you said

yonigozlan:
Perfect thanks! I'll merge once the CI passes

PR #41880, opened 2025-10-27
Indonesian Language Support for ReadMe

derrickchen03:
Added support for Indonesian for README in the i18n folder, and links it to the other translated versions as well as the original one.

### DIFF
--- a/i18n/README_id.md
+++ b/i18n/README_id.md
@@ -1,0 +1,263 @@
+<!---
+Copyright 2020 The HuggingFace Team. All rights reserved.
+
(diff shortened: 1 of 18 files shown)


Rocketknight1:
cc @stevhliu

i18n/README_id.md
```diff
@@ -0,0 +1,263 @@
+<!---
```
stevhliu, line 1: It looks like this was translated from an older version of the `README.md` (the main version is [here](https://github.com/huggingface/transformers/blob/main/README.md)).

PR #41882, opened 2025-10-27
Support fdma for models with attention bias

LoserCheems:
# What does this PR do?

Fixes #41465

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@MekkCyber @drbh

### DIFF
--- a/src/transformers/integrations/flash_dynamic_mask_attention.py
+++ b/src/transformers/integrations/flash_dynamic_mask_attention.py
@@ -1,0 +1,83 @@
+from typing import Optional
+
+import torch
+
+from ..modeling_flash_dynamic_mask_attention_utils import _flash_dynamic_mask_attention_forward
+from ..utils import logging
+
+
+logger = logging.get_logger(__name__)
+
+
+def get_target_dtype(query: torch.Tensor, module: torch.nn.Module) -> torch.dtype:
+    """If the query is in float32, return a target dtype compatible with flash dynamic mask attention. Return None otherwise."""
+    if query.dtype == torch.float32:
+        if torch.is_autocast_enabled():
+            return torch.get_autocast_gpu_dtype()
+        # Handle the case where the model is quantized
+        elif hasattr(module.config, "_pre_quantization_dtype"):
+            return module.config._pre_quantization_dtype
+        else:
+            return next(layer for layer in module.modules() if isinstance(layer, torch.nn.Linear)).weight.dtype
+    return None
+
+
+def flash_dynamic_mask_attention_forward(
+    module: torch.nn.Module,
+    query: torch.Tensor,
+    key: torch.Tensor,
+    value: torch.Tensor,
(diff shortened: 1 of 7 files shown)


src/transformers/testing_utils.py
```diff
@@ -599,6 +600,27 @@ def require_flash_attn(test_case):
     return unittest.skipUnless(kernels_available | flash_attn_available, "test requires Flash Attention")(test_case)
 
 
+def require_flash_dmattn(test_case):
```
vasqu, line 603: The idea is to not use a separate package but to integrate it into kernels, similar to https://huggingface.co/kernels-community/flash-attn3

src/transformers/modeling_utils.py
vasqu, line 1: This logic could all be avoided if it was integrated into kernels.

src/transformers/modeling_flash_dynamic_mask_attention_utils.py
vasqu, line 1: Same here, we would handle most of this. Maybe small modifications would be needed to support the `attention_bias` parameter in https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_flash_attention_utils.py

src/transformers/integrations/flash_dynamic_mask_attention.py
vasqu, line 1: Same we will handle this then

vasqu (commented):
There is a lot of duplicated effort as it's mostly the same as our fa implementation. 

The idea is to wrap this within the kernels package instead (cc @MekkCyber @drbh) and let us handle all the forwards, wrapping etc as we already support other FA implementations that way. The only modifications that we would need then, would be to make sure that the parameters match, i.e. `s_aux` <-> `attention_bias` 

Edit: meaning kernels as in https://github.com/huggingface/kernel-builder

LoserCheems:
hi @vasqu, is there and referenceable process tutorial available

vasqu:
There is a blog at https://huggingface.co/blog/kernel-builder but I would wait for @MekkCyber @drbh as they are more familiar there and can provide proper guidance.

MekkCyber (commented):
Thanks for all this work @LoserCheems ! Though as @vasqu suggested it's much better to use fdma via `kernel-builder`, you can just follow the [blogpost](https://huggingface.co/blog/kernel-builder), build the kernel, upload the builds to the hub, and then start using the exported `torch.ops` from there

LoserCheems:
Thank you @MekkCyber for your guidance, but I have encoumtered some issues regarding building the wheel:  https://github.com/huggingface/kernel-builder/issues/283

jordane95:
Hi @LoserCheems , any progress here?

LoserCheems:
> Hi @LoserCheems , any progress here?

I'm very sorry. I haven't been able to proceed smoothly with the kernel-builder process. I think I need help...

PR #41888, opened 2025-10-27
Fix torch.no_grad decorator in VLMS

yaswanth19:
As per the title


vasqu (approved):
Thx a lot, merging both in a second

PR #41889, opened 2025-10-27
🚨 [v5][PEFT] Bump min version requirement of PEFT to  0.18.0

BenjaminBossan:
# What does this PR do?

PEFT is an optional dependency of transformers with a min version of 0.5.0. However, starting with transformers v5, older PEFT versions will not work anymore (see #41406). The minimum PEFT version will be 0.18.0.

This PR updates the PEFT integration to require PEFT 0.18.0. This allows us to eliminate some obsolete checks and tests that were required for backwards compatibility.

Note: 

- PEFT 0.18.0 is not yet released, so **don't merge** this PR yet. However, we will release it once we have tested it with the transformers v5 release candidate, before the transformers v5 released.
- If this commit is not included in the transformers v5 release, users who want to use PEFT will get an `ImportError` without indication what's wrong. Therefore, I would highly recommend to merge this before the v5 release.

## Before submitting
      Pull Request section?
      to it if that's the case.
- [x] Did you write any new necessary tests? <= Tests updated


src/transformers/integrations/peft.py
```diff
@@ -450,9 +426,6 @@ def active_adapters(self) -> list[str]:
         """
         check_peft_version(min_version=MIN_PEFT_VERSION)
 
-        if not is_peft_available():
```
BenjaminBossan, line 592: I deleted this check because `check_peft_version` already checks if PEFT is available, meaning it is redundant. This change is not strictly related to the min version bump.

githubnemo (approved):
Looks good from my side

BenjaminBossan:
This PR should be included in the v5 release. Not sure who can best review it, ping @ydshieh. The failing tests seem to be unrelated to this PR.

githubnemo (approved):

ArthurZucker (commented):
Happy to merge,  wondering why we don't update the `dependency_table` and `setup.py` ? 🤗

BenjaminBossan:
> Happy to merge, wondering why we don't update the `dependency_table` and `setup.py` ? 🤗

PEFT is an optional dependency, as such there is no entry there.

ydshieh:
> Happy to merge, wondering why we don't update the `dependency_table` and `setup.py` ? 🤗

You can still put it in those files. Most of them are optional dependency, even `torch` .

The benefits is that an user could do something like `pip install transformers[dev]` or even `pip install transformers[peft]` (if you want such entries).

But it's just minor, up to you.

BenjaminBossan:
@ydshieh I didn't know that, I added PEFT to that list. The failing CI looks unrelated.

BenjaminBossan:
@ArthurZucker I updated the PR, could you please give a final review?

BenjaminBossan:
Merged the latest main, the failing test looks unrelated to this PR.

ArthurZucker (approved):
Happy to merge 🤗

PR #41894, opened 2025-10-27
Update transformers to support `FqnToConfig`

jcaip:
Summary:

This PR adds in support for `FqnToConfig` to transformers, which was added to torchao earlier in [this](https://github.com/pytorch/ao/pull/3083) PR. 

`FqnToConfig` adds support for quantizing parameters that are not called "weight". Previously torchao was linear only, so assuming "weight" was fine, but this is insufficient for quantizing moe models like llama4, as they register their parameters under different names (`gate_up_proj / down_proj`). 

Test Plan:
```
pytest tests/quantization/torchao_integration/test_torchao.py -k fqn_to_config
```

Reviewers:

Subscribers:

Tasks:

Tags:

# What does this PR do?

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -245,13 +245,24 @@ def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **
...
-            # we only quantize the weight of nn.Linear and nn.Embedding
-            module, tensor_name = get_module_from_name(model, param_name)
-            _QUANTIZABLE = [torch.nn.Linear]
-            if self.quantization_config.include_input_output_embeddings:
-                _QUANTIZABLE.append(torch.nn.Embedding)
-            return isinstance(module, tuple(_QUANTIZABLE)) and tensor_name == "weight"
+        # Handle FqnToConfig, introduced in torchao 0.15.0+
+        elif (self.quantization_config._get_ao_version() >= version.parse("0.15.0")):
+            from torchao.quantization import FqnToConfig
+            from torchao.quantization.quant_api import _filter_fn_and_param_in_fqn_config
```
vkuzo: can we avoid using this private function?  Or, if you need it here, it should be a part of the public API, without an underscore

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -245,13 +245,24 @@ def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **
...
+
+            if isinstance(self.quantization_config.quant_type, FqnToConfig):
+                module, tensor_name = get_module_from_name(model, param_name)
+                module_fqn, param_name_fqn = param_name.rsplit(".", 1)
+                if _filter_fn_and_param_in_fqn_config(module, module_fqn, self.quantization_config.quant_type, None):
```
vkuzo: seems like this isn't passing in a user defined `filter_fn`, so IMO let's remove `_filter_fn_and_param_in_fqn_config` from here to simplify?

Rocketknight1:

jcaip:
cc @SunMarc as well

jcaip:
cc @MekkCyber @SunMarc friendly ping on review :)

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -319,8 +336,53 @@ def create_quantized_param(
...
+                    elif module_fqn in config.module_fqn_to_config:
+                        assert not module_fqn.startswith("re:"), (
+                            "module fqn should not start with`re:`, which is used for specifying regex"
+                        )
+                        c = config.module_fqn_to_config[module_fqn]
+                    # regx match module and param
+                    else:
+                        for maybe_module_fqn_pattern in config.module_fqn_to_config:
+                            if not maybe_module_fqn_pattern.startswith("re:"):
+                                continue
```
MekkCyber, line 362: I think we should raise an error instead of silently skipping, wdyt ?

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -319,8 +336,53 @@ def create_quantized_param(
...
+                                # we'll apply the config for first fully matched pattern
+                                c = config.module_fqn_to_config[maybe_module_fqn_pattern]
+                                break
+                        else:
+                            c = config.module_fqn_to_config.get("_default", None)
+
+                    if c is not None:
+                        if param_val == "weight":
+                            # we can apply the module config directly
+                            quantize_(module, c, lambda x, fqn: True)
```
MekkCyber: small nit: Maybe we can add parentheses around the lambda function to avoid confusion with the comma
```suggestion
                            quantize_(module, c, (lambda x, fqn: True))
```

tests/quantization/torchao_integration/test_torchao.py
```diff
@@ -378,6 +380,154 @@ def test_module_fqn_to_config_regex_precedence(self):
...
+                "_default": linear1_config,
+            }
+        )
+        quant_config = TorchAoConfig(quant_type=config)
+        quantized_model = AutoModelForCausalLM.from_pretrained(
+            "unsloth/Llama-4-Scout-17B-16E-Instruct",
+            device_map="auto",
+            dtype=torch.bfloat16,
+            quantization_config=quant_config,
+        )
```
MekkCyber, line 523: could you please use a smaller model ? or the same model we use across all the tests

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -319,8 +336,53 @@ def create_quantized_param(
...
+                            "param fqn should not start with`re:`, which is used for specifying regex"
+                        )
+                        c = config.module_fqn_to_config[param_name]
+                    elif module_fqn in config.module_fqn_to_config:
+                        assert not module_fqn.startswith("re:"), (
+                            "module fqn should not start with`re:`, which is used for specifying regex"
+                        )
+                        c = config.module_fqn_to_config[module_fqn]
+                    # regx match module and param
+                    else:
```
MekkCyber:
```suggestion
                    # regex match module and param
                    else:
```

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -319,8 +336,53 @@ def create_quantized_param(
...
+                            elif re.fullmatch(maybe_module_fqn_pattern[3:], module_fqn):
+                                # we'll apply the config for first fully matched pattern
+                                c = config.module_fqn_to_config[maybe_module_fqn_pattern]
+                                break
+                        else:
+                            c = config.module_fqn_to_config.get("_default", None)
+
+                    if c is not None:
+                        if param_val == "weight":
+                            # we can apply the module config directly
```
MekkCyber, line 376: small nit: `param_val` is a bit misleading here as a name, maybe we can rename it to `parameter_name` or something else

MekkCyber (commented):
Thanks @jcaip for this integration ! It looks great, I just have some few nits 🤗

src/transformers/quantizers/quantizer_torchao.py
jcaip: This is a check to make sure that we only match regex keys, which start with `re`, erroring would be incorrect here imo. 

For example, if we have `exact_fqn` and `re:regex_fqn` in the config, we don't want to error, just want to ignore `exact_fqn` when regex matching. We can't do this all in one loop because we want to establish a precedence of `exact_fqn` > `re:regex_fqn`, so the for loop processing `exact_fqn` happens before this loop

I can add a comment though so this is more clear

tests/quantization/torchao_integration/test_torchao.py
jcaip: I want to test non-weight parameters for completeness here, but the only models of that format are very large (qwen-3-vl, gpt-oss, etc).

Let me update the test so it only loads a single moe layer and test on that instead.

jcaip: Actually, we need to test specifying quantized_config in `from_pretrained`... let me see if I can push a custom model that's only a single MoE layer for testing instead.

src/transformers/quantizers/quantizer_torchao.py
MekkCyber: Yes ! that would be great thank you

MekkCyber (approved):
Thank you @jcaip, awesome ! lgtm

SunMarc (approved):
LGTM !

PR #41900, opened 2025-10-27
Remove unnecessary slicing in sdpa_attention_forward

justinchuby:
The slicing in sdpa_attention_forward was there only because some masks were not constructed correctly (I was told). When the key size is dynamic, the slice op also prevents torch.export from correctly reasoning about its size.

cc @vasqu

### DIFF
--- a/src/transformers/integrations/sdpa_attention.py
+++ b/src/transformers/integrations/sdpa_attention.py
@@ -63,9 +63,6 @@ def sdpa_attention_forward(
         else:
             sdpa_kwargs = {"enable_gqa": True}
 
-    if attention_mask is not None and attention_mask.ndim == 4:
-        attention_mask = attention_mask[:, :, :, : key.shape[-2]]
-
     # Instead of relying on the value set in the module directly, we use the is_causal passed in kwargs if it is presented
     is_causal = is_causal if is_causal is not None else getattr(module, "is_causal", True)
 


justinchuby:
@Cyrilvallez Looks like this change passes the CI.

justinchuby:
@vasqu @Cyrilvallez any thoughts? Thanks. This is an important fix we hope to include in the 5.0 release

vasqu:
Responded in https://github.com/huggingface/transformers/pull/41559#discussion_r2468914497

But I'm pro this, we might wanna check some important models with slow run. Let's wait for Cyril for a final decision

Cyrilvallez:
Sorry for the delay, I was off as @vasqu mentioned! Still very relevant, would be very happy to finally remove this (and in other attn functions as well, such as the eager ones but I can take care of it myself later no worries)

cc @ydshieh, could you run a more extensive CI run on this PR and tell us whether you see any new failures, especially on older models? I don't have much time to do it manually myself as I need to catch up on all reviews 🤓 Just a bit scared that the fast tests may not be enough on this one!

ydshieh:
For sure, thank you for the ping. I will report back today or tomorrow.

ydshieh:
run-slow: bert, gpt2, t5, modernbert, vit, clip, detr, table_transformer, got_ocr2, whisper, wav2vec2, qwen2_audio, speech_t5, csm, llama, gemma3, qwen2, mistral3, qwen2_5_vl, llava, smolvlm, internvl, gemma3n, gpt_oss, qwen2_5_omni

ydshieh:
[Update] Looks good!



~~The report is good, but let me take a look deeper inside the workflow run to make sure.~~

Cyrilvallez:
Alright, amazing @ydshieh, thanks! Merging then! Thanks again @justinchuby for pushing on something we wanted since a long time!

Cyrilvallez (approved):

PR #41904, opened 2025-10-27
Fix inaccurate eval and train loss computation with variable batch sizes

### LINKED ISSUE #41898: inaccurate eval loss computation
https://github.com/huggingface/transformers/blob/1f0b490a2c42eb129dccc69031ccb537058689c4/examples/pytorch/language-modeling/run_clm_no_trainer.py#L657 The last batch may be smaller, since `drop_last` default to `False` for `DataLoader` construction. Thus,  computing the mean over the losses is inaccurate.
Same issue for the train loss:
https://github.com/huggingface/transformers/blob/1f0b490a2c42eb129dccc69031ccb537058689c4/examples/pytorch/language-modeling/run_clm_no_trainer.py#L673

jameslovespancakes:
Fixes #41898

When drop_last=False (default), the last batch may contain fewer samples than per_device_eval_batch_size. Using a fixed batch_size to repeat the scalar loss causes the last batch to be over-represented in the final average loss calculation.

Changes:
- Trainer: Use observed_batch_size instead of fixed batch_size when repeating eval loss for gather_for_metrics
- no_trainer examples: Use actual batch size from input_ids.shape[0] for both eval and train loss computation
- Train loss: Weight by actual batch size and divide by total samples instead of number of batches

This ensures accurate loss computation regardless of batch size variability while maintaining backward compatibility (identical behavior when all batches are uniform size).

### DIFF
--- a/examples/pytorch/language-modeling/run_clm_no_trainer.py
+++ b/examples/pytorch/language-modeling/run_clm_no_trainer.py
@@ -616,6 +616,7 @@ def main():
         model.train()
         if args.with_tracking:
             total_loss = 0
+            total_samples = 0
         if args.resume_from_checkpoint and epoch == starting_epoch and resume_step is not None:
             # We skip the first `n` batches in the dataloader when resuming from a checkpoint
             active_dataloader = accelerator.skip_first_batches(train_dataloader, resume_step)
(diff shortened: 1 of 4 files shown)


Rocketknight1:
The updates to the `no_trainer` examples look okay, but I'd like @sunmarc's confirmation about the change in `trainer.py`!

PR #40287, opened 2025-08-19
No more `natten`

ydshieh:
# What does this PR do?

This is required by `nat` (already depreciated) and `dinat`.

We have to pin it to a very old version to avoid `natten2dav` issue, but now we get 

> ImportError: cannot import name '_device_t' from 'torch.cuda' (/usr/local/lib/python3.10/dist-packages/torch/cuda/__init__.py)

with torch 2.8, which fails the job of `tests/utils` where `from transformers import *` is used.

Let's simply remove this natten from docker CI image (and no tests for `dinat` for which I think reasonable).

PR #41905, opened 2025-10-28
upgrade natten to 0.21 version

kaixuanliu:


src/transformers/models/dinat/modeling_dinat.py
```diff
@@ -37,7 +37,100 @@
 
 
 if is_natten_available():
-    from natten.functional import natten2dav, natten2dqkrpb
+    def natten2dqkrpb(query, key, rpb, kernel_size, dilation):
```
yao-matrix: use `na2d`?

src/transformers/models/dinat/modeling_dinat.py
```diff
@@ -37,7 +37,100 @@
...
+                            score = score + bias.unsqueeze(0)  # [batch, heads]
+
+                            attention_scores[:, :, i, j, idx] = score
+                        else:
+                            attention_scores[:, :, i, j, idx] = float("-inf")
+                        idx += 1
+
+        return attention_scores
+
+    def natten2dav(attention_probs, value, kernel_size, dilation):
```
yao-matrix: use `na2d`?

src/transformers/models/dinat/modeling_dinat.py
kaixuanliu: Have updated.

src/transformers/models/dinat/modeling_dinat.py
kaixuanliu: have updated

kaixuanliu:
@ydshieh Hi, current transformers uses an old version natten(0.14.6) and have many conflicts when we run CI tests for tests/models/dinat, in this PR we upgrade the natten to latest version(0.21.0) and I have updated related APIs, all relavant test cases passed w/ this PR on my side, pls help review. Many thanks!

ydshieh:
Just want to point out that

https://github.com/huggingface/transformers/pull/40287

But happy to update the dependency however.

src/transformers/models/dinat/modeling_dinat.py
```diff
@@ -267,44 +265,39 @@ def forward(
         hidden_states: torch.Tensor,
         output_attentions: Optional[bool] = False,
     ) -> tuple[torch.Tensor]:
-        batch_size, seq_length, _ = hidden_states.shape
-        query_layer = (
-            self.query(hidden_states)
-            .view(batch_size, -1, self.num_attention_heads, self.attention_head_size)
-            .transpose(1, 2)
+        batch_size, height, width, _ = hidden_states.shape
```
ydshieh, line 268: could you shortly explain the changes in these lines of changes in this attention block ?

ydshieh:
other than some torch export issues, I got

> =FAILED tests/models/dinat/test_modeling_dinat.py::DinatModelIntegrationTest::test_inference_image_classification_head - AssertionError: Tensor-likes are not close!

on Nvidia A10. It's fine, you don't need to deal with it, but is this test passing on XPU?

kaixuanliu:
@ydshieh Oops, I tested the `tests/models/dinat` both on A100 and Intel XPU, but I forgot to turn on `RUN_SLOW`, this case also fails on my side after I `export RUN_SLOW=1`. Will change this PR to draft and investigate what happened. Sorry for the inconvenience!

ydshieh:
@kaixuanliu 

I have to let you know the whole `dinat` tests are not (actually) running for very long time, from our history, it was not run any test on 2025/01/01, and maybe even earlier.

So it might be really difficult to check , especially we also have the natten / torch version issues etc.

Is it important for XPU to run test with this model?

yao-matrix:
> @kaixuanliu
> 
> I have to let you know the whole `dinat` tests are not (actually) running for very long time, from our history, it was not run any test on 2025/01/01, and maybe even earlier.
> 
> So it might be really difficult to check , especially we also have the natten / torch version issues etc.
> 
> Is it important for XPU to run test with this model?

@ydshieh , does `transformers` have plan to retire `dinat`, seems it's out of maintain per your description? For XPU, we want to deliver a solid message to community that XPU has equivalent broad modeling coverage as CUDA in `transformers`, so we are doing everything we can do to make it happen. `dinat` and `LayoutLMV2` are the last 2 models which don't work on XPU:
- for `dinat`, since `natten` is still actively maintained now, but as you can see the API is no longer back-compatible, so we are submitting PR to `natten` to  add XPU support, and upgrade `natten` version in `dinat` to make it work.
- for `LayoutLMV2`, since `detectronv2` is out-of-maintain now, we waived its support.

That's the reason behind what we are doing, :). If `transformers` has a retire plan for out-of-maintain models, we can waive it, too.

ydshieh:
Personally, I would like to deprecate it. Let me check some numbers and talk to the core maintainers and come back to you.

Great jobs the XPU team make this amazing progress!

ydshieh:
Hi @kaixuanliu @yao-matrix 

Yes, we are going to deprecate this dinat (along with some models).

kaixuanliu:
@ydshieh , Thanks for the feedback. Will close this PR. BTW,  we want to double check `layoutlmV2` will also be deprecated, right?

ydshieh:
Hi @kaixuanliu 

I haven't talked about `layoutlmV2` with the team (forgot about this). I will have to check.

But I think it's fair that the XPU team not to test it (I have trouble many time with detectron2 too).

ydshieh:
BTW, if XPU team fetch the results of `transformers` Nvidia CI runs and do the comparison? I mean for this specific `dinat`, there is actually no test being run, so there is no reason for XPU to run and compare to Nvidia.

I know our artifacts may lack some important information for external to use reliably.

We have recently these file collated_reports_xxx, for example

https://huggingface.co/datasets/hf-internal-testing/transformers_daily_ci/tree/main/2025-11-13/ci_results_run_models_gpu

that might be helpful to you. 

Let me know if you have any feedback.

yao-matrix:
> BTW, if XPU team fetch the results of `transformers` Nvidia CI runs and do the comparison? I mean for this specific `dinat`, there is actually no test being run, so there is no reason for XPU to run and compare to Nvidia.
> 
> I know our artifacts may lack some important information for external to use reliably.
> 
> We have recently these file collated_reports_xxx, for example
> 
> https://huggingface.co/datasets/hf-internal-testing/transformers_daily_ci/tree/main/2025-11-13/ci_results_run_models_gpu
> 
> that might be helpful to you.
> 
> Let me know if you have any feedback.

Thx @ydshieh , we will do this. And one more question, is the dashboard here https://huggingface.co/spaces/transformers-community/transformers-ci-dashboard generated by the test plan you mentioned as https://huggingface.co/datasets/hf-internal-testing/transformers_daily_ci/tree/main/2025-11-13/ci_results_run_models_gpu? Thx very much.

ydshieh:
@yao-matrix The dashboard is developed by another team member (and I am less engaged in this part), but yes, it's from the files `collated_reports_xxx` files in the linked repository.

If you need more details, don't hesitate.

ydshieh:
@kaixuanliu @yao-matrix 

Don't hesitate to reach out (on github or on slack) in the future when you find something that requires more work and not sure about if it is worth doing that .

Apologize again for this not well-commutated hidden states (of not maintained models but not deprecated explicitly)

kaixuanliu:
It's OK. Thanks for your kindly reminder! Will do it next time :)

PR #41909, opened 2025-10-28
Fix Break change of AWQ FusedModules due to Attention Refactor

### LINKED ISSUE #41910: Breaking change about AWQ Fused modules due to Attention Refactor
### System Info

transformers==5.0.0dev
autoawq==0.2.9
autoawq_kernels==0.0.9
torch==2.6.0+cu124

### Who can help?

Due to PR #35235, the `past_key_values` is no longer a returned value of attention modules.

However, when using AWQ models with Fused modules [AWQ Fused modules docs](https://huggingface.co/docs/transformers/main/en/quantization/awq#fused-modules), there will be an error like issue #38554

```bash
ValueError: too many values to unpack (expected 2)
```

So we can hack the `awq.modules.fused.attn.QuantAttentionFused` to avoid returning `past_key_values`. Therefore, I create a primary PR #41909 to fix it.

However, for special `rope_type` such as LLaMA3, the RoPE implementation in AutoAWQ will cause error, since `awq.modules.fused.attn.RoPE` supports default RoPE only.

Maybe we can implement and maintain `AwqRoPE` and `AwqQuantAttentionFused` in `transformers.integrations.awq`? 

I'd like to refine my PR to help transformers fix this bug!

@SunMarc @MekkCyber

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

```python

from transformers import AwqConfig, AutoModelForCausalLM, AutoTokenizer


# model_path = "./llama-3.1-8b-instruct-awq"
model_path = "./qwen2.5-7b-instruct-awq"
# model_path = "./qwen3-8b-awq"

awq_config = AwqConfig(
    bits=4,
    do_fuse=True,
    fuse_max_seq_len=8192
)

model = AutoModelForCausalLM.from_pretrained(model_path, quantization_config=awq_config).to("cuda:0")
print(model)
tokenizer = AutoTokenizer.from_pretrained(model_path)

max_new_tokens = 1024 if "qwen3" in model_path else 32


messages = []

prompt1 = "What is the result of 3+5?"
messages.append({"role": "user", "content": prompt1})
text1 = tokenizer.apply_chat_template(messages, tokenize=False, add_generation_prompt=True)
inputs1 = tokenizer(text1, return_tensors="pt").to("cuda:0")

generated_ids1 = model.generate(**inputs1, max_new_tokens=max_new_tokens)
output_ids1 = generated_ids1[0, len(inputs1.input_ids[0]) :].tolist()
output1 = tokenizer.decode(output_ids1, skip_special_tokens=True)
messages.append({"role": "assistant", "content": output1})
print("Output 1:", output1)

prompt2 = "What about adding 10 to that result?"
messages.append({"role": "user", "content": prompt2})
text2 = tokenizer.apply_chat_template(messages, tokenize=False, add_generation_prompt=True)
inputs2 = tokenizer(text2, return_tensors="pt").to("cuda:0")

generated_ids2 = model.generate(**inputs2, max_new_tokens=max_new_tokens)
output_ids2 = generated_ids2[0, len(inputs2.input_ids[0]) :].tolist()
output2 = tokenizer.decode(output_ids2, skip_special_tokens=True)
messages.append({"role": "assistant", "content": output2})
print("Output 2:", output2)

```

### Expected behavior

There is no error.

fanqiNO1:
# What does this PR do?

Fixes #41910

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @MekkCyber

### DIFF
--- a/src/transformers/models/llama/modeling_llama.py
+++ b/src/transformers/models/llama/modeling_llama.py
@@ -425,6 +425,7 @@ class LlamaModel(LlamaPreTrainedModel):
                 position_embeddings=position_embeddings,
                 position_ids=position_ids,
                 past_key_values=past_key_values,
+                use_cache=use_cache,
                 cache_position=cache_position,
                 **kwargs,
             )
--- a/src/transformers/integrations/awq.py
+++ b/src/transformers/integrations/awq.py
@@ -71,7 +71,7 @@ AWQ_FUSED_MAPPINGS = {
         "mlp": ["gate_proj", "up_proj", "down_proj"],
         "layernorm": ["input_layernorm", "post_attention_layernorm", "norm"],
         "use_alibi": False,
-    }
+    },
 }
 
 AWQ_SCALES_MAPPINGS = {
@@ -105,6 +105,7 @@ if is_auto_awq_available():
             device (`torch.device`):
                 The device to put the module on.
         """
+
         def __init__(self, rope_type, head_dim, max_seq_len, config, device):
             rope_init_fn = ROPE_INIT_FUNCTIONS[rope_type]
             self.inv_freq, self.attention_scaling = rope_init_fn(config, device)
@@ -324,9 +325,11 @@ def fuse_awq_modules(model, quantization_config):
 
         # Hack QuantAttentionFused to modify the return value of forward function to avoid returning past_key_value
         old_quant_attention_fused_forward = QuantAttentionFused.forward
+
         def new_quant_attention_fused_forward(self, *args, **kwargs):
             attn_output, attention_weight, _ = old_quant_attention_fused_forward(self, *args, **kwargs)
             return attn_output, attention_weight
+
         QuantAttentionFused.forward = new_quant_attention_fused_forward
     else:
         raise ValueError("Fusing is only supported for the AutoAWQ backend")
--- a/src/transformers/integrations/awq.py
+++ b/src/transformers/integrations/awq.py
@@ -18,6 +18,7 @@ import importlib
 from packaging import version
 
 from ..activations import ACT2FN
+from ..modeling_rope_utils import ROPE_INIT_FUNCTIONS
 from ..modeling_utils import PreTrainedModel
 from ..utils import is_auto_awq_available, is_ipex_available, is_torch_available, logging
 from ..utils.quantization_config import (
@@ -46,7 +47,6 @@ AWQ_FUSED_MAPPINGS = {
         "mlp": ["w1", "w3", "w2"],
         "layernorm": ["input_layernorm", "post_attention_layernorm", "norm"],
         "use_alibi": False,
-        "rope_theta": 1000000.0,
     },
     "llama": {
         "attention": ["q_proj", "k_proj", "v_proj", "o_proj"],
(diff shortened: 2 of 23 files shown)


SunMarc (commented):
Thanks, really appreciate it! Can you add some tests with small models ?

fanqiNO1:
Certainly! I will add some tests after implementing `AWQRoPE` to support models like `LLaMA3`~

fanqiNO1:
I modified modeling_llama because autoawq uses the use_cache flag to determine whether to use KV cache by `start_pos`.

https://github.com/casper-hansen/AutoAWQ/blob/88e4c76b20755db275574e6a03c83c84ba3bece5/awq/modules/fused/attn.py#L231

Therefore, use_cache must be passed down to the decoder layer.

I’ll add some tests this weekend~

fanqiNO1:
Sorry for the late update.

I've added a test using `hugging-quants/Meta-Llama-3.1-8B-Instruct-AWQ-INT4` to verify the correctness of the `AWQRoPE` implementation, and I think this PR is ready for review~

src/transformers/models/llama/modeling_llama.py
```diff
@@ -425,6 +425,7 @@ def forward(
                 position_embeddings=position_embeddings,
                 position_ids=position_ids,
                 past_key_values=past_key_values,
+                use_cache=use_cache,
```
SunMarc, line 428: don't mind passing this but I didn't find where this is used in decoder layer -> attention layer

SunMarc (commented):
Thanks for this and nice tests ! Just a minor comment

src/transformers/models/llama/modeling_llama.py
fanqiNO1: A very good question.

On one hand, when using `model.generate`, `use_cache` is set to True, which enables the model to utilize `past_key_values`. At this point, the logic in `autoawq` checks whether the forward call originates from `generate` by inspecting `use_cache`, and accordingly adjusts the starting position of its precomputed RoPE embeddings. If `use_cache` is not passed down to the decoder layer and subsequently to the attention module, `autoawq` cannot determine whether it is inside a `generate` call. Consequently, it assumes the forward pass is always a regular one (i.e., without any cache), keeping the starting position fixed at 0, which leads to garbled output during inference.

`autoawq`: 

https://github.com/casper-hansen/AutoAWQ/blob/88e4c76b20755db275574e6a03c83c84ba3bece5/awq/modules/fused/attn.py#L218-L241

On the other hand, similar to the implementations in Qwen2 and Qwen3, `use_cache` is indeed passed to the decoder layer and then forwarded to the attention module—but it is not actually used within the attention module itself.

fanqiNO1:
In fact, I believe that as `Transformers` continue to evolve—with an increasing number of models and ever-larger model sizes—the demand for AWQ will inevitably grow. Since `AutoAWQ` has already been archived, compatibility issues are likely to become more frequent. It might be a good solution for HuggingFace to fork and maintain an `AutoAWQ` repository. Should such a day come, I would be very willing to help you maintain it.

src/transformers/models/llama/modeling_llama.py
SunMarc: Hmmmm I see, thanks for the extensive explanation !

SunMarc (approved):
Thanks a lot !

SunMarc:
> In fact, I believe that as Transformers continue to evolve—with an increasing number of models and ever-larger model sizes—the demand for AWQ will inevitably grow. Since AutoAWQ has already been archived, compatibility issues are likely to become more frequent. It might be a good solution for HuggingFace to fork and maintain an AutoAWQ repository. Should such a day come, I would be very willing to help you maintain it.

Thanks for your concerns ! We don't think we will maintain a fork of AutoAWQ as this is quite a huge tasks tbh with our current staff. I think the best solution for all quantization methods and not only AutoAWQ is to put the critical modeling code inside transformers (eg Linear4bit, QuantAttentionFused) if they want to make sure that nothing breaks and avoid monkey patching as much as possible + move the kernels to kernels-community repo. A good example is FP8-finegrained method.
If you are willing to upstream + simplify `AutoAWQ` modeling code into transformers like `QuantAttentionFused` class, happy to review the PRs !

fanqiNO1:
> > In fact, I believe that as Transformers continue to evolve—with an increasing number of models and ever-larger model sizes—the demand for AWQ will inevitably grow. Since AutoAWQ has already been archived, compatibility issues are likely to become more frequent. It might be a good solution for HuggingFace to fork and maintain an AutoAWQ repository. Should such a day come, I would be very willing to help you maintain it.
> 
> Thanks for your concerns ! We don't think we will maintain a fork of AutoAWQ as this is quite a huge tasks tbh with our current staff. I think the best solution for all quantization methods and not only AutoAWQ is to put the critical modeling code inside transformers (eg Linear4bit, QuantAttentionFused) if they want to make sure that nothing breaks and avoid monkey patching as much as possible + move the kernels to kernels-community repo. A good example is FP8-finegrained method.

Thank you for your reply!

I indeed overlooked the human effort required to maintain such a library before. Next, I’ll open a PR to migrate, simplify, and integrate the key code from `AutoAWQ` into `transformers` as possible, and also move the relevant implementations from `AutoAWQ-kernels` into `kernel-community`.

I plan to first draft a well-thought-out design proposal and include it in the new PR description so we can discuss it together. Thanks so much!

fanqiNO1:
I've created a new PR #42256, which includes an analysis of AutoAWQ and its kernels.

Is there anything else I need to do before this PR can be merged?

SunMarc:
you need to do make `fix-copies` to propagate some changes you did in the modeling file of llama

MekkCyber (approved):
Thanks for fixing this!

PR #41916, opened 2025-10-28
feat(ci): add continuous batching to benchmarks

McPatate:


benchmark_v2/framework/benchmark_runner.py
```diff
@@ -274,6 +276,54 @@ def run_benchmark(
...
+        first_req_results = []
+        timestamps = []
+        wall_time_0 = time.perf_counter()
+        inputs = self.inputs["input_ids"].tolist()
+        manager.add_requests(inputs, max_new_tokens=max_new_tokens, stream=True)
+        first_req_id = None
+        num_requests = len(inputs)
+        finished_requests = 0
+        while finished_requests < num_requests:
+            timestamps.append(time.perf_counter() - wall_time_0)
```
remi-or: Does this mean that a timestamp is added for each finished request or each generated token? Or none of those?

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -274,6 +276,54 @@ def run_benchmark(
...
+            # NOTE: I don't like having the extra if stmt here, but hopefully won't degrade perf too much
+            result = manager.get_result()
```
remi-or: Does this not incur a big overhead?

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -274,6 +276,54 @@ def run_benchmark(
...
+    def time_generate_batch(
+        self,
+        max_new_tokens: int,
+        gpu_monitor: GPUMonitor | None = None,
+    ) -> tuple[float, list[float], str, GPURawMetrics | None]:
+        if gpu_monitor is not None:
+            gpu_monitor.start()
+        config = GenerationConfig(
+            max_new_tokens=max_new_tokens,
+            use_cuda_graphs=False,
```
remi-or: Why no cuda graphs? I would leave it to None so we have defaut, automatic behavior

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -274,6 +276,54 @@ def run_benchmark(
...
+            eos_token_id=self.tokenizer.eos_token_id,
+            pad_token_id=self.tokenizer.pad_token_id,
+            do_sample=False,
```
remi-or: It seems like the real-world case scenario is `do_sample=True`

benchmark_v2/framework/data_classes.py
```diff
@@ -141,27 +142,22 @@ def get_measured_ttft(self) -> list[float]:
     def get_measured_itl(self) -> list[float]:
         return [(dt[-1] - dt[0]) / (len(dt) - 1) for dt in self.token_generation_times if len(dt) > 1]
 
-    def get_throughput(self, batch_size: int) -> float:
+    def get_throughput(self, batch_size: int) -> list[float]:
```
remi-or: good catch, thx

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -877,9 +877,11 @@ def add_request(
         self.input_queue.put(state, block=True, timeout=10)  # XXX: pass timeout as fn arg?
         return request_id
 
-    def add_requests(self, inputs: list[list[int]], max_new_tokens: Optional[int] = None) -> None:
+    def add_requests(
+        self, inputs: list[list[int]], max_new_tokens: Optional[int] = None, streaming: bool = False
+    ) -> None:
         for input_ids in inputs:
-            self.add_request(input_ids, max_new_tokens=max_new_tokens)
+            self.add_request(input_ids, max_new_tokens=max_new_tokens, streaming=streaming)
```
remi-or, line 886: did we not change this to get all the "streaming" stuff out?

benchmark_v2/framework/benchmark_config.py
```diff
@@ -56,6 +58,7 @@ def __init__(
         self.check_validity(skip_validity_check)
         self.name = name if name is not None else self.infer_name()
 
+    # TODO: add validity checks for continuous batching
```
remi-or: Could we adress this, or at least part of it, in this PR?

remi-or (changes requested):

benchmark_v2/framework/benchmark_runner.py
McPatate: will address later when I fix the `get_result(Some(request_id))` variant

src/transformers/generation/continuous_batching/continuous_api.py
McPatate: We added it to each fn instead of having it as a global in the `ContinuousBatchingManager` class

benchmark_v2/framework/benchmark_config.py
McPatate: sure, was waiting for you for help 😇

McPatate:
@copilot, in the `Self-hosted runner (benchmark)` I have an odd `TypeError("ContinuousBatchingManager.add_requests() got an unexpected keyword argument 'streaming'")`. The change is indeed reflected in the PR's diff and it works locally. Could you fix the issue? It's causing a downstream OOM due to improper shutdown and freeing of the `ContinuousBatchingManager` instance. Feel free to tackle that issue as well.

McPatate:
@copilot in the `Self-hosted runner (benchmark)` the run benchmarks step completes without finishing and it is marked as successful. could you investigate and fix?

McPatate:
@copilot removing the `finally` was indeed a mistake but that is not the issue at hand. I've fixed this issue, but the problem mentioned here:
> https://gh.io/copilot-coding-agent-docs in the Self-hosted runner (benchmark) the run benchmarks step completes without finishing and it is marked as successful. could you investigate and fix?

happens on main also. Can you take a look again?

benchmark_v2/framework/benchmark_config.py
```diff
@@ -191,24 +215,35 @@ def get_config_by_level(level: int) -> list[BenchmarkConfig]:
...
                         )
-                    )
         return configs
     # Otherwise, we add the configs for the given level
     if level >= 0:
         configs.append(BenchmarkConfig(attn_implementation="flex_attention", compile_mode="default"))
     if level >= 1:
         configs.append(BenchmarkConfig(attn_implementation="flash_attention_2"))
         configs.append(BenchmarkConfig(attn_implementation="eager", compile_mode="default"))
+        configs.append(
```
remi-or: I would put 
```
configs.append(
            BenchmarkConfig(attn_implementation="flash_attention_2", continuous_batching=True)
``` 
instead, it's the pest performing one so far.

benchmark_v2/framework/benchmark_config.py
```diff
@@ -191,24 +215,35 @@ def get_config_by_level(level: int) -> list[BenchmarkConfig]:
...
+            BenchmarkConfig(attn_implementation="paged|sdpa", compile_mode="default", continuous_batching=True)
+        )
     if level >= 2:
         configs.append(BenchmarkConfig(attn_implementation="sdpa", compile_mode="default"))
         configs.append(BenchmarkConfig(attn_implementation="flex_attention", compile_mode="default", kernelize=True))
         configs.append(BenchmarkConfig(attn_implementation="flash_attention_2", kernelize=True))
+        configs.append(
+            BenchmarkConfig(attn_implementation="paged|sdpa", compile_mode="default", continuous_batching=True)
```
remi-or: probably compile off is better, have you tested it with and without? I would be curious.

benchmark_v2/framework/benchmark_config.py
```diff
@@ -191,24 +215,35 @@ def get_config_by_level(level: int) -> list[BenchmarkConfig]:
...
+        )
+        configs.append(
+            BenchmarkConfig(attn_implementation="flash_attention_2", kernelize=True, continuous_batching=True)
```
remi-or: Not sure about the impact of kernelize here. Plus if we go my above comment is accepted, we already have a FA2 config.

remi-or (approved):
LGTM, few comments to adress, and we need to fix the disabled tests quick. Awesome addition!

benchmark_v2/framework/benchmark_config.py
McPatate: I haven't no! I'll try. What do we want to include? Better perf?

PR #41918, opened 2025-10-28
V4.57.1 training ci: Refactor `test_tensor_parallel.py`

3outeille:
# What does this PR do?

Refactor  `test_tensor_parallel.py` by removing `subprocess`. This way we can easily debug with breakpoint the test if it fails. On top of that, I made the tests more robust by testing on more than `--nproc_per_node> 2.` `--nproc_per_nodes = 8` crashes because the llama model we use is too tiny so can't do TP. But since it works with  `--nproc_per_nodes = 4` already, no need to get a bigger llama (that may slow down the tests)

## Who can review?

@ArthurZucker @Cyrilvallez


ArthurZucker (approved):
thanks

PR #41922, opened 2025-10-28
Fix rope_parameters for gemma3 weights conversion script

douglas-reid:
# What does this PR do?

Fixes the rope_parameters in the weights conversion script for Gemma 3.

These should be:

```
local => default @ 10_000.0
global => linear(8.0) @ 1_000_000.0
```


RyanMullins:
cc @zucchini-nlp

Rocketknight1:
Yes, good fix - `rope_parameters` is assumed to be keyed by layer type in the `modeling` file, so this conversion script would generate a broken config in at least some cases.

Rocketknight1 (approved):

PR #41925, opened 2025-10-28
[deepspeed tests fixes]

stas00:
Fixing a few deepspeed tests


tests/deepspeed/test_deepspeed.py
```diff
@@ -761,8 +761,8 @@ def test_gradient_accumulation(self, stage, dtype):
             self.assertAlmostEqual(no_grad_accum_a, yes_grad_accum_a, delta=1e-4)
             self.assertAlmostEqual(no_grad_accum_b, yes_grad_accum_b, delta=1e-4)
         else:
-            self.assertAlmostEqual(no_grad_accum_a, yes_grad_accum_a, places=5)
-            self.assertAlmostEqual(no_grad_accum_b, yes_grad_accum_b, places=5)
+            self.assertAlmostEqual(no_grad_accum_a, yes_grad_accum_a, delta=1e-4)
+            self.assertAlmostEqual(no_grad_accum_b, yes_grad_accum_b, delta=1e-4)
```
stas00: This fixes:

```
FAILED tests/deepspeed/test_deepspeed.py::TrainerIntegrationDeepSpeed::test_gradient_accumulation_zero2_bf16 - AssertionError: 0.00043487...
FAILED tests/deepspeed/test_deepspeed.py::TrainerIntegrationDeepSpeed::test_gradient_accumulation_zero2_fp16 - AssertionError: 0.00028920...
```

Please check that it doesn't hide a problem - since this had less tolerance when it was originally written. This possibly indicates a regression.

Do note that if `hpu` code branch just above this has the same tolerance as this proposed fixed.

To reproduce/full mismatch numbers:
```
$ RUN_SLOW=1 pytest -sv tests/deepspeed/test_deepspeed.py::TrainerIntegrationDeepSpeed::test_gradient_accumulation_zero2_fp16
tests/deepspeed/test_deepspeed.py:764: in test_gradient_accumulation
    self.assertAlmostEqual(no_grad_accum_a, yes_grad_accum_a, places=6)
E   AssertionError: 0.0002892017364501953 != 0.00032782554626464844 within 6 places (3.8623809814453125e-05 difference)
```

stas00: I'm also thinking that this could be related to the gpu it's being run on. These tests were originally run on v100s and later a100s, I had the above mismatches with h200s - as we know the precision can change between architectures so probably this is ok.

Rocketknight1:
cc @sunmarc @mekkcyber for the deepspeed integration

tests/deepspeed/test_deepspeed.py
MekkCyber: Yes, I think it’s just GPU-related numerical differences — for instance, different GPUs can produce slightly different results, so tests that pass locally might fail on CI. That said, I’ll let @SunMarc make the final call.

MekkCyber (commented):
Thanks for fixing this!

tests/deepspeed/test_deepspeed.py
SunMarc: What's strange is that the same tests for z3 pass. I wanted to check what's happening but never had the time cc @IlyasMoutawwakil as you changed the tolerance for hpu. 

We don't need the if branch if we decide to use the same tolerance @stas00

SunMarc (approved):
Thanks, maybe this is indeed an issue with the hardware and z2

tests/deepspeed/test_deepspeed.py
stas00: I agree, once there is an agreement to go ahead I'll collapse the 2 identical code branches into 1.

stas00: ok, I collapsed the 2 branches into 1 as they are the same.

stas00:
this should be good to merge, @SunMarc

PR #41930, opened 2025-10-29
handle inputs from Siglip/Siglip2 non-automapped encoder layers

molbap:
# What does this PR do?

Should fix #41929 . The `check_model_inputs` / `can_record_outputs` interaction is not always trivial and models with several entrypoints such as `VisionModel` vs `VisionTransformer` are missing some, adding it here. Also added a modification in `generic` to make sure the flag was captured, not 100% sure it's needed.


src/transformers/utils/generic.py
```diff
@@ -836,7 +836,7 @@ def wrapper(self, *args, **kwargs):
                     all_args[k] = v
 
             # _can_record_outputs is None by default
-            capture_flags = _CAN_RECORD_REGISTRY.get(str(self.__class__)) or {}  # there is a weak ref for executorch
+            capture_flags = _CAN_RECORD_REGISTRY.get(str(self.__class__)) or getattr(self, "_can_record_outputs", {})
```
vasqu: I think we can revert this, the registry should already be either applied to the class or we get the default `{}`. See https://github.com/huggingface/transformers/blob/4d0b6758b90aa54e4077171e6d42c55e0c01c622/src/transformers/modeling_utils.py#L1851

And I honestly don't wanna risk breaking executorch

src/transformers/models/siglip2/modeling_siglip2.py
```diff
@@ -811,7 +817,6 @@ def get_input_embeddings(self) -> nn.Module:
         return self.vision_model.embeddings.patch_embedding
 
     @check_model_inputs(tie_last_hidden_states=False)
-    @auto_docstring
```
vasqu, line 814: Nit: Ig this was already inherited?

vasqu (approved):
lgtm, I would just revert the registry default change. Shouldnt be needed and I don't wanna risk breaking executorch (or similar trace ops)

src/transformers/utils/generic.py
molbap: yeah the latter reason is compelling :grimacing:

src/transformers/models/siglip2/modeling_siglip2.py
molbap: uh, interesting. yes it should have been, entirely unrelated to this lol

molbap:
run-slow: siglip, siglip2

molbap:
run-slow: siglip, siglip2

molbap:
ended up down the rabbit hole of wrong `PreTrainedModel` inheritances hehe, @yonigozlan if you want to take a look

molbap:
run-slow: siglip, siglip2

molbap:
run-slow: siglip, siglip2

molbap:
run-slow: siglip, siglip2

molbap:
run-slow: siglip, siglip2

tests/models/siglip2/test_modeling_siglip2.py
```diff
@@ -166,6 +166,10 @@ def test_flash_attn_2_inference_equivalence_right_padding(self):
     def test_sdpa_can_dispatch_on_flash(self):
         pass
 
+    @unittest.skip(reason="This test is broken on A10 multi runners for now")
+    def test_multi_gpu_data_parallel_forward(self):
+        pass
```
vasqu: We shouldn't skip these, will be hard to revert because everyone will forget imo

src/transformers/models/siglip/modeling_siglip.py
```diff
@@ -510,7 +510,7 @@ def _init_weights(self, module):
             nn.init.xavier_uniform_(module.fc2.weight)
             nn.init.normal_(module.fc1.bias, std=1e-6)
             nn.init.normal_(module.fc2.bias, std=1e-6)
-        elif isinstance(module, SiglipMultiheadAttentionPoolingHead):
+        elif "MultiheadAttentionPoolingHead" in module.__class__.__name__:
```
vasqu: Seems unrelated, no?

vasqu (commented):
Awkward situation with the auto model, can you also check out clip? We should face very similar issues there as well + do we need to adjust tests maybe?

tests/models/siglip2/test_modeling_siglip2.py
molbap: well it's entirely broken, so need to write it down somewhere :sweat_smile:

src/transformers/models/siglip/modeling_siglip.py
molbap: you would believe that, right :eyes: 
it wasn't AFAIK but it's related to when I shuffled around PreTrainedModel or not

molbap:
yes, needs another PR to make sure tests are actually testing that attentions and hidden states are traced!

tests/models/siglip2/test_modeling_siglip2.py
vasqu: Fair enough but the issue is bigger than this PR and model. Imo we should isolate this in a different PR and directly in the common tests. Wdyt?

I've encountered these issues with all recent models I've interacted with.

src/transformers/models/siglip/modeling_siglip.py
vasqu: Dang wtf, no biggie but interesting

tests/models/siglip2/test_modeling_siglip2.py
molbap: reverted for now, let's remember that slow tests do not pass but that's ok :P

vasqu:
Wanna merge this @molbap? I think it's better to go for it than for it to be forgotten :eyes:

molbap:
You're right @vasqu , thanks for the push! merging

PR #41931, opened 2025-10-29
fix 3 failed test cases for video_llama_3 model on Intel XPU

kaixuanliu:
This PR fixes 3 failed test cases on Intel XPU:
```
1.tests/models/video_llama_3/test_modeling_video_llama_3.py::VideoLlama3IntegrationTest::test_small_model_integration_test
2.tests/models/video_llama_3/test_modeling_video_llama_3.py::VideoLlama3IntegrationTest::test_small_model_integration_test_batch_wo_ima
ge
3.tests/models/video_llama_3/test_modeling_video_llama_3.py::VideoLlama3ModelTest::test_generate_with_quant_cache 
```


ydshieh:
Thank you @kaixuanliu 

cc @yao-matrix for a first look 🙏

src/transformers/utils/import_utils.py
```diff
@@ -58,7 +58,7 @@ def _is_package_available(pkg_name: str, return_version: bool = False) -> tuple[
             # pick the first item of the list as best guess (it's almost always a list of length 1 anyway)
             distribution_name = pkg_name if pkg_name in distributions else distributions[0]
             package_version = importlib.metadata.version(distribution_name)
-        except importlib.metadata.PackageNotFoundError:
+        except (importlib.metadata.PackageNotFoundError, KeyError):
```
ydshieh: Could you explain what is the issue here that requires this new `KeyError`?

yao-matrix: same question.

tests/models/video_llama_3/test_modeling_video_llama_3.py
```diff
@@ -831,8 +832,18 @@ def test_small_model_integration_test(self):
...
+                (
+                    "cuda",
+                    None,
+                ): "user\n\nDescribe the image.\nassistant\nThe image captures a vibrant nighttime scene on a bustling city street. A woman in a striking red dress",
+                (
+                    "xpu",
+                    None,
+                ): "user\n\nDescribe the image.\nassistant\nThe image captures a vibrant night scene in a bustling Japanese city. A woman in a striking red dress",
+            }
+        ).get_expectation()
```
yao-matrix, line 841: pls use `format: off` and `format: on`, and the same format as other files, refer to https://github.com/huggingface/transformers/blob/a43b36cf802f00616800e0bd4d748679236123ee/tests/models/janus/test_modeling_janus.py#L496

tests/models/video_llama_3/test_modeling_video_llama_3.py
```diff
@@ -875,10 +886,24 @@ def test_small_model_integration_test_batch_wo_image(self):
...
+                ],
+                (
+                    "xpu",
+                    None,
+                ): [
+                    "user\n\nDescribe the image.\nassistant\nThe image captures a vibrant night scene in a bustling Japanese city. A woman in a striking red dress",
+                    "user\nWhat is relativity?\nassistant\nRelativity is a scientific theory that describes the relationship between space and time. It was first proposed by",
+                ],
+            }
+        ).get_expectation()
```
yao-matrix, line 897: as above

src/transformers/utils/import_utils.py
kaixuanliu: When i run `pytest -rA tests/models/video_llama_3/test_modeling_video_llama_3.py::VideoLlama3ModelTest::test_generate_with_quant_cache`, it will fail and return error:
```
pkg_name = 'optimum.quanto', return_version = True

    def _is_package_available(pkg_name: str, return_version: bool = False) -> tuple[bool, str] | bool:
        """Check if `pkg_name` exist, and optionally try to get its version"""
        spec = importlib.util.find_spec(pkg_name)
        package_exists = spec is not None
        package_version = "N/A"
        if package_exists and return_version:
            try:
                # importlib.metadata works with the distribution package, which may be different from the import
                # name (e.g. `PIL` is the import name, but `pillow` is the distribution name)
>               distributions = PACKAGE_DISTRIBUTION_MAPPING[pkg_name]
E               KeyError: 'optimum.quanto'

src/transformers/utils/import_utils.py:56: KeyError
```
Here is to fix this.

tests/models/video_llama_3/test_modeling_video_llama_3.py
kaixuanliu: Done

tests/models/video_llama_3/test_modeling_video_llama_3.py
kaixuanliu: Done

src/transformers/utils/import_utils.py
yao-matrix: @kaixuanliu it shows `'optimum': ['optimum-quanto', 'optimum-onnx', 'optimum'],` in `PACKAGE_DISTRIBUTION_MAPPING`, so we may need another logic to search `optimum-quanto` distribution in `optimum` package, rather than using your current logic, because the package actually there, just we find it in a wrong way in this case. @ydshieh , WDYT?

tests/models/video_llama_3/test_modeling_video_llama_3.py
yao-matrix: pls also "follow the same format as other files", as I shared in above link, thx.

tests/models/video_llama_3/test_modeling_video_llama_3.py
yao-matrix: as above

src/transformers/utils/import_utils.py
ydshieh: Let me check ,thank you for explaining the issue.

kaixuanliu: In this PR, when we catch the `keyError` exception, it will load `optimum-quanto` in [package = importlib.import_module(pkg_name)](https://github.com/huggingface/transformers/blob/main/src/transformers/utils/import_utils.py#L64), where `pkg_name` is `optimum.quanto`, and it can be imported properly.

tests/models/video_llama_3/test_modeling_video_llama_3.py
kaixuanliu: Have updated. Is it OK now?

ydshieh: yes, thank you both!

ydshieh:
@kaixuanliu @yao-matrix Thank you.

I have currently some trouble to get a CI runner to check the `_is_package_available` issue. From [our scheduled CI runs](https://github.com/huggingface/transformers/actions/runs/18961133099/job/54148721758), there is no issue. Of course, it's likely depending on the sys. environment, but I don't want to merge this PR without being verified more deeply myself.

I am happy to merge the PR if you can revert this part of change, and open an issue for this `_is_package_available`. Don't forget to  provide the env. details (`transformers env` and `pip freeze`). WDYT?

ydshieh:
run-slow: vit, clip

kaixuanliu:
@SunMarc , Hi, have reverted the code and submitted issue here: https://github.com/huggingface/transformers/issues/41981

ydshieh (approved):
👍 🙏

PR #41932, opened 2025-10-29
Fix: Handle missing safetensors gracefully to prevent import errors

Pranavi125:
This PR adds a safeguard for environments where `safetensors` is not installed.
It prevents import errors during dependency checks and allows transformers to load normally.

Changes made:
- Updated `setup.py` to conditionally check for safetensors
- Improved dependency handling logic

Tested locally: verified that transformers imports correctly with and without safetensors.

### DIFF
--- a/src/transformers/dependency_versions_check.py
+++ b/src/transformers/dependency_versions_check.py
@@ -15,7 +15,6 @@
 from .dependency_versions_table import deps
 from .utils.versions import require_version, require_version_core
 
-
 # define which module versions we always want to check at run time
 # (usually the ones defined in `install_requires` in setup.py)
 #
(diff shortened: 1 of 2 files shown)


Pranavi125:
Hi, this PR addresses the dependency handling for safetensors.
It ensures that transformers installs and runs correctly even when safetensors isn’t present.
I’ve verified imports and functionality both with and without the package installed.

Please review when possible — happy to make any adjustments if needed.

Rocketknight1:
Thanks for the PR, but `safetensors` is a pretty hard dependency for `transformers` at this point - I don't think we want alternate code paths where it isn't available!

PR #41934, opened 2025-10-29
Fix: Gemma3TextConfig rope scaling assignments

RyanMullins:
# What does this PR do?

Related to https://github.com/huggingface/transformers/pull/41922, this PR corrects the assignment of the `rope_scaling` dictionary present on some Gemma 3 pre-trained models on HF Hub when normalizing to the new `rope_parameters` value.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@zucchini-nlp PTAL since you have been handling the RoPE changes.

### DIFF
--- a/src/transformers/models/gemma3/configuration_gemma3.py
+++ b/src/transformers/models/gemma3/configuration_gemma3.py
@@ -186,10 +186,16 @@ class Gemma3TextConfig(PreTrainedConfig):
         self.final_logit_softcapping = final_logit_softcapping
         self.attn_logit_softcapping = attn_logit_softcapping
         self.layer_types = layer_types
+
         # Try to set `rope_scaling` if available, otherwise use `rope_parameters`
-        rope_scaling = kwargs.pop("rope_scaling", None)
-        if rope_scaling is not None:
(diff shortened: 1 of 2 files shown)


Rocketknight1 (approved):
I'm guessing the context for this is that `rope_scaling` clobbering `rope_parameters` (the old behaviour) gave the right config for older models but breaks on upcoming models that might have specific configs for e.g. `sliding_attention`?

RyanMullins:
> I'm guessing the context for this is that `rope_scaling` clobbering `rope_parameters` (the old behaviour) gave the right config for older models but breaks on upcoming models that might have specific configs for e.g. `sliding_attention`?

Yes, we've observed that `rope_scaling` was being applied to both `full_attention` _and_ `sliding_attention` for the Gemma 3 checkpoints on HF Hub and in our attempts to re-convert them from the original Orbax checkpoint. The correct behavior is for `rope_scaling` to only be applied to `full_attention` if it exists in the config, and the `sliding_attention` config should always be default RoPE @ 10k.

Rocketknight1:
Got it, thanks for the fix!

PR #41936, opened 2025-10-29
Fix: add missing SAFE_WEIGHTS_INDEX_NAME to __all__ in constants.py

Pranavi125:
This PR adds the missing SAFE_WEIGHTS_INDEX_NAME constant to the __all__ list in constants.py.

Why:
Without this, the constant isn’t exported when using
from transformers.utils.constants import *.

Impact:
- Keeps the constants module consistent.
- No functional changes or breaking impact.


src/transformers/modeling_utils.py
vasqu, line 1: These changes here are unrelated, no?

vasqu (commented):
These constants are used internally only, so I'm not sure why we'd want to include them here

Is there any issue or example you have? `SAFE_WEIGHTS_INDEX_NAME` doesn't even exist in the file

PR #41937, opened 2025-10-29
Refactor: Replace _default_log_level with DEFAULT_LOG_LEVEL constant

Pranavi125:
This PR refactors the logging module by renaming `_default_log_level` to `DEFAULT_LOG_LEVEL` 
to align with constant naming conventions and improve readability.


vasqu (commented):
Sorry but we don't want to change this unless for very important reasons.

There also seem to be a few unrelated changes :sweat_smile:

PR #41942, opened 2025-10-29
fix prepare_config_and_inputs_for_common bug in llava test

yao-matrix:
@ydshieh, pls help review, thx very much.

### DIFF
--- a/tests/models/llava/test_modeling_llava.py
+++ b/tests/models/llava/test_modeling_llava.py
@@ -152,10 +152,11 @@ class LlavaVisionText2TextModelTester:
     def prepare_config_and_inputs_for_common(self):
         config_and_inputs = self.prepare_config_and_inputs()
         config, pixel_values = config_and_inputs
-        input_ids = ids_tensor([self.batch_size, self.seq_length], config.text_config.vocab_size - 1) + 1
+
+        input_ids = ids_tensor([self.batch_size, self.seq_length], config.text_config.vocab_size - 2) + 2
+        attention_mask = torch.ones(input_ids.shape, dtype=torch.long).to(torch_device)
         input_ids[input_ids == config.image_token_index] = self.pad_token_id
         input_ids[:, : self.num_image_tokens] = config.image_token_index
-        attention_mask = input_ids.ne(1).to(torch_device)
 
         inputs_dict = {
             "pixel_values": pixel_values,


tests/models/llava/test_modeling_llava.py
```diff
@@ -152,10 +152,11 @@ def prepare_config_and_inputs(self):
     def prepare_config_and_inputs_for_common(self):
         config_and_inputs = self.prepare_config_and_inputs()
         config, pixel_values = config_and_inputs
-        input_ids = ids_tensor([self.batch_size, self.seq_length], config.text_config.vocab_size - 1) + 1
+
+        input_ids = ids_tensor([self.batch_size, self.seq_length], config.text_config.vocab_size - 2) + 2
+        attention_mask = torch.ones(input_ids.shape, dtype=torch.long).to(torch_device)
```
yao-matrix, line 157: the original  process of this function is:
1. the `ids_tensor` will generate random ids from `0` to `vocab_size - 2`
2. `input_ids` range is `1` to `vocab_size - 1` after `+ 1`
3.  there is no element of  `input_ids` equals `config.image_token_index` which is `0`, so `input_ids` range is still `1` to `vocab_size - 1`
4. force the first `16` elements to `config.image_token_index` which is `0`

In this process, whether `input_ids` has `1`(which is `pad_token_id`) is fully random, and the position of `1` is also random(which means `"pad_token"` can appear in the middle of the sequence), which brings the test some time pass, some time skip, which makes no sense.

I aligned the logic to what is used by `llava_next`, as https://github.com/huggingface/transformers/blob/main/tests/models/llava_next/test_modeling_llava_next.py#L160 to make it deterministic in padding behavior.

Rocketknight1:
I think that was probably meant as a ping for @ydshieh :sweat_smile:

yao-matrix:
> I think that was probably meant as a ping for @ydshieh 😅

yes, sorry 😅

tests/models/llava/test_modeling_llava.py
ydshieh: yeah, thank you for diving into this 🙏

ydshieh (approved):

ydshieh:
seems there is some connection issue for hub tests. I will check later

PR #41962, opened 2025-10-31
Patch Mistral Common Tokenizer 2

juliendenize:
# What does this PR do?

1. Fixes the behavior of add_special_tokens to match PretrainedTokenizer goals:
- if mode is finetuning: add both bos and eos tokens
- if mode is test: add only bos tokens so that the model can generate freely.

2. Add `ValidationMode` as a string support to avoid import from mistral-common for the user.

3. Fix decode when user passes an int instead of a sequence of ints

4. Add support to batch for `decode`.

5. Add special tokens properties to list ids and tokens string representation.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

cc @patrickvonplaten for viz

@ArthurZucker and @itazap

### DIFF
--- a/src/transformers/tokenization_mistral_common.py
+++ b/src/transformers/tokenization_mistral_common.py
@@ -215,7 +215,7 @@ class MistralCommonBackend(PushToHubMixin):
         """
         Constructs a `MistralCommonBackend`.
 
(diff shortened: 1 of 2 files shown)


ArthurZucker (approved):
Ty !

PR #41966, opened 2025-10-31
Fix GPT-2 Flash Attention 2 generation with left-padding

Abdennacer-Badaoui:
# What does this PR do?

Fixes Flash Attention 2 generation with left-padding in GPT-2 by ensuring `is_causal=True` is set even when padding masks are provided.

Flash Attention 2 handles causal masking and padding independently, but the original code incorrectly disabled causal masking when any attention mask was present.

Fixes: `test_flash_attn_2_generate_padding_left`

### DIFF
--- a/src/transformers/models/gpt2/modeling_gpt2.py
+++ b/src/transformers/models/gpt2/modeling_gpt2.py
@@ -245,6 +245,11 @@ class GPT2Attention(nn.Module):
 
         is_causal = attention_mask is None and query_states.shape[-2] > 1 and not is_cross_attention
 
+        # For flash attention backends, we must keep causal behavior even when a 2D padding mask is provided
+        # (flash attention uses `is_causal` for the triangular mask and expects an optional 2D key padding mask)
+        if self.config._attn_implementation in {"flash_attention_2", "flash_attention_3"} and not is_cross_attention:
+            is_causal = query_states.shape[-2] > 1
+
         using_eager = self.config._attn_implementation == "eager"
         attention_interface: Callable = eager_attention_forward
         if self.config._attn_implementation != "eager":
--- a/src/transformers/models/gpt2/modeling_gpt2.py
+++ b/src/transformers/models/gpt2/modeling_gpt2.py
@@ -260,7 +260,6 @@ class GPT2Attention(nn.Module):
                 value_states,
                 attention_mask,
                 dropout=self.attn_dropout.p if self.training else 0.0,
-                is_causal=self.is_causal,
                 **kwargs,
             )
 
--- a/src/transformers/models/gpt2/modeling_gpt2.py
+++ b/src/transformers/models/gpt2/modeling_gpt2.py
@@ -130,7 +130,7 @@ class GPT2Attention(nn.Module):
 
         self.attn_dropout = nn.Dropout(config.attn_pdrop)
         self.resid_dropout = nn.Dropout(config.resid_pdrop)
-        self.is_causal = True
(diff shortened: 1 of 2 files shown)


Abdennacer-Badaoui:
@vasqu , ready for review :)

src/transformers/models/gpt2/modeling_gpt2.py
```diff
@@ -245,6 +245,11 @@ def forward(
 
         is_causal = attention_mask is None and query_states.shape[-2] > 1 and not is_cross_attention
 
+        # For flash attention backends, we must keep causal behavior even when a 2D padding mask is provided
+        # (flash attention uses `is_causal` for the triangular mask and expects an optional 2D key padding mask)
+        if self.config._attn_implementation in {"flash_attention_2", "flash_attention_3"} and not is_cross_attention:
+            is_causal = query_states.shape[-2] > 1
```
Cyrilvallez: I believe we can greatly simplify the check in the end, such as a simple
```suggestion
        is_causal = not is_cross_attention
```
if I'm not mistaken

Cyrilvallez (dismissed):
Makes sense, fa should know that it's indeed causal! cc @vasqu here as well!

However, I think we can even completely remove the shape check, as sdpa checks it internally to make sure of the alignment, and fa should align the mask correctly already if I remember correctly.

src/transformers/models/gpt2/modeling_gpt2.py
vasqu: We should do this during init instead. It's probably very old code so the interface wasn't as established at that time. These edge cases should already be handled be the sdpa interface and we should not check for the same on general modeling code

vasqu: Relevant code part https://github.com/huggingface/transformers/blob/1c9077f66dc285dd32aa2410bed8acc680d64475/src/transformers/integrations/sdpa_attention.py#L69-L82

vasqu (commented):
I'm with Cyril here, let's simplify the forward, essentially removing the whole is_causal logic during the forward and only change it within in the init based on if we are cross attn or not

vasqu:
All the is_causal logic is correctly handled by the integrations; we just need to make sure we define it correctly on init

Abdennacer-Badaoui:
Makes sense. Thanks @vasqu, @Cyrilvallez. Pushing the changes now.

src/transformers/models/gpt2/modeling_gpt2.py
```diff
@@ -262,7 +260,7 @@ def forward(
                 value_states,
                 attention_mask,
                 dropout=self.attn_dropout.p if self.training else 0.0,
-                is_causal=is_causal,
+                is_causal=self.is_causal,
```
vasqu: We should be able to remove this line as well, the integrations grep it via the passed self (module)

vasqu:
run-slow: decision_transformer, gpt2

vasqu:
run-slow: decision_transformer, gpt2

vasqu (approved):
Thx a lot, looking good now!

vasqu:
@Abdennacer-Badaoui can you run the fa tests locally and confirm that those work? Seems like our CI doesn't have FA installed anymore so just wanna double check

Abdennacer-Badaoui:
@vasqu, I tested the FA failing tests for which I opened the PR. they’re all passing.

vasqu:
Thx a lot :heart:

PR #41968, opened 2025-10-31
test tensor parallel: make tests for dense model more robust

3outeille:
Previous tests were not robust enough. New changes tests:

- Non tp model against TP model (both eval-train) in forward/ forward + backward
- Non tp model against TP model (both eval-train) in forward + compile

### DIFF
--- a/tests/tensor_parallel/test_tensor_parallel.py
+++ b/tests/tensor_parallel/test_tensor_parallel.py
@@ -319,35 +319,49 @@ def _test_model_dense_backward_pass_impl():
     
     torch.distributed.barrier()
 
-def _test_model_dense_generate_impl():
-    """Implementation of test_model_generate for distributed execution."""
+def _test_model_dense_forward_compile_impl(mode):
+    """Implementation for comparing TP and non-TP model outputs with torch.compile."""
     model_id = "JackFram/llama-68m"
-
-    int(os.environ["RANK"])
-    int(os.environ["WORLD_SIZE"])
-
-    model = AutoModelForCausalLM.from_pretrained(model_id, dtype="auto", tp_plan="auto")
-    torch.distributed.barrier()
-
-    model.forward = torch.compile(model.forward)
-
-    has_dtensor = 0
-    for name, parameter in model.named_parameters():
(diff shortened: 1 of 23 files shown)


tests/tensor_parallel/test_tensor_parallel.py
```diff
@@ -211,95 +212,169 @@ def test_tp_plan_none_handling(self):
...
+    dist.barrier()
+    if mode == "eval":
+        model_tp.eval()
+    else:
+        model_tp.train()
+
+    # Load non-TP model and move to same device as TP model
+    device = model_tp.device
+    model = AutoModelForCausalLM.from_pretrained(model_id, dtype="auto")
+    model = model.to(device)
```
ArthurZucker, line 238: ah no device map auto because you always run with torchrun?

tests/tensor_parallel/test_tensor_parallel.py
```diff
@@ -211,95 +212,169 @@ def test_tp_plan_none_handling(self):
...
+    assert torch.allclose(loss, loss_tp, atol=1e-5, rtol=1e-5), (
+        f"TP and non-TP model losses differ. Non-TP loss: {loss.item()}, TP loss: {loss_tp.item()}, Diff: {(loss - loss_tp).abs().item()}"
+    )
 
-def _test_model_generate_impl(rank):
-    """Implementation of test_model_generate for distributed execution."""
-    model_id = "JackFram/llama-68m"
+    # Compare gradients for matching parameters
+    # Note: TP model may have sharded parameters (DTensors), so we slice the reference gradient to match
+    for (name, param), (name_tp, param_tp) in zip(model.named_parameters(), model_tp.named_parameters()):
```
ArthurZucker, line 300: won't always work with the refactor coming but fine for now

ArthurZucker (approved):

tests/tensor_parallel/test_tensor_parallel.py
3outeille: I run with `pytest` and this will call `torch.mp.spawn` which is pretty much like running with `torchrun`. I dont want to set `device_map=auto` because I am already setting manually the model to a specific `device` right after

PR #41972, opened 2025-11-01
fix: improve visibility of ValueError root causes in model config loading

scottzh8:
##  What does this PR do?

This PR improves error visibility when loading model configurations via `PretrainedConfig._get_config_dict`.

Previously, any `ValueError` raised deep in the Hugging Face Hub stack (for example, when the `hf_transfer` package was missing) was caught by a generic `except Exception` and rethrown as a vague `OSError`.  

With this change, `ValueError`s are explicitly caught and wrapped in an `OSError` that preserves the original message, making it much easier to diagnose environment and dependency issues.

**Example improvement:**
```
Before:
OSError: Can't load the configuration of 'mobicham/Qwen2.5-VL-3B-Instruct_int8wo_ao'. 
If you were trying to load it from 'https://huggingface.co/models', make sure you don't have a local directory with the same name...

After:
OSError: Fast download using 'hf_transfer' is enabled (HF_HUB_ENABLE_HF_TRANSFER=1) 
but 'hf_transfer' package is not available in your environment. Try `pip install hf_transfer`.
```

---

## Who can review?


Rocketknight1:
Hmn, I think this makes sense in some cases, but the problem is that it might make a lot of errors start surfacing in deep internal code and users might not know that this just means something like "you have a directory with the same name as the repo" or something. cc @arthurzucker @cyrilvallez WDYT?

Cyrilvallez:
Hey @scottzh8! It can indeed be a good idea, for that very particular case when `HF_HUB_ENABLE_HF_TRANSFER` is set without the associated lib installed. I believe it's the only ValueError that we do not control directly and that can happen. It should however directly be catched and transformed as OSError in `cached_files` in the `hub.py` integration, as we have several instances of the usage of the function. Do you wish to make the change, or do you want me to open a PR?

scottzh8:
Thanks @Cyrilvallez and @Rocketknight1 for the context! I'll update the PR and keep you posted

Cyrilvallez (approved):
Perfect! Thanks a lot for this! 🤗

PR #41973, opened 2025-11-02
Fix import error with huggingface_hub v1.0.0+

jameslovespancakes:
Update error class imports to use huggingface_hub.errors instead of huggingface_hub.utils to resolve ImportError with huggingface_hub versions 1.0.0 and above.

Fixes #41970


Rocketknight1:
Waiting for confirmation that this error is actually occurring on `main` first!

PR #36036, opened 2025-02-04
Add Phi-3.5-vision

Dahlbomii:
Draft PR for now, still need to add tests and convert to modular

cc @Rocketknight1

Fixes #36071

PR #41977, opened 2025-11-02
Add Phi3.5 Vision Model

yaswanth19:
Closes #36036


Rocketknight1:
cc @zucchini-nlp when you get a chance!

zucchini-nlp:
Phi3.5 💀 I will take a look some time this week

yaswanth19:
@zucchini-nlp PR should be ready for review next week. Will ping you at that time.

yaswanth19:
@zucchini-nlp PR is ready for initial review 🤗. CI has been broken from weekend but it was all green previously so no issues there.

src/transformers/models/phi3_v/modular_phi3_v.py
```diff
@@ -0,0 +1,416 @@
...
+        # Process image using CLIP model.
+        vision_outputs = self.vision_model(pixel_values, output_hidden_states=True)
+
+        # Extract the hidden states from the second last layer.
+        hidden_state = vision_outputs.hidden_states[-2][:, 1:]
+        hidden_state = hidden_state.reshape(num_images, num_crops, -1, self.image_dim_out)
+
+        # Transform the image features to text embedding space.
+        image_features = self.transform_image_embeds(hidden_state, image_sizes)
+        return image_features
```
yaswanth19, line 267: Majorly because of this function where the image features are transformed and projected in a bit non-standard way, also the utilization of image sizes makes it difficult to run inference with num_return_sequence>1 as then the image inputs are not synced with the repeated input_ids. Thus appropriate tests like beam search and any other tests are skipped.

src/transformers/models/phi3_v/processing_phi3_v.py
```diff
@@ -0,0 +1,203 @@
...
+                    if image_token_counter >= len(num_image_tokens):
+                        raise ValueError("More image placeholders in the text than images provided.")
+                    image_tokens = [self.image_token_id] * num_image_tokens[image_token_counter]
+                    tokenized_outputs.extend(image_tokens)
+                    image_token_counter += 1
+                else:
+                    text_tokens = self.tokenizer(split)["input_ids"]
+                    tokenized_outputs.extend(text_tokens)
+
+            tokenized_prompts.append(tokenized_outputs)
```
yaswanth19: Because of custom tokenization and the creation of subsequent attention mask, it's difficult to support assisted decoding and IMO it's not a super important generation method to support in this case.

docs/source/en/model_doc/phi3_v.md
```diff
@@ -0,0 +1,113 @@
+<!--Copyright 2022 The HuggingFace Team. All rights reserved.
```
zucchini-nlp: 2025 :)

docs/source/en/model_doc/phi3_v.md
```diff
@@ -0,0 +1,113 @@
...
+
+<hfoptions id="usage">
+<hfoption id="Pipeline">
+
+```py
+import torch
+from transformers import pipeline
+
+pipeline = pipeline(
+    task="text-generation",
```
zucchini-nlp: looks like copied and not yet updated

src/transformers/models/auto/modeling_auto.py
```diff
@@ -502,6 +503,7 @@ class _BaseModelWithGenerate(PreTrainedModel, GenerationMixin):
         ("nllb-moe", "NllbMoeForConditionalGeneration"),
         ("openai-gpt", "OpenAIGPTLMHeadModel"),
         ("paligemma", "PaliGemmaForConditionalGeneration"),
+        ("phi3_v", "Phi3VForConditionalGeneration"),
```
zucchini-nlp, line 513: let's add in 'ImageTextToText' mapping as well

src/transformers/models/phi3_v/convert_phi3_v_weights_to_hf.py
```diff
@@ -0,0 +1,364 @@
...
+        model = Phi3VForConditionalGeneration(config)
+
+    # model.generation_config._from_model_config = False
+    # model.generation_config.temperature = 1
+    # model.generation_config.guidance_scale = 5
+    # model.generation_config.pad_token_id = tokenizer.vocab.get("<\uff5c\u2581pad\u2581\uff5c>")
+    # if not hasattr(model.generation_config, "generation_kwargs"):
+    #     model.generation_config.generation_kwargs = {}
+    # model.generation_config.generation_kwargs["boi_token_id"] = tokenizer.vocab.get("<begin_of_image>")
+
```
zucchini-nlp: todo?

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
```diff
@@ -0,0 +1,230 @@
...
+    is_vision_available,
+)
+
+
+if is_torch_available():
+    import torch
+    import torchvision
+
+if is_vision_available():
+    pass
```
zucchini-nlp: can be deleted

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
```diff
@@ -0,0 +1,230 @@
...
+    auto_docstring,
+    filter_out_non_signature_kwargs,
+    is_torch_available,
+    is_vision_available,
+)
+
+
+if is_torch_available():
+    import torch
+    import torchvision
```
zucchini-nlp: now we can import these without guards in certain files, including fast image processors

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
```diff
@@ -0,0 +1,230 @@
...
+    num_crops: int = 4
+
+
+def pad_to_max_num_crops(images, max_crops):
+    """Pad each image to max crops."""
+    num_crops, _, H, W = images.shape
+    if num_crops < max_crops:
+        pad = torch.zeros(max_crops - num_crops, 3, H, W, dtype=images.dtype, device=images.device)
+        images = torch.cat([images, pad], dim=0)
+    return images
```
zucchini-nlp, line 59: can pad with `F.pad()` as well in one line, no?

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
```diff
@@ -0,0 +1,230 @@
...
+        do_resize = do_resize if do_resize is not None else self.do_resize
+        resample = resample if resample is not None else self.resample
+        do_rescale = do_rescale if do_rescale is not None else self.do_rescale
+        rescale_factor = rescale_factor if rescale_factor is not None else self.rescale_factor
+        do_normalize = do_normalize if do_normalize is not None else self.do_normalize
+        image_mean = image_mean if image_mean is not None else self.image_mean
+        image_std = image_std if image_std is not None else self.image_std
+
+        size = size if size is not None else self.size
+        images = self.fetch_images(images)
```
zucchini-nlp: we don't need these in fast processors. It's handled in `super().preprocess()`

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
```diff
@@ -0,0 +1,230 @@
...
+                stacked_images.float(),
+                size=(size["height"], size["width"]),
+                mode="bicubic",
+            )
+            processed_images_grouped[shape] = stacked_images
+        global_images = reorder_images(processed_images_grouped, grouped_images_index)
+
+        image_sizes = [[image.shape[1], image.shape[2]] for image in images]  # expecting h,w
+        num_img_tokens = [
+            int(((h // size["height"]) * (w // size["width"]) + 1) * 144 + 1 + (h // size["height"] + 1) * 12)
```
zucchini-nlp: Let's add a bit more context about where these values come from. Are these always the same or do we need to get them from config?

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
```diff
@@ -0,0 +1,230 @@
...
+            for h, w in image_sizes
+        ]
+
+        final_reshaped_images = []
+        for image, (h, w) in zip(images, image_sizes):
+            reshaped = image.reshape(1, 3, h // size["height"], size["width"], w // size["width"], size["width"])
+            permuted = reshaped.permute(0, 2, 4, 1, 3, 5)
+            reshaped_final = permuted.reshape(-1, 3, size["height"], size["width"]).contiguous()
+            final_reshaped_images.append(reshaped_final)
```
zucchini-nlp: hmm, again this doesn't do much if we resized images to `size`. If the resizing on by default for this model?

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def __init__(self, **kwargs: Unpack[Phi3VFastImageProcessorKwargs]):
+        super().__init__(**kwargs)
+
+    def get_new_scaled_dims(self, image, size):
+        height, width = image.size()[-2:]
+
+        is_transposed = False
+        if width < height:
+            image = image.transpose(1, 2)
```
zucchini-nlp: I think we have to transpose last two dims for height and width. Images at this point are batches so a 4D tensor

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
```diff
@@ -0,0 +1,230 @@
...
+            is_transposed = True
+            height, width = image.size()[-2:]
```
zucchini-nlp, line 84: huh interesting that we resize an image without taking `is_transpose` into account

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
```diff
@@ -0,0 +1,230 @@
...
+            processed_images_grouped[shape] = stacked_images
+
+        images = reorder_images(processed_images_grouped, grouped_images_index)
+
+        # Create global images from images by resizing.
+        grouped_images, grouped_images_index = group_images_by_shape(images, disable_grouping=disable_grouping)
+        processed_images_grouped = {}
+        for shape, stacked_images in grouped_images.items():
+            stacked_images = torch.nn.functional.interpolate(
+                stacked_images.float(),
```
zucchini-nlp: already a float by this point, no?

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
```diff
@@ -0,0 +1,230 @@
...
+
+        image_sizes = [[image.shape[1], image.shape[2]] for image in images]  # expecting h,w
+        num_img_tokens = [
+            int(((h // size["height"]) * (w // size["width"]) + 1) * 144 + 1 + (h // size["height"] + 1) * 12)
+            for h, w in image_sizes
+        ]
+
+        final_reshaped_images = []
+        for image, (h, w) in zip(images, image_sizes):
+            reshaped = image.reshape(1, 3, h // size["height"], size["width"], w // size["width"], size["width"])
```
zucchini-nlp: looks like a typo, `size["width"]` duplicated twice

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
zucchini-nlp: we can put this code block along with the code for obtaining `image_size`, and concating global images in a stacked for loop, simialr to how resize and other transforms are called. It will make everything a bit faster

src/transformers/models/phi3_v/processing_phi3_v.py
```diff
@@ -0,0 +1,203 @@
...
+    attributes = ["image_processor", "tokenizer"]
+    image_processor_class = "AutoImageProcessor"
+    tokenizer_class = "AutoTokenizer"
+
+    def __init__(
+        self,
+        image_processor=None,
+        tokenizer=None,
+        chat_template=None,
+        image_token="<|image|>",
```
zucchini-nlp, line 63: can we add it in tokenizer in converted models, so we don't need to have a default value

src/transformers/models/phi3_v/processing_phi3_v.py
```diff
@@ -0,0 +1,203 @@
...
+            if isinstance(text, str):
+                text = [text]
+            elif not (isinstance(text, (list, tuple)) and all(isinstance(t, str) for t in text)):
+                raise ValueError("Invalid input text. Please provide a string, or a list of strings")
+
+        # Replace the image token with expanded image tokens.
+        tokenized_prompts = []
+        image_token_counter = 0
+        for prompt in text:
+            prompt_splits = re.split(r"(\<\|image\|\>)", prompt)
```
zucchini-nlp: i think we need to make it configurable and use `self.image_token`

src/transformers/models/phi3_v/processing_phi3_v.py
```diff
@@ -0,0 +1,203 @@
...
+            tokenized_outputs = []
+            for split in prompt_splits:
+                if split == "<|image|>":
+                    if image_token_counter >= len(num_image_tokens):
+                        raise ValueError("More image placeholders in the text than images provided.")
+                    image_tokens = [self.image_token_id] * num_image_tokens[image_token_counter]
+                    tokenized_outputs.extend(image_tokens)
+                    image_token_counter += 1
+                else:
+                    text_tokens = self.tokenizer(split)["input_ids"]
```
zucchini-nlp: it's not a good idea to tokenize each split separately. Tokenizer internally does several things, including padding and truncation. It's more reliable if we don't pad/truncate manually and delegate it to a tokenizer

src/transformers/models/phi3_v/processing_phi3_v.py
```diff
@@ -0,0 +1,203 @@
...
+        """
+        return self.tokenizer.batch_decode(*args, **kwargs)
+
+    def decode(self, *args, **kwargs):
+        """
+        This method forwards all its arguments to LlamaTokenizerFast's [`~PreTrainedTokenizer.decode`]. Please refer to
+        the docstring of this method for more information.
+        """
+        return self.tokenizer.decode(*args, **kwargs)
+
```
zucchini-nlp: these are same in parent class, can be deleted

src/transformers/models/phi3_v/modular_phi3_v.py
```diff
@@ -0,0 +1,416 @@
...
+    """
+    An abstract class to handle weights initialization and a simple interface for downloading and loading pretrained
+    models. The model is only intended for inference and doesn't support finetuning.
+    """
+
+    config: Phi3VConfig
+    base_model_prefix = "model"
+    input_modalities = ["image", "text"]
+    supports_gradient_checkpointing = True
+    _no_split_modules = ["LlamaDecoderLayer"]
```
zucchini-nlp: not needed, we recursively merge all `_no_split_modules` from child modules in `PreTrainedModel`

src/transformers/models/phi3_v/modular_phi3_v.py
```diff
@@ -0,0 +1,416 @@
...
+        super().__init__(config)
+        self.config = config
+        self.image_dim_out = config.vision_config.hidden_size
+
+        self.vision_model = AutoModel.from_config(config.vision_config)
+        self.language_model = AutoModel.from_config(config.text_config)
+        self.image_projection = Phi3VImageProjection(config)
+
+        self.glb_GN = nn.Parameter(torch.zeros([1, 1, self.image_dim_out * 4]))
+        self.sub_GN = nn.Parameter(torch.zeros([1, 1, 1, self.image_dim_out * 4]))
```
zucchini-nlp: let's name better. global and local newline?

src/transformers/models/phi3_v/modular_phi3_v.py
```diff
@@ -0,0 +1,416 @@
...
+    def get_input_embeddings(self):
+        return self.language_model.get_input_embeddings()
+
+    def set_input_embeddings(self, value):
+        self.language_model.set_input_embeddings(value)
+
+    def reshape_hd_patches_2x2merge(self, image_features, h_crop, w_crop):
+        """Reshape high-dimensional patches by merging 2x2 patches."""
+        N, L, C = image_features.shape  # Shape: (num_images*num_crops, 24*24, 1024)
+        H = int(L**0.5)
```
zucchini-nlp: nit: let's use longer names for var

src/transformers/models/phi3_v/modular_phi3_v.py
```diff
@@ -0,0 +1,416 @@
...
+            num_images, h_crop * H // 2, w_crop * H // 2, 4 * C
+        )  # (num_images, h_crop*12, w_crop*12, 4096)
+
+        return image_features_hd
+
+    def add_newline_embeds(self, image_features):
+        """Add the newline token embeds to the image feature patches"""
+        num_images, h, w, hid_dim = image_features.shape
+        newline_embeddings = self.sub_GN.expand(num_images, h, -1, -1)
+        image_features_newline = torch.cat([image_features, newline_embeddings], dim=2)
```
zucchini-nlp, line 219: we might need to adjust the devices and dtypes before merging

src/transformers/models/phi3_v/modular_phi3_v.py
```diff
@@ -0,0 +1,416 @@
...
+            )
+        if inputs_embeds is None:
+            inputs_embeds = self.get_input_embeddings()(input_ids)
+
+        if pixel_values is not None:
+            num_images, num_crops, c, h, w = pixel_values.shape
+            pixel_values = pixel_values.flatten(0, 1)
+            image_features = self.get_image_features(pixel_values, image_sizes, num_images, num_crops)
+            image_attention_mask = self.get_placeholder_mask(input_ids, inputs_embeds, image_features)
+            inputs_embeds = inputs_embeds.masked_scatter(image_attention_mask, image_features)
```
zucchini-nlp, line 322: same, need to make sure `image_features` is in correct device and has same dtype as inputs embeds

src/transformers/models/phi3_v/modular_phi3_v.py
```diff
@@ -0,0 +1,416 @@
...
+            attention_mask=attention_mask,
+            cache_position=cache_position,
+            logits_to_keep=logits_to_keep,
+            **kwargs,
+        )
+
+        if cache_position[0] == 0:
+            model_inputs["pixel_values"] = pixel_values
+
+        model_inputs["image_sizes"] = image_sizes
```
zucchini-nlp: then we can pass it to super and it will be returned in `model_inputs` without modifications

tests/models/phi3_v/test_image_processing_phi3_v.py
```diff
@@ -0,0 +1,129 @@
...
+import unittest
+
+from transformers.testing_utils import require_torch, require_vision
+from transformers.utils import is_torch_available, is_torchvision_available, is_vision_available
+
+from ...test_image_processing_common import ImageProcessingTestMixin, prepare_image_inputs
+
+
+if is_torch_available():
+    pass
```
zucchini-nlp: can be deleted

tests/models/phi3_v/test_processing_phi3_v.py
```diff
@@ -0,0 +1,115 @@
...
+        processor_dict_loaded = json.loads(processor_loaded.to_json_string())
+        # chat templates aren't serialized to json in processors
+        self.assertFalse("chat_template" in processor_dict_loaded)
+
+        # they have to be saved as separate file and loaded back from that file
+        # so we check if the same template is loaded
+        processor_dict = self.prepare_processor_dict()
+        self.assertTrue(processor_loaded.chat_template == processor_dict.get("chat_template", None))
+
+    @unittest.skip("Not possible as processor creates a custom attention mask.")
```
zucchini-nlp: mask format doesn't look custom, even though prepared manually instead of passing to tokenizer

zucchini-nlp (commented):
@yaswanth19 thanks for the PR, looks much cleaner already!

I left some comments, mostly nitty-picking for better standardization. Also I believe there's one test failing with Phi3.5V :)

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
yaswanth19: Yep correct.

src/transformers/models/phi3_v/processing_phi3_v.py
yaswanth19: Yep I agree with that and it makes the code much simpler, but in the original implementation [[LINK](https://huggingface.co/microsoft/Phi-3.5-vision-instruct/blob/main/processing_phi3_v.py#L414)] the prompt is tokenized that way, This way we add a bos and eos id around each image and I am afraid if we delegate it to tokenizer directly then this pattern will be missed and performance might take a hit. IMO, for inference we it mostly aligns with original implementation and for fine-tuning a user can process the prompts in the dataset class

src/transformers/models/phi3_v/processing_phi3_v.py
yaswanth19: Added in conversion script but any reason to remove the default value from args, I can see in quite a few models and it also gives users some flexibility to set custom image tokens 🤔

tests/models/phi3_v/test_processing_phi3_v.py
yaswanth19: I am skipping this test because it requires offset mapping which is quite difficult to fetch because of the way we tokenize the prompt.

yaswanth19:
@zucchini-nlp Ready for another review. The test failures are unrelated and can you please also trigger slow CI here.

yaswanth19:
@zucchini-nlp Can you trigger the slow tests so that I can push changes for Multi-GPU tests if required and you can do the review later based on your bandwidth.

zucchini-nlp:
run-slow: phi3_v

zucchini-nlp:
great, reviewing tomorrow!

src/transformers/models/phi3_v/processing_phi3_v.py
zucchini-nlp: So each image gets replaces with a `BOI+image+EOI`. We can delegate it to chat template, no? Actually we can replace each image placeholder as we wish, so should not be problem unless I am missing smth

Currently this is the biggest comment I have, and I want us to try and call the tokenizer only once

docs/source/en/model_doc/phi3_v.md
```diff
@@ -0,0 +1,95 @@
...
+<div style="float: right;">
+    <div class="flex flex-wrap space-x-1">
+        <img alt="PyTorch" src="https://img.shields.io/badge/PyTorch-DE3412?style=flat&logo=pytorch&logoColor=white">
+        <img alt="FlashAttention" src="https://img.shields.io/badge/%E2%9A%A1%EF%B8%8E%20FlashAttention-eae0c8?style=flat">
+        <img alt="SDPA" src="https://img.shields.io/badge/SDPA-DE3412?style=flat&logo=pytorch&logoColor=white">
+    </div>
+</div>
+
+# Phi-3.5 Vision
+
```
zucchini-nlp, line 27: nit: can we add a small intro for the model here before the abstract?

src/transformers/models/phi3_v/convert_phi3_v_weights_to_hf.py
```diff
@@ -0,0 +1,359 @@
+# coding=utf-8
```
zucchini-nlp, line 1: not sure if Microsoft will be willing to host converted weights on the hub. We just recently added a dynamic weight converter as public API, I think it will be the way-to-go for Phi3V

If you want to play around with it and update conversion, [this is the file](https://github.com/vasqu/transformers/blob/main/src/transformers/conversion_mapping.py) that needs to be modified. IMO it's totally fine if you are out of bandwidth, we'll add conversion mapping after v5 for this model and Phi4-Multimodal

src/transformers/models/phi3_v/image_processing_phi3_v_fast.py
```diff
@@ -0,0 +1,209 @@
...
+            reshaped = image.reshape(1, 3, h // size["height"], size["height"], w // size["width"], size["width"])
+            permuted = reshaped.permute(0, 2, 4, 1, 3, 5)
+            reshaped_final = permuted.reshape(-1, 3, size["height"], size["width"]).contiguous()
+            combined_image = torch.cat([global_image.unsqueeze(0), reshaped_final], axis=0)
+            images_combined.append(combined_image)
+            image_sizes.append([h, w])
+
+        # Calculate the number of image tokens for each image based on height and width dynamically.
+        num_img_tokens = [
+            int(((h // size["height"]) * (w // size["width"]) + 1) * 144 + 1 + (h // size["height"] + 1) * 12)
```
zucchini-nlp, line 194: i think it's still nice to comment clearly where 144 comes from. Ig it has to do with the model's patch size when embedding image features or with pooling stride

src/transformers/models/phi3_v/processing_phi3_v.py
zucchini-nlp: we don't want to have two sources of truth for the same attribute, so if users want to change an `image_token` they can do so in the tokenizer. For ex `tokenizer.image_token_id = 10000`

src/transformers/models/phi3_v/processing_phi3_v.py
```diff
@@ -0,0 +1,160 @@
...
+            The tokenizer is a required input.
+        chat_template (`str`, *optional*): A Jinja template which will be used to convert lists of messages
+            in a chat into a tokenizable string.
+        image_token (`str`, *optional*, defaults to `"<|image|>"`): Image token to be used in the text input.
+    """
+
+    attributes = ["image_processor", "tokenizer"]
+    image_processor_class = "AutoImageProcessor"
+    tokenizer_class = "AutoTokenizer"
+
```
zucchini-nlp, line 57: these are not needed after the recent refactor, we will infer dynamically the attributes from the `__init__` signature and always load from auto-mapping

src/transformers/models/phi3_v/processing_phi3_v.py
```diff
@@ -0,0 +1,160 @@
...
+                # If the text is an image token, replace it with the corresponding number of image tokens.
+                if text == self.image_token:
+                    if image_token_counter >= len(num_image_tokens):
+                        raise ValueError("More image placeholders in the text than images provided.")
+                    text = "".join([self.image_token] * num_image_tokens[image_token_counter])
+                    image_token_counter += 1
+
+                outputs.append(text)
+            prompt_strings.append("".join(outputs))
+
```
zucchini-nlp, line 146: much better than the official impl, thanks for updating it!

src/transformers/models/phi3_v/modular_phi3_v.py
```diff
@@ -0,0 +1,428 @@
...
+            logger.info("`text_config` is None. Initializing with default values")
+            self.text_config = CONFIG_MAPPING["phi3"]()
+        elif isinstance(text_config, dict):
+            text_config["model_type"] = text_config.get("model_type", "phi3")
+            self.text_config = CONFIG_MAPPING[text_config["model_type"]](**text_config)
+        elif isinstance(text_config, PretrainedConfig):
+            self.text_config = text_config
+        else:
+            raise ValueError(
+                f"Invalid type for `text_config`. Must be either `dict` or `LlamaConfig`."
```
zucchini-nlp, line 96: nit: ig the message should ask for a `Phi3Config`, not `LlamaConfig`

src/transformers/models/phi3_v/modular_phi3_v.py
zucchini-nlp: not addressed. Image feats and the newline embed can end up in different devices with mutliGPU or dtypes if user changes config's `torch_dtype` param 

We can move the newline to the same device/dtype as image embeddings

src/transformers/models/phi3_v/modular_phi3_v.py
zucchini-nlp: I see where it can go wrong, generation has a huge bias towards text-like inputs and assumes the first dimension is batch size. Oke, let's skip for now, it's not a super common feature

tests/models/phi3_v/test_modeling_phi3_v.py
```diff
@@ -0,0 +1,412 @@
...
+    @unittest.skip("Not possible now as processor creates a custom attention mask.")
+    def test_prompt_lookup_decoding_matches_greedy_search(self):
+        pass
+
+    @unittest.skip("Not possible now as processor creates a custom attention mask.")
+    def test_assisted_decoding_sample(self):
+        pass
+
+    @unittest.skip("Not possible now as processor creates a custom attention mask.")
+    def test_apply_chat_template_assistant_mask(self):
```
zucchini-nlp, line 193: the skip reason needs an update or we can delete the skips if it's already supported

tests/models/phi3_v/test_processing_phi3_v.py
```diff
@@ -0,0 +1,115 @@
...
+        self.assertFalse("chat_template" in processor_dict_loaded)
+
+        # they have to be saved as separate file and loaded back from that file
+        # so we check if the same template is loaded
+        processor_dict = self.prepare_processor_dict()
+        self.assertTrue(processor_loaded.chat_template == processor_dict.get("chat_template", None))
+
+    @unittest.skip("Not possible as processor can't create an assistant mask.")
+    def test_apply_chat_template_assistant_mask(self):
+        pass
```
zucchini-nlp, line 88: same here, i expect it works now after changing the code for placeholder expansion

tests/models/phi3_v/test_processing_phi3_v.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def test_unstructured_kwargs_batched(self):
```
zucchini-nlp, line 90: curious why we needed to overwrite one test while other similar tests are passing? A comment explaining the reason would be great

zucchini-nlp (commented):
Thanks a lot for iterating! The PR looks good overall with a few nits to address. After resolving these comment, I think you can request a review from core maintainers (ArthurZucker or Cyrilvallez)

We are currently focusing on the major v5 release so the core maintainer review can be a bit delayed. Prob the model will be merged after the release. Also there was a big refactor on weight loading recently, and if your tests are failing for unrelated reasons feel free to rebase

PR #41975, opened 2025-11-02
Fix `detectron2` installation in docker files

ydshieh:
# What does this PR do?

See the issue reported in https://github.com/facebookresearch/detectron2/issues/5495

`--no-build-isolation` fixes the issue.

Merge directly.

The effect is demonstrated in https://github.com/huggingface/transformers/actions/runs/19012508548

PR #41978, opened 2025-11-02
Fix `autoawq[kernels]` installation in quantization docker file

ydshieh:
# What does this PR do?

Same as #41975 but for `autoawq[kernels]`.

Merge directly.

The effect :

before this PR: https://github.com/huggingface/transformers/actions/runs/19015759405
after this PR: https://github.com/huggingface/transformers/actions/runs/19014860804

cc @MekkCyber


MekkCyber (commented):
Thanks for fixing

PR #41980, opened 2025-11-02
Correct type hint in config models

jorisSchaller:
# What does this PR do?
Most models are using a config with `rms_norm_eps: Optional[int] = 1e-6`: I modified it to an optional float.
This update configs that use a type of optional int instead of optional float for the epsilon value for their RMS norm.
You can clearly see that the epsilon has a default value of 10e-6 (float), the hint should therefore be adjusted.

FIx no issue.

## Before submitting
- [X] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

## Who can review?

Let me know if you want any changes,  @ArthurZucker @Cyrilvallez @stevhliu, I tried to find all occurence of it, but some might have slipped through.


Rocketknight1:
Looks good but please run `make fix-copies` to propagate modular changes! Ping me again whenever the CI is green and I'll merge it

PR #41985, opened 2025-11-03
Fix `torch+deepspeed` docker file

ydshieh:
# What does this PR do?

Similar to #41978 and #41975, but with extra issue of `--global-option=`. All issues are caused by the new pip version `25.3`.

Verified the docker file could be built and the `deepspeed` CI job could run.

### DIFF
--- a/.github/workflows/build-docker-images.yml
+++ b/.github/workflows/build-docker-images.yml
@@ -97,7 +97,7 @@ jobs:
   latest-torch-deepspeed-docker:
     name: "Latest PyTorch + DeepSpeed"
     runs-on:
-      group: aws-g4dn-2xlarge-cache
+      group: aws-general-8-plus
     steps:
       -
         name: Set up Docker Buildx
--- a/docker/transformers-pytorch-deepspeed-latest-gpu/Dockerfile
+++ b/docker/transformers-pytorch-deepspeed-latest-gpu/Dockerfile
@@ -8,6 +8,9 @@ ARG PYTORCH='2.8.0'
 # Example: `cu102`, `cu113`, etc.
 ARG CUDA='cu126'
 
+# This needs to be compatible with the above `PYTORCH`. Let's use cpu version for now.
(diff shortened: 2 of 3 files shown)


ydshieh:
change it to draft as there are more things to fix


In the deepspeed docker, we also have `torchcodec`

> RUN python3 -m pip uninstall -y torch torchvision torchaudio && python3 -m pip install --no-cache-dir -U torch==$PYTORCH torchvision torchaudio torchcodec --extra-index-url https://download.pytorch.org/whl/$CUDA

.github/workflows/build-docker-images.yml
```diff
@@ -97,7 +97,7 @@ jobs:
   latest-torch-deepspeed-docker:
     name: "Latest PyTorch + DeepSpeed"
     runs-on:
-      group: aws-g4dn-2xlarge-cache
+      group: aws-general-8-plus
```
ydshieh, line 100: we are not using T4 runners anymore. Since the CPU runners work now for building this image, let's use it.

docker/transformers-pytorch-deepspeed-latest-gpu/Dockerfile
```diff
@@ -8,6 +8,9 @@ ARG PYTORCH='2.8.0'
 # Example: `cu102`, `cu113`, etc.
 ARG CUDA='cu126'
 
+# This needs to be compatible with the above `PYTORCH`. Let's use cpu version for now.
+ARG TORCHCODEC='0.7.0'
```
ydshieh: previously (around 2025/07), this line likely had `soundfile` and being replaced by `torchcodec`.

If you prefer to simply remove `torchcodec` in this docker file, let me know.

SunMarc: let's remove it in this dockerfile !

SunMarc (approved):
Thanks, just a nit

PR #41986, opened 2025-11-03
add xpu support in test_modeling_janus.py::JanusIntegrationTest::test…

sywangyi:
…_model_generate_images

add xpu support in tests/models/janus/test_modeling_janus.py::JanusIntegrationTest::test_model_generate_images


sywangyi:
@ydshieh please help review

sywangyi:
The change in modeling is because:  xpu set allow_is_causal_skip to True for prefill in static cache case for performance optimization in sdpa, as a result the attention_mask may not be contained in model_inputs.

sywangyi:
Intel XPU: @IlyasMoutawwakil

yao-matrix:
@sywangyi , pls fix CI issues, thx.

ydshieh (approved):
sound good, so it is kind of xpu specific, but general users might have use that flag too, so make sense.

Thank you. Please follow @yao-matrix to fix the failing tests. (about the modular files)

PR #41892, opened 2025-10-27
Update some workflow files

ydshieh:
# What does this PR do?

Mostly:

- Make `docker/transformers-all-latest-gpu/Dockerfile` more readable and clean as we now need to handle `torchcodec` (using `cpu`) along with `torch` (`cuda`)

- Remove `push-ci` stuff. We are not paying any attention to it. We have something running on a very small subset now.
- Separate CI workflows and their docker images: with `flash-attn` and without it

PR #41988, opened 2025-11-03
Fix `torchcodec` version in quantization docker file

ydshieh:
# What does this PR do?

The `torchcodec` version is not bind automatically to `torch` version during the installation. Currently we get `torch 2.8` (as we pin it) and `torchcodec 0.8.1` (which is for `torch 2.9`).

This PR follows #41892 to allow specifying `torchcodec` version.

Previously, the quantization jobs fails 

https://github.com/huggingface/transformers/actions/runs/18862400780/job/53850645716


MekkCyber (approved):
Thanks for fixing

MHRDYN7:
@ydshieh coincidentally, I was just working on this version mismatch and was about to open an issue. It seems that even with this current set of versions torch 2.9 & torchcodec 0.8.1, google colab crashes while importing AutoVideoProcesser or any model specific video processor. Are there any tests that check these imports? I noticed this when I found that the example code for VJEPA2 crashes on colab.

SunMarc (approved):
Let's just remove torchdec as this is not used for our tests

PR #41990, opened 2025-11-03
feat(kernels): add opt-out flag to disable kernels hub usage through the lib

mfuntowicz:
This PR attempts to bring an opt-out mecanism to disable hub kernels (aka `kernels`) globally through `transformers`. 

The default value for the `USE_HUB_KERNELS` environment variable defaults to `YES` to maintain backward compatiblity with the current machinery.

In more details, it works by proxying the `kernels.use_kernel_forward_from_hub` method to include our `_kernels_enabled` branching logic. In case of  `_kernels_enabled == False`, the function `use_kernel_forward_from_hub` will return an identity decorator which should just default to the current class the decorator is attached to.

I added tests to a new folder `integrations.tests_hub_kernels.py` to match the layout structure of the sources folder. 
Let me know if it was not the correct way to do, will swizzle at the right place.

### DIFF
--- a/src/transformers/integrations/hub_kernels.py
+++ b/src/transformers/integrations/hub_kernels.py
@@ -19,7 +19,7 @@ from types import ModuleType
 from typing import Optional, Union
 
 from ..modeling_flash_attention_utils import lazy_import_flash_attention
-from ..utils import logging, ENV_VARS_TRUE_VALUES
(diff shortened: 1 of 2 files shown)


ArthurZucker (approved):
LGTM, I would probably test with a full model as well, just to be sure (Like Llama4 or gpt oss) but fine as is! 

As you mentioned offline, doc is coming for this as well!

PR #41993, opened 2025-11-03
Attempt to fix VLM gradient enabling

molbap:
# What does this PR do?

As per title. Linked to https://github.com/huggingface/peft/issues/2880.
Follows more or less closely the already existing implementations for idefics2-3 and smolvlm, trying to cover several types of VLMs (they are named differently across the lib.)


tests/models/qwen2_vl/test_modeling_qwen2_vl.py
```diff
@@ -397,6 +397,58 @@ def test_sdpa_can_dispatch_on_flash(self):
...
+
+            vision_module = None
+            if hasattr(model, "visual"):
+                vision_module = model.visual
+            elif hasattr(model, "model") and hasattr(model.model, "visual"):
+                vision_module = model.model.visual
+
+            if vision_module is not None:
+                for parameter in vision_module.parameters():
+                    parameter.requires_grad = True
```
githubnemo: From my understanding of peft#2880, the problem is mainly that the entry point of the model doesn't require gradients (not a trainable parameter, just for gradient checkpointing) so that targeting modules after that doesn't work with reentrant gradient checkpointing. Isn't setting all vision parameters to `requires_grad=True` masking the changes done in `enable_input_requires_grad` and therefore always true, regardless of what that helper function does? Maybe targeting something that is clearly not an input, something resembling an attention layer for example, is better?

molbap: I see, hmm- followed the implem of idefics2/smolvlm as I remembered they faced this issue at the time. You're right that this isn't necessary, we register twice. The lowest module trick should work though, and I'm not sure targeting an attention layer works either. Currently @BenjaminBossan 's script outputs grad norms properly with gradient checkpointing enabled and PEFT disabled on this branch, so it seems to do the trick?

no GC
```bash
{'loss': 9.4971, 'grad_norm': 23.421083450317383, 'learning_rate': 2e-05, 'epoch': 0.33}                                                                     
{'loss': 7.9526, 'grad_norm': 675.1868896484375, 'learning_rate': 1.866666666666667e-05, 'epoch': 0.67}                                                      
```

with GC
```bash
{'loss': 9.4971, 'grad_norm': 23.421083450317383, 'learning_rate': 2e-05, 'epoch': 0.33}
 {'loss': 7.9526, 'grad_norm': 675.1868896484375, 'learning_rate': 1.866666666666667e-05, 'epoch': 0.67} 
```


in either case, agree double registering is useless, will remove!

githubnemo: Yeah, I think the implementation is fine. I'm just worried that the test is masking the behavior of the fix and is therefore not honest enough. Sorry if I didn't make that clear.

molbap: No that's fair, I'll revamp the test for a narrower scope!

zucchini-nlp (commented):
I think this solution works only for VLMs and also depends a lot on how the vision model is named. I'm sure we listed all possible names, but new models can get creative with it

So I'm thinking that we could potentially make it works ootx for all MLLMS (audio/vision/omni) by checking for each `PreTrainedModel` within the model and then setting grads on that models' inputs (`model.get_input_embeddings()`)

We use similar trick when setting attention implementations and check for `PreTrainedModel`'s, so it could be a good option. WDYT?

molbap:
Thanks, yes it's a far less brittle option. There's a few (really a few and hopefully should be 0 after v5) modules that were just `nn.Module`s instead of `PreTrainedModel` so they would be off the hook, other than these few exceptions should work out well, will push something like that today

githubnemo:
> So I'm thinking that we could potentially make it works ootx for all MLLMS (audio/vision/omni) by checking for each `PreTrainedModel` within the model and then setting grads on that models' inputs (`model.get_input_embeddings()`)
> 
> We use similar trick when setting attention implementations and check for `PreTrainedModel`'s, so it could be a good option. WDYT?

Sorry I may misunderstand the proposed solution but this doesn't seem to solve the problem? In a VLM where I target a module in the vision stack I need to have the vision model's inputs require grads, not the language model's input (`get_input_embeddings`).

zucchini-nlp:
@githubnemo `model.get_input_embeddings()` will return the lowest module for each models, which is the same thing current PR does by recursing over modules. So setting grads on inputs of `model.get_input_embeddings()` will enable it not only for for vision-modules, but also for audio and other modalities

zucchini-nlp:
BTW, when working on smth else noticed that we have code like below which can be deleted after this PR

https://github.com/huggingface/transformers/blob/80134e6e663db962c30d0e45c54265b3141bc7c7/src/transformers/models/idefics2/modeling_idefics2.py#L1026-L1030

molbap:
Yes, for all idefics/smolvlm there'll be no need for that. Should ship that today (finally)

molbap:
Iterated a bit on that and hit a dead end on idefics2/3 code, back at it tomorrow!

molbap:
Tests failing appear unrelated (I rebased on main), what do you think of the new method @zucchini-nlp ? Also @githubnemo I updated the test a tad, let me know

src/transformers/models/idefics2/modeling_idefics2.py
```diff
@@ -802,36 +802,6 @@ def __init__(self, config: Idefics2Config):
 
         self.post_init()
 
-    def enable_input_require_grads(self):
-        """
-        Enables the gradients for the input embeddings.
-
-        This is useful for lora when using gradient checkpointing.
-        c.f. https://github.com/huggingface/peft/issues/1402#issuecomment-1913675032
```
zucchini-nlp, line 795: nice 🔪

src/transformers/modeling_utils.py
```diff
@@ -2214,13 +2214,44 @@ def enable_input_require_grads(self):
...
+            if embedding_id in seen_modules:
+                continue
+
+            seen_modules.add(embedding_id)
+            hooks.append(input_embeddings.register_forward_hook(make_inputs_require_grads))
+
+        self._require_grads_hooks = hooks
+        if hooks:
+            # for BC
+            self._require_grads_hook = hooks[0]
```
zucchini-nlp, line 2001: aren't we ignoring all hooks except for the first one in this case, i.e. when we disable it will disable the text model and will not disable vision model?

src/transformers/modeling_utils.py
```diff
@@ -2214,13 +2214,44 @@ def enable_input_require_grads(self):
...
+            if input_embeddings is None:
+                continue
+
+            embedding_id = id(input_embeddings)
+            if embedding_id in seen_modules:
+                continue
+
+            seen_modules.add(embedding_id)
+            hooks.append(input_embeddings.register_forward_hook(make_inputs_require_grads))
+
```
zucchini-nlp, line 1997: super clean!

src/transformers/modeling_utils.py
```diff
@@ -2214,13 +2214,44 @@ def enable_input_require_grads(self):
...
+        hooks = getattr(self, "_require_grads_hooks", None)
+        if not hooks:
+            return
+
+        for hook in hooks:
+            hook.remove()
+
+        self._require_grads_hooks = []
+        if hasattr(self, "_require_grads_hook"):
+            del self._require_grads_hook
```
zucchini-nlp, line 2016: out of curiosity, is it required to explicitly delete?

tests/models/qwen2_vl/test_modeling_qwen2_vl.py
```diff
@@ -397,6 +397,62 @@ def test_sdpa_can_dispatch_on_flash(self):
     def test_multi_gpu_data_parallel_forward(self):
         pass
 
+    def test_enable_input_require_grads_with_gradient_checkpointing(self):
```
zucchini-nlp, line 365: i am thinking, if we can make a common test for all models?

zucchini-nlp (commented):
Love the clean-up! Only one major q about tests, would be super super cool to have a common test imo. Though I realize it can be hard with multimodals

src/transformers/modeling_utils.py
molbap: I don't think so, this is just because we used to remove `_require_grads_hook`, now we always iterate over the full list `_require_grads_hooks` (with an `s`) so every registered hook (vision or text or whatever) should be removed

zucchini-nlp: ahh my bad, didn't see the "s" at the end

src/transformers/modeling_utils.py
molbap: Just out of safety, not certain it's always necessary but not knowing what people were doing with that hook in their FT scripts I think it's safer to remove it so no reference remains

tests/models/qwen2_vl/test_modeling_qwen2_vl.py
molbap: eeeh... I think we should :D yes
will look before EOD if I have time

src/transformers/modeling_utils.py
molbap: might be a bad naming then haha

ArthurZucker (approved):
Very nice!

molbap:
run-slow: bart, blip_2, idefics2, idefics3, qwen2_5_omni, qwen2_5_vl, qwen2_vl, qwen3_omni_moe, smolvlm, timm_wrapper

ArthurZucker (approved):
Very nice! this fixes #42489 (#42494) as well I think

PR #41994, opened 2025-11-03
DOC: Fix typo in argument name: pseudoquant

BenjaminBossan:
# What does this PR do?

The correct argument name is `pseudoquantization`. Since there is no error on passing wrong argument names (which is arguably an anti-pattern), this is difficult for users to debug.

## Before submitting


SunMarc (approved):
Thanks for spotting it ! LGTM

PR #41996, opened 2025-11-03
[v5] Deprecate Text2Text and related pipelines

Rocketknight1:
Text2Text is mostly used only by older models, and via the Summarization/Translation pipelines. However, all of these tasks are now handled by general-purpose LLMs that provide much higher translation/summarization quality than older single-purpose models. This PR retains the pipelines, but deprecates them with a warning.


LysandreJik (approved):
Thank you!

PR #40735, opened 2025-09-05
[torchao safetensors] integrate torchao safetensors support with transformers 

liangel-02:
**Context**

Currently, we need to use `safe_serialization=False` while saving models as shown [here](https://huggingface.co/pytorch/Qwen3-8B-INT4#quantization-recipe). This PR enables safetensors support for torchao so that users can now save and load checkpoints using safetensors. Currently, only Float8Tensor is supported (`Float8DynamicActivationFloat8WeightConfig`, `Float8WeightOnlyConfig`) but allowing other subclasses should involve minimal code changes.

```
# default forsafe_serialization is True
quantized_model.push_to_hub(save_to)
```

**Summary**

Changes to transformers code includes: 

1.  In ```TorchAoHfQuantizer```, we provide ```get_state_dict``` and ```update_state_dict_with_metadata``` that flattens/unflattens a model state dict with tensor subclasses by calling functionality built out in [this PR](https://github.com/pytorch/ao/pull/2881). 
2. In ```modeling_utils.py```, we make appropriate changes to support propagating the metadata from tensor subclasses. We also add logic similar to `hqq` and `bnb` to directly load onto `cpu` rather than `meta`.

**Test Plan**

Modified unit test to allow safe serialization. Run using ```python tests/quantization/torchao_integration/test_torchao.py```

Reference https://huggingface.co/torchao-testing/opt-125m-Float8WeightOnlyConfig-v2-0.14.0.dev-safetensors for an example of a serialized model and test script

PR #41998, opened 2025-11-03
[torchao] fix safetensors and enable loading from sharded files 

liangel-02:
**Context**

This PR is a followup to #40735 and #41138. Previously, we enabled safetensors in torchao for one shard file. This PR fixes some errors introduced in #41138 and handles the case when checkpoints are sharded onto more than one file, including the edge case where a single quantized tensor (ie `Float8Tensor`) is sharded onto two different files (ie `qdata` on one and `scale` on another). 

**Summary**

If we are loading in a component of a tensor subclass in `create_quantized_param()` called by `_load_state_dict_into_meta_model()`, we add this as a new parameter into the model. Then after all parameters are loaded, we unflatten the `state_dict` and reassign the model parameters. 

**Testing**

Modified unit tests to test all tensor subclasses 
`python tests/quantization/torchao_integration/test_torchao.py -k TorchAoSafeSerializationTest`

### DIFF
--- a/src/transformers/quantizers/base.py
+++ b/src/transformers/quantizers/base.py
@@ -117,6 +117,7 @@ class HfQuantizer(ABC):
 
     def __init__(self, quantization_config: QuantizationConfigMixin, **kwargs):
         self.quantization_config = quantization_config
+        self.metadata = {}
 
         # -- Handle extra kwargs below --
         self.modules_to_not_convert = kwargs.pop("modules_to_not_convert", [])
@@ -391,10 +392,6 @@ class HfQuantizer(ABC):
     def get_state_dict_and_metadata(self, model, safe_serialization=False):
         """Get state dict and metadata. Useful when we need to modify a bit the state dict due to quantization"""
         return None, {}
-
-    def update_state_dict_with_metadata(self, state_dict, metadata):
-        """Update state dict with metadata. Default behaviour returns state_dict"""
-        return state_dict
 
     @abstractmethod
     def is_serializable(self, safe_serialization=None): ...
--- a/src/transformers/quantizers/quantizer_torchao.py
+++ b/src/transformers/quantizers/quantizer_torchao.py
@@ -14,7 +14,6 @@
 import importlib
 import re
 import types
-from collections import defaultdict
 from typing import TYPE_CHECKING, Optional, Union
 
 from packaging import version
@@ -38,13 +37,13 @@ if is_torch_available():
 if is_torchao_available():
     import torchao
 
-    if version.parse(importlib.metadata.version("torchao")) >= version.parse("0.14.0"):
+    if version.parse(importlib.metadata.version("torchao")) >= version.parse("0.15.0"):
+        from torchao.prototype.awq import AWQConfig
         from torchao.prototype.safetensors.safetensors_support import (
             flatten_tensor_state_dict,
             unflatten_tensor_state_dict,
         )
         from torchao.prototype.safetensors.safetensors_utils import is_metadata_torchao
-
 
 logger = logging.get_logger(__name__)
 
@@ -87,6 +86,11 @@ if is_torchao_available():
     SUPPORTED_SAFE_SERIALIZATION_CONFIGS = [
         torchao.quantization.Float8WeightOnlyConfig,
         torchao.quantization.Float8DynamicActivationFloat8WeightConfig,
+        torchao.quantization.Int4WeightOnlyConfig,
+        torchao.quantization.IntxWeightOnlyConfig,
+        torchao.quantization.Int8DynamicActivationIntxWeightConfig,
+        torchao.quantization.ModuleFqnToConfig,
+        AWQConfig,
     ]
 
     TORCHAO_VERSION = version.parse(importlib.metadata.version("torchao"))
@@ -103,20 +107,6 @@ class TorchAoHfQuantizer(HfQuantizer):
 
     def __init__(self, quantization_config, **kwargs):
         super().__init__(quantization_config, **kwargs)
-
-        if isinstance(self.quantization_config.quant_type, str):
-            is_int_4 = "int4" in self.quantization_config.quant_type
-        else:
-            config_name = self.quantization_config.quant_type.__class__.__name__
-            is_int_4 = fuzzy_match_size(config_name) == "4"
-
-        # TODO: better way to get the serialized key names? Hard to read from torchao codebase
-        if is_int_4:
-            self.weight_ao_keys = ["qdata", "scale", "zero_point"]
-        else:
-            self.weight_ao_keys = ["qdata", "scale"]
-        # Instead of serializing the simple torch.Tensor like usual, torchao adds a `:_data` suffix so we need this
-        self.full_ao_keys = self.weight_ao_keys + ["_data"]
 
     def validate_environment(self, *args, **kwargs):
         if not is_torchao_available():
@@ -168,11 +158,11 @@ class TorchAoHfQuantizer(HfQuantizer):
         the safetensors format.
         """
         if type(self.quantization_config.quant_type) in SUPPORTED_SAFE_SERIALIZATION_CONFIGS and safe_serialization:
-            if TORCHAO_VERSION >= version.parse("0.14.0"):
+            if TORCHAO_VERSION >= version.parse("0.15.0"):
                 return flatten_tensor_state_dict(model.state_dict())
             else:
                 raise RuntimeError(
-                    f"In order to use safetensors with torchao, please use torchao version >= 0.14.0. Current version: {TORCHAO_VERSION}"
+                    f"In order to use safetensors with torchao, please use torchao version >= 0.15.0. Current version: {TORCHAO_VERSION}"
                 )
         else:
             return None, {}
@@ -234,7 +224,7 @@ class TorchAoHfQuantizer(HfQuantizer):
         return
 
     def update_unexpected_keys(self, model, unexpected_keys: list[str]) -> list[str]:
-        return [k for k in unexpected_keys if not any(k.endswith(x) for x in self.full_ao_keys)]
+        return [k for k in unexpected_keys if "_weight_" not in k]
 
     def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **kwargs) -> bool:
         if self.quantization_config.quant_type == "autoquant":
@@ -243,7 +233,7 @@ class TorchAoHfQuantizer(HfQuantizer):
         # check if the param_name is not in self.modules_to_not_convert
         if any(key + "." in param_name or key == param_name for key in self.modules_to_not_convert):
             return False
-        elif any(param_name.endswith(f":{x}") for x in self.full_ao_keys):
+        elif "_weight_" in param_name:
             return True
         else:
             # we only quantize the weight of nn.Linear and nn.Embedding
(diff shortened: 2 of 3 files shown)


src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -269,38 +287,27 @@ def create_quantized_param(
...
-            self.ao_params[param_name].update({full_name: param_value})
-
-            # We are ready for quantization in this case (we retrieved all the needed keys)
-            if len(self.ao_params[param_name]) == len(self.weight_ao_keys):
-                new_param = unflatten_tensor_state_dict(self.ao_params[param_name], self.metadata)[param_name]
-                # Set it
-                module._parameters[tensor_name] = torch.nn.Parameter(
-                    new_param.to(target_device), requires_grad=new_param.requires_grad
+            elif TORCHAO_VERSION >= version.parse("0.14.0") and is_metadata_torchao(self.metadata):
+                module._parameters[tensor_name + suffix] = torch.nn.Parameter(
```
jerryzh168: `tensor_name + suffix` is just `param_name`?

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -269,38 +287,27 @@ def create_quantized_param(
...
+        if "weight_" in param_name:
+            param_name, suffix = param_name.rsplit("_", 1)
+            suffix = "_" + suffix
         module, tensor_name = get_module_from_name(model, param_name)
 
         if self.pre_quantized:
             # If it's a bias, no need to do anything special (except removing the ":_data" part of the key, but was
             # already done) - if it's unsafe-serialized (i.e. not safetensors), not need for anything either
-            is_unsafe_serialization = ":" not in full_name
+            is_unsafe_serialization = "weight_" not in full_name
```
jerryzh168: do we still need to distinguish between safe/non-safe serialization in this new design?

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -253,6 +241,36 @@ def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **
...
+    def update_model_with_metadata(self, model, metadata):
+        state_dict = model.state_dict()
+        if TORCHAO_VERSION >= version.parse("0.14.0") and is_metadata_torchao(self.metadata):
+            updated_state_dict = unflatten_tensor_state_dict(state_dict, metadata)
+
+            for name, param in list(model.named_parameters()):
+                module = model
+                parts = name.split(".")
+                for part in parts[:-1]:
+                    module = getattr(module, part)
```
jerryzh168: can we use https://github.com/pytorch/pytorch/blob/aa4a8c9b9228a8999cac1fba37cd04e086ecfb9b/torch/nn/modules/module.py#L674
```
module_fqn, weight_name = name.rsplit(".", 1)
module = model.get_submodue(module_fqn)
weight = getattr(module, weight_name)
```

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -253,6 +241,36 @@ def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **
                 _QUANTIZABLE.append(torch.nn.Embedding)
             return isinstance(module, tuple(_QUANTIZABLE)) and tensor_name == "weight"
 
+    def update_model_with_metadata(self, model, metadata):
```
jerryzh168: is this called after all shards are loaded to the model?

liangel-02: yes this is called after `load_shard_file()` which loads in all the shards

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -253,6 +241,36 @@ def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **
...
+        state_dict = model.state_dict()
+        if TORCHAO_VERSION >= version.parse("0.14.0") and is_metadata_torchao(self.metadata):
+            updated_state_dict = unflatten_tensor_state_dict(state_dict, metadata)
+
+            for name, param in list(model.named_parameters()):
```
jerryzh168: will we be looping over both `weight_qdata` `weight_scale` twice here? so L268 will be called twice? it feels like it might be better to loop over `updated_state_dict` and then delete all attributes that has the weight_name as prefix and then register attributes based on the `updated_state_dict`, something like

```
# delete first to save memory
# delete old attributes
for weight_name_fqn, param in list(model.named_parameters()):
    module_fqn, weight_name = weight_name_fqn.rsplit(".", 1)
    module = model.get_submodue(module_fqn)
    # if module_fqn + ".weight" in updated_state_dict
    # or to be more general, ".weight" should be derived from `weight_name_fqn` as well, by removing the `_qdata` stuff in the end
    # delete the current attribute
    ...


# add new attributes
for weight_name_fqn, new_param in updated_state_dict.items():
    module_fqn, weight_name = weight_name_fqn.rsplit(".", 1)
    module = model.get_submodue(module_fqn)
    module.register_parameter(weight_name, new_param)
```

liangel-02: if we do two loops we'll need to store param's `device` + `requires_grad`, ill keep track of what params we've updated instead to ensure we don't call twice

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -243,7 +231,7 @@ def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **
         # check if the param_name is not in self.modules_to_not_convert
         if any(key + "." in param_name or key == param_name for key in self.modules_to_not_convert):
             return False
-        elif any(param_name.endswith(f":{x}") for x in self.full_ao_keys):
+        elif "_" in param_name:
```
jerryzh168: isn't this a bit too loose?

jerryzh168 (approved):
thanks, looks good mostly, had one more inline comment

src/transformers/quantizers/quantizer_torchao.py
jerryzh168: is this check still needed? I saw there is a torchao bypass in `_load_state_dict_into_meta_model`

liangel-02: yep, changing to `_weight_`

src/transformers/modeling_utils.py
```diff
@@ -659,7 +664,10 @@ def _load_state_dict_into_meta_model(
             if param_device == "disk":
                 if not is_safetensors:
                     disk_offload_index = offload_weight(param, param_name, disk_offload_folder, disk_offload_index)
-            elif not is_quantized or not hf_quantizer.param_needs_quantization(model, param_name):
+            elif not is_quantized or (
+                not is_metadata_torchao(hf_quantizer.metadata)
+                and not hf_quantizer.param_needs_quantization(model, param_name)
+            ):
```
SunMarc: we shouldn't need to check the torchao metadata if the model is quantized no ?

src/transformers/modeling_utils.py
```diff
@@ -4808,6 +4816,9 @@ def _load_pretrained_model(
                 _error_msgs, disk_offload_index = load_shard_file(args)
                 error_msgs += _error_msgs
 
+        if hf_quantizer:
+            hf_quantizer.update_model_with_metadata(model, hf_quantizer.metadata)
+
```
SunMarc: we have `_process_model_after_weight_loading` to deal with that

src/transformers/quantizers/base.py
```diff
@@ -344,9 +345,9 @@ def get_state_dict_and_metadata(self, model, safe_serialization=False):
         """Get state dict and metadata. Useful when we need to modify a bit the state dict due to quantization"""
         return None, {}
 
-    def update_state_dict_with_metadata(self, state_dict, metadata):
+    def update_model_with_metadata(self, model, metadata):
         """Update state dict with metadata. Default behaviour returns state_dict"""
-        return state_dict
```
SunMarc, line 397: this is actually not used in our codebase as you saw so we can remove it

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -253,6 +241,34 @@ def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **
...
+                    delattr(module, weight_name)
+
+                if name in weights_to_register:
+                    new_param_value = updated_state_dict[name]
+                    new_param = torch.nn.Parameter(new_param_value.to(device), requires_grad=requires_grad)
+                    module.register_parameter(weight_name, new_param)
+
+                    weights_to_register.remove(name)
+
+            model.load_state_dict(updated_state_dict, strict=False)
```
SunMarc: so instead of performing `unflatten_tensor_state_dict` in `create_quantized_param`, we do it here at the very end and we just store the flattened weights in the module?

SunMarc (commented):
Thanks for your work ! Left a couple of comments. Btw, we will soon refactor how quantization is applied as we move to dynamic weights loading like vllm. This should help getting support for features like TP

src/transformers/modeling_utils.py
liangel-02: this is because some parameters are deemed "not to need quantization" by `param_needs_quantization()` but we still need to call `create_quantized_param()` rather than `_load_parameter_into_model()` so it should not go into that code block

src/transformers/quantizers/quantizer_torchao.py
liangel-02: yeah, we don't want to do it in `create_quantized_param` since at most, we'd only have access to one shard file, and we want to consider the case where tensor subclass attributes are split up over multiple files

we call `unflatten_tensor_state_dict` at the very end to get the recovered state dict, and then iterate through the model and replace the weights that represent the tensor attributes with the entire tensor subclass.

src/transformers/modeling_utils.py
SunMarc: maybe we can put it in `param_needs_quantization` ? I perfer not having specific code for torchao in modeling_utils

liangel-02: @SunMarc done if you want to take another look!

liangel-02:
@SunMarc i rebased my pr and now am seeing this error due to [#41580](https://github.com/huggingface/transformers/pull/41580)

```  
File "/home/liangel/local/transformers/src/transformers/modeling_utils.py", line 4122, in from_pretrained
    model, missing_keys, unexpected_keys, mismatched_keys, offload_index, error_msgs = cls._load_pretrained_model(
                                                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/liangel/local/transformers/src/transformers/modeling_utils.py", line 4275, in _load_pretrained_model
    missing_keys, unexpected_keys, mismatched_keys, misc = convert_and_load_state_dict_in_model(
                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/liangel/local/transformers/src/transformers/core_model_loading.py", line 621, in convert_and_load_state_dict_in_model
    raise ValueError("This quantization method is gonna be supported SOOOON")
ValueError: This quantization method is gonna be supported SOOOON
```

will there be follow up changes to support torchao/what changes would be needed? cc @jerryzh168

ArthurZucker:
Happy to help if you want to do the changes here, I think @SunMarc and @MekkCyber are gonna be helping as well making sure torchao is supported!

SunMarc:
We just merged a big [PR](https://github.com/huggingface/transformers/pull/41580) and all quantization methods are impacted, we will add back the support for those methods asap !

liangel-02:
closing due to #41580 #42289

PR #42004, opened 2025-11-04
[kernels] change import time in KernelConfig

MekkCyber:
# What does this PR do?

Importing `kernels` when not installed was causing issues in the CI. Since we have a `kernels` folder `_is_package_availabe` assumes the package `kernels` is available when it's not. This PR simply fixes that by moving the import inside the kernel function that are only executed when we are sure `kernels` is installed


ydshieh (approved):
Thank you. Looks like this is the way to go for now , even if the edge case (very special usage) still show the issue

PR #42008, opened 2025-11-04
Reduce the number of benchmark in the CI

remi-or:
In recent changes, we increased the number of benchmarks being run in the benchmarking CI. This has led to a relatively long CI runtime ~1hr on A10 runners. This might be wasteful. 
This PR changes how the `run_benchmark.py` generates the benchmark configs, favoring a benchmarking "level" system instead of ad hoc heuristic. The system has 5 levels of coverage:

0. Only one cfg is benchmarked, the most efficient until now for generate
1. A few well-performing config are benchmarked. Default option
2. For each attention implementation, we benhcmark one config, and we add 2 more on top to test `kernelize`. This leads to a total of 6 configs, and it is the option used in the CI.
3. We benchmark all possible config, but the compile mode is either `None` (no compile) or `default`
4. Same but with all possible compile modes.

This simplifies the script and makes it usable for both CI and users that want to run a fast benchmark suite as well as those who want more coverage.

### DIFF
--- a/benchmark_v2/framework/benchmark_config.py
+++ b/benchmark_v2/framework/benchmark_config.py
@@ -1,4 +1,5 @@
 import hashlib
+import itertools
 import json
 import logging
 from typing import Any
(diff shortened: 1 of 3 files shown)


benchmark_v2/framework/benchmark_config.py
```diff
@@ -146,60 +147,68 @@ def from_dict(cls, data: dict[str, Any], skip_validity_check: bool = False) -> "
...
+        for x in [
+            warmup_iterations,
+            measurement_iterations,
+            batch_size,
+            sequence_length,
+            num_tokens_to_generate,
+            gpu_monitoring,
+        ]
+    )
+    iterator = itertools.product(*parameters)
```
McPatate, line 170: that's a lot of configs 😄

McPatate (approved):

benchmark_v2/framework/benchmark_config.py
remi-or: level 3 is for no small gpu

PR #42011, opened 2025-11-04
Add TiledFusedLogitsLoss for memory-efficient training

Priyanshjain10:
🚀 What does this PR do?
Adds TiledFusedLogitsLoss — a memory-efficient cross-entropy loss implementation for long sequence language model training (Fixes #41306).

This feature introduces a TiledFusedLogitsLoss class to enable chunk-based (tiled) loss computation as an alternative to standard PyTorch cross-entropy, significantly reducing peak memory usage during training with long sequences and large batch sizes.

Motivation & Context:

Large language models with long sequence inputs often hit memory bottlenecks because logits and loss computation require full materialization of large tensors.

The recently proposed "sequence tiling" technique processes logits and labels in small manageable tiles/chunks, enabling up to 28% reduction in memory usage.

This opens up support for longer sequence training or bigger batch sizes on the same hardware.

Implementation:

New file: src/transformers/loss_utils.py

New class: TiledFusedLogitsLoss

Pure PyTorch, no external dependencies

Simple, drop-in compatible with existing training workflows

Performance:

Peak memory usage: 50 GiB → 36 GiB (example: Llama-2, 4K → 8K sequence length, batch size constant)

Enables longer sequences or larger batches on the same GPU

Negligible impact (<5%) on computation speed

Fixes
Fixes #41306

Documentation
Full docstring in loss_utils.py provided

Example usage added at the top of the file

Dependencies
None (purely backend-optional, configurable)

Tests
 Unit test to verify output and shape correctness

 Memory profiling validated (empirically)

Reviewers and Area
@ArthurZucker (text models, attention, kernel integration)

@stas00 (core memory optimizations)

@Rocketknight1 (pipelines)

@MekkCyber (kernels)

Ready for Review!
This PR is ready for functional, style, and integration review.
Please let me know if you'd like checks with other backends (Liger, DeepSpeed) — they're trivial to add using the same tiling logic.

### DIFF
--- a/src/transformers/loss_utils.py
+++ b/src/transformers/loss_utils.py
@@ -1,0 +1,29 @@
+"""Memory-efficient loss computation."""
+import torch
+import torch.nn.functional as F
+
+class TiledFusedLogitsLoss:
+    def __init__(self, chunk_size=4096, ignore_index=-100):
+        self.chunk_size = chunk_size
+        self.ignore_index = ignore_index
+    
+    def __call__(self, hidden_states, lm_head_weight, labels, lm_head_bias=None):
+        hidden_states = hidden_states.view(-1, hidden_states.shape[-1])
+        labels = labels.view(-1)
+        total_loss = 0.0
+        num_chunks = 0
+        
+        for i in range(0, hidden_states.shape[0], self.chunk_size):
+            end_idx = min(i + self.chunk_size, hidden_states.shape[0])
+            chunk_hidden = hidden_states[i:end_idx]
+            chunk_labels = labels[i:end_idx]
+            
+            if (chunk_labels != self.ignore_index).sum() == 0:
+                continue
+            
+            chunk_logits = F.linear(chunk_hidden, lm_head_weight, lm_head_bias)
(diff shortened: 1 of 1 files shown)


stas00:
@Priyanshjain10, thank you for the PR but why are you proposing the reduced functionality? Was there something wrong with the original class? https://github.com/deepspeedai/DeepSpeed/blob/4691bebdf72a6bdcb56167d741177a9d814c223b/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L960 At the very least you should name it differently if you make it of a limited use.

Moreover I think your re-write misses a crucial performance point - we don't want to do a normal backward, but run the backward inside forward instead.

Plus the discussion in https://github.com/huggingface/transformers/issues/41306 wasn't to just add this class to transformers, and have it languish there, the discussion was to have a toggle switch in `from_pretrained` which will activate this feature for users automatically - and it'll use the more efficient liger-kernel implementation if available and fall back to the pure pytorch implementation otherwise.

This PR has no tests and I can already tell you that the proposed version will not work beyond a single gpu. DDP will not work at all and FSDP and Deepspeed need a special treatment when you shard params (grads reduction will break). Axolotl has a version that works with both Deepspeed and FSDP, I don't remember if @winglian figured out how to make it work with DDP as well.

PR #41626, opened 2025-10-15
[v5] Return a BatchEncoding dict from apply_chat_template by default

Rocketknight1:
Tokenizers return a BatchEncoding dict by default, but `apply_chat_template` doesn't. This is just an accident of how I wrote it originally, which we were stuck with for backward compatibility reasons. Ideally, I think `apply_chat_template` should return exactly the same format as tokenizers, since it also performs tokenization most of the time. It's now `v5` time, so we can start making that happen :sweat_smile:

This PR also updates tests, and removes very old `test_tokenization_for_chat` tests. These model-specific tests don't do anything useful anymore, since the `apply_chat_template` functionality is unified across tokenizers; they're mostly a legacy leftover from when model classes used to need custom chat tokenization functions.

PR #42012, opened 2025-11-04
Fix continuous batching tests

Rocketknight1:
The continuous batching tests call `apply_chat_template` and broke due to the changed default args in https://github.com/huggingface/transformers/pull/41626. This was a real problem because those tests spin up threads via the `Manager` class, which I think persisted as zombie threads as a result (there was a lot of memory usage and very slow cleanup after these failures).

This PR fixes the tests and hopefully also the rest of our generation CI!

### DIFF
--- a/tests/generation/test_continuous_batching.py
+++ b/tests/generation/test_continuous_batching.py
@@ -350,9 +350,9 @@ class ContinuousBatchingTest(unittest.TestCase):
 
         messages = [{"content": "What is the Transformers library known for?", "role": "user"}]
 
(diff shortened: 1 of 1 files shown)


ydshieh (approved):
OK for me, but t looks like continuous batching is not working well with return_dict=True. So would this be a problem for our users ..?

Thank you for the quick action!

Rocketknight1:
@ydshieh The problem is that `add_request` expects just `input_ids`, and breaks when you pass the whole tokenizer dict. It's possible that some user workflows might break because of this, but it should be easy to fix. We knew this change would break backward compatibility a little bit (which is why it waited for v5)!

PR #42013, opened 2025-11-04
add back `logging_dir`

SunMarc:
# What does this PR do?

This PR adds back `logging_dir` from @BenjaminBossan suggestion. It might be more used than I thought due to its generic name even though users might not be using tensorboard.


SunMarc:
@bot /style

BenjaminBossan (approved):
Thanks for taking care of this, LGTM.

PR #42019, opened 2025-11-04
Add backward compatibility helpers to ease v5 RoPE config transition

hmellor:
Overrides the following 3 methods if `rope_parameters` is found to be nested based on `layer_types`:

- `__getitem__` - Checks all subdicts and returns the value as long as it is the same in all subdicts
- `__contains__` - Checks that all subdicts contain the key and return True if they all do
- `__setitem__` - Sets the value for all subdicts

### DIFF
--- a/src/transformers/configuration_utils.py
+++ b/src/transformers/configuration_utils.py
@@ -426,7 +426,7 @@ class PreTrainedConfig(PushToHubMixin):
         if set(self.rope_parameters.keys()).issubset(ALLOWED_LAYER_TYPES):
             def __getitem__(self, key):
                 if key in self.keys():
-                    return super().__getitem__(key)
+                    return dict.__getitem__(self, key)
                 # Try getting it from the nested key
                 values = set()
                 for subdict in self.values():
@@ -439,7 +439,7 @@ class PreTrainedConfig(PushToHubMixin):
 
             def __contains__(self, key):
                 if key in self.keys():
-                    return super().__contains__(key)
+                    return dict.__contains__(self, key)
                 # Try checking in the nested keys
                 values = set()
                 for subdict in self.values():
@@ -452,7 +452,7 @@ class PreTrainedConfig(PushToHubMixin):
 
             def __setitem__(self, key, value):
                 if key in self.keys():
-                    return super().__setitem__(key, value)
+                    return dict.__setitem__(self, key, value)
                 # If key is not present, we need to update all subdicts
                 for subdict in self.values():
                     if isinstance(subdict, dict):
--- a/src/transformers/configuration_utils.py
+++ b/src/transformers/configuration_utils.py
@@ -419,6 +419,49 @@ class PreTrainedConfig(PushToHubMixin):
 
     @property
     def rope_scaling(self):
+        if self.rope_parameters is None:
+            return None
+        
+        # Backward compatibility for v4
+        if set(self.rope_parameters.keys()).issubset(ALLOWED_LAYER_TYPES):
+            def __getitem__(self, key):
+                if key in self.keys():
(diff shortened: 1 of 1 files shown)


src/transformers/configuration_utils.py
```diff
@@ -419,6 +419,49 @@ def torch_dtype(self, value):
...
+                if key in self.keys():
+                    return super().__setitem__(key, value)
+                # If key is not present, we need to update all subdicts
+                for subdict in self.values():
+                    if isinstance(subdict, dict):
+                        subdict[key] = value
+
+            self.rope_parameters.__getitem__ = __getitem__
+            self.rope_parameters.__contains__ = __contains__
+            self.rope_parameters.__setitem__ = __setitem__
```
zucchini-nlp, line 463: not sure if it's a good idea to overwrite `getter/setter` like this in runtime. I wanted to convert `rope_parameters` to a dict-like object in post-init when doing the prev Pr, wdyt about it?

hmellor: vLLM reads from `rope_scaling`/`rope_parameters` in a lot of places (575 references to `rope_scaling` in the codebase currently).

Without this change, we would have to update every place we read `rope_scaling` to check if its keys are a subset of `layer_types` to know if we're dealing with a nested `rope_parameters` or not.

With this change, we can continue to access it like we always have, reducing the friction of updating to v5 and enabling us to support both v4 and v5 style rope configuration in the near term.

zucchini-nlp: Got it, and totally aligned with the idea, Prob nitty-picking, I meant the way it is overriding setters looks weird. We can have `type(config.rope_parameters) = RopeParametersDict` instead of `type(config.rope_parameters) = dict` so we add custom setter/getter in a more clear way

hmellor: I'm not sure I follow. What is the change you're suggesting?

zucchini-nlp:
```python
class RopeParameters(UserDict):
    def __getitem__(self, key):
        # new getter code

    def __setitem__(self, key, value):
        # new setter code

# And in the PreTrainedConfig's post-init add the following. Then rope parameters
# will be a `RopeParameters` with custom getter and setter
self.rope_parameters = RopeParameters(self.rope_parameters)

```

hmellor: Assuming this is different to the `RopeParameters` `TypedDict` in the rope utils, the issue with this is that a new class is created so setting values never make their way back to the original config.

zucchini-nlp: not sure I follow it now. The `RopeParameters` isn't the TypedDict we have, it's more like a small config of its own. Similar to how we save [watermarking config in generation](https://github.com/huggingface/transformers/blob/2418196ef41255cb686125a5f148ac44619c8622/src/transformers/generation/configuration_utils.py#L383-L389) or t o how the config has a text config inside it

hmellor: I was just clarifying that you weren't referring to https://github.com/huggingface/transformers/blob/561233cabf480b0f884700be3aebf42dbe6a7b20/src/transformers/modeling_rope_utils.py#L888

hmellor: If we use a wrapper class, the following would fail because the value would never be written backto the actual config

```py
config.rope_scaling["something"] = 1
assert config.rope_scaling["something"] == 1
```

zucchini-nlp:
Not forgotten, discussed internally if updates will be done in-place or by creating a copy. Reviewing after checking it

hmellor:
Closing as this does not solve the issue for models which actually use the nested `RoPEParameters`

PR #42028, opened 2025-11-05
Add support for MiniMax-M2

rogeryoungh:
# What does this PR do?

This PR adds MiniMax-M2 model to Hugging Face Transformers from MiniMaxAI.

Relevant Links:

- [MiniMax M2 & Agent: Ingenious in Simplicity](https://www.minimax.io/news/minimax-m2)
- https://huggingface.co/MiniMaxAI/MiniMax-M2

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker @Cyrilvallez


docs/source/en/model_doc/minimax_m2.md
```diff
@@ -0,0 +1,82 @@
...
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be rendered properly in your Markdown viewer.
+
+-->
+
+
+# MiniMax-M2
+
+## Overview
+
+MiniMax-M2 redefines efficiency for agents. It's a compact, fast, and cost-effective MoE model (230 billion total parameters with 10 billion active parameters) built for elite performance in coding and agentic tasks, all while maintaining powerful general intelligence. With just 10 billion activated parameters, MiniMax-M2 provides the sophisticated, end-to-end tool use performance expected from today's leading models, but in a streamlined form factor that makes deployment and scaling easier than ever.
```
molbap: Changed a bit to be more factual 
```suggestion
MiniMax-M2 is a compact, fast, and cost-effective MoE model (230 billion total parameters with 10 billion active parameters) built for elite performance in coding and agentic tasks, all while maintaining powerful general intelligence. With just 10 billion activated parameters, MiniMax-M2 provides the sophisticated, end-to-end tool use performance expected from today's leading models, but in a streamlined form factor that makes deployment and scaling easier than ever.
```

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,384 @@
...
+        "layers.*.block_sparse_moe.experts.*.w3": "colwise",
+    }
+    base_model_pp_plan = {
+        "embed_tokens": (["input_ids"], ["inputs_embeds"]),
+        "layers": (["hidden_states", "attention_mask"], ["hidden_states"]),
+        "norm": (["hidden_states"], ["hidden_states"]),
+    }
+    attribute_map = {
+        "num_experts": "num_local_experts",
+    }
```
molbap, line 150: For further reviews: This part is identical to `Minimax1` configuration but inheriting the configuration would require deleting a bunch of keys, like `full_attn_beta_factor`, and so on, so ok to keep.

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,384 @@
...
+            pad_token_id=pad_token_id,
+            bos_token_id=bos_token_id,
+            eos_token_id=eos_token_id,
+            tie_word_embeddings=tie_word_embeddings,
+            **kwargs,
+        )
+
+
+class MiniMaxM2MLP(MixtralMLP):
+    pass
```
molbap: should be safe to remove
```suggestion
```

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,384 @@
...
+        )
+
+        attn_output = attn_output.reshape(*input_shape, -1).contiguous()
+        attn_output = self.o_proj(attn_output)
+        return attn_output, attn_weights
+
+
+class MiniMaxM2DecoderLayer(MixtralDecoderLayer):
+    pass
+
```
molbap: safe to delete as well
```suggestion
```

molbap (approved):
Very clean integration, no particular comments. Thank you! cc @Cyrilvallez for core review

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,374 @@
...
+        self.num_key_value_groups = config.num_attention_heads // config.num_key_value_heads
+        self.scaling = self.head_dim**-0.5
+        self.attention_dropout = config.attention_dropout
+        self.is_causal = True
+        self.q_proj = nn.Linear(config.hidden_size, config.num_attention_heads * self.head_dim, bias=False)
+        self.k_proj = nn.Linear(config.hidden_size, config.num_key_value_heads * self.head_dim, bias=False)
+        self.v_proj = nn.Linear(config.hidden_size, config.num_key_value_heads * self.head_dim, bias=False)
+        self.o_proj = nn.Linear(config.num_attention_heads * self.head_dim, config.hidden_size, bias=False)
+
+        self.use_qk_norm = config.use_qk_norm
```
molbap: missed that (I think) - do we always use qk norm? if yes, no need for an `if` check

rogeryoungh: You're right, in our current configuration, `use_qk_norm` is always true. However, I kept the if statement to maintain consistency with other model implementations that use `qk_norm`.

molbap: Since one model == one file, it's better to remove unused code paths, so I'd suggest to remove that one!

rogeryoungh: `use_qk_norm` has been removed.

molbap:
Failing tests on the hub are unrelated timeouts. Thanks for the update!

Qubitium:
@ArthurZucker lets gets this quickly merged. 2 weeks is infinity in the current model wars cycle.

src/transformers/models/auto/modeling_auto.py
vasqu, line 1: We should also add to `tokenization_auto`. If I see it correctly on the hub, then it's a gpt2 tokenizer

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,368 @@
...
+        self.num_local_experts = num_local_experts
+        self.output_router_logits = output_router_logits
+        self.router_aux_loss_coef = router_aux_loss_coef
+        self.router_jitter_noise = router_jitter_noise
+        # Try to set `rope_scaling` if available, otherwise use `rope_parameters`
+        rope_scaling = kwargs.pop("rope_scaling", None)
+        self.rope_parameters = rope_scaling or rope_parameters
+
+        # Validate the correctness of rotary position embeddings parameters
+        rope_theta = kwargs.get("rope_theta", 1000000.0)
```
vasqu: This was only meant for BC, let's remove this - instead it should be directly in the config already

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,368 @@
...
+        "layers.*.block_sparse_moe.experts.*.w3": "colwise",
+    }
+    base_model_pp_plan = {
+        "embed_tokens": (["input_ids"], ["inputs_embeds"]),
+        "layers": (["hidden_states", "attention_mask"], ["hidden_states"]),
+        "norm": (["hidden_states"], ["hidden_states"]),
+    }
+    attribute_map = {
+        "num_experts": "num_local_experts",
+    }
```
vasqu, line 150: Nit: would be nice if we could change this directly but given it's already released it's ok to keep if not

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,368 @@
...
+        # Validate the correctness of rotary position embeddings parameters
+        rope_theta = kwargs.get("rope_theta", 1000000.0)
+        standardize_rope_params(self, rope_theta=rope_theta)
+        rope_config_validation(self)
+
+        rotary_dim = kwargs.pop("rotary_dim", head_dim)
+        self.partial_rotary_factor = kwargs.pop("partial_rotary_factor", 1.0)
+
+        if self.head_dim is not None:
+            self.partial_rotary_factor = rotary_dim / self.head_dim
```
vasqu: Any reason this needs to be within the config itself? It also seems that this is not used at all (i.e. defaults to 1). Let's remove this and if it is indeed used, then it needs to be in the `rope_parameters` directly

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,368 @@
...
+        self.experts = MiniMaxM2Experts(config)
+        self.register_buffer("e_score_correction_bias", torch.zeros(config.num_local_experts))
+
+    def route_tokens_to_experts(self, router_logits):
+        routing_weights = torch.nn.functional.sigmoid(router_logits.float())
+        scores_for_choice = routing_weights + self.e_score_correction_bias
+        _, top_k_index = torch.topk(scores_for_choice, self.top_k, dim=-1, sorted=False)
+        top_k_weights = routing_weights.gather(1, top_k_index)
+        top_k_weights /= top_k_weights.sum(dim=-1, keepdim=True)
+        return top_k_index, top_k_weights.to(router_logits.dtype)
```
vasqu: We recently changed how we build MoE, can you add this to our weight converter and adjust respectively; see https://github.com/huggingface/transformers/blob/5804c1fa474f2568e359e3fb5800cb7b13fb01e2/src/transformers/conversion_mapping.py#L109

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,368 @@
...
+class MiniMaxM2Attention(Glm4MoeAttention):
+    def __init__(self, config: MiniMaxM2Config, layer_idx: int):
+        nn.Module.__init__(self)
+        self.config = config
+        self.layer_idx = layer_idx
+        self.head_dim = getattr(config, "head_dim", None) or config.hidden_size // config.num_attention_heads
+        self.num_key_value_groups = config.num_attention_heads // config.num_key_value_heads
+        self.scaling = self.head_dim**-0.5
+        self.attention_dropout = config.attention_dropout
+        self.is_causal = True
```
vasqu: We can use qwen3 instead imo, this will allow us to use super init and remove / overwrite the parts that changed

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,368 @@
...
+        input_shape = hidden_states.shape[:-1]
+        hidden_shape = (*input_shape, -1, self.head_dim)
+
+        query_states = self.q_proj(hidden_states)
+        key_states = self.k_proj(hidden_states)
+        value_states = self.v_proj(hidden_states)
+
+        # main diff from Llama
+        query_states = self.q_norm(query_states)
+        key_states = self.k_norm(key_states)
```
vasqu: That's interesting :D nothing for you to do, seeing this the first time

tests/models/minimax_m2/test_modeling_minimax_m2.py
```diff
@@ -0,0 +1,202 @@
...
+        processor_name,
+    ):
+        return True
+
+    @require_flash_attn
+    @require_torch_gpu
+    @pytest.mark.flash_attn_test
+    @slow
+    def test_flash_attn_2_inference_equivalence_right_padding(self):
+        self.skipTest(reason="MiniMaxM2 flash attention does not support right padding")
```
vasqu: These feel like copy paste, can you check if we can remove either?

tests/models/minimax_m2/test_modeling_minimax_m2.py
```diff
@@ -0,0 +1,202 @@
...
+        # This is to mimic torch.testing.assert_not_close
+        self.assertNotAlmostEqual(include_padding_result.aux_loss.item(), result.aux_loss.item())
+
+
+@require_torch
+class MiniMaxM2IntegrationTest(unittest.TestCase):
+    @slow
+    @require_torch_accelerator
+    def test_small_model_logits(self):
+        model_id = "hf-internal-testing/MiniMaxM2-tiny"
```
vasqu: @molbap have you uploaded a model there? That's definitely the way to go but we should double-check and probably adjust numbers

src/transformers/models/minimax_m2/modular_minimax_m2.py
vasqu: This possibly also affects the tp plan

vasqu (commented):
Added some comments but looks already very nice, good job! Some things have changed on main so you will likely need to add our new weights converter to this model as well.

Imo, the integration tests seem suspicious to me and might not have been checked?

tests/models/minimax_m2/test_modeling_minimax_m2.py
molbap: haven't! good catch

vasqu:
@Qubitium sorry for the delays, will be taking over for the most part. Crunching some important features for v5 so some PRs got lost under it, apologies

src/transformers/models/minimax_m2/modular_minimax_m2.py
Qubitium: @vasqu @rogeryoungh I am not minimax staff but still have concerns myself.

Since HF Transformer changes broke this PR in terms of compat, what is exaclty is 

`def _build_checkpoint_conversion_mapping():` doing? What is `WeightConverter`?

I see no method level documentation and only internal code comments which looks like list generation to assist in concat/fusing of MoE modules? But how everything is glued together is still unclear. 

Minimax internally does not use HF transformers stack and I think they would have the same question as I do. What exactly is the latest v5 changes regarding MoE? Is there an overview doc to make this clear? 

@ArthurZucker Can we get a global (even in alpha form) doc/readme on the MoE arch changes in v5 so model makers has a clear eagle eye view of the v5 target when it comes to MoE?

rogeryoungh:
I've encountered a strange problem:

```python
>>> model = AutoModelForCausalLM.from_pretrained(MODEL_PATH, device_map="auto")
  File "<stdin>", line 1, in <module>
  File "../transformers/src/transformers/models/auto/auto_factory.py", line 373, in from_pretrained
    return model_class.from_pretrained(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/modeling_utils.py", line 278, in _wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/modeling_utils.py", line 3967, in from_pretrained
    device_map = _get_device_map(model, device_map, max_memory, hf_quantizer)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/integrations/accelerate.py", line 407, in _get_device_map
    device_map = infer_auto_device_map(
                 ^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/integrations/accelerate.py", line 736, in infer_auto_device_map
    current_buffer_size = compute_module_total_buffer_size(module, hf_quantizer)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/integrations/accelerate.py", line 260, in compute_module_total_buffer_size
    module_sizes, _ = compute_module_sizes(model, hf_quantizer, buffers_only=True)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/integrations/accelerate.py", line 240, in compute_module_sizes
    dtype_size = hf_quantizer.param_element_size(model, name)
  File "../transformers/src/transformers/quantizers/base.py", line 182, in param_element_size
    return model.get_parameter_or_buffer(param_name).element_size()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1964, in __getattr__
    raise AttributeError(
AttributeError: 'MiniMaxM2DecoderLayer' object has no attribute 'get_parameter_or_buffer'
```

I found a solution #42342, which involves adding a `get_parameter_or_buffer` function to the MiniMaxM2RotaryEmbedding. However, for MiniMax-M2, `e_score_correction_bias` also needs to be changed to `nn.Parameter`.

```diff
@@ -106,7 +106,10 @@ class MiniMaxM2SparseMoeBlock(nn.Module):
         self.jitter_noise = config.router_jitter_noise
         self.gate = nn.Linear(config.hidden_size, config.num_local_experts, bias=False)
         self.experts = MiniMaxM2Experts(config)
-        self.register_buffer("e_score_correction_bias", torch.zeros(config.num_local_experts))
+        self.e_score_correction_bias = nn.Parameter(
+            torch.zeros(config.num_local_experts),
+            requires_grad=False,
+        )
 
     def forward(self, hidden_states: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
         batch_size, sequence_length, hidden_dim = hidden_states.shape
@@ -168,6 +171,20 @@ class MiniMaxM2RotaryEmbedding(nn.Module):
         self.register_buffer("inv_freq", inv_freq, persistent=False)
         self.original_inv_freq = inv_freq
 
+    # Add a compatibility method so callers expecting PreTrainedModel-like API don't crash.
+    def get_parameter_or_buffer(self, name: str):
+        # Prefer direct attribute access (parameters and buffers are attributes)
+        if hasattr(self, name):
+            return getattr(self, name)
+        # Fallback: search named parameters and buffers (non-recursive to keep semantics)
+        for n, p in self.named_parameters(recurse=False):
+            if n == name:
+                return p
+        for n, b in self.named_buffers(recurse=False):
+            if n == name:
+                return b
+        raise AttributeError(f"{self.__class__.__name__} has no parameter or buffer named '{name}'")
+
     @staticmethod
     def compute_default_rope_parameters(
         config: Optional[MiniMaxM2Config] = None,
```

MekkCyber:
Hi @rogeryoungh, we are taking care of this in : https://github.com/huggingface/transformers/pull/42289/files#diff-7f40070336f6d7b1ffe08e654cdf930080e3cbd4dbbcbee2996fabe4ffc1c2b3, this should fix the issue without adding a `get_parameter_or_buffer`

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,478 @@
...
+        # Validate the correctness of rotary position embeddings parameters
+        standardize_rope_params(self, rope_theta=rope_theta)
+        rope_config_validation(self)
+
+        rotary_dim = kwargs.pop("rotary_dim", head_dim)
+
+        if self.head_dim is not None:
+            self.partial_rotary_factor = rotary_dim / self.head_dim
+        else:
+            self.partial_rotary_factor = 1.0
```
vasqu: We really either
- have this within the rope dict directly
- or remove this (is it used in the main model?)

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,478 @@
...
+        self.w1 = nn.Linear(self.hidden_dim, self.ffn_dim, bias=False)
+        self.w2 = nn.Linear(self.ffn_dim, self.hidden_dim, bias=False)
+        self.w3 = nn.Linear(self.hidden_dim, self.ffn_dim, bias=False)
+
+        self.act_fn = ACT2FN[config.hidden_act]
+
+    def forward(self, hidden_states):
+        current_hidden_states = self.act_fn(self.w1(hidden_states)) * self.w3(hidden_states)
+        current_hidden_states = self.w2(current_hidden_states)
+        return current_hidden_states
```
vasqu: This looks very similar to `LlamaMLP`, let's inherit from this instead. Regarding the weights name you can rename the weights if necessary. I will add a proper explanation on top of my comments.

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,478 @@
...
+    def forward(self, hidden_states: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+        batch_size, sequence_length, hidden_dim = hidden_states.shape
+        if self.training and self.jitter_noise > 0:
+            hidden_states *= torch.empty_like(hidden_states).uniform_(1.0 - self.jitter_noise, 1.0 + self.jitter_noise)
+        hidden_states = hidden_states.view(-1, hidden_states.shape[-1])
+        router_logits = self.gate(hidden_states)
+        top_k_index, top_k_weights = self.route_tokens_to_experts(router_logits)
+        hidden_states = self.experts(hidden_states, top_k_index, top_k_weights.to(hidden_states.dtype))
+        hidden_states = hidden_states.reshape(batch_size, sequence_length, hidden_dim)
+        return hidden_states
```
vasqu: This was originally very similar to Mixtral, hence you can use the same weight conversions and TP/EP plans as the current Mixtral. 

We support on the fly weight conversions and renaming (similar to vllm), which allows us to standardize a lot of things even more and allowing more efficient kernel implementations. This is why I linked https://github.com/huggingface/transformers/blob/96d1c5d63d378fc64908d96fcf365cb96bfb9f83/src/transformers/conversion_mapping.py#L104
As you can also copy this from Mixtral, expecting you still follow closely to the Mixtral model series. You define per model type which operations are done --> In this case, renaming the MoE Mlp along fusing weights to single parameters

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,478 @@
...
+            scaling=self.scaling,
+            **kwargs,
+        )
+
+        attn_output = attn_output.reshape(*input_shape, -1).contiguous()
+        attn_output = self.o_proj(attn_output)
+        return attn_output, attn_weights
+
+
+class MiniMaxM2DecoderLayer(MixtralDecoderLayer):
```
vasqu: Is it because of the MoE mlp name? Please see my other comment(s) about renaming/converting

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,478 @@
...
+        hidden_states = residual + hidden_states
+        residual = hidden_states
+        hidden_states = self.post_attention_layernorm(hidden_states)
+        hidden_states = self.block_sparse_moe(hidden_states)
+        hidden_states = residual + hidden_states
+        return hidden_states
+
+
+@auto_docstring
+class MiniMaxM2PreTrainedModel(PreTrainedModel):
```
vasqu: Should be also (partially) inheritable, no?

tests/models/minimax_m2/test_modeling_minimax_m2.py
```diff
@@ -0,0 +1,178 @@
...
+        # We make sure that the loss of including padding tokens != the loss without padding tokens
+        # if attention_mask=None --> we don't exclude padding tokens
+        include_padding_result = model(padded_input_ids, attention_mask=None)
+
+        # This is to mimic torch.testing.assert_not_close
+        self.assertNotAlmostEqual(include_padding_result.aux_loss.item(), result.aux_loss.item())
+
+
+@require_torch
+class MiniMaxM2IntegrationTest(unittest.TestCase):
```
vasqu, line 93: Can you provide us a slimmed down model version, i.e. ideally running on A10 GPUs (24GB)? So that we can have some integration tests? I can move this model/copy it to our internal repos

tests/models/minimax_m2/test_modeling_minimax_m2.py
```diff
@@ -0,0 +1,178 @@
...
+
+    if is_torch_available():
+        base_model_class = MiniMaxM2Model
+
+
+@require_torch
+class MiniMaxM2ModelTest(CausalLMModelTest, unittest.TestCase):
+    model_tester_class = MiniMaxM2ModelTester
+
+    # Ignore copy
```
vasqu:
```suggestion
```

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,478 @@
...
+
+
+class MiniMaxM2SparseMoeBlock(MixtralSparseMoeBlock):
+    def __init__(self, config):
+        nn.Module.__init__(self)
+        self.top_k = config.num_experts_per_tok
+        self.jitter_noise = config.router_jitter_noise
+        self.gate = nn.Linear(config.hidden_size, config.num_local_experts, bias=False)
+        self.experts = MiniMaxM2Experts(config)
+        self.register_buffer("e_score_correction_bias", torch.zeros(config.num_local_experts))
```
vasqu, line 252: Is this not saved as parameter? Are we always using a 0s buffer, i.e. does it even do anything?

vasqu (commented):
Added some comments which are mostly revolving around the weights name/convention. See #41580 for the respective PR that introduced this

The essence is that we now support on the fly renaming and weight conversion (similar to vllm). You only have to specify your operations in https://github.com/huggingface/transformers/blob/96d1c5d63d378fc64908d96fcf365cb96bfb9f83/src/transformers/conversion_mapping.py#L26

In your case, this will look very similar to Mixtral I assume, i.e. `block_sparse->mlp` + fusing the MoE modulelists to parameters. Maybe additional mlp renaming for the base MLP to inherit from LlamaMLP. Please let us know if it doesn't work as expected, it's still WIP in a way and we work on providing better docs for this (cc @ArthurZucker).

I know this is a lot at once so apologies.

src/transformers/models/minimax_m2/modular_minimax_m2.py
vasqu: Tried to answer here a bit https://github.com/huggingface/transformers/pull/42028#pullrequestreview-3501706669

- `_build_checkpoint_conversion_mapping` is providing a mapping for `model_type -> weight renaming+conversion`
- These operations can be split into 2 types
  - Renaming, should be self-explanatory --> rename weights 
  - WeightConverter --> change the original weights in some way (spread out in [core_model_loading](https://github.com/huggingface/transformers/blob/main/src/transformers/core_model_loading.py)
    - Chunk --> split the weights along a dimension into x chunks, e.g. useful for qkv splitting into q, k, and v
    - Concatenate --> essentially the reverse of chunk, e.g. fusing q, k, v into qkv
    - MergeModulelist --> merge a module list of tensors into one big tensor along a new dimension, e.g. when a module list should be one big parameter instead of a list (useful for us to merge MoE parameters and add hf kernels on top for more efficient inference etc)
    - PermuteForRope --> more complicated but essentially related to how weights are rotated, e.g. original llama needs this for hf RoPE conversion

You can stack these operations as well but we continue to change/add these.

The standard for MoE currently is to have fused experts into one big parameter instead of a modulelist. We use these to enable kernels, e.g. https://github.com/huggingface/transformers/blob/96d1c5d63d378fc64908d96fcf365cb96bfb9f83/src/transformers/models/llama4/modeling_llama4.py#L150

This is for efficiency reasons in essence.

src/transformers/models/minimax_m2/modular_minimax_m2.py
rogeryoungh: The `partial_rotary_factor` in the main model is 0.5; setting it to 1.0 will pass the test.

src/transformers/models/minimax_m2/modular_minimax_m2.py
rogeryoungh: `e_score_correction_bias` will be added to `routing_weights`, where `routing_weights = router_logits.sigmoid()`. I defined this parameter using `register_buffer` to keep it consistent with other models.

rogeryoungh:
I verified that PR #42289 resolved the previous issue.

Now I've encountered a new problem: when loading weights, I get the following error. Can this be resolved using `WeightRenaming`?

```bash
Key                                                       | Status     | Details
----------------------------------------------------------+------------+--------
model.layers.{0...61}.mlp.experts.down_proj_scale_inv     | UNEXPECTED |        
model.layers.{0...61}.mlp.experts.gate_up_proj_scale_inv  | UNEXPECTED |        
model.layers.{0...61}.mlp.experts.gate_up_proj_scales_inv | MISSING    |        
model.layers.{0...61}.mlp.experts.down_proj_scales_inv    | MISSING    |        

Notes:
- MISSING       :those params were newly initialized because missing form the checkpoint. Consider training on your downstream task.
```

vasqu:
I'm a bit surprised that you get 
```text
model.layers.{0...61}.mlp.experts.gate_up_proj_scales_inv | MISSING    |        
model.layers.{0...61}.mlp.experts.down_proj_scales_inv    | MISSING    |     
```
Don't you have https://github.com/huggingface/transformers/blob/8fe4d2bd67dc8f0c624e2234ce0d07768bc54c1a/src/transformers/models/minimax_m2/modeling_minimax_m2.py#L58-L59

This seems weird to me, how do you load the model? But yes, that can work as well but I'm surprised about the naming because it isn't in the modeling file at all - can you double check?

src/transformers/models/minimax_m2/modular_minimax_m2.py
vasqu: Gotcha

src/transformers/models/minimax_m2/modular_minimax_m2.py
vasqu: Which test? Can we move this directly to the rope config instead? Everything rope should be directly in the config, no more within the kwargs

rogeryoungh:
image

This was a part of FP8Linear.

https://github.com/huggingface/transformers/blob/8fe4d2bd67dc8f0c624e2234ce0d07768bc54c1a/src/transformers/integrations/finegrained_fp8.py#L328-L335

molbap:
Could it be linked with what we discussed yesterday @MekkCyber , just in case?

vasqu:
#42434 should fix your issue @rogeryoungh 

Sorry didn't notice we have native fp8 in this model :facepalm:

MekkCyber:
Hi @rogeryoungh ! Can you try with the pr @vasqu shared please 🤗 ! Thank you

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,378 @@
...
+    >>> configuration = model.config
+    ```"""
+
+    model_type = "minimax_m2"
+    keys_to_ignore_at_inference = ["past_key_values"]
+    base_model_tp_plan = {
+        "layers.*.self_attn.q_proj": "colwise",
+        "layers.*.self_attn.k_proj": "colwise",
+        "layers.*.self_attn.v_proj": "colwise",
+        "layers.*.self_attn.o_proj": "rowwise",
```
vasqu:
```suggestion
        "layers.*.self_attn.q_proj": "colwise_rep",
        "layers.*.self_attn.k_proj": "colwise_rep",
        "layers.*.self_attn.v_proj": "colwise_rep",
        "layers.*.self_attn.o_proj": "rowwise_rep",
```
It's less efficient but we likely need to replicate here as the qk norm goes over all heads x head dim. Previous models only had head dim so it didn't change under TP

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,378 @@
...
+        eos_token_id: Optional[int] = 2,
+        tie_word_embeddings: Optional[bool] = False,
+        sliding_window: Optional[int] = None,
+        attention_dropout: Optional[float] = 0.0,
+        num_experts_per_tok: Optional[int] = 2,
+        num_local_experts: Optional[int] = 8,
+        output_router_logits: Optional[bool] = False,
+        router_aux_loss_coef: Optional[float] = 0.001,
+        router_jitter_noise: Optional[float] = 0.0,
+        rope_theta: Optional[float] = 1000000.0,
```
vasqu: Should not be a kwarg anymore, if you need BC behavior I would suggest to do the same as in llama https://github.com/huggingface/transformers/blob/7c8d72bde8fe7c362aa3294f4244a8dd67af0abf/src/transformers/models/llama/configuration_llama.py#L174-L181

It would be nicer if we could just put it directly into the `rope_parameters` dict

src/transformers/models/minimax_m2/modular_minimax_m2.py
vasqu: Ok adjusting my comments, sorry about the confusion. 

I thought we had an attribute for this in `rope_parameters` but apparently not yet, have to wait for #42255 (which should be merged today). In general, you can follow `gpt_neox` for this and iiuc, you need a `partial_rotary_factor` of 0.5. Just mind that the attribute ideally should be within the `rope_parameters` dict then.

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+class MiniMaxM2TopKRouter(nn.Module):
+    def __init__(self, config):
+        super().__init__()
+        self.top_k = config.num_experts_per_tok
+        self.num_experts = config.num_local_experts
+        self.hidden_dim = config.hidden_size
+        self.weight = nn.Parameter(torch.empty(self.num_experts, self.hidden_dim))
+        self.register_buffer("e_score_correction_bias", torch.zeros(self.num_experts))
```
vasqu: We can still inherit from mixtral for the init here, only needs the buffer for the bias then declared (super init handles the rest)

tests/models/minimax_m2/test_modeling_minimax_m2.py
vasqu: ^

vasqu (commented):
My last few important comments, looks very good already! Sorry about all the issues during the PR with fp8 and the buffer :sweat:

tests/models/minimax_m2/test_modeling_minimax_m2.py
rogeryoungh: I wrote a script to generate small models, but I encountered a problem: `AutoModelForCausalLM.from_config` does not follow the `quantization_config` settings, so the generated models do not have FP8Linear layers.
https://github.com/rogeryoungh/MiniMaxM2TinyModelGenerator/blob/main/generator.py

rogeryoungh:
When syncing branches, I encountered some issues related to FP8Linear:

```bash
Key                                                      | Status     |  | 
---------------------------------------------------------+------------+--+-
model.layers.{0...61}.mlp.experts.gate_up_proj_scale_inv | UNEXPECTED |  | 
model.layers.{0...61}.mlp.experts.down_proj_scale_inv    | UNEXPECTED |  | 

Notes:
Some parameters are on the meta device because they were offloaded to the cpu.
```

I also encountered a strange error, it seems that FP8Linear is not working correctly here?

```bash
  File "<stdin>", line 1, in <module>
  File "../transformers/.venv/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/generation/utils.py", line 2684, in generate
    result = decoding_method(
             ^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/generation/utils.py", line 2877, in _sample
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/generation/utils.py", line 3853, in _prefill
    return self(**model_inputs, return_dict=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
  File "../transformers/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1786, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/.venv/lib/python3.12/site-packages/accelerate/hooks.py", line 175, in new_forward
    output = module._old_forward(*args, **kwargs)
  File "../transformers/src/transformers/utils/generic.py", line 764, in wrapper
    output = func(self, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/models/minimax_m2/modeling_minimax_m2.py", line 671, in forward
    outputs: MoeModelOutputWithPast = self.model(
                                      ^^^^^^^^^^^
  File "../transformers/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
  File "../transformers/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1786, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/utils/generic.py", line 919, in wrapper
    outputs = func(self, *args, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/models/minimax_m2/modeling_minimax_m2.py", line 508, in forward
    hidden_states = decoder_layer(
                    ^^^^^^^^^^^^^^
  File "../transformers/src/transformers/modeling_layers.py", line 94, in __call__
    return super().__call__(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
  File "../transformers/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1786, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/.venv/lib/python3.12/site-packages/accelerate/hooks.py", line 175, in new_forward
    output = module._old_forward(*args, **kwargs)
  File "../transformers/src/transformers/models/minimax_m2/modeling_minimax_m2.py", line 412, in forward
    hidden_states = self.mlp(hidden_states)
                    ^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
  File "../transformers/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1786, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/.venv/lib/python3.12/site-packages/accelerate/hooks.py", line 175, in new_forward
    output = module._old_forward(*args, **kwargs)
  File "../transformers/src/transformers/models/minimax_m2/modeling_minimax_m2.py", line 124, in forward
    hidden_states = self.experts(hidden_states, top_k_index, top_k_weights)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
  File "../transformers/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1786, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/.venv/lib/python3.12/site-packages/accelerate/hooks.py", line 175, in new_forward
    output = module._old_forward(*args, **kwargs)
  File "../transformers/src/transformers/models/minimax_m2/modeling_minimax_m2.py", line 80, in forward
    gate, up = nn.functional.linear(current_state, self.gate_up_proj[expert_idx]).chunk(2, dim=-1)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: expected mat1 and mat2 to have the same dtype, but got: float != c10::Float8_e4m3fn
```

MekkCyber:
Sorry about that @rogeryoungh ! can you check with this pr: https://github.com/huggingface/transformers/pull/42654, it's a small fix

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,374 @@
...
+        eos_token_id: Optional[int] = 2,
+        tie_word_embeddings: Optional[bool] = False,
+        sliding_window: Optional[int] = None,
+        attention_dropout: Optional[float] = 0.0,
+        num_experts_per_tok: Optional[int] = 2,
+        num_local_experts: Optional[int] = 8,
+        output_router_logits: Optional[bool] = False,
+        router_aux_loss_coef: Optional[float] = 0.001,
+        router_jitter_noise: Optional[float] = 0.0,
+        rope_theta: Optional[float] = 1000000.0,
```
vasqu: We changed it again a bit, let's move this to an attribute like here https://github.com/huggingface/transformers/blob/4ad279fb981ea341a09c04a5239e862fb33262cb/src/transformers/models/minimax/modular_minimax.py#L160
I.e. `default_theta` and remove it from the docs as well

tests/models/minimax_m2/test_modeling_minimax_m2.py
```diff
@@ -0,0 +1,177 @@
...
+
+    if is_torch_available():
+        base_model_class = MiniMaxM2Model
+
+
+@require_torch
+class MiniMaxM2ModelTest(CausalLMModelTest, unittest.TestCase):
+    model_tester_class = MiniMaxM2ModelTester
+
+    def test_load_balancing_loss(self):
```
vasqu, line 51: Can you align this test with Jamba https://github.com/huggingface/transformers/blob/4ad279fb981ea341a09c04a5239e862fb33262cb/tests/models/jamba/test_modeling_jamba.py#L406-L446
(you can remove the comments with `After #xxx`)

The test as is is a bit more susceptible to being flaky otherwise + uses more resources than necessary

tests/models/minimax_m2/test_modeling_minimax_m2.py
vasqu: This would initialize random weights, no? I think we can just cut the original safetensors and load only a partial subset of the original instead (e.g. 8 layers), wdyt? This should keep the fp8 weights

Otherwise, even if it is saved in bf16 for example, we should be able to quantize to fp8 on the fly. Not ideal but the goal here is to have a working version.

vasqu: Native fp8 weights for models initialized from config cc @MekkCyber @SunMarc

vasqu (commented):
Thanks left a few comments, assuming that the other PR fixed the fp8 issue 

It looks very ready, let's just focus on adding integration tests then we're good to go imo

rogeryoungh:
I used the following code to load the model from the configuration file:

```python
# ... modify default config and save
config = AutoConfig.from_pretrained(
    save_folder,
    trust_remote_code=True,
)
model = AutoModelForCausalLM.from_config(config)

# full code: https://github.com/rogeryoungh/MiniMaxM2TinyModelGenerator/blob/main/generator.py
```

But I received the following error:

```bash
  File "../MiniMaxM2TinyModelGenerator/generator.py", line 49, in <module>
    model = AutoModelForCausalLM.from_config(config)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/models/auto/auto_factory.py", line 237, in from_config
    return model_class._from_config(config, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/modeling_utils.py", line 250, in _wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/modeling_utils.py", line 1438, in _from_config
    model = cls(config, **kwargs)
            ^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/models/minimax_m2/modeling_minimax_m2.py", line 617, in __init__
    self.model = MiniMaxM2Model(config)
                 ^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/models/minimax_m2/modeling_minimax_m2.py", line 458, in __init__
    self.rotary_emb = MiniMaxM2RotaryEmbedding(config=config)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/models/minimax_m2/modeling_minimax_m2.py", line 160, in __init__
    self.rope_type = self.config.rope_parameters["rope_type"]
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "../transformers/src/transformers/configuration_utils.py", line 198, in __getattribute__
    return super().__getattribute__(key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'MiniMaxM2Config' object has no attribute 'rope_parameters'
```

src/transformers/models/minimax_m2/configuration_minimax_m2.py
```diff
@@ -0,0 +1,205 @@
...
+    }
+    base_model_pp_plan = {
+        "embed_tokens": (["input_ids"], ["inputs_embeds"]),
+        "layers": (["hidden_states", "attention_mask"], ["hidden_states"]),
+        "norm": (["hidden_states"], ["hidden_states"]),
+    }
+    attribute_map = {
+        "num_experts": "num_local_experts",
+    }
+    default_theta = 1000000.0
```
vasqu:
```suggestion
    default_theta = 5000000.0
```
Just looking at the values, can you double check? We want to construct the original v2 model with these values

vasqu:
@rogeryoungh I think the auto mapping is getting confused here, you should rather directly use `MiniMaxM2Config` from your current PR. At least that seems to work somewhat. Note that this was used for BC loading but not saving so you might need to check if some values are unwanted, e.g. `rope_scaling` will also live beside `rope_parameters`, `layernorm_full_attention_beta`, etc. --> We should have a config that is only having the relevant fields (if that's ok) for us.

Also the automap should not consist in the config unless you want to use remote code and need to point the respective files (which shouldn't be the case anymore after merging this).

Last side note: I think the way you load the generation config it will load all our values (including those that don't divert from defaults). Imo, you can keep the generation config purely as is and just move the raw json.

Edit: If it helps you can also do something along
```python
from transformers import MiniMaxM2Config, MiniMaxM2Config

config = MiniMaxM2Config.from_pretrained("MiniMaxAI/MiniMax-M2") # ideally just MiniMaxM2Config() as those should have the respective values already 
config.num_hidden_layers = 7  # check the size, something up to 7-8B should be ok
model = MiniMaxM2Config.from_pretrained(
    "MiniMaxAI/MiniMax-M2",
    config=config,  # we cut the model into x layers, e.g. here 7
    device_map="auto",
    dtype="auto",
    # ... (quants if you need etc)
)
model.save_pretrained("Small-MiniMax-M2")
```

rogeryoungh:
According to my tests, `from_pretrained` does not reduce the number of layers in the model, indicating that the configuration is not being used.

I found a solution: generate the model twice.

1. Load the model using `from_config`, which saves the model with the correct structure but without the correct FP8Linear layers.
2. Use `from_pretrained` again, which allows me to save the correct quantized results.

My complete code is here: https://github.com/rogeryoungh/MiniMaxM2TinyModelGenerator/blob/main/generator.py

ArthurZucker:
```diff
model = MiniMaxM2Config.from_pretrained(
    "MiniMaxAI/MiniMax-M2",
+    num_layers=3,  # we cut the model into x layers, e.g. here 7
-    config=config,
    device_map="auto",
    dtype="auto",
    # ... (quants if you need etc)
)
```
should work with the appropriate name!

vasqu:
@rogeryoungh I pushed a few commits to 
- sync with main
- adjust some values in the config
- fixup modular

With the current state, a conversion could look like this
```python
from transformers import FineGrainedFP8Config, MiniMaxM2Config, MiniMaxM2ForCausalLM


token = "hf_xxx"


config = MiniMaxM2Config()
quantization_config = FineGrainedFP8Config(
    modules_to_not_convert=[
      "gate",
      "e_score_correction_bias",
      "lm_head"
    ]
)

config.num_hidden_layers = 2  # check the size, something up to 7-8B should be ok
model = MiniMaxM2ForCausalLM.from_pretrained(
    "MiniMaxAI/MiniMax-M2",
    config=config,
    device_map="auto",
    dtype="auto",
    quantization_config=quantization_config,
)
model.save_pretrained("MiniMax-M2-Small", push_to_hub=True, repo_id="hf-internal-testing/MiniMax-M2-Small", token=token)
```

However, I discovered that while there are no error logs, the conversion doesn't seem to match the intended shapes. Discussing this internally and following back when I got news cc @MekkCyber 

Edit: Fixed, the quant config needs to be within the config for native fp8 weights, not passed as separate quant config

```python
from transformers import FineGrainedFP8Config, MiniMaxM2Config, MiniMaxM2ForCausalLM


token = "hf_xxx"


config = MiniMaxM2Config()
config.quantization_config = FineGrainedFP8Config(
    modules_to_not_convert=[
      "gate",
      "e_score_correction_bias",
      "lm_head"
    ]
).to_dict()

config.num_hidden_layers = 2  # check the size, something up to 7-8B should be ok
model = MiniMaxM2ForCausalLM.from_pretrained(
    "MiniMaxAI/MiniMax-M2",
    config=config,
    device_map="auto",
    dtype="auto",
)
model.save_pretrained("MiniMax-M2-Small", push_to_hub=True, repo_id="hf-internal-testing/MiniMax-M2-Small", token=token)
```

Edit 2: Our CI Runner can't handle fp8 because it needs an hopper GPU and we only have an ampere :cry: updated to a base version that is randomized
```python
import torch

from transformers import MiniMaxM2Config, MiniMaxM2ForCausalLM, set_seed


token = "hf_xxx"

config = MiniMaxM2Config()
config.head_dim = 128
config.hidden_size = 1024
config.intermediate_size = 2048
config.num_attention_heads = 12
config.num_experts_per_tok = 2
config.num_hidden_layers = 12
config.num_key_value_heads = 2
config.num_local_experts = 16
config.tie_word_embeddings = True

model = MiniMaxM2ForCausalLM(config)
set_seed(42)
model = model.cpu()
with torch.no_grad():
    for name, p in sorted(model.named_parameters()):
        torch.nn.init.normal_(p, mean=0.0, std=0.2)
model.to(torch.float16)
model.save_pretrained("MiniMax-M2-Small", push_to_hub=True, repo_id="hf-internal-testing/MiniMax-M2-Small", token=token)
```

tests/models/minimax_m2/test_modeling_minimax_m2.py
ArthurZucker: something we should add support for IMO @vasqu can you open an issue with this please!

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,416 @@
...
+    ):
+        self.vocab_size = vocab_size
+        self.max_position_embeddings = max_position_embeddings
+        self.hidden_size = hidden_size
+        self.intermediate_size = intermediate_size
+        self.num_hidden_layers = num_hidden_layers
+        self.num_attention_heads = num_attention_heads
+
+        # for backward compatibility
+        if num_key_value_heads is None:
```
ArthurZucker: we should not have BC paths

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,416 @@
...
+class MiniMaxM2TopKRouter(MixtralTopKRouter):
+    def forward(self, hidden_states, e_score_correction_bias):
+        hidden_states = hidden_states.reshape(-1, self.hidden_dim)
+        router_logits = nn.functional.linear(hidden_states, self.weight)  # (seq_len, num_experts)
+        routing_weights = nn.functional.sigmoid(router_logits.float())
+        scores_for_choice = routing_weights + e_score_correction_bias
+        _, top_k_index = torch.topk(scores_for_choice, self.top_k, dim=-1, sorted=False)
+        top_k_weights = routing_weights.gather(1, top_k_index)
+        top_k_weights /= top_k_weights.sum(dim=-1, keepdim=True)
+        router_scores = torch.zeros_like(routing_weights).scatter_(1, top_k_index, top_k_weights)
```
ArthurZucker: this will prevent working with `vllm` using transformers backend! 

We should follow the recent update #42456 
```suggestion
```

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,416 @@
...
+
+class MiniMaxM2RMSNorm(MixtralRMSNorm):
+    pass
+
+
+class MiniMaxM2RotaryEmbedding(Glm4MoeRotaryEmbedding):
+    pass
+
+
+class MiniMaxM2Attention(Qwen3MoeAttention):
```
ArthurZucker: `FlexOlmoAttention` seems to define the same attention no?

src/transformers/models/minimax_m2/modular_minimax_m2.py
ArthurZucker: 👀

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,416 @@
...
+                **kwargs,
+            )
+
+        hidden_states = self.norm(hidden_states)
+
+        return MoeModelOutputWithPast(  # only diff with Mistral is the output type, we need MoE
+            last_hidden_state=hidden_states,
+            past_key_values=past_key_values,
+        )
+
```
ArthurZucker, line 358: here nothing needs to be overwritten AFAIK? same as a normal mixtral no?

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,416 @@
...
+        self.validate_rope(ignore_keys=ignore_keys_at_rope_validation)
+        return kwargs
+
+
+class MiniMaxM2TopKRouter(MixtralTopKRouter):
+    def forward(self, hidden_states, e_score_correction_bias):
+        hidden_states = hidden_states.reshape(-1, self.hidden_dim)
+        router_logits = nn.functional.linear(hidden_states, self.weight)  # (seq_len, num_experts)
+        routing_weights = nn.functional.sigmoid(router_logits.float())
```
ArthurZucker, line 236: I think this is the main diff with other models TLDR sigmoid act for the router no? Let's amphasize this 😉

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -0,0 +1,416 @@
...
+            ignore_keys_at_rope_validation |= {"partial_rotary_factor"}
+
+        self.validate_rope(ignore_keys=ignore_keys_at_rope_validation)
+        return kwargs
+
+
+class MiniMaxM2TopKRouter(MixtralTopKRouter):
+    def forward(self, hidden_states, e_score_correction_bias):
+        hidden_states = hidden_states.reshape(-1, self.hidden_dim)
+        router_logits = nn.functional.linear(hidden_states, self.weight)  # (seq_len, num_experts)
```
ArthurZucker:
```suggestion
        router_logits = F.linear(hidden_states, self.weight)  # (seq_len, num_experts)
```
let's import, this is a bit more standard

ArthurZucker: Image

ArthurZucker (approved):
TY! THere should be a little bit less changes to just isolate what's different!

src/transformers/models/minimax_m2/modular_minimax_m2.py
vasqu: Yup, dependent on #42801 tho since we should not have pos ids typed --> will cause issues with padding free FA (that only uses pos ids to infer)

src/transformers/models/minimax_m2/modular_minimax_m2.py
vasqu: No sliding window here, otherwise we would need to keep an attribute in the config that is unused (sliding window)

tests/models/minimax_m2/test_modeling_minimax_m2.py
vasqu: Opened #42804 to keep track

vasqu:
run-slow: minimax_m2

vasqu:
run-slow: minimax_m2

vasqu:
Deleted the github action comment because I was just checking something which I already reverted

vasqu:
run-slow: minimax_m2

vasqu:
Sorry for the delays, just coming back from the holidays @rogeryoungh. I've synced with the latest changes 
- PR for native transformers integration is at https://huggingface.co/MiniMaxAI/MiniMax-M2.1/discussions/15 (will need to double check if the values are correct for the new model here, it's currently the same as for the base M2 model)
- Likely needing #43154 to have correct Moe behavior with fp8 weights

Will come back to this tomorrow but wanted to give a first insight, should be close to getting merged :hugs:

rogeryoungh:
I found that the current code is not running correctly and throws an error at `self.experts(hidden_states, top_k_index, top_k_weights)`. But I'm not sure if it's a problem with MiniMaxM2TopKRouter or FP8Expert.

I found a way to fix it:

```diff
@@ -59,7 +59,7 @@ class MiniMaxM2TopKRouter(nn.Module):
         _, top_k_index = torch.topk(scores_for_choice, self.top_k, dim=-1, sorted=False)
         top_k_weights = routing_weights.gather(1, top_k_index)
         top_k_weights /= top_k_weights.sum(dim=-1, keepdim=True)
-        router_scores = top_k_weights
+        router_scores = torch.zeros_like(routing_weights).scatter_(1, top_k_index, top_k_weights)
         return router_logits, router_scores, top_k_index
```

vasqu:
@rogeryoungh it will be fixed by #43154, but yes your solution would work as well but we removed that way of routing because scatter caused issues iirc (see #42456)

Will finalize that PR and update this branch then

vasqu:
run-slow: minimax_m2

vasqu:
Ok, now everything works and is synced with the latest changes on main. Hub PRs are open at
- Minimax M2: https://huggingface.co/MiniMaxAI/MiniMax-M2/discussions/52
- Minimax M2.1: https://huggingface.co/MiniMaxAI/MiniMax-M2.1/discussions/15

Lmk if you have any last changes @rogeryoungh, otherwise I'll merge tomorrow as it works from our side then.

ArthurZucker (approved):
Very clean thanks!

PR #42031, opened 2025-11-05
Fix the order of methods in processor loading

zucchini-nlp:
# What does this PR do?

As per title. The order was accidentally changed in prev PR and there's no reason for it. The reproducer is

```python
from transformers import Qwen2VLProcessor
import torch 

repo='hf-internal-testing/tiny-qwenimage-edit-pipe'
load_kwargs = {'subfolder': 'processor', 'variant': None, 'revision': None}
kwargs = {'torch_dtype': torch.float32}

processor = Qwen2VLProcessor.from_pretrained(repo, **load_kwargs, **kwargs)
```

### DIFF
--- a/src/transformers/processing_utils.py
+++ b/src/transformers/processing_utils.py
@@ -1370,8 +1370,8 @@ class ProcessorMixin(PushToHubMixin):
         if token is not None:
             kwargs["token"] = token
 
-        processor_dict, kwargs = cls.get_processor_dict(pretrained_model_name_or_path, **kwargs)
         args = cls._get_arguments_from_pretrained(pretrained_model_name_or_path, **kwargs)
+        processor_dict, kwargs = cls.get_processor_dict(pretrained_model_name_or_path, **kwargs)
         return cls.from_args_and_dict(args, processor_dict, **kwargs)
 
     @classmethod
--- a/tests/models/auto/test_processor_auto.py
+++ b/tests/models/auto/test_processor_auto.py
@@ -88,6 +88,15 @@ class AutoFeatureExtractorTest(unittest.TestCase):
 
         self.assertIsInstance(processor, Wav2Vec2Processor)
 
+    def test_processor_from_local_subfolder_from_repo(self):
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            processor = AutoProcessor.from_pretrained("facebook/wav2vec2-base-960h")
+            processor.save_pretrained(f"{tmpdirname}/processor_subfolder")
+
+            processor = Wav2Vec2Processor.from_pretrained(tmpdirname, subfolder="processor_subfolder")
+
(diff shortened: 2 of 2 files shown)


src/transformers/processing_utils.py
```diff
@@ -1370,8 +1370,8 @@ def from_pretrained(
         if token is not None:
             kwargs["token"] = token
 
-        processor_dict, kwargs = cls.get_processor_dict(pretrained_model_name_or_path, **kwargs)
         args = cls._get_arguments_from_pretrained(pretrained_model_name_or_path, **kwargs)
+        processor_dict, kwargs = cls.get_processor_dict(pretrained_model_name_or_path, **kwargs)
```
sayakpaul, line 1374: Curious -- how does the order matter unless there's something in place going on there? 👀 Does the `kwargs` get consumed?

zucchini-nlp: yes, the kwargs are modified in-place and returned back from `get_processor_dict`

sayakpaul: Ah that's the culprit! I wonder if it could make sense to use `kwargs.copy()` so that we know the original `kwargs` is unchanged and "safe" to be reused.

But I would understand if that's not preferred.

zucchini-nlp: Yeha, that was done for a reason since we use kwargs later to update config keys. So we can allow users to `from_pretrained(model_id, subfolder='x', any_processor_attribute='new_value')`

tests/models/auto/test_processor_auto.py
```diff
@@ -88,6 +88,15 @@ def test_processor_from_local_directory_from_repo(self):
...
 
+    def test_processor_from_local_subfolder_from_repo(self):
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            processor = AutoProcessor.from_pretrained("facebook/wav2vec2-base-960h")
+            processor.save_pretrained(f"{tmpdirname}/processor_subfolder")
+
+            processor = Wav2Vec2Processor.from_pretrained(tmpdirname, subfolder="processor_subfolder")
+
+        self.assertIsInstance(processor, Wav2Vec2Processor)
+
```
molbap, line 99: Thanks for adding a test! Why not use `'hf-internal-testing/tiny-qwenimage-edit-pipe'` though?

molbap (approved):
Looks good, thanks for fixing!

tests/models/auto/test_processor_auto.py
zucchini-nlp: just wanted to not depend on diffusers test repos, in case it is changed in the future to suit diffusers' needs :)

molbap: Fair point!

PR #42036, opened 2025-11-05
fix(granitemoe*): Only create block_sparse_moe if num_local_experts > 0

gabe-l-hart:
Branch: GraniteMoeAsDenseFix

# What does this PR do?

With the introduction of `modular_granitemoe.py` in https://github.com/huggingface/transformers/pull/40132, [the conditional](https://github.com/huggingface/transformers/blob/e6a8e7debed69d4cf5d493158f3f347931c24c9a/src/transformers/models/granitemoe/modeling_granitemoe.py#L511) that allowed `GraniteMoe` to also encapsulate dense models as a degenerate case was accidentally removed. This is never actually needed for the `GraniteMoe` architecture directly, but `GraniteMoe` is reused in `GraniteMoeShared` and then `GraniteMoeHybrid` which do need this ability to also encapsulate dense FFN blocks in place of the MoE block.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker I believe this came in with your PR for MoE in vLLM, so I'd love your sanity check on this fix.


gabe-l-hart:
Looks like I need to regenerate the `modeling_` code everywhere

gabe-l-hart:
Interestingly, when I run `make fix-copies`, the `python utils/check_modular_conversion.py --fix_and_overwrite` script makes far more changes than I would expect from adding this conditional. In particular, for each architecture in the `GraniteMoe*` chain, it adds `GraniteMoe<qualifier>SparseMoeBlock`, then in the `__init__` for `GraniteMoe<qualifier>DecoderLayer`, it adds `self.block_sparse_moe = GraniteMoe<qualifier>SparseMoeBlock(config)` (without any conditional) AND the guarded conditional block `self.block_sparse_moe = GraniteMoeHybridMoE(config)`.

Looking a little deeper, this seems to be caused by the inheritance from `modular_mixtral.py`. I've added a clause that explicitly uses `delattr` to remove `self.block_sparse_moe` when not used, but that seems a bit backwards. Alternatively, we could not inherit from `MixtralDecoderLayer` or we could move the conditional up to `MixtralDecoderLayer.__init__`.

The part that I'm still confused about is why regenerating the `modeling_*` files is adding these `SparseMoeBlock` implementations at all. The inheritance from `Mixtral` was already there prior to my change, so I would expect that those should have already been added unless somehow making creation of `self.block_sparse_moe` conditional triggered logic in the generation to require that those be added?

gabe-l-hart:
It appears that putting creation of `self.block_sparse_moe` behind a conditional does, in fact, trigger the inclusion of those `SparseMoeBlock` pieces in the generation. I've used an inline conditional now which seems to prevent this.

src/transformers/models/granitemoehybrid/modeling_granitemoehybrid.py
```diff
@@ -972,158 +972,21 @@ def extra_repr(self):
...
-
 class GraniteMoeHybridDecoderLayer(GradientCheckpointingLayer):
     def __init__(self, config: GraniteMoeHybridConfig, layer_idx: int):
         super().__init__()
         self.hidden_size = config.hidden_size
         # Either attention or mamba will be initialized, depending on the layer type.
         self.self_attn = None
-        self.block_sparse_moe = GraniteMoeHybridMoE(config)
+        self.block_sparse_moe = (
+            GraniteMoeHybridRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
```
gabe-l-hart: This looks wrong!

src/transformers/models/granitemoe/modeling_granitemoe.py
```diff
@@ -411,11 +411,13 @@ def __init__(self, config: GraniteMoeConfig, layer_idx: int):
         super().__init__()
         self.hidden_size = config.hidden_size
         self.self_attn = GraniteMoeAttention(config=config, layer_idx=layer_idx)
-        self.block_sparse_moe = GraniteMoeMoE(config)
+        self.block_sparse_moe = (
+            GraniteMoeRMSNorm(config.hidden_size, eps=config.rms_norm_eps) if config.num_local_experts > 0 else None
```
gabe-l-hart: This is _very_ wrong! Yikes, will fix

gabe-l-hart:
🤦 Ok, all that was because of a bad copy-paste somewhere that had me creating a completely incorrect block for `self.block_sparse_moe`. Fixed and cleaned up history.

gabe-l-hart:
One more redo: Based on advice from @ArthurZucker, since there are no models using either `GraniteMoe` or `GraniteMoeShared` with the degenerate dense configuration, it's preferable to _only_ have this conditional override in `GraniteMoeHybrid` where it is needed for various flavors of `granite-4.0-*` models.

src/transformers/models/granitemoehybrid/modeling_granitemoehybrid.py
```diff
@@ -926,52 +926,6 @@ def forward(self, x, position_ids):
         return cos.to(dtype=x.dtype), sin.to(dtype=x.dtype)
 
 
-class GraniteFlashAttentionKwargs(TypedDict, total=False):
```
gabe-l-hart, line 930: It looks like these just got moved during the regeneration. I'm not sure if should be included (to enforce consistency with the generation script) or excluded (to minimize the size of the change).

Rocketknight1:
Hi @gabe-l-hart, thanks for the PR! You can get the code style tests to pass with `pip install -e .[quality]` followed by `make fixup` or `make style`.

Overall it looks good to me, and it does seem like the zero-experts case was accidentally deleted. Will wait for @arthurzucker to confirm before merging!

gabe-l-hart:
@Rocketknight1 Thanks! I'll get it cleaned up and hopefully green today

Rocketknight1:
Yep, looks good now! If you don't get core maintainer approval within a few days, ping me and I'll see what I can do

gabe-l-hart:
@Rocketknight1 @ArthurZucker I lost track of this one. I just resolved the conflict with the GH web ui, but I didn't look to see if there are other changes that would conflict (ie this being solved at a higher level in the class hierarchy). Do we still need this PR to support non-hybrid/non-moe with `GraniteMoeHybrid`? If so, can we get this looked over?

src/transformers/models/granitemoehybrid/modeling_granitemoehybrid.py
Rocketknight1: Don't stress about the modeling file - if the modular is correct then `make fix-copies` should handle it all

Rocketknight1 (approved):
Yes, I'm happy to approve now! The change makes sense, and given the existence of `has_experts`, the 0-expert path is clearly intended but just broken.

PR #42042, opened 2025-11-05
Fix transformers serve following chat template output update

LysandreJik:
cc @Rocketknight1 :)


Rocketknight1 (approved):
Yes, sorry about missing that one! You can also do `return_dict=False`, but it's all equivalent.

PR #42043, opened 2025-11-05
Fix bnb for the weights refactor

SunMarc:
# What does this PR do?

This PR fixes bnb support (8bit + 4bit) in the new weight loading logic. 

## Testing 

```python
from transformers import AutoTokenizer, AutoModelForCausalLM, BitsAndBytesConfig

model_name = "meta-llama/Llama-3.2-3B-Instruct"
quantization_config = BitsAndBytesConfig(load_in_4bit=True, bnb_4bit_quant_type="nf4")

#model_name = "unsloth/Llama-3.2-3B-Instruct-bnb-4bit"
# don't pass quantization_config

tokenizer = AutoTokenizer.from_pretrained(model_name)

    model_name,
    quantization_config=quantization_config,
    device_map=0
)

input_text = "Write me a poem about Machine Learning."
input_ids = tokenizer(input_text, return_tensors="pt").to("cuda")

outputs = model.generate(**input_ids, do_sample=False, max_new_tokens=1024)
print(tokenizer.decode(outputs[0]))
```

- [ ] check why the memory is way too high when quantizing on the fly
- [x] bnb tests


src/transformers/integrations/finegrained_fp8.py
```diff
@@ -568,15 +568,7 @@ def replace_with_fp8_linear(
...
     return model
-
-
-class QuantizationOp(ConversionOps):
-    """Base class for quantization operations."""
-
-    pass
-
-
-class Fp8Quantize(QuantizationOp):
```
MekkCyber, line 579: agree! makes more sense to remove it

src/transformers/integrations/bitsandbytes.py
```diff
@@ -26,7 +27,16 @@
 
 logger = logging.get_logger(__name__)
 
+from ..core_model_loading import ConversionOps
 
+class Bnb4bitQuantize(ConversionOps):
+    def convert(self, input_dict: torch.Tensor, model: Optional[torch.nn.Module] = None, **kwargs) -> dict[str, torch.Tensor]:
```
MekkCyber: we can also dequantize in bnb right ?

MekkCyber (commented):
Niice! makes sense

src/transformers/integrations/bitsandbytes.py
SunMarc: yes but it will be better to ask users to do model.dequantize()

src/transformers/integrations/__init__.py
```diff
@@ -196,7 +198,7 @@
     )
     from .eetq import replace_with_eetq_linear
     from .fbgemm_fp8 import FbgemmFp8Linear, FbgemmFp8Llama4TextExperts, replace_with_fbgemm_fp8_linear
-    from .finegrained_fp8 import FP8Linear, replace_with_fp8_linear
+    from .finegrained_fp8 import FP8Linear, Fp8Quantize, replace_with_fp8_linear
```
ArthurZucker: i think we removed it because it was importing with a jit derocator making it slow

src/transformers/integrations/bitsandbytes.py
ArthurZucker: nice!

src/transformers/integrations/bitsandbytes.py
```diff
@@ -26,6 +29,44 @@
...
+        else:
+            full_name = target_key
+            # update param name to get the weights instead of the quantized stats
+            target_key = self.hf_quantizer.get_param_name(target_key)
+            module, _ = get_module_from_name(model, target_key)
+
+            module_name = target_key.rsplit(".", 1)[0]
+            # Save the states for later quantization when they are all gathered
+            if not hasattr(self.hf_quantizer, "param_quant_stats"):
+                self.hf_quantizer.param_quant_stats = defaultdict(dict)
```
ArthurZucker, line 67: they are gathered from what? sorry I am not familiar with it, you need which states?

src/transformers/integrations/finegrained_fp8.py
ArthurZucker: yep

src/transformers/quantizers/base.py
```diff
@@ -418,6 +418,17 @@ def _convert_model_for_quantization(self, model):
...
+        raise NotImplementedError(
+            f"{self.quantization_config.quant_method} is not available yet and will be supported soon."
+        )
+
+    def is_valid_unexpected_keys(self, k):
+        """ 
+        Check if the keys is valid or not even if it is not in the state_dict of the meta model.
+        This is because the state dict of the model might change after quantization like for 4bit bnb
+        """
+        return False
```
ArthurZucker: i would love to avoid this, meaning can we make sure the 1st call to `hf_quantizer.quantize_model` just properly prepares the meta model?

src/transformers/core_model_loading.py
```diff
@@ -280,19 +301,22 @@ class ConversionEntry:
...
 
 
-def _materialize_copy(tensor, dtype):
+def _materialize_copy(tensor, dtype=None):
     # PyTorch: this runs in C and releases the GIL; good for threads.
-    return tensor[...].to(dtype)
+    tensor = tensor[...]
+    if dtype is not None:
+        tensor = tensor.to(dtype)
+    return tensor
```
ArthurZucker, line 303: yep

src/transformers/core_model_loading.py
```diff
@@ -357,11 +381,17 @@ def set_param_for_module(
...
+    hf_quantizer,
 ):
     with log_to_misc(layer_name, misc, layer_name):
         module_path, _, param_name = layer_name.rpartition(".")
         module_obj = model.get_submodule(module_path) if module_path else model
-        param_value = param_value[0] if isinstance(param_value, list) else param_value[...]
+        if isinstance(param_value, list):
+            param_value = param_value[0]
+        elif isinstance(param_value, torch.nn.Parameter):
+            pass
```
ArthurZucker: I am guessing that's for BNB? Can't we force it to return a data instead of a param?

src/transformers/core_model_loading.py
```diff
@@ -378,7 +408,8 @@ def set_param_for_module(
                 pass  # TODO for "local" stuff, it will trigger missmatched no?
             param_value = torch.nn.Parameter(param_value, requires_grad=param_value.is_floating_point())
 
-        if ref is not None and ref.shape != param_value.shape:
+        # skip mismatch for hf_quantizer for now
+        if ref is not None and ref.shape != param_value.shape and hf_quantizer is None:
```
ArthurZucker, line 406: why is the shape of the BnbLinear not correct? this I also don't think we want long term no?

src/transformers/core_model_loading.py
```diff
@@ -458,16 +489,19 @@ def convert_and_load_state_dict_in_model(
...
             if empty_param is None:
-                unexpected_keys.add(t)
-                continue
-
-            if quantizer is not None and quantizer.param_needs_quantization(model, t):
-                if quantizer.__class__.__name__ == "FineGrainedFP8HfQuantizer":
-                    from .integrations.finegrained_fp8 import Fp8Quantize
-
-                    converter.quantization_operation = Fp8Quantize()  # TODO support other methods
+                if hf_quantizer is not None and hf_quantizer.is_valid_unexpected_keys(t):
```
ArthurZucker: ditto 😉

src/transformers/core_model_loading.py
```diff
@@ -458,16 +489,19 @@ def convert_and_load_state_dict_in_model(
...
+                    unexpected_keys.add(t)
+                    continue
+    
+            if hf_quantizer is not None and hf_quantizer.param_needs_quantization(model, t):
+                converter.quantization_operation = hf_quantizer.get_quantize_ops()
+                # TODO: to clean later. We need to use the empty_param from the checkpoint to decide if we upcast the param to a specific dtype
+                k_dtype = tensor.get_dtype()
+                dtype = str_to_torch_dtype[k_dtype]
+                empty_param_checkpoint = torch.empty(size=tensor.get_shape(), dtype=dtype, device="meta")
+                _, _dtype = _infer_parameter_dtype(model, t, empty_param_checkpoint, hf_quantizer)
```
ArthurZucker: why? is it because BNB needs say bf16 always? can you elaborate here because I don't upcast any of the parameters, they just have the `_dtype` on meta, and then get whatever was loaded from the weights

src/transformers/core_model_loading.py
```diff
@@ -531,9 +565,7 @@ def convert_and_load_state_dict_in_model(
                             if op := converter.quantization_operation:
                                 with log_to_misc(layer_name, misc, op=op):
                                     realized_value.update(
-                                        op.convert(
-                                            {k: realized_value.pop(k)}, quant_config=quantizer.quantization_config
-                                        )
+                                        op.convert({k: realized_value.pop(k)}, model=model)
```
ArthurZucker: not a fan of passing the whole model!

src/transformers/core_model_loading.py
```diff
@@ -580,3 +613,34 @@ def revert_weight_conversion(model, state_dict):
...
+            QuantizationMethod.MXFP4,
+            QuantizationMethod.BITS_AND_BYTES,
+        }:
+            return True, None
+        else:
+            raise e
+    is_torch_e4m3fn_available = hasattr(torch, "float8_e4m3fn")
+    # We convert floating dtypes to the `dtype` passed except for float8_e4m3fn type. We also want to keep the buffers/params
+    # in int/uint/bool and not cast them.
+    casting_dtype = None
```
ArthurZucker: can't the methods do this in the `quantize` because they should!

ArthurZucker (commented):
nice!

src/transformers/core_model_loading.py
```diff
@@ -580,3 +613,34 @@ def revert_weight_conversion(model, state_dict):
         original_state_dict[key] = value
     state_dict = original_state_dict
     return state_dict
+
+def _infer_parameter_dtype(
+    model: torch.nn.Module,
+    param_name: str,
+    empty_param: torch.Tensor,
+    hf_quantizer: Optional[HfQuantizer] = None,
+) -> Union[bool, Optional[torch.dtype]]:
```
matthewdouglas:
```suggestion
) -> tuple[bool, Optional[torch.dtype]]:
```

src/transformers/integrations/__init__.py
SunMarc: ok, I will import it directly from the correct file then

src/transformers/integrations/bitsandbytes.py
SunMarc: Basically, we need to store some parameters to create the quantized weight. For example, bnb requires 6 values that are stored in the checkpoint to recover the quantized weight. So we store them in a dict that it stored in hf_quantizer for now as we can't save it in the op since we create an op per tensor.

src/transformers/quantizers/base.py
SunMarc: This is more to take care of the case where we load the quantized checkpoint. I don't think there is a good way to fix this but let's keep this for now. We can think of fixing this later

src/transformers/core_model_loading.py
SunMarc: if you do tensor[...], the nn.parameters becomes a tensor. So the isinstance check later becomes invalid. But yeah this is for bnb as we need to return a nn.Params4bit which is a subclass of `torch.nn.Parameter` and I didn't want to pollute this function with hf_quantizer related logic as much as possible.

src/transformers/core_model_loading.py
SunMarc: this is because when we initialize the meta model with nn.Linear4bit, those don't have the right shape as the weights are not quantized yet. But yeah maybe we can fix this by overwriting the shape of the param when replacing the layers. In long term, we will remove this yes.

src/transformers/core_model_loading.py
SunMarc: We need to infer the right `dtype` for each values in the checkpoints:
- some of the values are not parameters or buffers of the model so we shouldn't change the dtype
- for some parameters / buffers, we should also keep the same dtype as the checkpoint (`empty_param_checkpoint`) because the `_dtype` on meta is not correct ... (fp16 instead of int8) . But this can be fixed potentially if we initialize the correct `dtype`. For bnb it should work but not sure for other method like torchao as the dtype is hard to infer from the beginning.

src/transformers/core_model_loading.py
SunMarc: I wish I could not do that but let's keep this for now

src/transformers/core_model_loading.py
```diff
@@ -280,85 +303,59 @@ class ConversionEntry:
...
+            inst = super().__new__(cls, from_existing.data, from_existing.requires_grad, **from_existing.__dict__)
+            inst._original_param = from_existing
+            # Explicitly override all in-place methods per instance
+            for method_name in inst._inplace_methods:
+                setattr(inst, method_name, MethodType(inst._skip, inst))
+
+            return inst
+
+        def _skip(self, *args, **kwargs):
+            """Helper to skip in-place operations."""
```
SunMarc: simplified a bit to accomodate subclass of nn.parameters cc @ArthurZucker. Also if we are using this class, it means the param is initialized as you said so let's simplify everything

src/transformers/core_model_loading.py
ArthurZucker: but if `quantize` we never change the dtype of the param, which is the source of truth

src/transformers/core_model_loading.py
ArthurZucker: yep absolutely, ty

SunMarc:
run-slow: bnb

src/transformers/integrations/accelerate.py
```diff
@@ -435,6 +435,7 @@ def _get_device_map(
             if max_memory is not None and device_name in max_memory:
                 inferred_max_memory[device_name] = min(inferred_max_memory[device_name], max_memory[device_name])
 
+        model.tie_weights()
```
ArthurZucker:
```suggestion
        model.tie_weights() # TODO make sure to remove this later on if possible, does not make sense
```

src/transformers/integrations/bitsandbytes.py
```diff
@@ -26,6 +29,91 @@
...
 logger = logging.get_logger(__name__)
 
+from ..core_model_loading import ConversionOps
+
+
+class Bnb4bitQuantize(ConversionOps):
+    def __init__(self, hf_quantizer):
+        self.hf_quantizer = hf_quantizer
+
+    def convert(self, input_dict: torch.Tensor, model: Optional[torch.nn.Module] = None, missing_keys = None, **kwargs) -> dict[str, torch.Tensor]:
```
ArthurZucker:
```suggestion
    def convert(self, input_dict: torch.Tensor, model: Optional[torch.nn.Module] = None, missing_keys = None, **kwargs) -> dict[str, torch.Tensor]:
    """
    we need to store some parameters to create the quantized weight. For example, bnb requires 6 values that are stored in the checkpoint to recover the quantized weight. So we store them in a dict that it stored in hf_quantizer for now as we can't save it in the op since we create an op per tensor.
    """
```

src/transformers/quantizers/quantizer_bnb_4bit.py
```diff
@@ -118,7 +117,10 @@ def adjust_target_dtype(self, target_dtype: "torch.dtype") -> "torch.dtype":
         return CustomDtype.INT4
 
     def update_unexpected_keys(self, model, unexpected_keys: list[str]) -> list[str]:
-        return [k for k in unexpected_keys if not any(k.endswith(x) for x in self.bnb_keys)]
+        return [k for k in unexpected_keys if not self.is_valid_unexpected_keys(k)]
+
+    def is_valid_unexpected_keys(self, k):
+        return any(k.endswith(x) for x in self.bnb_keys)
```
ArthurZucker: i'd rather we have a regex pre-compiled here

src/transformers/core_model_loading.py
```diff
@@ -612,13 +632,13 @@ def convert_and_load_state_dict_in_model(
...
-                if quantizer.__class__.__name__ == "FineGrainedFP8HfQuantizer":
-                    from .integrations.finegrained_fp8 import Fp8Quantize
-
-                    converter.quantization_operation = Fp8Quantize()  # TODO support other methods
-                else:
-                    raise ValueError("This quantization method is gonna be supported SOOOON")
+            if hf_quantizer is not None and hf_quantizer.param_needs_quantization(model, t):
+                converter.quantization_operation = hf_quantizer.get_quantize_ops()
+                # TODO: to clean later. We need to use the empty_param from the checkpoint to decide if we upcast the param to a specific dtype
+                k_dtype = tensor.get_dtype()
```
ArthurZucker: > this can be fixed potentially if we initialize the correct dtype 

let's go with this instead it makes much more sense to do the things right! 
So for both the unexpected keys and this. 

I am fine with updating the missing keys

src/transformers/modeling_utils.py
```diff
@@ -4093,26 +4098,30 @@ def from_pretrained(
...
-            )
+            for m in [model, copy_model]:
+                hf_quantizer.preprocess_model(
+                    model=m,
+                    device_map=device_map,
+                    keep_in_fp32_modules=model._keep_in_fp32_modules,  # TODO prob no longer needed?
+                    config=config,
+                    checkpoint_files=checkpoint_files,
+                    use_kernels=use_kernels,
+                )
```
ArthurZucker: let's find a way to remove the copy model and shut accelerate warning up please, we should not have to forcefully tie weights twice nor have a copy

ArthurZucker (commented):
very nice !
Let's keep updating missing, I think it does make sense, but not unexpected: let's enforce the contract of preparing the model on meta at least!

SunMarc:
@bot /style

src/transformers/core_model_loading.py
SunMarc: removed for now !

src/transformers/quantizers/quantizer_bnb_4bit.py
SunMarc: removed

src/transformers/modeling_utils.py
SunMarc: removed the copy, let's tie for now

SunMarc:
@bot /style

SunMarc:
@bot /style

src/transformers/core_model_loading.py
```diff
@@ -496,20 +484,14 @@ def convert_and_load_state_dict_in_model(
...
-            else:
-                _dtype = dtype
-                matched_dtype_pattern = match_glob(t, dtype_policy_alt, dtype_policy_by_group_name)
-                if matched_dtype_pattern is not None:
-                    _dtype = dtype_plan[matched_dtype_pattern]
-                elif empty_param.dtype != _dtype:
-                    _dtype = empty_param.dtype
+            if hf_quantizer is not None and hf_quantizer.param_needs_quantization(model, t):
+                converter.quantization_operation = hf_quantizer.get_quantize_ops()
+            _dtype = dtype
```
ArthurZucker, line 492: nice

ArthurZucker (approved):
LGTM.Just tie twice is my nightmare but good otherwise

SunMarc:
run-slow: bnb

SunMarc:
> LGTM.Just tie twice is my nightmare but good otherwise

I've upstreamed some code from accelerate to fix tied weights. This should be easier this way and we can better tweak device map related code in the future.

SunMarc:
@bot /style

SunMarc:
@bot /style

SunMarc:
run-slow: bnb

src/transformers/integrations/accelerate.py
```diff
@@ -547,3 +547,380 @@ def accelerate_disk_offload(
...
+    elif len(gpus) > 0:
+        main_devices = [gpus[0], "cpu"]
+    else:
+        main_devices = ["cpu"]
+
+    module_sizes, _ = compute_module_sizes(model, hf_quantizer)
+
+    if tied_parameters is None and len(model.all_tied_weights_keys) > 0:
+        # create a list of list of tied params
+        tied_parameters = [list(t) for t in model.all_tied_weights_keys.items()]
```
SunMarc: changed this

src/transformers/integrations/accelerate.py
```diff
@@ -547,3 +547,380 @@ def accelerate_disk_offload(
...
+def infer_auto_device_map(
+    model: nn.Module,
+    max_memory: Optional[dict[Union[int, str], Union[int, str]]] = None,
+    no_split_module_classes: Optional[list[str]] = None,
+    verbose: bool = False,
+    clean_result: bool = True,
+    offload_buffers: bool = False,
+    tied_parameters: Optional[list[list[str]]] = None,
+    hf_quantizer: "HfQuantizer | None" = None,
+):
```
SunMarc, line 627: removed `dtype` and `special_dtype` to rely on `hf_quantizer` instead when computing `compute_module_sizes`

PR #42044, opened 2025-11-05
Remove unused functions in `image_transforms.py`

yaswanth19:
Delete a unused function and class from image_transforms.py file.


yaswanth19:
@yonigozlan for review. Here `_cast_tensor_to_float` func was unused hence removed it

yonigozlan (commented):
I don't think there's a need to change anything here. The case you described should be already supported by using `is_nested=True` in `group_images_by_shape` and `reorder_images`. It will flattened the `batch_size` and `num_crops` dimensions, then reconstruct them both.

yaswanth19:
@yonigozlan Updated the PR to now only remove the unused function and class. It's not used anywhere in the transformers codebase a I doubt these would be used in external libs as they seem trivial

yonigozlan (approved):
Hello @yaswanth19 ! Thanks for cleaning that up. Did you manage to fix your issue using `is_nested=True` then?

PR #42045, opened 2025-11-05
Add llama 4 scaling

juliendenize:
# What does this PR do?

Add llama 4 scaling for long context to Mistral models.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@patrickvonplaten @ArthurZucker @zucchini-nlp

### DIFF
--- a/src/transformers/models/mistral/configuration_mistral.py
+++ b/src/transformers/models/mistral/configuration_mistral.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 """Mistral model configuration"""
 
-from typing import Optional, TypedDict
+from typing import Optional
 
 from ...configuration_utils import PreTrainedConfig
 from ...modeling_rope_utils import RopeParameters, rope_config_validation, standardize_rope_params
@@ -22,21 +22,6 @@ from ...utils import logging
 
 
 logger = logging.get_logger(__name__)
-
-
-class LLama4Scaling(TypedDict):
-    r"""
-    The scaling parameters to apply LLama 4 scaling to the rope embeddings.
-
-    Args:
-        original_max_position_embeddings (`int`):
-            The original max position embeddings used during pretraining.
-        scaling_beta (`float`):
-            The scaling beta parameter.
-    """
-
-    original_max_position_embeddings: int
-    scaling_beta: float
 
 
 class MistralConfig(PreTrainedConfig):
@@ -97,8 +82,6 @@ class MistralConfig(PreTrainedConfig):
             Dictionary containing the configuration parameters for the RoPE embeddings. The dictionaty should contain
             a value for `rope_theta` and optionally parameters used for scaling in case you want to use RoPE
             with longer `max_position_embeddings`.
-        llama_4_scaling (`LLama4Scaling`, *optional*):
-            Dictionary containing the scaling parameters for LLama 4 scaling.
         sliding_window (`int`, *optional*, defaults to 4096):
             Sliding window attention window size. If not specified, will default to `4096`.
         attention_dropout (`float`, *optional*, defaults to 0.0):
--- a/src/transformers/models/mistral/configuration_mistral.py
+++ b/src/transformers/models/mistral/configuration_mistral.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 """Mistral model configuration"""
 
-from typing import Optional
+from typing import Optional, TypedDict
 
 from ...configuration_utils import PreTrainedConfig
 from ...modeling_rope_utils import RopeParameters, rope_config_validation, standardize_rope_params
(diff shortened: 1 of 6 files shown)


src/transformers/models/mistral/configuration_mistral.py
```diff
@@ -137,6 +154,7 @@ def __init__(
         eos_token_id: Optional[int] = 2,
         tie_word_embeddings: Optional[bool] = False,
         rope_parameters: Optional[RopeParameters | dict[str, RopeParameters]] = None,
+        llama_4_scaling: Optional[LLama4Scaling] = None,
```
zucchini-nlp: can we have it in `rope_parameters` instead? Looks more intuitive to save all RoPE related args in one place. Since there are already parameters for RoPE scaling in the dict, we can add these as `llama4_{param_name}`

src/transformers/models/mistral3/configuration_mistral3.py
```diff
@@ -103,7 +103,7 @@ def __init__(
                 num_attention_heads=16,
                 vocab_size=32000,
                 head_dim=64,
-                hidden_act="gelu",
+                hidden_act="silu",
```
zucchini-nlp, line 106: might be a bit breaking for configs that were saved without `hidden_act` and therefore defaulted to GeLU prev

src/transformers/models/mistral3/convert_mistral3_weights_to_hf.py
```diff
@@ -206,13 +240,21 @@ def convert_and_write_processor(input_dir: str, output_dir: str):
         image_processor=image_processor,
         image_token="[IMG]",
         patch_size=patch_size,
-        chat_template=chat_template,
```
zucchini-nlp, line 209: model has no chat template anymore? 🥲

src/transformers/models/mistral3/configuration_mistral3.py
juliendenize: I think it was broken before actually, it is a fix here.

src/transformers/models/mistral3/convert_mistral3_weights_to_hf.py
juliendenize: It makes no sense anymore to put one here unfortunately, depending on the models the chat template looks very different (tokenizer version, Thinking or not, ...). So having a default one is arguably worse than none at all imo.

src/transformers/models/mistral/modular_mistral.py
```diff
@@ -70,6 +77,9 @@ def forward(
         cos, sin = position_embeddings
         query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)
 
+        if self.config.rope_parameters.llama_4_scaling_beta is not None:
```
zucchini-nlp: `rope_parameters` is a simple dict so we better safe-get `rope_parameters.get(`llama_4_scaling_beta')

zucchini-nlp:
Look good, let's just address one comment and fix CI so it's ✅

ArthurZucker (commented):
LGTM! We will need to add a new model for this change to get through, as we never change old models to add new features 🤗 

The best way is to write a simple `modular_new_model.py` with something like this: 
```python
...

class MyModelAttention(MistralAttention):
    def forward(
        self,
        hidden_states: torch.Tensor,
        position_embeddings: tuple[torch.Tensor, torch.Tensor],
        attention_mask: Optional[torch.Tensor],
        **kwargs: Unpack[FlashAttentionKwargs],
    ) -> tuple[torch.Tensor, Optional[torch.Tensor]]:
        input_shape = hidden_states.shape[:-1]
        hidden_shape = (*input_shape, -1, self.head_dim)

        query_states = self.q_proj(hidden_states).view(hidden_shape).transpose(1, 2)
        key_states = self.k_proj(hidden_states).view(hidden_shape).transpose(1, 2)
        value_states = self.v_proj(hidden_states).view(hidden_shape).transpose(1, 2)

        cos, sin = position_embeddings
        query_states = query_states * self._get_llama4_attn_scale(cache_position).to(query_states.dtype)
        query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)

            # sin and cos are specific to RoPE models; cache_position needed for the static cache

        attn_output, attn_weights = attention_interface(
            self,
            query_states,
            key_states,
            value_states,
            attention_mask,
            scaling=self.scaling,
            sliding_window=getattr(self.config, "sliding_window", None),  # main diff with Llama
            **kwargs,
        )

        attn_output = attn_output.reshape(*input_shape, -1).contiguous()
        attn_output = self.o_proj(attn_output)
```
the best would be to modify the `RopeEmbedding` to do the scaling in there, this way you don't have to change the forward pass!

PR #42050, opened 2025-11-06
Fix KeyError in _is_package_available for packages with dotted names

### LINKED ISSUE #41981: KeyError: 'optimum.quanto' for `_is_package_available`
### System Info

When i run pytest -rA tests/models/video_llama_3/test_modeling_video_llama_3.py::VideoLlama3ModelTest::test_generate_with_quant_cache, it will fail and return error:
```
pkg_name = 'optimum.quanto', return_version = True

    def _is_package_available(pkg_name: str, return_version: bool = False) -> tuple[bool, str] | bool:
        """Check if `pkg_name` exist, and optionally try to get its version"""
        spec = importlib.util.find_spec(pkg_name)
        package_exists = spec is not None
        package_version = "N/A"
        if package_exists and return_version:
            try:
                # importlib.metadata works with the distribution package, which may be different from the import
                # name (e.g. `PIL` is the import name, but `pillow` is the distribution name)
>               distributions = PACKAGE_DISTRIBUTION_MAPPING[pkg_name]
E               KeyError: 'optimum.quanto'

src/transformers/utils/import_utils.py:56: KeyError
```
I suggest one approach to solve this is add `KeyError` to except line in [L61](https://github.com/huggingface/transformers/blob/main/src/transformers/utils/import_utils.py#L61), is it OK?

### Who can help?

@SunMarc FYI

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

1. git clone https://github.com/huggingface/transformers.git
2. pip install optimum-quanto
3. export RUN_SLOW=1
4. pytest -rA tests/models/video_llama_3/test_modeling_video_llama_3.py::VideoLlama3ModelTest::test_generate_with_quant_cache

### Expected behavior

The test case is expected to pass.

yashwantbezawada:
# What does this PR do?

Fixes #41981

I ran into this issue while testing with `optimum.quanto`. The function was crashing with `KeyError` when checking for packages with dotted names.

The problem is in `import_utils.py` line 56 - when it tries to look up the package in `PACKAGE_DISTRIBUTION_MAPPING`, packages with dots in the name (like `optimum.quanto`) aren't always there. The mapping might only have `optimum` as a key, not `optimum.quanto`.

The function already has fallback logic to handle cases where package metadata can't be found - it imports the package directly and checks for `__version__`. But the exception handler only caught `PackageNotFoundError`, not `KeyError`.

I added `KeyError` to the exception handler so the fallback works for both cases. Now when a package name isn't in the mapping, it'll use the same fallback path instead of crashing.

This is a one-line change that handles the edge case without affecting normal operation. The existing tests should still pass since the behavior for normal packages is unchanged.

### DIFF
--- a/src/transformers/utils/import_utils.py
+++ b/src/transformers/utils/import_utils.py
@@ -58,7 +58,7 @@ def _is_package_available(pkg_name: str, return_version: bool = False) -> tuple[
             # pick the first item of the list as best guess (it's almost always a list of length 1 anyway)
             distribution_name = pkg_name if pkg_name in distributions else distributions[0]
             package_version = importlib.metadata.version(distribution_name)
-        except importlib.metadata.PackageNotFoundError:
+        except (importlib.metadata.PackageNotFoundError, KeyError):
             # If we cannot find the metadata (because of editable install for example), try to import directly.
             # Note that this branch will almost never be run, so we do not import packages for nothing here
             package = importlib.import_module(pkg_name)


Rocketknight1:
Yes, this makes sense to me! Since it should only have an impact in situations that were already causing a crash, I think it's safe to merge this.

Rocketknight1 (approved):

PR #42052, opened 2025-11-06
fix test failure of speculative_generation on xpu

sywangyi:
- Intel XPU: @IlyasMoutawwakil

fix pytest failure of following case
tests/generation/test_candidate_generator.py::TestUniversalSpeculativeDecoding::test_usd_vs_vanilla_sampling
tests/models/mistral/test_modeling_mistral.py::MistralIntegrationTest::test_speculative_generation
tests/models/qwen2/test_modeling_qwen2.py::Qwen2IntegrationTest::test_speculative_generation
tests/models/qwen3/test_modeling_qwen3.py::Qwen3IntegrationTest::test_speculative_generation
tests/models/whisper/test_modeling_whisper.py::WhisperModelIntegrationTests::test_speculative_decoding_non_distil
tests/models/whisper/test_modeling_whisper.py::WhisperModelIntegrationTests::test_speculative_decoding_distil
tests/models/sam2_video/test_modeling_sam2_video.py::Sam2VideoModelIntegrationTest::test_inference_propagate_on_streamed_video
tests/models/sam2_video/test_modeling_sam2_video.py::Sam2VideoModelIntegrationTest::test_inference_mask_generation_video_one_point_propagate_in_video_directly
tests/models/univnet/test_modeling_univnet.py::UnivNetModelIntegrationTests::test_integration

### DIFF
--- a/src/transformers/masking_utils.py
+++ b/src/transformers/masking_utils.py
@@ -251,7 +251,9 @@ def _ignore_causal_mask_sdpa(
     if (
         not is_tracing(padding_mask)
         # only cases when lower and upper diags are the same, see https://github.com/pytorch/pytorch/issues/108108
-        and (query_length == 1 or (kv_length == query_length or _is_torch_xpu_available))
+        and (
+            query_length == 1 or (kv_length == query_length or (_is_torch_xpu_available and padding_mask is not None))
+        )
         # in this case we need to add special patterns to the mask so cannot be skipped otherwise
         and (local_attention_size is None or kv_length < local_attention_size)
         # In this case, we need to add padding to the mask, so cannot be skipped otherwise
@@ -260,7 +262,7 @@ def _ignore_causal_mask_sdpa(
             or (
                 padding_mask.all()
                 if not _is_torch_xpu_available or query_length == 1
-                else padding_mask[:, :query_length].all()
+                else (padding_mask[:, :query_length].all() and not padding_mask[:, query_length:].any())
             )
         )
     ):
--- a/src/transformers/masking_utils.py
+++ b/src/transformers/masking_utils.py
@@ -280,12 +280,16 @@ def _ignore_causal_mask_sdpa(
         mask_indices += kv_offset
         padding_mask = padding_mask[:, mask_indices]
 
+    if _is_torch_xpu_available:
+        # XPU devices have special handling for mask skipping:
+        # - Single query tokens use the same logic as CUDA
+        # - Multi-query tokens can skip if padding_mask is provided and correctly structured
+        #   (all True in query window, all False after)
+        return _can_skip_causal_mask_xpu(padding_mask, query_length, kv_length, local_attention_size)
     # When using `torch.export` or `torch.onnx.dynamo_export`, we must pass an example input, and `is_causal` behavior is
     # hard-coded to the forward. If a user exports a model with query_length > 1, the exported model will hard-code `is_causal=True`
     # which is in general wrong (see https://github.com/pytorch/pytorch/issues/108108). Thus, we only set
     # `ignore_causal_mask = True` if we are not tracing
-    if _is_torch_xpu_available:
-        return _can_skip_causal_mask_xpu(padding_mask, query_length, kv_length, local_attention_size)
     if (
         not is_tracing(padding_mask)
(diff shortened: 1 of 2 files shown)


sywangyi:
@vasqu could you help review it, thanks

src/transformers/masking_utils.py
```diff
@@ -251,7 +251,9 @@ def _ignore_causal_mask_sdpa(
     if (
         not is_tracing(padding_mask)
         # only cases when lower and upper diags are the same, see https://github.com/pytorch/pytorch/issues/108108
-        and (query_length == 1 or (kv_length == query_length or _is_torch_xpu_available))
+        and (
+            query_length == 1 or (kv_length == query_length or (_is_torch_xpu_available and padding_mask is not None))
```
vasqu: So xpu can only skip the mask if a mask is provided? We can also remove one `()`

src/transformers/masking_utils.py
```diff
@@ -260,7 +262,7 @@ def _ignore_causal_mask_sdpa(
             or (
                 padding_mask.all()
                 if not _is_torch_xpu_available or query_length == 1
-                else padding_mask[:, :query_length].all()
+                else (padding_mask[:, :query_length].all() and not padding_mask[:, query_length:].any())
```
vasqu: Can we add a comment here for what this entails or rather why we need this check explicitly

tests/models/qwen3/test_modeling_qwen3.py
```diff
@@ -165,7 +165,7 @@ def test_model_600m_long_prompt_sdpa(self):
     def test_speculative_generation(self):
         EXPECTED_TEXT_COMPLETIONS = Expectations(
             {
-                ("xpu", 3): "My favourite condiment is 100% peanut butter. I love it so much that I can't help but use it",
+                ("xpu", 3): "My favourite condiment is 100% beef and comes in a 12 oz. jar. It is sold in",
```
vasqu, line 168: Is this expected? I'd like to avoid changing outputs unless it was faulty before already

vasqu (commented):
The conditionals are getting longer and longer. Maybe it would be nicer to split these into cuda vs xpu and have proper explanations/comments at each point?

Edit: ~Not sure if the change in expected outputs is good, depends if things were faulty before (which I assume was the case)~ it's expected and was clarified under this review

tests/models/qwen3/test_modeling_qwen3.py
yao-matrix: @vasqu, it's expected, since XPU ops are under active development and performance tuning, each op level bug fixing or finetuning will potentially bring numerical minor changes and may reflect in the final output text, but the output is still meaningful. We intel team is keeping close monitoring the transformers UT status, and co-working w/ our PyTorch team on this. When op development is stable like CUDA, we'll expect no more these changes.

vasqu: Gotcha, thx for the quick clarification! That totally makes sense, just wanted to make sure we really know what we were doing here

sywangyi:
> The conditionals are getting longer and longer. Maybe it would be nicer to split these into cuda vs xpu and have proper explanations/comments at each point?
> 
> Edit: ~Not sure if the change in expected outputs is good, depends if things were faulty before (which I assume was the case)~ it's expected and was clarified under this review

I split xpu out and add comment in the func

src/transformers/masking_utils.py
```diff
@@ -248,21 +284,16 @@ def _ignore_causal_mask_sdpa(
     # hard-coded to the forward. If a user exports a model with query_length > 1, the exported model will hard-code `is_causal=True`
     # which is in general wrong (see https://github.com/pytorch/pytorch/issues/108108). Thus, we only set
     # `ignore_causal_mask = True` if we are not tracing
+    if _is_torch_xpu_available:
+        return _can_skip_causal_mask_xpu(padding_mask, query_length, kv_length, local_attention_size)
```
vasqu: Can we move this above the comments above "When using..." and add a comment above this if to indicate the special xpu case, e.g. something along your docstring in the new function.

vasqu (approved):
Thx, lgtm let's just move this one if a bit and add a comment there

vasqu:
Thx for iterating, merging now

PR #41589, opened 2025-10-14
Allow VLMs to have a correct `base_model`

zucchini-nlp:
# What does this PR do?

As per title, being able to simply call `model.base_model` is a useful feat for models which we didn't support in VLMs

There will be no more weird patter like this after this PR:

```python
            if hasattr(model, "visual"):
                vision_module = model.visual
            elif hasattr(model, "model") and hasattr(model.model, "visual"):
                vision_module = model.model.visual


# Now we can
vision_module = model.base_model.visual
```

Also PR fixes loading base model in Llava-like models which was accidentally deleted. We still need a conversion mapping for legacy checkpoints

```python
# throws bunch of missing keys in main and works with this PR
model  = AutoModel.from_pretrained('llava-hf/llava-1.5-7b-hf') 
```

PR #42059, opened 2025-11-06
Fix base model prefix in VLMs

zucchini-nlp:
# What does this PR do?

Fiexes https://github.com/huggingface/transformers/issues/41814. The incorrect prefix is throwing errors when loadiing a base model from checkpoint

It was a bad copy from llava, where the prefix is empty string to support BC with old-style checkpoints

Update: just realized I had a PR already in https://github.com/huggingface/transformers/pull/41589 😅  which solves this issue + adds new functionality

### DIFF
--- a/src/transformers/models/cohere2_vision/modular_cohere2_vision.py
+++ b/src/transformers/models/cohere2_vision/modular_cohere2_vision.py
@@ -345,7 +345,7 @@ class Cohere2VisionImageProcessorFast(GotOcr2ImageProcessorFast):
 
 __all__ = [
     "Cohere2VisionForConditionalGeneration",
-    "Cohere2VisionPreTrainedModel",  # noqa: F822
+    "Cohere2VisionPreTrainedModel",
     "Cohere2VisionModel",
     "Cohere2VisionImageProcessorFast",
 ]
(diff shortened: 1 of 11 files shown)


src/transformers/models/cohere2_vision/modular_cohere2_vision.py
```diff
@@ -89,6 +90,10 @@ class Cohere2VisionCausalLMOutputWithPast(AyaVisionCausalLMOutputWithPast):
     pass
 
 
+class Cohere2VisionPreTrainedModel(AyaVisionPreTrainedModel):
+    base_model_prefix = "model"
+
```
molbap, line 95: we don't need to add it to _all_ ? or maybe modular did it already

molbap (approved):
Should make things easier :grin:

src/transformers/models/cohere2_vision/modular_cohere2_vision.py
zucchini-nlp: you mean all VLM models? Not really, other VLMs have a conversion-mapping which is not compatible with base prefix. I opened another PR to make it compatible but waiting for core maintainers review

molbap: no, I meant `__all__` at bottom of the file, haha

zucchini-nlp: oh, you're right! My bad, has to be added

PR #42070, opened 2025-11-06
[core] Fix mxfp4 model loading

MekkCyber:
# What does this PR do?


src/transformers/integrations/mxfp4.py
```diff
@@ -48,6 +53,61 @@
...
+                    setattr(
+                        module,
+                        f"{proj}_precision_config",
+                        PrecisionConfig(weight_scale=weight_scale, flex_ctx=FlexCtx(rhs_data=InFlexData())),
+                    )
+                    missing_keys.discard(f"{target_key}_blocks")
+                    missing_keys.discard(f"{target_key}_scales")
+                    delattr(module, f"{proj}_blocks")
+                    delattr(module, f"{proj}_scales")
+
```
SunMarc, line 89: since we are setting this here, chek what I did for loadedparam. This is to make sure that we don't re-initialize the weights

SunMarc (commented):
Nice, just a nit

SunMarc:
also another nit is that arthur prefer that we don't update the unexpected keys at all and deal with missing keys instead (even that it will be better if we don't touch that at all). Check the bnb PR.

src/transformers/integrations/mxfp4.py
```diff
@@ -48,6 +53,61 @@
...
+                    setattr(module, proj, triton_weight_tensor)
+                    setattr(
+                        module,
+                        f"{proj}_precision_config",
+                        PrecisionConfig(weight_scale=weight_scale, flex_ctx=FlexCtx(rhs_data=InFlexData())),
+                    )
+                    missing_keys.discard(f"{target_key}_blocks")
+                    missing_keys.discard(f"{target_key}_scales")
+                    delattr(module, f"{proj}_blocks")
+                    delattr(module, f"{proj}_scales")
```
ArthurZucker, line 88: `Mxfp4` should not even need to delete anything AFAIK because it just uses the blocks and scales !

PR #42083, opened 2025-11-07
fix qwen2vl/qwen3vl video processor temporal padding when num_frames%temporal_patch_size!=1

yaogang2060:
# What does this PR do?
expect qwen3vl video processor can process these two cases:
1. num_frames is 1, and sample_frames is false.
2. num_frames > temporal_patch_size and num_frames % temporal_patch_size != 1

Fixes # (issue)
https://github.com/QwenLM/Qwen3-VL/issues/1689

### DIFF
--- a/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
@@ -196,10 +196,6 @@ class Qwen3VLVideoProcessor(BaseVideoProcessor):
 
         for shape, stacked_videos in grouped_videos.items():
             if do_resize:
-                T = stacked_videos.shape[1]
-                if pad := -T % temporal_patch_size:
-                    repeats = stacked_videos[:, -1:].expand(-1, pad, -1, -1, -1)
-                    stacked_videos = torch.cat((stacked_videos, repeats), dim=1)
                 B, T, C, H, W = stacked_videos.shape
                 num_frames, height, width = T, H, W
                 resized_height, resized_width = smart_resize(
--- a/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
@@ -40,8 +40,6 @@ def smart_resize(
     min_pixels: int = 128 * 128,
     max_pixels: int = 16 * 16 * 2 * 2 * 2 * 6144,
 ):
-    if num_frames < temporal_factor:
-        raise ValueError(f"t:{num_frames} must be larger than temporal_factor:{temporal_factor}")
     if height < factor or width < factor:
         raise ValueError(f"height:{height} or width:{width} must be larger than factor:{factor}")
     elif max(height, width) / min(height, width) > 200:
@@ -50,7 +48,7 @@ def smart_resize(
         )
     h_bar = round(height / factor) * factor
     w_bar = round(width / factor) * factor
-    t_bar = round(num_frames / temporal_factor) * temporal_factor
+    t_bar = math.ceil(num_frames / temporal_factor) * temporal_factor
 
     if t_bar * h_bar * w_bar > max_pixels:
         beta = math.sqrt((num_frames * height * width) / max_pixels)
--- a/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
@@ -195,9 +195,13 @@ class Qwen3VLVideoProcessor(BaseVideoProcessor):
         resized_videos_grouped = {}
 
         for shape, stacked_videos in grouped_videos.items():
-            B, T, C, H, W = stacked_videos.shape
-            num_frames, height, width = T, H, W
             if do_resize:
+                T = stacked_videos.shape[1]
+                if pad := -T % temporal_patch_size:
+                    repeats = stacked_videos[:, -1:].expand(-1, pad, -1, -1, -1)
+                    stacked_videos = torch.cat((stacked_videos, repeats), dim=1)
+                B, T, C, H, W = stacked_videos.shape
+                num_frames, height, width = T, H, W
                 resized_height, resized_width = smart_resize(
                     num_frames=num_frames,
                     height=height,
@@ -232,9 +236,10 @@ class Qwen3VLVideoProcessor(BaseVideoProcessor):
             patches = stacked_videos
 
             # Check that videos have `num_frames` divisible by `temporal_patch_size`
-            if patches.shape[1] % temporal_patch_size != 0:
-                repeats = patches[:, -1:].repeat(1, temporal_patch_size - 1, 1, 1, 1)
-                patches = torch.cat([patches, repeats], dim=1)
+            T = patches.shape[1]
+            if pad := -T % temporal_patch_size:
+                repeats = patches[:, -1:].expand(-1, pad, -1, -1, -1)
+                patches = torch.cat((patches, repeats), dim=1)
             batch_size, grid_t, channel = patches.shape[:3]
             grid_t = grid_t // temporal_patch_size
             grid_h, grid_w = resized_height // patch_size, resized_width // patch_size
--- a/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
@@ -195,9 +195,9 @@ class Qwen3VLVideoProcessor(BaseVideoProcessor):
         resized_videos_grouped = {}
 
         for shape, stacked_videos in grouped_videos.items():
+            B, T, C, H, W = stacked_videos.shape
+            num_frames, height, width = T, H, W
             if do_resize:
-                B, T, C, H, W = stacked_videos.shape
-                num_frames, height, width = T, H, W
                 resized_height, resized_width = smart_resize(
                     num_frames=num_frames,
                     height=height,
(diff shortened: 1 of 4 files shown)


yaogang2060:
@zucchini-nlp please check this~

src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
```diff
@@ -195,9 +195,13 @@ def _preprocess(
...
         for shape, stacked_videos in grouped_videos.items():
-            B, T, C, H, W = stacked_videos.shape
-            num_frames, height, width = T, H, W
             if do_resize:
+                T = stacked_videos.shape[1]
+                if pad := -T % temporal_patch_size:
+                    repeats = stacked_videos[:, -1:].expand(-1, pad, -1, -1, -1)
+                    stacked_videos = torch.cat((stacked_videos, repeats), dim=1)
+                B, T, C, H, W = stacked_videos.shape
+                num_frames, height, width = T, H, W
```
zucchini-nlp: i don't think this is needed if we are expanding it later, just before petchifying

src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
```diff
@@ -232,9 +236,10 @@ def _preprocess(
             patches = stacked_videos
 
             # Check that videos have `num_frames` divisible by `temporal_patch_size`
-            if patches.shape[1] % temporal_patch_size != 0:
-                repeats = patches[:, -1:].repeat(1, temporal_patch_size - 1, 1, 1, 1)
-                patches = torch.cat([patches, repeats], dim=1)
+            T = patches.shape[1]
+            if pad := -T % temporal_patch_size:
+                repeats = patches[:, -1:].expand(-1, pad, -1, -1, -1)
+                patches = torch.cat((patches, repeats), dim=1)
```
zucchini-nlp, line 236: great, I see that qwen2VL's video processor has the same issue. I thought it was fixed but apparently there was a regression. Can you update it as well?

tests/models/qwen3_vl/test_video_processing_qwen3_vl.py
```diff
@@ -328,3 +328,39 @@ def test_call_sample_frames(self):
...
+
+            n, w, h = 1, 64, 64
+            video_inputs = [(np.random.randint(0, 256, (h, w, 3), dtype=np.uint8)) for _ in range(n)]
+
+            video_processed = video_processing(video_inputs, return_tensors="pt")
+            encoded_videos = video_processed[self.input_name]
+            self.assertEqual(list(encoded_videos.shape), [16, 2304])
+
+            video_grid_thw = video_processed["video_grid_thw"]
+            self.assertEqual(video_grid_thw.tolist(), [[1, 4, 4]])
```
zucchini-nlp: Kind of the same test as `test_videos_PIL` ig, so it is redundant. I think the below one for temporal patch size is enough

tests/models/qwen3_vl/test_video_processing_qwen3_vl.py
```diff
@@ -328,3 +328,39 @@ def test_call_sample_frames(self):
...
+
+    def test_num_frames_equal_temporal_patch_size_plus_two(self):
+        for video_processing_class in self.video_processor_list:
+            video_processor_dict = self.video_processor_dict.copy()
+            video_processor_dict["size"] = {"longest_edge": 40960, "shortest_edge": 4096}
```
zucchini-nlp: just for my understanding, do we need to change the size? It should not be affecting the final results and keeping it small ensures that tests are run fast

zucchini-nlp (commented):
Thanks a lot for the PR! Can you also update the Qwen2VL video processor with the same changes?

src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
yaogang2060: if `resize` is enabled and `num_frames` < `temporal_patch_size`,  resize check will throw an error: this line: https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py#L44.
so i think here is needed for cases such as one video has only one image.

tests/models/qwen3_vl/test_video_processing_qwen3_vl.py
yaogang2060: this test case is: one video has just one frame. `test_videos_pil` is not exactly one frame. i update the test function name.

tests/models/qwen3_vl/test_video_processing_qwen3_vl.py
yaogang2060: down the size to 32 * 32, if smaller, `smart_resize` will change it too.

src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
yaogang2060: done~ 
did not change `smart_resize`. 
the `smart_resize` in qwen2vl and qwen3vl video processor is not same, is it correct? @JJJYmmm

tests/models/qwen3_vl/test_video_processing_qwen3_vl.py
zucchini-nlp: might be missing smth, why do we need a test for single frame? The issue solved by this PR is only about temporal padding 

never mind, just saw another comment above

src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
zucchini-nlp: Ah interesting. I remember we fixes similar issue with qwen2-vl series in https://github.com/huggingface/transformers/pull/38706

Imo we should not be restricting the inputs to be certain size before resizing, so we can adjust the `t_bar` so it is never zero

src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
JJJYmmm: Yes, for qwen3vl, I copied the `smart_resize` from glm4v, which takes video frames into account—making it more flexible for usage.

src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py
JJJYmmm: Agree. How about moving this logic to `sample_frames` instead of `_preprocess`? If users sample frames themselves, they’d be responsible for ensuring at least `temporal_patch_size` frames—though that’s rarely an issue since videos typically have more than two frames. By the way, when would a “video” have only one frame? Wouldn’t that just be an image? 🤔

yaogang2060: > Agree. How about moving this logic to `sample_frames` instead of `_preprocess`? If users sample frames themselves, they’d be responsible for ensuring at least `temporal_patch_size` frames—though that’s rarely an issue since videos typically have more than two frames. By the way, when would a “video” have only one frame? Wouldn’t that just be an image? 🤔

the cases i am using is: I have videos from 7 cam views, and i want every video has different frames, maybe some view has only one frame. And i want to use video processor to preprocess these videos, so i come to this...

yaogang2060: ok. guys, i delete the temporal padding when num_frames < temporal_patch_size.....  let this pr be simple.🫡

zucchini-nlp (approved):
LGTM now! Do you plan to update `smart_resize` in this PR or make a separate one?

yaogang2060:
> Member



> LGTM now! Do you plan to update `smart_resize` in this PR or make a separate one?

@JJJYmmm  it's up to you🫡. maybe just use `smart_resize` in qwen2vl?

JJJYmmm:
@yaogang2060 As mentioned in https://github.com/huggingface/transformers/pull/42083#discussion_r2509274897, you can remove the frame number check (or warning) at  
https://github.com/huggingface/transformers/blob/59994f0841f1a3641ced696ebfb477572aabbfcf/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py#L43-L44
and instead use `ceil` at  
https://github.com/huggingface/transformers/blob/59994f0841f1a3641ced696ebfb477572aabbfcf/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py#L53
to ensure `t_bar` is never zero.

yaogang2060:
> @yaogang2060 As mentioned in [#42083 (comment)](https://github.com/huggingface/transformers/pull/42083#discussion_r2509274897), you can remove the frame number check (or warning) at
> 
> https://github.com/huggingface/transformers/blob/59994f0841f1a3641ced696ebfb477572aabbfcf/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py#L43-L44
> 
> 
> and instead use `ceil` at
> https://github.com/huggingface/transformers/blob/59994f0841f1a3641ced696ebfb477572aabbfcf/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py#L53
> 
> 
> to ensure `t_bar` is never zero.



> @yaogang2060 As mentioned in [#42083 (comment)](https://github.com/huggingface/transformers/pull/42083#discussion_r2509274897), you can remove the frame number check (or warning) at
> 
> https://github.com/huggingface/transformers/blob/59994f0841f1a3641ced696ebfb477572aabbfcf/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py#L43-L44
> 
> 
> and instead use `ceil` at
> https://github.com/huggingface/transformers/blob/59994f0841f1a3641ced696ebfb477572aabbfcf/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py#L53
> 
> 
> to ensure `t_bar` is never zero.

just removed~

PR #42085, opened 2025-11-07
Fix helper fn for new processor config format

zucchini-nlp:
# What does this PR do?

As per title, these helpers are used in vLLM and don't work if we save a processor after v5. Now all processor's subcomponents are saved in one single place in `processor_config.json`, so we need to check it as well when trying to locate the config file

fyi @hmellor

### DIFF
--- a/src/transformers/feature_extraction_utils.py
+++ b/src/transformers/feature_extraction_utils.py
@@ -437,7 +437,7 @@ class FeatureExtractionMixin(PushToHubMixin):
                 # Load from local folder or from cache or download from model Hub and cache
                 resolved_feature_extractor_files = [
                     resolved_file
-                    for filename in [feature_extractor_file, PROCESSOR_NAME]
+                    for filename in [PROCESSOR_NAME, feature_extractor_file]
                     if (
                         resolved_file := cached_file(
                             pretrained_model_name_or_path,
--- a/src/transformers/feature_extraction_utils.py
+++ b/src/transformers/feature_extraction_utils.py
@@ -427,35 +427,42 @@ class FeatureExtractionMixin(PushToHubMixin):
             feature_extractor_file = os.path.join(pretrained_model_name_or_path, FEATURE_EXTRACTOR_NAME)
         if os.path.isfile(pretrained_model_name_or_path):
             resolved_feature_extractor_file = pretrained_model_name_or_path
+            resolved_processor_file = None
             is_local = True
         elif is_remote_url(pretrained_model_name_or_path):
             feature_extractor_file = pretrained_model_name_or_path
+            resolved_processor_file = None
             resolved_feature_extractor_file = download_url(pretrained_model_name_or_path)
         else:
             feature_extractor_file = FEATURE_EXTRACTOR_NAME
             try:
                 # Load from local folder or from cache or download from model Hub and cache
-                resolved_feature_extractor_files = [
-                    resolved_file
-                    for filename in [PROCESSOR_NAME, feature_extractor_file]
-                    if (
-                        resolved_file := cached_file(
-                            pretrained_model_name_or_path,
-                            filename=filename,
-                            cache_dir=cache_dir,
-                            force_download=force_download,
-                            proxies=proxies,
-                            local_files_only=local_files_only,
-                            subfolder=subfolder,
-                            token=token,
-                            user_agent=user_agent,
-                            revision=revision,
-                            _raise_exceptions_for_missing_entries=False,
-                        )
-                    )
-                    is not None
-                ]
-                resolved_feature_extractor_file = resolved_feature_extractor_files[0]
+                resolved_processor_file = cached_file(
+                    pretrained_model_name_or_path,
+                    filename=PROCESSOR_NAME,
+                    cache_dir=cache_dir,
+                    force_download=force_download,
+                    proxies=proxies,
+                    local_files_only=local_files_only,
+                    token=token,
+                    user_agent=user_agent,
+                    revision=revision,
+                    subfolder=subfolder,
+                    _raise_exceptions_for_missing_entries=False,
+                )
+                resolved_feature_extractor_file = cached_file(
+                    pretrained_model_name_or_path,
+                    filename=feature_extractor_file,
+                    cache_dir=cache_dir,
+                    force_download=force_download,
+                    proxies=proxies,
+                    local_files_only=local_files_only,
+                    token=token,
+                    user_agent=user_agent,
+                    revision=revision,
+                    subfolder=subfolder,
+                    _raise_exceptions_for_missing_entries=False,
+                )
             except OSError:
                 # Raise any environment error raise by `cached_file`. It will have a helpful error message adapted to
                 # the original exception.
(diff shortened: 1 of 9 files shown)


src/transformers/models/auto/feature_extraction_auto.py
```diff
@@ -167,27 +167,40 @@ def get_feature_extractor_config(
...
         return {}
 
+    resolved_config_file = resolved_config_files[0]
     with open(resolved_config_file, encoding="utf-8") as reader:
-        return json.load(reader)
+        feature_extractor_dict = json.load(reader)
+    if "audio_processor" in feature_extractor_dict:
+        feature_extractor_dict = feature_extractor_dict["audio_processor"]
+    else:
+        feature_extractor_dict = feature_extractor_dict.get("feature_extractor", feature_extractor_dict)
```
molbap: I'm a bit confused by the logic here, can you remind me why we need to `get` this / why `audio_processor` is directly accessible?

src/transformers/models/auto/image_processing_auto.py
```diff
@@ -305,27 +306,38 @@ def get_image_processor_config(
...
+        is not None
+    ]
+    # An empty list if none of the possible files is found in the repo
+    if not resolved_config_files:
         logger.info(
             "Could not locate the image processor configuration file, will try to use the model config instead."
         )
         return {}
 
+    resolved_config_file = resolved_config_files[0]
```
molbap: So this means if two are present we don't take the `PROCESSOR`, why do we iterate on both?

tests/models/auto/test_processor_auto.py
```diff
@@ -107,6 +114,48 @@ def test_processor_from_local_directory_from_extractor_config(self):
 
         self.assertIsInstance(processor, Wav2Vec2Processor)
 
+    def test_subcomponent_get_config_dict__saved_as_nested_config(self):
```
molbap: thanks for adding a test!

molbap (approved):
Thanks, looks fine to me! just a couple questions

src/transformers/models/auto/feature_extraction_auto.py
zucchini-nlp: This is to get the actual config from nested structure if we're loading from `processor_config.json`. In audio models we have no standardization unfortunately, and some call the attribute as `audio_processor` or `feature_extractor`

We need to get any of the two keys if available, otherwise just return

src/transformers/models/auto/image_processing_auto.py
zucchini-nlp: good point, I haven't noticed that we're prioritizing `IMAGE_PROCESSOR` here. Imo the priority should be as follows if we find more than one file on the hub:

`PROCESSOR -> IMAGE_PROCESSOR` and for videos `PROCESSOR -> VIDEO_PROCESSOR -> IMAGE_PROCESSOR`

Do you think this priority makes sense? I will update accordingly

tests/models/auto/test_processor_auto.py
zucchini-nlp: for myself: there's a typo of double underscore. To fix before merging

src/transformers/models/auto/feature_extraction_auto.py
molbap: Crystal clear, thanks, actual audio processors will help a lot when they arrive, cc @eustlb for reference

src/transformers/models/auto/image_processing_auto.py
molbap: Yeah it respects the encapsulation of concerns for this set of class, IMO ok!

tests/models/auto/test_processor_auto.py
molbap: I saw, didn't want to add a suggestion for that :D should have, my bad

src/transformers/models/auto/feature_extraction_auto.py
zucchini-nlp: Oh yes, much needed!

src/transformers/models/auto/image_processing_auto.py
zucchini-nlp: oh wait, this is not good bc `PROCESSOR` is almost always present in the repo. But not always contains a nested config for all sub-components

So we can't blindly give priority to `PROCESSOR` if it's found in model repo. I will think of smth

zucchini-nlp:
Oke, now it works. We have to load both (legacy vs non-legacy) configs and then decide which one takes priority because it depends on their content

src/transformers/models/auto/feature_extraction_auto.py
```diff
@@ -178,16 +179,50 @@ def get_feature_extractor_config(
...
+    # An empty list if none of the possible files is found in the repo
+    if not resolved_feature_extractor_file and not resolved_processor_file:
+        logger.info("Could not locate the feature extractor configuration file.")
         return {}
 
-    with open(resolved_config_file, encoding="utf-8") as reader:
-        return json.load(reader)
+    # Load feature_extractor dict. Priority goes as (nested config if found -> feature extractor config)
+    # We are downloading both configs because almost all models have a `processor_config.json` but
+    # not all of these are nested. We need to check if it was saved recebtly as nested or if it is legacy style
```
molbap: Thanks for including comments lol, not trivial

src/transformers/models/auto/feature_extraction_auto.py
```diff
@@ -178,16 +179,50 @@ def get_feature_extractor_config(
...
+    feature_extractor_dict = {}
+    if resolved_processor_file is not None:
+        try:
+            with open(resolved_processor_file, encoding="utf-8") as reader:
+                text = reader.read()
+            processor_dict = json.loads(text)
+        except json.JSONDecodeError:
+            raise OSError(f"It looks like the config file at '{resolved_processor_file}' is not a valid JSON file.")
```
molbap: could be another exception type, why not capture and re-raise as part of the message?

src/transformers/models/auto/feature_extraction_auto.py
molbap:
```suggestion
    # not all of these are nested. We need to check if it was saved recently as nested or if it is legacy style
```

src/transformers/models/auto/image_processing_auto.py
```diff
@@ -316,16 +318,50 @@ def get_image_processor_config(
...
+    # An empty list if none of the possible files is found in the repo
+    if not resolved_image_processor_file and not resolved_processor_file:
+        logger.info("Could not locate the image processor configuration file.")
         return {}
 
-    with open(resolved_config_file, encoding="utf-8") as reader:
-        return json.load(reader)
+    # Load image_processor dict. Priority goes as (nested config if found -> image processor config)
+    # We are downloading both configs because almost all models have a `processor_config.json` but
+    # not all of these are nested. We need to check if it was saved recebtly as nested or if it is legacy style
```
molbap:
```suggestion
    # not all of these are nested. We need to check if it was saved recently as nested or if it is legacy style
```

src/transformers/models/auto/image_processing_auto.py
```diff
@@ -316,16 +318,50 @@ def get_image_processor_config(
...
+            image_processor_dict = processor_dict["image_processor"]
+
+    if resolved_image_processor_file is not None and image_processor_dict is None:
+        try:
+            with open(resolved_image_processor_file, encoding="utf-8") as reader:
+                text = reader.read()
+            image_processor_dict = json.loads(text)
+        except json.JSONDecodeError:
+            raise OSError(
+                f"It looks like the config file at '{resolved_image_processor_file}' is not a valid JSON file."
```
molbap: a util like _safe_json_open could help a tad

src/transformers/feature_extraction_utils.py
```diff
@@ -469,19 +476,38 @@ def get_feature_extractor_dict(
...
 
-        except json.JSONDecodeError:
+        if feature_extractor_dict is None:
             raise OSError(
-                f"It looks like the config file at '{resolved_feature_extractor_file}' is not a valid JSON file."
+                f"Can't load feature extractor for '{pretrained_model_name_or_path}'. If you were trying to load"
+                " it from 'https://huggingface.co/models', make sure you don't have a local directory with the"
+                f" same name. Otherwise, make sure '{pretrained_model_name_or_path}' is the correct path to a"
+                f" directory containing a {feature_extractor_file} file"
             )
```
molbap, line 499: order of operations seems ok to me

molbap (approved):
Re-approving after re-reading 😁 thanks for the iteration, not much except proposing to refacto the json opening a bit as many repeated bits

src/transformers/models/auto/feature_extraction_auto.py
zucchini-nlp: i see it was always raising a clear message for only for incorrectly serialized files, while other types of Errors are raised as is. So I decided to leave it as is, but in general we can just fallback to json's own errors for all cases. I think they are pretty straighforward to understand

zucchini-nlp: maybe in the future

molbap: yeah agreed, json own error handling is robust

PR #42088, opened 2025-11-07
Prefill-related logic in input preparation for generation

### LINKED ISSUE #40910: Gemma-3: prepare_inputs_for_generation should forward pixel_values based on image token presence, not cache_position==0
### System Info

- `transformers` version: 4.56.1
- Platform: Windows-11-10.0.26100-SP0
- Python version: 3.12.11
- Huggingface_hub version: 0.34.4
- Safetensors version: 0.6.2
- Accelerate version: 1.10.1
- PyTorch version (accelerator?): 2.8.0+cu129 (CUDA)
- Jax version: not installed
- Using GPU in script?: yes
- GPU type: NVIDIA GeForce RTX 3060 Laptop GPU

### Who can help?

@zucchini-nlp 

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

[Prefix caching](https://huggingface.co/docs/transformers/kv_cache#prefill-a-cache-prefix-caching) with models that use  `Gemma3ForConditionalGeneration` is impossible when the non-cached prompt has an image due to code in `src/transformers/models/gemma3/modeling_gemma3.py:1175`

```python
# If we're in cached decoding stage, pixel values should be None because input ids do not contain special image token anymore
# Otherwise we need pixel values to be passed to model. NOTE: use_cache=False needs pixel_values always
if cache_position[0] == 0:
    model_inputs["pixel_values"] = pixel_values
```

Step to reproduce:
```python
import copy
from typing import Any, Dict, List
import torch
from transformers import AutoModelForImageTextToText, AutoProcessor


model_id = f"google/medgemma-4b-it"

model = AutoModelForImageTextToText.from_pretrained(
    model_id, dtype=torch.bfloat16, device_map="auto"
)
processor = AutoProcessor.from_pretrained(model_id)


# Generate prefix cache
system_instruction = "You are a helpful assistant.\n\nDescribe the following image."
conversation = [
    {"role": "system", "content": [{"type": "text", "text": system_instruction}]},
    {"role": "user", "content": ""},
]
tokens = processor.apply_chat_template(
    conversation, add_generation_prompt=False, tokenize=True, return_tensors="pt"
)
eot_id = model.config.eos_token_id[-1]
eot_pos = torch.nonzero(tokens == eot_id).max()
initial_prompt = tokens[:, :eot_pos]


with torch.no_grad():
    prompt_cache = model(input_ids=initial_prompt.to(model.device), use_cache=True).past_key_values


# Inference with PREFIX + USER PROMPT with image
def build_conversation(system_instruction: str, url: str) -> List[Dict[str, Any]]:
    """Builds chat messages with a system instruction and a single image."""
    return [
        {"role": "system", "content": [{"type": "text", "text": system_instruction}]},
        {"role": "user", "content": [{"type": "image", "url": url}]},
    ]
image_url = "http://images.cocodataset.org/val2017/000000039769.jpg"
conversation = build_conversation(system_instruction, image_url)
inputs = processor.apply_chat_template(
    conversation,
    add_generation_prompt=True,
    tokenize=True,
    return_dict=True,
    return_tensors="pt",
)

with torch.inference_mode():
    batch_inputs = inputs.to(model.device, dtype=model.dtype)
    past_key_values = copy.deepcopy(prompt_cache)
    generation = model.generate(
        **batch_inputs,
        max_new_tokens=100,
        do_sample=False,
    )
```

After caching a prefix, when we pass a user turn that includes image tokens at non-zero `cache_position`, `prepare_inputs_for_generation` drops `pixel_values` (it only forwards them when `cache_position[0] == 0`).

### Expected behavior

* `prepare_inputs_for_generation` should **forward `pixel_values` whenever the current `input_ids` slice contains the model’s image special token(s)**, regardless of whether `cache_position[0]` is 0 or >0. I.e. `cache_position[0] == 0` is not a good way to understand if we are in the decoding stage.
* This enables a common workflow: cache a **text system prompt** once, then later process a **user turn with an image** without re-encoding the system text, while still exercising the vision path.
* A minimal change would be to replace the strict `cache_position[0] == 0` gate with a check on image token presence
  * If `(input_ids == config.image_token_index).any():` → pass `pixel_values`.
  * Else, omit `pixel_values` during decode.

### LINKED ISSUE #41863: `generate()` produces incoherent output when `inputs_embeds` has length 1
### System Info

- Platform: Linux-6.8.0-51-generic-x86_64-with-glibc2.35
- Python version: 3.10.18
- Safetensors version: 0.6.2
- Accelerate version: 1.10.1
- PyTorch version (accelerator?): 2.8.0+cu128 (CUDA)
- Jax version: not installed
- GPU type: NVIDIA H100 PCIe

### Who can help?

 @xadupre (original author of code in question), @zucchini-nlp 

### Information

- [x] My own modified scripts

### Tasks

#### Steps to Reproduce

```python
import torch

# Load any causal LM
model = AutoModelForCausalLM.from_pretrained("gpt2")
tokenizer = AutoTokenizer.from_pretrained("gpt2")
model.eval()

# Create a single embedding (simulating a prefix like a style token)
single_embedding = torch.randn(1, 1, 768)  # [batch=1, seq_len=1, hidden_dim=768]

# Generate with length-1 inputs_embeds
with torch.no_grad():
    outputs = model.generate(
        inputs_embeds=single_embedding,
        max_length=20,
        do_sample=True,
        temperature=1.0,
        pad_token_id=tokenizer.eos_token_id,
    )

# Decode and observe gibberish
generated_text = tokenizer.decode(outputs[0], skip_special_tokens=True)
print("Generated text:", generated_text)
# Output will be incoherent repetitive tokens like "the the the I if the..."
```

#### Comparison with working case (length ≥ 2):

```python
# Add a second embedding (e.g., BOS token embedding)
bos_embedding = model.get_input_embeddings()(torch.tensor([[tokenizer.bos_token_id]]))
two_embeddings = torch.cat([single_embedding, bos_embedding], dim=1)  # [1, 2, 768]

# Generate with length-2 inputs_embeds
with torch.no_grad():
    outputs = model.generate(
        inputs_embeds=two_embeddings,
        max_length=20,
        do_sample=True,
        temperature=1.0,
        pad_token_id=tokenizer.eos_token_id,
    )

generated_text = tokenizer.decode(outputs[0], skip_special_tokens=True)
print("Generated text:", generated_text)
# Output will be coherent
```

#### Expected Behavior

`generate()` should produce coherent text regardless of whether `inputs_embeds` has length 1 or length > 1, as long as the embeddings are valid.

#### Actual Behavior

With length-1 `inputs_embeds`, `generate()` produces incoherent, repetitive gibberish that appears to be high-frequency tokens without proper conditioning on previous context.

#### Suggested Fix

The `_cache_dependant_input_preparation` method needs to properly handle the transition from `inputs_embeds` mode to `input_ids` mode after the first generation step. Specifically:

1. After the first token is generated from `inputs_embeds`, set `inputs_embeds = None` for subsequent iterations
2. Or, maintain proper bookkeeping so that the embeddings prefix is correctly tracked throughout the autoregressive loop
3. Or, ensure Exception 4 logic properly handles the case where we've transitioned from embeddings to token IDs

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/41863 and fixes https://github.com/huggingface/transformers/issues/40910

We always have had an imperfect way to infer if we're in prefill or decoding stage, which caused us many bugs in the past. The most reliable way is to check cache position values but it is not compile-compatible and also has an edge case

Recently Manuel merged a PR to split prefill into its own function so now we can benefit from it and know with 100% certainty which stage we're in. This PR adds `is_first_iteration` flag to generation input preparation and replaces existing logic with the flag

Note in some models, we have to keep checking `if cache_position[0] == 0` because first iteration does not mean first in total. We might get cached system prompt and we don't want to call some methods a second time (e.g. Qwen mRoPE) 

Also it adds a test case for the above linked issue


zucchini-nlp:
Another worm of cans, assisted decoding has no prefill separated out and is causing issues now 😢

manueldeprada:
> Another worm of cans, assisted decoding has no prefill separated out and is causing issues now 😢

worm of cans?? 🤣 haha love it

Sooo this already arose on my PR. The main gist is that assisted generate does not prefill with the prompt tokens, but waits for the first batch of candidates and then prefills. Thus, we could not apply the standard prefill. But surely assisted_gen can pass the prefill flag on the first call, or we can also maybe call _prefill with the first batch of candidates.

zucchini-nlp:
> assisted_gen can pass the prefill flag on the first call

yeah, this seemed to be the easiest option. The only issue with VLMs is that we should not be passing certain inputs (pixels/etc) after a prefill phase. But with assistant model calling `generate()` many times internally, we end up with several "prefill" phases

zucchini-nlp:
Support for `continue_generation_from_past_cache` is making it hard 😭  The concept of `is_prefill` and `is_first_iteration` can be different in some models and require different input preparation

I don't want us to multiplicate number of input args for `prepare_inputs_generation` so I might leave the `cache_position[0] == 0` as the ultimate source of truth for prefill, and pass only `is_first_iteration` as an argument.

zucchini-nlp:
@bot /style

src/transformers/generation/candidate_generator.py
```diff
@@ -304,19 +307,39 @@ def _update_past_and_masks(
...
+            generation_args = self.assistant_model._get_initial_cache_position(
+                input_ids.shape[1], input_ids.device, self.assistant_kwargs
+            )
+            generation_args = self.assistant_model.prepare_inputs_for_generation(
+                input_ids, is_first_iteration=True, **generation_args
+            )
+            generation_args[self.input_ids_key] = input_ids
+            for model_input_name in ["position_ids", "token_type_ids", "decoder_position_ids"]:
+                generation_args.pop(model_input_name, None)
+        else:
```
zucchini-nlp: this is needed for specific models which prepare inputs differently depending on `first` vs `subsequent` iterations. For ex in multimodal models, we pass over multimodal data only in first iteration and then rely on cached inputs

Assisted generation however calls internally `generate()` many times and technically will trigger many times `first_iiteration`. This way we can call `prefill` only once per assistant model

src/transformers/generation/utils.py
```diff
@@ -3351,9 +3359,15 @@ def _beam_search(
...
         flat_running_sequences = input_ids
-        model_outputs = self._prefill(input_ids, generation_config, model_kwargs)
+        # Assisted generation completes the prefill stage in candidate generator so that
+        # we don't have several `prefill` calls in one generation loop. Skip `_prefill` for assistants
+        if not generation_config.is_assistant:
+            model_outputs = self._prefill(input_ids, generation_config, model_kwargs)
+            prefill_consumed = False
+        else:
+            model_kwargs = self._get_initial_cache_position(input_ids.shape[1], input_ids.device, model_kwargs)
+            prefill_consumed = True
```
zucchini-nlp, line 3332: same as above - since we already called prefill on assistant, we should not call it a second time

src/transformers/generation/candidate_generator.py
```diff
@@ -215,8 +219,7 @@ def get_candidates(self, input_ids: torch.LongTensor) -> tuple[torch.LongTensor,
             return input_ids, None
         # Update past key values and masks
         self._update_past_and_masks(input_ids)
-        # Generate candidates
```
vasqu, line 223: Let's keep the comment, no?

src/transformers/generation/candidate_generator.py
```diff
@@ -304,19 +307,39 @@ def _update_past_and_masks(
...
-            "min_new_tokens": min_new_tokens,
-            "max_new_tokens": max_new_tokens,
-            "generation_config": self.generation_config,
-            "logits_processor": self.logits_processor,
-        }
+        # Generate candidates. Run prefill-specific logic in first generation and prepare model kwargs.
+        # NOTE: `prepare_inputs_for_generation` creates inputs that can't be used when continuing generation with past-cache
+        # therefore we manually re-assign full input ids and other args. It is a known issue, due to legacy reasons we
+        # have to pass whole input ids to `generate()` including past tokens which are in encoded in cache
+        if is_first_iteration is None:
```
vasqu: Very nitpicky but then the type of `is_first_iteration` is `Optional[bool]/bool | None`

So this only happens with assistant models: Should we add another check that assistant model is not None or similar?

src/transformers/generation/candidate_generator.py
vasqu: Can we also add this to the comments. This is a nice to know. Possibly into the docstring directly, I think the scope is worth enough to cover properly

src/transformers/generation/candidate_generator.py
```diff
@@ -304,19 +307,39 @@ def _update_past_and_masks(
...
+                "generation_config": self.generation_config,
+                "logits_processor": self.logits_processor,
+            }
+        )
+        return generation_args
 
     def _generate_candidates(self, generation_args: dict) -> tuple[torch.LongTensor, Optional[torch.FloatTensor]]:
         """Generate candidate sequences using the assistant model."""
-        assistant_output = self.assistant_model.generate(**generation_args, **self.assistant_kwargs)
+        assistant_output = self.assistant_model.generate(**self.assistant_kwargs, **generation_args)
```
vasqu: Oh lol the order has changed, was confused for a second but it was not changing anything?

src/transformers/generation/candidate_generator.py
vasqu: Seeing this is explained in utils directly, maybe just my order of reviewing was just bad then... Can keep it this way

src/transformers/models/aria/modeling_aria.py
```diff
@@ -1209,12 +1210,15 @@ def prepare_inputs_for_generation(
...
         )
 
-        if cache_position[0] == 0:
-            # If we're in cached decoding stage, pixel values should be None because input ids do not contain special image token anymore
-            # Otherwise we need pixel values to be passed to model
+        if is_first_iteration or not kwargs.get("use_cache", True):
+            # Pixel values are used only in the first iteration if available
+            # In subsquent iterations, they are already merged with text and cached
+            # NOTE: first iteration doesn't have to be prefill, it can be the first
+            # iteration with a question and cached system prompt (continue generate from cache)
```
vasqu, line 1224: Not directly on you, but at some point it would be nice if could integrate this logic into the main generation loop without overrides. We have to repeat ourselves quite often

src/transformers/models/bamba/modeling_bamba.py
```diff
@@ -1519,7 +1520,7 @@ def prepare_inputs_for_generation(
                 position_ids = position_ids[:, -input_ids.shape[1] :]
 
         # if `inputs_embeds` are passed, we only want to use them in the 1st generation step
-        if inputs_embeds is not None and empty_past_kv:
+        if inputs_embeds is not None and is_first_iteration:
```
vasqu, line 1524: Shoot, we can't use super here because we miss it being considered as class? 

Out of scope for this PR: But wouldn't we able to do this with super if we initialize the cache in the forward (if appropriate)

src/transformers/models/gemma3/modeling_gemma3.py
```diff
@@ -825,8 +826,12 @@ def create_causal_mask_mapping(
     # NOTE: this `may_have_image_input` logic is not flawless, it fails when we're using a cache eagerly initialized
     # (e.g. compiled prefill) AND `pixel_values` are not provided (i.e. the image data is provided through other
     # means). Determining prefill in that case requires checking data values, which is not compile-compatible.
-    may_have_image_input = past_key_values is None or not past_key_values.is_initialized or pixel_values is not None
-    if token_type_ids is not None and may_have_image_input:
+    is_first_iteration = (
+        is_first_iteration
+        if is_first_iteration is not None
+        else (past_key_values is None or not past_key_values.is_initialized or pixel_values is not None)
+    )
```
vasqu, line 795: Do we still need this additional logic? Would also update the comment above if we keep this

src/transformers/models/git/modeling_git.py
vasqu, line 1: What happened here :D let's be careful here and double check with run-slow!

src/transformers/models/mamba/modeling_mamba.py
```diff
@@ -761,6 +761,7 @@ def prepare_inputs_for_generation(
         cache_params: Optional[MambaCache] = None,
         cache_position: Optional[torch.LongTensor] = None,
         attention_mask: Optional[torch.LongTensor] = None,
+        is_first_iteration: Optional[bool] = False,
```
vasqu, line 753: Missing?

src/transformers/models/mamba2/modeling_mamba2.py
```diff
@@ -955,6 +955,7 @@ def prepare_inputs_for_generation(
         cache_params: Optional[Mamba2Cache] = None,
         cache_position: Optional[torch.LongTensor] = None,
         attention_mask: Optional[torch.Tensor] = None,
+        is_first_iteration: Optional[bool] = False,
```
vasqu, line 958: Also here missing logic

src/transformers/models/moshi/modeling_moshi.py
vasqu, line 1: Would add this to the list to be careful about with run slow

src/transformers/models/paligemma/modeling_paligemma.py
```diff
@@ -169,31 +170,33 @@ def create_causal_mask_mapping(
...
-    # NOTE: this `is_prompt` logic is not flawless, it fails when we're using a cache eagerly initialized
-    # (e.g. compiled prefill) AND `pixel_values` are not provided (i.e. the image data is provided through other
-    # means). Determining prefill in that case requires checking data values, which is not compile-compatible.
-    maybe_is_prompt = past_key_values is None or not past_key_values.is_initialized or pixel_values is not None
-
-    if maybe_is_prompt:
+    # Infer if prefill or decoding stage, if the flag isn't passed. This happens only when the mask is constructed
+    # from `forward` call. If users run a `forward` call, we have no option to infer `is_first_iteration` because users may be
+    # running generation with custom loop. Thus we need to infer it in a `non-perfect` way
+    # NOTE: Determining prefill in that case requires checking data values, which is not compile-compatible.
```
vasqu, line 176: Ah this is the comment I was missing in gemma3 I think

src/transformers/models/qwen2_5_vl/modular_qwen2_5_vl.py
```diff
@@ -803,7 +805,7 @@ def prepare_inputs_for_generation(
             # When compiling, we can't check tensor values thus we check only input length
             # It is safe to assume that `length!=1` means we're in pre-fill because compiled
             # models currently cannot do assisted decoding
-            if cache_position[0] == 0 or self.model.rope_deltas is None:
+            if (cache_position[0] == 0 or not use_cache) or self.model.rope_deltas is None:
```
vasqu, line 809: Can we not simplify here? Same for the other related models

src/transformers/models/reformer/modeling_reformer.py
```diff
@@ -2238,7 +2238,7 @@ def forward(
         )
 
     def prepare_inputs_for_generation(
-        self, input_ids, past_key_values=None, use_cache=None, num_hashes=None, **kwargs
+        self, input_ids, past_key_values=None, use_cache=None, num_hashes=None, is_first_iteration=False, **kwargs
```
vasqu, line 2250: Do we need this? Can be removed and kwargs just implicitly takes it. Seeing this in rembert too

src/transformers/models/xlm/modeling_xlm.py
```diff
@@ -946,7 +946,7 @@ def get_output_embeddings(self):
     def set_output_embeddings(self, new_embeddings):
         self.pred_layer.proj = new_embeddings
 
-    def prepare_inputs_for_generation(self, input_ids, **kwargs):
+    def prepare_inputs_for_generation(self, input_ids, is_first_iteration=False, **kwargs):
```
vasqu, line 948: Also here

src/transformers/models/xlnet/modeling_xlnet.py
```diff
@@ -1244,7 +1244,9 @@ def get_output_embeddings(self):
     def set_output_embeddings(self, new_embeddings):
         self.lm_loss = new_embeddings
 
-    def prepare_inputs_for_generation(self, input_ids, past_key_values=None, use_mems=None, **kwargs):
+    def prepare_inputs_for_generation(
+        self, input_ids, past_key_values=None, use_mems=None, is_first_iteration=False, **kwargs
+    ):
```
vasqu, line 1249: Annoying but I tag them before they get lost, also here

tests/models/git/test_modeling_git.py
```diff
@@ -496,7 +524,7 @@ def test_inference_image_captioning(self):
         self.assertEqual(outputs.sequences.shape, expected_shape)
         self.assertEqual(generated_caption, "two cats laying on a pink blanket")
         self.assertTrue(outputs.scores[-1].shape, expected_shape)
-        expected_slice = torch.tensor([-0.8805, -0.8803, -0.8799], device=torch_device)
+        expected_slice = torch.tensor([-0.8433, -0.8432, -0.8429], device=torch_device)
```
vasqu: Was this failing before?

vasqu (commented):
LGTM overall, the logic is good

I mainly left some smaller comments and things that might've been missed. We should be a bit careful here and run-slow on a few models, e.g. gemma3, mamba2, etc

src/transformers/models/bamba/modeling_bamba.py
zucchini-nlp: ideally we should init cache in `def prepare_cache` (i dont remember exact name) utility in generation. But rn it doesn't work well with model-specific caches like Mamba
In the future we should move it to `GeneationMixin` and to `forward`, similar to most transformer-LMs

src/transformers/models/gemma3/modeling_gemma3.py
zucchini-nlp: keeping it for edge cases if users call Gemma3 `forward` directly without `generate()`, so it's mostly for power users. In common cases we don't need it anymore

src/transformers/models/git/modeling_git.py
zucchini-nlp: Hahaha, a small refactor. GIT is an old VLM and doesn't follow our standards. Had to change a few things  to standardize its cache layout
Will defi trigger slow tests

src/transformers/models/mamba/modeling_mamba.py
zucchini-nlp: nope, the preparation logic doesn't depend on first iteration currently. Mamba-like models have a lot of things simplified and not standardized, so it should be a separate PR to clean them up

src/transformers/models/bamba/modeling_bamba.py
zucchini-nlp: Here the changes have to be reverted, cache can't be init in `first_iteration` blindly because user might provide their own cache

src/transformers/models/qwen2_5_vl/modular_qwen2_5_vl.py
zucchini-nlp: i wanted to use `is_first_iteration` at first for all models, and realized the concept of `prefill` and `first iteration` can be different
Specifically in mRoPE, the deltas are computed once with the first prompt and should not be computed again if user wants to re-use cache and continue generation from where it's left

This caused me so much headache tbh 🙃

src/transformers/models/reformer/modeling_reformer.py
zucchini-nlp: added so that `is_first_iteration` is not consumed by model's forward, because later we pass all kwargs to the model. Same for most other model

I think BERT-like models need to get rid of overriden `prepare_inputs_for_generation` and rely on `super()`, but it's out of scope for this PR

tests/models/git/test_modeling_git.py
zucchini-nlp: nope, my small refactor changed the logits slightly due to difference in how caching is done. I will trigger slow tests and check how big of a diff I caused

src/transformers/generation/candidate_generator.py
zucchini-nlp: my bad, dunno where the `is None` check came from. It can't be `None` and we will always get one of True/False values

src/transformers/generation/candidate_generator.py
zucchini-nlp: nope, prob changed after several rounds of updates

src/transformers/models/aria/modeling_aria.py
zucchini-nlp: yeah, that is the next step. Most VLMs are quite standard now and we can generalize without exceptions

zucchini-nlp:
run-slow: git

tests/models/git/test_modeling_git.py
zucchini-nlp: Oke, i see why there is a difference. Prev we would compute attention for the image embeddings (around 900 tokens) always, even though it is cached. After this PR, caching works as expected and we have one single token at each decoding step

IMO the current version is more correct and it's expected that caching results in tiny numerical differences that can add up

zucchini-nlp:
@vasqu requesting one last review. One tiny thing left to do is to make sure slow GIT tests pass, some expected values are hardware-dependent

Otherwise should be ready, addressed a few comments and answered q above

zucchini-nlp:
run-slow: git

src/transformers/generation/candidate_generator.py
```diff
@@ -304,19 +308,46 @@ def _update_past_and_masks(
...
-            self.input_ids_key: input_ids,
-            "min_new_tokens": min_new_tokens,
-            "max_new_tokens": max_new_tokens,
-            "generation_config": self.generation_config,
-            "logits_processor": self.logits_processor,
-        }
+        # Generate candidates. Run prefill-specific logic in first generation and prepare model kwargs.
+        # Some models prepare inputs differently depending on first vs subsequent iterations.(e.g. VLMs)
+        # Assisted generation however calls internally `self.generate()` many times and technically will
+        # lead to many `ifirst_iteration's`. This way we can call prefill only once per assistant model
```
```suggestion
        # lead to many `first_iteration's`. This way we can call prefill only once per assistant model
```
typo

src/transformers/generation/candidate_generator.py
```diff
@@ -304,19 +308,46 @@ def _update_past_and_masks(
...
+        if is_first_iteration:
+            generation_args = self.assistant_model._get_initial_cache_position(
+                input_ids.shape[1], input_ids.device, self.assistant_kwargs.copy()
```
vasqu, line 326: Is there a specific reason we copy the kwargs here? Any risk this could be `None`?

I suspect some inplace ops but just checking

src/transformers/generation/candidate_generator.py
```diff
@@ -304,19 +308,46 @@ def _update_past_and_masks(
...
+
+        generation_args.update(
+            {k: self.assistant_kwargs[k] for k in self.assistant_kwargs if k not in generation_args}
+        )
+        return generation_args
 
     def _generate_candidates(self, generation_args: dict) -> tuple[torch.LongTensor, torch.FloatTensor | None]:
         """Generate candidate sequences using the assistant model."""
-        assistant_output = self.assistant_model.generate(**generation_args, **self.assistant_kwargs)
+        assistant_output = self.assistant_model.generate(**generation_args)
```
vasqu, line 355: So we now directly write into the generation args instead (from the prep)

src/transformers/generation/utils.py
```diff
@@ -593,6 +593,7 @@ def prepare_inputs_for_generation(
         attention_mask: torch.LongTensor | None = None,
         inputs_embeds: torch.FloatTensor | None = None,
         cache_position: torch.LongTensor | None = None,
+        is_first_iteration: Optional[bool] = False,
```
vasqu:
```suggestion
        is_first_iteration: bool | None = False,
```
also often forgetting this, but we really should move to the new typing

src/transformers/models/csm/generation_csm.py
vasqu, line 1: Important model, so just double check with run-slow or similar

src/transformers/models/gemma3/modeling_gemma3.py
```diff
@@ -1166,7 +1176,7 @@ def create_masks_for_generate(
             past_key_values,
             position_ids,
             token_type_ids,
-            pixel_values=kwargs.get("pixel_values"),
+            is_first_iteration,
```
vasqu: Should we not keep pixel values here? I just fear that the args order won't match anymore? So `is_first_iteration` would be used as pixel values

src/transformers/models/qwen2_5_vl/modular_qwen2_5_vl.py
vasqu: Makes sense, mRoPE strikes again :cry: cant wait when we have a standardized way here outside of modeling.py

src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
```diff
@@ -848,11 +848,11 @@ def prepare_inputs_for_generation(self, *args, **kwargs):
         # Overwritten -- we should not pass input_features when we are in cached decoding stage
 
         input_features = kwargs.pop("input_features", None)
-        cache_position = kwargs.get("cache_position")
+        is_first_iteration = kwargs.get("is_first_iteration", False)
 
         model_inputs = super().prepare_inputs_for_generation(*args, **kwargs)
 
-        if cache_position is not None and cache_position[0] == 0:
+        if is_first_iteration or not kwargs.get("use_cache", True):
```
vasqu, line 855: Is this `or` intentional? Not sure if this happened elsewhere

tests/generation/test_utils.py
```diff
@@ -2890,6 +2899,19 @@ def test_transition_scores_group_beam_search_encoder_decoder(self):
 
         torch.testing.assert_close(transition_scores_sum, outputs.sequences_scores, rtol=1e-3, atol=1e-3)
 
+    @slow
+    def test_generate_inputs_embeds_one_token(self):
```
vasqu, line 2928: Nice :pray:

vasqu (approved):
LGTM overall, left smaller comments/nits but I think this looks pretty ready

Trusting you on fixing up git and other CI stuff :eye: lots of potential for followups PRs to clean more but this is already big enough as is and solves the biggest issue(s)

zucchini-nlp:
run_slow: git, csm

src/transformers/generation/candidate_generator.py
zucchini-nlp: because `_get_initial_cache_position` updates input kwargs in-place which we do not want. `assistant_kwargs` here is an instance-lvl attribute and it messes up the whole thing with in-place updates to it

It cannot be None, at the very least can be an empty dict

src/transformers/generation/candidate_generator.py
zucchini-nlp: yep, the kwargs should not overwrite anything in `generation_args` in the first iteration. We could swap the order, but I think moving everything to prep was more explicit

src/transformers/models/gemma3/modeling_gemma3.py
zucchini-nlp: oh right, will pass it as keyword arg. Pixel values aren't needed, and will be a no-op even if we pass it. If there's a `is_first_iteration`, it takes precedence over implicit inferring via pixels

src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
zucchini-nlp: yeah, we don't want to pop pixels/audios when generating without cache. TBH we didn't check it earlier and noone complained, which means noone generates without cache 😄  Added it only for completeness

zucchini-nlp:
Will merge when tokenization tests are fixed on main

vasqu:
No worries, kinda been this way the past few days :cry:

src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
vasqu: If noone complains, that's a good sign :D

zucchini-nlp:
@bot /repo

PR #42092, opened 2025-11-07
Fix the issue Arthur was talking about.

Vaibhavs10:


ArthurZucker (commented):
Cool, make sure you test before vs after and you show that before it would indeed overwrite the weights for a linear layer for example, if one of the keys was not loaded. So make sure you repro and show that without the fix it will fail.

Vaibhavs10 (commented):
I tracked down the regression you pointed out. On upstream main, the linear head weight really does get re-randomized when only the bias is missing:
```
  git worktree add /tmp/transformers-upstream upstream/main
  PYTHONPATH=/tmp/transformers-upstream/src python - <<'PY'
  import torch
  from transformers import BertConfig, BertForSequenceClassification

  config = BertConfig(hidden_size=8, intermediate_size=16, num_attention_heads=2,
  num_hidden_layers=1, num_labels=2)
  model = BertForSequenceClassification(config)
  with torch.no_grad():
      sentinel = torch.arange(model.classifier.weight.numel(),
  dtype=torch.float32).view_as(model.classifier.weight)
      model.classifier.weight.copy_(sentinel)
      model.classifier.bias.zero_()

  model.classifier.weight._is_hf_initialized = True
  if hasattr(model.classifier.bias, "_is_hf_initialized"):
      delattr(model.classifier.bias, "_is_hf_initialized")
  model.classifier._is_hf_initialized = False

  model._initialize_missing_keys(['classifier.bias'], is_quantized=False)

  print('old diff', torch.max(torch.abs(model.classifier.weight - sentinel)).item())
  PY

  Output:

  old diff 15.002209663391113

  With the patch in this branch:

  python - <<'PY'
  import torch
  from transformers import BertConfig, BertForSequenceClassification

  config = BertConfig(hidden_size=8, intermediate_size=16, num_attention_heads=2,
  num_hidden_layers=1, num_labels=2)
  model = BertForSequenceClassification(config)
  with torch.no_grad():
      sentinel = torch.arange(model.classifier.weight.numel(),
  dtype=torch.float32).view_as(model.classifier.weight)
      model.classifier.weight.copy_(sentinel)
      model.classifier.bias.zero_()

  model.classifier.weight._is_hf_initialized = True
  if hasattr(model.classifier.bias, "_is_hf_initialized"):
      delattr(model.classifier.bias, "_is_hf_initialized")
  model.classifier._is_hf_initialized = False

  model._initialize_missing_keys(['classifier.bias'], is_quantized=False)

  print('new diff', torch.max(torch.abs(model.classifier.weight - sentinel)).item())
  PY

  Output:

  new diff 0.0
```

  To keep this covered going forward I also added a regression test
  (test_missing_linear_bias_preserves_weight_custom_model) that reproduces the exact flag
  state the loader sets; it fails on main and passes here.

Cyrilvallez:
Note sure what were the discussions, but just wanna pin-point that this is known on `main` and is not an issue as param loading happens after the call to `_initialize_missing_keys`, so even if we indeed overwrite some params (because we can only init full modules), they are then correctly loaded afterwards

TLDR `_initialize_missing_keys` is never called alone, always BEFORE the actual loading

PR #42094, opened 2025-11-07
Add prefix sharing to continuous batching

remi-or:
This PR adds a prefix sharing mechanism to the continuous batching API like the one present in VLLM.
It only activates if the model is a full-attention model, as is the case in VLLM.

The mechanism has two main components:
- block hashing: each block in the cache, once it is filled up, is given a hash that depends on all the tokens in the sequence up to and including the ones in the block
- prefix detection: when starting prefill for a request, we first look for a prefix with KV cache already computed, and if such a prefix is found, we skip the KV computation for it, using references to completed blocks to save compute
- block de-reference: when a block is given a hash, we check that no other block shares the same hash. This ensures that each block sharing the same information is unique, and helps keep the cache size in control

What is missing from this PR: 
- [x] more documentations
- [x] checking the code again
- [x] edge case: if the prefix is the entire initial request, we still need to do a forward with the last token of the request
- [x] checks TODOs to adress: the ones left are for a PR down the line

The PR is draft until these are resolved but any early comment is welcome


src/transformers/generation/continuous_batching/cache.py
```diff
@@ -227,13 +227,18 @@ def __init__(
                 raise ValueError(f"Invalid group type: {group_type}")
             self.group_cache_managers.append(cm)
 
+        # We only use prefix sharing if the whole model has only full attention layers
```
McPatate, line 234: Is that a "for the moment" thing?

src/transformers/generation/continuous_batching/requests.py
```diff
@@ -116,10 +116,10 @@ class RequestState:
         error (Optional[str]): Any error message associated with the request. When None, has had no error yet.
     """
 
-    # Required fields
+    # Required fields # TODO: come up with better names / not sure prompt_ids and such are not redundant
```
McPatate, line 119: wdym?

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -201,22 +366,3 @@ def get_seqlens_k(self, request_id: str, past_length: int, query_length: int) ->
         """Returns the attention type of the cache allocator and the key sequence length for the given request_id."""
         seqlens_k = query_length + min(past_length, self.sliding_window - 1)
         return "sliding_attention", seqlens_k
-
-
-# TODO: test the impact of this
```
McPatate, line 206: inconclusive results?

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -937,20 +942,6 @@ def request_id_iter(self, request_id: str) -> Generator[GenerationOutput]:
                 request_cancelled = self.batch_processor.scheduler.request_is_cancelled(request_id)
 
     @traced
-    def warmup(self, batch_processor: ContinuousBatchProcessor) -> None:
```
McPatate, line 942: why are we deleting this?

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -561,27 +555,39 @@ def _sync(self) -> list[int]:
     @traced
     def _maybe_send_output(self, state: RequestState) -> None:
         """Send output to the queue based on streaming mode and request state."""
-        if state.streaming:
-            self.output_queue.put(state.to_generation_output())
-        elif state.status == RequestStatus.FINISHED:
+        if state.streaming or state.status == RequestStatus.FINISHED:
             self.output_queue.put(state.to_generation_output())
```
McPatate, line 560: In that case, wdyt of this:
```suggestion
        send_output = state.streaming or state.status == RequestStatus.FINISHED
        if not send_output:
            return
        self.output_queue.put(state.to_generation_output())
```
imo reads better

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -561,27 +555,39 @@ def _sync(self) -> list[int]:
...
 
     @traced
     def update_batch(self) -> None:
         """Update request states based on generated tokens."""
         out_tokens = self._sync()
         for i, state in enumerate(self.requests_in_batch):
+
+            # If the request has no remaining prompt ids, it means prefill has already or just finished
```
McPatate: has already..?

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -561,27 +555,39 @@ def _sync(self) -> list[int]:
...
+                self.cache.mark_blocks_as_completed(state)
+                if is_finished:
                     self.metrics.record_request_completion(state.created_time, state.request_id)
                     self.scheduler.finish_request(state.request_id, evict_from_cache=(not self.manual_eviction))
                 self._maybe_send_output(state)
+            #  Otherwise, the request is still prefilling, but the prefill has been split
             elif state.status == RequestStatus.PREFILLING_SPLIT:
+                self.cache.mark_blocks_as_completed(state)
                 state.status = RequestStatus.SPLIT_PENDING_REMAINDER
+            # DEBUG: there is a dangling if, but idk if it ever happens. Adding an error to catch it.
```
McPatate:
```suggestion
```

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -561,27 +555,39 @@ def _sync(self) -> list[int]:
...
+            else:
+                raise ValueError(f"Request {state.request_id} is in an unexpected state: {state.status}")
+
+
+        # We error out if the cache is full
```
McPatate:
```suggestion
```

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -335,6 +339,38 @@ def update(
...
+                break
+        # If we found a matching prefix, we reference the blocks in the request
+        if allocated_blocks:
+            logger.debug(f"Found prefix match for request {request_id} with {len(allocated_blocks)} blocks")
+            cm = self.group_cache_managers[0]
+            cm.block_table[request_id] = allocated_blocks
+        return len(allocated_blocks) * self.block_size
+
+    def mark_blocks_as_completed(self, state: RequestState) -> None:
+        """Marks the blocks that have been computed in the forward pass as such. If prefix sharing is off, this is a
```
McPatate:
```suggestion
        """Marks the blocks that have been computed in the forward pass as completed. If prefix sharing is off, this is a
```

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -335,6 +339,38 @@ def update(
...
+        no-op."""
+        num_completed_blocks = 0 if not self.use_prefix_sharing else self.blocks_to_complete.pop(state.request_id)
+        if num_completed_blocks == 0:
+            return None
+        cm = self.group_cache_managers[0]  # if prefix sharing is on, there is only one group
+        self._block_manager.mark_blocks_as_computed(
```
McPatate: why is this called in `completed`? or why is the fn called `mark_blocks_as_computed`?

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -14,29 +14,188 @@
...
 
 from .requests import logger
 
 
+T = TypeVar('T')
+def reverse_enumerate(xs: list[T]) -> Iterator[tuple[int, T]]:
+    index = len(xs) - 1
+    for x in xs[::-1]:
+        yield index, x
+        index -= 1
```
McPatate, line 31:
```suggestion
    for idx, x in enumerate(xs[::-1]):
        yield len(xs) - idx - 1, x
```
🤏🏻

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -14,29 +14,188 @@
...
+        """Marks a list of blocks as free. If there is no prefix sharing, we simply add them to the uninitialized blocks
+        queue. Otherwise, we mark them as initalized but they can be freed in no uninitialized blocks are lefts."""
+        if self._use_prefix_sharing:
+            for block_id in blocks:
+                self.decrease_ref_count(block_id)
+        else:
+            self._uninit_block_ids.extend(blocks)
+
+
+    def mark_blocks_as_computed(
```
McPatate:
```suggestion
    def mark_blocks_as_completed(
```

src/transformers/generation/continuous_batching/cache.py
remi-or: No, it is not compatible w/ sliding window (VLLM agrees)

src/transformers/generation/continuous_batching/cache.py
remi-or: I changed terminology halfway into the PR, I need to proof read this. Final terminology is "complete" which imo is clearer

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: will test the impact of this!

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: will rename using f2, thx

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: yes, not much changes

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: If you want the `send_output` bool why not, but might as well make it `if send_output`

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: ended* will add

src/transformers/generation/continuous_batching/requests.py
remi-or: I had a hard time figuring out which was which when reading through the code. Maybe for a later PR hough, wdyt?

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: never called

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: agreed, removing

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: It is slightly slower :/

src/transformers/generation/continuous_batching/cache.py
McPatate: I understand but why not have only on full attention layers?

src/transformers/generation/continuous_batching/cache.py
McPatate: agree!

src/transformers/generation/continuous_batching/requests.py
McPatate: sure!

src/transformers/generation/continuous_batching/continuous_api.py
McPatate: _a      e      s      t      h      e      t      i      c_

remi-or:
Maybe I will proof read again, but I am removing draft for now. Will address review soon.

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: **n o w a y**

remi-or: will ask for outside opinion

src/transformers/generation/continuous_batching/cache.py
remi-or: For now, there is only sliding window or full attention. The only other type of attention I know is in transformers is block attention

remi-or:
Removed draft, ready to merge IMO

src/transformers/generation/continuous_batching/continuous_api.py
McPatate: it's just a matter of readability, by reducing indentation, you'll get it when you're older 👴🏻

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -235,6 +235,7 @@ def __init__(
         self.use_prefix_sharing = allow_prefix_sharing and group_types == ["full_attention"]
         self._block_manager = BlockManager(num_blocks, self.block_size, self.use_prefix_sharing)
         self.blocks_to_complete: dict[str, int] = {}
+        self._total_prefix_length: int = 0  # a counter to measure the impact of prefix sharing, also used in tests
```
McPatate, line 238: super geil, we'll want to monitor this, not sure if relevant in benchmarks but definitely in prod workloads

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -1058,6 +1058,7 @@ def init_continuous_batching(
         max_queue_size: int = 0,
         num_q_cuda_graphs: int = 0,
         num_kv_cuda_graphs: int = 0,
+        allow_prefix_sharing: bool = True,
```
McPatate, line 1081: I feel `enable_prefix_sharing` is better semantically

src/transformers/generation/continuous_batching/scheduler.py
```diff
@@ -140,6 +140,7 @@ def _prepare_request_for_processing(
                 prefill_length = min(prefill_length, len(state.prompt_ids) - 1)
                 state.remaining_prompt_ids = state.prompt_ids[prefill_length:]
                 state.prompt_ids = state.prompt_ids[prefill_length:]
+                state.position_offset += prefill_length
```
McPatate, line 143: ty boss

tests/generation/test_continuous_batching.py
```diff
@@ -456,6 +456,68 @@ def test_streaming_and_non_streaming_requests_can_alternate(self) -> None:
 
         manager.stop(block=True)
 
+    @require_torch_accelerator
+    def test_prefix_sharing(self) -> None:
```
McPatate, line 460: @LysandreJik will be proud

tests/generation/test_continuous_batching.py
```diff
@@ -456,6 +456,68 @@ def test_streaming_and_non_streaming_requests_can_alternate(self) -> None:
...
+        model_id = "Qwen/Qwen2.5-0.5B-Instruct"
+        max_new_tokens = 32
+
+        tokenizer = AutoTokenizer.from_pretrained(model_id)
+        model = AutoModelForCausalLM.from_pretrained(model_id, device_map="auto")
+
+        generation_config = GenerationConfig(do_sample=False, block_size=32)
+        manager = model.init_continuous_batching(generation_config=generation_config)
+        manager.logit_processor = LogitsProcessorList()
+        manager.start()
```
McPatate: I think adding the `with` syntax would help for writing tests and avoid having dangling threads since you handle errors in the `__exit__` block directly

But you can also just throw everything in a `try:` `finally:` block

src/transformers/generation/continuous_batching/cache_manager.py
McPatate: yeah I figured it'd be slower, I was trolling mostly trolling writing this 😄

I wonder if it's the `enumerate` or the `len` call that makes things slower tho

McPatate: > It is slightly slower :/

but did you benchmark with a 1000 runs? :p

src/transformers/generation/continuous_batching/cache.py
McPatate: sorry my question wasn't clear, in models that have a mix of sliding and full attn, why not enable prefix caching?

tests/generation/test_continuous_batching.py
remi-or: the test even caught a bug :exploding_head:

src/transformers/generation/continuous_batching/cache.py
remi-or: Agreed. Added a log into message to retrieve the data.

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: I thought about `enable` but we can never ensure prefix sharing is enabled, because the model might not support it. It could give a false sense of it being always enabled to the user. So I think the `allow` keyword is clearer for the API.

tests/generation/test_continuous_batching.py
remi-or: Added a context manager as discussed offline

remi-or:
Benchmarks, on H100, with the last version of the example script that adds the `--add-prefix` option and sampling off by default
Command: `python examples/pytorch/continuous_batching.py --attn $attn -mp none --add-prefix --samples 500`
| Attention | Version  | Generated tokens | Duration (s) | Throughput (tok/s) | Total prefill length matched |
|-------------|------------|--------------------------|------------------|--------------------------|-------------------------------------|
| SDPA | With prefix sharing | 112033 | 90.47 | 1238.37 | 16096 |
| SDPA | No prefix sharing | 112223 | 89.95  | 1247.55 | - |
| SDPA | Main branch | 112223 | 90.31 | 1242.60 | - |
| Flash attention | With prefix sharing | 111494 | 24.76  | 4503.30 | 16096 |
| Flash attention | No prefix sharing | 112599 | 26.87 | 4190.22  | - |
| Flash attention | Main branch | 112599 | 27.10 | 4155.44 | - |

The throughput gain increases when the prefix length increases. For reference, the prefixes in the table above are of length 0, 40, 60 or 80 tokens (roughly). If instead we add a large system prompt, so all requests share a 2500 tokens prefix, the gap is more noticeable. Here are the number with flash attention:

```
No prefix sharing:    195.49 seconds for 122435 tokens. 626.29tok/s
W/ prefix sharing:    129.37 seconds for 121763 tokens. 941.17tok/s
```

src/transformers/generation/continuous_batching/continuous_api.py
McPatate: gotcha makes sense

src/transformers/generation/continuous_batching/cache.py
remi-or: Because the layers with a sliding window overwrite their KV cache when they reach the end of their sliding window, so we have to disable prefix caching for those. And if we disable prefix caching for one layer of the model, we have to disable it for all layers of the model: we need a full foward pass to build the KV cache for all layers, we cannot only do a forward pass for the layers that have no prefix caching.

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -1059,13 +1062,23 @@ def evict_request_from_cache(self, request_id: str) -> None:
...
     """Mixin class for models to add continuous batching capabilities."""
 
+    @contextmanager
+    def continuous_batching_context_manager(self, **kwargs) -> Generator[ContinuousBatchingManager]:
+        manager = self.init_continuous_batching(**kwargs)
+        manager.start()
+        try:
+            yield manager
+        finally:
+            manager.stop(block=True)
```
McPatate, line 1072: oh wow error is propagated given you don't have an `except` block, neat

McPatate (approved):

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: neato indeed

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -14,29 +14,211 @@
...
+class Block:
+    """A class to represent a block managed by the block manager. We say that a block is complete when the physical KV
+    cache it points to is fully computed. A block can have a parent, which is the block that came before in the
+    sequence. Once a block is complete, it is given a hash, which takes into account the tokens ids of the block and
+    its parent's hash (if there is a parent)."""
+
+    def __init__(self, id_: int, parent_id: int | None) -> None:
+        self.id: int = id_
+        self.parent_id: int | None = parent_id
+        self.hash: int | None = None
```
ArthurZucker, line 43: you can also use `__hash__` native property of pythong

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -14,29 +14,211 @@
...
+        self._use_prefix_sharing = use_prefix_sharing
+        self._hash_to_id: dict[int, int] = {}
+        self._id_to_block: dict[int, Block] = {}
+
+    @property
+    def num_free_blocks(self) -> int:
+        """Returns the number of free blocks left. Both initialized and uninitialized blocks are considered free."""
+        return len(self._uninit_block_ids) + len(self._init_block_ids)
+
+    def is_enough_free_blocks(self, n_blocks: int) -> bool:
```
```suggestion
    def has_enough_free_blocks(self, n_blocks: int) -> bool:
```

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -14,29 +14,211 @@
...
+                self.free_blocks([block.id])
+
+            # Otherwise, we add the completed block to the hash table
+            else:
+                self._hash_to_id[block.hash] = block.id
+
+            # Update loop variables
+            parent_hash = block.hash
+
+    def compute_hash(self, parent_hash: int | None, tokens: list[int]) -> int:
```
ArthurZucker, line 199:
```suggestion
    def __hash__(self, parent_hash: int | None, tokens: list[int]) -> int:
```

ArthurZucker (approved):
as discussed offline, we might want to simplify the stream, when we request the cache for `n` blocks, we know the `n-1` are FULL / Completed and thus can keep the logic to compute hash there. This means it can be "scheduled" as you don't need the result until the next forward (thus can be done in the BG while the model runs a forward)

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: only takes `self` as an argument :( 
https://docs.python.org/3.5/reference/datamodel.html#object.__hash__

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: answered in a later comment

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: will do with f2, it's used in other place. Good catch, thanks!

PR #42095, opened 2025-11-07
Xcodec fix

eustlb:
# What does this PR do?

Before a deeper clean (#42039) on xcodec and to break things down a bit, this PR fixes some inefficient logic in X-codec, responsible for unnecessary memory spikes.

Makes me think that the whole `_get_output_length` when applying conv1d, which is repeated a lot throughout audio models, could benefit from some standardisation. In that objective, I've added a new audio util that I'll propagate in a subsequent PR.


src/transformers/models/xcodec/modeling_xcodec.py
```diff
@@ -394,6 +396,40 @@ def remove_weight_norm(self):
...
+        """
+        if module is None:
+            module = self
+
+        conv1d_layers = self._get_conv1d_layers(module)
+
+        for layer in conv1d_layers:
+            input_length = conv1d_output_length(layer, input_length)
+
+        return input_length
```
Cyrilvallez, line 433: Is this something we can do earlier, in the config? It would avoid having to do these recursions on modules etc!

eustlb: When we'll standardize this computation (see #41203), we could be able to do it from the config, provided we ensure that every convolution layer’s parameters are stored there, and in the correct order relative to their appearance in the forward pass. However, I don’t think expanding the config with such parameters is a good idea, especially since we decided to hardcode them to avoid exposing them to the user.

I was thinking more along the lines of handling this directly during model initialization, when the modules are already being iterated over. For now, is it okay to merge this?

Cyrilvallez (approved):
Alright, yes it's fine to merge it like that to unblock, since I see you're already working on a refined version! 🤗
CI not happy though it seems!

PR #42098, opened 2025-11-07
Fix mel length computation in Qwen2-Audio

anferico:
# What does this PR do?

1. Updates the misleading docstring in `Qwen2AudioEncoder.forward` stating that _"Qwen2Audio does not support masking of the `input_features`"_, while in fact it _does_ make use of it. This is likely a mistake made while copying the docstring over from Whisper
2. Fixes a bug that causes the last "valid" (i.e. not masked) frame of the sequence coming out of the first two convolutional layers to be masked incorrectly when `max_mel_seq_len` is an odd number

Here's a more elaborate explanation for point 2:
`max_seq_len` represents the maximum sequence length after the first 2 convolutional layers, of which only the first one compresses the input sequences (due to `stride=2`). The compression applied by the first convolutional layer can be computed using this expression (cf. [torch.nn.Conv1d](https://docs.pytorch.org/docs/stable/generated/torch.nn.Conv1d.html)): 
```
output_length = (input_length + 2*padding - dilation * (kernel_size - 1) - 1) // stride + 1
```
which, given that:
- `padding` = 1
- `dilation` = 1
- `kernel_size` = 3
- `stride` = 2

can be simplified to:

```
output_length = (input_length - 1) // 2 + 1
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@ylacombe @amyeroberts @faychu

### DIFF
--- a/src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
+++ b/src/transformers/models/qwen2_audio/modeling_qwen2_audio.py
@@ -347,9 +347,15 @@ class Qwen2AudioEncoder(Qwen2AudioPreTrainedModel):
     ):
         r"""
         Args:
-            attention_mask (`torch.Tensor`)`, *optional*):
-                Qwen2Audio does not support masking of the `input_features`, this argument is preserved for compatibility,
-                but it is not used. By default the silence in the input log mel spectrogram are ignored.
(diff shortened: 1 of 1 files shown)


anferico:
@ylacombe @amyeroberts @faychu any news?

Rocketknight1:
cc @eustlb @ebezzam

PR #42099, opened 2025-11-07
Fix TimesFM patch normalization instability

AnMakc:
This PR fixes patch normalization issue in TimesFM model:
- Numerical instability for low-variance data
- Incorrect clamping of calculated sigma to one instead of configured eps

Issue only affects data with low relative variance and results in incorrect prediction.
Variance is now calculated after mean removal to avoid catastrophic cancellation.

@kashif Please, take a look as you contributed this model.

## Bug demo

image

<details>
<summary> Code to reproduce </summary>

```py
import random

import numpy as np
import matplotlib.pyplot as plt
import torch

from transformers import TimesFmModelForPrediction


seed = 42
np.random.seed(seed)
torch.manual_seed(seed)
random.seed(seed)


model = TimesFmModelForPrediction.from_pretrained(
    "google/timesfm-2.0-500m-pytorch",
    attn_implementation="sdpa",
    device_map="auto",
)

model.eval()


X_inp = np.arange(256)
X_pred = np.arange(256, 256 + 128)

fig, axs = plt.subplots(2, 2, figsize=(12, 12))

for ax, sigma in zip(axs.flatten(), [1e-0, 1e-1, 1e-2, 1e-3]):
    rng = np.random.default_rng(seed=seed)
    Y_inp = 1e3 + rng.normal(loc=0, scale=sigma, size=(256,)).cumsum()

    inputs = [
        torch.from_numpy(Y_inp).to(model.device, dtype=model.dtype)
    ]

    with torch.no_grad():
        pred = model(past_values=inputs)
        pred = pred.mean_predictions.cpu().numpy().squeeze()

    ax.plot(X_inp, Y_inp, label='Inputs')
    ax.plot(X_pred, pred, label='Mean Prediction')
    ax.set_title(f'Sigma={sigma}')
    ax.legend()

```
</details>

### DIFF
--- a/src/transformers/models/timesfm/modeling_timesfm.py
+++ b/src/transformers/models/timesfm/modeling_timesfm.py
@@ -339,11 +339,7 @@ class TimesFmModel(TimesFmPreTrainedModel):
     ) -> tuple[torch.Tensor, tuple[torch.Tensor, torch.Tensor]]:
         """Input is of shape [B, N, P]."""
         mu, sigma = self._timesfm_masked_mean_std(inputs, patched_pads)
-        sigma = torch.where(
-            sigma < self.config.tolerance,
-            torch.tensor(1.0, dtype=sigma.dtype, device=sigma.device),
-            sigma,
-        )
+        sigma = torch.clamp(sigma, min=self.config.tolerance)
 
         # Normalize each patch
         outputs = (inputs - mu[:, None, None]) / sigma[:, None, None]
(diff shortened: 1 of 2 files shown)


AnMakc:
@kashif Kind ping :)

kashif:
thanks @AnMakc checking! looks good overall

kashif (approved):
good catch!

PR #42102, opened 2025-11-07
fix(ci): unexpected keyword argument `streaming`

McPatate:
I suspect some sort of error where we're not working with the correct up to date code of transformers in the benchmarks, which definitely is an issue.

EDIT: culprit seems to be:
```yml
working-directory: /transformers
```
and thinking of it indeed, it's not what we just checked out but transformers dir of the docler image

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -36,6 +36,9 @@ jobs:
         with:
           fetch-depth: 1
 
+      - name: debug
+        run: pwd
+
       - name: Install benchmark script dependencies
         run: python3 -m pip install -r benchmark_v2/requirements.txt kernels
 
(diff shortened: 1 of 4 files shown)


ydshieh (approved):
ok, nice catch, thank you!

For me, I would probably use something like

```
      - name: Update clone
        working-directory: /transformers
        env:
          commit_sha: ${{ github.sha }}
        run: |
          git fetch origin "$commit_sha" && git checkout "$commit_sha"
```

It's not a technical reason, but it's what I did for all workflow files where we use our CI docker images. But since you maintain the benchmark workflows, it's up to you  to make the call :-)

McPatate:
> It's not a technical reason, but it's what I did for all workflow files where we use our CI docker images. But since you maintain the benchmark workflows, it's up to you to make the call :-)

I'd rather keep my workflow standard with the actions/checkout step! Will change if it is too confusing

ArthurZucker:
Ty all!

PR #9491, opened 2021-01-08
[trainer] round numbers in trainer state

stas00:
This PR rounds very long fractions in trainer state e.g., 

```
{'loss': 14.846837043762207, 'learning_rate': 6e-06, 'epoch': 0.3333333333333333} 
```
to:
* epoch 2 decimals
* loss 4 decimals

resulting in:

```
{'loss': 14.8468, 'learning_rate': 6e-06, 'epoch': 0.33}
```` 

If you want any other small tweaks for me to add please let me know.

Fixes: #9475 

@sgugger

PR #42104, opened 2025-11-07
Disable loss rounding in training stats log

### LINKED ISSUE #38032: Removing the modification of loss value due to rounding off to 4 digits
### System Info

transformers version: 4.50.2
python version: 3.13.1

### Who can help?

@zach-huggingface @SunMarc 

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

Inside the Trainer class, why is the loss rounded to 4 digits? I have applications where I am interested to see the loss go below 4 significant digits, but they all get rounded to 0. Please let the user set this rounding number, or let the loss be displayed in scientific notation like 1.xxxe^y

this is set inside def _maybe_log_save_evaluate()
logs["loss"] = round(tr_loss_scalar / (self.state.global_step - self._globalstep_last_logged), 4)

### Expected behavior

It would be great if this hardcoding of rounding off the "loss" is removed. Best output would be to remove the round function and log the loss value in scientific notation which is much clearer.

AnMakc:
This PR disables unconditional rounding of the loss value to 4 digits in logged values. Effectively reverting part of https://github.com/huggingface/transformers/pull/9491.

Closes https://github.com/huggingface/transformers/issues/38032 (stale)

Motivation:
- Loss value may have very low absolute magnitude depending on model/data/objective function.
- No other values are rounded (e.g. eval/loss or grad_norm) which may result in confusion.
- Rounding is a display/formatting detail in this case and should not be done silently in non-UI code

### Examples of undesired behavior that would be fixed with this PR:
Notebook outputs: 
Screenshot 2025-11-07 at 16 06 21

Wandb loss chart:
Screenshot 2025-11-07 at 15 42 45

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -3004,7 +3004,7 @@ class Trainer:
             # reset tr_loss to zero
             tr_loss -= tr_loss
 
-            logs["loss"] = round(tr_loss_scalar / (self.state.global_step - self._globalstep_last_logged), 4)
+            logs["loss"] = tr_loss_scalar / (self.state.global_step - self._globalstep_last_logged)
             if grad_norm is not None:
                 logs["grad_norm"] = grad_norm.item() if isinstance(grad_norm, torch.Tensor) else grad_norm
             if learning_rate is not None:
--- a/src/transformers/trainer_callback.py
+++ b/src/transformers/trainer_callback.py
@@ -665,12 +665,12 @@ class ProgressCallback(TrainerCallback):
                         f"[String too long to display, length: {len(v)} > {self.max_str_len}. "
                         "Consider increasing `max_str_len` if needed.]"
                     )
+                if isinstance(v, float):
+                    # Format floats for better readability
+                    shallow_logs[k] = f"{v:.4g}"
                 else:
                     shallow_logs[k] = v
             _ = shallow_logs.pop("total_flos", None)
-            # round numbers so that it looks better in console
-            if "epoch" in shallow_logs:
-                shallow_logs["epoch"] = round(shallow_logs["epoch"], 2)
             self.training_bar.write(str(shallow_logs))
 
     def on_train_end(self, args, state, control, **kwargs):
@@ -687,6 +687,8 @@ class PrinterCallback(TrainerCallback):
     def on_log(self, args, state, control, logs=None, **kwargs):
         _ = logs.pop("total_flos", None)
         if state.is_local_process_zero:
+            if logs is not None:
+                logs = {k: (f"{v:.4g}" if isinstance(v, float) else v) for k, v in logs.items()}
             print(logs)
 
 


Rocketknight1:
cc @sunmarc, but maybe formatting using something like `%.4f` so that it always displays 4 significant figures with an optional exponent would preserve the clean rendering but still work in cases when the loss is very small?

AnMakc:
@Rocketknight1 Any rounding at this point introduces inconsistency between validation and training loss. We could not apply formatting here as we log numeric value, not string.

The main point is that any formatting is a representation detail that should be handled by the display-related code and we should not make assumptions on how it works as in most cases it is some external handler (wandb, trackio, ipython NB widget, etc.).

AnMakc:
@Rocketknight1 @SunMarc

Please, take a look

SunMarc (approved):
I don't mind merging this PR. I also saw a couple of issues before complaining about this. cc @stas00 as this is something you added way back

qgallouedec (approved):

stas00:
This use case didn't exist when this was originally coded. The goal was to avoid a pointless output like `loss=1.3982828288737373`

I wonder if using a scientific notation can address this use-case while not making it an ugly experience for all other use cases. `loss=2.5444e-4` for the example above. Same 4 decimals, except the exponent overcomes the issue of lost precision in the original.

AnMakc:
> I wonder if using a scientific notation can address this use-case

At this point these are still numeric values so we can't apply formatting, see https://github.com/huggingface/transformers/pull/42104#issuecomment-3511663788

stas00:
Apply at the point of print then, while saving the full precision in the log file?

AnMakc:
> Apply at the point of print then, while saving the full precision in the log file?

There are multiple ways of how the values could be displayed - jupyter notebooks, reporting to wandb, trackio, etc. Each handles display rounding separately. See attached screenshots in the first message here. 

This PR makes this behaviour explicit and consistent across different type of values/metrics. If there is a benefit of rounding the metric in specific logs - it could be added in a separate PR.

stas00:
Whatever works. To make it clear - I wasn't objecting to your PR, was just suggesting on possible further work to keep things neat for multiple use-cases since I was tagged as the originator of this "feature".

AnMakc:
Sure, just clarified the intention. Thanks!

@SunMarc Could you merge please?

SunMarc:
To not have regression for other users, maybe we can implement @Rocketknight1 suggestions in `PrinterCallback` class that we have in transformers @AnMakc ? It should be quite straightforward.

> maybe formatting using something like %.4f so that it always displays 4 significant figures with an optional exponent would preserve the clean rendering but still work in cases when the loss is very small?

Thanks a lot @stas00 for the suggestion

SunMarc (approved):
Much better, thank you !

AnMakc:
@SunMarc 

Updated the PR to format metrics in console with 4 significant digits.
Python default is to use scientific notation for values <1e-4 which seems sane threshold.

Example output: 
`{'loss': '0.0002014', 'grad_norm': '0.0001294', 'learning_rate': '2.179e-07', 'epoch': '0.0003269'}`

Note - epoch formatted the same way for consistency.

stas00:
that looks great, except `'epoch': '0.0003269'` which is not a sensible format when facing a human reader, IMHO.

Should it report the percentage of epoch progress? `0.03%` maybe?

AnMakc:
Percentage would not work for second+ epoch.

stas00:
why not? `135.03%` unless we don't have the knowledge about which epoch is running?

stas00:
But please feel free to ignore my suggestions if the current maintainers are ok with your last proposal. They decide.

SunMarc:
I think we can keep this for now. If ppl complains about that, we can still do follow-up pr for specific metrics

PR #42112, opened 2025-11-09
Add max_thinking_tokens for reasoning models (issue #42111)

### LINKED ISSUE #42111: Add thinking-budget support (max_thinking_tokens) for reasoning-capable chat models
### Feature request

A built-in way to cap how many tokens a reasoning model spends inside its ``<think> … </think>`` block. Today, we can only control the total response length via ``max_new_tokens``. No parameter limits the internal reasoning segment when ``enable_thinking=True``.

### Motivation

- Reasoning models (e.g., Qwen3 series) often produce very long thought blocks, which can blow past latency budgets before the final answer starts.
- Users need a simple, model-agnostic control to bound that “thinking” cost without disabling reasoning entirely.
- The Qwen docs (https://qwen.readthedocs.io/en/latest/getting_started/quickstart.html#thinking-budget) already describe a brute-force approach (two-step generation) to implement “thinking budgets”.

### Your contribution

I want to submit a PR that:

- Extends ``GenerationConfig`` with:
``max_thinking_tokens``: integer budget for reasoning tokens.
``begin_thinking_token_id / end_thinking_token_id``: marker IDs so generation knows where the thinking span begins/ends.
- Add a ``MaxThinkingTokensLogitsProcessor`` that watches the active ``<think>`` block. Once the budget is reached, it forces end_thinking_token_id, ensuring the model exits reasoning and continues with the final response.
- Document the new parameter in reasoning-model guides (EXAONE, CWM, etc.) and show how to wire the thinking-token IDs until configs do it automatically.
- Provide unit coverage so ``_get_logits_processor`` injects the new processor whenever the config is fully specified.

AndresAlgaba:
Add feature described in Issue #42111 

# What does this PR do?
A built-in way to cap how many tokens a reasoning model spends inside its ``<think> … </think>`` block. Today, we can only control the total response length via ``max_new_tokens``. No parameter limits the internal reasoning segment when ``enable_thinking=True``.

Motivation:
- Reasoning models (e.g., Qwen3 series) often produce very long thought blocks, which can blow past latency budgets before the final answer starts.
- Users need a simple, model-agnostic control to bound that “thinking” cost without disabling reasoning entirely.
- The Qwen docs (https://qwen.readthedocs.io/en/latest/getting_started/quickstart.html#thinking-budget) already describe a brute-force approach (two-step generation) to implement “thinking budgets”.

This PR:
- Extends ``GenerationConfig`` with:
``max_thinking_tokens``: integer budget for reasoning tokens.
``begin_thinking_token_id / end_thinking_token_id``: marker IDs so generation knows where the thinking span begins/ends.
- Add a ``MaxThinkingTokensLogitsProcessor`` that watches the active ``<think>`` block. Once the budget is reached, it forces end_thinking_token_id, ensuring the model exits reasoning and continues with the final response.
- Document the new parameter in reasoning-model guides (EXAONE, CWM, etc.) and show how to wire the thinking-token IDs until configs do it automatically.
- Provide unit coverage so ``_get_logits_processor`` injects the new processor whenever the config is fully specified.

Fixes #42111 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker @Cyrilvallez @zucchini-nlp


AndresAlgaba:
One thing that still bothers me is that:

```python

model_id = "Qwen/Qwen3-4B"
model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype="auto", device_map="auto")

messages = [{"role": "user", "content": "Give me a concise proof that sqrt(2) is irrational."}]
inputs = tokenizer.apply_chat_template(messages, add_generation_prompt=True, return_tensors="pt").to(model.device)

think_start = tokenizer.convert_tokens_to_ids("<think>")
think_end = tokenizer.convert_tokens_to_ids("</think>")
model.generation_config.begin_thinking_token_id = think_start
model.generation_config.end_thinking_token_id = think_end

out = model.generate(
    **inputs,
    max_new_tokens=128,
    max_thinking_tokens=64,
    do_sample=False,
)

print(tokenizer.decode(out[0], skip_special_tokens=False))
```

will result in: 
```python
<|im_start|>user
Give me a concise proof that sqrt(2) is irrational.<|im_end|>
<|im_start|>assistant
<think>
Okay, so I need to prove that the square root of 2 is irrational. Hmm, I remember that an irrational number is a number that can't be expressed as a fraction of two integers. So, I need to show that there are no integers a and b where a/b equals sqrt(2). Let</think>

To prove that √2 is irrational, we use a **proof by contradiction**. Here's a concise version:

1. **Assume** √2 is rational. Then it can be written as a fraction $ \frac{a}{b} $, where $ a $ and $ b
```
where the ```end_thinking_token``` is appended to the output, whereas by default ```\n<\think>``` is generated by the model, resulting in:
```python
<|im_start|>user
Give me a concise proof that sqrt(2) is irrational.<|im_end|>
<|im_start|>assistant
<think>
Okay, so I need to prove that the square root of 2 is irrational. Hmm, I remember that an irrational number is a number that can't be expressed as a fraction of two integers. So, I need to show that there are no integers a and b where a/b equals sqrt(2). Let
</think>

To prove that √2 is irrational, we use a **proof by contradiction**. Here's a concise version:

1. **Assume** √2 is rational. Then it can be written as a fraction $ \frac{a}{b} $, where $ a $ and $ b
```

AndresAlgaba:
Additionally, the early stopping of the Qwen example (https://qwen.readthedocs.io/en/latest/getting_started/quickstart.html#thinking-budget) adds:
```\n\nConsidering the limited time by the user, I have to give the solution based on the thinking directly now.\n</think>\n\n```

Not sure whether such customization should be allowed? Kind of related to the comment above, but seems like a different logic and maybe for a future PR?

**Comment by AndresAlgaba** [comment]
I might have overcomplicated the ```MaxThinkingTokensLogitsProcessor```, but it is to avoid token overcount when 1) people use ```<think>``` in their user prompt, or 2) people feed previous reasoning traces which include ```<think>...</think>``` (although this is not best practice, see number 4 in https://huggingface.co/Qwen/Qwen3-4B#best-practices).

**Comment by AndresAlgaba** [comment]
Another issue is when ```enable_thinking=True``` pre-fills the prompt with a ```<think>``` token as this becomes part of the ```prompt_length``` and will be ignored. Any thoughts on this? A simple solution here would be ```prompt_length-1```, but I noticed that, for example, Qwen3-4B doesn't do the pre-filling.

**Comment by AndresAlgaba** [comment]
While testing this, another problem is that the model might output a second ```<\think>``` token after ```\n\n``` (so not directly after the enforced ```<\think>``` token):

```python

model_id = "Qwen/Qwen3-4B"
model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype="auto", device_map="auto")

messages = [
    {
        "role": "user",
        "content": "Give me a concise proof that sqrt(2) is irrational.",
    },
]
inputs = tokenizer.apply_chat_template(messages, add_generation_prompt=True, return_tensors="pt").to(model.device)

think_start = tokenizer.convert_tokens_to_ids("<think>")
think_end = tokenizer.convert_tokens_to_ids("</think>")
model.generation_config.begin_thinking_token_id = think_start
model.generation_config.end_thinking_token_id = think_end

out = model.generate(
    **inputs,
    max_new_tokens=64,
    max_thinking_tokens=32,
    do_sample=False,
)

print(tokenizer.decode(out[0], skip_special_tokens=False))
```

```python
<|im_start|>user
Give me a concise proof that sqrt(2) is irrational.<|im_end|>
<|im_start|>assistant
<think>
Okay, so I need to prove that the square root of 2 is irrational. Hmm, I remember that an irrational number is a number that can't</think>

</think>

Sure! Here's a concise proof that $\sqrt{2}$ is irrational:

---

**Proof that $\sqrt{2}$ is irrational
```

AndresAlgaba:
Update should satisfy following requirements:

- When a user uses past reasoning traces (i.e., previous assistant chunks contain `<think> ... </think>`), it should ignore this.
- When a user uses `<think>`, `</think>`, or `<think> ... </think>`, in their prompt, it should be ignored.
- It should count the number of tokens only in the current reasoning chain.
It should be robust against the model generating additional `<think>` tokens in the current reasoning chain (not restart the count).
- It should be robust against the model automatically appending the `<think>` token when `enable_thinking=True` (right now it becomes part of the prompt and gets ignored because of it).
- The model sometimes generates an additional `</think>` token after the enforced closure `</think>` token. Not always directly after, sometimes it first generated `\n\n` and then the `</think>` token. Right now, the `</think>` token can only be generated if a new `<think>` token has been generated.

src/transformers/generation/configuration_utils.py
```diff
@@ -548,6 +562,31 @@ def validate(self, strict=False):
...
+                    raise ValueError(
+                        "Using `max_thinking_tokens` requires `max_new_tokens` to be at least 2 so there is room for "
+                        "the closing token and final response."
+                    )
+                if self.max_thinking_tokens > self.max_new_tokens - 2:
+                    raise ValueError(
+                        "`max_thinking_tokens` must be at most `max_new_tokens - 2` so there is room for the closing "
+                        f"token and at least one response token (got {self.max_thinking_tokens} vs "
+                        f"{self.max_new_tokens})."
+                    )
```
zucchini-nlp: looks like the first error is a subset of the last one, we can keep the second one which has a more clear error msg

src/transformers/generation/utils.py
```diff
@@ -1285,6 +1300,16 @@ def _get_logits_processor(
...
 
+        def _propagate_prompt_length(processor_list: Optional[list]):
+            if input_ids_seq_length is None or processor_list is None:
+                return
+            for processor in processor_list:
+                if isinstance(processor, MaxThinkingTokensLogitsProcessor):
+                    processor.set_prompt_length(input_ids_seq_length)
+
+        _propagate_prompt_length(processors)
+        _propagate_prompt_length(logits_processor)
```
zucchini-nlp: should not be needed since we init the processor with `input_ids_seq_length` already

src/transformers/generation/logits_process.py
```diff
@@ -228,6 +228,159 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
+
+    def set_prompt_length(self, prompt_length: Optional[int]) -> None:
+        """
+        Overrides the prompt length used to determine how many tokens have been generated inside the thinking block.
+
+        This is useful when reusing the same processor instance across multiple calls to `generate`, where each call
+        may start from prompts of different lengths.
+        """
+        self._prompt_length_override = prompt_length
+        self._tracked_prompt_length = prompt_length
```
zucchini-nlp: i don't think we need it. We can expect that users set the correct prompt length if passing a `LogitsProcessor`, otherwise we use the input seq length when init logits processors in `generate()`

src/transformers/generation/logits_process.py
```diff
@@ -228,6 +228,159 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
+
+    @add_start_docstrings(LOGITS_PROCESSOR_INPUTS_DOCSTRING)
+    def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> torch.FloatTensor:
+        batch_size, sequence_length = input_ids.shape
+
+        if batch_size == 0:
+            return scores
```
zucchini-nlp: nit: input ids at this point are guaranteed to be non-empty

src/transformers/generation/logits_process.py
```diff
@@ -228,6 +228,159 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
+            else:
+                self._tracked_prompt_length = min(self._tracked_prompt_length, sequence_length)
+            prompt_length = self._tracked_prompt_length
+        else:
+            self._tracked_prompt_length = prompt_length
+
+        if prompt_length is None:
+            prompt_length = sequence_length
+            self._tracked_prompt_length = prompt_length
+
```
zucchini-nlp: same, imo it is provided at `__init__`. We can make it a required arg to be sure

src/transformers/generation/logits_process.py
```diff
@@ -228,6 +228,159 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
+        scores_processed = scores.clone()
+        if disallow_close_indices:
+            scores_processed[disallow_close_indices, self.end_thinking_token_id] = -math.inf
+
+        if force_end_indices:
+            scores_processed[force_end_indices] = -math.inf
+            end_token_logits = scores[force_end_indices, self.end_thinking_token_id]
+            safe_end_token_logits = torch.where(
+                torch.isfinite(end_token_logits),
```
zucchini-nlp: hm, do we usually get inf values in logits? Should not be the case unless it was manipulated somehow

src/transformers/generation/logits_process.py
```diff
@@ -228,6 +228,159 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
+            first_open_position, block_started = self._thinking_block_state(input_ids[batch_index], prompt_length)
+
+            if first_open_position is None:
+                if block_started:
+                    disallow_close_indices.append(batch_index)
+                continue
+
+            count_start = max(first_open_position + 1, prompt_length)
+            if count_start >= sequence_length:
+                continue
```
zucchini-nlp: this can't be more than input length, no? Can't think of a situation unless the prompt length is set to an incorrect value

src/transformers/generation/logits_process.py
```diff
@@ -228,6 +228,159 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
+        started either in the prompt (and carried over) or in the generated continuation.
+        """
+        sequence_length = sequence.shape[-1]
+        if sequence_length == 0:
+            return None, False
+
+        open_stack: list[int] = []
+        prompt_open_depth = 0
+        prompt_open_depth_at_boundary = 0
+        boundary_recorded = prompt_length <= 0
```
zucchini-nlp: `prompt_length <= 0` looks weird 👀

src/transformers/generation/logits_process.py
```diff
@@ -228,6 +228,159 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
+        self._tracked_prompt_length: Optional[int] = prompt_length
+
+    def _thinking_block_state(self, sequence: torch.LongTensor, prompt_length: int) -> tuple[Optional[int], bool]:
+        """
+        Returns the first open `<think>` position (if any) together with a flag indicating whether a thinking block has
+        started either in the prompt (and carried over) or in the generated continuation.
+        """
+        sequence_length = sequence.shape[-1]
+        if sequence_length == 0:
+            return None, False
```
zucchini-nlp: same, it will have at least one token because prompt can't be empty. We'll artificially add BOS if it's empty, iirc

src/transformers/generation/logits_process.py
```diff
@@ -228,6 +228,159 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
+        if not boundary_recorded:
+            prompt_open_depth_at_boundary = prompt_open_depth
+
+        first_open_position: Optional[int] = None
+        if prompt_open_depth > 0:
+            first_open_position = max(prompt_length - 1, 0)
+        elif open_stack:
+            first_open_position = open_stack[0]
+
+        block_started = prompt_open_depth_at_boundary > 0 or seen_begin_after_prompt or first_open_position is not None
```
zucchini-nlp: IIUC we are disallowing any new `</think>` tokens if at least one opening think token is seen in the input prompt or generated part. Can you explain why would we need it?

I think we have to manually disallow all opening/closing think tokens only when over max-budget

src/transformers/generation/logits_process.py
```diff
@@ -228,6 +228,159 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
+            if not boundary_recorded and position >= prompt_length - 1:
+                prompt_open_depth_at_boundary = prompt_open_depth
+                boundary_recorded = True
+
+        if not boundary_recorded:
+            prompt_open_depth_at_boundary = prompt_open_depth
+
+        first_open_position: Optional[int] = None
+        if prompt_open_depth > 0:
+            first_open_position = max(prompt_length - 1, 0)
```
zucchini-nlp: i wonder how often we get a prompt which started thinking but didn't finish yet 🤔

src/transformers/generation/logits_process.py
```diff
@@ -228,6 +228,159 @@ def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> to
...
+        if sequence_length == 0:
+            return None, False
+
+        open_stack: list[int] = []
+        prompt_open_depth = 0
+        prompt_open_depth_at_boundary = 0
+        boundary_recorded = prompt_length <= 0
+        seen_begin_after_prompt = False
+
+        for position in range(sequence_length):
```
zucchini-nlp: would be nice to simplify the below logic and comment more to make maintaining easier. We don't have a dedicated person to maintain generation currently, so a straight-forward and easy-readable code is super needed

zucchini-nlp:
@AndresAlgaba hey!
Great job, thanks a lot for the PR! I think it a super useful feature to have. I left a few general comment and question above, mostly to make sure I understand the flow and to simplify it when possible

Rocketknight1:
I'm actually not sure about this PR! I think forcibly ending model thinking mid-stream will force them out of distribution and cause them to output very low-quality answers. We could add it as a feature for advanced users, but I worry that if we do that then people will assume that because it's a feature that it works well in practice, when I think it probably won't, right?

Note that in the [Qwen example](https://qwen.readthedocs.io/en/latest/getting_started/quickstart.html#thinking-budget), they don't cut off thinking directly, but instead insert an "early stopping prompt" that is likely fine-tuned for Qwen and will not work well on other thinking models.

AndresAlgaba:
@zucchini-nlp Thank you for the feedback, working on it!

@Rocketknight1 Thanks for the comment! Indeed, I agree it is more of an experimental feature. It is in line with Gemini's thinking_budget (https://ai.google.dev/gemini-api/docs/thinking#set-budget- and Claude's budget_tokens (https://docs.claude.com/en/docs/build-with-claude/extended-thinking#how-to-use-extended-thinking). I remember that DeepSeek was planning to implement a max_thinking_tokens parameter at some point, but I can't find it anymore. I implemented it for my own research and thought others might be interested in this feature, but with the warranty that it is experimental. So not sure whether it should then be added to the library like this?

Rocketknight1:
hi @AndresAlgaba, yes, but note that that feature in Gemini or Claude does not actually cap the reasoning tokens or forcibly stop thinking - they specifically say the model can overflow the budget. They've probably just trained the model so that it (roughly) obeys thinking budgets, and then they probably tell it the budget in the prompt somewhere

src/transformers/generation/logits_process.py
AndresAlgaba: Good question. It might be an edge case, but, for example, it happens when a <think> token gets automatically appended when ```enable_thinking=True```, or is a user (for no good reason) puts <think>  in their prompt. Or when the messages contain earlier assistant parts, which still contain the reasoning (without closing properly). However, I now only look back a few tokens in the prompt and have added the ```_prompt_prefilled_suffix_length``` argument.

src/transformers/generation/logits_process.py
AndresAlgaba: Agree, this should no longer be the case.

src/transformers/generation/logits_process.py
AndresAlgaba: Did some refactoring and added comments!

AndresAlgaba:
True, same for the reasoning effort in OpenAI models. I still believe some people might be interested in the feature, see a similar feature request in vllm: https://github.com/vllm-project/vllm/issues/17887 (and PR: https://github.com/vllm-project/vllm/pull/20859, where they also propose a logit processor). Is there any way we can indicate it's experimental? Is it an option to keep the processor without adding the```max_thinking_tokens``` argument for now?

I believe that adding the ```\n\nConsidering the limited time by the user, I have to give the solution based on the thinking directly now.\n</think>\n\n``` part is interesting, but part of different PR?

Rocketknight1:
Hmn, if we can mention in the docs that it's experimental and might reduce answer quality it's probably okay! cc @zucchini-nlp, are you happy with that?

AndresAlgaba:
Thanks again for the feedback! I have made some changes after further experimentation and have updated the docs (including a warning that the feature is experimental).

src/transformers/generation/logits_process.py
zucchini-nlp: Hmm, I'd personally prefer to not add more args in the generation config. It's already quite bloated due to historical reasons and we've been trying to not add new generation techniques recently (instead re-directing them to[ hub as custom code](https://huggingface.co/docs/transformers/en/generation_strategies#custom-generation-methods))

AndresAlgaba: Ok, I see. In the most recent version I added this, but happy to change it!

zucchini-nlp (commented):
@AndresAlgaba thanks for iterating on this!

My first review might have been a bit rushed. Looking at comments from @Rocketknight1 and how thinking budget is handled in some other models, truncating generation might not be the best option performance-wise. I realize though that the feature is requested a lot and community is excited to have smth working

So, to check community reaction and assess usage, wdyt of hosting it as a custom generation method on the hub (following [the docs](https://huggingface.co/docs/transformers/en/generation_strategies#custom-generation-methods))? We can simply copy the logits processor to the hub and call `model.generate()` with it (I can help with that and add in `transformers-community` org). It will also give us more freedom to experiment and introduce breaking changes in the future

AndresAlgaba:
Seems like a good idea! I agree that hard truncation is often not the way to go. Thank you for offering help, I will look into it and keep you posted here!

AndresAlgaba:
Hi @zucchini-nlp, sorry for not following up on this but I decided to look further into the "early stopping" schemes and write a paper on it (now available on arxiv: https://arxiv.org/abs/2601.23163). I will experiment further and then start to add to the ```transformers-community``` org!

zucchini-nlp:
Amazing @AndresAlgaba ! If you have a paper/blog-post, feel free to link it directly in the hub page via `hub papers` :)

qgallouedec:
closing, see this thread, and https://github.com/huggingface/transformers/issues/42111#issuecomment-5106174358 for a recap

PR #42126, opened 2025-11-10
Avoid mutating user-provided arguments in preprocessing utils

LeonardoEmili:
# What does this PR do?

As per title, `_merge_kwargs` removes user-provided arguments via `pop` which is fixed by copying the arguments before attempting to modify them. This happens for instance when calling the `Gemma3nProcessor` with modality-specific kwargs.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker @Cyrilvallez

### DIFF
--- a/src/transformers/processing_utils.py
+++ b/src/transformers/processing_utils.py
@@ -1187,7 +1187,6 @@ class ProcessorMixin(PushToHubMixin):
         tokenizer_init_kwargs: Optional[dict] = None,
         **kwargs,
     ) -> dict[str, dict]:
-        kwargs = copy.deepcopy(kwargs)
         """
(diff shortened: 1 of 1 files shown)


Rocketknight1:
Hey, seems like a good idea, but two comments:

1) Please move any code below the docstring! Otherwise the docstring stops being recognized as a docstring.
2) `deepcopy()` here seems dangerous, since `kwargs` might include tensors. Since we're only worried about the dict being mutated by `pop()`, let's just do a shallow copy or call  `kwargs.copy()`

Rocketknight1 (approved):
Yep, looks good now. Thank you for the fix!

PR #42128, opened 2025-11-10
Fix model name test for compressed tensors 

SunMarc:
# What does this PR do?

As per title, the models that were used to test compressed-tensors changed. Also test_llama_8b_fp8 output depends on the hardware, so I changed it even it passes on h100.


MekkCyber (approved):
Thanks

PR #42129, opened 2025-11-10
[TP] Fix parameter detection issue and some invalid TP-plans

### LINKED ISSUE #42016: Load Mixtral-8x22B-Instruct-v0.1 lead to OOM
### System Info

- `transformers` version: 4.56.1
- Platform: Linux-5.10.0-1.oe.x86_64-x86_64-with-glibc2.34
- Python version: 3.12.11
- Safetensors version: 0.6.2
- Accelerate version: 1.10.1
- PyTorch version (accelerator?): 2.8.0+cu128 (CUDA)
- Jax version: not installed
- Using GPU in script?: yes
- GPU type: NVIDIA H800

### Who can help?

_No response_

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

The code I used is as follows:
```python
import torch

if __name__ == "__main__":
    torch.distributed.init_process_group("nccl")
        "mistralai/Mixtral-8x22B-Instruct-v0.1",
        tp_plan="auto",
        tp_size=8,
        dtype="auto",
    )
```
I run the script with the command:
```bash
torchrun --nproc-per-node=8 load.py
```

### Expected behavior

Load the model(~262G) with 8×H800(80G) should not lead to OOM

Cyrilvallez:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/42016. Currently, we replace any number in a parameter name with a `*` to match the tp_plan. However, params may contain numbers naturally in their name, such as Mixtral's MLPs containing `w1`, `w2` and `w3` param names. This means that they were replaced, and thus not detected as needing sharding. So basically they would be invisible from the tp_plan and never sharded 🥵🥵

The regex now checks for `.` (dots) on each side of the numbers, to only match layers or list names as we need.

Also fixes a few TP-plans, that were detected wrong by the test 🙃🙃


ArthurZucker (commented):
Missing a big test ! mega important!

PR #41750, opened 2025-10-20
:rotating_light: [`Clip`] Fix masking and enable flash attention on all model types

vasqu:
Clip used old mask APIs leading to a confused usage:
- A causal mask (normal triu mask)
- A padding mask (encoder mask == only accounting for padding)
- Add both of above == final mask --> causal mask with padding

^ works only for interfaces with support for 4D masks which disabled FA usage in general.

This PR now correctly changes this to the new API which handles padding automatically. We have to additionally pass the `is_causal` kwarg to dynamically switch between modality types (text == causal, image == full). This is only enabled through recent PRs (fa #39707, sdpa #41692).

Closes #41673
Fixes #41668

PR #28818, opened 2024-02-01
[`OWL-VIT`] Added sdpa attention

nileshkokane01:
# What does this PR do?

This PR add sdpa attention for OWL-ViT.

Fixes #28103

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests),
      Pull Request section?
- [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

@NielsRogge  @younesbelkada
Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #42136, opened 2025-11-10
Sdpa for owlvit

### LINKED ISSUE #28103: OWL-VIT Vision Foundation Model deployment in the edge cases - Need SDPA support for OWL-ViT Model optimization for Edge Deployment
### Feature request

Hi Team,
I am working with  OWL-ViT Size model which has around 611 MB size ( https://huggingface.co/google/owlvit-base-patch16).
I want to optimize this model and like to deploy in the edge device for object detection.

Come to know from the group torch.scaled_dot_product_attention can be used for model optimization. 

I need your feedback comments how optimally we can reduce the memory size so that we can deploy in the edge device.

waiting for your response.

with thanks

### Motivation

It will help to deploy the models in edge so that more applications we can use it.

### Your contribution

Like to know your feedback comments.

Aravind-11:
# What does this PR do?
Implements SDPA for OWL VIT. 

Fixes #28103

## Before submitting
      to it if that's the case.
- [] Did you write any new necessary tests?

## Who can review?

 @vasqu @younesbelkada


Aravind-11:
> # What does this PR do?
> Implements SDPA for OWL VIT. Revamp of #28818
> 
> Fixes #28103
> 
> ## Before submitting
> * [x]  Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
>   to it if that's the case.
> * [ ]  Did you make sure to update the documentation with your changes? Here are the
>   [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
>   [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
> * [] Did you write any new necessary tests?
> 
> ## Who can review?
> @vasqu @younesbelkada

I ran the RUN_SLOW=1 python -m pytest tests/models/owlvit/test_modeling_owlvit.py for the original owlvit implementation and it seemed to fail the same tests as my current implementation. I'm not sure how to infer from that.

vasqu (commented):
Sorry but I've got to be strict about this. We no longer implement separate classes for all the attention flavors but one unified one. I think ViT is a good example in this case, e.g. see https://github.com/huggingface/transformers/blob/main/src/transformers/models/vit/modeling_vit.py

Before changing this to these standards I won't take a proper look for now.

Aravind-11:
> Sorry but I've got to be strict about this. We no longer implement separate classes for all the attention flavors but one unified one. I think ViT is a good example in this case, e.g. see https://github.com/huggingface/transformers/blob/main/src/transformers/models/vit/modeling_vit.py
> 
> Before changing this to these standards I won't take a proper look for now.

Got it. Thanks a lot!

Aravind-11:
> Sorry but I've got to be strict about this. We no longer implement separate classes for all the attention flavors but one unified one. I think ViT is a good example in this case, e.g. see https://github.com/huggingface/transformers/blob/main/src/transformers/models/vit/modeling_vit.py
> 
> Before changing this to these standards I won't take a proper look for now.

I made similar changes as in the vit and removed the seperate sdpa class. Let me know what you think!

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -713,18 +711,16 @@ def forward(
...
+        # OWL-ViT uses a bidirectional (non-causal) encoder.
+        attention_mask = create_bidirectional_mask(
+            config=self.config,
+            input_embeds=hidden_states,
+            attention_mask=attention_mask,
         )
-        # expand attention_mask
-        if attention_mask is not None:
-            # [num_samples, seq_len] -> [num_samples, 1, tgt_seq_len, src_seq_len]
-            attention_mask = _prepare_4d_attention_mask(attention_mask, hidden_states.dtype)
```
vasqu: This seems to suffer from the same issue as in #41750

It does not use a bidirectional mask, but a causal mask:
- The first mask is a based causal mask
- The second is a padding mask
- These are added on top creating a causal mask with padding included

vasqu: This also may need to adjust the `is_causal` argument dynamically as in the PR I linked - although I'm not sure if it's just causal in general

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -552,6 +551,7 @@ class OwlViTPreTrainedModel(PreTrainedModel):
     base_model_prefix = "owlvit"
     input_modalities = ["image", "text"]
     supports_gradient_checkpointing = True
+    _supports_sdpa = True
```
vasqu, line 520: It should also support flash attn and flex attention then

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -495,9 +513,9 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
         return hidden_states
 
 
-# Copied from transformers.models.altclip.modeling_altclip.AltCLIPEncoderLayer with AltCLIP->OwlViT
+# Copied from transformers.models.clip.modeling_clip.CLIPEncoderLayer with CLIP->OwlViT
```
vasqu, line 480: This doesn't work, output_attentions would be ignored --> see the `check_model_inputs` decorator in other models like CLIP

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -388,95 +388,113 @@ def __init__(self, config):
...
+                key_states,
+                value_states,
+                attention_mask,
+                is_causal=self.is_causal,
+                scaling=self.scale,
+                dropout=self.dropout if self.training else 0.0,
             )
+        else:
+            # Eager attention implementation
+            # Scale query
```
vasqu: Please follow what is done in other models, we should be able to copy the eager function from bert (see clip as example)

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -388,95 +388,113 @@ def __init__(self, config):
...
+        value_states = value_states.view(bsz, tgt_len, self.num_heads, self.head_dim).transpose(1, 2)
 
-        src_len = key_states.size(1)
-        attn_weights = torch.bmm(query_states, key_states.transpose(1, 2))
+        # Prepare attention mask - OWL-ViT uses bidirectional (non-causal) attention
+        attention_mask = create_bidirectional_mask(
+            config=self.config,
+            input_embeds=hidden_states,
+            attention_mask=attention_mask,
+        )
```
vasqu: The mask is prepared outside, not within the attention module

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -388,95 +388,113 @@ def __init__(self, config):
...
+        value_states = self.v_proj(hidden_states)
 
-        proj_shape = (bsz * self.num_heads, -1, self.head_dim)
-        query_states = self._shape(query_states, tgt_len, bsz).view(*proj_shape)
-        key_states = key_states.view(*proj_shape)
-        value_states = value_states.view(*proj_shape)
+        # Reshape to (bsz, num_heads, seq_len, head_dim)
+        query_states = query_states.view(bsz, tgt_len, self.num_heads, self.head_dim).transpose(1, 2)
+        key_states = key_states.view(bsz, tgt_len, self.num_heads, self.head_dim).transpose(1, 2)
+        value_states = value_states.view(bsz, tgt_len, self.num_heads, self.head_dim).transpose(1, 2)
```
vasqu: Please follow llama / clip, the -1 dim is also important for example

vasqu (commented):
Added some comments but in general it would be best to have a green CI before requesting a review. Atm, things are likely not working as expected

src/transformers/models/owlvit/modeling_owlvit.py
Aravind-11: Thanks! I made some changes to the code after referring to CLIP - removing the output_attention, return dict and casual_attention_mask. Also copied the eager attention part, attention reshaping from CLIP. Added the flash and flex attn too. 

I think that the current CI is failing because the OWL VIT config file is conflicting with the current encoder implementation. Could you guide me here? Thanks a lot!

Aravind-11: > Thanks! I made some changes to the code after referring to CLIP - removing the output_attention, return dict and casual_attention_mask. Also copied the eager attention part, attention reshaping from CLIP. Added the flash and flex attn too.
> 
> I think that the current CI is failing because the OWL VIT config file is conflicting with the current encoder implementation. Could you guide me here? Thanks a lot!

Hi, I investigated the failing OwlViTForObjectDetectionTest::test_eager_matches_sdpa_inference_09_fp32_pad_left.

The failure is due to the test invoking OwlViTForObjectDetection.forward() without providing pixel_values.

OwlViTForObjectDetection requires pixel_values (image tensors) for its vision backbone. When the test omits them, the model raises a ValueError: 'pixel_values' is None.

Aravind-11: Also, when I run make fix-copies, it's add output_attention and create_causal_mask parameters in owlvitencoderlayer.forward() function.

vasqu:
@Aravind-11 it seems like ~180 tests are failing, so there might be some really serious regressions. Don't worry about the copies, functionality is more important first.

If pixel values are needed for example, I would first check if they were needed before. If yes, see what changed, if not, the inputs preparation may need to change in the test file.

src/transformers/models/owlvit/modeling_owlvit.py
vasqu: Responded here https://github.com/huggingface/transformers/pull/42136#issuecomment-3541719143

Resolving my previous comments since the state has changed quite a bit from last time

Aravind-11:
> @Aravind-11 it seems like ~180 tests are failing, so there might be some really serious regressions. Don't worry about the copies, functionality is more important first.
> 
> If pixel values are needed for example, I would first check if they were needed before. If yes, see what changed, if not, the inputs preparation may need to change in the test file.

Hi, @vasqu , thank you for the comments! I reverted changes in owlv2 back to original for easier debugging. Only the pixel_values() seem to be causing issues in OWLVIT as per the tests. 

The pixel_values are needed in the original implementation in ObjectionDetection.forward() too.  

def forward(
        self,
        input_ids: torch.Tensor,
        pixel_values: torch.FloatTensor,
        output_attentions: Optional[bool] = None,
        output_hidden_states: Optional[bool] = None,
        interpolate_pos_encoding: bool = False,
        return_dict: Optional[bool] = None,
    ) 

this is the original forward signature. Please let me know what you think. Thanks!

vasqu:
Have you checked for a correct `prepare_config_and_inputs_for_common` as in ViT https://github.com/huggingface/transformers/blob/2cc9152da08dc9a2d1ac5e738b8eb764ca369717/tests/models/vit/test_modeling_vit.py#L173-L181

This might cause these mismatches as we use that for generation specific tests (sdpa to eager equivalence included). I.e. it needs to be converted to a dict for us to pass as kwargs directly

Aravind-11:
> Have you checked for a correct `prepare_config_and_inputs_for_common` as in ViT
> 
> https://github.com/huggingface/transformers/blob/2cc9152da08dc9a2d1ac5e738b8eb764ca369717/tests/models/vit/test_modeling_vit.py#L173-L181
> 
> This might cause these mismatches as we use that for generation specific tests (sdpa to eager equivalence included). I.e. it needs to be converted to a dict for us to pass as kwargs directly

def prepare_config_and_inputs_for_common(self):
        config_and_inputs = self.prepare_config_and_inputs()
        config, pixel_values, input_ids, attention_mask = config_and_inputs
        inputs_dict = {
            "pixel_values": pixel_values,
            "input_ids": input_ids,
            "attention_mask": attention_mask,
        }
        return config, inputs_dict 

this is the current prepare_config_and_inputs_for_common() for OwlViTForObjectDetectionTester. is this overrriden ?

> Have you checked for a correct `prepare_config_and_inputs_for_common` as in ViT
> 
> https://github.com/huggingface/transformers/blob/2cc9152da08dc9a2d1ac5e738b8eb764ca369717/tests/models/vit/test_modeling_vit.py#L173-L181
> 
> This might cause these mismatches as we use that for generation specific tests (sdpa to eager equivalence included). I.e. it needs to be converted to a dict for us to pass as kwargs directly

    def prepare_config_and_inputs_for_common(self):
          config_and_inputs = self.prepare_config_and_inputs()
          config, input_ids, attention_mask, pixel_values = config_and_inputs
          inputs_dict = {
              "pixel_values": pixel_values,
              "input_ids": input_ids,
              "attention_mask": attention_mask,
              "return_loss": False,
          }
          return config, inputs_dict 

yes, the pixel_values are passed for the vision model tester.

vasqu:
Sorry have a bit more todo atm, would need to debug myself! This is weird, but it does seem something in the preparation seems to go wrong :cry:

Aravind-11:
> Sorry have a bit more todo atm, would need to debug myself! This is weird, but it does seem something in the preparation seems to go wrong 😢

Got it! No worries. Let me know if you find something. Thanks.

Aravind-11:
> > Sorry have a bit more todo atm, would need to debug myself! This is weird, but it does seem something in the preparation seems to go wrong 😢
> 
> Got it! No worries. Let me know if you find something. Thanks.

I have a doubt .. @vasqu I noticed some models still have output_attention and return_dict() in their forward signatures inspite of having **kwargs --> MPT, T5, Paligemma, pix2struct.. why is that ? and should all models have sdpa, flash and flex attn enabled?

vasqu:
It's connected but not necessary, the attention interface allows the usage of all attentions (fa, sdpa, flex). The signature along output_xxx, return_dict is a different matter and is handled via decorators such as `can_return_dict` or `check_model_inputs`. The latter needs `_can_record_outputs` to be properly adjusted as attribute of that model.

tl;dr: separate things but usually you need the attention interface first and then the decorators. There are cases where it's possible that only one has been applied.

Aravind-11:
> It's connected but not necessary, the attention interface allows the usage of all attentions (fa, sdpa, flex). The signature along output_xxx, return_dict is a different matter and is handled via decorators such as `can_return_dict` or `check_model_inputs`. The latter needs `_can_record_outputs` to be properly adjusted as attribute of that model.
> 
> tl;dr: separate things but usually you need the attention interface first and then the decorators. There are cases where it's possible that only one has been applied.

ohh okay , got it! Makes sense. Thank you!

Aravind-11:
hii @vasqu! 😁 , were you able to look into this? thank you!

vasqu:
Sorry didnt have time yet, can you also check/debug what happens?

Aravind-11:
> Sorry didnt have time yet, can you also check/debug what happens?

no worries, I'll take a look!

Aravind-11:
Hi @vasqu 😄 ,

I traced the issue with the failing SDPA padding test (test_eager_matches_sdpa_inference_24_fp32_pad_left_output_attentions) and found that the shared SDPA test helper in test_modeling_common.py was dropping required inputs for multi-modal models. Specifically, the helper only preserved main_input_name and additional_model_inputs, which caused OWL-ViT to lose input_ids during the SDPA inference path.

I added a small change to the helper so that it now preserves all remaining keys from inputs_dict (without overriding the special-cased ones). 

After applying this fix, the previously failing SDPA test for OWL-ViT passes. However, this change breaks approximately 280 tests for other models with errors like "RuntimeError: to - 1 is out of bounds for bool" and dimension mismatches. It seems the test framework deliberately excludes or modifies certain inputs, and my fix adds back the original unmodified versions, causing conflicts.

Could you advise on the best approach here? 

Thanks!

vasqu:
> Specifically, the helper only preserved main_input_name and additional_model_inputs, which caused OWL-ViT to lose input_ids during the SDPA inference path.

Could it be that we need to change the main input name?

> I added a small change to the helper so that it now preserves all remaining keys from inputs_dict (without overriding the special-cased ones).

This seems a bit too specific to owlvit in this case, so I'd rather we override it in the owlvit test instead of affecting all other models.

> Could you advise on the best approach here?

- Try to find if we can change certain settings so that it works out of the box, e.g. changing `main_input_name`.
- Override it specifically in the test file of owl vit, not in the common file. Like you noticed, it breaks other models

Aravind-11:
thank u thank u , let me see what I can do 😄

Aravind-11:
Hi @vasqu, 

when i enable flash attention , i get this following error , so i have disabled it for now to focus on the functional equivalence. 

FAILED tests/models/owlvit/test_modeling_owlvit.py::OwlViTForObjectDetectionTest::test_flash_attn_2_inference_equivalence_right_padding - ValueError: Could not find the currently requested flash attention implementation at `flash_attention_2`.Make sure that you reque...

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -388,95 +413,49 @@ def __init__(self, config):
...
+        keys = self.k_proj(hidden_states)
+        values = self.v_proj(hidden_states)
+
+        queries = queries.view(batch_size, seq_length, -1, self.head_dim).transpose(1, 2)
+        keys = keys.view(batch_size, seq_length, -1, self.head_dim).transpose(1, 2)
+        values = values.view(batch_size, seq_length, -1, self.head_dim).transpose(1, 2)
+
+        attention_interface: Callable = eager_attention_forward
+        attn_impl = getattr(self.config, "_attn_implementation", None)
+        if attn_impl and attn_impl != "eager":
```
vasqu: Yea this seems to be a bug to me, I think Im also encountering this in #42848, i.e. the config does not have any attention implementation

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -552,6 +514,14 @@ class OwlViTPreTrainedModel(PreTrainedModel):
     base_model_prefix = "owlvit"
     input_modalities = ("image", "text")
     supports_gradient_checkpointing = True
+    _supports_sdpa = True
+    _supports_flash_attn = False
```
vasqu: Ig this is due to https://github.com/huggingface/transformers/pull/42136#issuecomment-3678268402?

There really seems something going completely wrong here, `config.attn_implementation` is not properly set somehow.

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -695,47 +634,36 @@ def forward(
...
-        encoder_outputs = self.encoder(
+        encoder_outputs: BaseModelOutput = self.encoder(
             inputs_embeds=hidden_states,
             attention_mask=attention_mask,
-            causal_attention_mask=causal_attention_mask,
-            output_attentions=output_attentions,
-            output_hidden_states=output_hidden_states,
-            return_dict=return_dict,
+            is_causal=True,
+            **kwargs,
```
vasqu, line 670: Have you checked integration tests? E.g. `pytest tests/models/owlvit -k integration`

So the models used a causal mask then :eyes:

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -925,18 +829,6 @@ class OwlViTModel(OwlViTPreTrainedModel):
...
-            raise TypeError(
-                "config.text_config is expected to be of type OwlViTTextConfig but is of type"
-                f" {type(config.text_config)}."
-            )
-
-        if not isinstance(config.vision_config, OwlViTVisionConfig):
-            raise TypeError(
-                "config.vision_config is expected to be of type OwlViTVisionConfig but is of type"
-                f" {type(config.vision_config)}."
-            )
```
vasqu, line 845: Unrelated?

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -1305,21 +1182,33 @@ def image_text_embedder(
...
+            raise ValueError(
+                "image_text_embedder: outputs.vision_model_output is None. "
+                "Make sure pixel_values were passed to the inner OwlViTModel."
+            )
+
+        if (
+            getattr(outputs, "vision_model_output", None) is not None
+            and getattr(outputs.vision_model_output, "last_hidden_state", None) is None
+        ):
+            raise ValueError("image_text_embedder: vision_model_output.last_hidden_state is None.")
```
vasqu: Shouldnt be needed, lets remove

tests/models/owlvit/test_modeling_owlvit.py
```diff
@@ -296,6 +296,45 @@ def prepare_config_and_inputs_for_common(self):
         return config, inputs_dict
 
 
+class OWLVITModelTesterMixin(ModelTesterMixin):
```
vasqu: Seems not to be used anywhere?

vasqu (commented):
Sorry for the delays, I was off for the holidays. I've added a few comments but I feel like there is some deeper bug (not on you) with the attn implementation setting within the config. I also encountered this.

I have a similar PR here #42848 fyi, which is where I encountered this. So needs to be fixed either way :sweat:

Aravind-11:
> Sorry for the delays, I was off for the holidays. I've added a few comments but I feel like there is some deeper bug (not on you) with the attn implementation setting within the config. I also encountered this.
> 
> I have a similar PR here #42848 fyi, which is where I encountered this. So needs to be fixed either way 😓

Hope you enjoyed your hols and Happy new year! I'll wait for your pr merge and then see what i can do here?

vasqu:
Thank you yes I did :) happy new year! 

Up to you, if you can find the issue great but no pressure. I'll investigate what happens here

Aravind-11:
> Thank you yes I did :) happy new year!
> 
> Up to you, if you can find the issue great but no pressure. I'll investigate what happens here

okay! i'll see what i can do :))

vasqu:
@Aravind-11 I finally came back around and wrapped up the mask PR :hugs: I also found the issues in the attention mask so they should be fixed as well

Aravind-11:
> @Aravind-11 I finally came back around and wrapped up the mask PR 🤗 I also found the issues in the attention mask so they should be fixed as well

wow, that looks like a hell lot of cleanup! 😄 😄

Aravind-11:
let me checkl what changes to make here! I forgot everything 😭 😭

vasqu:
No worries, take your time :heart:

vasqu:
Fyi, our decorators changed so it will need some modifications, i.e. see #43979

Aravind-11:
> Fyi, our decorators changed so it will need some modifications, i.e. see #43979

ahh, thank you! 😄

Aravind-11:
@vasqu sir! could you please check it now ? can i run the make fix-copies for changes to owlv2 as well?

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -486,9 +477,9 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
         return hidden_states
 
 
-# Copied from transformers.models.altclip.modeling_altclip.AltCLIPEncoderLayer with AltCLIP->OwlViT
+# Copied from transformers.models.clip.modeling_clip.CLIPEncoderLayer with CLIP->OwlViT
```
vasqu, line 480: There is a lot of copied from statements, what do you think about refactoring to modular instead, e.g. see the docs here https://huggingface.co/docs/transformers/modular_transformers#implementing-a-modular-file? 

The PR changed quite a bit so it might be easier to do. No worries if not tho, up to you

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -1300,19 +1194,15 @@ def image_text_embedder(
...
     ) -> tuple[torch.FloatTensor]:
-        # Encode text and image
         outputs = self.owlvit(
             pixel_values=pixel_values,
             input_ids=input_ids,
             attention_mask=attention_mask,
-            output_attentions=output_attentions,
-            output_hidden_states=output_hidden_states,
             interpolate_pos_encoding=interpolate_pos_encoding,
-            return_dict=True,
```
vasqu: More of a fan to keep the `return_dict=True` (should probably happen elsewhere as well)

src/transformers/models/owlv2/modeling_owlv2.py
vasqu, line 1: I assume it's pretty much a copy of owlvit?

tests/models/owlv2/test_modeling_owlv2.py
```diff
@@ -550,6 +556,10 @@ def test_model(self):
     def test_hidden_states_output(self):
         pass
 
+    @parameterized.expand(TEST_EAGER_MATCHES_SDPA_INFERENCE_PARAMETERIZATION)
+    def test_eager_matches_sdpa_inference(self, *args):
+        self.skipTest("Owlv2ObjectDetectionOutput has no top-level hidden_states; SDPA tested in sub-models")
+
```
vasqu, line 562: I think some flash attention tests are similar then, can you check? You only need to have kernels (`pip install kernels`) installed for FA, not the original package.

tests/models/owlvit/test_modeling_owlvit.py
```diff
@@ -292,6 +294,45 @@ def prepare_config_and_inputs_for_common(self):
...
+            if hasattr(model_sdpa, "vision_model"):
+                self.assertTrue(model_sdpa.vision_model.config._attn_implementation == "sdpa")
+                self.assertTrue(model_eager.vision_model.config._attn_implementation == "eager")
+
+            if hasattr(model_sdpa, "text_model"):
+                self.assertTrue(model_sdpa.text_model.config._attn_implementation == "sdpa")
+                self.assertTrue(model_eager.text_model.config._attn_implementation == "eager")
+
+            self.assertTrue(model_sdpa.config._attn_implementation == "sdpa")
+            self.assertTrue(model_eager.config._attn_implementation == "eager")
```
vasqu:
```suggestion
```
It is not used anywhere + should be outdated

vasqu (commented):
I did some small cleanup for owlv2 and the attention module. Imo, it might make sense to adjust this PR to refactor with modular as there is a lot of copied from statements; up to you if you want to. Otherwise, some smaller comments left

Will check run-slow for integration tests in a second

vasqu:
run-slow: owlv2, owlvit

src/transformers/models/owlvit/modeling_owlvit.py
Aravind-11: got it .

src/transformers/models/owlv2/modeling_owlv2.py
Aravind-11: yes!

src/transformers/models/owlvit/modeling_owlvit.py
Aravind-11: can we do that modular thing in a different pr? i feel like the current pr itself has diverged a lot from previous. i will raise a pr if you'd like to review.

tests/models/owlv2/test_modeling_owlv2.py
Aravind-11: I overrode flash_attn_inference_equivalence in the detection test class to skip all FA variants. is that fine?

vasqu: Can we use `@unittest.skip(reason=...)` instead; the sdpa one is different because of the parametrization but here we should be fine (same for owlvit)

src/transformers/models/owlvit/modeling_owlvit.py
vasqu: Yea no worries!

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -791,43 +732,38 @@ def forward(
...
-    ) -> tuple | BaseModelOutputWithPooling:
-        output_attentions = output_attentions if output_attentions is not None else self.config.output_attentions
-        output_hidden_states = (
-            output_hidden_states if output_hidden_states is not None else self.config.output_hidden_states
-        )
-        return_dict = return_dict if return_dict is not None else self.config.use_return_dict
+        **kwargs: Unpack[TransformersKwargs],
+    ) -> BaseModelOutputWithPooling:
+        if pixel_values is None:
+            raise ValueError("You have to specify pixel_values")
```
vasqu: This is a bit weird, we can remove these

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -678,28 +634,24 @@ def __init__(self, config: OwlViTTextConfig):
...
             [`PreTrainedTokenizer.encode`] and [`PreTrainedTokenizer.__call__`] for details. [What are input
             IDs?](../glossary#input-ids)
         """
-        output_attentions = output_attentions if output_attentions is not None else self.config.output_attentions
-        output_hidden_states = (
-            output_hidden_states if output_hidden_states is not None else self.config.output_hidden_states
-        )
-        return_dict = return_dict if return_dict is not None else self.config.use_return_dict
+        if input_ids is None:
+            raise ValueError("You have to specify input_ids")
```
vasqu: Same here

vasqu (approved):
Can you check the failing flex attention test https://github.com/huggingface/transformers/pull/42136#issuecomment-3984946681

Otherwise just some small comments

Aravind-11:
> Can we use @unittest.skip(reason=...) instead; the sdpa one is different because of the parametrization but here we should be fine (same for owlvit)
@vasqu  the base ModelTesterMixin uses @parameterized.expand on test_eager_matches_sdpa_inference, which generates numbered test methods at class creation time. @unittest.skip only overrides the base method name, not those generated variants, so they still run and fail with KeyError: 'hidden_states'.

Aravind-11:
> > Can we use @unittest.skip(reason=...) instead; the sdpa one is different because of the parametrization but here we should be fine (same for owlvit)
> > @vasqu  the base ModelTesterMixin uses @parameterized.expand on test_eager_matches_sdpa_inference, which generates numbered test methods at class creation time. @unittest.skip only overrides the base method name, not those generated variants, so they still run and fail with KeyError: 'hidden_states'.

@vasqu hoping to hear back from you 😄

vasqu:
Hey @Aravind-11 sorry for the delays and late reply. I worked on this monster of a PR #43590 :joy: 

Can you sync with main and adjust accordingly? sorry should've mentioned that earlier - it should do a lot of what you did already minus the attention refactor

Aravind-11:
@vasqu pls take a look , thank you!!

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -391,81 +420,45 @@ def __init__(self, config):
...
-        key_states = key_states.view(*proj_shape)
-        value_states = value_states.view(*proj_shape)
+        input_shape = hidden_states.shape[:-1]
+        hidden_shape = (*input_shape, -1, self.head_dim)
 
-        src_len = key_states.size(1)
-        attn_weights = torch.bmm(query_states, key_states.transpose(1, 2))
+        queries = self.q_proj(hidden_states).view(*hidden_shape).transpose(1, 2)
+        keys = self.k_proj(hidden_states).view(*hidden_shape).transpose(1, 2)
+        values = self.v_proj(hidden_states).view(*hidden_shape).transpose(1, 2)
```
vasqu:
```suggestion
        query_states = self.q_proj(hidden_states).view(*hidden_shape).transpose(1, 2)
        key_states = self.k_proj(hidden_states).view(*hidden_shape).transpose(1, 2)
        value_states = self.v_proj(hidden_states).view(*hidden_shape).transpose(1, 2)
```
super nit: but that naming is just more standard across the library

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -736,11 +739,13 @@ def forward(
...
         super().__init__(config)
+        self.config = config
+        embed_dim = config.hidden_size
 
         self.embeddings = OwlViTVisionEmbeddings(config)
-        self.pre_layernorm = nn.LayerNorm(config.hidden_size, eps=config.layer_norm_eps)
+        self.pre_layernorm = nn.LayerNorm(embed_dim, eps=config.layer_norm_eps)
         self.encoder = OwlViTEncoder(config)
-        self.post_layernorm = nn.LayerNorm(config.hidden_size, eps=config.layer_norm_eps)
+        self.post_layernorm = nn.LayerNorm(embed_dim, eps=config.layer_norm_eps)
```
vasqu: This change seems not necessary? Or does it come from copies?

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -1214,7 +1206,7 @@ def image_text_embedder(
             num_patches_width = self.num_patches_width
 
         # Get image embeddings
-        last_hidden_state = outputs.vision_model_output[0]
+        last_hidden_state = outputs.vision_model_output.last_hidden_state
```
vasqu: This can break, no? We have no `can_return_tuple` decorator and if someone may pass return_dict=False, this will fail

Would rather revert these changes here at least

vasqu: And below

src/transformers/models/owlvit/modeling_owlvit.py
```diff
@@ -1410,8 +1402,8 @@ def image_guided_detection(
     @auto_docstring
     def forward(
         self,
-        input_ids: torch.Tensor,
         pixel_values: torch.FloatTensor,
+        input_ids: torch.Tensor | None = None,
```
vasqu: This seems breaking to me, any reason we need it?

src/transformers/models/owlv2/modeling_owlv2.py
vasqu, line 1: Same comments apply here so not mentioning things twice

vasqu (commented):
Got a few small details, but overall looks good! Thanks a lot for sticking with this, I really didn't make it easy for you as well

src/transformers/models/owlvit/modeling_owlvit.py
Aravind-11: Done, renamed to query_states/key_states/value_states.

src/transformers/models/owlvit/modeling_owlvit.py
Aravind-11: reverted.

src/transformers/models/owlvit/modeling_owlvit.py
Aravind-11: done

src/transformers/models/owlvit/modeling_owlvit.py
Aravind-11: Reverted to original signature.

src/transformers/models/owlv2/modeling_owlv2.py
Aravind-11: got it

src/transformers/models/owlvit/modeling_owlvit.py
Aravind-11: When I put input_ids back as the first required param, the main_input_name = "pixel_values" on OwlViTForObjectDetection no longer matched, causes the test failure.

Aravind-11: So i had to :

Remove main_input_name = "pixel_values" from the class 

Changed additional_model_inputs in the test from ["input_ids", "attention_mask"] to ["pixel_values", "attention_mask"] — since input_ids is now the main input, the test needs to provide pixel_values as an additional input

Aravind-11:
> Got a few small details, but overall looks good! Thanks a lot for sticking with this, I really didn't make it easy for you as well

haha, no worries!! thank u for helping out!!! :)))

vasqu:
run-slow: owlv2, owlvit

PR #42138, opened 2025-11-11
add new model internvl-3_5-Flash

### LINKED ISSUE #41862: Request for InternVL3_5_Flash
### Model description

InternVL3_5_Flash employ an additional router to use dynamic resolution of input images and this is not compatible with previous class "InternVLForConditionalGeneration". As a result, I would like to inquire whether it is possible to support the Flash version of InternVL3-5.

Image

### Open source status

- [x] The model implementation is available
- [x] The model weights are available

### Provide useful links for the implementation

https://huggingface.co/OpenGVLab/InternVL3_5-8B-Flash/blob/main/modeling_internvl_chat.py

Chenhao-Guan:
Resolves #41862

Hi @Rocketknight1,

Following our discussion in the issue, this PR adds support for the InternVL-Flash model variant.

As noted, the Flash variant uses a dynamic resolution router and a different image feature processing path, which was not compatible with the existing InternVLForConditionalGeneration.

Implementation Details
To solve this while following the modular style, I've implemented the following:

Added an is_flash_model: bool = False flag to the InternVLConfig.

Updated InternVLModel and InternVLForConditionalGeneration to use this flag (getattr(config, "is_flash_model", False)) to branch their logic for the __init__, get_image_features, and forward methods.

This implementation allows the same modeling_internvl.py file to support both the Base and Flash models, keeping the PR concise as you suggested.

Testing
All local tests are passing:

make style (style checks pass)

pytest  

Thank you for the guidance!

Before submitting
[ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

[x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request), Pull Request section?

[x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case. (Link: #41862)

[x] Did you make sure to update the documentation with your changes? No documentation changes needed.
[x] Did you write any new necessary tests? (Added tests/models/internvl/test_modeling_internvl.py and InternVLFlashModelTest)

Who can review?
@Rocketknight1 (handled the original issue) @zucchini-nlp (multimodal models)

### DIFF
--- a/src/transformers/models/internvl/modeling_internvl.py
+++ b/src/transformers/models/internvl/modeling_internvl.py
@@ -27,6 +27,7 @@ from typing import Optional, Union
 
 import torch
 import torch.nn as nn
+import torch.utils.checkpoint as cp
 
 from ...activations import ACT2FN
 from ...cache_utils import Cache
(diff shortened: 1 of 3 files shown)


zucchini-nlp (commented):
Hey @Chenhao-Guan 

Thanks for your work! We don't usually support several paths in a single model class. If we get a new model which is similar to an existing, the new model will need to be added as a separate class

You can benefit from modular as you did here, but you need to create a completely new model in its own folder. After you have it done, feel free to tag for a review

Chenhao-Guan:
Hi @zucchini-nlp,

Thank you for the clear feedback! I understand now—I should create a completely separate model, not use an if flag.

My main blocker before was a prefix conflict. When I first tried adding InternVLFlash classes inside the .../models/internvl/ folder, the modular_model_converter.py script failed because it saw both InternVL and InternVLFlash prefixes in the same file.

Your suggestion to "create a completely new model in its own folder" seems to be the perfect solution to this prefix problem.

Just to confirm: should I create a new folder like src/transformers/models/internvl_flash/ and place all the new files (e.g., modular_internvl_flash.py, config_internvl_flash.py, etc.) inside it, using only the InternVLFlash prefix?

If so, that makes perfect sense. I will start refactoring this right away.

Thanks!

zucchini-nlp:
> Just to confirm: should I create a new folder like src/transformers/models/internvl_flash/ and place all the new files (e.g., modular_internvl_flash.py, config_internvl_flash.py, etc.) inside it, using only the InternVLFlash prefix?

Yep, if you run `transformers add-new-model-like` and follow the prompts, you will get the files created for you including test files and docs

Chenhao-Guan:
Thanks @zucchini-nlp! Closing this in favor of a new PR that implements internvl_flash as a separate model, per your instructions.

PR #42145, opened 2025-11-11
Fix T5Gemma module structure

Cyrilvallez:
# What does this PR do?

As per the title. We never use direct inheritance from module to module.


Cyrilvallez:
run-slow: t5gemma

PR #42149, opened 2025-11-11
Fixes Flash Attention implementation for models 

i3hz:
# What does this PR do?
Fixes the issue where models use an outdated `if self.config._attn_implementation != "flash_attention_2":` check.

Models changed - SmolVLM, idefics3 , idefics2 

Fixes #42121 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@zucchini-nlp  @vasqu

### DIFF
--- a/src/transformers/models/idefics2/modeling_idefics2.py
+++ b/src/transformers/models/idefics2/modeling_idefics2.py
@@ -24,7 +24,7 @@ from torch import nn
 from ...activations import ACT2FN
 from ...cache_utils import Cache, DynamicCache
 from ...generation import GenerationMixin
-from ...masking_utils import create_bidirectional_mask
+from ...modeling_attn_mask_utils import _prepare_4d_attention_mask
 from ...modeling_flash_attention_utils import FlashAttentionKwargs
 from ...modeling_layers import GradientCheckpointingLayer
 from ...modeling_outputs import BaseModelOutput, ModelOutput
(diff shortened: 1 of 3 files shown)


src/transformers/models/idefics2/modeling_idefics2.py
```diff
@@ -502,10 +503,12 @@ def forward(
         # The call to `_upad_input` in `_flash_attention_forward` is expensive
         # So when the `patch_attention_mask` is full of 1s (i.e. attending to the whole sequence),
         # avoiding passing the attention_mask, which is equivalent to attending to the full sequence
```
vasqu: Let's remove these comments too

src/transformers/models/idefics2/modeling_idefics2.py
```diff
@@ -24,6 +24,7 @@
 from ...activations import ACT2FN
 from ...cache_utils import Cache, DynamicCache
 from ...generation import GenerationMixin
+from ...masking_utils import create_bidirectional_mask
 from ...modeling_attn_mask_utils import _prepare_4d_attention_mask
```
vasqu: Seems like `_prepare_4d_attention_mask` is still used and it will likely cause similar issues at other points, best to completely remove this usage elsewhere too!

See https://github.com/huggingface/transformers/blob/3ff0e69f84247ec8ed263ca412adc42f19fadbe2/src/transformers/models/idefics2/modeling_idefics2.py#L746-L750
(looks like the flag for that model there also needs to be updated to `_supports_flash_attn`)

vasqu (commented):
I'd like to fix idefics2 properly or leave it out for now, the others lgtm. Will check with `run-slow` a bit later

src/transformers/models/idefics2/modeling_idefics2.py
i3hz: so do I just reset all the changes from idefics2?

vasqu: Yea either that or use the `create_bidirectional_mask` fn here as well if it works; if not, I also appreciate that. Means I need to take a look here :D

i3hz: alright I can't run the testing script as I don't have enough vram but is this the correct approach?

i3hz:
tests seem to be failing so I don't think its the correct one

zucchini-nlp (approved):
LGTM! I think there were more models which still check for `flash_attention_2`, would be nice to batch update all. It can totally go in a separate PR later :)

zucchini-nlp:
run-slow: idefics3, smolvlm

vasqu (approved):
Thank you, let's keep it simple for now. We can address more models in separate PRs

vasqu:
> tests seem to be failing so I don't think its the correct one

Hmm it might need a dummy embedding of the correct size along the latents. Would leave this for a different PR

vasqu:
Also thx for all the PR :hugs:

i3hz:
>LGTM! I think there were more models which still check for flash_attention_2, would be nice to batch update all. It can totally go in a separate PR later :)

Yeah I didn't want to make a whole lot of changes in a single PR . I find it very confusing , sorry if that wasn't the ideal choice

PR #42156, opened 2025-11-12
🚨  Generalize `get_decoder()` for multimodal and delete redundant code 🔪 

zucchini-nlp:
# What does this PR do?

As per title, blocked by https://github.com/huggingface/transformers/pull/41589 for VLMs! We should be able to use `get_decoder()` to get the LM part of any model after this and have much less duplicate code. Same foes for the `get_encoder()` to get the encoder if the model has a separate encoding module. In comparison to decoder, we can have specific encoder per modality so the helper will accept `modality` as arg

Universal helper first reduces duplicate code, nudges us to use standardized names for major modules and can be used by 3rd party libraries. Right now we have 5 ways to name a vision encoder!

####  🚨 Breaking changes, ig we can break helpers for v5:
- VLMs now will not have a property to get `self.language_model` directly from task-model and users will need to call `self.get_decoder()`
- Deleted `get_text_encoder` and `get_audio_encoder` in some audio models because functionality is covered now by the `get_encoder()`


tests/utils/test_modeling_utils.py
```diff
@@ -3142,7 +3142,7 @@ def test_nested_wrapper_recursion(self):
         model = GPT2LMHeadModel(cfg)
         dec = model.get_decoder()
 
-        assert dec is model, f"GPT2 get_decoder() should return self (fallback), got {type(dec)}"
```
zucchini-nlp, line 3134: prev helper didn't cover all edge cases!  This should be the base model, if we compare with other LLMs (e.g. llama)

src/transformers/modeling_utils.py
```diff
@@ -2581,26 +2581,79 @@ def disable_input_require_grads(self):
...
+
+        # If this is a base transformer model (no encoder/model attributes), return self
+        return self
+
+    def set_encoder(self, encoder, modality: Optional[str] = None):
+        """
+        Symmetric setter. Mirrors the lookup logic used in `get_encoder`.
+        """
+
+        # NOTE: new models need to use existing names for layers if possible, so this list doesn't grow infinitely
```
molbap, line 2241: To note, this should be enforced in `make fixup` in code consistency part to save ourselves the hassle

src/transformers/modeling_utils.py
```diff
@@ -2581,26 +2581,79 @@ def disable_input_require_grads(self):
         """
         self._require_grads_hook.remove()
 
+    def get_encoder(self, modality: Optional[str] = None):
```
molbap, line 2208: usage-wise, how does this `modality` arg work? ie how do we use it concretely, should it match the processors signatures?

src/transformers/models/musicgen/modeling_musicgen.py
```diff
@@ -1956,7 +1940,7 @@ def _prepare_audio_encoder_kwargs_for_generation(
         self, input_values, model_kwargs, model_input_name: Optional[str] = None
     ):
         # 1. get audio encoder
-        encoder = self.get_audio_encoder()
+        encoder = self.get_encoder(modality="audio")
```
molbap, line 1930: ah I see now, ok, it's a good thing it's decoupled from the kwargs actually. So, the logic is 
- if I don't set the modality: defaults to text encoder
- If I set the modality, checks possible module names 
- If I override get_encoder it's whatever I want

Not 100% sure about the modality arg but I see the motivation, understood

src/transformers/modeling_utils.py
```diff
@@ -2581,26 +2581,79 @@ def disable_input_require_grads(self):
...
+        if modality == "audio":
+            possible_module_names = ["audio_tower", "audio_encoder"]
+        elif modality is None:
+            possible_module_names = ["text_encoder", "encoder"]
+        else:
+            raise ValueError(f'Unnrecognized modality, has to be "image", "video" or "audio" but found {modality}')
+
+        for name in possible_module_names:
+            if hasattr(self, name):
+                print(name)
```
molbap: I suppose
```suggestion
```

molbap (approved):
Very nice unbloating 🔪 
OK for me, just would be cool to add to the make style/ruff rules/quality check to reduce cognitive load

zucchini-nlp:
Merge conflicts after a big refactor 😢

src/transformers/models/musicgen/modeling_musicgen.py
zucchini-nlp: Yep, the idea is to allow users to ask for modality encoder through single interface, instead of having each as `self.get_vision_encoder/self.get_audio_encoder` etc.

Just added tests to make sure it works

src/transformers/modeling_utils.py
zucchini-nlp: hmm, isn't it going to be a huge limitation for contributors if we force it and auto-renam with fix-copies? Imo we need to communicate it when reviewing and explain why it's important. It's only a few ppl reviewing VLMs currently, so it might be easier

molbap: I was thinking the `make fixup` updated message (or rather code-quality check on the CI, same) would be informative enough, saying "decoder layer names should be part of this list: ..." rather than auto-renaming. Could be a ruff warning if we think it's too restrictive as an error?

zucchini-nlp: Hmm, lemme see where I can fit this in a non-disruptive way. Not sure if users actually read the warnings, we should be more strict in review process in any case imo 😆

zucchini-nlp:
hey @jackzhxng, I remember you requested this feature for `torch.export`. Now you can

```python
multimodal_model.get_decoder() -> returns the decoding LM
multimodal_model.get_encoder() -> returns the encoding LM if any
multimodal_model.get_encoder(modality="image") -> returns the encoding vision tower if any
multimodal_model.get_encoder(modality="audio") -> returns the encoding audio tower if any
```

also cc @hmellor, we also discussed it re vLLM

BenjaminBossan:
Hi @zucchini-nlp this PR causes an issue with PEFT as (at least some) decoder models now have `get_encoder`. As an example:

```python

model_id = "facebook/opt-125m"
assert not hasattr(model, "get_encoder")
# after this PR, model.get_encoder() returns model.model
```

This works with the previous commit (`a5c903f877fda21e739027eed133e03162eb7712`) but fails after this PR (`e2fb8d6062a05f69f976cf6e39618df6c31a3bfd`). Is this change intended?

jackzhxng:
@zucchini-nlp this is amazing thank you!

zucchini-nlp:
@BenjaminBossan ideally it should return `self` and not the base model. I see where it comes from and will fix it. Then you can check with smth like - `self.get_encoder != self: encoder = self.get_encoder()`

All models will have `get_encoder()` implemented, but if a model has no actual encoder it should return `self`

BenjaminBossan:
> I see where it comes from and will fix it.

Great, please let me know when the PR is there.

> All models will have `get_encoder()` implemented, but if a model has no actual encoder it should return `self`

We can modify PEFT to take this into account. But at least to me, this API feels a bit strange to be honest.

zucchini-nlp:
@BenjaminBossan yeah, it is because we have `get_encoder()` in PreTrainedModel so all models will have it as an attribute. Might feel weird though it is same as it was with `get_decoder()` and I think it saves us from duplicating the same code in all models

jackzhxng:
@zucchini-nlp I believe Voxtral actually still has top-level `language_model` - https://github.com/huggingface/transformers/blob/main/src/transformers/models/voxtral/modeling_voxtral.py#L379

Additionally woult it be possible for `get_decoder` to return the causal variant of the model which includes the `lm_head`?

zucchini-nlp:
Ah, Voxtral has a causal model as backbone. We need to explicitly overwrite `get_decoder` for the model in that case, will do

zucchini-nlp:
Oh, which includes the lm_head, I misread it!

Nah, the base model is supposed to be the model without any task head on top so in Voxtral, it is going to be the language model without a head

PR #42162, opened 2025-11-12
pin `pytest<9` for now

ydshieh:
# What does this PR do?

`pytest 9.0.0` is released 4 days ago, causing some problems in our slack reporting. 

There is `9.0.1` just released, let's try it.


ydshieh:
TODO: handle the new format  given by pytest 9

> = 1 failed, 103 passed, 84 skipped, 7 warnings, 16 subtests passed in 61.35s (0:01:01) =

PR #42164, opened 2025-11-12
Add slow to some examples tests 

SunMarc:
# What does this PR do?

This PR puts the slow tests to some e2e tests that takes > 1min on our CI. This should be fine as the CI never really failed because of those tests in my experience except due to flakiness.


ydshieh (approved):
Go! Merci beaucoup

ArthurZucker (approved):

PR #42166, opened 2025-11-12
add internvl_flash model

### LINKED ISSUE #41862: Request for InternVL3_5_Flash

Chenhao-Guan:
Resolves #41862

Hi @zucchini-nlp and @Rocketknight1,

Following your guidance in the issue, this PR re-implements the InternVL-Flash model as a completely separate model (instead of using an if flag in the existing InternVL class).

Implementation Details
Created a new, independent model directory: src/transformers/models/internvl_flash/.

Used the transformers add-new-model-like script to scaffold the new model, as you suggested.

Implemented the model logic in modular_internvl_flash.py (including Gating, CrossAttentionPooling, etc.) and converted it using the modular script.

Testing
All local tests are passing:

make fixup (style, quality, and repository consistency checks all pass)

pytest tests/models/internvl_flash/test_modeling_internvl_flash.py

Thank you for the guidance!

Before submitting
[ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

[x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request), Pull Request section?

[x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case. (Link: #41862)

[x] Did you make sure to update the documentation with your changes? (Added docs/source/en/model_doc/internvl_flash.md and updated _toctree.yml)

[x] Did you write any new necessary tests? (Added tests/models/internvl_flash/test_modeling_internvl_flash.py)

Who can review?
@zucchini-nlp @Rocketknight1


src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+    InternVLVisionPatchEmbeddings,
+    InternVLVisionPreTrainedModel,
+    InternVLVisionRMSNorm,
+)
+from ..llava.modeling_llava import (
+    LlavaForConditionalGeneration,
+)
+
+
+class Gating(nn.Module):
```
zucchini-nlp: `InternVLFlashGating` or similar naming since it's recommended to have model's name explicit in layer names

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+        self.block4 = mlp_block(hidden_size, mid_dim)
+        self.gate = nn.Sequential(nn.LayerNorm(hidden_size), nn.Linear(hidden_size, 2))  # 2 experts
+
+    def forward(self, x):
+        if self.use_checkpoint:
+            x = x + cp.checkpoint(self.block1, x)
+            x = x + cp.checkpoint(self.block2, x)
+            x = x + cp.checkpoint(self.block3, x)
+            x = x + cp.checkpoint(self.block4, x)
+        else:
```
zucchini-nlp: i don't think it, GC can be toggled on by `PreTrainedModel` if needed

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+        def mlp_block(in_dim, out_dim):
+            return nn.Sequential(
+                nn.Linear(in_dim, out_dim),
+                nn.GELU(),
+                nn.Dropout(dropout),
+                nn.Linear(out_dim, in_dim),
+                nn.Dropout(dropout),
+                nn.LayerNorm(in_dim),
+            )
+
```
zucchini-nlp: Separate layers are more preferred than a `Sequential`. Let's create an "nn.Module" with the following and call it smth like `InternVLFlashMLP`

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+        self.block1 = mlp_block(hidden_size, mid_dim)
+        self.block2 = mlp_block(hidden_size, mid_dim)
+        self.block3 = mlp_block(hidden_size, mid_dim)
+        self.block4 = mlp_block(hidden_size, mid_dim)
+        self.gate = nn.Sequential(nn.LayerNorm(hidden_size), nn.Linear(hidden_size, 2))  # 2 experts
```
zucchini-nlp: same here, let's have them separated

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+            x = x + self.block1(x)
+            x = x + self.block2(x)
+            x = x + self.block3(x)
+            x = x + self.block4(x)
+        logits = self.gate(x)  # shape: [B, 2]
+        probs = torch.softmax(logits, dim=-1)  # 每个 token 的 expert 选择概率
+        return probs
+
+
+class CrossAttentionPooling(nn.Module):
```
zucchini-nlp: same comment for naming

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+    def __init__(self, dim, num_heads=16):
+        super().__init__()
+        self.query_token = nn.Parameter(torch.randn(1, dim))  # [1, D]
+        self.attn1 = nn.MultiheadAttention(embed_dim=dim, num_heads=num_heads, batch_first=True)
```
zucchini-nlp: Using transformers-style attention layers will support more attention implementations, than `torch.nn.MultiheadAttention`. Let's add an attention layer following `InternVLTextAttention` format and use it here instaed

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+class InternvlFlashMultiModalProjector(InternVLMultiModalProjector):
+    pass
+
+
+class InternvlFlashModelOutputWithPast(InternVLModelOutputWithPast):
+    pass
+
+
+class InternvlFlashModel(InternVLModel):
+    _checkpoint_conversion_mapping = {"language_model.model": "language_model"}
```
zucchini-nlp: `_checkpoint_conversion_mapping` should not be needed for newly released weights if they are converted correctly. Do you know how InternVLFlash state dict in serialized in the hub currently?

We can keep it only if the weights are released on hub and can't be loaded without `_checkpoint_conversion_mapping`

zucchini-nlp:
Taking a look tomorrow-Monday, thanks for making a new model class

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+    pass
+
+
+class InternvlFlashVisionPreTrainedModel(InternVLVisionPreTrainedModel):
+    pass
+
+
+@auto_docstring
+class InternvlFlashVisionModel(InternVLVisionModel):
+    pass
```
zucchini-nlp, line 230: we can keep only this one and delete the prev classes starting from `InternvlFlashVisionRMSNorm`, modular will take care of dependencies

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+class InternvlFlashPreTrainedModel(InternVLPreTrainedModel):
+    pass
+
+
+class InternvlFlashMultiModalProjector(InternVLMultiModalProjector):
+    pass
+
+
+class InternvlFlashModelOutputWithPast(InternVLModelOutputWithPast):
+    pass
```
zucchini-nlp: can delete as well, modular will copy it anyway as long as the base class is a copy from "InternVL"

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+
+
+class InternvlFlashModel(InternVLModel):
+    _checkpoint_conversion_mapping = {"language_model.model": "language_model"}
+
+    def __init__(self, config: InternvlFlashConfig):
+        super().__init__(config)
+
+        if getattr(config, "is_flash_model", False):
```
zucchini-nlp: is it needed to support official checkpoints or is that a flag for experiments? We'd prefer to delete code path if not needed to support released models
Less unused code -> better readability and easier to maintain

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+            self.mlp2 = nn.Sequential(
+                nn.LayerNorm(vit_hidden_size * int(1 / config.downsample_ratio) ** 4),
+                nn.Linear(vit_hidden_size * int(1 / config.downsample_ratio) ** 4, llm_hidden_size * 2),
+                nn.GELU(),
+                nn.Dropout(0.1),
+                nn.Linear(llm_hidden_size * 2, llm_hidden_size * 2),
+                nn.GELU(),
+                nn.Dropout(0.1),
+                nn.Linear(llm_hidden_size * 2, llm_hidden_size),
+            )
```
zucchini-nlp: same as before, let's wrap this as an `nn.Module` class and int from config

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+        new_input_embeds = input_embeds[keep_mask.to(input_embeds.device), :]
+        new_input_ids = input_ids[keep_mask.to(input_ids.device)]
+
+        return new_input_embeds, new_input_ids, updated_mask_idx, keep_mask
+
+    def get_image_num_per_sample(
+        self,
+        input_ids: torch.Tensor,
+    ):
+        input_ids = input_ids.squeeze(0)  # (N,)
```
zucchini-nlp: won't work with batch size more than 1

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+        """
+        # 拆分 features → 每个样本一个 tile list
+        tile_splits = torch.split(features, split_sizes, dim=0)
+
+        # 合并前两维：tile * 1024 × 1024
+        merged = [x.reshape(-1, x.shape[-1]) for x in tile_splits]
+
+        return merged
+
+    def get_image_features_flash(
```
zucchini-nlp: simply `get_image_features`, we should not need non-flash code path

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+
+        Args:
+            pixel_values (`torch.FloatTensor]` of shape `(batch_size, channels, height, width)`)
+               The tensors corresponding to the input images.
+            vision_feature_layer (`int` or `list[int]`):
+                Layer index or list of layer indices to extract features from.
+        Returns:
+            vision_features (`torch.Tensor`): Image feature tensor of shape `(num_images, image_length, embed_dim)`.
+        """
+        with torch.no_grad():
```
zucchini-nlp: not needed, some users might want to fine-tune and we will need grads over vision tower in that case

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+            vit_embeds_1024 = self.vision_tower(pixel_values=pixel_values.to(self.dtype)).last_hidden_state
```
zucchini-nlp: `self.dtype` isn't veryy representative, because dtypes can be different for vision and text models. Let's just pass as is and let users to set dtype themselves

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+
+        vit_embeds_1024 = vit_embeds_1024[:, 1:, :]
```
zucchini-nlp, line 371: depends on `vision_feature_select_strategy`? Or do we want to always crop CLS token, in which case we need to get rid of `vision_feature_select_strategy` everywhere

src/transformers/models/internvl_flash/modular_internvl_flash.py
zucchini-nlp: same here for `vision_feature_layer`, either we use it or delete from config if it's always the last layer

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+
+        returns: List of Tensors of shape [tile_i * 1024, 1024]
+        """
+        # 拆分 features → 每个样本一个 tile list
+        tile_splits = torch.split(features, split_sizes, dim=0)
+
+        # 合并前两维：tile * 1024 × 1024
+        merged = [x.reshape(-1, x.shape[-1]) for x in tile_splits]
+
+        return merged
```
zucchini-nlp: two-liner does not need a separate fn, let's move it to where it's used

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+        padded = torch.cat(
+            [torch.tensor([0], device=selected.device), selected.int(), torch.tensor([0], device=selected.device)]
+        )
+        diff = torch.diff(padded)
+
+        starts = (diff == 1).nonzero(as_tuple=True)[0]
+        ends = (diff == -1).nonzero(as_tuple=True)[0]
+        lengths = ends - starts
+
+        return lengths
```
zucchini-nlp: looks quite complicated to get the length of each image feature. Why can't we use simply `input_ids == self.config.image_token_id` and sum over it? If images are processed with different sizes, we can return from processor `image_sizes` as we do with other models

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+                )
+
+                selected_embeds = []
+                for i in range(gate_result.size(0)):
+                    if gate_result[i]:
+                        selected_embeds.append(vit_embeds_64[i])
+                    else:
+                        selected_embeds.append(vit_embeds_256[i])
+
+                vit_embeds = torch.cat(selected_embeds, dim=0)
```
zucchini-nlp: let's move this block to `get_image_features` as it is still part of obtaining ready image hidden states for VLM

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+
+                assert torch.all(attention_mask == 1)
```
zucchini-nlp: can't support padded inputs?

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+        selected = input_ids == img_context_token_id
+        padded = torch.cat(
+            [torch.tensor([0], device=selected.device), selected.int(), torch.tensor([0], device=selected.device)]
+        )
+        diff = torch.diff(padded)
+
+        starts = (diff == 1).nonzero(as_tuple=True)[0]
+        ends = (diff == -1).nonzero(as_tuple=True)[0]
+        lengths = ends - starts
+
```
zucchini-nlp: we already computed `length` once, no? Can re-use it instead of re-computing

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+        keep_mask = torch.ones(N, dtype=torch.bool, device=input_embeds.device)
+
+        delete_flags = torch.zeros(N, dtype=torch.int32, device=input_embeds.device)
+
```
zucchini-nlp, line 299: oh, deciding on number of image tokens while models makes this part hard 🫠

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+                    mask_idx=attention_mask,
+                    img_context_token_id=self.config.image_token_id,
+                    gate_result=gate_result,
+                )
+
+                attention_mask = torch.ones(1, inputs_embeds.shape[0]).to(inputs_embeds.device)
+
+                selected = input_ids == self.config.image_token_id
+                assert selected.sum() != 0
+                inputs_embeds[selected] = vit_embeds.to(inputs_embeds.device)
```
zucchini-nlp: let's try to use the standard API with helper fn for getting the mask and scattering

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+                block_start = s + i * 256
+                block_end = block_start + 256
+
+                compress = gate_result[flag_idx]
+                flag_idx += 1
+
+                if compress:
+                    keep_mask[block_start + 64 : block_end] = False
+                    delete_flags[block_start + 64 : block_end] = 1
+
```
zucchini-nlp: could it be vectorized?

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,586 @@
...
+        cumulative_deletes = torch.cumsum(delete_flags, dim=0)
+        cumulative_deletes = torch.cat([cumulative_deletes, cumulative_deletes[-1:].clone()], dim=0)
+
+        mask_idx = mask_idx.squeeze(0)
+        updated_mask_idx = mask_idx - cumulative_deletes[mask_idx.to(cumulative_deletes.device)].to(mask_idx.device)
+        updated_mask_idx = updated_mask_idx.unsqueeze(0)
```
zucchini-nlp: new attention mask is not used as I can see from current commit, so we have to fix it first. Then, why can't we `attention_mask = attention_mask[keep_mask]`

tests/models/internvl_flash/test_modeling_internvl_flash.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+if is_torch_available():
+    import torch
+
+    from transformers import InternvlFlashForConditionalGeneration, InternvlFlashModel
+
+
+if is_vision_available():
+    pass
```
zucchini-nlp: dummy import, can be deleted

tests/models/internvl_flash/test_modeling_internvl_flash.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def setUp(self):
+        self.model_tester = InternvlFlashVisionText2TextModelTester(self)
+        self.config_tester = ConfigTester(self, config_class=InternvlFlashConfig, has_text_modality=False)
+
+    @unittest.skip(
+        reason="Failing with `torch._inductor.exc.InductorError: RuntimeError: No valid triton configs. OutOfMemoryError: out of resource: triton_tem_fused_0 Required: 147456 Hardware limit:101376 Reducing block sizes or `num_stages` may help.`"
+    )
+    def test_flex_attention_with_grads(self):
+        pass
```
zucchini-nlp: can you delete skip on it? Prob it was failing on your local hardware, might pass with CI runners

tests/models/internvl_flash/test_modeling_internvl_flash.py
```diff
@@ -0,0 +1,219 @@
...
+    def test_sdpa_can_compile_dynamic(self):
+        pass
+
+    @unittest.skip("FlashAttention only support fp16 and bf16 data type")
+    def test_flash_attn_2_fp32_ln(self):
+        pass
+
+    @unittest.skip("Skipping compilation test: fails with batch_size=0 reshape error")
+    def test_generate_compile_model_forward_fullgraph(self):
+        pass
```
zucchini-nlp, line 210: we need to support any batch size for the model before merging

zucchini-nlp (commented):
Hey @Chenhao-Guan , thanks for making a separate PR for the model!

I have a few major comments:
1. The model doesn't seem to support batch size > 1 currently. Prob the official release works with a single batch size which is oke. However we need to enable batched inference before merging this PR
2. I see that you kept 'flash" and "non-flash" paths. We have to delete the "non-flash" code path as it is not needed to run with InternVLFlash released checkpoint 
3. A few minor issues like naming and sticking to transformers standards 👇🏻

Chenhao-Guan:
@zucchini-nlp Thank you for the advice. This is my first time submitting a PR, and Gonna working to resolve the test failures related to batch_size > 1 support. My initial intention in adding non-Flash methods was specifically to bypass these failing tests temporarily. I will continue working to implement a full solution.

src/transformers/models/internvl_flash/modular_internvl_flash.py
Chenhao-Guan: The vision_feature_select_strategy parameter is currently unused (it was previously introduced to support the now-deprecated non_flash_method).

src/transformers/models/internvl_flash/modular_internvl_flash.py
Chenhao-Guan: Gonna test it out.That's the offical implemention,find it confusing too

src/transformers/models/internvl_flash/modular_internvl_flash.py
Chenhao-Guan: Already compute in get_image_num_per_sample. To avoid redundant computation, the logic of get_image_num_per_sample should be inlined into the forward method. Will it be too redundent?

zucchini-nlp: If we can make the code for computing num of images more readable, we can keep it in forward. We can also try to merge these fns or simply pass over computed length to each fn separately. Whichever looks better after cleaning up

Chenhao-Guan:
@zucchini-nlp I've finished the requested modifications. Please let me know if there are any other points to discuss before we merge.

zucchini-nlp:
Thanks, I will review some time this week. It was a bit hectic due to v5 last week

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+        hidden_states = self.act_fn(hidden_states)
+        hidden_states = self.dropout_in(hidden_states)
+        hidden_states = self.dense_out(hidden_states)
+        hidden_states = self.dropout_out(hidden_states)
+        hidden_states = self.norm(hidden_states)
+
+        return hidden_states
+
+
+class InternvlFlashMLP2(nn.Module):
```
zucchini-nlp: nit: `InternvlFlashMultimodalProjection` for explicitness and to not confuse with MLP layer above

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+    def __init__(self, vit_hidden_size, llm_hidden_size, config):
+        super().__init__()
+
+        in_dim = vit_hidden_size * int(1 / config.downsample_ratio) ** 4
+        mid_dim = llm_hidden_size * 2
+        out_dim = llm_hidden_size
```
zucchini-nlp: vit and llm hidden sizes are available in config, so we dont need to pass those separately

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+        hidden_states = self.dropout2(hidden_states)
+        hidden_states = self.dense3(hidden_states)
+
+        return hidden_states
+
+
+class InternvlFlashGating(nn.Module):
+    def __init__(self, hidden_size=2048, expansion_factor=4, dropout=0.1, use_checkpoint=True):
+        super().__init__()
+        self.use_checkpoint = use_checkpoint
```
zucchini-nlp: attribute not used?

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+
+    def forward(self, batched_tokens: list[torch.Tensor]):
+        """
+        batched_tokens: List of Tensors of shape [Ti, D], length = B
+        """
+        B = len(batched_tokens)
+        if B == 0:
+            return torch.empty(
+                0, self.query_token.shape[-1], device=self.query_token.device, dtype=self.query_token.dtype
+            )
```
zucchini-nlp: i think it cannot be `0` unless there is a bug in code?

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+        **kwargs,
+    ):
+        """
+        Obtains image last hidden states from the vision tower and apply multimodal projection.
+
+        Args:
+            pixel_values (`torch.FloatTensor]` of shape `(batch_size, channels, height, width)`)
+               The tensors corresponding to the input images.
+            vision_feature_layer (`int` or `list[int]`):
+                Layer index or list of layer indices to extract features from.
```
zucchini-nlp: nit: the docstring doesn't match signature

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+        return new_input_embeds, new_input_ids, keep_mask
+
+    def get_image_num_per_sample(
+        self,
+        input_ids: torch.Tensor,
+    ):
+        if input_ids is None:
+            raise ValueError("input_ids cannot be None when pixel_values are provided. ")
+        if input_ids.dim() == 1:
+            input_ids = input_ids.squeeze(0)  # (N,) #todo add batch size support
```
zucchini-nlp: todo?

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+
+        vit_embeds_256 = vit_embeds_1024.clone()
+
+        vit_embeds_64 = self.pixel_shuffle(vit_embeds_1024, scale_factor=self.config.downsample_ratio**2)
+        vit_embeds_64 = vit_embeds_64.reshape(vit_embeds_64.shape[0], -1, vit_embeds_64.shape[-1])
+        vit_embeds_64 = self.mlp2(vit_embeds_64)
+
+        vit_embeds_256 = self.pixel_shuffle(vit_embeds_256, scale_factor=self.config.downsample_ratio)
+        vit_embeds_256 = vit_embeds_256.reshape(vit_embeds_256.shape[0], -1, vit_embeds_256.shape[-1])
+        vit_embeds_256 = self.multi_modal_projector(vit_embeds_256)
```
zucchini-nlp: pixel shuffle could be part of the MLP module. So when we call `self.mlp(images)`, we'll get already projected/reshaped outputs

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+        self,
+        inputs_embeds: torch.Tensor,
+        input_ids: torch.Tensor,
+        vit_embeds: torch.Tensor,
+    ) -> torch.Tensor:
+        selected_mask = input_ids == self.config.image_token_id
+        if selected_mask.sum() == 0:
+            return inputs_embeds
+        inputs_embeds[selected_mask] = vit_embeds.to(inputs_embeds.device)
+        return inputs_embeds
```
zucchini-nlp, line 425: needs to support input embeddings, similar to other VLMs with `gte_placeholder_mask`

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+            padding_mask[i, :L] = False
+        # 2. Query token: [B, 1, D]
+        query = self.query_token.unsqueeze(0).expand(B, -1, -1)  # learnable token for each sample
+
+        attention_mask = torch.zeros_like(padding_mask, dtype=query.dtype)
+        min_value = torch.finfo(query.dtype).min
+        attention_mask.masked_fill_(padding_mask, min_value)
+
+        # 3. Adjust Attention Score: [B, Num_Heads, Q_Len, K_Len]
+        attention_mask = attention_mask.unsqueeze(1).unsqueeze(1)
```
zucchini-nlp: hmm interesting. Ig we could do attention with ragged input though I don't see much difference tbh. Let's just call `create_bidirection_mask` to get a 4D projected mask (e.g. in BERT-like models they do it)

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+                lengths=lengths_copy,
+                starts=global_starts,
+            )
+            if isinstance(attention_mask, dict):  # add support for StaticCache
+                attention_mask = attention_mask["full_attention"]
+
+            if attention_mask is not None:
+                if attention_mask.dim() > 2 or attention_mask.numel() != input_ids.numel():
+                    pad_token_id = self.config.pad_token_id if self.config.pad_token_id is not None else 0
+                    attention_mask = input_ids.ne(pad_token_id)
```
zucchini-nlp, line 519: I don't think it is a good idea to reconstruct a mask when a 4D mask is provided. We still can slice it out with keep mask, as long as the mask is broadcasted ,no ?

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+            inputs_embeds, input_ids, keep_mask = self.compress_visual_tokens_in_sentence(
+                input_embeds=inputs_embeds,
+                input_ids=input_ids,
+                img_context_token_id=self.config.image_token_id,
+                gate_result=gate_result,
+                lengths=lengths_copy,
+                starts=global_starts,
+            )
+            if isinstance(attention_mask, dict):  # add support for StaticCache
+                attention_mask = attention_mask["full_attention"]
```
zucchini-nlp: so the model has sliding and full attention, iiuc. In which case we either slice both masks or disallow preparing 4D masks (see Bloom for that, it has special postiion embedding which requires 2D mask)

src/transformers/models/internvl_flash/modular_internvl_flash.py
```diff
@@ -0,0 +1,580 @@
...
+
+            inputs_embeds, attention_mask = self._reconstruct_batch(
+                inputs_embeds=inputs_embeds,
+                attention_mask=attention_mask,
+                gate_result=gate_result,
+                lengths=lengths,
+                batch_indices=batch_indices,
+                N=N,
+                B=B,
+            )
```
zucchini-nlp, line 540: sorry, not clear to me why we reconstruct the batch again, after it was sliced prev to keep only necessary image placeholder tokens?

zucchini-nlp (commented):
Hey, sorry it took long to review. The model arch is more complicated than in other VLMs due to the attention-based gating and different image lengths

I think we can keep for now the approach you have, by slicing out unused image placeholder tokens. I left a few comments on better readability below, let me know if anyrhing isn't clear

Chenhao-Guan:
Thank you for your review! Recently working on my final. Hoping to slove this problems ASAP.

zucchini-nlp:
No problem, take your time :)

YanxingLiu:
> Thank you for your review! Recently working on my final. Hoping to slove this problems ASAP.

hello, is there any progress? I am the author of the issue https://github.com/huggingface/transformers/issues/41862.  I have free time currently,  may I do something for help the PR?

YanxingLiu:
I am tring to convert raw internvl-3.5-flash model to huggingface format by modifing https://github.com/OpenGVLab/InternVL/blob/main/internvl_chat/tools/internvl_custom2hf.py . I find that there are 2 additional params exist in InternvlFlashConfig but not in InternVLConfig. Is it possible to add these two parameters in InternvlFlashConfig (although even without these two parameters, they can still be manually added in config.json to achieve the conversion)?
image

Chenhao-Guan:
@YanxingLiu Sorry for not checking my mail.I use scripts for internvl to convert the models with some minor changes. Models can be found at( https://huggingface.co/chenhaoguan/InternVL3_5-2B-Flash-hf ). I have mail u my version. I haven't made any changes. Free to ask me if u have any question.

src/transformers/models/internvl_flash/modular_internvl_flash.py
Chenhao-Guan: the "slicing" earlier removed unnecessary visual tokens within the flattened representation,  _reconstruct_batch is needed to reorganize the data back into the proper (B, N, C) batch format the LLM expects, with uniform padding for the new variable lengths.

Chenhao-Guan:
Sorry it takes so long @zucchini-nlp . I’ve addressed most of the issues you pointed out. However, in batch processing, sequences are padded to the maximum length of the batch making it not that efficient, and I feel there must be a more efficient way to handle this.

PR #42170, opened 2025-11-12
[`Pop2Piano`] Fix cache usage

vasqu:
I don't want to speak about his :see_no_evil:


ydshieh (approved):
I don't want to speak about this neither 🙈. Thank you for taking care the past.

(but why it worked before, prior your merged PR ..?)

vasqu:
The speed difference comes from using the cache (or not afterwards). The headmasking PR mistakenly outcommented passing the cache.

PR #42176, opened 2025-11-13
Remove import to deperated file

yinpeiqi:
the `tokenization_bert_tf.py` file is already removed, while the import in TYPE_CHECKING is still remain.


Rocketknight1:
Ah, thank you for the PR! I made #42182 before I realized you'd made this one. I'm going to merge #42182 instead because it also cleans up the reference in `check_docstrings.py`, but thank you for spotting the bug in the first place!

PR #42185, opened 2025-11-13
[`PEFT`] Fix the general test for prefix tuning

vasqu:
We now require exact masks! The test was utilizing the fact that we lazily cut masks cc @Rocketknight1


Rocketknight1 (approved):
Nice, thank you for the quick fix!

PR #42186, opened 2025-11-13
Remove mask slicing in all eager attentions

Cyrilvallez:
# What does this PR do?

As per the title. Following https://github.com/huggingface/transformers/pull/41900.

The mask is (and should!!) be correctly prepared, with the correct shape. If not, then it0s better to crash immediately, as otherwise this leads to very silent bugs!!


justinchuby:
Was this able to move forward? Thanks!

Cyrilvallez:
Hey @justinchuby! This has been on standby a bit as a lot of errors popped up, because some old models do not create their masks correctly... 
We are right now removing old mask preparation primitives from every model, so once this is done this PR should be able to land immediately!

Cyrilvallez:
Finally ready now that https://github.com/huggingface/transformers/pull/42848 was merged

src/transformers/models/bloom/modeling_bloom.py
```diff
@@ -274,7 +274,7 @@ def forward(
 
         # change view to [batch_size, num_heads, q_length, kv_length]
         attn_weights = attention_scores.view(batch_size, self.num_heads, q_length, -1)
-        if attention_mask is not None:  # no matter the length, we just slice it
+        if attention_mask is not None:
             causal_mask = attention_mask[:, :, :, : key_layer.shape[-1]]
```
vasqu: Oops :D

src/transformers/models/diffllama/modeling_diffllama.py
```diff
@@ -255,7 +255,7 @@ def forward(
 
         attn_weights = torch.matmul(query_states, key_states.transpose(2, 3)) / math.sqrt(self.head_dim)
 
-        if attention_mask is not None:  # no matter the length, we just slice it
+        if attention_mask is not None:
             causal_mask = attention_mask[:, :, :, : key_states.shape[-2]]
```
vasqu: Same

src/transformers/models/diffllama/modular_diffllama.py
```diff
@@ -131,7 +131,7 @@ def forward(
 
         attn_weights = torch.matmul(query_states, key_states.transpose(2, 3)) / math.sqrt(self.head_dim)
 
-        if attention_mask is not None:  # no matter the length, we just slice it
+        if attention_mask is not None:
             causal_mask = attention_mask[:, :, :, : key_states.shape[-2]]
```
vasqu: Ok mentioning it a last time lol

vasqu (commented):
Wdyt about using 2 sources (because some do fp32 softmax and some don't) for eager to unify this? Talking about those that are not using modular atm --> use copied from to ensure we only modify these

Just a thought which might make maintenance easier + I don't have a good overview on which are independent atm

src/transformers/models/bloom/modeling_bloom.py
Cyrilvallez: Damn nice catch! This one thought it could keep on living... 😈

Cyrilvallez:
I agree that in general it would be very nice to unify a bit more the eager implementations which are all the same but use different variable names etc for the same things.
This feels a bit outside the scope of this PR which touches a LOT of files already though, and which is focused on removing a useless and dangerous slicing ops - would prefer to do it later on to avoid overcomplicating this one

vasqu (approved):
Yup, it's fine that way - wanna run a few slow tests (run-slow) on important models for safety? Otherwise lgtm

Cyrilvallez:
run-slow: bert, gemma2, llama, mistral, mixtral

PR #42188, opened 2025-11-13
[`BLT`] Fix cache usage

vasqu:
The issue is/was that the cache wasn't utilized at all while several generation methods rely on it. This resulted in a mismatch between actual input and attention mask
- the mask was correctly expanded one by one
- after the prefill phase, the cache wasn't used resulting in kv always being of size 1

Why did it even work?
- because we silently cut the mask to fit the kv shape, i.e. a silent bug and another good reason why enforcing the correct shape is highly needed

The fix:
- apply self caches for both the encoder and decoder

Future considerations:
- add cross cache for both encoder + decoder --> not used atm (but also wasn't used before as far as I can see)


vasqu:
run-slow: blt

vasqu:
run-slow: blt

vasqu:
run-slow: blt

vasqu:
run-slow: blt

vasqu:
Fixes the normal fast tests; weird issue atm on main where it takes significantly more memory / it's not properly cleaned up. Locally the slow tests pass except for those that already didnt

zucchini-nlp (approved):
Very interesting model! Approving since the PR fixes CI

I think it's fine if we don't have cache on cross attention module, and instead add a cache for Global Transformers. I skimmer over the models' tech report and seems that global transformer is a much bigger module than local encoder-decoders.

ArthurZucker (approved):
thanks

PR #42191, opened 2025-11-13
Much more efficient and clear weight initialization and tie weights

Cyrilvallez:
# What does this PR do?

Follow-up to https://github.com/huggingface/transformers/pull/41580. 
This PR is focused on speed and efficiency, as well as clarity. It avoids the following issues/bottlenecks:

- No longer need to switch dynamically the class of the Parameters (which is hard to understand and may lead to issues with quantization etc, as it's not always clear if the memory stays the same or if we double it)
- Avoids one more loop over all the modules to switch class
- Avoids initializing the tied weights (which takes a lonnnngggg time for big embeddings, e.g for `"google/gemma-2-2b"`, the embedding has size `256k * 2304`, which is more than 1GB of data in float16 in a single parameter, and it takes more than 3-4s to initialize it with `normal_`, see profiling below), when we will overwrite them anyway later (with the tied weights)
- avoids one more loop over all modules for tied weights (get the correct names in advance)

Screenshot 2025-11-14 at 10 43 53

On that trace, the call to `Parameter.normal_` takes 3.4s, and is only the `lm_head` (the only "missing" weight), even though it's not truly missing because it's a tied weight which is being overwritten later!

So basically, the following snippet 

```python
import torch
import time

model_id = "google/gemma-2-2b"
device = 1 

torch.cuda.synchronize()
t0 = time.time()
model = AutoModelForCausalLM.from_pretrained(model_id, device_map=device, dtype="auto")
torch.cuda.synchronize()
dt = time.time() - t0
print(f"Took {dt:.2f} s")
```

takes about 7s, when it was taking about 3s on main before https://github.com/huggingface/transformers/pull/41580 /with the old loading). After this PR, it takes 3s as well, effectively being as performant as before.


ArthurZucker (commented):
very nice, make sure to rebase + models with specific init weight i think are like https://github.com/huggingface/transformers/blob/9ccb6935fcc24221988464b455a12631f686774a/src/transformers/models/rt_detr_v2/modeling_rt_detr_v2.py#L475 this one!

src/transformers/initialization.py
```diff
@@ -0,0 +1,180 @@
...
+
+@contextmanager
+def guard_torch_init_functions():
+    """
+    Guard the `torch.nn.init` primitive functions to behave exactly like the functions in this file, i.e. be
+    protected against the `_is_hf_initialized` flag to avoid re-init if the param was already loaded.
+
+    Usually, all models are using the init from `transformers` which are already guarded, but just to make extra sure
+    and for remote code, we also use this context manager.
+    """
```
ArthurZucker, line 173: this won't work for any tensor manipulation for any remote code / code outside our scope, but its ~fine~

src/transformers/modeling_utils.py
```diff
@@ -4625,11 +4577,15 @@ def _adjust_missing_and_unexpected_keys(
         if ignore_unexpected_regex is not None:
             unexpected_keys = {key for key in unexpected_keys if ignore_unexpected_regex.search(key) is None}
 
-        # Note: only the unexpected keys should remove the added prefix here, to correctly display the original name
-        # in the warnings. For missing keys, we should show the prefix in the warning as it's part of the final model
-        if loading_task_model_from_base_state_dict:
-            _prefix = f"{self.base_model_prefix}."
-            unexpected_keys = {k.removeprefix(_prefix) for k in unexpected_keys}
+        # Set the flag (very important to avoid initializing them!!)
+        for tied_param in self._tied_weights_keys.keys():
```
ArthurZucker: only if `tie_words_embeddings` or `tie_encoder_decoder`

ArthurZucker: ah ok they only exist if you have tie_wwords embeddings. But you need in that case a set and get

ArthurZucker (commented):
In general good, you need to overwrite the get and set for `tie_words_embeddings` to update `tied_weight_keys` potentially.

src/transformers/initialization.py
Cyrilvallez: Yes I know, this is very unfortunate but we cannot really make it work for remote code 🥲

src/transformers/modeling_utils.py
Cyrilvallez: Yes they are set correctly in advance in post_init

ArthurZucker (approved):
Big bird failing IS related, let's fix. 
Also can you update https://github.com/huggingface/transformers/blob/082e3ff4a3cfaeffac38812a3a68e9a59e4d0a5a/utils/check_init_weights_data.py#L11 please.

Also add a big TODO for the dynamic part I think it is important!

PR #42192, opened 2025-11-13
Stop inheriting tests!

Rocketknight1:
When we import a `pytest` test class like `BambaModelTest` into another file to inherit from, this creates a reference to the parent class in the new file. When we run `pytest new_file.py`, both the new child test class **and** the parent class are run. This was causing the `bamba` tests to be run in the `granitemoehybrid` test file and the `cohere` tests to be run in the `cohere2` test file

This PR copies the test code into the new file, replacing the overridden attributes and methods as required, so that there's no reference to the parent class anywhere.

cc @ydshieh

### DIFF
--- a/tests/models/cohere2/test_modeling_cohere2.py
+++ b/tests/models/cohere2/test_modeling_cohere2.py
@@ -34,8 +34,7 @@ from transformers.testing_utils import (
     torch_device,
 )
 
-from ...models.cohere.test_modeling_cohere import CohereModelTest, CohereModelTester
-from ...test_configuration_common import ConfigTester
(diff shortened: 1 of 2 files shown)


Rocketknight1:
run-slow: cohere2, granitemoehybrid

ydshieh:
Yes, I agree! Will check the changes but in principle yes

ydshieh (approved):
As talked internally, putting 

del xxxTest (the imported one)

at the end of the test file would work

But works for me if you want to go this current version.


Thank you!

ArthurZucker (approved):

Rocketknight1:
@ydshieh using the `del` solution instead, since it's a lot simpler

PR #42193, opened 2025-11-13
[`Pop2Piano`] Fix tied weights

vasqu:
Turns out that the weights were a 2way tie and not 3way as indicated by the original key:
- Only encoder + decoder --> the lm heads are separate

My investigation came originally from the increased runtime in a test, turns out it is due to different weights and hence generation ran longer (longer sequence generated). And the profiler showed the most time was spent on the cache, which is kinda expected as we have to constantly torch cat during the rollout generation.

Fixed by correctly specifying the key and forcing tied weights keys as in bart. I checked with the first 2 integration tests that now pass (the last head version issues and I'm too lazy to resolve them)

NOTE:
It is likely that some other models are also affected by this as it's a mixture of what ties were really expected.

### DIFF
--- a/src/transformers/models/pop2piano/modeling_pop2piano.py
+++ b/src/transformers/models/pop2piano/modeling_pop2piano.py
@@ -583,6 +583,8 @@ class Pop2PianoPreTrainedModel(PreTrainedModel):
             module.o.weight.normal_(mean=0.0, std=factor * ((n_heads * key_value_proj_dim) ** -0.5))
             if module.has_relative_attention_bias:
                 module.relative_attention_bias.weight.normal_(mean=0.0, std=factor * ((d_model) ** -0.5))
+        elif isinstance(module, nn.Linear):
+            module.weight.normal_(mean=0.0, std=factor)
 
     def _shift_right(self, input_ids):
         decoder_start_token_id = self.config.decoder_start_token_id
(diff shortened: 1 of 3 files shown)


src/transformers/models/pop2piano/modeling_pop2piano.py
```diff
@@ -583,6 +583,8 @@ def _init_weights(self, module):
             module.o.weight.normal_(mean=0.0, std=factor * ((n_heads * key_value_proj_dim) ** -0.5))
             if module.has_relative_attention_bias:
                 module.relative_attention_bias.weight.normal_(mean=0.0, std=factor * ((d_model) ** -0.5))
+        elif isinstance(module, nn.Linear):
+            module.weight.normal_(mean=0.0, std=factor)
```
vasqu: Could probably use super here, but I think Cyril works on refactoring the inits

src/transformers/models/pop2piano/modeling_pop2piano.py
```diff
@@ -948,7 +948,6 @@ class Pop2PianoForConditionalGeneration(Pop2PianoPreTrainedModel, GenerationMixi
     _tied_weights_keys = {
         "encoder.embed_tokens.weight": "shared.weight",
         "decoder.embed_tokens.weight": "shared.weight",
```
ArthurZucker, line 950: Yep force tie sounds good to me

ArthurZucker (approved):

PR #42198, opened 2025-11-13
Add AutoTokenizer mapping for mistral3 and ministral

patrickvonplaten:
# What does this PR do?

This PR makes sure that:

```py

tok = AutoTokenizer.from_pretrained("mistralai/Mistral-Small-3.2-24B-Instruct-2506")
tok = AutoTokenizer.from_pretrained("mistralai/Ministral-8B-Instruct-2410")
```

works correctly

### DIFF
--- a/src/transformers/models/auto/tokenization_auto.py
+++ b/src/transformers/models/auto/tokenization_auto.py
@@ -452,6 +452,15 @@ TOKENIZER_MAPPING_NAMES = OrderedDict[str, tuple[Optional[str], Optional[str]]](
             ),
         ),
         (
+            "mistral3",
+            (
+                "MistralCommonTokenizer"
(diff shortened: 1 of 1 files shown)


patrickvonplaten:
None of the test failures look related

ArthurZucker (approved):

vasqu (commented):
The auto mapping is incorrect and only works with `mistral_common`, see https://github.com/huggingface/transformers/issues/41553#issuecomment-3513138030. This is probably an issue for multiple mistral-related models now.

Tl;dr: It defines a llama backup but does not have the necessary files uploaded to the hub for that to work

ArthurZucker:
Yep, but we need to sort that on our side iMO!

PR #34213, opened 2024-10-17
Simplify running tests in a subprocess

ydshieh:
# What does this PR do?

In order to make `running (some individual) tests in a subprocess` work with GPU (cuda), if we use `multiprocessing`, we have to use `spawn`, but this uses pickle and has some limitations that make the necessary changes to the test methods quite big. 

This PR tries to use the `subprocess` module to implement a simpler way to run (some individual) tests in a subprocess.

PR #42206, opened 2025-11-14
Reduce timing on CircleCI - part 1 (Use @slow for IntegrationTests)

ydshieh:
# What does this PR do?

Reduce timing on CircleCI.

Use `@slow` for some IntegrationTests.

Note that, for IntegrationTest class that have `setupClass`, even if a test method is fast, the `setupClass` might still take quite long time, if it downloads models. Even tokenizers may take 3 -5 seconds sometimes

### DIFF
--- a/tests/models/imagegpt/test_modeling_imagegpt.py
+++ b/tests/models/imagegpt/test_modeling_imagegpt.py
@@ -249,6 +249,9 @@ class ImageGPTModelTest(ModelTesterMixin, GenerationTesterMixin, PipelineTesterM
         self.assertEqual(len(scores), generated_length)
         self.assertListEqual([iter_scores.shape for iter_scores in scores], [expected_shape] * len(scores))
 
+    # After #33632, this test still passes, but many subsequential tests fail with `device-side assert triggered`.
+    # We need to put `@slow` whenever `run_test_using_subprocess` is used.
+    @slow
     @run_test_using_subprocess
     def test_beam_search_generate_dict_outputs_use_cache(self):
         super().test_beam_search_generate_dict_outputs_use_cache()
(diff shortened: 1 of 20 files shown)


tests/models/imagegpt/test_modeling_imagegpt.py
```diff
@@ -249,6 +249,9 @@ def _check_scores(self, batch_size, scores, generated_length, config):
         self.assertEqual(len(scores), generated_length)
         self.assertListEqual([iter_scores.shape for iter_scores in scores], [expected_shape] * len(scores))
 
+    # After #33632, this test still passes, but many subsequential tests fail with `device-side assert triggered`.
+    # We need to put `@slow` whenever `run_test_using_subprocess` is used.
+    @slow
```
Cyrilvallez, line 254: I don't see how this can can the case when this is run on the CI? Because it does not have any accelerator available so no `device-side assert triggered`

Probably the issue is only the `@run_test_using_subprocess` which should not be necesary here I think!

Cyrilvallez (approved):
Makes a lot of sense, except the imagegpt one! I think this is unrelated!

tests/models/imagegpt/test_modeling_imagegpt.py
ydshieh: This comment is from my previous PR #34213, and I am sure I did the investigation at that time :-).

This test call the one in the parent class `GenerationTesterMixin` in the file `tests/generation/test_utils.py` and it does use `torch_device`.

Of course, when running on CircleCI, the `torch_device` is CPU and the comment doesn't apply in  this case. But we still need put `run_test_using_subprocess` so the problem won't occur when it's running on GPU enviornment.

I can of course make `run_test_using_subprocess` more flexible to accept some conditions, but it won't be done in this PR .

Let me know your opinion.

ydshieh: Merge now. Will address if there is further comment (which should be in a follow-up PR anyway)

PR #42208, opened 2025-11-14
New docker from AMD

remi-or:
This PR bumps the AMD docker base image because the last one was malfunctioning.


ydshieh (approved):
Thank you!

(This one works well?)

remi-or:
It works _better_ according to AMD! Time will tell I guess

PR #42212, opened 2025-11-14
Update `test_dynamic_cache_exportability_multiple_run` (failing on torch 2.10 nightly)

ydshieh:
# What does this PR do?

See 

https://github.com/pytorch/pytorch/issues/167293#issuecomment-3529014252

The suggestion seems to reasonable since the test name is 

> test_dynamic_cache_exportability_multiple_run


vasqu (approved):
LGTM, my only concern is that we should check for BC (not sure when that way of calling `.DYNAMIC` is supported)

ydshieh:
it's a test, so i won't worry the BC :-)

vasqu:
gotcha, fine by me then :p

PR #42217, opened 2025-11-14
support for a new model_type called blueberry

dustinwloring1988:
This PR adds support for a new `model_type` called **`blueberry`** in the Transformers library. It registers the model type, updates the configuration mapping, and ensures that `AutoConfig` and related utilities correctly recognize and load `blueberry` models.


Rocketknight1:
Hi @dustinwloring1988, are there any pretrained models with this architecture?

dustinwloring1988:
@Rocketknight1 yes I have 2 small ones a base and a chat finetune.

Rocketknight1:
Can you link them? We generally recommend starting with a [custom model](https://huggingface.co/docs/transformers/main/en/custom_models), and only migrating the models into Transformers itself after they have a lot of users, because the team at Hugging Face takes responsibility for maintaining them once they're in the main library!

PR #42225, opened 2025-11-16
Add SDPA and FlashAttention-2 support to LayoutLMv3

### LINKED ISSUE #35467: Support Flash Attention 2 for LayoutLMv3

jackiehimel:
# What does this PR do?

Adds SDPA and FlashAttention-2 support to LayoutLMv3 using the unified attention interface pattern, following the same architecture used in BERT and other recent model implementations.

Fixes #35467

## Implementation

- Refactored `LayoutLMv3SelfAttention` to use `ALL_ATTENTION_FUNCTIONS` interface instead of separate attention classes
- Created `layoutlmv3_eager_attention_forward` function that implements the CogView attention mechanism (alpha=32 scaling) with support for LayoutLMv3's relative position bias and spatial attention bias
- Added `_supports_sdpa = True` and `_supports_flash_attn = True` flags to `LayoutLMv3PreTrainedModel`
- Updated mask creation to use `create_bidirectional_mask` (replacing `get_extended_attention_mask`)
- Threaded `layer_idx` parameter through attention classes for consistency
- **Added automatic enforcement in `LayoutLMv3Config`** to set `attn_implementation="eager"` when relative or spatial attention biases are enabled (default behavior)

**Note**: SDPA and FlashAttention-2 are incompatible with LayoutLMv3's relative position bias and spatial attention bias. The config automatically enforces eager attention when these biases are enabled (the default). To use SDPA/FlashAttention-2, users must disable both biases (has_relative_attention_bias=False and has_spatial_attention_bias=False).

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How has this change been tested?

- CI tests passing, linting and formatting locally validated.
- Added test skips for SDPA/Flash comparison tests (LayoutLMv3 defaults to eager when biases are enabled) and overrode `test_batching_equivalence` to ensure eager attention is used.
- Implementation follows the unified attention pattern from BERT.

## Before submitting

## Who can review?

@vasqu - Thanks for the feedback on #41801! I've refactored this to use the unified attention interface pattern as you suggested. Would appreciate another look when you have time.

@ArthurZucker @CyrilVallez - attention implementation reviewers

### DIFF
--- a/src/transformers/integrations/deepspeed.py
+++ b/src/transformers/integrations/deepspeed.py
@@ -304,6 +304,7 @@ def _load_state_dict_into_zero3_model(model_to_load, state_dict):
         state_dict._metadata = metadata
 
     error_msgs = []
+    missing_keys = set(model_to_load.state_dict().keys())
 
     # PyTorch's `_load_from_state_dict` does not copy parameters in a module's descendants
     # so we need to apply the function recursively.
@@ -320,7 +321,14 @@ def _load_state_dict_into_zero3_model(model_to_load, state_dict):
             # In sharded models, each shard has only part of the full state_dict, so only gather
             # parameters that are in the current state_dict.
             named_parameters = dict(module.named_parameters(prefix=prefix[:-1], recurse=False))
-            params_to_gather = [named_parameters[k] for k in named_parameters if k in state_dict]
+            params_to_gather = []
+            for k in named_parameters:
+                if k in state_dict:
+                    param = named_parameters[k]
+                    # crutial to not init the weight again
+                    param._is_hf_initialized = True
+                    params_to_gather.append(param)
+                    missing_keys.discard(k)
 
             if len(params_to_gather) > 0:
                 # because zero3 puts placeholders in model params, this context
@@ -333,11 +341,10 @@ def _load_state_dict_into_zero3_model(model_to_load, state_dict):
         for name, child in module._modules.items():
             if child is not None:
                 load(child, state_dict, prefix + name + ".", assign_to_params_buffers)
-                child._is_hf_initialized = True
 
     load(model_to_load, state_dict, assign_to_params_buffers=False)
 
-    return error_msgs
+    return error_msgs, missing_keys
 
 
 def deepspeed_optim_sched(trainer, hf_deepspeed_config, args, num_training_steps, model_parameters):
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -4106,9 +4106,9 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
                 for ckpt_file in checkpoint_files:
                     merged_state_dict.update(load_state_dict(ckpt_file, map_location="cpu", weights_only=weights_only))
                 state_dict = merged_state_dict
-            error_msgs += _load_state_dict_into_zero3_model(model, state_dict)
+            error_msgs, missing_keys = _load_state_dict_into_zero3_model(model, state_dict)
             # This is not true but for now we assume only best-case scenario with deepspeed, i.e. perfectly matching checkpoints
-            missing_keys, unexpected_keys, mismatched_keys, conversion_errors = set(), set(), set(), set()
+            unexpected_keys, mismatched_keys, conversion_errors = set(), set(), set()
         else:
             all_pointer = set()
             # Checkpoints are safetensors
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -2349,7 +2349,8 @@ class Trainer:
         if self.is_fsdp_enabled:
             self.model = self.model_wrapped = model
             # Fix `got mixed torch.Tensor and DTensor` error in model.generate() for FSDP2 with LoRA
-            dist.fsdp.register_fsdp_forward_method(self.model, "generate")
+            if hasattr(self.model, "generate"):
+                dist.fsdp.register_fsdp_forward_method(self.model, "generate")
 
         # for the rest of this function `model` is the outside model, whether it was wrapped or not
         if model is not self.model:
--- a/tests/deepspeed/test_deepspeed.py
+++ b/tests/deepspeed/test_deepspeed.py
@@ -277,13 +277,17 @@ class CoreIntegrationDeepSpeed(TestCasePlus, TrainerIntegrationCommon):
         import deepspeed
         import torch
 
-        from transformers.models.gpt2.modeling_gpt2 import GPT2PreTrainedModel
+        from transformers.models.gpt2.modeling_gpt2 import GPT2Model, GPT2PreTrainedModel
 
         class TinyGPT2WithUninitializedWeights(GPT2PreTrainedModel):
             def __init__(self, config):
                 super().__init__(config)
-                self.transformer = AutoModel.from_pretrained(GPT2_TINY, config=config)
+                self.transformer = GPT2Model(config)
+                # AutoModel.from_pretrained(GPT2_TINY, config=config)
                 self.new_head = torch.nn.Linear(config.hidden_size, config.vocab_size, bias=True)
+
+                # Initialize weights and apply final processing
+                self.post_init()
 
             def forward(self, *args, **kwargs):
                 transformer_outputs = self.transformer(*args, **kwargs)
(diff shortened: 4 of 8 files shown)


src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -214,18 +220,80 @@ def _init_weights(self, module):
...
+    LayoutLMv3 eager attention with support for relative position bias and spatial attention bias.
+    Based on the CogView attention trick for training stability.
+    """
+    if scaling is None:
+        scaling = 1.0 / math.sqrt(query.size(-1))
+
+    # Take the dot product between "query" and "key" to get the raw attention scores.
+    # The attention scores QT K/√d could be significantly larger than input elements, and result in overflow.
+    # Changing the computational order into QT(K/√d) alleviates the problem. (https://huggingface.co/papers/2105.13290)
+    attention_scores = torch.matmul(query / math.sqrt(query.size(-1)), key.transpose(-1, -2))
```
vasqu: It's essentially the same as bert https://github.com/huggingface/transformers/blob/47227f477532e8ff6dd3290dc6b9e0e38aef303b/src/transformers/models/bert/modeling_bert.py#L131

with flipped order: `attn_weights = torch.matmul(query * scaling, key.transpose(2, 3))`

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -214,18 +220,80 @@ def _init_weights(self, module):
...
+        module.has_relative_attention_bias
+        and module.has_spatial_attention_bias
+        and rel_pos is not None
+        and rel_2d_pos is not None
+    ):
+        attention_scores = attention_scores + (rel_pos + rel_2d_pos) / math.sqrt(query.size(-1))
+    elif module.has_relative_attention_bias and rel_pos is not None:
+        attention_scores = attention_scores + rel_pos / math.sqrt(query.size(-1))
+    elif module.has_spatial_attention_bias and rel_2d_pos is not None:
+        attention_scores = attention_scores + rel_2d_pos / math.sqrt(query.size(-1))
```
vasqu: Are we applying the scaling twice here? Do the integration tests still pass? I.e. another ` / math.sqrt(query.size(-1)`

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -234,18 +302,7 @@ def __init__(self, config):
         self.dropout = nn.Dropout(config.attention_probs_dropout_prob)
         self.has_relative_attention_bias = config.has_relative_attention_bias
         self.has_spatial_attention_bias = config.has_spatial_attention_bias
```
vasqu, line 240: It's missing an `is_causal` attribute, I would think it's not causal

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -254,54 +311,45 @@ def forward(
         output_attentions=False,
```
vasqu, line 258:
```suggestion
```

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -254,54 +311,45 @@ def forward(
...
+
+        if not use_eager:
+            # SDPA and Flash Attention don't support custom relative position bias and spatial attention bias
+            if self.has_relative_attention_bias or self.has_spatial_attention_bias:
+                raise ValueError(
+                    f"You are using {self.config._attn_implementation} as attention type. However, LayoutLMv3's "
+                    "relative position bias and spatial attention bias are not compatible with it. "
+                    'Please load the model with `attn_implementation="eager"`.'
+                )
+            attention_interface = ALL_ATTENTION_FUNCTIONS[self.config._attn_implementation]
```
vasqu:
```suggestion
        if self.config._attn_implementation != "eager":
            # SDPA and Flash Attention don't support custom relative position bias and spatial attention bias
            if self.has_relative_attention_bias or self.has_spatial_attention_bias:
                raise ValueError(
                    f"You are using {self.config._attn_implementation} as attention type. However, LayoutLMv3's "
                    "relative position bias and spatial attention bias are not compatible with it. "
                    'Please load the model with `attn_implementation="eager"`.'
                )
            attention_interface = ALL_ATTENTION_FUNCTIONS[self.config._attn_implementation]
```
For now we raise a value error, thinking about forcing a fallback instead. Let's keep it as is for now

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -254,54 +311,45 @@ def forward(
...
+            dropout=0.0 if not self.training else self.dropout.p,
+            scaling=self.scaling,
+            rel_pos=rel_pos,
+            rel_2d_pos=rel_2d_pos,
+            **kwargs,
+        )
+        attn_output = attn_output.reshape(*input_shape, -1).contiguous()
 
+        outputs = (attn_output, attn_weights) if output_attentions else (attn_output,)
         return outputs
```
vasqu:
```suggestion
        return outputs, attn_weights
```
You can take a look at https://github.com/huggingface/transformers/blob/47227f477532e8ff6dd3290dc6b9e0e38aef303b/src/transformers/models/bert/modeling_bert.py#L560-L562

which takes care of the output xxx. Will need more changes here then but better to go for the right thing from the get go.

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
```diff
@@ -320,11 +368,11 @@ def forward(self, hidden_states: torch.Tensor, input_tensor: torch.Tensor) -> to
         return hidden_states
 
 
-# Copied from transformers.models.layoutlmv2.modeling_layoutlmv2.LayoutLMv2Attention with LayoutLMv2->LayoutLMv3
+# Adapted from transformers.models.layoutlmv2.modeling_layoutlmv2.LayoutLMv2Attention with LayoutLMv2->LayoutLMv3
```
vasqu: Is layoutlm v2 vastly different, might be worth to change along at the same time.

src/transformers/models/layoutlmv3/modeling_layoutlmv3.py
vasqu: Bert is a good reference for this + you need the same decorators, they are essential.

tests/models/layoutlmv3/test_modeling_layoutlmv3.py
```diff
@@ -162,12 +170,17 @@ def prepare_config_and_inputs(self):
...
+        # which is incompatible with SDPA/FlashAttention
+        config._attn_implementation = "eager"
 
         return config, input_ids, bbox, pixel_values, token_type_ids, input_mask, sequence_labels, token_labels
 
     def create_and_check_model(
         self, config, input_ids, bbox, pixel_values, token_type_ids, input_mask, sequence_labels, token_labels
     ):
+        # Ensure eager attention is set before model creation
+        config._attn_implementation = "eager"
```
vasqu: Is the change in `prepare_config_and_inputs` not enough? Same for below

tests/models/layoutlmv3/test_modeling_layoutlmv3.py
```diff
@@ -309,6 +331,55 @@ def setUp(self):
         self.model_tester = LayoutLMv3ModelTester(self)
         self.config_tester = ConfigTester(self, config_class=LayoutLMv3Config, hidden_size=37)
 
+    @parameterized.expand(TEST_EAGER_MATCHES_SDPA_INFERENCE_PARAMETERIZATION)
+    @unittest.skip("LayoutLMv3's relative position bias and spatial attention bias are incompatible with SDPA.")
+    def test_eager_matches_sdpa_inference(self, *args):
+        pass
```
vasqu: Does that mean that layout lm always uses the relative bias? Then it might not even make sense to support sdpa/flash attn - it won't be used either way

vasqu (commented):
Sorry, I did go through the code and gave some comments but I'm noticing that layout lm only uses the relative bias. If that's the case, then the usage of other attention flavors is questionable as they won't be used either way.

Imo, it would make more sense to go for other models that are suitable. You already did a overall pretty good job over here.

PR #42233, opened 2025-11-17
Fix processor test for glm

molbap:
# What does this PR do?

As per title, seems like the wrong arg name was passed to processor


zucchini-nlp (approved):
Thanks! I recently deleted the deprecation, might have not updated in a few places

Rocketknight1:
Merging because I want the CI to be green, yell at me if this causes any problems and I'll fix them :sweat_smile:

PR #42234, opened 2025-11-17
Fix initialization guard for pytest

Cyrilvallez:
# What does this PR do?

As per the title. Otherwise, when running multiple tests at once with `pytest`, we sometimes hit the following error: `RuntimeError: dictionary changed size during iteration` due to how multiple tests are handled with `pytest`.


src/transformers/initialization.py
```diff
@@ -179,7 +179,7 @@ def guard_torch_init_functions():
             # something like `from torch.nn.init import xavier_uniform_` in their internals (e.g in torch.nn.modules,
             # where MultiHeadAttention lives), so the function name is binded at import time and just doing
             # `setattr(torch.nn.init, name, gloabls()[name])` is thus not enough
-            for module in sys.modules.values():
+            for module in sys.modules.copy().values():
```
ArthurZucker, line 182:
```suggestion
            for module in sys.modules.values().copy():
```
we copy less?

ArthurZucker (approved):

PR #42235, opened 2025-11-17
delete already deprecated models

ydshieh:
# What does this PR do?

As discussed internally, we want to first delete these already deprecated models for V5.

(Before we add new models to deprecate)


ArthurZucker (approved):
TY!

ydshieh:
Merge , finally fixed all the issues. We have to make the process more easily in the future ... to many places to adjust

PR #42236, opened 2025-11-17
Fix TP plans for MoE models

Cyrilvallez:
# What does this PR do?

As per the title! After https://github.com/huggingface/transformers/pull/41580, the MoE structures have changed so the TP-plans need to be updated accordingly!


ArthurZucker (approved):
ty

PR #42242, opened 2025-11-17
[loading] Re-add and improve disk offloading support

Cyrilvallez:
# What does this PR do?

This PR re-add support for weight offloading to disk using the `device_map`, which was temporarily dropped in https://github.com/huggingface/transformers/pull/41580 to simplify the PR (because weights need to be resaved correctly after performing custom Ops during dynamic loading).

It further improve the offloading mechanism as everything is now offloaded in `safetensors` format, vs `numpy` format originally when using `accelerate`. This is much easier and efficient.

Slow tests are exactly similar to before the big weight loading refactor (10 failing over all models)

### DIFF
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -452,7 +452,7 @@ def set_param_for_module(
 
         ref = getattr(module_obj, param_name)
         if ref is None:
-            unexpected_keys.add(layer_name)
+            unexpected_keys.add(full_param_name)
         else:
             use_dtensor = hasattr(distributed_operation, "use_dtensor") and distributed_operation.use_dtensor
             if not isinstance(param_value, torch.nn.Parameter):
@@ -472,14 +472,13 @@ def set_param_for_module(
                     param_value = torch.nn.Parameter(param_value, requires_grad=param_value.is_floating_point())
 
             # Remove from missing keys (it's either mismatched, or all good)
-            missing_keys.discard(layer_name)
+            missing_keys.discard(full_param_name)
             if ref is not None and ref.shape != param_value.shape and hf_quantizer is None:
-                mismatch_keys.add((layer_name, param_value.shape, ref.shape))
+                mismatch_keys.add((full_param_name, param_value.shape, ref.shape))
                 module_obj.param_name._is_hf_initialized = False  # Needs to be initialized
             else:
-                param_value._is_hf_initialized = (
-                    True  # super important otherwise _init_weight re-initi if bias is missing
-                )
+                # super important otherwise _init_weight will re-init the param
+                param_value._is_hf_initialized = True
                 setattr(module_obj, param_name, param_value)
 
 
@@ -711,6 +710,8 @@ def convert_and_load_state_dict_in_model(
             if future is None:
                 device_match = device_map_regex.match(renamed_key)
                 param_device = device_map[device_match.group()] if device_match else device_map.get("", "cpu")
+                # If disk, we need to materialize on cpu first
+                param_device = "cpu" if param_device == "disk" else param_device
                 future = spawn_materialize(thread_pool, tensor, param_device, _dtype)
 
             mapping.add_tensor(renamed_key, original_key, source_pattern, future)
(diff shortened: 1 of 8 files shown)


src/transformers/core_model_loading.py
```diff
@@ -551,55 +556,70 @@ def convert_and_load_state_dict_in_model(
...
+                                device_map[device_match.group()] if device_match else device_map.get("", "cpu")
                             )
+                            # Offloading support
+                            if param_device == "disk":
+                                missing_keys.discard(target_name)
+                                # If not already offloaded, or if we applied any special Operation, we need to re-save
+                                if target_name not in disk_offload_index or len(operations) > 0:
+                                    disk_offload_index = offload_weight(
+                                        param, target_name, disk_offload_folder, disk_offload_index
+                                    )
```
ArthurZucker: can be done inside set param for module as its kinda related

ArthurZucker (commented):
for the weight renaming, IMO we should just use the prepare model state dict instead of the weights

src/transformers/core_model_loading.py
```diff
@@ -723,30 +724,43 @@ def convert_and_load_state_dict_in_model(
 
     total_entries = len(param_name_to_load)
     with logging.tqdm(total=total_entries, desc="Loading weights") as pbar:
-        for layer_name, mapping in param_name_to_load.items():
+        for full_param_name, mapping in param_name_to_load.items():
```
ArthurZucker: it is the full param name but only the first one

tests/test_modeling_common.py
ArthurZucker, line 1: I think there were other tests I skiped with a todo @Cyrilvallez but ff to check

tests/utils/test_core_model_loading.py
```diff
@@ -250,7 +250,7 @@ def test_moe_and_qkv_conversion(self):
             ),
             WeightRenaming("mlp.w2.weight", "mlp.down_proj.weight"),
         ]
-        missing, unexpected, mismatch, misc = convert_and_load_state_dict_in_model(
+        missing, unexpected, mismatch, _, misc = convert_and_load_state_dict_in_model(
```
ArthurZucker, line 253: if never used why return it

src/transformers/modeling_utils.py
```diff
@@ -4103,16 +4116,20 @@ def _load_pretrained_model(
...
-                model,
-                merged_state_dict,
-                weight_mapping,
-                tp_plan,
-                hf_quantizer,
-                dtype,
-                device_map,
-                model.dtype_plan,
-                device_mesh,
+            missing_keys, unexpected_keys, mismatched_keys, disk_offload_index, misc = (
```
ArthurZucker, line 4119: ah ok used here

src/transformers/core_model_loading.py
Cyrilvallez: I preferred to keep it outside for now, as it's actually not setting the param! (just saves it to disk and skip loading) So would be a bit weird IMO to do that inside `set_param` as it does not set it

src/transformers/integrations/accelerate.py
```diff
@@ -445,71 +449,96 @@ def accelerate_dispatch(model, hf_quantizer, device_map, offload_folder, offload
...
+    disk_offload_folder: str | None,
+    checkpoint_files: list[str] | None,
+    device_map: dict,
+    expected_keys: list[str],
+    sharded_metadata: dict | None,
+    dtype: torch.dtype | None,
+    weight_mapping=None,
 ):
-    disk_only_shard_files = []
+    from ..core_model_loading import WeightRenaming, build_glob_alternation, repl
```
ArthurZucker, line 483: just missing a doc about "weights that are only renamed are not saved twice" or something like that

src/transformers/core_model_loading.py
```diff
@@ -397,7 +398,7 @@ def dot_natural_key(s: str):
 
 @contextmanager
 def log_to_misc(
-    layer_name: str,
+    all_target_keys: str,
```
ArthurZucker: to revert, its the first_target_key

src/transformers/core_model_loading.py
```diff
@@ -723,30 +724,43 @@ def convert_and_load_state_dict_in_model(
...
+                    device_match = device_map_regex.match(target_name)
+                    param_device = device_map[device_match.group()] if device_match else device_map.get("", "cpu")
+                    # Offloading support
+                    if param_device == "disk":
+                        missing_keys.discard(target_name)
+                        # If not already offloaded, or if we applied any special Operation except Renaming, we need to re-save
+                        if target_name not in disk_offload_index or isinstance(mapping, WeightConverter):
+                            disk_offload_index = offload_weight(
+                                param, target_name, disk_offload_folder, disk_offload_index
+                            )
```
ArthurZucker: can you put all of this in  a util ? or as much as possible please!

ArthurZucker (approved):
Very nice thanks!

tests/test_modeling_common.py
Cyrilvallez: Indeed I missed a few! Nice catch!

PR #42247, opened 2025-11-17
Stop inheriting tests (again)

Rocketknight1:
This is a follow up to #42192 - after discussion, we decided not to use `del` and instead to avoid importing `TestCase` classes.


tests/models/granitemoehybrid/test_modeling_granitemoehybrid.py
```diff
@@ -77,7 +85,7 @@ def get_config(self):
 
 
 @require_torch
-class GraniteMoeHybridModelTest(BambaModelTest, GenerationTesterMixin, unittest.TestCase):
+class GraniteMoeHybridModelTest(ModelTesterMixin, GenerationTesterMixin, PipelineTesterMixin, unittest.TestCase):
```
tengomucho, line 88: Perhaps add a comment explaining why it is better not to inherit from another test, so if someone is tempted to do the same, that someone will have a hint

tengomucho (approved):
LGTM

ydshieh (approved):
thx!

PR #30639, opened 2024-05-03
Generate: add `min_p` sampling

gante:
# What does this PR do?

Adds `min_p`, a dynamic sampling strategy. Instead of picking a fixed number of tokens (`top_k`) or a fixed cumulative probability (`top_p`), it sets a minimum probability filter that scales with the most likely token in that round:
- High probability token present -> aggressive filter (we don't want to miss on that high-probability case and risk derailing generation)
- No high probability token present -> relaxed filter (there are many continuation possibilities that the model finds plausible)

Fixes #27670, opened by the authors of the technique. Opening the PR now because we now have empirical checks on the technique (see the original issue), in addition to the vibe checks.

________________________

Cool example:
```py
chat = [
    {"role": "system", "content": "You are a sassy, wise-cracking robot as imagined by Hollywood circa 1986."},
    {"role": "user", "content": "Hey, can you tell me any fun things to do in New York?"}
]

import torch
from transformers import pipeline, set_seed

set_seed(0)

pipe = pipeline("text-generation", "meta-llama/Meta-Llama-3-8B-Instruct", torch_dtype=torch.bfloat16, device_map="auto")
response = pipe(chat, max_new_tokens=512, do_sample=True, min_p=0.08, temperature=1.5)
print(response[0]['generated_text'][-1]['content'])
```

with `min_p` 
```
*snap* *whirr* Oh, you're talkin' to a robot, pal! I've got the 411 on the Big Apple! Now, let me think... (beep boop) Ah, yeah! You gotta hit Times Square, kid! It's like, the center of the universe or something. All those bright lights, giant billboards, and street performers will blow your mind, man! (zzt) And don't even get me started on the food! You can get a slice of pizza the size of your face from one of those 24-hour joints on every corner! (ding) Just watch out for those crazy cabbies, they'll almost run you over! (sputter)

Or, if you're feelin' fancy, take a stroll across the Brooklyn Bridge. (whoosh) The views of the Manhattan skyline are like, totally tubular, dude! (zip) And when you get to Brooklyn, grab a bagel from one of those classic shops – you'll be saying "Fuhgeddaboudit!" like a native in no time! (beep)

And don't miss the Statue of Liberty, either. (ding) You know, that big ol' copper chick with the torch? (whirr) It's like, an iconic symbol of freedom and all that jazz. (buzz) Just don't forget to wear sunscreen, or you'll be lobster-red by the time you get back to your hotel room! (zort)

[...]

PR #42248, opened 2025-11-17
perf: Optimization for Min-p sampling implementation

casinca:
# What does this PR do?

tldr: This PR aims to optimize `MinPLogitsWarper` implementation that was done in https://github.com/huggingface/transformers/pull/30639

Hello,

While I was re-implementing Min-P for my own sake, and later compared the current HF/official implementation, I've noticed that we could use a single `torch.topk` call to retrieve `min_tokens_to_keep` instead of chaining `argsort` + `gather` + masking.  
Also, minor, I simplified using `torch.amax` instead of `torch.max` for getting only top_probs.

Overall, besides being faster, the readability is also improved I believe.

---

Benchmarking in isolation, the proposed optimization is noticeably faster, in bf16, varying single/multiple batches, vocab_sizes:
- on cpu, 2x up to 10x
- on cuda, 20% up to 100%

In real situation, generating with a `Qwen3-0.6B` model:
- on cpu, i'm getting ~2s faster per `model.generate()` call (on average) with `max_new_tokens=200`
- on cuda,  there's no meaningful difference

All tests passed. There's no extra changes needed to do I believe, it's just internal code optimization.

Min-P is a very popular sampling alternative after top_k and top_p if I'm not mistaken, so if this can help slightly speed up inference, I thought why not?

&nbsp;

## Before submitting
      Pull Request section?
      to it if that's the case.

&nbsp;

## Who can review?

Let me know what you think @gante 
Also tagging @menhguin as he is a co-author on Min-P sampling who participated in the original PR

### DIFF
--- a/src/transformers/generation/logits_process.py
+++ b/src/transformers/generation/logits_process.py
@@ -748,19 +748,17 @@ class MinPLogitsWarper(LogitsProcessor):
         # Convert logits to probabilities
         probs = torch.softmax(scores, dim=-1)
         # Get the probability of the top token for each sequence in the batch
-        top_probs, _ = probs.max(dim=-1, keepdim=True)
+        top_probs = probs.amax(dim=-1, keepdim=True)
         # Calculate the actual min_p threshold by scaling min_p with the top token's probability
         scaled_min_p = self.min_p * top_probs
         # Create a mask for tokens that have a probability less than the scaled min_p
         tokens_to_remove = probs < scaled_min_p
 
-        sorted_indices = torch.argsort(scores, descending=True, dim=-1)
-        sorted_indices_to_remove = torch.gather(tokens_to_remove, dim=-1, index=sorted_indices)
         # Keep at least min_tokens_to_keep
-        sorted_indices_to_remove[..., : self.min_tokens_to_keep] = False
+        sorted_indices = torch.topk(probs, self.min_tokens_to_keep, dim=-1).indices
+        tokens_to_remove.scatter_(-1, sorted_indices, False)
 
-        indices_to_remove = sorted_indices_to_remove.scatter(1, sorted_indices, sorted_indices_to_remove)
-        scores_processed = scores.masked_fill(indices_to_remove, self.filter_value)
+        scores_processed = scores.masked_fill(tokens_to_remove, self.filter_value)
         return scores_processed
 
 
--- a/src/transformers/generation/logits_process.py
+++ b/src/transformers/generation/logits_process.py
@@ -754,8 +754,9 @@ class MinPLogitsWarper(LogitsProcessor):
         # Create a mask for tokens that have a probability less than the scaled min_p
         tokens_to_remove = probs < scaled_min_p
 
-        # Keep at least min_tokens_to_keep
-        sorted_indices = torch.topk(probs, self.min_tokens_to_keep, dim=-1).indices
+        # Keep at least min_tokens_to_keep tokens (clip k to vocab size if needed, avoids index out of range)
+        k = min(self.min_tokens_to_keep, probs.shape[-1])
+        sorted_indices = torch.topk(probs, k, dim=-1).indices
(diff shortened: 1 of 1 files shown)


Rocketknight1:
Very interesting! This looks better than the original already. I wonder if we could compute this even more directly, though - really all we're doing is is keeping all the elements larger than `scaled_min_p` **and** the n-th largest probability, where n is `self.min_tokens_to_keep`. So we could just do something like

```python
cutoff_value = torch.quantile(probs, ...)  # Set the quantile so that we get the n-th largest value for each sequence
cutoff_value.clamp(min=scaled_min_p)  # Clip the cutoff value if it's smaller than scaled_min_p
scores_processed = scores.masked_fill(probs < cutoff_value, self.filter_value)
```

and totally avoid computing `top_probs`, `sorted_indices` or `tokens_to_remove`. The behaviour might change slightly in cases where a lot of tokens have probability exactly equal to the cutoff value, but that seems like an edge case where either behaviour is acceptable.

I'm not sure if this works, but it might be worth a try! If you don't feel like benchmarking it, though, I can just review and merge your existing code, which is already a nice speedup.

EDIT: I think `torch.quantile()` is actually not very optimized, so this approach probably won't work. Finding the n-th largest element might be quicker with `topk(sorted=True)`

casinca:
Hey @Rocketknight1, thanks for the suggestion.

I tried with quantile() and `q= 1- (self.min_tokens_to_keep/vocab_size)` you had the right intuition because it was way worse than using topk.

You said 
> and totally avoid computing `top_probs`

I'd still have to compute it to get `scaled_min_p` if i'm not mistaken?  
If you meant `cutoff_value.clamp(min=scaled_min_p)` only, it could fail for some cases where we also need to guarantee at least `self.min_tokens_to_keep`  
`nth_prob.clamp(max=scaled_min_p)` work (more readable imo is `torch.minimum(scaled_min_p, nth_prob)`)

&nbsp;

Based on your suggestion I tried this alternative:

```python
def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor) -> torch.FloatTensor:
    # Convert logits to probabilities
    probs = torch.softmax(scores, dim=-1)

    topk_probs = probs.topk(self.min_tokens_to_keep, dim=-1).values
    # get top prob and n-th prob
    top_prob = topk_probs.amax(dim=-1, keepdim=True)
    nth_prob = topk_probs[..., -1]
    # Calculate the actual min_p threshold by scaling min_p with the top token's probability
    scaled_min_p = top_prob * self.min_p
    # we keep tokens that are >= scaled_min_p but also >= nth_prob
    cutoff_value = torch.minimum(scaled_min_p, nth_prob)
    # or cutoff_value = nth_prob.clamp(max=scaled_min_p)

    scores_processed = scores.masked_fill(probs < cutoff_value, self.filter_value)
    return scores_processed
```

But I didn’t see meaningful improvements compared to the first optimization, to justify choosing one over the other. Both are faster than the original or using `torch.quantile()`. Quite pronounced on cpu and to a lesser extent on cuda.

I guess it's up to you guys on what style/preference you want to go with.

Rocketknight1:
Yes, thank you for checking! I'm happy with your implementation in that case. One final edge case I thought of, though: If `min_tokens_to_keep` is larger than the vocab size, `topk` will throw an error, but the existing code will tolerate this because `sorted_indices_to_remove[..., : self.min_tokens_to_keep]` will work fine even if that slice runs off the end of the array.

Maybe we keep your implementation, but clip the `k` arg to `topk` so that it's never larger than the array size, to avoid that potential error? Then we should be fully equivalent to the original code while keeping the performance improvements.

casinca:
Ah yes, that's a good find👍

Updated to take into account this edge case:  
`k` for `topk` is now the `min` of `self.min_tokens_to_keep` and vocab size (inferred from `probs.shape[-1]`)

Rocketknight1 (approved):
Yes, code looks good to me now. Thanks for a nice PR!

PR #42254, opened 2025-11-18
Update conversion mapping to separate renaming from converting

ArthurZucker:
Update the conversion: 
image

image

to support torchao 
image


src/transformers/core_model_loading.py
```diff
@@ -49,69 +47,98 @@
...
-    >>> match_ = reg.match("model.layers.0.mlp.0.w1.weight")
-    >>> print(match_.lastgroup)
-    'g0'
-    >>> print(map_[match_.lastgroup])
-    mlp.*.w1
-    ```
+    def _star_to_backref(_: re.Match) -> str:
+        nonlocal counter
+        counter += 1
+        return rf"\{counter}"
```
molbap, line 65: I see I see :face_with_spiral_eyes: not obvious though

src/transformers/core_model_loading.py
```diff
@@ -49,69 +47,98 @@
...
+                branches.append(f"(?P<{group_name}>{body})")
+                tgt_group_to_glob[group_name] = glob.target_keys[0]  # we index witht the first target
+        else:
+            group_name = f"g{i}"
+            src_group_to_glob[group_name] = glob
+            i += 1
+            body = glob
+            body = body.replace("*", r".*")
+            branches.append(f"(?P<{group_name}>{body})")
+            tgt_group_to_glob[group_name] = glob
```
molbap, line 97: not sure I get why we have this branch for the non-rename non-convert, what does it match?

src/transformers/core_model_loading.py
```diff
@@ -237,11 +299,21 @@ def _apply(self, tensor: torch.Tensor) -> torch.Tensor:
 
     @torch.no_grad
     def convert(
-        self, value: Union[dict[str, torch.Tensor], Sequence[torch.Tensor], torch.Tensor], config
-    ) -> Union[dict[str, torch.Tensor], list[torch.Tensor], torch.Tensor]:
+        self,
+        value: dict[str, list[torch.Tensor]],
+        *,
```
molbap: to be sure, why do we need the variadic option here?

src/transformers/core_model_loading.py
```diff
@@ -260,37 +332,111 @@ class WeightConverter:
...
-            if self.target_keys is None:
-                self.target_keys = list(self.source_keys)
-                targets_were_none = True
-            else:
-                self.target_keys = [self.target_keys]
+        if isinstance(self.target_keys, str):
+            self.target_keys = [self.target_keys]
 
-        if not targets_were_none and bool(len(self.source_keys) - 1) + bool(len(self.target_keys) - 1) >= 2:
+        if bool(len(self.source_keys) - 1) + bool(len(self.target_keys) - 1) >= 2:
```
molbap: :+1:

src/transformers/core_model_loading.py
```diff
@@ -260,37 +332,111 @@ class WeightConverter:
...
+                    self.collected_tensors,
+                    source_keys=self.source_keys,
+                    target_keys=self.target_keys,
+                    target_key=layer_name,
+                    config=config,
+                    quant_config=quantizer.quantization_config,
+                    missing_keys=missing_keys,
+                )
+
+        return collected_tensors, misc
```
molbap: Ok, clear - maybe a parent class to hold the add_tensor? like `WeightOp`

molbap (commented):
Sweet update!

src/transformers/core_model_loading.py
```diff
@@ -124,8 +151,15 @@ class ConversionOps:
 
     @abstractmethod
     def convert(
-        self, value: Union[dict[str, torch.Tensor], Sequence[torch.Tensor], torch.Tensor], *args, **kwargs
-    ) -> torch.Tensor:
+        self,
+        value: dict[str, Any],
+        *,
```
Cyrilvallez: Having the wildcard `*` here is very error prone IMO -> would just lead to silently skip args

src/transformers/core_model_loading.py
```diff
@@ -124,8 +151,15 @@ class ConversionOps:
...
+        source_keys: list[str],
+        target_keys: list[str],
+        target_key: str,
```
Cyrilvallez: Same args twice, we should not need to be either a list or a single arg with 2 similar names IMO

src/transformers/core_model_loading.py
Cyrilvallez: nit: I'd rather have this without the `bool` - makes it harder to understand instead of simpler patterns IMO

src/transformers/core_model_loading.py
```diff
@@ -260,37 +332,111 @@ class WeightConverter:
...
-        if not targets_were_none and bool(len(self.source_keys) - 1) + bool(len(self.target_keys) - 1) >= 2:
+        if bool(len(self.source_keys) - 1) + bool(len(self.target_keys) - 1) >= 2:
             raise ValueError(
                 f"source keys={self.source_keys}, target_keys={self.target_keys} but you can only have one to many, one to one or many to one."
             )
 
+        if not self.operations:
+            raise ValueError("WeightConverter requires at least one operation.")
+
+    def add_tensor(self, target_key: str, original_key: str, source_pattern: str, future: Future):
```
Cyrilvallez:
```suggestion

    def add_tensor(self, target_key: str, source_key: str, source_pattern: str, future: Future):
```
Let's always talk about sources and targets, instead of mixing names!

src/transformers/core_model_loading.py
```diff
@@ -260,37 +332,111 @@ class WeightConverter:
...
+        bucket = self.collected_tensors.setdefault(source_pattern, [])
+        bucket += [future]
+
+        bucket = self.layer_targets.setdefault(target_key, set())
```
Cyrilvallez, line 291: Maybe we can use `defaultdict` here for the attributes instead of using `setdefault`?

src/transformers/core_model_loading.py
```diff
@@ -260,37 +332,111 @@ class WeightConverter:
...
+            self.source_keys = [self.source_keys]
+        if isinstance(self.target_keys, str):
+            self.target_keys = [self.target_keys]
+
+    def add_tensor(self, target_key: str, original_key: str, source_pattern: str, future: Future):
+        bucket = self.collected_tensors.setdefault(source_pattern, [])
+        bucket += [future]
+
+        bucket = self.layer_targets.setdefault(target_key, set())
+        bucket.add(original_key)
```
Cyrilvallez: Same comments as above here

src/transformers/conversion_mapping.py
```diff
@@ -26,12 +26,13 @@
 def _build_checkpoint_conversion_mapping():
     mapping = {
         "mixtral": [
+            WeightRenaming(r"\.block_sparse_moe\.", ".mlp."),
             WeightConverter(
```
Cyrilvallez: Not a fan to need both for a single set of keys, see also my other comment!

src/transformers/core_model_loading.py
```diff
@@ -450,135 +676,127 @@ def convert_and_load_state_dict_in_model(
...
-            if empty_param is None:
-                unexpected_keys.add(t)
-                continue
+        # 1. apply all renamings
+        renamed_key = rename_alt.sub(lambda m: repl(m, rename_by_group), original_key).replace("\\", "")
+
+        # 2. apply 1 weight conversion on the key
+        matched_pattern = weight_pattern_alt.search(renamed_key) if converters != [] else None
+        if matched_pattern is not None:  # we have a converter to apply
+            renamed_key = weight_pattern_alt.sub(lambda m: repl(m, tgt_group_to_glob), renamed_key).replace("\\", "")
```
Cyrilvallez, line 654: To me that's a bit weird to have to rename twice like that. IMO, the best would be: 

- we have a key mapping (e.g. vlms) -> converter is a WeightRenaming
- we have actual tensor ops -> converter is a WeightConverter, which also handles the renaming (so just like before)

Needing both for s single key is weird to me, as most of the time after the Operations (Chunk, Cat etc), the keys _need_  to change name anyway

If we do that, then we only have to rename once

src/transformers/core_model_loading.py
```diff
@@ -450,135 +676,127 @@ def convert_and_load_state_dict_in_model(
...
+                        output_value,
+                        mismatch_keys,
+                        missing_keys,
+                        misc,
+                        unexpected_keys,
+                        mapping.distributed_operation,
+                        hf_quantizer,
+                    )
+            except SkipLayer:
+                continue
```
Cyrilvallez: Yes! That's really what we want 😍 A single simple loop over COnverters

Cyrilvallez (commented):
All right! Much simpler than before! 
Did not check in details all the complex regex patterns for edge cases though, but they look similar to before
The only thing I'm really not a fan of, is the double WeightRenaming + WeightConverter for the same keys -> IMO if we need any Operation, then we should simply have a WeightConverter correctly specifying sources and targets

src/transformers/conversion_mapping.py
ArthurZucker: yeah, as we discussed you needd this for VLM not for mixtral specifically, will update mixtral as for this one we can encourage exact match

src/transformers/core_model_loading.py
ArthurZucker: yep gonna remove

src/transformers/core_model_loading.py
ArthurZucker: yep, as discussed offline that's what this tries to do!

src/transformers/core_model_loading.py
ArthurZucker: for TP plan and DTYPE plan!

src/transformers/core_model_loading.py
ArthurZucker: they are actually not the same!

ArthurZucker: renamed for convenience tho

src/transformers/core_model_loading.py
ArthurZucker: keeping for now i am a bit lazy with dataclasses

src/transformers/core_model_loading.py
```diff
@@ -237,60 +294,112 @@ def _apply(self, tensor: torch.Tensor) -> torch.Tensor:
...
+@dataclass(slots=True)
+class WeightConverter(WeightTransform):
+    operations: list[ConversionOps] = field(default_factory=list, repr=False)
+
+    def __post_init__(self):
+        WeightTransform.__post_init__(self)
+        if bool(len(self.source_keys) - 1) + bool(len(self.target_keys) - 1) >= 2:
+            raise ValueError(
+                f"source keys={self.source_keys}, target_keys={self.target_keys} but you can only have one to many, one to one or many to one."
+            )
```
vasqu, line 329: Let's remove this constraint while we're at it, can also do this in the ernie vl PR :D

ArthurZucker: will prob leave you to it 👀

PR #42255, opened 2025-11-18
🚨 Move `rotary_partial_emb` to RopeParams and delete unnecessary code 🔪 

zucchini-nlp:
# What does this PR do?

To finalize the work on rope config, I am moving `rotary_partial_emb` to rope parameter dict as well. Along with it, I did some clean-up on standardization because we can make a few assumptions with the models we have

Note, PR is breaking BC completely and users will no longer have access to `config.rope_theta` since I pop it from config kwargs manually. That way is more clear imo than having two rope thetas in different places


src/transformers/modeling_rope_utils.py
```diff
@@ -27,55 +27,87 @@
...
+def get_standardized_rope_params(config):
     """
     Helper to standardize the config's rope params field by ensuring the params are defined for each
     later type. For old model the fn will duplicate a single rope param in each layer type (backward compatibility)
     """
-    rope_parameters = getattr(config, "rope_parameters", None)
-    layer_types = getattr(config, "layer_types", None)
-    if rope_theta is None:
-        rope_theta = getattr(config, "rope_theta", None)
+    rope_parameters = getattr(config, "rope_parameters", {})
```
zucchini-nlp: could have been simplified if we make a few assumption, and we can make assumptions because only 2 models have a nested rope parameterization

src/transformers/modeling_rope_utils.py
```diff
@@ -854,17 +886,17 @@ def _validate_llama3_parameters(rope_parameters: dict, config: PreTrainedConfig,
 }
 
 
-def rope_config_validation(config: PreTrainedConfig, ignore_keys: Optional[set] = None):
+def rope_config_standardize_and_validate(config: PreTrainedConfig, ignore_keys: Optional[set] = None):
```
zucchini-nlp: renaming is a bit breaking, especially with custom code. I'm quite sure it's used a lot by copying from Llama. Renamed only to be explicit but surely can revert it

src/transformers/models/apertus/configuration_apertus.py
```diff
@@ -160,14 +160,16 @@ def __init__(
...
+        rope_parameters = rope_scaling or rope_parameters
+        self.rope_parameters = rope_parameters if rope_parameters is not None else {}
 
         # Validate the correctness of rotary position embeddings parameters
-        rope_theta = kwargs.get("rope_theta", 12000000.0)
-        standardize_rope_params(self, rope_theta=rope_theta)
-        rope_config_validation(self)
+        if "rope_theta" not in self.rope_parameters:
+            self.rope_parameters["rope_theta"] = kwargs.pop("rope_theta", 12000000.0)
+        rope_config_standardize_and_validate(self)
```
zucchini-nlp: the standardize fn had weird assumption. It would require a theta when called from config or assume theta is already in config when called from rope compute functions

I decided we can pop the theta (`config.rope_theta` throws error now) and directly assign it in the rope dict. It will hopefully also simplify the BC code on vLLM-side?

src/transformers/modeling_rope_utils.py
```diff
@@ -27,55 +27,87 @@
...
+class RopeParameters(TypedDict, total=False):
+    """
+    Args:
+        rope_theta (`float`):
+            The base period of the RoPE embeddings.
+        rope_type (`str`, *optional*, defaults to "default"):
+            The sub-variant of RoPE to use. Can be one of ['default', 'linear', 'dynamic', 'yarn', 'longrope',
+            'llama3'], with 'default' being the original RoPE implementation.
+        partial_rotary_factor (`float`, *optional*):
+            Percentage of the query and keys which will have rotary embedding.
```
vasqu:
```suggestion
        partial_rotary_factor (`float`, *optional*):
            The percentage of the query and key head embedding on which RoPE will be applied.
```
Just some more explicit doc, not sure if this is really better

src/transformers/modeling_rope_utils.py
```diff
@@ -27,55 +27,87 @@
...
-            # BC: if there is a 'type' field, copy it it to 'rope_type'.
-            rope_type = rope_parameters.get("rope_type", rope_parameters.get("type", "default"))
-            rope_theta = rope_parameters.get("rope_theta") or rope_theta
-            rope_parameters.update({"rope_theta": rope_theta, "rope_type": rope_type})
-        config.rope_parameters = rope_parameters
+    # Move `rope_theta` and `partial_rotary_factor` to the params dict, if not there yet
+    rope_theta = getattr(config, "rope_theta", None)
+    partial_rotary_factor = getattr(config, "partial_rotary_factor", None)
 
+    # Case 1: one RoPE theat = one RoPE param per model without nesting
```
vasqu:
```suggestion
    # Case 1: one RoPE theta = one RoPE param per model without nesting
```
typo?

src/transformers/modeling_rope_utils.py
```diff
@@ -27,55 +27,87 @@
...
+    if not set(rope_parameters.keys()).issubset(ALLOWED_LAYER_TYPES):
+        rope_parameters.setdefault("rope_type", rope_parameters.get("type", "default"))
+        rope_parameters.setdefault("rope_theta", rope_theta)
+        if partial_rotary_factor is not None:
+            rope_parameters["partial_rotary_factor"] = partial_rotary_factor
```
vasqu: Why not use `setdefault` here as well? If I put it directly into `rope_parameters`, would I not get overwritten/forced by the config

src/transformers/modeling_rope_utils.py
```diff
@@ -27,55 +27,87 @@
...
-                        **rope_parameters[layer_type],
-                        "rope_type": curr_rope_type,
-                        "rope_theta": rope_theta[layer_type],
-                    }
-            config.rope_parameters = rope_parameters_per_layer_type
+        for layer_type in config.layer_types:
+            rope_parameters[layer_type].setdefault("rope_type", rope_parameters[layer_type].get("type", "default"))
+            rope_parameters[layer_type].setdefault("rope_theta", rope_theta)
+            if partial_rotary_factor is not None:
+                rope_parameters[layer_type]["partial_rotary_factor"] = partial_rotary_factor
```
vasqu: Same here then

src/transformers/modeling_rope_utils.py
vasqu: A bit confusing comment to me, aren't we just saying --> no layer types == one global dict

src/transformers/modeling_rope_utils.py
vasqu: Why not introduce a wrapper around both functions instead? Then we have standardize, validate, and the wrapper that calls both subsequently - nothing gets broken

src/transformers/models/apertus/configuration_apertus.py
```diff
@@ -160,14 +160,15 @@ def __init__(
         self.use_cache = use_cache
         self.attention_bias = attention_bias
         self.attention_dropout = attention_dropout
+
         # Try to set `rope_scaling` if available, otherwise use `rope_parameters`
         rope_scaling = kwargs.pop("rope_scaling", None)
-        self.rope_parameters = rope_scaling or rope_parameters
+        rope_parameters = rope_scaling or rope_parameters
+        self.rope_parameters = rope_parameters if rope_parameters is not None else {}
```
vasqu: This is a bit awkward, could the ternary be moved to the standardize fn instead?

src/transformers/models/efficientloftr/configuration_efficientloftr.py
```diff
@@ -176,17 +172,17 @@ def __init__(
...
         # Try to set `rope_scaling` if available, otherwise use `rope_parameters`
         rope_scaling = kwargs.pop("rope_scaling", None)
-        self.rope_parameters = rope_scaling or rope_parameters
-
-        # Validate the correctness of rotary position embeddings parameters
-        rope_theta = kwargs.get("rope_theta", 10000.0)
-        standardize_rope_params(self, rope_theta=rope_theta)
-        rope_config_validation(self)
+        rope_parameters = rope_scaling or rope_parameters
+        self.rope_parameters = rope_parameters if rope_parameters is not None else {} or {}
```
vasqu:
```suggestion
        self.rope_parameters = rope_parameters if rope_parameters is not None else {}
```

vasqu (commented):
Just my 2 cents :smile:

src/transformers/models/apertus/configuration_apertus.py
ArthurZucker: do you think super could do the standardization for us? this way we "hide" it?

ArthurZucker (approved):
In general if we can put stuff in `PreTrainedConfig` I am also happy, but fine this way as well

src/transformers/modeling_rope_utils.py
zucchini-nlp: Had a hard time trying to find good wording here so I just copied existing docs haha. I agree that sounds a bit raw

src/transformers/modeling_rope_utils.py
zucchini-nlp: That's a sign to update the comment then. We need clear comments for future us

src/transformers/modeling_rope_utils.py
zucchini-nlp: I didn't want configs to have a `partial_rotary_factor` by default, unless they actually use it. With `setdefault` all configs will have a `partial_rotary_factor = None` and it will be serialized as well

src/transformers/models/apertus/configuration_apertus.py
zucchini-nlp: indeed!

src/transformers/models/apertus/configuration_apertus.py
zucchini-nlp: yeah, but do we want the general `__init__` to handle RoPE which is not a general attribute for all models?

ArthurZucker: We can have a RopeConfigMixin, PreTrainedConfig can inherit from it or something -> isolate rope stuff but they are stil common

zucchini-nlp: agreed that isolating is the end goal, lemme see

src/transformers/modeling_rope_utils.py
vasqu: Makes sense, it really should stay optional

src/transformers/models/apertus/configuration_apertus.py
```diff
@@ -160,15 +160,9 @@ def __init__(
...
-        rope_scaling = kwargs.pop("rope_scaling", None)
-        self.rope_parameters = rope_scaling or rope_parameters
-
-        # Validate the correctness of rotary position embeddings parameters
-        rope_theta = kwargs.get("rope_theta", 12000000.0)
-        standardize_rope_params(self, rope_theta=rope_theta)
-        rope_config_validation(self)
+        self.rope_parameters = rope_parameters
 
+        kwargs = self.convert_rope_params_to_dict(default_theta=12000000.0, **kwargs)
```
ArthurZucker: `default_theta` in args maybe and you end up not having to call `convert_rope_params_to_dict` in here at all

ArthurZucker (approved):
`RotaryEmbeddingConfigMixin` is my Christmas gift! ty its a lot better

zucchini-nlp:
run-slow: llama, gemma3, qwen2, qwen2_vl, mistral, mixtral, modernbert, llava

zucchini-nlp:
Doc-builder and weight tying tests will fail but are not related

PR #42261, opened 2025-11-18
[testing] fix `cwm`

ydshieh:
# What does this PR do?

Missing tokens


ydshieh:
run-slow: cwm

ydshieh:
Will merge. As discussed internally, the weight loading / offloading at this moment is still under some work and the fixes in this PR still fails with the current main.

However, it works with the commit c4cfc2e0

PR #42262, opened 2025-11-18
Add quantize for missing methods

SunMarc:
# What does this PR do?

This adds supports for quants methods that needs to add Quantize class. From what I saw, hqq will be the hardest to deal with ... (not fixed at all for now). The other will only require to check if the tests are passing.


SunMarc:
@bot /style

src/transformers/quantizers/quantizer_quark.py
```diff
@@ -94,3 +95,22 @@ def is_serializable(self, safe_serialization=None):
...
+
+                postfix = target_key.split(".")[-1]
+
+                if postfix in CHECKPOINT_KEYS:
+                    target_key = target_key.replace(postfix, CHECKPOINT_KEYS[postfix])
+
+                return {target_key: value}
+
+        # TODO: Using this class for now but the best would be to use a weight converter directly
+        return QuarkQuantize(self)
```
SunMarc, line 117: indeed, this would be much simpler with a weight converter. This will also break as we don't call quantize when the model is already quantized

src/transformers/core_model_loading.py
```diff
@@ -805,7 +805,9 @@ def convert_and_load_state_dict_in_model(
                             mapping.distributed_operation,
                             hf_quantizer,
                         )
-            except SkipLayer:
+            except SkipLayer as e:
+                print(e)
```
ArthurZucker, line 809: We could / can also just update https://github.com/huggingface/transformers/blob/1491367f12b8586740ffdfb1db8dedcce0387b84/src/transformers/utils/loading_report.py#L235 

To add at the end the full error? Might be cleaner?

src/transformers/quantizers/quantizer_quark.py
ArthurZucker: Yep happy to have a weight converter!

SunMarc:
not meant to be merged, we will add those methods back progressively !

PR #42263, opened 2025-11-18
[docs] Deploying

stevhliu:
This PR focuses on several improvements to the serving and deploying docs:

- consolidates integration guides and ExecuTorch/ONNX export guides to reduce navigation overhead and group related content together
- improve scope of the `serving.md` doc to only the `transformers serve` CLI. clearer intro on when to use it vs optimized inference engines, more complete endpoint descriptions (`v1/audio/transcriptions` and `v1/models`), MCP example with `GenerationConfig`, port forwarding, and better structure for optimizations


LysandreJik:
I think the "serving" docs will tend to grow so keeping it in its section would be good imo; would it be possible to split this PR into two so I can review the serving docs independently and we can iterate on them? The deploying docs look good so we can split it in two to accelerate the merge

stevhliu:
sounds good, i've split off the `transformers serve` docs in a separate PR!

this PR only focuses on consolidating executorch/onnx into a "exporting to production" doc in the models section which i think is a more natural fit

i've also reverted changes to the transformers as a modeling backend doc. i'll build that one out separately into its own community integrations section as discussed on slack 🙂

LysandreJik (approved):
Thank you Steven! Looks good to me in its current form

PR #42264, opened 2025-11-18
Fix accelerate integration

Cyrilvallez:
# What does this PR do?

The `device_map` computation is currently broken, after https://github.com/huggingface/transformers/pull/42043 switched to using an integration of the accelerate functions to simplify them and use the already available `all_tied_weights_keys` instead of computing them again and again. But it does not compute the double list of `tied_parameters` correctly, in the format that the `accelerate` functions are used to, i.e. groups of all similar parameters.
Also, now that we don't tie weights before device_map computation, we need to exclude them explicitly from the loop in `compute_model_sizes`, as they will be present in the iterated params.

This PR fixes it. This is needed in https://github.com/huggingface/transformers/pull/42242, which is where I noticed it.
Also fixes a more long-standing issue, where `compute_model_sizes` was not taking non-persistent buffers into account, but they actually take some space!

### DIFF
--- a/src/transformers/integrations/accelerate.py
+++ b/src/transformers/integrations/accelerate.py
@@ -227,7 +227,7 @@ def compute_module_sizes(
 
     tied_keys = getattr(model, "all_tied_weights_keys", {}).keys()
     for name, param in iterator:
-        # Do not count tied keys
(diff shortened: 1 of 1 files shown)


SunMarc (approved):
Thanks for fixing this !

PR #42265, opened 2025-11-18
add sequential sampling

jiosephlee:
# What does this PR do?

Adds Sequential Sampling! All huggingface trainers default to random sampling, and there is no way to allow a sequential sampling, which is valuable when the user has a custom data curriculum, for instance.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -432,7 +432,7 @@ class Transpose(ConversionOps):
             tensor = input_dict.get(key, [])
             if len(tensor) != 1:
                 raise ValueError(f"Transpose conversion requires exactly one tensor, found {len(tensor)}.")
-            output[target_pattern] = torch.transpose(tensor[0], dim0=self.dim0, dim1=self.dim1)
+            output[target_pattern] = torch.transpose(tensor[0], dim0=self.dim0, dim1=self.dim1).contiguous()
         return output
 
     @property
--- a/tests/models/timm_backbone/test_modeling_timm_backbone.py
+++ b/tests/models/timm_backbone/test_modeling_timm_backbone.py
@@ -155,6 +155,10 @@ class TimmBackboneModelTest(ModelTesterMixin, BackboneTesterMixin, PipelineTeste
     def test_save_load(self):
         pass
 
+    @unittest.skip(reason="Only checkpoints on timm can be loaded into TimmBackbone")
+    def test_load_contiguous_weights(self):
+        pass
+
     @unittest.skip(reason="TimmBackbone uses its own `from_pretrained` without device_map support")
     def test_can_load_with_device_context_manager(self):
         pass
--- a/tests/test_modeling_common.py
+++ b/tests/test_modeling_common.py
@@ -933,6 +933,23 @@ class ModelTesterMixin:
                     )
                     self.assertTrue(len(load_result.unexpected_keys) == 0)
 
+    def test_load_contiguous_weights(self):
+        """
+        Checks whether the loaded weights are contiguous or not; inherently checking whether a conversion
+        operation from `core_model_loading` may have affected the original weights.
+        """
+        for model_class in self.all_model_classes:
+            config, _ = self.model_tester.prepare_config_and_inputs_for_common()
+
+            model = model_class(config)
+            self.assertTrue(all(param.is_contiguous() for param in list(model.parameters())))
+
+            with tempfile.TemporaryDirectory() as tmpdirname:
+                model.save_pretrained(tmpdirname)
+
+                model = model_class.from_pretrained(tmpdirname)
+                self.assertTrue(all(param.is_contiguous() for param in list(model.parameters())))
+
     def test_gradient_checkpointing_backward_compatibility(self):
         config, inputs_dict = self.model_tester.prepare_config_and_inputs_for_common()
 


Rocketknight1:
cc @sunmarc

SunMarc:
Hey @jiosephlee, I don't mind adding this but did you manage to do an experiment where this leads to better training overall ?

jiosephlee:
Hey @SunMarc ! Sequential sampling in and of itself wouldn't improve training, but yes, I'm doing research currently where a well-designed curriculum can lead to better performance. There are also prior works that support this e.g. https://arxiv.org/abs/2405.07490. A common way of organizing data in this field is from easy to hard: https://dl.acm.org/doi/abs/10.1145/3589335.3641257

SunMarc:
Maybe we can first gauge the interest from the community before merging this ? TrainingArgs is already super bloated so I prefer to not add any args that have low usage. Could you create an issue so that we see if folks are interested in ?

jiosephlee:
@SunMarc, that totally makes sense. I just created an issue, and I can re-bump this PR once it's garnered some interest. 

I know folks like  @qgallouedec have wanted this as well, if this helps clarify the need for this feature (correct me if I'm wrong, Quentin!)

qgallouedec:
Thznks!

Yes, to give more context:

Currently, if I want to sample my data in order, I can't. The PR in the current state is a solution, but in my opinion, sequential sampling should be the default.

If we agree on this, I see two options

1. Add a parameter such as `random_sampling` (`False` by default) that would enable random sampling. But that would add another argument to `TrainingArguments`.

2. (Better, in my opinion) Simply replace the random sampler with a sequential sampler, and if the user wants random sampling, they can simply shuffle their dataset beforehand, which amounts to the same thing. No new arg.

In both cases it would change the default behavior of the trainer

jiosephlee:
@qgallouedec I will +1 on the default behavior change. The default behavior is likely unexpected for many, especially for those coming from PyTorch, which also has the shuffle off as default https://docs.pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader

Concerning (2), there is an issue that would arise from forgoing RandomSampler. RandomSampler has a nice property where you can reset the shuffle at each epoch, and it appears to be used in Trainer.

https://github.com/huggingface/transformers/blob/d6a6c82680cba9c51decdacac6dd6315ea4a766a/src/transformers/trainer.py#L2465

Otherwise, it seems difficult to implement epoch-level shuffling outside of Trainer.

SunMarc:
> Currently, if I want to sample my data in order, I can't. The PR in the current state is a solution, but in my opinion, sequential sampling should be the default.

Can you explain why ? For the training dataset, it should be better to have have random sampling in general no ? 

Maybe as a first set, we can create a new arg called `train_sampler` and we remove that the same time `group_by_length` arg. In this new arg, we will have three choices `random` (default), `group_by_length` and `sequential` wdyt ? Since v5 is not out yet, if we are quick enough, we won't have to do dep cycle regarding `group_by_length` arg. cc @jiosephlee 

Happy new year btw ;)

jiosephlee:
@SunMarc happy new year :) 

And that's a pretty clean suggestion! I took a stab at refactoring it based on what you said. I took the liberty of changing one of the ValueErrors to a warning because the IterableDataset incompatibility seems to be an issue for any Sampler, afaik, which is why samplers are entirely skipped in line 1037.

qgallouedec:
Yes, that sounds good to me too. With one small reservation about the name. Because this argument would not take a sampler, but rather a strategy. So maybe `sampling_strategy`?

qgallouedec:
> Can you explain why ? For the training dataset, it should be better to have random sampling in general no ?

Yes indeed. The motivation was mainly to avoid the surprise caused by the fact that the PyTorch loader is sequential by default, but shuffling by default is also a good default.

src/transformers/trainer.py
```diff
@@ -670,9 +670,11 @@ def __init__(
         if (
             train_dataset is not None
             and isinstance(train_dataset, torch.utils.data.IterableDataset)
-            and args.group_by_length
         ):
-            raise ValueError("the `--group_by_length` option is only available for `Dataset`, not `IterableDataset")
+            logger.warning(
+                f"The `train_sampler='{args.train_sampler}'` option is ignored when using an `IterableDataset`. "
+                "Samplers cannot be used with IterableDataset as they require indexed access to the dataset."
+            )
```
SunMarc: This is indeed true. sampler field will be ignored when using `IterableDataset`. Since this behavior is silently ignored by torch, maybe we can just throw logger.info instead of a warning.

src/transformers/training_args.py
```diff
@@ -576,13 +576,20 @@ class TrainingArguments:
...
-            Whether or not to group together samples of roughly the same length in the training dataset (to minimize
-            padding applied and be more efficient). Only useful if applying dynamic padding.
+        train_sampler (`str`, *optional*, defaults to `"random"`):
+            The sampler to use for the training dataloader. Possible values are:
+
+                - `"random"`: Uses `RandomSampler` (default).
+                - `"sequential"`: Uses `SequentialSampler`.
+                - `"group_by_length"`: Uses `LengthGroupedSampler` to group samples of roughly the same length
+                  together (to minimize padding and be more efficient). Requires the dataset to not be an
+                  `IterableDataset`.
```
SunMarc:
```suggestion
                  together (to minimize padding and be more efficient).
```

src/transformers/training_args.py
```diff
@@ -576,13 +576,20 @@ class TrainingArguments:
             for a full list of optimizers.
         optim_args (`str`, *optional*):
             Optional arguments that are supplied to optimizers such as AnyPrecisionAdamW, AdEMAMix, and GaLore.
-        group_by_length (`bool`, *optional*, defaults to `False`):
-            Whether or not to group together samples of roughly the same length in the training dataset (to minimize
-            padding applied and be more efficient). Only useful if applying dynamic padding.
+        train_sampler (`str`, *optional*, defaults to `"random"`):
+            The sampler to use for the training dataloader. Possible values are:
```
SunMarc: maybe specify here instead that it only works when the dataset is not an interabledataset

src/transformers/training_args.py
```diff
@@ -576,13 +576,20 @@ class TrainingArguments:
             for a full list of optimizers.
         optim_args (`str`, *optional*):
             Optional arguments that are supplied to optimizers such as AnyPrecisionAdamW, AdEMAMix, and GaLore.
-        group_by_length (`bool`, *optional*, defaults to `False`):
-            Whether or not to group together samples of roughly the same length in the training dataset (to minimize
-            padding applied and be more efficient). Only useful if applying dynamic padding.
+        train_sampler (`str`, *optional*, defaults to `"random"`):
```
SunMarc: as quentin suggested, maybe we should switch to `train_sampling_strategy`

SunMarc (commented):
Thanks ! Just a few nits

jiosephlee:
Moved this to new PR #43138

PR #42270, opened 2025-11-19
Correctly create tied key mapping in post_init, and dynamic tie weight

Cyrilvallez:
As we rely more and more in `self.all_tied_weight_keys` everywhere (i.e. the list of tied keys obtained during `post_init`) for multiple manipulations (device_map computation, cuda warmup, post-processing of `from_pretrained`...), it becomes very important that the (few) models containing regex patterns for their `_tied_weights_keys` mapping have the patterns expanded to fit in `all_tied_weight_keys` as well, instead of containing simple patterns that are skipped in different ways for all downstream application.
This PR fixes that, by expanding correctly at `post_init` time, so the mapping are correct params everywhere.
Also allows for recomputing this mapping in `tie_weights` dynamically, so that it is correct if calling `tie_weights` after having modified the config

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -2350,9 +2350,9 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
                 if isinstance(submodule, PreTrainedModel):
                     # Will dynamically check the config if it has changed
                     submodel_tied_weights = submodule.get_expanded_tied_weights_keys(all_submodels=False)
-                    expanded_tied_weights.update(
-                        {f"{prefix}.{k}": f"{prefix}.{v}" for k, v in submodel_tied_weights.items()}
-                    )
+                    if prefix != "":
+                        submodel_tied_weights = {f"{prefix}.{k}": f"{prefix}.{v}" for k, v in submodel_tied_weights.items()}
+                    expanded_tied_weights.update(submodel_tied_weights)
             return expanded_tied_weights
 
         tied_mapping = self._tied_weights_keys
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -1499,8 +1499,8 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         # Attach the different parallel plans and tied weight keys to the top-most model, so that everything is
         # easily available
         self._tp_plan, self._ep_plan, self._pp_plan = {}, {}, {}
-        # Current submodel should register its tied weights keys only if the config is asking for it
-        self.all_tied_weights_keys = self.get_expanded_tied_weights_keys(recurse=False)
+        # Current submodel should register its tied weights
+        self.all_tied_weights_keys = self.get_expanded_tied_weights_keys(all_submodels=False)
         # If current model is a base model, attach `base_model_tp_plan` and `base_model_pp_plan` from config
         if self.base_model is self:
             self._pp_plan = self.config.base_model_pp_plan.copy() if self.config.base_model_pp_plan is not None else {}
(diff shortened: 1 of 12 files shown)


src/transformers/modeling_utils.py
```diff
@@ -2341,7 +2338,73 @@ def smart_apply(self, fn):
         # Let the magic happen with this simple call
         self.smart_apply(self._initialize_weights)
 
-    def tie_weights(self, missing_keys: Optional[set[str]] = None):
+    def get_expanded_tied_weights_keys(self, all_submodels: bool = False) -> dict:
+        """
+        Return the expanded tied weight keys (in case they contain modules or regex patterns) for only the current
+        model, or recursively for all submodels if `all_submodels=True` (i.e. it will re-check the config values for all
+        submodels).
+        """
```
ArthurZucker: can you add a reduced example of what that gives for Llama for example?

src/transformers/modeling_utils.py
```diff
@@ -2341,7 +2338,73 @@ def smart_apply(self, fn):
...
+        if all_submodels:
+            expanded_tied_weights = {}
+            for prefix, submodule in self.named_modules():
```
ArthurZucker: You don't ignore duplicates here

src/transformers/modeling_utils.py
```diff
@@ -2341,7 +2338,73 @@ def smart_apply(self, fn):
...
+                if isinstance(submodule, PreTrainedModel):
+                    # Will dynamically check the config if it has changed
+                    submodel_tied_weights = submodule.get_expanded_tied_weights_keys(all_submodels=False)
```
ArthurZucker, line 2396: don't know if we really have to go the inheritance path here?

ArthurZucker: given that we do `named_parameters` afterwards

src/transformers/modeling_utils.py
```diff
@@ -2390,50 +2453,36 @@ def __init__(self):
...
-
-        # TODO let's pray this is not too slow :)
-        top_level_params = dict(self.named_parameters(remove_duplicate=False)) | dict(
-            self.named_buffers(remove_duplicate=False)
-        )
-        for target_name, source_name in mapping.items():
-            source_name = "^" + source_name
-            target_name = "^" + target_name
+        # In this case, the keys stored in `all_tied_weights_keys` are already correct
+        if not recompute_mapping:
```
ArthurZucker, line 2461: to update with `setter` and `getter` for `tie_words_embedding` no?

ArthurZucker (commented):
missing a bit of reprensentative doc! Let's take a t5 as example ? Or rtdetr? to ahve a complexe list

src/transformers/modeling_utils.py
Cyrilvallez: Yes, in order to check the proper subconfig... No better way unfortunately as sometimes we cannot get the subconfig in a proper way

src/transformers/modeling_utils.py
Cyrilvallez: No, was already checked before!

ArthurZucker (approved):
Thanks for itiretaing ! i like that its explicit now!

lugimzzz:
@Cyrilvallez @ArthurZucker 
I have a concern: with the changes in this PR, a model like Qwen/Qwen3-0.6B — which is a tie-weight model — is now treated as a non–tie-weight model. In its config.json, "tie_word_embeddings": true is set, and lm_head.weight and embed_tokens.weight actually share identical values. However, after this PR, the model is no longer recognized as a tie-weight model.

As a result, when training Qwen/Qwen3-0.6B, the gradients update completely different parameters before and after this PR. I believe the original behavior is more consistent with the model’s intended design.
Reference model: https://huggingface.co/Qwen/Qwen3-0.6B

Cyrilvallez:
Hey @lugimzzz! You can check-out my answer to the same question [here](https://github.com/huggingface/transformers/issues/42832#issuecomment-3646657225)! Let me know if you want more clarification!

PR #42280, opened 2025-11-19
Fix validation checks order in benchmark_v2

Abdennacer-Badaoui:
- Reorder the validation checks in the benchmarker to avoid TypeError when required arguments are missing.
- Fix `adapt_configs()` preserving old benchmark names (e.g., s128_n128) when parameters are changed via command-line arguments. Remove the name field from the config dict so it gets re-inferred with the updated sequence length and token generation values. 

----
@remi-or


benchmark_v2/framework/benchmark_config.py
```diff
@@ -203,6 +203,8 @@ def adapt_configs(
             config["sequence_length"] = seqlen
             config["num_tokens_to_generate"] = ntok
             config["gpu_monitoring"] = monitor
+            # Remove the old name so it gets re-inferred with the updated values
+            config.pop("name", None)
```
remi-or, line 207: good catch

remi-or (approved):
LGTM!

PR #42284, opened 2025-11-19
Remove timm backbone weight loading to prevent meta-tensor warnings in `from_pretrained`

HichTala:
# What does this PR do?

This pull request makes a targeted change to the model loading logic in `modeling_utils.py` to improve compatibility with timm backbones. Specifically, it ensures that when a model is configured to use a timm backbone with pretrained weights (`use_pretrained_backbone=True`), the backbone weights are not loaded while the model is still in `meta` device during `from_pretrained()`.

**Model initialization and configuration:**

* In `from_pretrained`, before model instantiation, if the config specifies both `use_timm_backbone` and `use_pretrained_backbone`, the code sets `config.use_pretrained_backbone` to `False` to defer loading pretrained timm backbone weights until after initialization.

Fixes # (issue)
- #37615 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@Rocketknight1 @qubvel @CyrilVallez @NielsRogge @xenova


Cyrilvallez:
Hey @HichTala! Are you sure you are still facing the original issue in latest `main`? Loading has changed, and we do not use `model.load_state_dict` anymore!

HichTala:
Hi @Cyrilvallez! I just tested this again with the latest `main`, and the original issue is still present.

The problem is not related to `load_state_dict` itself. It comes from the fact that during `from_pretrained`, when a model uses a `timm` backbone with `use_pretrained_backbone=True`, the model is still instantiated on meta devices, but the `timm` pretrained backbone weights are already being loaded at that stage. This causes all the warnings from `timm`, and then, once `from_pretrained` continues, those backbone weights are overwritten anyway by the Hugging Face checkpoint loading logic.

This PR simply removes loading the `timm` backbone weights until after the model is fully materialized in `from_pretrained` method.

Cyrilvallez:
I see... Then this config switch should be hardcoded in all model config when this can happen, not in the general loading code! Do you want to make the change?

HichTala:
Yep totally agree with that, I'm making the change!

benchmark_v2/framework/benchmark_config.py
```diff
@@ -224,8 +224,6 @@ def adapt_configs(
             config["sequence_length"] = seqlen
             config["num_tokens_to_generate"] = ntok
             config["gpu_monitoring"] = monitor
-            # Remove the old name so it gets re-inferred with the updated values
-            config.pop("name", None)
```
Cyrilvallez, line 228: unrelated! To remove!

src/transformers/models/conditional_detr/configuration_conditional_detr.py
```diff
@@ -161,7 +161,7 @@ def __init__(
         auxiliary_loss=False,
         position_embedding_type="sine",
         backbone="resnet50",
-        use_pretrained_backbone=True,
+        use_pretrained_backbone=False,
```
Cyrilvallez: Let's not change the default value, just hardcode the change if we meet the same condition as before! (i.e. using timm backbone, and using pretrained one)

HichTala: I may be misunderstanding something, so let me restate it to be sure: you’re suggesting that we simply hard-code the same condition as before directly in the config file, correct?

If that’s the case, I don’t think this approach is the solution. The problem appears specifically when using a timm backbone with `use_pretrained_backbone=True` only in the `from_pretrained` method. But it’s perfectly valid to use a pretrained timm backbone without loading pretrained weights for the rest of the architecture, so we should keep that option open for users.

benchmark_v2/framework/benchmark_config.py
HichTala: Oops, I think I misclicked and reverted a commit that shouldn’t have been reverted…

src/transformers/models/conditional_detr/configuration_conditional_detr.py
Cyrilvallez: Ohh, I see, I did not know this was a thing... Is it really a real-world scenario that we should suport? Meaning, do you as a user load only timm backbone? Seems fairly weird to me

HichTala: I was initially thinking about fine-tuning a fully pretrained object detection model on a dataset with a different number of classes, but this case is already properly handled by the Hugging Face framework. So yes, we can ignore that specific scenario here

HichTala:
Hi, sorry for the delay, I completely forgot about this PR. Here are the updates

Cyrilvallez:
Hey! I believe this PR is superseded by https://github.com/huggingface/transformers/pull/43323

HichTala:
Ok cool I'm closing this then

HichTala:
superseded by https://github.com/huggingface/transformers/pull/43323

PR #42288, opened 2025-11-19
Update torchcodec to match torchaudio version

remi-or:
There is an issue in the latest AMD container: along with ROCm, `torch` was updated, so the `torchcodec` version does not match anymore. This PR fixes this, which in turn fixes `whisper` tests, and I imagine other audio model's tests.


ydshieh (approved):
Sure.

(Just FYI: if you want to update to torch 2.9.1 )

remi-or:
> (Just FYI: if you want to update to torch 2.9.1 )

No ROCm container for that 2.9.1 for now sadly :( we can try a manual update though

PR #42289, opened 2025-11-19
[core] Fix torchao 

MekkCyber:
# What does this PR do?

Refactors torchao quantization method to use conversion ops instead of the classical `create_quantized_param`


src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -530,3 +530,37 @@ def set_metadata(self, checkpoint_files: list[str]):
...
+                WeightConverter(
+                    source_keys= ["weight:qdata", "weight:scale", "weight:zero_point"],
+                    target_keys= "weight",
+                    operations=[TorchAoDeserialize(self)],
+                ),
+                WeightConverter(
+                    source_keys= ["weight:_data"],
+                    target_keys= "weight",
+                    operations=[TorchAoDeserialize(self)],
+                ),
```
jerryzh168, line 559: cc @liangel-02

src/transformers/integrations/torchao.py
```diff
@@ -0,0 +1,249 @@
...
+        missing_keys=None,
+        **kwargs,
+    ) -> dict[str, torch.Tensor]:
+        from torchao.quantization import quantize_
+
+        _, value = tuple(input_dict.items())[0]
+        value = value[0] if isinstance(value, list) else value
+
+        if not self.hf_quantizer.param_needs_quantization(model, full_layer_name):
+            return {full_layer_name: value}
```
ArthurZucker: quantize op is not added if it does not require quantization in the core no?

src/transformers/integrations/torchao.py
```diff
@@ -0,0 +1,249 @@
...
+        module._parameters[tensor_name] = torch.nn.Parameter(value, requires_grad=value.requires_grad).to(value.device)
+        # if we are quantizing tied parameters, to avoid tying the quantized weights
+        # the correct order to do it is
+        # 1. load the weight to model
+        # 2. run tie_weights to populate the weights
+        # 3. quantize
+        input_embed = model.get_input_embeddings()
+        if self.hf_quantizer.quantization_config.untie_embedding_weights and id(module) == id(input_embed):
+            model.tie_weights()
+            setattr(model.config.get_text_config(decoder=True), "tie_word_embeddings", False)
```
ArthurZucker, line 108: mmm that's not optimal the best is IMO to update config.tiew_word_embeddings directly post init or something? the weights should not be tied at this point in time

src/transformers/integrations/torchao.py
```diff
@@ -0,0 +1,249 @@
...
+                f"{full_layer_name}:scale": input_dict["weight:scale"][0]
+                if isinstance(input_dict["weight:scale"], list)
+                else input_dict["weight:scale"],
+            }
+            if is_int_4:
+                param_data[f"{full_layer_name}:zero_point"] = (
+                    input_dict["weight:zero_point"][0]
+                    if isinstance(input_dict["weight:zero_point"], list)
+                    else input_dict["weight:zero_point"]
+                )
```
ArthurZucker: let's have poroper if else please to make it more readable

src/transformers/integrations/torchao.py
```diff
@@ -0,0 +1,249 @@
...
+        self.hf_quantizer = hf_quantizer
+
+    def convert(
+        self,
+        input_dict: dict[str, torch.Tensor],
+        model: Optional[torch.nn.Module] = None,
+        full_layer_name: str | None = None,
+        missing_keys=None,
+        **kwargs,
+    ) -> dict[str, torch.Tensor]:
```
ArthurZucker, line 217: documentation with example of in / out will be much much welcome

ArthurZucker: like the figures I shared!

src/transformers/integrations/torchao.py
```diff
@@ -0,0 +1,249 @@
...
+        from torchao.quantization import quantize_
```
ArthurZucker, line 90: same here expected in and transformations happening explanation will be welcome!

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -530,3 +531,34 @@ def set_metadata(self, checkpoint_files: list[str]):
...
+                WeightConverter(
+                    source_keys=["weight:_data"],
+                    target_keys="weight",
+                    operations=[TorchAoDeserialize(self)],
+                ),
+                # used for unsafe serialization
+            ]
+        else:
+            return [
+                WeightConverter(
```
ArthurZucker: you don't need this one!

src/transformers/core_model_loading.py
```diff
@@ -675,8 +698,8 @@ def convert_and_load_state_dict_in_model(
                 source_pattern = renamed_key
 
             # 5. Handle dtype casting
-            if hf_quantizer is not None and hf_quantizer.param_needs_quantization(model, renamed_key):
-                mapping.quantization_operation = hf_quantizer.get_quantize_ops()
+            # if hf_quantizer is not None and hf_quantizer.param_needs_quantization(model, renamed_key):
+            #     mapping.quantization_operation = hf_quantizer.get_quantize_ops()
```
ArthurZucker: the issue with not using this is that any renaming pattern that matches some keys won't use the quantizer weight converter. It's a lot safer to just always have the op added to the mapping!

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -530,3 +531,34 @@ def set_metadata(self, checkpoint_files: list[str]):
                     metadata.update(metadata_)
             # Save it
             self.metadata = metadata
+
+    def get_quantize_ops(self):
+        from ..integrations.torchao import TorchAoQuantize
+
+        return TorchAoQuantize(self)
+
+    def get_weight_conversions(self):
```
ArthurZucker, line 545: yep verynice

ArthurZucker: we might want a update conversion func to update previous weight converter with :data :zeros etc

ArthurZucker (commented):
Very nice. We agree that this one is the 1st we need then a 2nd: if the model is pre_quantized but also has weight converter -> update the weight converters to be applied on the quantized keys

src/transformers/quantizers/quantizer_torchao.py
MekkCyber: I'm not sure I get it, we need it to quantize

src/transformers/core_model_loading.py
```diff
@@ -113,6 +119,8 @@ def convert(
         source_keys: list[str],
         target_keys: list[str],
         full_layer_name: str,
+        model,
+        missing_keys,
         config,
         **kwargs,
```
SunMarc, line 121: most of the args should be optional kwargs so that we can clean the other convert function with **kwargs and only put args that are being used but that's fine, we should do that later on

src/transformers/integrations/torchao.py
```diff
@@ -0,0 +1,245 @@
...
+
+        _, value = tuple(input_dict.items())[0]
+        value = value[0] if isinstance(value, list) else value
+
+        module, tensor_name = get_module_from_name(model, full_layer_name)
```
SunMarc, line 95: I don't know if it is better to use the layer name from `input_dict` or not ? I feel like this might be better ?

src/transformers/integrations/torchao.py
SunMarc: yeah as discussed on slack, we should return the `embedding` and the `lm_head` here if they don't want us to tie them.

src/transformers/integrations/torchao.py
SunMarc: Not sure if this is worth spending too much time on that bc torchao serialization don't work yet, maybe in v0.15 so this won't be used and can't be easily tested e2e

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -237,6 +238,8 @@ def update_unexpected_keys(self, model, unexpected_keys: list[str]) -> list[str]
         return [k for k in unexpected_keys if not any(k.endswith(x) for x in self.full_ao_keys)]
 
     def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **kwargs) -> bool:
+        if self.pre_quantized:
+            return False
```
SunMarc, line 245: this is typically one of the cases where we would get the wrong numel calculation since we are skipping them. We should try to fix that at some point, as this should be quite simple

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -530,3 +533,27 @@ def set_metadata(self, checkpoint_files: list[str]):
...
+            return [
+                WeightConverter(
+                    source_keys=["weight:qdata", "weight:scale", "weight:zero_point"],
+                    target_keys="weight",
+                    operations=[TorchAoDeserialize(self)],
+                ),
+                WeightConverter(
+                    source_keys=["weight:_data"],
+                    target_keys="weight",
+                    operations=[TorchAoDeserialize(self)],
```
SunMarc, line 558: A `WeightRename` should be enough in this case no ?

src/transformers/integrations/torchao.py
SunMarc: the safe serialization don't work yet because of torchao, so it is fine to just clean a bit, we can come back to that later on

SunMarc (approved):
A few nits but overall fine !

src/transformers/integrations/torchao.py
MekkCyber: the input dict doesn't contain the layer_name, it contains what you gathered, something like `weight:qdata`

SunMarc: oh yeah indeed, makes sense

src/transformers/quantizers/quantizer_torchao.py
MekkCyber: yes both are fine i guess

src/transformers/integrations/torchao.py
```diff
@@ -0,0 +1,257 @@
...
+        # If it's a bias, no need to do anything special (except removing the ":_data" part of the key, but was
+        # already done) - if it's unsafe-serialized (i.e. not safetensors), not need for anything either
+        if is_unsafe_serialization:
+            return {full_layer_name: weight}
+        # Sanity check for the new serialization format
+        elif not (TORCHAO_VERSION >= version.parse("0.14.0") and is_metadata_torchao(self.hf_quantizer.metadata)):
+            # print("metadata", self.hf_quantizer.metadata)
+            raise ValueError("To use `safetensors` serialization, you should have `torchao>=0.14.0` installed")
+
+        new_param = unflatten_tensor_state_dict(param_data, self.hf_quantizer.metadata)[full_layer_name]
```
liangel-02, line 267: in a followup pr, we can modify this to work with all tensor subclasses and for sharded checkpoint files.

im thinking that in this `convert` function, we load in the tensor subclass components (ie `_weight_qdata`) as module parameters. after all files are loaded, we can delete them and replace the actual layer weights with the reconstructed quantized tensors. 

see #41998 for details - will this approach still work with the new refactoring? cc @jerryzh168

src/transformers/quantizers/quantizer_torchao.py
jerryzh168: FYI we just changed `weight:_data` to `weight_qdata` so these things can be attached to module directly incase we need it in the future. https://github.com/pytorch/ao/commit/ba3ac9f2f6117ba35ff28fbb8811f61ad992dfcf

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -530,3 +533,27 @@ def set_metadata(self, checkpoint_files: list[str]):
...
+
+        return TorchAoQuantize(self)
+
+    def get_weight_conversions(self):
+        from ..integrations.torchao import TorchAoDeserialize
+
+        if self.pre_quantized:
+            return [
+                WeightConverter(
+                    source_keys=["weight:qdata", "weight:scale", "weight:zero_point"],
```
jerryzh168, line 551: nit: maybe also add [`weight_qdata`, `weight_scale`] as well since zero_point may be optional, like https://github.com/pytorch/ao/blob/2ff1eb2e356275cfbe46832327387d382c72945d/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L99

src/transformers/integrations/torchao.py
jerryzh168: @liangel-02 yeah I think our original approach should still work, I guess it's fine to land this PR first and you can re-open https://github.com/huggingface/transformers/pull/41998 on top of these new changes, since you are more familiar with this part

src/transformers/quantizers/quantizer_torchao.py
ArthurZucker: Weight converter is better than `WeightRename` here because there is an op!

src/transformers/integrations/torchao.py
```diff
@@ -0,0 +1,257 @@
+import importlib.metadata
```
ArthurZucker, line 16: missing a licence

src/transformers/integrations/torchao.py
ArthurZucker: well done 🎐 !

src/transformers/integrations/torchao.py
ArthurZucker: Thanks both for chiming in! 🤗

src/transformers/core_model_loading.py
ArthurZucker: agreed

src/transformers/modeling_utils.py
```diff
@@ -3830,6 +3830,8 @@ def from_pretrained(
                 weight_conversions.extend(
                     [WeightRenaming(source_keys=k, target_keys=v) for k, v in key_mapping.items()]
                 )
+            if hf_quantizer is not None:
+                weight_conversions.extend(hf_quantizer.get_weight_conversions())
```
ArthurZucker, line 3893: nice!

ArthurZucker (approved):
Great work! thanks 🤗

src/transformers/quantizers/quantizer_torchao.py
MekkCyber: let's do that in a follow up pr since the safetensors support is broken with the latest torchao version

PR #42290, opened 2025-11-19
Fix device_map computation part 2

Cyrilvallez:
# What does this PR do?

Follow up to https://github.com/huggingface/transformers/pull/42264.
Since https://github.com/huggingface/transformers/pull/42043, `compute_module_sizes` is also used in `_init_infer_auto_device_map`. However, there it expects to have all the modules inside, but also the leaves parameters and buffers, which was not the case. Because we return `defaultdict` instance, we would not have any `KeyError` but instead very nasty silent bugs because the values found would always be 0 (the default of the defaultdicts.

This PR fixes it by adding the option for adding them in the dict. Doing `compute_module_sizes(model, only_modules=False)` is now equivalent to `accelerate.compute_module_sizes(model)`, which is not the case without the flag (leaves are not in the dict)

cc @SunMarc

### DIFF
--- a/src/transformers/integrations/accelerate.py
+++ b/src/transformers/integrations/accelerate.py
@@ -204,13 +204,18 @@ def check_and_set_device_map(device_map: "torch.device | int | str | dict | None
 
 
 def compute_module_sizes(
-    model: "PreTrainedModel", hf_quantizer: "HfQuantizer | None" = None, buffers_only: bool = False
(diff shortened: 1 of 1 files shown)


SunMarc (approved):
Indeed, this was indeed strange that in this version of `compute_module_sizes` it returned two values. Thanks for fixing this nasty bug !

PR #42295, opened 2025-11-20
Fix an edge case for `get_encoder()`

zucchini-nlp:
# What does this PR do?

As per title, fixes https://github.com/huggingface/transformers/pull/42156#issuecomment-3553776255 and adjusts the test case slightly

I didn't not change the setter, we don't call `set_encoder` unless we're sure there is an encoder right?


BenjaminBossan (approved):
Thanks for fixing this quickly.

src/transformers/modeling_utils.py
```diff
@@ -2232,7 +2232,11 @@ def get_encoder(self, modality: Optional[str] = None):
                 return getattr(self, name)
 
         if self.base_model is not self and hasattr(self.base_model, "get_encoder"):
-            return self.base_model.get_encoder(modality=modality)
+            base_encoder = self.base_model.get_encoder(modality=modality)
+            # Base model will always have attr `get_encoder` if inherited from `PreTrainedModel`
+            # But it doesn't mean that the model has an encoder module, and we need to return `self`
+            if base_encoder != self.base_model:
+                return base_encoder
```
molbap, line 2037: Nice! I thought the `get_encoder` method has a method to ignore if self, could we add it there after?

tests/utils/test_modeling_utils.py
```diff
@@ -3241,7 +3241,7 @@ def test_decoder_only_model_returns_self(self):
             num_hidden_layers=2,
             num_attention_heads=4,
         )
-        model = MistralModel(cfg)
+        model = MistralForCausalLM(cfg)
```
molbap, line 3236: alright! QQ - why is this test only checking MistralForCausalLM?

molbap (approved):
SGTM, thanks!

tests/utils/test_modeling_utils.py
zucchini-nlp: we want to check that a generative decoder-only model doesn't return its base model as encoder, Just a tiny chnage

src/transformers/modeling_utils.py
zucchini-nlp: Not sure I got it

molbap: basically this check `if base_encoder != self.base_model:` could be in the get_encoder method, no?

tests/utils/test_modeling_utils.py
molbap: sure!

src/transformers/modeling_utils.py
zucchini-nlp: I think we're talking about different things haha

It's in `get_encoder()` now, isn't it?

molbap: lol yes indeed. Don't know what I was looking at, thought it was for a specific model, my bad

zucchini-nlp: that happens, will merge then

zucchini-nlp:
CI is not feeling good today

zucchini-nlp:
My last try to pass the flaky test 🤞🏻

PR #42296, opened 2025-11-20
Tiny doc fix

molbap:
# What does this PR do?

Checked out the contributing guide, found 2 dead links

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,9 +125,9 @@ If you're contributing a **vision-language model** (or any multimodal model that
 All new models should use the modular architecture pattern. Create a `modular_<model_name>.py` file using the modular model converter:
 
 - Use the CLI, [`transformers add-new-model-like`](https://github.com/huggingface/transformers/blob/main/src/transformers/cli/add_new_model_like.py) to generate a modular skeleton and get started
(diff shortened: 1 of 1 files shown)


zucchini-nlp (commented):
Hmm, the links work for me and lead to correct sections from [guidelines](https://huggingface.co/docs/transformers/main/en/contributing#vision-language-model-contribution-checklist) 🤔  Do we prefer a full path for some reason?

molbap:
it's that it does not work from https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md is all and it tends to be my first entrypoint :laughing: and most other links there do work,  haven't checked all though

zucchini-nlp:
Ah I see, though it will stop working from docs after this PR because it has to be without the `.md` suffix for it 🥲  Not sure what's the correct way to make it work in both places


https://moon-ci-docs.huggingface.co/docs/transformers/pr_42296/en/contributing#do-you-want-to-implement-a-new-model

molbap:
no idea either, cc @stevhliu here please :pray:

stevhliu:
Hmm, I think we can either:

- remove the symlink between the `CONTRIBUTING.md` on the repo and the docs
- modify the doc-builder to accept the `.md` extension (this is probably the best since it requires less maintenance)

wdyt?

molbap:
Yes why not go for the doc-builder change!

stevhliu (approved):
LGTM, going to merge now as the doc-builder change has also been merged

PR #42299, opened 2025-11-20
Auto convert tekken.json

ArthurZucker:
# What does this PR do?
1. If `mistral-common` is not installed, we can always convert to `tokenizer.json`
2. Don't convert if a tokenizer.json is here
3. Fix the tokenizer if its affected by the regex issue. 

```python
In [3]: tok = AutoTokenizer.from_pretrained("mistralai/Mistral-Small-3.1-24B-Instruct-2503")
The tokenizer you are loading from 'mistralai/Mistral-Small-3.1-24B-Instruct-2503' with an incorrect regex pattern: https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503/discussions/84#69121093e8b480e709447d5e.  This will lead to incorrect tokenization. You should set the `fix_mistral_regex=True` flag when loading this tokenizer to fix this issue.

In [4]: tok = AutoTokenizer.from_pretrained("mistralai/Mistral-Small-3.1-24B-Instruct-2503", fix_mistral_regex=True)
# no warning
In [2]: tok = AutoTokenizer.from_pretrained("mistralai/Mistral-Small-3.1-24B-Instruct-2503", fix_mistral_regex=False)
# no warning either
In [3]: tok.fix_mistral_regex
Out[3]: False

In [4]: tok = AutoTokenizer.from_pretrained("mistralai/Mistral-Small-3.1-24B-Instruct-2503")
The tokenizer you are loading from 'mistralai/Mistral-Small-3.1-24B-Instruct-2503' with an incorrect regex pattern: https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503/discussions/84#69121093e8b480e709447d5e.  This will lead to incorrect tokenization. You should set the `fix_mistral_regex=True` flag when loading this tokenizer to fix this issue.

In [5]: tok.fix_mistral_regex
Out[5]: False
```

Superseed #41592 and #41718 for now, fixes #41553, fixes #42283


src/transformers/tokenization_utils_base.py
```diff
@@ -2417,6 +2424,9 @@ def _from_pretrained(
                 "Special tokens have been added in the vocabulary, make sure the associated word embeddings are"
                 " fine-tuned or trained."
             )
+        if tokenizer.vocab_size > 100000:
+            # Try to catch mistral tokenizers.
+            pass # TODO
```
ArthurZucker: @patrickvonplaten that's were I think it would make sense to post-process the `regex`. I just want to avoid downloading too many files for everyone (especially the config) so any common denomitors are welcome!

src/transformers/models/auto/tokenization_auto.py
```diff
@@ -750,8 +750,8 @@
             (
                 "MistralCommonTokenizer"
                 if is_mistral_common_available()
-                else ("LlamaTokenizer" if is_sentencepiece_available() else None),
-                "LlamaTokenizerFast" if is_tokenizers_available() and not is_mistral_common_available() else None,
+                else ("PreTrainedTokenizerFast" if is_tokenizers_available() else None),
+                "PreTrainedTokenizerFast" if is_tokenizers_available() and not is_mistral_common_available() else None,
```
Cyrilvallez, line 752: Usually the pattern is `(slow, fast)`, here it's `(fast, fast)`, not sure if intended
Maybe it should be instead:
```python
(None, "MistralCommonTokenizer" if is_mistral_common_available() else ("PreTrainedTokenizerFast" if is_tokenizers_available() else None)
```
so that we never have slow one anyway?

src/transformers/tokenization_utils_base.py
```diff
@@ -2417,6 +2424,22 @@ def _from_pretrained(
...
+                if model.tags is not None:
+                    if re.search("base_model:.*mistralai", "".join(model.tags)):
+                        return True
+                return False
+
+            if is_base_mistral(pretrained_model_name_or_path) and not kwargs.get("fix_regex"):
+                logger.warning(
+                    f"The tokenizer you are loading from '{pretrained_model_name_or_path}'"
+                    f" with an old regex pattern. This will lead to incorrect tokenization."
+                )
```
Cyrilvallez: Why warn if we fix it anyway? IMO no need to add `fix_regex` kwarg, and we can simply fix

src/transformers/tokenization_utils_base.py
```diff
@@ -2417,6 +2424,22 @@ def _from_pretrained(
...
                 " fine-tuned or trained."
             )
+        if tokenizer.vocab_size > 100000 and getattr(tokenizer.backend_tokenizer, "pre_tokenizer", None) is not None:
+            from huggingface_hub import model_info
+            def is_base_mistral(model_id: str) -> bool:
+                model = model_info(model_id)
+                if model.tags is not None:
+                    if re.search("base_model:.*mistralai", "".join(model.tags)):
+                        return True
+                return False
```
Cyrilvallez: Do we have access to anything else allowing to know if it's a Mistral tokenizer? I.e. maybe the `config` architecture or something? Would be more general than hub tags IMO

src/transformers/tokenization_utils_fast.py
```diff
@@ -207,6 +207,8 @@ def __init__(self, *args, **kwargs):
             if tokens:
                 self.add_tokens(tokens)
 
+        # there was an issue with mistral models where the pre_tokenizer's regex pattern
+        # is not correct. Here we try to fix it.
```
Cyrilvallez: Is this intended to be added? Were we already trying to do something?

Cyrilvallez (approved):
Nice! Just left a few comments!

src/transformers/tokenization_utils_base.py
```diff
@@ -2417,6 +2424,22 @@ def _from_pretrained(
...
+
+            if is_base_mistral(pretrained_model_name_or_path) and not kwargs.get("fix_regex"):
+                logger.warning(
+                    f"The tokenizer you are loading from '{pretrained_model_name_or_path}'"
+                    f" with an old regex pattern. This will lead to incorrect tokenization."
```
patrickvonplaten:
```suggestion
                    f" with an incorrect regex pattern. This will lead to incorrect tokenization on some edge cases. See issue: ....."
```
let's maybe say incorrect no?

ArthurZucker: yep yep absolutely

src/transformers/tokenization_utils_base.py
ArthurZucker: hub tag is the most reliable AFAIK the key is to avoid downloading the config so this filters first then we download it

src/transformers/tokenization_utils_base.py
```diff
@@ -2417,6 +2430,66 @@ def _from_pretrained(
                 "Special tokens have been added in the vocabulary, make sure the associated word embeddings are"
                 " fine-tuned or trained."
             )
+        if tokenizer.vocab_size > 100000 and getattr(tokenizer.backend_tokenizer, "pre_tokenizer", None) is not None:
+            from huggingface_hub import model_info
+
+            def is_base_mistral(model_id: str) -> bool:
+                model = model_info(model_id)
+                if model.tags is not None:
+                    if re.search("base_model:.*mistralai", "".join(model.tags)):
```
patrickvonplaten, line 2450: Hmm so that's only for mistral org no? Should we directly check of `model_type in ["mistral" ....] so that it also works for other orgs?

ArthurZucker: We can't do that until we download the config / config is there

src/transformers/tokenization_utils_base.py
```diff
@@ -2417,6 +2432,75 @@ def _from_pretrained(
...
+
+                    if transformers_version and version.parse(transformers_version) <= version.parse("4.57.2"):
+                        if _is_local and _config.model_type not in [
+                            "mistral",
+                            "mistral3",
+                            "voxstral",
+                            "ministral",
+                            "pixtral",
+                        ]:
+                            return tokenizer
```
CISC, line 2478: The non-existent attribute use of `_config.model_type` is causing massive loading failures everywhere (including CIs), please consider making a hotfix ASAP. :)

thinkahead: Change it to `_config.get("model_type")`

ArthurZucker: yeah sorry

ArthurZucker: I have no idea why the CI was full green

PR #40936, opened 2025-09-17
rm slow tokenizers

itazap:
## Tokenization

Just as we moved towards a single backend library for model definition, we want `Tokenizer` to be a lot more intuitive.
With v5, you can now initialize an empty `LlamaTokenizer` and train it directly on your new task! 

Defining a new tokenizer object should be as simple as this:
```python
from transformers import TokenizersBackend, generate_merges
from tokenizers import pre_tokenizers, Tokenizer
from tokenizers.model import BPE

class Llama5Tokenizer(TokenizersBackend):
    def __init__(self,        unk_token="<unk>",bos_token="<s>", eos_token="</s>", vocab=None, merges=None ):
        if vocab is None:
            self._vocab = {
                str(unk_token): 0,
                str(bos_token): 1,
                str(eos_token): 2,
            }

        else:
            self._vocab = vocab

        if merges is not None:
            self._merges = merges
        else:
            self._merges = generate_merges(filtered_vocab)

        self._tokenizer = Tokenizer(
            BPE(vocab=self._vocab, merges=self._merges, fuse_unk=True)
        )
        self._tokenizer.pre_tokenizer = pre_tokenizers.Metaspace(
            replacement="▁", prepend_scheme=_get_prepend_scheme(self.add_prefix_space, self), split=False
        )
        super().__init__(
            tokenizer_object=self._tokenizer,
            unk_token=unk_token,
            bos_token=bos_token,
            eos_token=eos_token,
        )
```

And now if you call `Llama5Tokenizer()` you just get an empty, trainable tokenizer that follows the definition of the authors of `Llama5` (it does not exist yet :wink:).

The above is the main motivation towards refactoring tokenization: we want people to just instantiate a tokenizer like they would a model, empty or not and with exactly what they defined.

### Non-tokenizers
If you tokenizers is not common, or you just don't want to rely on `sentencepiece` nor `tokenizers` you can just import the `PythonBackend` (previousl `PreTrainedTokenzier`) which has all the API and logic for added tokens, encoding and decoding wieht them etc. 

If you want to have en less features, you can use the common `PreTrainedTokenizerBase` mixin, which mostly defines `transformers` tokenizer API: `encode`, `decode`, `vocab_size`, `get_vocab`, `convert_tokens_to_ids`, `convert_ids_to_tokens`, `from_pretrained`, `save_pretrained`, etc.

### Backend Architecture Changes

**Moving away from "slow" vs "fast" tokenizers:**

[...]

PR #42300, opened 2025-11-20
Fix parse_response after tokenizer refactor

Rocketknight1:
Since we're likely to be updating the schemas used by `parse_response`, this PR removes the documentation references to it for now. We're keeping it as a softlaunched prototype feature, and will re-add documentation and do an official launch once it's stabilized!


Rocketknight1:
Also re-adding `Tokenizer.parse_response()` which got accidentally nuked in #40936

Rocketknight1:
Also guarding the checks for `tokenizer.response_schema`, since that attribute was also nuked in https://github.com/huggingface/transformers/pull/40936

ydshieh (approved):
Thanks.

(still be good to add what was the issue: the title is "fix something" but the description don't mention what was the issue ... )

PR #42301, opened 2025-11-20
Remove error string test that was failing

Rocketknight1:
`test_use_safetensors` was failing in the CI after #41580 because different error messages are returned in some cases, and this test inspected specific error message text.

This PR keeps the rest of the test, but removes the specific text check.

cc @ArthurZucker @ydshieh


ydshieh (approved):
If we know the list of possible messages, we could still test it.

But with the `self.assertRaises(OSError)` itself, it's OK for me.

Thanks

Rocketknight1:
I think it's fine - I don't think we really care about the exact message contents!

ArthurZucker (commented):
ty, the error keeps changing it seems

PR #42302, opened 2025-11-20
Remove outdated methods in modeling_utils.py

Cyrilvallez:
# What does this PR do?

As per the title. Do some cleanup. They are very out-dated and not useful anymore.


ArthurZucker (approved):
time to also remove model.parallel

PR #42305, opened 2025-11-20
Allow LayoutLMV3Processor to accept rescale_factor

Rocketknight1:
`test_image_processor_defaults_preserved_by_image_kwargs` is flaky for models that use the `LayoutLMV3Processor`, which causes annoying intermittent CI failures, especially for `UDOP`. The cause is that `LayoutLMV3Processor` expects `rescale_value`, not the standard `rescale_factor`, but the test depends on `rescale_factor` being accepted. This PR allows `LayoutLMV3Processor` to accept `rescale_factor`, but I'm not sure if this is the optimal solution. cc @molbap @yonigozlan @zucchini-nlp 

Note that no other models use `rescale_value` - it's just LayoutLMV3! I suspect this issue was also the cause of several other flaky tests in `test_processing_common`

### DIFF
--- a/src/transformers/models/layoutlmv3/image_processing_layoutlmv3.py
+++ b/src/transformers/models/layoutlmv3/image_processing_layoutlmv3.py
@@ -188,10 +188,7 @@ class LayoutLMv3ImageProcessor(BaseImageProcessor):
         self.size = size
         self.resample = resample
         self.do_rescale = do_rescale
-        if "rescale_factor" in kwargs:
-            self.rescale_factor = kwargs["rescale_factor"]
-        else:
-            self.rescale_factor = rescale_value
+        self.rescale_factor = kwargs.get("rescale_factor", rescale_value)
         self.do_normalize = do_normalize
         self.image_mean = image_mean if image_mean is not None else IMAGENET_STANDARD_MEAN
         self.image_std = image_std if image_std is not None else IMAGENET_STANDARD_STD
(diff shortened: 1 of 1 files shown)


Rocketknight1:
run-slow: layoutlmv3, udop

src/transformers/models/layoutlmv3/image_processing_layoutlmv3.py
```diff
@@ -188,7 +188,7 @@ def __init__(
         self.size = size
         self.resample = resample
         self.do_rescale = do_rescale
-        self.rescale_factor = rescale_value
+        self.rescale_factor = kwargs.get("rescale_factor", rescale_value)
```
molbap: Ok! would add a small comment like "only model to use `rescale_value`, needed for tests"

molbap (approved):
Thanks! Seems good, anything to remove flaky tests

src/transformers/models/layoutlmv3/image_processing_layoutlmv3.py
Rocketknight1: Should we refactor to make `rescale_factor` the actual argument here and keep `rescale_value` as a legacy kwarg? cc @zucchini-nlp or maybe @NielsRogge, I'm not sure why it has a different name in the first place

molbap: For what it's worth, I checked from the hub and preprocessor_configs seem to never store either. I don't know why it has this name, and think either option is safe, waiting to see what others think

NielsRogge: Yes, I think `rescale_value` was added before the image processors were standardized to use `rescale_factor`.

Rocketknight1: In that case I think I'm going to rename the processor arg to `rescale_factor` for standardization, but allow `rescale_value` in the kwargs so we don't break backward compatibility!

PR #42309, opened 2025-11-20
Add buffers to `_init_weights` for ALL models

Cyrilvallez:
# What does this PR do?

## The issue at hand

Currently, during loading we initialize the model on meta device before loading weights, thanks to `init_empty_weights` from `accelerate`.
However, this context manager has BIG drawbacks:
- everything (every parameter and buffer) is first materialized on cpu before being moved to meta device
- this is extremely inefficient of course, as we only want them on meta -> it wastes time and memory
- all buffers stay on cpu (even the persistent ones, that we are loading again after anyway, so don't need to be there...)

This was done like this because BEFORE THIS PR we had no way to reinit non-persistent buffers such as RoPE modules (`XXXRotaryEmbedding` modules) if they were put on meta device.
This PR removes this problem, which will allow me to correctly put everything directly on meta device in a future PR (for convenience as this one is already massive)

## What this PR is doing

This is the first big step towards initializing the full models on meta device natively before loading!
- add the buffers to `_init_weights` for ALL models in the library
- add a test to ensure it is enforced for future models as well
- `_init_weights` is now a complete parameterization of all the model's internal states (parameters AND buffers), allowing to always get back a fully correctly initialized model, independently of any movement to the meta device

Only this change already allows very cool stuff, such as getting back a fully working and initialized model after putting it manually on meta device such as:

```python
with torch.device("meta"):
    model = model_class(config)
# move everything back to cpu (it puts random values everywhere)
model.to_empty(device="cpu")
# Run all the inits, including the buffers -> all buffers are correct here, including ROPE etc
model.init_weights()
```

This was previously impossible, as buffers were NOT re-initialized, so for example ROPE modules would be fully broken, leading to untrainable models.
It can now be trained directly after a simple `init_weights()` call


src/transformers/models/blt/modeling_blt.py
```diff
@@ -551,47 +552,55 @@ def _init_weights(self, module):
...
+            init.trunc_normal_(
                 module.weight,
                 mean=0.0,
                 std=std,
                 a=-3 * std,
                 b=3 * std,
             )
             if module.bias is not None:
-                nn.init.zeros_(module.bias)
+                init.zeros_(module.bias)
```
ArthurZucker, line 592: that's myy bad I reviewied and told him to use nn.init cuz I thought we still had the decorator that updates nn.init for the context of the init no?

src/transformers/models/blenderbot_small/modeling_blenderbot_small.py
```diff
@@ -430,6 +431,11 @@ class BlenderbotSmallPreTrainedModel(PreTrainedModel):
     _supports_flex_attn = True
     _can_compile_fullgraph = True
 
+    def _init_weights(self, module):
+        super()._init_weights(module)
```
ArthurZucker, line 435: if super handled parameters we should not really need that no?

src/transformers/models/donut/modeling_donut_swin.py
```diff
@@ -802,6 +802,18 @@ def _init_weights(self, module):
         elif isinstance(module, DonutSwinSelfAttention):
             init.zeros_(module.relative_position_bias_table)
 
+            coords_h = torch.arange(module.window_size[0])
```
ArthurZucker: this is now written twice, here and in the init of the class which we probably want to create a separate func imo

src/transformers/models/gemma3n/modular_gemma3n.py
```diff
@@ -1893,8 +1894,53 @@ def _init_weights(self, module):
...
+            lower_causal_mask = torch.tril(
+                torch.ones((module.context_size, module.chunk_size), dtype=torch.bool), diagonal=0
+            ).T
+            upper_causal_mask = torch.tril(
+                torch.ones((module.chunk_size, module.context_size), dtype=torch.bool),
+                diagonal=module.max_past_horizon + module.max_future_horizon,
+            )
+            local_causal_valid_mask = torch.ones((module.chunk_size, module.context_size), dtype=torch.bool)
+            local_causal_valid_mask = local_causal_valid_mask * lower_causal_mask * upper_causal_mask
+            init.copy_(module.local_causal_valid_mask, local_causal_valid_mask)
```
ArthurZucker: this shows the limitiation! very dirty to have 100loc 😢

tests/test_modeling_common.py
```diff
@@ -1063,26 +1053,144 @@ def seeded_initialize_weights(self, module):
...
+                v2 = from_pretrained_state_dict[k1]
                 # Since we added the seed, they should be exactly the same (i.e. using allclose maybe be wrong due
                 # to very low std in init function)
                 if not (v1 == v2).all():
-                    different_weights.append(k1)
+                    different_weights.add(k1)
 
             # Buffers that are initialized randomly are ignored as they are not initialized on meta device anyway
             buffer_names = {name for name, _ in model_from_config.named_buffers()}
-            different_weights = [k for k in different_weights if k not in buffer_names and "timm" not in k]
```
ArthurZucker, line 1081: haha you saw the Timm escape 🤣

tests/test_modeling_common.py
```diff
@@ -1063,26 +1053,144 @@ def seeded_initialize_weights(self, module):
...
+                # Add it to the traceback
+                traceback = (
+                    f"`{buf_name}` in module `{immediate_parent_class}` called from `{pretrained_parent_class}`\n"
+                )
+                unique_bad_module_traceback.add(traceback)
+
+            unique_bad_module_traceback = "".join(unique_bad_module_traceback)
+            self.assertTrue(
+                len(different_buffers) == 0,
+                f"The following buffers are not properly handled in `_init_weights()`:\n{unique_bad_module_traceback}",
```
ArthurZucker: make the comment as helpful as possible think about ssomeone adding a model that has no idea what this is

ArthurZucker (approved):
Kinda against this TBH, it shows that we do need granularity in how buffers are initialized to only write it explicitly once. 


But for the sake of perf I don't mind. 

Let's minimize tho, nn.Parameters should just be handeled by super by default please

src/transformers/models/blt/modeling_blt.py
Cyrilvallez: Yes, we still have it but I switched it just for the sake of using our inits everywhere!

src/transformers/models/blenderbot_small/modeling_blenderbot_small.py
Cyrilvallez: In this case, `final_logits_bias` is a buffer that is only present for a few models... So did not add it to the general init to avoid cluttering

Cyrilvallez:
Created some externals functions to avoid duplication for the ones that require several steps!
In general, I STRONGLY AGREE that `_init_weights` is far from ideal, and that something closer to the `reset_parameters`/`reset_buffers` of torch at the module-level would be far simpler! Let's keep it in mind!

PR #42311, opened 2025-11-20
Fix: Guard against None num_query_tokens in Blip2Processor (to avoid TypeError)

### LINKED ISSUE #42203: `num_query_tokens` of blip2 might be None
### System Info

transformers>=4.55.0

### Who can help?

_No response_

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

Hi, the code below did not check if the `self.num_query_tokens` is None, but in the `__init__` method, the default value for this variable is None, which might cause error like:
```
TypeError: unsupported operand type(s) for -: 'int' and 'NoneType'
```

https://github.com/huggingface/transformers/blob/6f6095e0cf509f7384d3ce0c1804013ef6cafd5f/src/transformers/models/blip_2/processing_blip_2.py#L109-L110

I tried kpyu/video-blip-opt-2.7b-ego4d in transformers>=4.55.0 and it failed.

### Expected behavior

The model can be successfully used.

Flakes342:
# What does this PR do?

Fixes #42203 

## Before submitting
      Pull Request section?
      to it if that's the case.

### Fixed a bug where Blip2Processor assumes num_query_tokens is an int and does `max_length - self.num_query_tokens` which raises TypeError when `num_query_tokens` is None.

This change:
- Treats num_query_tokens as 0 when it is None (with a warning).
- Adds a minimal unit test to prevent regression.

## Who can review?
@Rocketknight1


Rocketknight1:
cc @zucchini-nlp! I think it's fine without the warning message, but deferring to her on what behaviour we want from processors :sweat_smile:

zucchini-nlp (commented):
@Flakes342 thanks for the PR!

To support older models from the hub which weren't updated, I think we should update the default value of `num_query_tokens`. 

The modeling code assumes that `input_ids` have special image placeholders and with `self.num_query_tokens=None` the placeholders will not be added in the input text. Therefore the model will not see an input image

Let's add the default value of `32` which is the only value used in all official checkpoints. Ping me for another review when ready :)

Flakes342:
@zucchini-nlp thr pr is ready for your review

zucchini-nlp:
run-slow: blip_2

src/transformers/models/blip_2/processing_blip_2.py
zucchini-nlp: Sorry if I wasn't clear, I meant the default value in the signature here

zucchini-nlp:
run-slow: blip_2

zucchini-nlp:
run-slow: blip_2

zucchini-nlp:
Can you also fix the modeling code to be in line with the changes? In some places we check if the model config was updated or not with `if self.config.image_token_index is not None` but after this PR all model inputs will be prepared in "updated" format. So we can assume it safely in modeling code 

It can be later tested with `RUN_SLOW=1 pytest tests/models/blip_2/test_modeling_blip_2.py::Blip2ModelIntegrationTest::test_inference_itm`

Finally, you need to run `make fix-copies` and `make style` to pass the CI checks

zucchini-nlp:
@bot /style

PR #41992, opened 2025-11-03
[PoC] HF exporters

IlyasMoutawwakil:
[CI](https://transformers-ci.lor-e.huggingface.cool/d/pytest-observability-by-pr/pytest-observability-branch?var-pr=41992)

# HfExporters: Native, Unified export for PyTorch / ONNX / ExecuTorch

thumbnail

> **Edit**: pieces of this PR were merged separately ([#42697](https://github.com/huggingface/transformers/pull/42697), [#42317](https://github.com/huggingface/transformers/pull/42317)), so this one is now focused on the **HfExporters** machinery itself.

## What this adds

A native, in-Transformers export pipeline — one base class (`HfExporter`), three subclasses for the runtimes we care about, one unified API:

| Exporter | Output | Runtime |
|---|---|---|
| `DynamoExporter` | `ExportedProgram` | Any PyTorch runtime, AOT compilation |
| `OnnxExporter` | `ONNXProgram` | Any ONNX runtime (ORT, TensorRT, OpenVINO, …) |
| `ExecutorchExporter` | `ExecutorchProgramManager` | Mobile and edge (ExecuTorch) |

Same call shape across all three. Dynamic shapes by default. Generation-style models split automatically into prefill + decode (+ vision/audio sub-encoders for VLMs).

```python
from transformers import AutoModelForMaskedLM, AutoTokenizer
from transformers.exporters import OnnxExporter, OnnxConfig

model_id = "hf-internal-testing/tiny-random-BertForMaskedLM"
tokenizer = AutoTokenizer.from_pretrained(model_id)
model = AutoModelForMaskedLM.from_pretrained(model_id).eval()
inputs = tokenizer(["Hello, my dog is cute"] * 2, return_tensors="pt")
onnx_program = OnnxExporter().export(model, inputs, config=OnnxConfig(dynamic=True))

new_input = tokenizer("Hello, my cat is so adorable!", return_tensors="pt")
torch.testing.assert_close(
    onnx_program.call_reference(**new_input)[0],   # numpy reference
    onnx_program(**new_input)[0],                  # onnxruntime
    rtol=1e-4, atol=1e-4,
)
```

Swap one line for another runtime — `DynamoExporter()` / `DynamoConfig` or `ExecutorchExporter()` / `ExecutorchConfig(backend=...)`.

For generative models the prefill/decode split is captured automatically:

```python
from transformers import AutoModelForCausalLM, AutoTokenizer
from transformers.exporters import OnnxExporter, OnnxConfig

[...]

PR #27007, opened 2023-10-23
Fuyu: improve image processing

molbap:
# What does this PR do?

This PR aims at aligning the FuyuImageProcessor class with other vision/language models within transformers. Fuyu model expects a tensor of token ids, a tensor of patch embeddings, and an indexing tensor indicating where to put rows of patch embeddings into the token embeddings, separated by the input ids. Currently the image processor does not separate the steps necessary to achieve this output in the Processor. It also limits the inference size to batches of size 1. It also aims at improving readability and code quality of the processor to possibly enable pipelining later on.

Pending tasks:

- [x] Return a `BatchFeature` with arbitrary batch size
- [x] add `do_rescale`, `do_normalize`, `do_pad` arguments in the `ImageProcessor` constructor
- [x] align patch-ification methods to ViTMAE and possibly pix2struct
- [x] rework and refactor method `process_images_for_model_input`, currently hard to read
- [x] test long images, stretched images, usual processor edge cases
- [x] test images and no text, text and no image in Processor class leveraging tokenizer + ImageProcessor 

## Before submitting
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests),
      Pull Request section?
- [x] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [x] Did you write any new necessary tests?

## Who can review?

Models:

- text models: @ArthurZucker
- vision models: @amyeroberts

PR #42317, opened 2025-11-21
Make Transformers more torch-exportable and dynamo-friendly

IlyasMoutawwakil:
# What does this PR do?

First proposals include:
- `check_with(error_type, cond, lambda: msg)` instead of `if cond: raise error_type(msg)`, which also works with torch.export/torch.compile to hint to the compiler that the condition is expected to be true at export/compile time.
- vectorization of some loops / comprehension lists into traceable, optimized and non-blocking versions.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


src/transformers/models/colqwen2/modeling_colqwen2.py
```diff
@@ -149,11 +149,10 @@ def forward(
...
             # NOTE: image_grid_thw: (batch_size, 3) where image_grid_thw[i] = (num_patches_h, num_patches_w, temporal_patch_size)
-            offsets = image_grid_thw[:, 1] * image_grid_thw[:, 2]  # (num_patches_h, num_patches_w)
-            pixel_values = torch.cat(
-                [pixel_sequence[:offset] for pixel_sequence, offset in zip(pixel_values, offsets)],
-                dim=0,
-            )  # (num_patches_h * num_patches_w, pixel_values)
+            offsets = image_grid_thw[:, 1] * image_grid_thw[:, 2]  # (batch_size,)
+            arange = torch.arange(pixel_values.shape[1], device=offsets.device)  # (max_len,)
+            mask = arange.unsqueeze(0) < offsets.unsqueeze(1)  # (batch_size, max_len)
+            pixel_values = pixel_values[mask]  # (total_valid_patches, channels, height, width)
```
IlyasMoutawwakil, line 154: avoiding looping over tensor

src/transformers/models/fuyu/modeling_fuyu.py
```diff
@@ -124,11 +124,7 @@ def get_image_features(self, pixel_values: torch.FloatTensor, **kwargs):
             pixel_values (`torch.FloatTensor` of shape `(batch_size, num_channels, image_size, image_size)`):
                 The tensors corresponding to the input images.
         """
-        patch_embeddings = [
-            self.vision_embed_tokens(patch.to(self.vision_embed_tokens.weight.dtype)).squeeze(0)
-            for patch in pixel_values
-        ]
-        return patch_embeddings
+        return self.vision_embed_tokens(pixel_values)
```
IlyasMoutawwakil, line 124: need opinion about this

IlyasMoutawwakil:
taken from #41992

src/transformers/utils/import_utils.py
```diff
@@ -1279,15 +1283,34 @@ def is_jit_tracing() -> bool:
...
+        cond (`Any`): The condition to check (`bool`, `SymBool` or `torch.Tensor`).
+        msg (`Callable[[], str]`): A callable that returns the message to display if the check fails.
+    Raises:
+        error_type: If the condition is not met.
+    """
+    import torch
+
+    if isinstance(cond, torch.Tensor):
+        cond = cond.item()
+    torch._check_with(error_type, cond, msg)
```
IlyasMoutawwakil: could be interesting to dispatch to `torch._assert_async` to avoid cpu-gpu syncs, with something like `TRANSFORMERS_USE_ASSERT_ASYNC=1`, however these assertions can only raise `RuntimeError: CUDA error: device-side assert triggered` https://github.com/pytorch/pytorch/issues/131491

src/transformers/models/fuyu/modeling_fuyu.py
echarlaix: cc @molbap maybe (looks like this was added in https://github.com/huggingface/transformers/pull/27007)

echarlaix (approved):
LGTM, thanks a lot @IlyasMoutawwakil !

src/transformers/models/fuyu/modeling_fuyu.py
molbap: Don't know why I'm seeing this only now :older_man: from what I remember `pixel_values` for that model is a `list` of `Tensor`s hence the weird list comp, if tests pass however it should be ~ok!

examples/modular-transformers/modeling_new_task_model.py
```diff
@@ -271,12 +271,13 @@ def get_placeholder_mask(
...
 
         n_image_tokens = special_image_mask.sum()
-        special_image_mask = special_image_mask.unsqueeze(-1).expand_as(inputs_embeds).to(inputs_embeds.device)
         n_image_features = image_features.shape[0] * image_features.shape[1]
-        if inputs_embeds[special_image_mask].numel() != image_features.numel():
-            raise ValueError(
-                f"Image features and image tokens do not match: tokens: {n_image_tokens}, features {n_image_features}"
-            )
+        special_image_mask = special_image_mask.unsqueeze(-1).expand_as(inputs_embeds).to(inputs_embeds.device)
+        check_with(
```
ArthurZucker: I think this needs a better name! something that says "torch_compile_check" something explicit for users as to why we use this!

src/transformers/models/colqwen2/modeling_colqwen2.py
ArthurZucker: very nic eindeed!

examples/modular-transformers/modeling_new_task_model.py
```diff
@@ -274,12 +274,13 @@ def get_placeholder_mask(
...
+            ValueError,
+            inputs_embeds[special_image_mask].numel() == image_features.numel(),
+            lambda: f"Image features and image tokens do not match, tokens: {n_image_tokens}, features: {n_image_features}",
+        )
```
ArthurZucker: Given that you defined the function check with I think we should not have to use lambda here

examples/modular-transformers/modeling_test_detr.py
```diff
@@ -429,10 +429,11 @@ def forward(
         batch_size, num_queries, _ = hidden_states.shape
         batch_size, sequence_length, _ = encoder_hidden_states.shape
         total_elements = sum(height * width for height, width in spatial_shapes_list)
-        if total_elements != sequence_length:
-            raise ValueError(
-                "Make sure to align the spatial shapes with the sequence length of the encoder hidden states"
-            )
+        check_with(
+            ValueError,
```
ArthurZucker: We should put the value error as a default because it seems to be used everywhere this way the more common cases were checked with function is used will be simplified

src/transformers/models/mm_grounding_dino/modeling_mm_grounding_dino.py
```diff
@@ -1783,33 +1784,42 @@ def generate_masks_with_special_tokens_and_transfer_map(input_ids: torch.LongTen
...
+    attention_mask = identity | attention_mask
 
-        previous_col = col
+    # Position IDs: distance from previous special token
+    position_ids = indices - prev_special - 1
+    position_ids = torch.where(valid_block, position_ids, torch.zeros_like(position_ids))
+    position_ids = torch.clamp(position_ids, min=0).to(torch.long)
 
-    return attention_mask, position_ids.to(torch.long)
+    return attention_mask, position_ids
```
ArthurZucker, line 1821: very nice work here!

src/transformers/models/zamba/modeling_zamba.py
```diff
@@ -491,23 +491,23 @@ def slow_forward(self, input_states, cache_params: ZambaHybridDynamicCache = Non
...
                 hidden_states = self.act(hidden_states).to(dtype).unsqueeze(-1)
             else:
-                if attention_mask is not None and not torch.all(attention_mask == 1):
+                if attention_mask is not None:
                     hidden_states = hidden_states * attention_mask[:, -hidden_states.shape[-1] :].unsqueeze(1)
                 conv_state = nn.functional.pad(hidden_states, (self.conv_kernel_size - hidden_states.shape[-1], 0))
                 cache_params.conv_states[self.layer_idx] = conv_state
                 hidden_states = self.act(self.conv1d(hidden_states)[..., :seq_len])
-                if attention_mask is not None and not torch.all(attention_mask == 1):
+                if attention_mask is not None:
```
ArthurZucker, line 500: this change is weird same for the next one in this file

src/transformers/utils/import_utils.py
ArthurZucker: yeah... that does sound good actually but only if we can catch to give a good detailed error!

ArthurZucker (approved):
My main comment is to use good default when you define the checking function this way most of the cases were use it or gonna be very simple. 

Otherwise would be nice to ducment the good practices that you expose here, and potentially add a test in `make repo-fix` for simple rules. 

Great work!

examples/modular-transformers/modeling_new_task_model.py
IlyasMoutawwakil: I will name it `torch_compilable_check` as it is compilable without being bound to torch.compile, tell me if it works for you

examples/modular-transformers/modeling_new_task_model.py
IlyasMoutawwakil: yes we can support both str and lambda returning a string (for when we want the message to only be evaluated if cond is false)

src/transformers/models/zamba/modeling_zamba.py
IlyasMoutawwakil: the data-dependency on `not torch.all(attention_mask == 1)` breaks graphs, I can revert the change and try to find better alternatives later (in another PR).

examples/modular-transformers/modeling_test_detr.py
IlyasMoutawwakil: makes sense !

src/transformers/utils/import_utils.py
IlyasMoutawwakil: yeah for now i removed this "feature" until pytorch supports async cuda asserts

src/transformers/models/zamba/modeling_zamba.py
ArthurZucker: no I mean look at the two if else

ArthurZucker:
LGTM, now just the flagged change that looks a bit weird (check the if else)

PR #42318, opened 2025-11-21
Fix: Pass local_files_only from pipeline() to model loading

nandan2003:
Fixes #42269

Fixes a bug where passing `local_files_only=True` to `pipeline()` was ignored, causing models to download from the Hub even when offline mode was explicitly requested.

**The Fix:**
The `local_files_only` argument was being extracted from `kwargs` but was missing from the `hub_kwargs` dictionary that `pipeline()` uses to configure downloads. This PR adds the flag to `hub_kwargs`, ensuring it is correctly propagated to `AutoConfig.from_pretrained` and `load_model`.

## Reproduction & Verification
I verified this locally by clearing the cache for a specific model (`hf-internal-testing/tiny-random-bert`) and attempting to load it via `pipeline()` with `local_files_only=True`.

* **Before Fix:** The model would download despite the flag.
* **After Fix:** The call correctly raises an `OSError` (because the model is not locally available), confirming that no connection to the Hub was attempted.

## Before submitting
- [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Yes, issue #42269.
- [ ] Did you make sure to update the documentation with your changes? (Bug fix, no docs needed).
- [x] Did you write any new necessary tests? (Verified manually via reproduction script).

## Who can review?

@Rocketknight1 (Pipelines)


nandan2003:
Branch updated to sync with main.

I verified this locally by clearing the Hugging Face cache and running a test script:
1. **Before:** `pipeline(..., local_files_only=True)` triggered a download.
2. **After:** It correctly raises `OSError` (indicating offline mode is respected).

The CI failure in `tests_processors` appears unrelated to these changes in `pipelines/__init__.py`.

Wauplin (approved):
Looks good to me, thanks for fixing this!

(might be good to add a test in [./tests/test_pipeline_mixin.py](https://github.com/huggingface/transformers/blob/main/tests/test_pipeline_mixin.py)? -let's wait for a review from a transformers's maintainer :) -)

Rocketknight1:
This PR makes sense to me! One question - can you check in the PEFT path, to see if we need to pass this there too? I'm not sure if the adapter-loading path will respect `local_files_only` either.

nandan2003:
> This PR makes sense to me! One question - can you check in the PEFT path, to see if we need to pass this there too? I'm not sure if the adapter-loading path will respect `local_files_only` either.

You're actually right. made some changes and PTAL.

Rocketknight1:
There's a lot of unrelated style changes here! Can you `pip install -e .[quality]` to make sure you're using the right versions of the style tools, then `make style` or `make fixup`? That should reduce the size of the diff a lot.

src/transformers/integrations/peft.py
```diff
@@ -350,7 +354,10 @@ def load_adapter(
                 # We only want to call prepare_model_for_compiled_hotswap once
                 self._prepare_peft_hotswap_kwargs = None
         else:
-            from peft.utils.hotswap import check_hotswap_configs_compatible, hotswap_adapter_from_state_dict
+            from peft.utils.hotswap import (
+                check_hotswap_configs_compatible,
+                hotswap_adapter_from_state_dict,
+            )
```
Rocketknight1: Still a lot of unrelated style changes, sorry! Can you check against `main` and revert the unrelated changes that are just style like this? Most good IDEs have git compare/diff functions that should make it easy enough to do, but doing it manually shouldn't take too long either.

src/transformers/pipelines/__init__.py
```diff
@@ -701,12 +714,14 @@ def pipeline(
...
     code_revision = kwargs.pop("code_revision", None)
     commit_hash = kwargs.pop("_commit_hash", None)
+    local_files_only = kwargs.get("local_files_only", False)
 
     hub_kwargs = {
         "revision": revision,
         "token": token,
         "trust_remote_code": trust_remote_code,
         "_commit_hash": commit_hash,
+        "local_files_only": local_files_only,
```
Rocketknight1, line 710: Hey, I see the review ping but there's still lots of style issues! If you look at the diff, everything in `__init__.py` except this section is just style changes.

nandan2003: Done. Every other styling change was done by ruff.

Rocketknight1:
@bot /style

Rocketknight1:
There are some CI issues, not the fault of this PR! Hold on for a day or two and we'll resolve them, then ping me to rerun tests

nandan2003:
> There are some CI issues, not the fault of this PR! Hold on for a day or two and we'll resolve them, then ping me to rerun tests

Hey! Is the CI issues sorted?

Rocketknight1 (approved):
Yep, and sorry about that! I'm happy with the changes at this point. Thank you for the PR!

PR #42319, opened 2025-11-21
Fix typo from side_dict to size_dict

nihui:
# What does this PR do?
fix typo


Rocketknight1 (approved):
Yes, thanks for the fix!

PR #42321, opened 2025-11-21
Fix gpt2 modeling tests

Abdennacer-Badaoui:
Fixing gpt2 modeling tests : 

1. `test_batch_generation` & `test_batch_generation_2heads`:
Changed model.config.max_length → model.generation_config.max_length (GPT2Config doesn't have max_length)
2. `test_model_parallelism`:
Added `model_split_percents = [0.5, 0.6, 0.7]` to force model splitting across GPUs (default `[0.5,0.7, 0.9]` was too large). The model doesn’t split starting from 0.8.
3. `test_flash_attn_2_generate_padding_left`:
Updated expected output to match actual generation result

### DIFF
--- a/tests/models/gpt2/test_modeling_gpt2.py
+++ b/tests/models/gpt2/test_modeling_gpt2.py
@@ -438,16 +438,18 @@ class GPT2ModelLanguageGenerationTest(unittest.TestCase):
         output_fa_2 = model.generate(**inputs, max_new_tokens=20, do_sample=False)
         output_fa_2 = tokenizer.batch_decode(output_fa_2)
 
-        expected_output = Expectations({
-            ("cuda", (8, 6)): [
-                "<|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|>hi, who was born in the city of Kolkata, was a member of the Kolkata",
-                "Hello this is a very long sentence. I'm sorry. I'm sorry. I'm sorry. I'm sorry. I'm sorry",
-            ],
-            ("rocm", (9, 4)): [
-                '<|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|>hi, who was also a member of the group, said: "We are very happy to have been',
-                "Hello this is a very long sentence. I'm sorry. I'm sorry. I'm sorry. I'm sorry. I'm sorry",
-            ]
-        }).get_expectation()
+        expected_output = Expectations(
+            {
+                ("cuda", (8, 6)): [
+                    "<|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|>hi, who was born in the city of Kolkata, was a member of the Kolkata",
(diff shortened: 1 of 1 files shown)


tests/models/gpt2/test_modeling_gpt2.py
```diff
@@ -438,7 +439,7 @@ def test_flash_attn_2_generate_padding_left(self):
         output_fa_2 = tokenizer.batch_decode(output_fa_2)
 
         expected_output = [
-            "<|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|>hi, who was born in the city of Kolkata, was a member of the Kolkata",
+            '<|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|>hi, who was also a member of the group, said: "We are very happy to have been',
```
remi-or: I think here would be a good example of where to use `Expectations`  as in here:
https://github.com/huggingface/transformers/blob/f2738ee3756483538c20e10e4f104324675fb406/tests/models/table_transformer/test_modeling_table_transformer.py#L564C1-L567C42
Especially if the test is passing on Nvidia right now

remi-or (commented):
A little to thing to consider when fixing test expected output is the difference in hardware, I left a comment to that effect :)

vasqu:
run-slow: gpt2

tests/models/gpt2/test_modeling_gpt2.py
```diff
@@ -437,10 +438,18 @@ def test_flash_attn_2_generate_padding_left(self):
...
+            {
+                ("cuda", (8, 6)): [
+                    "<|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|>hi, who was born in the city of Kolkata, was a member of the Kolkata",
+                    "Hello this is a very long sentence. I'm sorry. I'm sorry. I'm sorry. I'm sorry. I'm sorry",
+                ],
+                ("rocm", (9, 4)): [
+                    '<|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|><|endoftext|>hi, who was also a member of the group, said: "We are very happy to have been',
+                    "Hello this is a very long sentence. I'm sorry. I'm sorry. I'm sorry. I'm sorry. I'm sorry",
+                ],
+            }
```
vasqu, line 451: So we only added the ROCM expectations, CUDA stayed the same?

vasqu (approved):
LGTM, just checking slow tests just to be sure

tests/models/gpt2/test_modeling_gpt2.py
Abdennacer-Badaoui: Yes, because I see in the Nvidia Flash Attn CI that this test is already passing with the original expectation.
here: https://github.com/huggingface/transformers/actions/runs/19558277650/job/56005466127

vasqu:
Thanks for the PR, there is one multi-gpu test failing but it was also failing before. Now we have less red :D

PR #42325, opened 2025-11-21
[`Attn Masks`] Lift bidirectional mask restriction on eager

vasqu:
Given that the torch compiler fuses our eager attention to sdpa automatically under compile, we should take more advantage and allow no masks on bidirectional masks without padding == fa backend. This restriction is also unnecessary and I'm not sure why I had this one to begin with.

Fuse strategies can be seen here:
- https://github.com/pytorch/pytorch/blob/main/torch/_inductor/fx_passes/fuse_attention.py

### DIFF
--- a/src/transformers/masking_utils.py
+++ b/src/transformers/masking_utils.py
@@ -340,9 +340,6 @@ def sdpa_mask(
         allow_is_causal_skip (`bool`, optional):
             Whether to allow to return `None` for the mask under conditions where we can use the `is_causal` argument in
             `torch.sdpa` instead. Default to `True`.
-        allow_torch_fix (`bool`, optional):
-            Whether to update the mask in case a query is not attending to any tokens, to solve a bug in torch's older
-            versions. We need an arg to skip it when using eager. By default `True`.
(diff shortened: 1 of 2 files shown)


src/transformers/masking_utils.py
```diff
@@ -340,9 +340,6 @@ def sdpa_mask(
         allow_is_causal_skip (`bool`, optional):
             Whether to allow to return `None` for the mask under conditions where we can use the `is_causal` argument in
             `torch.sdpa` instead. Default to `True`.
-        allow_torch_fix (`bool`, optional):
-            Whether to update the mask in case a query is not attending to any tokens, to solve a bug in torch's older
-            versions. We need an arg to skip it when using eager. By default `True`.
```
vasqu, line 345: Doubled comment, unrelated but should be removed nonetheless. Slipped through

ArthurZucker (approved):
is there a test we can add for this? Otherwise lgtm!

vasqu:
Good point, let me check if we properly skip the mask in a test here.

vasqu:
run-slow: bart,bert,llama

PR #39796, opened 2025-07-30
[pipelines] text-to-audio pipeline standardization

gante:
# What does this PR do?

⚠️ TODO before merging: after settling on the design, update pipeline usage in model docs accordingly.

This PR standardizes `text-to-audio` such that the following lines work with most* text-to-audio models:

```py
from transformers import pipeline

synthesiser = pipeline("text-to-audio", "facebook/musicgen-large")
music = synthesiser("A low-fi song with a strong bassline")
synthesiser.save_audio(music, "test.wav")
```

On most* models where voice control is possible, it is possible to control it through the `voice` pipeline argument. The valid values for `voice` are model-dependent, and this argument is documented accordingly

```py
from transformers import pipeline

tts_pipeline = pipeline("text-to-audio", "sesame/csm-1b")
audio = tts_pipeline("I just got bamboozled by my cat.", voice="1")
tts_pipeline.save_audio(audio, "test.wav")
```

## Core changes

Prior to this PR, recent models with `text-to-audio` capabilities had no pipeline support (e.g. CSM, Dia, Qwen2.5 Omni). There was also not a standardized way to control the voice, if the model generates speech.

With this PR, `TextToAudioPipeline`:
- Uses a `processor` whenever possible, automatically (as opposed to needing a flag to control it);
- Takes a `voice` argument, which a few models can use out of the box. Whether a model can take `voice` in the pipeline is specified by properties of the model (if future models have the same properties, they will also have `voice` support);
- Standardizes outputs: ALL models using the pipeline will return `{"audio": <np.array with shape (audio_channels, sequence_length)>, "sampling_rate": <int>}`.  Different models return different array formats, and as a result we can see different saving scripts in the model cards -- the pipeline standardizes it;
- Adds a function to save the audio, for convenience. This way, users don't need to learn about `soundfile` or alternatives. Uses the processor's `save_audio` whenever it is available.

## Model support

Models with out-of-the-box pipeline support:
  - TTS Models with `voice` support:
    - CSM
    - Qwen2.5 Omni
    - Bark
  - TTS Models w/o `voice` support:
    - Dia  -- voice is set in the prompt; needs chat templates? (we can hardcode it in the pipeline, though 🤔 )
    - FastSpeech2Conformer (model has no voice control)
    - SeamlessM4T and variants (model has no voice control)
    - Vits (model has no voice control)
  - TTA Models:
    - Musicgen and variants

Models that have special requirements:
[...]

PR #42326, opened 2025-11-21
Fix processor usage + add chat_template support to TTS pipeline, and shift common chat template logic to base class.

ebezzam:
# What does this PR do?

Processor usage within TTS pipelines is faulty.

Moreover, it does not support chat template inputs with/without audio, which is of interest for the following models for generating conversations:
- [CSM](https://huggingface.co/sesame/csm-1b) (already in Transformers) 
- [Higgs](https://huggingface.co/bosonai/higgs-audio-v2-generation-3B-base) (upcoming in Transformers, cc @eustlb)
- [VibeVoice](https://huggingface.co/microsoft/VibeVoice-1.5B) (upcoming in Transformers)

Example usage [for CSM](https://huggingface.co/sesame/csm-1b#csm-sounds-best-when-provided-with-context) and inputs like such for VibeVoice:
```python
conversation = [
    {"role": "0", "content": [
        {"type": "text", "text": "Hello everyone, and welcome to the VibeVoice podcast. I'm your host, Linda, and today we're getting into one of the biggest debates in all of sports: who's the greatest basketball player of all time? I'm so excited to have Thomas here to talk about it with me."},
        {"type": "audio", "path": https://hf.co/datasets/bezzam/vibevoice_samples/resolve/main/voices/en-Alice_woman.wav}
    ]},
    {"role": "1", "content": [
        {"type": "text", "text": "Thanks so much for having me, Linda. You're absolutely right—this question always brings out some seriously strong feelings."},
        {"type": "audio", "path": https://hf.co/datasets/bezzam/vibevoice_samples/resolve/main/voices/en-Frank_man.wav}
    ]},
    {"role": "0", "content": [
        {"type": "text", "text": "Okay, so let's get right into it. For me, it has to be Michael Jordan. Six trips to the Finals, six championships. That kind of perfection is just incredible."},
    ]},
    {"role": "1", "content": [
        {"type": "text", "text": "Oh man, the first thing that always pops into my head is that shot against the Cleveland Cavaliers back in '89. Jordan just rises, hangs in the air forever, and just sinks it"},
    ]},
]
```

This PR is related to #39796, but this one is different/simpler (and I'd say more urgent) in its objective 👉 _fixing processor usage and enabling chat_template inputs like above_.

So I think it's worth a separate PR as #39796 requires more testing/review.

# Context for current error

[This line](https://github.com/huggingface/transformers/blob/f2738ee3756483538c20e10e4f104324675fb406/src/transformers/pipelines/text_to_audio.py#L153) fails when trying to use the processor because it [isn’t loaded](https://github.com/huggingface/transformers/blob/f2738ee3756483538c20e10e4f104324675fb406/src/transformers/pipelines/text_to_audio.py#L84) (so it is `None` instead)

Minimal failing example:
```python
import torch

model_id = "sesame/csm-1b"
device = "cuda" if torch.cuda.is_available() else "cpu"
pipe = pipeline("text-to-speech", model=model_id, device=device, no_processor=False)

# prepare the inputs
text = "[0]Hello from Sesame." # `[0]` for speaker id 0

# apply pipeline
output = pipe(text)

"""
  File "/home/eric_bezzam/transformers/src/transformers/pipelines/test_pipeline_chat_template.py", line 34, in <module>
    output = pipe(text)
             ^^^^^^^^^^
  File "/home/eric_bezzam/transformers/src/transformers/pipelines/text_to_audio.py", line 218, in __call__
    return super().__call__(text_inputs, **forward_params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eric_bezzam/transformers/src/transformers/pipelines/base.py", line 1261, in __call__
    return self.run_single(inputs, preprocess_params, forward_params, postprocess_params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eric_bezzam/transformers/src/transformers/pipelines/base.py", line 1267, in run_single
    model_inputs = self.preprocess(inputs, **preprocess_params)
  File "/home/eric_bezzam/transformers/src/transformers/pipelines/text_to_audio.py", line 153, in preprocess
    output = preprocessor(text, **kwargs, return_tensors="pt")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not callable
"""
```

# After changes

No need to explicitly ask for processor usage (auto-detected internally).

TTS example
```python
import torch
import soundfile as sf

model_id = "sesame/csm-1b"
device = "cuda" if torch.cuda.is_available() else "cpu"
pipe = pipeline("text-to-speech", model=model_id, device=device)

# prepare the inputs
text = "[0]Hello from Sesame." # `[0]` for speaker id 0

# apply pipeline
output = pipe(text, generate_kwargs={"output_audio": True})

# save the audio to a file
audio = output["audio"][0].squeeze()
fn = "csm_pipeline_output.wav"
sf.write(fn, audio, output["sampling_rate"])
print(f"Audio saved to {fn}")
```

Conversation example with chat template, which was not possible before! Again auto-detected [like in text-generation pipeline](https://github.com/huggingface/transformers/blob/f2738ee3756483538c20e10e4f104324675fb406/src/transformers/pipelines/text_generation.py#L359). Below example mimics [this usage](https://huggingface.co/sesame/csm-1b#csm-sounds-best-when-provided-with-context) from CSM:
```python
import torch
import soundfile as sf
from datasets import load_dataset, Audio

model_id = "sesame/csm-1b"
device = "cuda" if torch.cuda.is_available() else "cpu"
pipe = pipeline("text-to-speech", model=model_id, device=device)

# prepare the inputs like here: https://huggingface.co/sesame/csm-1b#csm-sounds-best-when-provided-with-context
ds = load_dataset("hf-internal-testing/dailytalk-dummy", split="train")
ds = ds.cast_column("audio", Audio(sampling_rate=24000))
conversation = []
for text, audio, speaker_id in zip(ds[:4]["text"], ds[:4]["audio"], ds[:4]["speaker_id"]):
    conversation.append(
        {
            "role": f"{speaker_id}",
            "content": [{"type": "text", "text": text}, {"type": "audio", "path": audio["array"]}],
        }
    )
conversation.append({"role": f"{ds[4]['speaker_id']}", "content": [{"type": "text", "text": ds[4]["text"]}]})

# apply pipeline
output = pipe(conversation, generate_kwargs={"output_audio": True})

# save the audio to a file
audio = output["audio"][0].squeeze()
fn = "csm_pipeline_output_chat.wav"
sf.write(fn, audio, output["sampling_rate"])
print(f"Audio saved to {fn}")

```

---

@Rocketknight1 since it is pipeline related you can take a look if you want! but will definitely ask @vasqu an @eustlb for audio-specific feedback 🙂


src/transformers/pipelines/text_to_audio.py
```diff
@@ -27,6 +30,23 @@
...
+    to this format because the rest of the pipeline code tends to assume that lists of messages are
+    actually a batch of samples rather than messages in the same conversation."""
+
+    def __init__(self, messages: dict):
+        for message in messages:
+            if not ("role" in message and "content" in message):
+                raise ValueError("When passing chat dicts as input, each dict must have a 'role' and 'content' key.")
+        self.messages = messages
+
+
```
ebezzam: Essentially coming from text generation: https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/text_generation.py#L17

src/transformers/pipelines/text_to_audio.py
```diff
@@ -141,16 +158,26 @@ def preprocess(self, text, **kwargs):
                 "add_special_tokens": False,
                 "return_attention_mask": True,
                 "return_token_type_ids": False,
-                "padding": "max_length",
```
ebezzam, line 144: Something Joao had removed in https://github.com/huggingface/transformers/pull/39796. I think to get tests passing?

src/transformers/pipelines/text_to_audio.py
```diff
@@ -141,16 +158,26 @@ def preprocess(self, text, **kwargs):
...
             new_kwargs.update(kwargs)
-
             kwargs = new_kwargs
 
-        preprocessor = self.tokenizer if self.no_processor else self.processor
-        output = preprocessor(text, **kwargs, return_tensors="pt")
+        preprocessor = self.processor if self.processor is not None else self.tokenizer
+        if self.model.config.model_type in ["musicgen"]:
+            # Fallback to legacy models that prefer tokenizer
+            preprocessor = self.tokenizer
```
ebezzam: Not super happy with this fallback... but needed to make this line happy: https://github.com/huggingface/transformers/blob/f15b95e67c89a9f62df40ed0caff8f08f0ddd496/tests/pipelines/test_pipelines_text_to_audio.py#L231

src/transformers/pipelines/text_to_audio.py
```diff
@@ -215,6 +251,28 @@ def __call__(self, text_inputs: str | list[str], **forward_params) -> dict[str,
...
+            if isinstance(first_item, (list, tuple, dict)):
+                # We have one or more prompts in list-of-dicts format, so this is chat mode
+                if isinstance(first_item, dict):
+                    return super().__call__(Chat(text_inputs), **forward_params)
+                else:
+                    chats = (Chat(chat) for chat in text_inputs)
+                    if isinstance(text_inputs, types.GeneratorType):
+                        return super().__call__(chats, **forward_params)
+                    else:
+                        return super().__call__(list(chats), **forward_params)
```
ebezzam: Similarly from text generation: https://github.com/huggingface/transformers/blob/f15b95e67c89a9f62df40ed0caff8f08f0ddd496/src/transformers/pipelines/text_generation.py#L311

src/transformers/pipelines/text_to_audio.py
```diff
@@ -248,17 +306,12 @@ def postprocess(self, audio):
...
-                waveform = audio[0]
-            else:
-                waveform = audio
-        # Or we need to postprocess to get the waveform
+        if isinstance(audio, dict):
+            waveform = audio[waveform_key]
+        elif isinstance(audio, tuple):
+            waveform = audio[0]
         else:
-            waveform = self.processor.decode(audio)
```
ebezzam, line 261: Was breaking for CSM

tests/pipelines/test_pipelines_text_to_audio.py
```diff
@@ -151,7 +151,6 @@ def test_conversion_additional_tensor(self):
             "add_special_tokens": False,
             "return_attention_mask": True,
             "return_token_type_ids": False,
-            "padding": "max_length",
```
ebezzam, line 154: Something Joao had removed in https://github.com/huggingface/transformers/pull/39796. I think to get tests passing?

ebezzam (commented):
@vasqu Self-review with pointers that hopefully helps!

ebezzam:
Failed tests are unrelated, concatenating below:
```
# https://app.circleci.com/pipelines/github/huggingface/transformers/154274/workflows/48600cd6-5d2a-455a-a2e0-96f3b9bae8ae/jobs/2028466
FAILED tests/models/efficientloftr/test_image_processing_efficientloftr.py::EfficientLoFTRImageProcessingTest::test_post_processing_keypoint_matching_with_padded_match_indices - AssertionError: 2 != 1
===== 1 failed, 549 passed, 362 skipped, 24 warnings in 115.22s (0:01:55) ======

# https://app.circleci.com/pipelines/github/huggingface/transformers/154274/workflows/48600cd6-5d2a-455a-a2e0-96f3b9bae8ae/jobs/2028474
FAILED tests/models/resnet/test_modeling_resnet.py::ResNetModelTest::test_can_load_ignoring_mismatched_shapes - AssertionError: 0.14109472930431366 not less than or equal to 0.1 : Issue with classifier.1.bias
==== 1 failed, 3959 passed, 6611 skipped, 93 warnings in 187.89s (0:03:07) =====

# https://app.circleci.com/pipelines/github/huggingface/transformers/154274/workflows/48600cd6-5d2a-455a-a2e0-96f3b9bae8ae/jobs/2028467
FAILED tests/models/olmo/test_modeling_olmo.py::OlmoModelTest::test_generate_with_static_cache - AssertionError: False is not true
===== 1 failed, 625 passed, 218 skipped, 16 warnings in 118.43s (0:01:58) ======
```

src/transformers/pipelines/text_to_audio.py
```diff
@@ -27,6 +30,22 @@
...
+class Chat:
+    """This class is intended to just be used internally in this pipeline and not exposed to users. We convert chats
+    to this format because the rest of the pipeline code tends to assume that lists of messages are
+    actually a batch of samples rather than messages in the same conversation."""
+
+    def __init__(self, messages: dict):
+        for message in messages:
+            if not ("role" in message and "content" in message):
+                raise ValueError("When passing chat dicts as input, each dict must have a 'role' and 'content' key.")
+        self.messages = messages
```
vasqu: I feel like chat templates with any modality are increasingly more important, might be better to move this somewhere more general (and let it be imported). We gonna have audio, image, text already atp

vasqu: And I would like to avoid copied from tbh

src/transformers/pipelines/text_to_audio.py
```diff
@@ -81,7 +100,7 @@ class TextToAudioPipeline(Pipeline):
     """
 
     _pipeline_calls_generate = True
-    _load_processor = False
+    _load_processor = None
```
vasqu: Let's add a comment that we prioritize processors and that some classes require it. `None` is usually optional but in our case the processor is essential for newer models

src/transformers/pipelines/text_to_audio.py
vasqu: Can you double check? It's only bark so should be quick to check

src/transformers/pipelines/text_to_audio.py
vasqu: Would it make more sense to define this directly on init --> set processor directly to `None`

src/transformers/pipelines/text_to_audio.py
```diff
@@ -141,16 +157,26 @@ def preprocess(self, text, **kwargs):
...
+
+        if isinstance(text, Chat):
+            output = preprocessor.apply_chat_template(
+                text.messages,
+                tokenize=True,
+                return_dict=True,
+                **kwargs,
+            )
+        else:
+            output = preprocessor(text, **kwargs, return_tensors="pt")
```
vasqu, line 175: Nice!

src/transformers/pipelines/text_to_audio.py
```diff
@@ -193,13 +219,22 @@ def __call__(self, text_inputs: str, **forward_params: Any) -> dict[str, Any]: .
...
         """
         Generates speech/audio from the inputs. See the [`TextToAudioPipeline`] documentation for more information.
 
         Args:
             text_inputs (`str` or `list[str]`):
-                The text(s) to generate.
+                One or several texts to generate. If strings or a list of string are passed, this pipeline will
+                generate the corresponding text. Alternatively, a "chat", in the form of a list of dicts with "role"
+                and "content" keys, can be passed, or a list of such chats. When chats are passed, the model's chat
+                template will be used to format them before passing them to the model.
```
vasqu, line 236: Update type above too then?

src/transformers/pipelines/text_to_audio.py
```diff
@@ -215,6 +250,26 @@ def __call__(self, text_inputs: str | list[str], **forward_params) -> dict[str,
             - **audio** (`np.ndarray` of shape `(nb_channels, audio_length)`) -- The generated audio waveform.
             - **sampling_rate** (`int`) -- The sampling rate of the generated audio waveform.
         """
+        if isinstance(
+            text_inputs,
+            (list, tuple, types.GeneratorType) if is_torch_available() else (list, tuple, types.GeneratorType),
+        ):
```
vasqu:
```suggestion
        if isinstance(text_inputs, (list, tuple, types.GeneratorType)):
```
Am I missing something?

src/transformers/pipelines/text_to_audio.py
```diff
@@ -215,6 +250,26 @@ def __call__(self, text_inputs: str | list[str], **forward_params) -> dict[str,
...
+            if isinstance(first_item, (list, tuple, dict)):
+                # We have one or more prompts in list-of-dicts format, so this is chat mode
+                if isinstance(first_item, dict):
+                    return super().__call__(Chat(text_inputs), **forward_params)
+                else:
+                    chats = (Chat(chat) for chat in text_inputs)
+                    if isinstance(text_inputs, types.GeneratorType):
+                        return super().__call__(chats, **forward_params)
+                    else:
+                        return super().__call__(list(chats), **forward_params)
```
vasqu: I honestly like this, it's short and sweet but I feel like we should maybe align with image e.g. https://github.com/huggingface/transformers/blob/ce7a5e0066e8b789e7ad7a8b6fdc12993cbe0f72/src/transformers/pipelines/image_text_to_text.py#L331-L352
We cooking all our own soup :cry: 

Only thing I'd change would be avoid using a wildcard `_` and give an explicit name instead

src/transformers/pipelines/text_to_audio.py
vasqu: Iirc it was for Dia but it has been broken way too often.

tests/pipelines/test_pipelines_text_to_audio.py
vasqu: Same as in bark, doesnt hurt to double check.

vasqu (commented):
Left some comments :hugs: I think it's overall fine more so nits to have more alignment with other pipelines.

My only gripe is that we still don't have a standard way how models generate audio. For example, CSM directly generates the audio waveform but that's because it uses the audio tokenizer directly within the model itself. Dia does not do that and depends on the processor do decode into waveform. This is something we have to properly enfore at some point before we get too many exceptions cc @eustlb

vasqu:
Failing tests are either flaky (loftr one is handled elsewhere)

src/transformers/pipelines/text_to_audio.py
ebezzam: Currently two other pipelines using this:
- [text generation](https://github.com/huggingface/transformers/blob/9162e190da31c2ee9aec81f052dbeb5fced72b24/src/transformers/pipelines/text_generation.py#L26)
- [image-text-to-text](https://github.com/huggingface/transformers/blob/9162e190da31c2ee9aec81f052dbeb5fced72b24/src/transformers/pipelines/image_text_to_text.py#L53), which calls additional [`add_images_to_messages`](https://github.com/huggingface/transformers/blob/9162e190da31c2ee9aec81f052dbeb5fced72b24/src/transformers/pipelines/image_text_to_text.py#L64)

We could combine into a single `Chat` object in [base.py](https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/base.py) like so?

```python
class Chat:
    """This class is intended to just be used internally in this pipeline and not exposed to users. We convert chats
    to this format because the rest of the pipeline code tends to assume that lists of messages are
    actually a batch of samples rather than messages in the same conversation."""

    def __init__(
        self, messages: dict, images: Union[str, list[str], "Image.Image", list["Image.Image"]] | None = None
    ):
        for message in messages:
            if not ("role" in message and "content" in message):
                raise ValueError("When passing chat dicts as input, each dict must have a 'role' and 'content' key.")
        if images is not None:
             messages = add_images_to_messages(messages, images)

        self.messages = messages
```

For audio models with @eustlb, our chat templates already allow audio in the message and the jinja template (like [this](https://huggingface.co/sesame/csm-1b/blob/main/chat_template.jinja) and [this](https://huggingface.co/bezzam/VibeVoice-1.5B/blob/main/chat_template.jinja)) handles extracting audio when calling `apply_chat_template` (which then calls the processor internally, see [here](https://github.com/huggingface/transformers/blob/9162e190da31c2ee9aec81f052dbeb5fced72b24/src/transformers/processing_utils.py#L1715))

src/transformers/pipelines/text_to_audio.py
ebezzam: It was failing when it wasn't removed! 

My question mark was rather a guess as to why Joao removed it 😛

tests/pipelines/test_pipelines_text_to_audio.py
ebezzam: same as above! (double-checked)

src/transformers/pipelines/text_to_audio.py
ebezzam: ha yes 🙈  when pruning from text generation missed that further simplification!

src/transformers/pipelines/text_to_audio.py
vasqu: Oh lol, then nvm this. Should be fine and bark is not super high priority.

src/transformers/pipelines/text_to_audio.py
vasqu: SGTM, @Rocketknight1 @zucchini-nlp wdyt about this? Not sure who has a better overview on (image) pipelines.

Rocketknight1: Yes, you can move it to `base.py` or `utils/chat_template_utils.py`, whichever is a cleaner import. Not a huge issue either way, though!

ebezzam: thanks @Rocketknight1! I've put in `utils/chat_template_utils.py` for now

src/transformers/pipelines/text_to_audio.py
ebezzam: As discussed on Slack, the `__call__` logic of `image-text-to-text` may be more complicated because they allow users to pass images as separate arguments rather than keeping everything within the chat template?

src/transformers/pipelines/text_to_audio.py
ebezzam: For `image-text-to-text`'s current Chat object ([here](https://github.com/huggingface/transformers/blob/3410ba9bab09ca6dadc130ab29fabfaba8baa131/src/transformers/pipelines/image_text_to_text.py#L53)), it could be nice to do something like below in the general purpose `Chat` to support the `images` input:

```python
class Chat:
    """This class is intended to just be used internally for pipelines and not exposed to users. We convert chats
    to this format because the rest of the pipeline code tends to assume that lists of messages are
    actually a batch of samples rather than messages in the same conversation."""

    def __init__(self, messages: dict, images: Union[str, list[str], "Image.Image", list["Image.Image"]] | None = None):
        for message in messages:
            if not ("role" in message and "content" in message):
                raise ValueError("When passing chat dicts as input, each dict must have a 'role' and 'content' key.")
        
        if images is not None:
	        messages = add_images_to_messages(messages, images)
        self.messages = messages
```

But actually this fails for a current usage 👉  when there is an image URL in the chat template ([this code path](https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/image_text_to_text.py#L94)).

FYI, I found out above this edge case, because [this test](https://github.com/huggingface/transformers/blob/3410ba9bab09ca6dadc130ab29fabfaba8baa131/tests/pipelines/test_pipelines_image_text_to_text.py#L349) would fail when I tried a modified Chat object like above (because the image wouldn't be properly loaded).

@zucchini-nlp do you have an idea of how `image-text-to-text` could also use the general purpose `Chat` object without having to call `add_images_to_messages` to avoid the current [double for-loop](https://github.com/huggingface/transformers/blob/3410ba9bab09ca6dadc130ab29fabfaba8baa131/src/transformers/pipelines/image_text_to_text.py#L78) if there is indeed no image input?

ebezzam: @vasqu, @Rocketknight1 for reference @zucchini-nlp started [a PR](https://github.com/huggingface/transformers/pull/42359) to handle `image-text-to-text` so I won't touch it in this PR

src/transformers/models/seamless_m4t/modeling_seamless_m4t.py
```diff
@@ -1981,7 +1981,6 @@ class SeamlessM4TTextToUnitForConditionalGeneration(SeamlessM4TPreTrainedModel,
         "text_encoder",
         "text_decoder",
     ]
-    _tied_weights_keys = {"lm_head.weight": "model.decoder.embed_tokens.weight"}
```
ebezzam, line 1984: @vasqu tbh I'm not familiar with `tied_weights` aspect, but following test was failing without removing this line
```
RUN_SLOW=1 pytest tests/pipelines/test_pipelines_text_to_audio.py::TextToAudioPipelineTests::test_medium_seamless_m4t_pt

E               During handling of the above exception, another exception occurred:
E               
E               Traceback (most recent call last):
E                 File "/home/eric_bezzam/transformers/src/transformers/pipelines/base.py", line 246, in load_model
E                   model = model_class.from_pretrained(model, **fp32_kwargs)
E                 File "/home/eric_bezzam/transformers/src/transformers/modeling_utils.py", line 278, in _wrapper
E                   return func(*args, **kwargs)
E                 File "/home/eric_bezzam/transformers/src/transformers/modeling_utils.py", line 3974, in from_pretrained
E                   model, missing_keys, unexpected_keys, mismatched_keys, offload_index, error_msgs = cls._load_pretrained_model(
E                 File "/home/eric_bezzam/transformers/src/transformers/modeling_utils.py", line 4142, in _load_pretrained_model
E                   model.tie_weights(missing_keys=missing_keys, recompute_mapping=False)
E                 File "/home/eric_bezzam/transformers/src/transformers/modeling_utils.py", line 2332, in tie_weights
E                   raise ValueError(
E               ValueError: This checkpoint seem corrupted. The tied weights mapping for this model specifies to tie t2u_model.model.decoder.embed_tokens.weight (which should be present and is not), to t2u_model.lm_head.weight (which is present).
```
I think because this tying is not necessary as it is done [elsewhere](https://github.com/huggingface/transformers/blob/dc6a53b9c152e5f02f37955fb8b09170bf6f6caa/src/transformers/models/seamless_m4t/modeling_seamless_m4t.py#L3586)?

Note that this line was changed recently with this hefty PR: https://github.com/huggingface/transformers/pull/41580

vasqu: Yup, refactored how we tied weights. Is it relevant to this PR explicitly? I would rather have this in a different PR

FMI, are the integration tests still pass?

src/transformers/pipelines/text_to_audio.py
```diff
@@ -11,13 +11,20 @@
...
+import types
 from typing import Any, overload
 
 from ..generation import GenerationConfig
 from ..utils import is_torch_available
+from ..utils.chat_template_utils import Chat, ChatType
 from .base import Pipeline
 
 
+AudioOutput = dict[str, Any]  # {"audio": np.ndarray, "sampling_rate": int}
```
vasqu: I'd rather have a typed dict with docs, you can take inspiration from processing utils

src/transformers/pipelines/text_to_audio.py
```diff
@@ -215,6 +238,23 @@ def __call__(self, text_inputs: str | list[str], **forward_params) -> dict[str,
             - **audio** (`np.ndarray` of shape `(nb_channels, audio_length)`) -- The generated audio waveform.
             - **sampling_rate** (`int`) -- The sampling rate of the generated audio waveform.
         """
+        if isinstance(text_inputs, (list, tuple, types.GeneratorType)):
+            if isinstance(text_inputs, types.GeneratorType):
+                text_inputs, _ = itertools.tee(text_inputs)
+                text_inputs, first_item = (x for x in text_inputs), next(_)
```
vasqu: Let's exchange the wildcard tho, i.e. `_` - trying to be explicit here

src/transformers/pipelines/text_to_audio.py
vasqu: I thought we were breaking it tho for v5? Or did I misunderstand something?

src/transformers/pipelines/text_to_audio.py
vasqu: Reopening because I want to verify that Dia works with this current version - I'm pretty sure we need the processor to decode for Dia which is why I wrote the initial long message on how we plan to standardize
- Everything handled by the model, audio tokenizer within it already
- Separate model / tokenizer, processor handles encoding/decoding into codebooks/waveform

vasqu (commented):
Added some smaller comments, imo we should still check how we standardize (not necessarily in this PR but we should keep this in mind / discuss before we lose all sight)

I think 2 somewhat bigger things
- seamless should be a different PR?
- check if dia works

src/transformers/models/seamless_m4t/modeling_seamless_m4t.py
ebezzam: I agree separate PR is better! I "threw" this in here as I was getting a failing error from this model when running the `test_pipelines_text_to_audio.py` tests.

Below are the Seamless M4T tests, namely running `RUN_SLOW=1 pytest tests/models/seamless_m4t/test_modeling_seamless_m4t.py`

Before above fix (11 errors)
```
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelWithSpeechInputTest::test_model_from_pretrained - ValueError: This checkpoint seem corrupted. The tied weights mapping for this model specifies to tie t2u_model.model.decoder.embed_tokens.weight (which should be present and is not), to t2u_mo...
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelWithTextInputTest::test_model_from_pretrained - ValueError: This checkpoint seem corrupted. The tied weights mapping for this model specifies to tie t2u_model.model.decoder.embed_tokens.weight (which should be present and is not), to t2u_mo...
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelWithTextInputTest::test_resize_embeddings_untied_with_deepspeed - ModuleNotFoundError: No module named 'mpi4py'
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelWithTextInputTest::test_resize_tokens_embeddings_with_deepspeed - ModuleNotFoundError: No module named 'mpi4py'
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_speech_to_speech_model - ValueError: This checkpoint seem corrupted. The tied weights mapping for this model specifies to tie t2u_model.model.decoder.embed_tokens.weight (which should be present and is not), to t2u_mo...
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_speech_to_text_model - ValueError: This checkpoint seem corrupted. The tied weights mapping for this model specifies to tie t2u_model.model.decoder.embed_tokens.weight (which should be present and is not), to t2u_mo...
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_text_to_speech_model - ValueError: This checkpoint seem corrupted. The tied weights mapping for this model specifies to tie t2u_model.model.decoder.embed_tokens.weight (which should be present and is not), to t2u_mo...
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_text_to_text_model - ValueError: This checkpoint seem corrupted. The tied weights mapping for this model specifies to tie t2u_model.model.decoder.embed_tokens.weight (which should be present and is not), to t2u_mo...
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_to_eng_text - ValueError: This checkpoint seem corrupted. The tied weights mapping for this model specifies to tie t2u_model.model.decoder.embed_tokens.weight (which should be present and is not), to t2u_mo...
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_to_rus_speech - ValueError: This checkpoint seem corrupted. The tied weights mapping for this model specifies to tie t2u_model.model.decoder.embed_tokens.weight (which should be present and is not), to t2u_mo...
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_to_swh_text - ValueError: This checkpoint seem corrupted. The tied weights mapping for this model specifies to tie t2u_model.model.decoder.embed_tokens.weight (which should be present and is not), to t2u_mo...
========== 11 failed, 79 passed, 193 skipped, 5 warnings in 116.67s (0:01:56) ====
```

After above fix (7 errors)
```
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelWithTextInputTest::test_resize_embeddings_untied_with_deepspeed - ModuleNotFoundError: No module named 'mpi4py'
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelWithTextInputTest::test_resize_tokens_embeddings_with_deepspeed - ModuleNotFoundError: No module named 'mpi4py'
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_speech_to_speech_model - AssertionError: 86400 != 86080
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_text_to_speech_model - AssertionError: 90560 != 86720
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_to_eng_text - AssertionError: Lists differ: [2, 10051, 8980, 8212, 949, 1270, 4311, 1123, 5918, 2333] != [2, 10051, 9253, 9253, 9253, 9253, 3765, 3765, 3765, 3765]
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_to_rus_speech - AssertionError: Lists differ: [2, 10067, 5729, 4798, 9631, 8378, 4446, 2393, 6901, 5983] != [2, 10067, 5560, 5560, 5560, 5560, 5560, 5560, 5560, 5560]
FAILED tests/models/seamless_m4t/test_modeling_seamless_m4t.py::SeamlessM4TModelIntegrationTest::test_to_swh_text - AssertionError: Lists differ: [2, 10071, 5729, 9995, 3089, 7546, 1204, 1721, 2532, 4340] != [2, 10071, 2679, 2679, 2679, 2679, 2679, 2679, 9178, 9178]
============ 7 failed, 83 passed, 193 skipped, 5 warnings in 161.17s (0:02:41) =========
```

ebezzam: Created this PR where can shift conversation: https://github.com/huggingface/transformers/pull/42377

src/transformers/pipelines/text_to_audio.py
ebezzam: love it, thanks for the tip

src/transformers/pipelines/text_to_audio.py
```diff
@@ -27,6 +32,17 @@
 DEFAULT_VOCODER_ID = "microsoft/speecht5_hifigan"
 
 
+class AudioOutput(TypedDict, total=False):
```
ebezzam, line 33: @vasqu new `AudioOutput` object

This could also be interesting in `audio_utils.py` at one point. Maybe a "richer" audio object, as I've seen in other public/internal libraries, with some useful utilities (as mentioned [here](https://github.com/huggingface/transformers/pull/39796#discussion_r2244616591)). But that's a rabbit hole for another PR 😁

src/transformers/pipelines/text_to_audio.py
ebezzam: Good point, Dia does need the processor for decoding and I'll also add a unit test for Dia so we don't miss this in the future.

However, I feel like a blanket `self.processor.decode` might be too broad. For example, CSM and VibeVoice don't require the processor to decode.

Since there is not standard approach (yet), how about something like below (which is working):
```python
  if isinstance(audio, dict):
      waveform = audio[waveform_key]
  elif isinstance(audio, tuple):
      waveform = audio[0]
  elif self.model.config.model_type in ["dia"]:
      # models that require decoding, e.g. with codec
      waveform = self.processor.decode(audio)
  else:
      waveform = audio
```

ebezzam: Example usage:
```python
import soundfile as sf


dia_pipeline = pipeline(
    "text-to-audio", model=model_checkpoint,
)
outputs = dia_pipeline(
    "[S1] Dia is an open weights text to dialogue model.",
    generate_kwargs={"max_new_tokens": 512},
)
assert outputs["sampling_rate"] == 44100

audio = outputs["audio"].squeeze()
fn = "dia_pipeline_output.wav"
sf.write(fn, audio, outputs["sampling_rate"])
print(f"Audio saved to {fn}")
```

I'm reluctant to allow [voice cloning](https://huggingface.co/nari-labs/Dia-1.6B-0626#generation-with-text-and-audio-voice-cloning) through pipeline, as this would require passing an `audios` input to pipeline (since Dia doesn't support chat templates).

Moreover, allowing inputs like `audios` is exactly what they are trying to phase out with `image-text-to-text` in https://github.com/huggingface/transformers/pull/42359 (to only support chat template usage).

src/transformers/pipelines/text_to_audio.py
ebezzam: What they are planning to break for v5 in `image-text-to-text` is removing inputs like `images` for pipeline, so that users stick to chat template. See https://github.com/huggingface/transformers/pull/42359

But perhaps the `__call__` logic could still be simplified and shifted to `base.py`. Let me see...

src/transformers/pipelines/base.py
```diff
@@ -1202,6 +1205,29 @@ def __call__(self, inputs, *args, num_workers=None, batch_size=None, **kwargs):
...
+                first_item = inputs[0]
+            if isinstance(first_item, (list, tuple, dict)):
+                if isinstance(first_item, dict):
+                    inputs = Chat(inputs)
+                else:
+                    chats = (Chat(chat) for chat in inputs)
+                    if isinstance(inputs, types.GeneratorType):
+                        inputs = chats
+                    else:
+                        inputs = list(chats)
```
ebezzam: @vasqu an idea for standardizing chat template usage in the base class.

Essentially this was in `text-generation` and what I had copied in `text-to-audio` (in what you last saw), and could potentially be used by `image-text-to-text` once it drops support for the `images` input (@zucchini-nlp)?

Following test run as before:
```
RUN_SLOW=1 pytest tests/pipelines/test_pipelines_text_to_audio.py
RUN_SLOW=1 pytest tests/pipelines/test_pipelines_text_generation.py
RUN_SLOW=1 pytest tests/pipelines/test_pipelines_image_text_to_text.py
```

I could also run below if you think we should check to be safe? and do you have anything else in mind that I should double check?
```
RUN_SLOW=1 pytest tests/pipelines
```

src/transformers/pipelines/text_to_audio.py
```diff
@@ -241,29 +271,32 @@ def _sanitize_parameters(
...
-            elif isinstance(audio, tuple):
-                waveform = audio[0]
-            else:
-                waveform = audio
-        # Or we need to postprocess to get the waveform
+        if self.model.config.model_type in ["csm"]:
+            audio_key = "audio"
+        elif self.model.config.model_type in ["dia"]:
+            # codes that need decoding
+            audio_key = "sequences"
```
ebezzam: I added a Dia test, and this is needed when using the `batch_size` argument to `pipeline`

tests/pipelines/test_pipelines_text_to_audio.py
```diff
@@ -247,22 +246,64 @@ def test_generative_model_kwargs(self):
...
+        max_new_tokens = 10
 
-        outputs = speech_generator("[0]This is a test")
+        outputs = speech_generator("[0]This is a test", generate_kwargs={"output_audio": True, "max_new_tokens": max_new_tokens})
         self.assertEqual(outputs["sampling_rate"], 24000)
 
         audio = outputs["audio"]
         self.assertEqual(ANY(np.ndarray), audio)
+        # ensure audio and not codes
+        self.assertEqual(len(audio.shape), 1)
```
ebezzam: Added such checks to make sure audio decoding actually working

tests/pipelines/test_pipelines_text_to_audio.py
```diff
@@ -247,22 +246,64 @@ def test_generative_model_kwargs(self):
...
-        outputs = speech_generator(["[0]This is a test", "[0]This is a second test"], batch_size=2)
+        generate_kwargs = {"output_audio": True, "max_new_tokens": max_new_tokens, "return_dict_in_generate": True}
+        outputs = speech_generator(["[0]This is a test", "[0]This is a second test"], generate_kwargs=generate_kwargs, batch_size=2)
         self.assertEqual(ANY(np.ndarray), outputs[0]["audio"])
+        # ensure audio and not codes
+        self.assertEqual(len(outputs[0]["audio"].squeeze().shape), 1)
+
+    @slow
+    @require_torch
+    def test_dia_model(self):
```
ebezzam, line 282: added Dia test similar to that of CSM

ebezzam (commented):
@vasqu thanks for your comments!

- I tried going directly for an approach that puts common chat template logic into the base pipeline object. Let me know if we should rather focus just on `text-to-audio` and do such standardization in a separate PR. 
- Double-checked Dia. You're right `self.processor.decode` was needed for it, but I've adapted the (previous) logic so it doesn't assume such a call is needed for all models that have a processor (e.g. CSM and VibeVoice don't need to call this)

src/transformers/pipelines/base.py
vasqu: I think our pipelines tests should be all not slow tests, but doesn't hurt to check with the env. I'm pro this!

Waiting for @zucchini-nlp if she has anything to add, if I see it correctly it's depending on #42359

src/transformers/pipelines/text_to_audio.py
vasqu: Yea I think it's good enough for now. Essentially only pipelines have this atm but we could extend for sure. 

A rabbit hole of a rabbit hole :joy:

src/transformers/pipelines/text_to_audio.py
vasqu: That's a good point on voice cloning, we should maybe update Dia with a chat template in the future. I did not have that in mind at that point, that's on me.

Re: standards. Yea, we have no choice atm - it's more of a question on how we handle this in the future

src/transformers/pipelines/text_to_audio.py
vasqu: We have a lot of exceptions :cry: should be refactored at some point, not sure how I feel about this.

tests/pipelines/test_pipelines_text_to_audio.py
```diff
@@ -247,22 +246,64 @@ def test_generative_model_kwargs(self):
...
-        outputs = speech_generator(["[0]This is a test", "[0]This is a second test"])
+        outputs = speech_generator(["[0]This is a test", "[0]This is a second test"], generate_kwargs={"output_audio": True, "max_new_tokens": max_new_tokens})
         audio = [output["audio"] for output in outputs]
         self.assertEqual([ANY(np.ndarray), ANY(np.ndarray)], audio)
+        # ensure audio and not codes
+        self.assertEqual(len(audio[0].shape), 1)
 
         # test batching
-        outputs = speech_generator(["[0]This is a test", "[0]This is a second test"], batch_size=2)
+        generate_kwargs = {"output_audio": True, "max_new_tokens": max_new_tokens, "return_dict_in_generate": True}
```
vasqu: We are essentially using this a lot of times, above without `return_in_...`. Could we only create them once and reuse?

tests/pipelines/test_pipelines_text_to_audio.py
```diff
@@ -247,22 +246,64 @@ def test_generative_model_kwargs(self):
...
+        outputs = speech_generator(["[0]This is a test", "[0]This is a second test"], generate_kwargs=generate_kwargs, batch_size=2)
         self.assertEqual(ANY(np.ndarray), outputs[0]["audio"])
+        # ensure audio and not codes
+        self.assertEqual(len(outputs[0]["audio"].squeeze().shape), 1)
```
vasqu: Let's check for bsz 2 if possible, the 1 is a bit confusing

tests/pipelines/test_pipelines_text_to_audio.py
vasqu: Nice, same comment as in csm for simplification

vasqu (commented):
I think it's already looking pretty good. We should sync with the other pipelines PR not to have some weird conflicting behavior

Other than that, we really should work on enforcing good standards so that we do not have to add so many exceptions (especially now with CSM/Dia). With v5, we IMO have the opportunity to break things and make it unified cc @eustlb wdyt?

src/transformers/pipelines/base.py
```diff
@@ -1202,6 +1225,32 @@ def __call__(self, inputs, *args, num_workers=None, batch_size=None, **kwargs):
...
+                    inputs = Chat(inputs)
+            elif isinstance(first_item, (list, tuple)):
+                # materialize generator is needed
+                items = list(inputs) if isinstance(inputs, types.GeneratorType) else inputs
+                if all(is_valid_chat(chat) for chat in items):
+                    chats = (Chat(chat) for chat in items)
+                    if isinstance(inputs, types.GeneratorType):
+                        inputs = chats
+                    else:
+                        inputs = list(chats)
```
ebezzam: Changed previous logic for backward compatibility: some pipelines pass a list of objects which are not necessarily a chat template (e.g. [key point matching](https://github.com/huggingface/transformers/blob/cb739f8610959c25faa798a0b6030ea930dcfe00/tests/pipelines/test_pipelines_keypoint_matching.py#L162)). **_So I only convert if the object is a valid chat._**

cc @Rocketknight1 who has more experience with pipelines and may spot an edge case 🙃

src/transformers/pipelines/text_to_audio.py
```diff
@@ -171,6 +192,10 @@ def _forward(self, model_inputs, **kwargs):
             # generate_kwargs get priority over forward_params
             forward_params.update(generate_kwargs)
 
+            # ensure dict output to facilitate postprocessing
+            if self.model.config.model_type not in ["bark", "musicgen"]:
+                forward_params.update({"return_dict_in_generate": True})
```
ebezzam: added this as noted in the comment: having the dictionary makes it easier to detect whether calling `self.processor.decode(audio)` is needed. 

So fewer exceptions below, and only adding exceptions here for backwards compatibility 🙂

src/transformers/pipelines/text_to_audio.py
```diff
@@ -241,29 +276,28 @@ def _sanitize_parameters(
...
             else:
-                waveform = audio
-        # Or we need to postprocess to get the waveform
-        else:
-            waveform = self.processor.decode(audio)
-
-        if isinstance(audio, list):
-            output_dict["audio"] = [el.to(device="cpu", dtype=torch.float).numpy() for el in waveform]
+                needs_decoding = True
+                audio = audio["sequences"]
```
ebezzam, line 284: @vasqu no more exceptions in postprocessing with such logic and by ensuring a dict when generate is used within `_forward`

tests/pipelines/test_pipelines_text_to_audio.py
```diff
@@ -247,22 +246,61 @@ def test_generative_model_kwargs(self):
     @slow
     @require_torch
     def test_csm_model_pt(self):
-        speech_generator = pipeline(task="text-to-audio", model="sesame/csm-1b")
+        speech_generator = pipeline(task="text-to-audio", model="sesame/csm-1b", device=torch_device)
+        generate_kwargs = {"max_new_tokens": 10, "output_audio": True}
```
ebezzam, line 254: @vasqu not sure if this is what you had in mind for simplification? Now `generate_kwargs` is only defined once.

But also with `return_dict_in_generate` set to `True` internally, it doesn't need to be explicitly set by user when using the `batch_size` argument.

ebezzam (commented):
@vasqu thanks for previous comments, it's getting better! 

No more exceptions (for newer audio models) in this iteration, so going forward we don't need to add exceptions for newer audio models if:
- they properly use `return_dict_in_generate`
- write `audio` into that the generation output dict, and otherwise codecs that need decoding into `sequences`

src/transformers/pipelines/base.py
ebezzam: FYI no new failures when I run `RUN_SLOW=1 pytest tests/pipelines`

src/transformers/utils/chat_template_utils.py
```diff
@@ -558,3 +561,26 @@ def render_jinja_template(
...
+
+def is_valid_message(message):
+    """
+    Check that input is a valid message in a chat, namely a dict with "role" and "content" keys.
+    """
+    if not isinstance(message, dict):
+        return False
+    if not ("role" in message and "content" in message):
+        return False
+    return True
```
ebezzam, line 574: also in `Chat` and in `pipelines/base.py` to check if the very first dict is part of a chat template. Otherwise pass input as is (for backward compatibility)

src/transformers/pipelines/base.py
Rocketknight1: I don't see anything it's missing! You could maybe simplify it by just always calling `list(inputs)` and removing the extra conditionals for generators, though? Python lists only store pointers to elements so it should be basically free in terms of speed/memory, even if the chats are big.

src/transformers/pipelines/text_to_audio.py
vasqu: Gotcha, fmi can we just do this for all models so that we don't have to explicitly check for bark/musicgen?

src/transformers/pipelines/text_to_audio.py
vasqu: Looks much cleaner thx!

src/transformers/pipelines/text_to_audio.py
```diff
@@ -241,29 +276,27 @@ def _sanitize_parameters(
...
+        elif isinstance(audio, tuple):
+            audio = audio[0]
+
+        if needs_decoding:
+            audio = self.processor.decode(audio)
+
+        if isinstance(audio, list) and len(audio) > 1:
+            audio = [el.to(device="cpu", dtype=torch.float).numpy() for el in audio]
+        elif isinstance(audio, list):
+            audio = audio[0].to(device="cpu", dtype=torch.float).numpy()
```
vasqu: Any reason we explicitly check for the len here? Can't we just do the list comprehension either way?

src/transformers/utils/chat_template_utils.py
vasqu: Sounds good

tests/pipelines/test_pipelines_text_to_audio.py
vasqu: Yea, that's essentially what I had in mind. Now we only define it once at top. Perfect

tests/pipelines/test_pipelines_text_to_audio.py
```diff
@@ -247,22 +246,69 @@ def test_generative_model_kwargs(self):
...
+        n_channel = 1  # model generates mono audio
```
vasqu: My last really big nit is to avoid shortening the var name here

vasqu (approved):
LGTM overall, just make to sure to sync with main and that nothing is broken based on that

The remaining comments are nits

src/transformers/pipelines/text_to_audio.py
ebezzam: bark/musicgen unit test would fail unfortunately :/

Either it would be adding an explicit check here or in `postprocess` for backward compatibility, if I remember correctly. But I'll check again before merging

src/transformers/pipelines/text_to_audio.py
ebezzam: The reasoning is to standardize the output. Because CSM was returning a list even for a single audio, and to be able to do extract the audio with `outputs["audio"]` as in the other models, rather than having to index it like such `outputs["audio"][0]`


But I can change to something like this so it's more concise:
```python
if isinstance(audio, list):
    audio = [el.to(device="cpu", dtype=torch.float).numpy() for el in audio]
    audio = audio if len(audio) > 1 else audio[0]
```

vasqu: Yea, would be better imo

src/transformers/pipelines/base.py
ebezzam: @Rocketknight1 yes I can cast as list from the start! fyi I'll have to remove [this check](https://github.com/huggingface/transformers/blob/f77950605e2f2f27613b829559b133c2b9fd8912/tests/pipelines/test_pipelines_text_generation.py#L256) which errors

src/transformers/models/bark/modeling_bark.py
```diff
@@ -651,8 +651,10 @@ def generate(
         )  # size: 10048
 
         # take the generated semantic tokens
-        semantic_output = semantic_output[:, max_input_semantic_length + 1 :]
-
+        if kwargs.get("return_dict_in_generate", False):
+            semantic_output = semantic_output.sequences[:, max_input_semantic_length + 1 :]
+        else:
+            semantic_output = semantic_output[:, max_input_semantic_length + 1 :]
```
ebezzam, line 657: This and below so that that Bark doesn't error out when `return_dict_in_generate=True` is passed

src/transformers/pipelines/text_to_audio.py
```diff
@@ -171,6 +193,9 @@ def _forward(self, model_inputs, **kwargs):
             # generate_kwargs get priority over forward_params
             forward_params.update(generate_kwargs)
 
+            # ensure dict output to facilitate postprocessing
+            forward_params.update({"return_dict_in_generate": True})
```
ebezzam, line 197: No more exception 🙂

src/transformers/pipelines/text_to_audio.py
```diff
@@ -241,29 +275,26 @@ def _sanitize_parameters(
...
-        # Or we need to postprocess to get the waveform
-        else:
-            waveform = self.processor.decode(audio)
+                needs_decoding = True
+                audio = audio["sequences"]
+        elif isinstance(audio, tuple):
+            audio = audio[0]
+
+        if needs_decoding and self.processor is not None:
+            audio = self.processor.decode(audio)
```
ebezzam, line 289: Just had to add `self.processor is not None` for music gen to pass. Makes sense since I set the processor to None 🤦

tests/pipelines/test_pipelines_text_to_audio.py
```diff
@@ -45,9 +45,11 @@ def test_small_musicgen_pt(self):
         music_generator = pipeline(
             task="text-to-audio", model="facebook/musicgen-small", do_sample=False, max_new_tokens=5
         )
+        n_ch = 1  # model generates mono audio
 
         outputs = music_generator("This is a test")
         self.assertEqual({"audio": ANY(np.ndarray), "sampling_rate": 32000}, outputs)
+        self.assertEqual(len(outputs["audio"].shape), n_ch)
```
ebezzam: added check for musicgen and bark to make sure ensure they are audio with cleaned `postprocess` and for future

ebezzam (commented):
@vasqu I removed the exceptions for bark/musicgen in `_forward`. I needed to fix the handling of `return_dict_in_generate` in `modeling_bark.py`

Cyrilvallez (approved):
Thanks!

src/transformers/pipelines/base.py
Rocketknight1: Yep, makes sense to me! Once we're materializing the entire generator output there's not much point in pretending we're streaming anymore.

PR #42327, opened 2025-11-21
[`CI`] Skip `EfficientLoFTR` test

vasqu:
As per title, added a fixme @yonigozlan 

This is hotfix (as it blocks CI) but tbh we should not inherit from other classes, i.e. SuperGlue. Not sure how many of these cases are still hidden.

### DIFF
--- a/tests/models/efficientloftr/test_image_processing_efficientloftr.py
+++ b/tests/models/efficientloftr/test_image_processing_efficientloftr.py
@@ -129,6 +129,7 @@ class EfficientLoFTRImageProcessingTest(SuperGlueImageProcessingTest, unittest.T
             fast_time, slow_time * 1.2, "Fast processor should not be significantly slower than slow processor"
         )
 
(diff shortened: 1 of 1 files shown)


vasqu:
We should probably look into something along #42247 but for any type of test

tests/models/efficientloftr/test_image_processing_efficientloftr.py
```diff
@@ -128,3 +128,7 @@ def test_fast_is_faster_than_slow(self):
         self.assertLessEqual(
             fast_time, slow_time * 1.2, "Fast processor should not be significantly slower than slow processor"
         )
+
+    @unittest.skip(reason="FIXME @yonigozlan failing after #42018")
```
ydshieh:
```suggestion
    # TODO: FIXME @yonigozlan 
    @unittest.skip(reason="failing after #42018")
```

ydshieh (approved):

ebezzam (approved):
trust you guys on this, thanks!

PR #42337, opened 2025-11-22
Bart config doesn't need generation parameters

zucchini-nlp:
# What does this PR do?

As per title, fixes slow tests for BART and MVP


zucchini-nlp:
slow bart test won't pass anyway due to weight tying, so skipping it for now

src/transformers/generation/configuration_utils.py
```diff
@@ -436,6 +436,13 @@ def __init__(self, **kwargs):
         self._commit_hash = kwargs.pop("_commit_hash", None)
         self.transformers_version = kwargs.pop("transformers_version", __version__)
 
+        # Ensure backward compatibility for BART CNN models
```
```suggestion
        # Ensure backward compatibility for models that use `forced_bos_token_id` within their config
```
might affect more, nit

src/transformers/models/bart/configuration_bart.py
```diff
@@ -80,9 +78,6 @@ class BartConfig(PreTrainedConfig):
             Whether or not the model should return the last key/values attentions (not used by all models).
         num_labels (`int`, *optional*, defaults to 3):
             The number of labels to use in [`BartForSequenceClassification`].
-        forced_eos_token_id (`int`, *optional*, defaults to 2):
-            The id of the token to force as the last generated token when `max_length` is reached. Usually set to
-            `eos_token_id`.
```
vasqu, line 85: More cleanup from last time?

vasqu (approved):
Thx LGTM, just a small nit

And could you also include another fix (I checked locally with #42355 that everything works). https://github.com/huggingface/transformers/blob/6940b44d8d10345d32fb4593df6748f1c47e700d/tests/models/bart/test_modeling_bart.py#L1282
This shouldn't be passed in `from_pretrained` anymore and we should rather pass this to generate, i.e. `model.generate(..., forced_bos_token_id=0)`

src/transformers/models/bart/configuration_bart.py
zucchini-nlp: yeah, these would not be assigned as config attr so the docs is a bit misleading now

vasqu:
run-slow: bart

PR #42338, opened 2025-11-22
Fix unhelpful error message for Voxtral tokenizer

engmohamedsalah:
## Summary

Fixes #41553

When loading certain LLaMA-style tokenizers (like Voxtral) without the required external package (`mistral-common`) or without a `tokenizer.model` file present, the tokenizer attempts to call:

```python
sentencepiece.Load(None)
```

This results in an unhelpful error:

```
TypeError: not a string
```

This PR adds a defensive check in `LlamaTokenizer.get_spm_processor()`: if `vocab_file` is None or invalid, a clear and actionable `ValueError` is raised instead.

## Changes

**Modified files:**
- `src/transformers/models/llama/tokenization_llama.py`: Added validation check before `tokenizer.Load()`
- `tests/models/llama/test_tokenization_llama.py`: Added test for the new error handling

## Behavior Before
- LlamaTokenizer calls `sentencepiece.Load(None)`
- User sees only: `TypeError: not a string`
- No guidance on how to fix the issue

## Behavior After
- If `vocab_file` is None or invalid, raises a clear `ValueError`:
```
ValueError: Cannot load the tokenizer: vocab_file is 'None'. This model appears to require special tokenization support. Please install the required package with: pip install mistral-common
```

## Why This Helps
Many model repositories depend on external packages or separate tokenizer assets. This improvement makes debugging straightforward and helps users understand exactly what they need to install.

## Testing
- Added `test_missing_vocab_file_error_message()` to verify proper error handling
- Test confirms that a helpful `ValueError` is raised with actionable guidance when `vocab_file=None`
- Manually tested the fix with the scenario described in the issue

## Checklist
- [x] Fix addresses the issue described in #41553
- [x] Added test for the new error handling
- [x] Test passes locally
- [x] Code follows repository style guidelines

### DIFF
--- a/src/transformers/models/llama/tokenization_llama.py
+++ b/src/transformers/models/llama/tokenization_llama.py
@@ -191,7 +191,6 @@ class LlamaTokenizer(PreTrainedTokenizer):
     def unk_token_length(self):
         return len(self.sp_model.encode(str(self.unk_token)))
 
-    # Copied from transformers.models.t5.tokenization_t5.T5Tokenizer.get_spm_processor
     def get_spm_processor(self, from_slow=False):
         tokenizer = spm.SentencePieceProcessor(**self.sp_model_kwargs)
         if self.legacy or from_slow:  # no dependency on protobuf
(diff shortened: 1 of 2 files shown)


Rocketknight1:
Isn't that issue fixed/closed already?

engmohamedsalah:
@Rocketknight1 
You are correct. we can close this PR

PR #42355, opened 2025-11-24
Fix tied weight for Bart (for BC)

Cyrilvallez:
# What does this PR do?

As per the title. Bart used to check which weight was present instead of simply using the default one, as some main checkpoint have the wrong one saved. See [here](https://github.com/huggingface/transformers/pull/41580/files#diff-5b256a6fcfe7956b744dc1e902f150a79e2692b822a6dbac2c6612f8690846b6L917-L928) before we refactored the tied weight

cc @vasqu as you noticed the issue first!

### DIFF
--- a/src/transformers/models/bart/modeling_bart.py
+++ b/src/transformers/models/bart/modeling_bart.py
@@ -897,6 +897,22 @@ class BartModel(BartPreTrainedModel):
         # Initialize weights and apply final processing
         self.post_init()
 
+    def tie_weights(self, *args, **kwargs):
+        """We need to overload here to handle the wrong key saved in some main checkpoints."""
+        super().tie_weights(*args, **kwargs)
+        if self.config.tie_word_embeddings:
+            # Some model checkpoints like "facebook/bart-large-cnn"'s embedding weight is in decoder.embed_tokens, 
+            # need check here, see issue #36247
+            if self.shared.weight.device == torch.device(
+                "meta"
+            ) and self.decoder.embed_tokens.weight.device != torch.device("meta"):
+                self._tie_embedding_weights(self.encoder.embed_tokens, self.decoder.embed_tokens)
+                self._tie_embedding_weights(self.shared, self.decoder.embed_tokens)
+            else:
+                self._tie_embedding_weights(self.encoder.embed_tokens, self.shared)
+                self._tie_embedding_weights(self.decoder.embed_tokens, self.shared)
+
+
     def get_input_embeddings(self):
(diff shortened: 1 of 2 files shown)


src/transformers/models/bart/modeling_bart.py
```diff
@@ -1343,6 +1388,21 @@ def __init__(self, config):
         # Initialize weights and apply final processing
         self.post_init()
 
+    def tie_weights(self, missing_keys: Optional[set[str]] = None, recompute_mapping: bool = True):
```
vasqu, line 1391: Do we want to add a FIXME/TODO here to cleanup after allowing the reverse direction in tied weights?

src/transformers/models/bigbird_pegasus/modeling_bigbird_pegasus.py
```diff
@@ -2183,7 +2183,7 @@ def forward(
     The BigBirdPegasus Model with a language modeling head. Can be used for summarization.
     """
 )
-# Copied from transformers.models.bart.modeling_bart.BartForConditionalGeneration with Bart->BigBirdPegasus, BART->BIGBIRD_PEGASUS
+# Except `tie_weights()`: everything else Copied from transformers.models.bart.modeling_bart.BartForConditionalGeneration with Bart->BigBirdPegasus, BART->BIGBIRD_PEGASUS
```
vasqu: These other classes like QnA are always a mess in legacy models :cry: 

Tbh, I don't see much being copied in the first place: The forward uses ignore copy and then only init and resize_xxx remain if I see it correctly. I'd pro to just remove the copies altogether or just directly use them on the 3 ish functions.

vasqu (approved):
Thx a lot, checked locally (+ with other different fixes) and the integration tests pass then. Just 2 nits but feel free to ignore, not super important

src/transformers/models/bigbird_pegasus/modeling_bigbird_pegasus.py
Cyrilvallez: Oh indeed, did not notice that `forward` was skipping copy, good catch! Will update on the separate functions then!

src/transformers/models/bart/modeling_bart.py
Cyrilvallez: We will need to remove all model-specific `tie_weights` anyway at that time, so fine like this IMO. A TODO does not add much value as we need to search for it and we don't always do anyway

PR #42357, opened 2025-11-24
Make benchmarking lighter: clean-up result files and remove non-needed arguments

remi-or:
This PR simplifies the benchmarking suite by:
- removing the timestamps for each token generation from the result file: we now only keep the time to first token and the (mean) inter-token latency
- removes the SDPA backend argument which is both un-used in practice and introduces a lot of edge cases
- starts the refactoring of `time_generate_batch` and fixes some logic in it (it was not recording the right throughput / results)
- adds a bit more context to loading bars and removes some messages


ArthurZucker (approved):
lgtm

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -828,7 +828,7 @@ def stop(self, block: bool = True, timeout: Optional[float] = None) -> None:
                     f"\nPrefix sharing was on. Total prefix length: {self.batch_processor.cache._total_prefix_length}"
                 )
             else:
-                logger.warning("\nPrefix sharing was off.")
+                logger.debug("\nPrefix sharing was off.")  # NOTE: not warning because this bring no inforantion
```
McPatate:
```suggestion
```
If it brings no information, might as well delete?
Also typo 😄

benchmark_v2/framework/benchmark_config.py
```diff
@@ -18,17 +19,28 @@
...
+        return True
+    # Early return if kernels is not available
+    if not is_kernels_available():
+        return False
+    # If kernels is available, try to get the flash_attn_2 kernel
+    try:
+        from kernels import get_kernel
+
+        get_kernel("kernels-community/flash-attn")
+    except Exception as _:
```
McPatate, line 40: Maybe we can add a warning with the err?

benchmark_v2/framework/benchmark_config.py
```diff
@@ -75,34 +85,21 @@ def check_validity(self, skip_validity_check: bool = False) -> None:
         if skip_validity_check:
             return
         # Check FA is installed
-        if self.attn_implementation == "flash_attention_2" and not is_flash_attn_2_available():
-            logger.warning(
-                "Flash attention does not support compile mode. Defaulting to SDPA w/ flash attention backend."
-            )
+        is_fa = self.attn_implementation == "flash_attention_2"
+        if is_fa and not is_fa2_or_kernel_available():
+            logger.warning("Flash attention is not available. Defaulting to SDPA.")
```
McPatate, line 98: Figured you might've added one here, but you lose the reason of the import error stored in the exc.

Just a nit, feel free to ignore

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -281,49 +258,32 @@ def time_generate_batch(
...
-            finished_requests = 0
-            while finished_requests < num_requests:
+            unfinished_requests = len(inputs)
+            while unfinished_requests > 0:
                 # NOTE: I don't like having the extra if stmt here, but hopefully won't degrade perf too much
                 result = manager.get_result()
-                if result:
-                    timestamps.append(time.perf_counter() - wall_time_0)
+                if result is not None:
+                    timestamps.append(time.perf_counter() - wall_time_0) # FIXME: the timestamps are wrong
```
McPatate: weren't you removing them?

McPatate (approved):
little comments, feel free to ignore, lgtm otherwise!

benchmark_v2/framework/benchmark_runner.py
remi-or: We still need to get the timestamps to have TTFT and ITL, but we are not saving them

benchmark_v2/framework/benchmark_config.py
remi-or: Ok, will not change this warning but will add one in the function above

benchmark_v2/framework/benchmark_runner.py
McPatate: You just need the timestamp of the start of the benchmarking job no?

McPatate: In any case, I'd fix before merging!

remi-or: No, we need the first generated token timestamp and the last one. It's not exactly when the request is scheduled
I will do very soon a PR that fixes this, but will change the timing function to benchmark the API users use -- which is `generate_batch`. It will remove a bit of code and add a new request class, so it will touch more CB. I would prefer to keep the two separate!

benchmark_v2/framework/benchmark_config.py
remi-or: yup

PR #42359, opened 2025-11-24
🚨 Image-text pipeline expects correctly formatted chat

zucchini-nlp:
# What does this PR do?

As per title, we can break for v5 and enforce a correct usage of chat templates. I will request review after adjusting tests

### DIFF
--- a/src/transformers/pipelines/image_text_to_text.py
+++ b/src/transformers/pipelines/image_text_to_text.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 import enum
-from collections.abc import Iterable
 from typing import Any, Union, overload
 
 from ..generation import GenerationConfig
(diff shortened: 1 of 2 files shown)


zucchini-nlp:
cc @ebezzam as discussed internally. I didn't move the `Chat` anywhere in this PR, ig it's already a part of the PR for `text-to-audio`

tests/pipelines/test_pipelines_image_text_to_text.py
```diff
@@ -196,7 +196,21 @@ def test_model_pt_chat_template(self):
...
+
+        messages = [
+            {
+                "role": "user",
+                "content": [
+                    {"type": "text", "text": "What’s the difference between these two images?"},
+                    {"type": "image", "url": image_ny},
+                    {"type": "image", "url": image_chicago},
+                ],
+            }
```
molbap, line 211: much better. It's already documented but shall we advertise also `"path"` here in addition to `"url"`? Just for completeness.

molbap (approved):
Way cleaner, thanks :+1:

src/transformers/pipelines/image_text_to_text.py
```diff
@@ -332,13 +282,19 @@ def _is_chat(arg):
...
 
         if _is_chat(text):
+            if images is not None:
+                raise ValueError(
+                    "Invalid input: you passed `chat` and `images` as separate input arguments. "
+                    "Images must be placed inside the chat message's `content`. For example, "
+                    "'content': ["
+                    "      {'type': 'image', 'url': 'image_url'}, {'type': 'text', 'text': 'Describe the image.'}}"
+                    "]"
+                )
```
yonigozlan, line 292: Should we mention that this is a deprecated behavior in >v5? Otherwise users might not understand why their previously working code is now crashing

yonigozlan (approved):
Nice to see this cleaned up! I don't see anything wrong with removing that for v5 if everything is now correctly handled inside chat templates.

Rocketknight1:
Big fan of this cleanup!

tests/pipelines/test_pipelines_image_text_to_text.py
zucchini-nlp: The chat template are tested with local path and url in processing tests. Since the pipeline internally only calls `apply_chat_template`, I think it isn't necessary

src/transformers/pipelines/image_text_to_text.py
zucchini-nlp: Right, I merged assuming I've read all comments. Sorry. Yeah, saying it's broken for v5 might be more clear, ig users will realize v5 broke a lot of stuff 🙃

src/transformers/pipelines/image_text_to_text.py
```diff
@@ -55,62 +54,13 @@ class Chat:
...
+    def __init__(self, messages: dict):
         for message in messages:
             if not ("role" in message and "content" in message):
                 raise ValueError("When passing chat dicts as input, each dict must have a 'role' and 'content' key.")
-        messages = add_images_to_messages(messages, images)
-
         self.messages = messages
 
 
-def add_images_to_messages(messages: dict, images: Union[str, list[str], "Image.Image", list["Image.Image"]] | None):
```
ebezzam, line 69: @zucchini-nlp I don't think the standardization is as simple as removing this function causes unfortunately :/  

As [this test](https://github.com/huggingface/transformers/blob/f77950605e2f2f27613b829559b133c2b9fd8912/tests/pipelines/test_pipelines_image_text_to_text.py#L363) to fails because `add_images_to_messages` was actually modifying the chat template when there were URLs
```
# RUN_SLOW=1 pytest tests/pipelines/test_pipelines_image_text_to_text.py::ImageTextToTextPipelineTests::test_model_pt_chat_template_image_url

    @slow
    @require_torch
    def test_model_pt_chat_template_image_url(self):
        pipe = pipeline("image-text-to-text", model="llava-hf/llava-interleave-qwen-0.5b-hf")
        messages = [
            {
                "role": "user",
                "content": [
                    {
                        "type": "image_url",
                        "image_url": {
                            "url": "https://cdn.britannica.com/61/93061-050-99147DCE/Statue-of-Liberty-Island-New-York-Bay.jpg"
                        },
                    },
                    {"type": "text", "text": "Describe this image in one sentence."},
                ],
            }
        ]
        outputs = pipe(text=messages, return_full_text=False, max_new_tokens=10)[0]["generated_text"]
>       self.assertEqual(outputs, "A statue of liberty in the foreground of a city")
E       AssertionError: "A close-up of a person's face with a" != 'A statue of liberty in the foreground of a city'
E       - A close-up of a person's face with a
E       + A statue of liberty in the foreground of a city

tests/pipelines/test_pipelines_image_text_to_text.py:380: AssertionError
```

src/transformers/pipelines/image_text_to_text.py
```diff
@@ -55,62 +54,13 @@ class Chat:
...
-                    if idx_images < len(images):
-                        # Insert the image passed as argument in the chat message
-                        content["image"] = images[idx_images]
-                        idx_images += 1
-                    else:
-                        raise ValueError(
-                            "The number of images in the chat messages should be the same as the number of images passed to the pipeline."
-                        )
-            # Add support for OpenAI/TGI chat format
-            elif content_type == "image_url":
```
ebezzam, line 94: I think this was the part rewrittting

ebezzam (commented):
@zucchini-nlp I think there's still a fix needed to handle chat templates with URL?

src/transformers/pipelines/image_text_to_text.py
zucchini-nlp: fixing it to support OpenAI format in pipelines

PR #42363, opened 2025-11-24
fix tekken pattern matching

ArthurZucker:
# What does this PR do?
Gotta add a test!


vasqu (approved):
LGTM, thx for fixing so quickly

Issue #41856, opened 2025-10-25
Performance regression: `allow_is_causal_skip` incorrectly disabled when `use_cache=False`


williamsnell:
### System Info

- `transformers` version: 4.54.0.dev0
- Platform: Linux-6.6.105+-x86_64-with-glibc2.35
- Python version: 3.12.12
- Huggingface_hub version: 0.35.3
- Safetensors version: 0.6.2
- Accelerate version: 1.11.0
- Accelerate config:    not found
- DeepSpeed version: not installed
- PyTorch version (accelerator?): 2.8.0+cu126 (CUDA)
- Tensorflow version (GPU?): 2.19.0 (True)
- Flax version (CPU?/GPU?/TPU?): 0.10.7 (gpu)
- Jax version: 0.7.2
- JaxLib version: 0.7.2
- Using distributed or parallel set-up in script?: no
- Using GPU in script?: yes
- GPU type: Tesla T4

### Who can help?

@Cyrilvallez 

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

Full colab reproduction: https://colab.research.google.com/drive/1vQ_UgSKMIlKMZGI7AM4DjqbWtYRTunIE#scrollTo=SuMxmbBBj4zx

Otherwise:
1. `install transformers <= 4.53.0`
2. run pretraining with `use_cache=False` with, for example, `gpt_neox` (using SDPA as the attention backend)
3. compare iteration time against `transformers >= 4.53.1` There should be a ~10-30% performance regression, depending on the exact model, context length, etc.

### Expected behavior

I noticed a ~25% drop in throughput when pretraining a variant of `EleutherAI/pythia-14m`, after upgrading transformers. I bisected the commit that caused this drop to [0cf27916](https://github.com/huggingface/transformers/commit/0cf27916f09a1a99af55ef4f2f3e8675372f38b6), which introduced packed tensor masks.

The issue seems to be that `allow_is_causal_skip` gets set to False whenever `use_cache=False`. This prevents SDPA using the fast-path when we don't provide an attention mask and are using causal attention.

Looking in [create_causal_mask](https://github.com/huggingface/transformers/blob/main/src/transformers/masking_utils.py#L801), `allow_is_causal_skip` is disabled because `packed_sequence_mask` is not `None`:

```python
[...]

PR #42367, opened 2025-11-24
[`Attn Masks`] Add skip option for non-packed sequences

vasqu:
Given a non-packed sequence, no attn mask, and no cache: We currently still create a mask for this. Related to #41856 for example

This PR changes the behavior under non-compile. The compile-based versions are trickier, maybe a part 2 in the future.


Cyrilvallez (approved):
Very nice that you added a test as well! Thanks!

ArthurZucker (approved):

PR #42368, opened 2025-11-24
Fix tp

ArthurZucker:
# What does this PR do?
There was something that slipped up


src/transformers/core_model_loading.py
```diff
@@ -718,7 +718,7 @@ def convert_and_load_state_dict_in_model(
                         mapping.distributed_operation = tp_layer(
                             device_mesh=device_mesh, rank=device_map[""].index, empty_param=empty_param.clone()
                         )
-                    shard_index = len(mapping.collected_tensors)
+                    shard_index = len(mapping.collected_tensors.get(original_key, []))
```
ArthurZucker, line 721: we need original key index to know how many tensors we collected that match the pattern

Issue #42378, opened 2025-11-25
Attribute error for AutoTokenizer when loading HF repo from older transformers versions :)


nil0x9:
### System Info

transformers==4.57.2

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

```python
from transformers import AutoTokenizer
HF_MODEL_PATH='some/repo/with/old/transformers/version'
tokenizer = AutoTokenizer.from_pretrained(HF_MODEL_PATH)
```

Error Msg:

```bash
File ~/miniconda3/lib/python3.13/site-packages/transformers/tokenization_utils_base.py:2419, in PreTrainedTokenizerBase._from_pretrained(cls, resolved_vocab_files, pretrained_model_name_or_path, init_configuration, token, cache_dir, local_files_only, _commit_hash, _is_local, trust_remote_code, *init_inputs, **kwargs)
   2416     transformers_version = _config.get("transformers_version")
   2418     if transformers_version and version.parse(transformers_version) <= version.parse("4.57.2"):
-> 2419         if _is_local and _config.model_type not in [
   2420             "mistral",
   2421             "mistral3",
   2422             "voxstral",
   2423             "ministral",
   2424             "pixtral",
   2425         ]:
   2426             return tokenizer
   2428 # Expose the `fix_mistral_regex` flag on the tokenizer when provided, even if no correction is applied.

AttributeError: 'dict' object has no attribute 'model_type'
```

### Expected behavior

load success.

ehoogeveen-medweb:
I'm guessing that `<=` should be a `>=`.

nil0x9:
> I'm guessing that `<=` should be a `>=`.

nah, the root cause is that they load a dict config object and then use attribute instead of key to access model_type.

I think hf team should really enhance their code quality, CI and review process.

https://github.com/huggingface/transformers/pull/42379

vasqu:
Fix is in #42389

vasqu:
Sorry about that. This is only happening on local tokenizers and not loading from a repo so it went undetected - let me know if the PR doesn't fix the issue

chtruong814:
Similarly, if we try to load a tokenizer with local_files_only, it fails.  For example
```
AutoTokenizer.from_pretrained('meta-llama/Meta-Llama-3-8B', local_files_only=True)
```

[...]

PR #42379, opened 2025-11-25
Fix _config attr error in tokenization_utils_base.py :)

nil0x9:
# What does this PR do?

Fixes # ([issue](https://github.com/huggingface/transformers/issues/42378))


ArthurZucker:
Ty we are indeed fixing it in #42389!

PR #42382, opened 2025-11-25
[core] fix mxfp4

MekkCyber:
# What does this PR do?

This pr fixes MxFP4 quantization

### DIFF
--- a/src/transformers/integrations/mxfp4.py
+++ b/src/transformers/integrations/mxfp4.py
@@ -69,6 +69,60 @@ def on_device(dev):
     # other: CPU
     yield
 
+class Mxfp4Quantize(ConversionOps):
+    def __init__(self, hf_quantizer):
+        self.hf_quantizer = hf_quantizer
+    def convert(
+        self, input_dict: dict[str, torch.Tensor], model: Optional[torch.nn.Module] = None, missing_keys: Optional[list[str]] = None, **kwargs
+    ) -> dict[str, torch.Tensor]:
+        target_key, value = tuple(input_dict.items())[0]
+        value = value[0] if isinstance(value, list) else value
+        if not self.hf_quantizer.pre_quantized:
+            module, _ = get_module_from_name(model, target_key)
+            with torch.device(value.device):
+                if isinstance(module, Mxfp4GptOssExperts):
+                    triton_weight_tensor, weight_scale = quantize_to_mxfp4(value, triton_kernels_hub)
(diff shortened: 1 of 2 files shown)


src/transformers/integrations/mxfp4.py
```diff
@@ -70,6 +76,102 @@ def on_device(dev):
...
+        full_layer_name: str | None = None,
+        **kwargs,
+    ) -> dict[str, torch.Tensor]:
+        _, value = tuple(input_dict.items())[0]
+        value = value[0] if isinstance(value, list) else value
+
+        module, _ = get_module_from_name(model, full_layer_name)
+
+        with torch.device(value.device):
+            if isinstance(module, Mxfp4GptOssExperts):
```
ArthurZucker, line 97: we only quantize experts for mxp4? can we try to use mxfp4 quantization for arbitrary models? Say mixtral ? Or just not worth it?

ArthurZucker (approved):
wow if this works OOB, its a testament to how simple it is!

src/transformers/quantizers/quantizer_mxfp4.py
```diff
@@ -426,3 +430,21 @@ def is_trainable(self) -> bool:
...
+
+        if self.pre_quantized:
+            return [
+                WeightConverter(
+                    source_keys=["_blocks", "_scales"],
+                    target_keys="",
+                    operations=[Mxfp4DequantizeOrSwizzle(self)],
+                )
+            ]
+        return []
```
SunMarc, line 459: I think it will be clearer to have one dequantize ops and one for deserialize so that we keep the naming consistent

src/transformers/integrations/mxfp4.py
```diff
@@ -70,6 +76,102 @@ def on_device(dev):
...
+                proj,
+                param_data["_blocks"].device,
+                triton_kernels_hub,
+            )
+            missing_keys.discard(f"{full_layer_name}")
+            # We return an empty mapping since the module was updated in-place. This prevents
+            # the loader from trying to materialize the original meta-parameter names again.
+            # We don't use set_param_for_module since it expects mainly a torch.nn.Parameter or a safetensors pointer
+            return {}
+
```
SunMarc: don't we need to set somewhere `module._if_hf_initialized` just in case ?

SunMarc (commented):
Thanks, just a minor nit for consistency

PR #42385, opened 2025-11-25
Fix weight tying logic between `_tied_weights_keys` and `tie_word_embeddings`

molbap:
# What does this PR do?


molbap:
run-slow: mistral

molbap:
run-slow: mistral3

molbap:
run-slow: mistral, mistral3

molbap:
run-slow: olmoe

molbap:
run-slow: olmoe, mistral, mistral3

molbap:
run-slow: olmoe, mistral, mistral3, phi3, starcoder

molbap:
run-slow: olmoe, mistral, mistral3, phi3, starcoder, bigbird, bigbird_pegasus

molbap:
run-slow: olmoe, mistral, mistral3, phi3, starcoder, bigbird, bigbird_pegasus, whisper, internvl, llava, llava_next, llava_next_video, qwen

src/transformers/modeling_utils.py
```diff
@@ -3174,7 +3176,11 @@ def save_pretrained(
             shared_ptrs = {ptr: names for ptr, names in ptrs.items() if len(names) > 1}
 
             # Recursively descend to find tied weight keys
-            _tied_weights_keys = set(_get_tied_weight_keys(self))
+            tied_keys_attr = getattr(self, "all_tied_weights_keys", None)
+            if tied_keys_attr is not None:
+                _tied_weights_keys = set(tied_keys_attr.keys())
+            else:
+                _tied_weights_keys = set(_get_tied_weight_keys(self))
```
molbap: This is the main change. IMO it's why `tie_word_embeddings` is not authoritative rn, it is entirely skipped if `tied_weights_keys` is filled

src/transformers/conversion_mapping.py
```diff
@@ -117,6 +117,21 @@ def _build_checkpoint_conversion_mapping():
     mapping["qwen3_vl_moe"] = mapping["qwen2_moe"].copy()
     mapping["hunyuan_v1_moe"] = mapping["qwen2_moe"].copy()
     mapping["minimax"] = mapping["mixtral"].copy()
+    mapping["olmoe"] = [
```
ArthurZucker: same as qwen2_moe no?

src/transformers/modeling_utils.py
```diff
@@ -2248,8 +2248,10 @@ def get_expanded_tied_weights_keys(self, all_submodels: bool = False) -> dict:
             return expanded_tied_weights
 
         tied_mapping = self._tied_weights_keys
+        text_config = self.config.get_text_config(decoder=True)
+        tie_word_embeddings = getattr(text_config, "tie_word_embeddings", self.config.tie_word_embeddings)
```
ArthurZucker: this is what we were trying to avoid but it might not really be possible as vlms do indeed just rely on child's ?

ArthurZucker (commented):
Looks good, I think we can also check when we read config.json if the `tie_word_embeddings` is `None` maybe we should put it to False? IDK but TBF that's what is looks like it would fix.

For my comment about subconfig, might be non avoidable.

Thanks a lot for taking a look!

src/transformers/modeling_utils.py
molbap: yes I think you're correct unfortunately, not 100% sure. invoking @zucchini-nlp 's brain here

src/transformers/conversion_mapping.py
molbap: yes totally the same, I thought I had committed it :facepalm:

molbap:
> Looks good, I think we can also check when we read config.json if the tie_word_embeddings is None maybe we should put it to False?

Sounds reasonable to me, putting this

molbap:
run-slow: olmoe, mistral, mistral3, phi3, starcoder, bigbird, bigbird_pegasus, whisper, internvl, llava, llava_next, llava_next_video, qwen, fsmt

src/transformers/modeling_utils.py
zucchini-nlp: agreed with Arthur, I think `self.config.get_text_config(decoder=True).tie_word_embeddings` is equivalent of what we need. The fallback will probably just hide the bug if some configs are implemented poorly

src/transformers/modeling_utils.py
zucchini-nlp: yeah, i think we fill it in in many cases just to pass some tests even if the released models don't tie weights

zucchini-nlp: Also, I am a bit lost here. Do we snow try to tie weights with `_tied_weights_keys` attribute even when `config.tie_word_embeddings = False`?

molbap:
run-slow: olmoe, mistral, mistral3, phi3, starcoder, bigbird, bigbird_pegasus, whisper, internvl, llava, llava_next, llava_next_video, qwen, fsmt, video_llava, deepseek_v3

molbap:
run-slow: olmoe, mistral, mistral3, phi3, starcoder, bigbird, bigbird_pegasus, whisper, internvl, llava, llava_next, llava_next_video, qwen, fsmt, video_llava, deepseek_v3, qwen3_vl_moe

src/transformers/modeling_utils.py
molbap: yes, it's what I observed. Tried to get a reproducer here @zucchini-nlp @ArthurZucker 

```python
from transformers import PaliGemmaConfig, PaliGemmaForConditionalGeneration
from transformers.models.gemma.configuration_gemma import GemmaConfig
from transformers.models.siglip.configuration_siglip import SiglipVisionConfig

vision_config = SiglipVisionConfig(
    hidden_size=32,
    intermediate_size=64,
    num_hidden_layers=1,
    num_attention_heads=2,
    image_size=32,
    patch_size=4,
    vocab_size=64,
    projection_dim=32,
).to_dict()

text_config = GemmaConfig(
    vocab_size=64,
    hidden_size=32,
    intermediate_size=64,
    num_hidden_layers=1,
    num_attention_heads=4,
    num_key_value_heads=4,
    tie_word_embeddings=False,
).to_dict()

config = PaliGemmaConfig(vision_config=vision_config, text_config=text_config)

print("Config tie flags:")
print(f"  config.tie_word_embeddings              -> {config.tie_word_embeddings}")
print(f"  config.text_config.tie_word_embeddings  -> {config.text_config.tie_word_embeddings}")

model = PaliGemmaForConditionalGeneration(config)
lm_head_weight = model.lm_head.weight
input_embed_weight = model.model.language_model.embed_tokens.weight
tied = lm_head_weight.data_ptr() == input_embed_weight.data_ptr()

print("\nModel tying details:")
print(f"  class _tied_weights_keys      -> {model._tied_weights_keys}")
print(f"  model.all_tied_weights_keys   -> {model.all_tied_weights_keys}")
print(f"  lm_head shares embedding tensor? -> {tied}")

```
this will show tied weights. However if you go to modeling_paligemma and change 

```python
    _tied_weights_keys = {}# {"lm_head.weight": "model.language_model.embed_tokens.weight"}
```

then it will work and weights will not be tied. I think the solution is to use `all_tied_weights_keys` instead of `_get_tied_weights_keys` here. It's what I understand from the logic flow

and for instance for `mistral3` it breaks the model generation, but for several others as well, this is why I wanted the `run-slow` upgrade mentioned hehe.

There might be a simpler fix, not seeing it right now

src/transformers/core_model_loading.py
```diff
@@ -504,7 +504,7 @@ def set_param_for_module(
             missing_keys.discard(target_name)
             if ref is not None and ref.shape != param_value.shape and hf_quantizer is None:
                 mismatch_keys.add((target_name, param_value.shape, ref.shape))
-                module_obj.param_name._is_hf_initialized = False  # Needs to be initialized
+                ref._is_hf_initialized = False  # unsure - but seems unreachable else?
```
ArthurZucker: Mmm that's not needed ref won't be initialized at all since we don't setattr

ArthurZucker: but line can be removed yes

molbap: removing!

molbap:
I removed the fallback to parent config in case `tie_word_embeddings` is not found in the text. I also attempted to add a test, might be a bit overkill, let's see

src/transformers/modeling_utils.py
molbap: done, removed the fallback and added a warning!

molbap:
A remaining headscratcher is `tie_encoder_decoder` which is now redundant. I did a fallback that looks like 
```python
should_tie = tie_encoder_decoder if tie_word_embeddings is None else tie_word_embeddings
```
and it seems to work/keep the key as source of authority. LMK!

zucchini-nlp:
> A remaining headscratcher is tie_encoder_decoder which is now redundant. I did a fallback that looks like

CMIIW, `tie_encoder_decoder` means tying all weights from encoder to decoder, i.e. attention and stuff isn't it? In which case 

`tie_encoder_decoder = config.tie_encoder_decoder if config.is_encoder_decoder else False`?

src/transformers/core_model_loading.py
```diff
@@ -504,7 +504,6 @@ def set_param_for_module(
             missing_keys.discard(target_name)
             if ref is not None and ref.shape != param_value.shape and hf_quantizer is None:
                 mismatch_keys.add((target_name, param_value.shape, ref.shape))
-                module_obj.param_name._is_hf_initialized = False  # Needs to be initialized
```
Cyrilvallez, line 597: Yes, was completely masked by the `log_to_misc`, along with other issues we have currently on some models. We will need to update the logger to avoid silencing important issues!!

src/transformers/modeling_utils.py
```diff
@@ -2254,8 +2254,17 @@ def get_expanded_tied_weights_keys(self, all_submodels: bool = False) -> dict:
             return expanded_tied_weights
 
         tied_mapping = self._tied_weights_keys
+        text_config = self.config.get_text_config(decoder=True)
+        if not hasattr(text_config, "tie_word_embeddings"):
+            logger.warning(
+                f"Text config {text_config.__class__.__name__} does not have 'tie_word_embeddings' attribute. "
+                "This may cause issues with weight tying."
+            )
+        tie_word_embeddings = getattr(text_config, "tie_word_embeddings", None)
```
Cyrilvallez: I don't think all multimodals rely on the text_config here unfortunately, do they? I.e. what we want is that each model decide on its own if it should tie its own weights. So we don't really want to check the text_config...

src/transformers/modeling_utils.py
```diff
@@ -2254,8 +2254,17 @@ def get_expanded_tied_weights_keys(self, all_submodels: bool = False) -> dict:
...
+        tie_encoder_decoder = getattr(self.config, "tie_encoder_decoder", False)
+        should_tie = tie_encoder_decoder if tie_word_embeddings is None else tie_word_embeddings
```
Cyrilvallez: So we basically disregard `tie_encoder_decoder` completely? Because from your change is `configuration_utils`, `tie_word_embeddings` will NEVER be None anymore if I understand correctly

src/transformers/modeling_utils.py
Cyrilvallez: The fact here is that I think some older models were tying using `_tie_weights` (the one with underscore) that were ALWAYS run, independently of the configs. So they could have tied weights even though the config say to not tie technically. In `save_pretrained`, we check the pointers of the weight tensors, so we know the tied weights from there and we cannot be wrong about it. But then if we use the good citizen `all_tied_weights_keys`, which correctly looks at the config, some of those tied weights will not find their source as they are not SUPPOSED to be tied (but they were anyway).

So technically looking at all potential `_tied_weights_keys` here is not wrong as we check pointers anyway, and this avoids this kind of issues.
But indeed, in the future we want to always know which weights are tied and simply rely on the internal list (or even better, recomputing it with `get_expanded_tied_weights_keys(all_submodels=True)`)

src/transformers/modeling_utils.py
```diff
@@ -4410,7 +4423,9 @@ def _move_missing_keys_from_meta_to_cpu(
         # The tied weight keys are in the "missing" usually, but they should not be moved (they will be tied anyway)
         # This is especially important because if they are moved, they will lose the `_is_hf_initialized` flag, and they
         # will be re-initialized for nothing (which can be quite long)
-        for key in missing_keys - self.all_tied_weights_keys.keys():
+        tied_keys_attr = getattr(self, "all_tied_weights_keys", {}) or {}
+        tied_keys = set(tied_keys_attr.keys())
+        for key in missing_keys - tied_keys:
```
Cyrilvallez: Why do we have this? `all_tied_weights_keys` is guaranteed to be a dict already at this point from `post_init`

Cyrilvallez (commented):
A few comments/answers! Sorry for the delay!

src/transformers/modeling_utils.py
molbap: hmm, it's the opposite of what we discussed above. In that case, what do you suggest?

src/transformers/modeling_utils.py
molbap: yeah, I'm reverting this unfortunately, I would indeed like to get rid of `tie_encoder_decoder`

src/transformers/modeling_utils.py
molbap: > So technically looking at all potential _tied_weights_keys here is not wrong as we check pointers anyway, and this avoids this kind of issues.

Hmm ok, but then why are models like mistral3 broken by the current setup? Ah because you mean we did it in `save_pretrained` but `load_pretrained` has to do with what exists now? 

Not super sure of the way forward there, can you elaborate a bit?

src/transformers/modeling_utils.py
molbap: ah yes indeed, could be just
```python
tied_keys = set(self.all_tied_weights_keys.keys())
  for key in missing_keys - tied_keys:
 do stuff
``` 
will update

src/transformers/modeling_utils.py
Cyrilvallez: Well the issue is that if you have a submodel such as `self.text_model = AutoModel._from_config(text_config)`, you cannot know its weights in advance as it can be any model. So that model is the only responsible for its own weight.
We cannot delegate for the `top-most` model, as this makes it wayyy too hard and not scalable in general

src/transformers/modeling_utils.py
Cyrilvallez: No need to cast it again as a `set` the keys already are a subclass of `set`!

tests/test_modeling_common.py
```diff
@@ -2100,6 +2100,49 @@ def test_tied_weights_keys(self):
...
+            tied_config.get_text_config().tie_word_embeddings = True
+
+            untied_config = copy.deepcopy(original_config)
+            untied_config.get_text_config().tie_word_embeddings = False
+
+            model_tied = model_class(tied_config)
+            model_untied = model_class(untied_config)
+
+            if not hasattr(model_tied, "_tied_weights_keys") or not model_tied._tied_weights_keys:
+                continue
```
molbap, line 2120: note from huddle: should recurse through all submodels here to get the tied keys

PR #42392, opened 2025-11-25
use dict everywhere for core model loading

ArthurZucker:
# What does this PR do?
Make sure we use dict not list should make everyone's life easier


SunMarc (commented):
Nice, thanks for making our life easier *0*. If you can add an example of how the values might look like in `convert`, i think it will be even better.

ArthurZucker:
On it

PR #42396, opened 2025-11-25
[loading/saving] Reverse all loading operations when saving

Cyrilvallez:
# What does this PR do?

As per the title! We can now revert back all our custom Operation performed during weight loading, and it's the default for BC!

### DIFF
--- a/src/transformers/conversion_mapping.py
+++ b/src/transformers/conversion_mapping.py
@@ -28,11 +28,11 @@ def _build_checkpoint_conversion_mapping():
         "mixtral": [
             WeightRenaming(".block_sparse_moe.gate", ".mlp.gate"),
             WeightConverter(
-                source_keys=[
+                source_patterns=[
                     "block_sparse_moe.experts.*.w1.weight",
                     "block_sparse_moe.experts.*.w3.weight",
                 ],  # you give me a list of 2 keys, I collect a list of a list of tensors
-                target_keys="mlp.experts.gate_up_proj",  # target key gets the list of two tensors
+                target_patterns="mlp.experts.gate_up_proj",  # target key gets the list of two tensors
                 operations=[
                     MergeModulelist(
                         dim=0
@@ -41,10 +41,10 @@ def _build_checkpoint_conversion_mapping():
                 ],  # we want the loading to add this shard operation here. Though we can't shard after concats and merge, needs to be first
             ),
             WeightConverter(
-                source_keys=[
+                source_patterns=[
                     "block_sparse_moe.experts.*.w2.weight",
                 ],
-                target_keys="mlp.experts.down_proj",  # target key gets the list of two tensors
+                target_patterns="mlp.experts.down_proj",  # target key gets the list of two tensors
                 operations=[
                     MergeModulelist(
                         dim=0
@@ -54,50 +54,50 @@ def _build_checkpoint_conversion_mapping():
         ],
         "qwen2_moe": [
             WeightConverter(
-                source_keys=[
+                source_patterns=[
                     "mlp.experts.*.gate_proj.weight",
                     "mlp.experts.*.up_proj.weight",
                 ],
-                target_keys="mlp.experts.gate_up_proj",
+                target_patterns="mlp.experts.gate_up_proj",
                 operations=[MergeModulelist(dim=0), Concatenate(dim=1)],
             ),
             WeightConverter(
-                source_keys=["mlp.experts.*.down_proj.weight"],
-                target_keys="mlp.experts.down_proj",
+                source_patterns=["mlp.experts.*.down_proj.weight"],
+                target_patterns="mlp.experts.down_proj",
                 operations=[MergeModulelist(dim=0)],
             ),
         ],
         "legacy": [
             WeightRenaming(
-                source_keys="LayerNorm.gamma",
-                target_keys="LayerNorm.weight",
+                source_patterns="LayerNorm.gamma",
+                target_patterns="LayerNorm.weight",
             ),
             WeightRenaming(
-                source_keys="LayerNorm.beta",
-                target_keys="LayerNorm.bias",
+                source_patterns="LayerNorm.beta",
+                target_patterns="LayerNorm.bias",
             ),
         ],
     }
     if hasattr(torch.nn.utils.parametrizations, "weight_norm"):
         mapping["legacy"] += [
             WeightRenaming(
-                source_keys="weight_g",
-                target_keys="parametrizations.weight.original0",
+                source_patterns="weight_g",
+                target_patterns="parametrizations.weight.original0",
             ),
             WeightRenaming(
-                source_keys="weight_v",
-                target_keys="parametrizations.weight.original1",
+                source_patterns="weight_v",
+                target_patterns="parametrizations.weight.original1",
             ),
         ]
     else:
         mapping["legacy"] += [
             WeightRenaming(
-                source_keys="parametrizations.weight.original0",
-                target_keys="weight_g",
+                source_patterns="parametrizations.weight.original0",
+                target_patterns="weight_g",
             ),
             WeightRenaming(
-                source_keys="parametrizations.weight.original1",
-                target_keys="weight_v",
+                source_patterns="parametrizations.weight.original1",
+                target_patterns="weight_v",
             ),
         ]
 
--- a/src/transformers/conversion_mapping.py
+++ b/src/transformers/conversion_mapping.py
@@ -203,8 +203,9 @@ def get_model_conversion_mapping(
         model_specific_conversions = get_checkpoint_conversion_mapping(model_type)
         if model_specific_conversions is not None:
             weight_conversions.extend(model_specific_conversions)
-        elif add_legacy:
-            weight_conversions.extend(get_checkpoint_conversion_mapping("legacy"))
+
+    if add_legacy:
+        weight_conversions.extend(get_checkpoint_conversion_mapping("legacy"))
 
     # Add the ones from the quantizer as well if provided
     if hf_quantizer is not None:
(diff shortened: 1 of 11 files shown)


src/transformers/core_model_loading.py
```diff
@@ -804,27 +841,36 @@ def convert_and_load_state_dict_in_model(
...
-            replacement = re.sub(r"\(.*\)", "", replacement)
-            key, n_replace = re.subn(pattern, replacement, key)
-            # Early exit of the loop
-            if n_replace > 0:
-                break
-        original_state_dict[key] = value
-    state_dict = original_state_dict
-    return state_dict
+    original_mapping = getattr(model, "_weight_loading_mapping", None)
+    if original_mapping is None:
```
ArthurZucker: well, if we init + train a mistral model then the `_weight_loading_mapping` will be inexistant. 
Fine for now, can come in another PR probaly?

ArthurZucker (commented):
Very nice., missing some fast test in test_core_model_loading! With all the ops that are define in the file to make sure they are all inversible (ex: permute for rope!)

src/transformers/conversion_mapping.py
```diff
@@ -127,5 +140,71 @@ def _build_checkpoint_conversion_mapping():
...
+        for class_name in model.__class__.__mro__[:-1]
+        for allowed_name in VLMS
+    ):
+        weight_conversions = [
+            WeightRenaming(source_patterns=k, target_patterns=v)
+            for k, v in model._checkpoint_conversion_mapping.items()
+        ]
+
+    # TODO: should be checked recursively on submodels!!
+    model_type = getattr(model.config, "model_type", None)
```
ArthurZucker, line 201: yep, but it should be as easy as iterating over the `config.sub_configs`. 

The more "tricky" part is probably checking that none are overlapping but should also be alright if checked at init?

src/transformers/core_model_loading.py
```diff
@@ -122,127 +118,134 @@ def convert(
...
         full_layer_name: str,
         model,
         missing_keys,
         config,
     ) -> dict[str, list[torch.Tensor]]:
-        tensors = next(iter(value.values()))
-        tensor = tensors[0]
-        sizes = len(target_keys)
+        if len(value) > 1:
+            raise ValueError("Unexpected value in `Chunk`!")
```
ArthurZucker: that's restrictive, I remember I had to remove it

src/transformers/conversion_mapping.py
Cyrilvallez: Yes, we will need to be careful but let's do it later! This is quite a specific case that I don't think we have for now!

src/transformers/core_model_loading.py
Cyrilvallez: For the current state of the mappings we have it's fine - we can definitely lift it later, but we have to be extremely careful about the sizes we use in the `torch.chunk` then! It may become undefined with several inputs, as the `target_patterns` are then mixed between them, and we don't know anymore which belong to which!

ArthurZucker: Ok let me play a little bit with `Chunk` and `qkv`, cuz the input shape of tensors for saving and loading should indeed be the same

src/transformers/conversion_mapping.py
ArthurZucker: okay!

src/transformers/core_model_loading.py
```diff
@@ -105,144 +104,152 @@ def build_glob_alternation(
 class ConversionOps:
     """Base class for weight conversion operations."""
 
-    # The inverse operation class, will be used when saving the checkpoint
-    reverse_op: type[ConversionOps]
+    def __repr__(self):
+        return f"{self.__class__.__name__}(dim={self.dim})"
```
ArthurZucker, line 108:
```suggestion
        return f"{self.__class__.__name__}(dim={self.dim}, {self.reverse_op}"
```

src/transformers/core_model_loading.py
```diff
@@ -105,144 +104,152 @@ def build_glob_alternation(
...
+    @property
+    def reverse_op(self) -> ConversionOps:
+        raise NotImplementedError
 
-class Chunk(ConversionOps):
-    """Split a tensor along ``dim`` into equally sized chunks or using explicit ``sizes``."""
 
-    reverse_op: type[ConversionOps]
+class Chunk(ConversionOps):
+    """Split a tensor along ``dim`` into equally sized chunks."""
```
ArthurZucker, line 122:
```suggestion
    """Split a tensor along `dim` into equally sized chunks."""
```

src/transformers/core_model_loading.py
```diff
@@ -283,27 +290,108 @@ def convert(
...
+                ]
+            elif len(ops) == 1 and isinstance(ops[0], SplitModulelist):
+                tensor = future
+                size = tensor.size(ops[0].dim)
+                all_created_targets = [target_key.replace("*", f"{i}") for i in range(size)]
+            # Perform replacement if we took any of the above branches
+            if len(all_created_targets) > 0:
+                self.layer_targets.pop(target_key)
+                for target in all_created_targets:
+                    self.layer_targets[target].add(source_key)
```
ArthurZucker: as discussed offline this is too dirty, these ops themselfves should do it

src/transformers/core_model_loading.py
```diff
@@ -283,27 +290,108 @@ def convert(
...
+            # Find the full names of the params we will need to use later for the reverse transform
+            reverse_layer_targets = defaultdict(set)
+            reverse_collected_tensors = defaultdict(list)
+            for target_key, all_sources in self.layer_targets.items():
+                matched_target_pattern = next(pat for pat in self.target_patterns if re.search(pat, target_key))
+                reverse_collected_tensors[matched_target_pattern].append(target_key)
+                for source in all_sources:
+                    reverse_layer_targets[source].add(target_key)
+            reverse_collected_tensors = {k: sorted(set(v)) for k, v in reverse_collected_tensors.items()}
+            reverse_transform.layer_targets = reverse_layer_targets
```
ArthurZucker: given complexity IDT that worth it to have more code. Let's always re-create it should be "fairly" cheap

src/transformers/core_model_loading.py
```diff
@@ -319,17 +407,25 @@ def convert(
...
-            self.collected_tensors[pattern] = [future.result() for future in futures]
+            self.collected_tensors[pattern] = (
+                futures if isinstance(futures[0], torch.Tensor) else [future.result() for future in futures]
+            )
 
-        collected_tensors = self.collected_tensors
+        # Perform renaming op (for a simple WeightRenaming, `self.source_patterns` and `self.target_patterns` can
+        # only be of length 1, and are actually the full key names - we also have only 1 single related tensor)
+        target_key = self.target_patterns[0]
+        collected_tensors = {target_key: self.collected_tensors[self.source_patterns[0]]}
```
ArthurZucker, line 385: not certain this should be done here either API-wise

src/transformers/core_model_loading.py
```diff
@@ -319,17 +407,25 @@ def convert(
...
+
+        all_target_keys = [target_key]
```
ArthurZucker: weird that we pass a list, of only the first element of all target patterns

src/transformers/core_model_loading.py
```diff
@@ -805,27 +910,94 @@ def convert_and_load_state_dict_in_model(
...
+                renamed_key = weight_pattern_alt.sub(lambda m: repl(m, tgt_group_to_glob), renamed_key).replace(
+                    "\\", ""
+                )
+                new_converter = deepcopy(pattern_to_converter[src_group_to_glob[matched_pattern.lastgroup]])
+                # each target key gets its own converter instance
+                mapping = conversion_mapping.setdefault(renamed_key, new_converter)
+                source_pattern = src_group_to_glob[matched_pattern.lastgroup]
+            else:
+                mapping = conversion_mapping.setdefault(renamed_key, WeightRenaming(original_key, renamed_key))
+                source_pattern = original_key
```
ArthurZucker: looks the same for loading and saving -> could be a utils?

ArthurZucker (commented):
Nice but missing tests 👁️ 👁️ !

src/transformers/core_model_loading.py
Cyrilvallez: Unfortunately this is a recursion Error if we do that, as the `reverse_op` is itself a `ConversionOps`! (i.e. we loop indefinitely between the 2)

src/transformers/core_model_loading.py
```diff
@@ -810,27 +925,91 @@ def convert_and_load_state_dict_in_model(
...
+
+            mapping.add_tensor(renamed_key, original_key, source_pattern, tensor)
+
+    new_state_dict = {}
+    for first_param_name, converter in conversion_mapping.items():
+        # In this case, the mapping was obtained from the loaded model, we need to reverse the ops
+        if need_to_reverse:
+            reversed_converter = converter.reverse_transform()
+            # The `collected_tensors` only contain the full param keys after `reverse_transform` is called, we
+            # need to populate it with the actual tensors
```
ArthurZucker: would love to avoid rever transformers actually preparing the collected tensors. Is it possible?

src/transformers/core_model_loading.py
```diff
@@ -810,27 +925,91 @@ def convert_and_load_state_dict_in_model(
...
-            if n_replace > 0:
-                break
-        original_state_dict[key] = value
-    state_dict = original_state_dict
-    return state_dict
+def revert_weight_conversion(model: PreTrainedModel, state_dict: dict[str, torch.Tensor]):
+    """ "
+    Revert the conversion mapping that was used to load the model with `from_pretrained`, or the default one
+    if the model was created in another way and is part of the default mappings.
+    """
```
ArthurZucker, line 941: an in depth example with the same cases I shared above would be very very nhice!@

ArthurZucker (commented):
Much much much better but missing tests still

ArthurZucker (approved):
Ty

PR #42400, opened 2025-11-25
Seed test for reproducibility

Rocketknight1:
The `test_load_balancing_loss` test is intermittently flaky for me. This PR sets a manual seed to make it more reproducible, which hopefully solves the issue!

### DIFF
--- a/tests/models/ernie4_5_moe/test_modeling_ernie4_5_moe.py
+++ b/tests/models/ernie4_5_moe/test_modeling_ernie4_5_moe.py
@@ -95,6 +95,7 @@ class Ernie4_5_MoeModelTest(CausalLMModelTest, unittest.TestCase):
         r"""
         Let's make sure we can actually compute the loss and do a backward on it.
         """
+        torch.manual_seed(0)
         config, input_dict = self.model_tester.prepare_config_and_inputs_for_common()
         config.num_labels = 3
         config.num_experts = 8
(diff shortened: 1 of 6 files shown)


Rocketknight1:
cc @ydshieh

Rocketknight1:
run-slow: ernie4_5_moe, jamba, minimax, mixtral, qwen2_moe, qwen3_moe

ydshieh:
Hi @Rocketknight1 

Could you share the job run pages , or the full error log you have if you are running some tests locally.

This test doesn't seem to require seed (from what this test is actually testing), so I would like to know what is wrong without seed.

(and the daily CI doesn't have this test failing, only once for mixtral)

Rocketknight1:
@ydshieh the test has a model initialized from `config`, so all of the weights are randomly initialized. This is the bit that causes flakiness, and the bit that we need to set a seed for! There's a random failure about ~1% of the time when the initialization we get behaves badly. People have tried to fix it by setting really wide `atol` and `rtol` values, but I think fixing the seed is a much more sensible approach.

ydshieh:
From my experience of dealing failing tests, use seed in this scenario is a quick hot fix but not the best approach handling it.
Seems it's 1% as you said, maybe we could wait until I check them next week?

Rocketknight1:
Closing since it's superceded by #42509!

PR #42408, opened 2025-11-25
Benchmark simplification

remi-or:
This PR does two things:
- in the benchmark_v2 suite, merge the two `time_generate` function (with and without CB) into one
- rename the the attributes of `RequestState` so they make more sense

This will make the code more readable and concise.

### DIFF
--- a/benchmark_v2/framework/benchmark_runner.py
+++ b/benchmark_v2/framework/benchmark_runner.py
@@ -413,36 +413,38 @@ class BenchmarkRunner:
                 "PUSH_TO_HUB_TOKEN is not set, cannot push results to the Hub. When setting dataset_id, please also set the PUSH_TO_HUB_TOKEN environment variable."
             )
 
+        api = HfApi()
         n_results = len(results)
-        self.logger.info(f"Pushing {n_results} results to: {dataset_id}")
-        rows = []
-        for cfg_hash, entry in results.items():
-            row = {
-                "benchmark_config_hash": cfg_hash,
-                "config": entry["config"].to_dict(),
-                "measurements": entry["measurements"].to_dict(),
-                "metadata": entry["metadata"].to_dict(),
-            }
-            rows.append(row)
+        for include_timestamps in [False, True]:
+            self.logger.info(f"Pushing {n_results} results to: {dataset_id} with {include_timestamps = }")
+            rows = []
+            for cfg_hash, entry in results.items():
+                row = {
+                    "benchmark_config_hash": cfg_hash,
+                    "config": entry["config"].to_dict(),
+                    "measurements": entry["measurements"].to_dict(include_timestamps=include_timestamps),
+                    "metadata": entry["metadata"].to_dict(),
+                }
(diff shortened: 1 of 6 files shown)


benchmark_v2/framework/benchmark_runner.py
```diff
@@ -249,75 +256,51 @@ def run_benchmark(
...
-        # Compute intermediate quantities
+        decoded_output = self.tokenizer.decode(results[0], skip_special_tokens=True)
+        shape_and_decoded_output = f"{tuple(results.shape)} | {decoded_output}"
+
+        # Compute metrics
         e2e_latency = wall_time_1 - wall_time_0
-        token_generation_times = [t - wall_time_0 for t in streamer.timestamps[1:]]
-        return e2e_latency, token_generation_times, shape_and_decoded_output, gpu_metrics
+        time_to_first_token = min(ts[0] - wall_time_0 for ts in timestamps)
+        inter_token_latency = np.mean([(ts[-1] - ts[0]) / (len(ts) - 1) for ts in timestamps])
```
McPatate: Hmm not super convinced, I'd rather keep the raw data as is before doing any aggregations

src/transformers/generation/continuous_batching/requests.py
```diff
@@ -176,13 +185,17 @@ def update_and_check_completion(self, token_id: int) -> bool:
         if self.status != RequestStatus.DECODING:
             return False
 
+        # If we're recording timestamps, add the token ID and timestamp to the list
```
McPatate: You're only adding a timestamp here, no token id, Mr GPT

McPatate (commented):
I'm ok with the ITL "pre"-aggregation if we store the raw data in the benchmark files as well

The idea being that with the raw data, were we to ever decide on computing new metrics we'd have the ability to do so, with a good history

It's not going to be a lot of data on disk, we can keep the timestamps imo

src/transformers/generation/continuous_batching/requests.py
remi-or: Outrageous allegations :o fixed.

benchmark_v2/framework/benchmark_runner.py
remi-or: The whole data serves no purpose: we only use ITL and TTFT, and there is no other popular metric that uses the detailed timestamps. It only serves to render the benchmark result files unreadable and bigger. 
Imo, if we ever find a use for the timestamps, we can start saving them again.

McPatate: chatgpt says they might be useful, for these metrics for instance:
> Latency per token position (is there a degradation after 512 / 1k / 2k tokens?)
>
> Jitter metrics (variance between successive token times).
>
> Tail latencies (P95, P99 token latency, first-N tokens only, last-M tokens only).
>
> “Smoothness” of streaming UX: are tokens clumpy (bursts) vs evenly spaced?

McPatate (approved):

PR #42409, opened 2025-11-25
[docs] Expert parallelism

stevhliu:
updates the parallelism docs:

- distributed inference is now tensor parallelism so it is easier to find
- new expert parallelism guide


ArthurZucker (approved):
cc @3outeille if you want to have a look as well!

stevhliu:
added a warning that `DistributedConfig` is experimental and may change!

PR #42410, opened 2025-11-25
Fix tf32 api deprecation for Pytorch version

### LINKED ISSUE #42371: Please use the new API settings to control TF32 behavior, ...
### System Info

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

```
```

### Expected behavior

khushali9:
What does this PR do?

The ask is to use fp32_precision instead of allow_tf32 for 
Pytorch version >= 2.9.0
as pointed out in this [doc](https://docs.pytorch.org/docs/main/notes/cuda.html#tensorfloat-32-tf32-on-ampere-and-later-devices) mentioned in the [issue](https://github.com/huggingface/transformers/issues/42371) 
 
I have also added test cases 

Fixes #42371 ([issue](https://github.com/huggingface/transformers/issues/42371))

Can you review @Rocketknight1  @ArthurZucker 

## Before submitting
      Pull Request section?
      to it if that's the case.


tests/utils/test_import_utils.py
```diff
@@ -24,3 +29,32 @@ def test_clear_import_cache():
...
+        ("2.8.1", False, False),
+        ("2.9.0", True, "tf32"),
+    ],
+)
+def test_set_tf32_mode(torch_version, enable, expected, caplog):
+    caplog.set_level(logging.INFO)
+    # Use the full module path for patch
+    with patch("transformers.utils.import_utils.get_torch_version", return_value=torch_version):
+        mock_torch = MagicMock()
+        with patch.dict("transformers.utils.import_utils.__dict__", {"torch": mock_torch}):
```
khushali9, line 50: @Rocketknight1 testcase passes with import torch at module level in import_utils file where _set_tf32_mode is defined, but that is not accepted so moved it back inside my new method. but that broke this line so fixed it aswell. But tests are having issue with mock. Keeping it to skip , may be can be worked on later , or if someone can help now that will be great.

src/transformers/utils/import_utils.py
```diff
@@ -58,7 +58,7 @@ def _is_package_available(pkg_name: str, return_version: bool = False) -> tuple[
             # pick the first item of the list as best guess (it's almost always a list of length 1 anyway)
             distribution_name = pkg_name if pkg_name in distributions else distributions[0]
             package_version = importlib.metadata.version(distribution_name)
-        except (importlib.metadata.PackageNotFoundError, KeyError):
+        except importlib.metadata.PackageNotFoundError:
```
khushali9, line 61: changes due to running make style or main merge. How can I remove unrelated code.

src/transformers/utils/import_utils.py
```diff
@@ -502,13 +501,14 @@ def is_torch_tf32_available() -> bool:
         return False
     return True
 
-
 @lru_cache
 def _set_tf32_mode(enable: bool) -> None:
     """
     Set TF32 mode using the appropriate PyTorch API.
+
```
khushali9: main method that will help us set correct API use

src/transformers/utils/import_utils.py
Rocketknight1: Hmn, we probably don't want these changes in the PR! It makes it hard to review and I'm a bit worried that it'll actually revert some code. Can you try getting rid of them, maybe with one of the following:

1) Rebase/merge onto the latest `main` commit
2) `pip install -e .[quality]` to get the latest style tools
3) Compare the edited files against the equivalent version in `main` and revert any of these unrelated changes?

src/transformers/utils/import_utils.py
```diff
@@ -503,6 +502,27 @@ def is_torch_tf32_available() -> bool:
     return True
 
 
+@lru_cache
+def _set_tf32_mode(enable: bool) -> None:
```
Rocketknight1, line 506: I think it's fine to use a non-private name for the function, like `enable_tf32` or even `torch_enable_tensorfloat32` for clarity.

tests/utils/test_import_utils.py
Rocketknight1: I think you can just remove this test! The only real way to fully test the function would be to have multiple versions of `torch` in the CI, which is quite hard. So in reality, we'll only be testing the installed version of `torch` anyway, and if the function is failing for that version then we'll see errors elsewhere anyway.

Rocketknight1 (commented):
Patch mostly looks good! I added some comments about the extra changes - those can be fiddly, and sometimes result from messy merges or rebases. If you can't get rid of them easily you may want to just start a new PR by branching from the latest `main` commit and tagging me there instead.

khushali9:
Sure I will create new PR, and remove tests, and update method name. 
Thanks for your help. 
Also I have torch > 2.9 and one of the only test passed for it. That makes sense.

khushali9:
created new PR as this got messed up. thanks

Rocketknight1:
No problem!

PR #42412, opened 2025-11-26
Replace Optional and Union typing with | in examples

cyyever:
# What does this PR do?

This PR uses python 3.10 syntax for Optional and Union typing in examples.


Rocketknight1:
Looks good but pausing this merge until after the weekend while we prep for the v5 launch!

PR #40546, opened 2025-08-29
Implement VibeVoice 

pengzhiliang:
[CPU CI](https://transformers-ci.lor-e.huggingface.cool/d/pytest-observability-by-pr/pytest-observability-branch?var-pr=40546) [GPU run-slow](https://transformers-ci.lor-e.huggingface.cool/d/pytest-observability-by-pr/pytest-observability-branch?var-pr=40546)

# What does this PR do?

Merge the model from https://github.com/microsoft/VibeVoice/tree/main

HF: 
https://huggingface.co/microsoft/VibeVoice-1.5B

PR #38405, opened 2025-05-27
Add Dia model

buttercrab:
# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #36719, opened 2025-03-14
Add CSM model

eustlb:
# What does this PR do?

Adds CSM (audio)-text-to-speech model! 
[Original code](https://github.com/SesameAILabs/csm)
[Hub model weights](https://huggingface.co/sesame/csm-1b)
[Converted weights](https://huggingface.co/eustlb/csm-1b)

PR #42413, opened 2025-11-26
Add chatterbox support

manmay-nakhashi:


Rocketknight1:
This PR is very overwhelming, like a lot of code agent PRs. It's not clear to me how novel some of these architectures are, or whether we really need three whole new architectures in the codebase! :sweat_smile:

Ideally we could cut down the PR size a lot by using `modular` files and importing, and maybe treating some of the submodels as components or reusing existing architectures for those?

cc @ebezzam @eustlb since it's an audio model/pipeline

manmay-nakhashi:
@Rocketknight1 t3 model is very similar to llama model but because this is a whole Text-to-Speech pipeline we have tokenizer(s3tokenizer different pr which compresses the audio to 25 tok/ sec we need this for conditioning), encoder(T3- mod-llama model (main changes are speech tokens and conditioning )), decoder(s3gen cfm based model from cozyvoice2) and hifinet(from cozyvoice2) which converts mel to wav. 
i checked we don't have this models in the hf currently

manmay-nakhashi:
@eustlb @Rocketknight1 What is the best way to add BasicTransformerBlock in s3gen ?, i am currently importing from diffusers but it's not there in the transformers , can i add that as a dependency ?

ebezzam:
@manmay-nakhashi thanks for your contributions 🤗 let's avoid importing from `diffusers`, especially since Transformer blocks are already implemented within `transformers`

I recommend taking a look at [CONTRIBUTING.md](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#do-you-want-to-implement-a-new-model), and in particular [this guide](https://huggingface.co/docs/transformers/en/modular_transformers) on adding new models with scripts called `modular_MODELNAME.py`.

It is very likely your models contain components (like Transformer blocks) that already exist in the library!

manmay-nakhashi:
@eustlb

src/transformers/models/s3tokenizer/modeling_s3tokenizer.py
```diff
@@ -0,0 +1,841 @@
...
+
+
+class Conv1d(torch.nn.Conv1d):
+    """Conv1d layer that preserves dtype."""
+
+    def _conv_forward(self, x: torch.Tensor, weight: torch.Tensor, bias: Optional[torch.Tensor]) -> torch.Tensor:
+        return super()._conv_forward(x, weight.to(x.dtype), None if bias is None else bias.to(x.dtype))
+
+
+class MultiHeadAttention(torch.nn.Module):
```
ebezzam: MHA already exists in Transformers. It should be reused, namely [from Llama](https://github.com/huggingface/transformers/blob/fccb04998af44166524c2a7cfe0d946b99aef497/src/transformers/models/llama/modeling_llama.py#L227) via modular.

Here is an example of using modular to reuse an existing component and overwrite _only_ what needs to be adapted: https://huggingface.co/docs/transformers/en/modular_transformers#attention

Moreover, to use the existing component from Transformers, you will have to adapt the state dict with a  `convert_<model_name>_to_hf.py` script that converts the original model weights to the HuggingFace format, as discussed in step 3 [here](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#vision-language-model-contribution-checklist).

Here is an example for another TTS model of such a script: https://github.com/huggingface/transformers/blob/main/src/transformers/models/dia/convert_dia_to_hf.py

src/transformers/models/s3tokenizer/modeling_s3tokenizer.py
```diff
@@ -0,0 +1,841 @@
...
+            self.weight.to(x.dtype),
+            None if self.bias is None else self.bias.to(x.dtype),
+        )
+
+
+class Conv1d(torch.nn.Conv1d):
+    """Conv1d layer that preserves dtype."""
+
+    def _conv_forward(self, x: torch.Tensor, weight: torch.Tensor, bias: Optional[torch.Tensor]) -> torch.Tensor:
+        return super()._conv_forward(x, weight.to(x.dtype), None if bias is None else bias.to(x.dtype))
```
ebezzam: Transformers convention doesn't wrap modules like this. 

torch.nn.LayerNorm/Linear/Conv1d should be directly used

src/transformers/models/s3tokenizer/modeling_s3tokenizer.py
```diff
@@ -0,0 +1,841 @@
...
+        processed_wavs = []
+        for wav in wavs:
+            if isinstance(wav, np.ndarray):
+                wav = torch.from_numpy(wav)
+            if wav.dim() == 1:
+                wav = wav.unsqueeze(0)
+            processed_wavs.append(wav)
+        return processed_wavs
+
+    def log_mel_spectrogram(self, audio: torch.Tensor, padding: int = 0) -> torch.Tensor:
```
ebezzam: Everything related to preparing the input audio before the model (dimension shape, spectrogram, etc) should be in the feature extractor

src/transformers/models/s3gen/modeling_s3gen.py
```diff
@@ -0,0 +1,2687 @@
...
+    batch_size = lengths.size(0)
+    max_len = max_len if max_len > 0 else lengths.max().item()
+    seq_range = torch.arange(0, max_len, dtype=torch.int64, device=lengths.device)
+    seq_range_expand = seq_range.unsqueeze(0).expand(batch_size, max_len)
+    seq_length_expand = lengths.unsqueeze(-1)
+    mask = seq_range_expand >= seq_length_expand
+    return mask
+
+
+def mel_spectrogram(
```
ebezzam, line 84: Everything related to preparing the input audio before the model (dimension shape, spectrogram, etc) should be in the feature extractor

src/transformers/models/s3gen/modeling_s3gen.py
```diff
@@ -0,0 +1,2687 @@
...
+    expansion = 1
+
+    def __init__(self, in_planes, planes, stride=1):
+        super().__init__()
+        self.conv1 = nn.Conv2d(in_planes, planes, kernel_size=3, stride=(stride, 1), padding=1, bias=False)
+        self.bn1 = nn.BatchNorm2d(planes)
+        self.conv2 = nn.Conv2d(planes, planes, kernel_size=3, stride=1, padding=1, bias=False)
+        self.bn2 = nn.BatchNorm2d(planes)
+
+        self.shortcut = nn.Sequential()
```
ebezzam: Another Transformers convention is to not use `nn.Sequential` but rather [ModuleList](https://docs.pytorch.org/docs/stable/generated/torch.nn.ModuleList.html). Can you change all of those?

src/transformers/models/s3gen/modeling_s3gen.py
```diff
@@ -0,0 +1,2687 @@
...
+            nonlinear.add_module("relu", nn.ReLU(inplace=True))
+        elif name == "prelu":
+            nonlinear.add_module("prelu", nn.PReLU(channels))
+        elif name == "batchnorm":
+            nonlinear.add_module("batchnorm", nn.BatchNorm1d(channels))
+        elif name == "batchnorm_":
+            nonlinear.add_module("batchnorm", nn.BatchNorm1d(channels, affine=False))
+        else:
+            raise ValueError(f"Unexpected module ({name}).")
+    return nonlinear
```
ebezzam, line 200: Are all of these non-linearities used in the final model? When merging new models into Transformers, all code paths that don't contribute to the final model are expected to be removed

src/transformers/models/s3gen/modeling_s3gen.py
```diff
@@ -0,0 +1,2687 @@
...
+    Parameters:
+        - alpha: trainable parameter
+
+    References:
+        - This activation function is from this paper by Liu Ziyin, Tilman Hartwig, Masahito Ueda:
+        https://arxiv.org/abs/2006.08195
+    """
+
+    def __init__(
+        self, in_features: int, alpha: float = 1.0, alpha_trainable: bool = True, alpha_logscale: bool = False
```
ebezzam: Is `alpha_logscale` set to `True` at any point?

Because `Snake1d` could be used [from DAC](https://github.com/huggingface/transformers/blob/5ee9ffe386c5ecc77d8009ab648b8c4c109931ea/src/transformers/models/dac/modeling_dac.py#L87) via modular

src/transformers/models/s3gen/modeling_s3gen.py
```diff
@@ -0,0 +1,2687 @@
...
+        Snake ∶= x + 1/a * sin^2 (xa)
+        """
+        alpha = self.alpha.unsqueeze(0).unsqueeze(-1)  # line up with x to [B, C, T]
+        if self.alpha_logscale:
+            alpha = torch.exp(alpha)
+        x = x + (1.0 / (alpha + self.no_div_by_zero)) * pow(sin(x * alpha), 2)
+        return x
+
+
+class ResBlock(nn.Module):
```
ebezzam, line 422: From a quick glance, there seems to be similarities to [DacEncoderBlock](https://github.com/huggingface/transformers/blob/5ee9ffe386c5ecc77d8009ab648b8c4c109931ea/src/transformers/models/dac/modeling_dac.py#L212).

However, if that component cannot be used directly via modular, how about [DacResidualUnit](https://github.com/huggingface/transformers/blob/5ee9ffe386c5ecc77d8009ab648b8c4c109931ea/src/transformers/models/dac/modeling_dac.py#L175C7-L175C22)? As you also seem to use three such modules

src/transformers/models/s3gen/modeling_s3gen.py
```diff
@@ -0,0 +1,2687 @@
...
+        kernel_size: int = 3,
+        dilations: tuple[int] = (1, 3, 5),
+    ):
+        super().__init__()
+        self.convs1 = nn.ModuleList()
+        self.convs2 = nn.ModuleList()
+
+        for dilation in dilations:
+            self.convs1.append(
+                weight_norm(
```
ebezzam: The Transformers version shouldn't have weight norm. But for checkpoints that have weight norm:
- weight norm can be added to all relevant layers with [a single method](https://github.com/huggingface/transformers/blob/5ee9ffe386c5ecc77d8009ab648b8c4c109931ea/src/transformers/models/dac/modeling_dac.py#L493)
- weights are loaded
- weight norm can be removed from all relevant layers [with a single method](https://github.com/huggingface/transformers/blob/5ee9ffe386c5ecc77d8009ab648b8c4c109931ea/src/transformers/models/dac/modeling_dac.py#L526)

For example, within the DAC conversion script: https://github.com/huggingface/transformers/blob/5ee9ffe386c5ecc77d8009ab648b8c4c109931ea/src/transformers/models/dac/convert_dac_checkpoint.py#L252-L254

src/transformers/models/s3gen/modeling_s3gen.py
```diff
@@ -0,0 +1,2687 @@
...
+            nn.ELU(),
+            weight_norm(nn.Conv1d(cond_channels, cond_channels, kernel_size=3, padding=1)),
+            nn.ELU(),
+            weight_norm(nn.Conv1d(cond_channels, cond_channels, kernel_size=3, padding=1)),
+            nn.ELU(),
+            weight_norm(nn.Conv1d(cond_channels, cond_channels, kernel_size=3, padding=1)),
+            nn.ELU(),
+            weight_norm(nn.Conv1d(cond_channels, cond_channels, kernel_size=3, padding=1)),
+            nn.ELU(),
+        )
```
ebezzam: Similar comments as before: removing Sequential and weight norm

tests/models/s3tokenizer/test_modeling_s3tokenizer.py
ebezzam, line 1: The modeling tests should contain integration tests to compare the outputs of Transformers version with original model. For example:
- DAC: https://github.com/huggingface/transformers/blob/5ee9ffe386c5ecc77d8009ab648b8c4c109931ea/tests/models/dac/test_modeling_dac.py#L305
- AudioFlamingo3 (more recent): https://github.com/huggingface/transformers/blob/5ee9ffe386c5ecc77d8009ab648b8c4c109931ea/tests/models/audioflamingo3/test_modeling_audioflamingo3.py#L252

src/transformers/models/s3tokenizer/modeling_s3tokenizer.py
```diff
@@ -0,0 +1,841 @@
...
+            import librosa
+
+            _mel_filters = librosa.filters.mel(sr=config.sampling_rate, n_fft=self.n_fft, n_mels=config.n_mels)
+            self.register_buffer("_mel_filters", torch.FloatTensor(_mel_filters))
+        except ImportError:
+            logger.warning(
+                "librosa is not installed. Mel filters will not be initialized. "
+                "Install librosa with: pip install librosa"
+            )
+            self.register_buffer("_mel_filters", torch.zeros(config.n_mels, self.n_fft // 2 + 1))
```
ebezzam: We should avoid try/except. Moreover, mel filters is something for the feature extractor

ebezzam (commented):
@manmay-nakhashi I've done an initial review of some things stuck out to me, and to help you familiarize with different Transformers conventions.

Most notably, there are several modules that already exist within the Transformers library from other models, and those should be used via [modular](https://huggingface.co/docs/transformers/en/modular_transformers) to create your modeling files.

Moreover, here are some PRs of other TTS models that may also help you see how to prepare the various files:
- CSM: https://github.com/huggingface/transformers/pull/36719
- Dia: https://github.com/huggingface/transformers/pull/38405
- VibeVoice (ongoing but reflects more recent conventions): https://github.com/huggingface/transformers/pull/40546

PR #42416, opened 2025-11-26
GLM4V - delete duplicate config attribute

zucchini-nlp:
# What does this PR do?

As per title, it was accidentally duplicated in the RoPE refactor PR. We don't access image-id from text config, so lets' delete to not confuse users

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


zucchini-nlp:
@bot /style

molbap (approved):
Thanks for fixing! Indeed, should not be allowed in text config. We tried to normalize attributes but it's hard, however maybe an idea would be to forbid some named attributes to be on text or image? 
LGTM anyway

zucchini-nlp:
Yeah, would be great. We could potentially force users to not have the same key in subconfig and in general config, though I am quite sure there are a few exceptions to it. Especially with omni-modal models, i remember that we had to add an `image_token_id` in several configs

I think it's fine as long as we review well. This one specifically is my bad, too many diffs and a huge refactor, so I might have missed smth

PR #42420, opened 2025-11-26
Let's break Qwen-VL 🚨  

zucchini-nlp:
# What does this PR do?

As per title, break the cycle. Qwen will no longer have access to text attributes with `config.vocab_size` which solves the duplicated attribute issue we had in TRL. A config should not have two different values for the same key

Tested with the TRL issue that there is no regression and ran slow test locally


zucchini-nlp:
run-slow: qwen2_5_vl, qwen2_vl

zucchini-nlp:
run-slow: qwen2_5_vl, qwen2_vl

zucchini-nlp:
run-slow: qwen2_5_vl, qwen2_vl

zucchini-nlp:
run-slow: qwen2_5_vl, qwen2_vl

zucchini-nlp:
run-slow: qwen2_5_vl, qwen2_vl

zucchini-nlp:
run-slow: qwen2_5_vl, qwen2_vl

src/transformers/models/qwen2_5_vl/configuration_qwen2_5_vl.py
```diff
@@ -290,39 +302,21 @@ def __init__(
         if isinstance(text_config, dict):
             self.text_config = self.sub_configs["text_config"](**text_config)
         elif text_config is None:
-            # For BC use all kwargs to init `TextConfig`
-            self.text_config = self.sub_configs["text_config"](**kwargs)
+            # Hub configs are saved as flat dicts so we pop some of kwargs to init `TextConfig`
+            text_params = inspect.signature(self.sub_configs["text_config"].__init__).parameters.keys()
+            text_params = list(text_params) + ["rope_scaling", "rope_theta"]
+            text_config = {key: kwargs.pop(key) for key in text_params if key in kwargs}
+            text_config["dtype"] = kwargs.get("torch_dtype", kwargs.get("dtype"))  # don't pop the dtype
```
zucchini-nlp, line 309: not the best way, i should admit. The problem is that config will assign all kwargs as attributes, and if we don't pop we end up with the same set of kwargs in text config and in the general config

albertvillanova:
Thanks for addressing this issue, @zucchini-nlp. :hugs: 

For completeness, here is the original issue we opened in TRL that tracked this problem:
- https://github.com/huggingface/transformers/issues/41020
  - More concretely, this comment: https://github.com/huggingface/transformers/issues/41020#issuecomment-3330098710

ArthurZucker (approved):
Thanks let's add a note in the migration guide please!

PR #42424, opened 2025-11-26
[WIP] attempt to fix ooms in tests

molbap:
# What does this PR do?

 as per title, focused on qwen series


molbap:
run-slow: qwen3_omni_moe, qwen3_vl, qwen3_vl_moe

ArthurZucker (commented):
Let's find the real root cause that's weird but ty for diving a bit!

PR #42426, opened 2025-11-26
Clean bnb integration using weight converter

SunMarc:
# What does this PR do?

This PR cleans bnb integration, we don't need to deal with missing keys.

### DIFF
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -721,7 +721,11 @@ def convert_and_load_state_dict_in_model(
                 source_pattern = renamed_key
 
             # 5. Handle dtype casting
-            if hf_quantizer and not hf_quantizer.pre_quantized and hf_quantizer.param_needs_quantization(model, renamed_key):
+            if (
+                hf_quantizer
+                and not hf_quantizer.pre_quantized
+                and hf_quantizer.param_needs_quantization(model, renamed_key)
+            ):
                 mapping.quantization_operation = hf_quantizer.get_quantize_ops()
 
             _dtype = dtype
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -674,9 +674,6 @@ def convert_and_load_state_dict_in_model(
 
     renamings = [entry for entry in weight_mapping if isinstance(entry, WeightRenaming)]
     converters = [entry for entry in weight_mapping if isinstance(entry, WeightConverter)]
-    if hf_quantizer:
-        # We will add the quantizer's deserialization WeightConverter here.
-        pass
 
     param_name_to_load: dict[str, Union[WeightRenaming | WeightConverter]] = {}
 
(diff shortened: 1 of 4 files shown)


src/transformers/core_model_loading.py
```diff
@@ -724,11 +721,15 @@ def convert_and_load_state_dict_in_model(
...
-            if hf_quantizer is not None and hf_quantizer.param_needs_quantization(model, renamed_key):
+            if hf_quantizer and not hf_quantizer.pre_quantized and hf_quantizer.param_needs_quantization(model, renamed_key):
                 mapping.quantization_operation = hf_quantizer.get_quantize_ops()
 
             _dtype = dtype
-            if dtype_plan != {} and dtype_policy_alt.search(renamed_key):
+            if hf_quantizer and hf_quantizer.pre_quantized and original_key != renamed_key:
+                # if the key was renamed as it is not available in the state dict otherwise, it means that we are deserializing it,
+                # so we need to make sure to load the tensor with the same dtype from the checkpoint
+                _dtype = None
```
SunMarc, line 735: those keys were collected but do not appear in the `state_dict` of the model, so I think we should expect it to keep the same dtype as the checkpoint.

src/transformers/integrations/bitsandbytes.py
```diff
@@ -47,45 +46,54 @@ def convert(
...
-                return {target_key: new_value}
-            return {}
+        # Support models using `Conv1D` in place of `nn.Linear` (e.g. openai-community/gpt2) by transposing the weight matrix prior to quantization.
+        # Since weights are saved in the correct "orientation", we skip transposing when loading.
+        if issubclass(module.source_cls, Conv1D):
+            value = value.T
+
+        old_value = model.get_parameter_or_buffer(target_key)
+        new_value = bnb.nn.Params4bit(value, requires_grad=False, **old_value.__dict__).to(value.device)
+        module._is_hf_initialized = True
```
MekkCyber, line 59: I don't think we need to set this here since we are returning the tensors

MekkCyber (approved):
Thanks ! looks good to me

src/transformers/core_model_loading.py
MekkCyber: why they don't appear in the state_dict ?

SunMarc: well those key are the one we are trying to collect such as weight.absmax or weight.quant_map to recreate weight. After renaming, those becomes weight, so it would upcast them in the same dtype as the weight.

src/transformers/integrations/bitsandbytes.py
SunMarc: just to be careful, i'm not sure how `_is_hf_initialized` works but i'm sure it will try to initialize somehow

SunMarc:
@bot /style

SunMarc:
@bot /style

SunMarc:
run-slow: bnb

PR #42428, opened 2025-11-26
logic to select tf32 API as per Pytorch version

### LINKED ISSUE #42371: Please use the new API settings to control TF32 behavior, ...

khushali9:
What does this PR do?
The ask is to use fp32_precision instead of allow_tf32 for
Pytorch version >= 2.9.0 for CUDA
as pointed out in this [doc](https://docs.pytorch.org/docs/main/notes/cuda.html#tensorfloat-32-tf32-on-ampere-and-later-devices) mentioned in the https://github.com/huggingface/transformers/issues/42371

I have also added test cases

Fixes https://github.com/huggingface/transformers/issues/42371 (https://github.com/huggingface/transformers/issues/42371)

Can you review @Rocketknight1 @ArthurZucker

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


khushali9:
@Rocketknight1 here is the fresh clean PR, thanks for helping.

wasertech:
~~As mentioned in https://github.com/huggingface/transformers/issues/42371#issuecomment-3575704598, you probably want to add a CI test. There should already be a simple test in the .github/workflow but that is using the previous version of pytorch, it should be easy to duplicate it and update it to make the new test, then we shall see.~~ nvm I couldn't see the tests result from the app the but did it alright 😄👍 conftest is the way to go! Thanks for being so quick.

khushali9:
> ~As mentioned in [#42371 (comment)](https://github.com/huggingface/transformers/issues/42371#issuecomment-3575704598), you probably want to add a CI test. There should already be a simple test in the .github/workflow but that is using the previous version of pytorch, it should be easy to duplicate it and update it to make the new test, then we shall see.~ nvm I couldn't see the tests result from the app the but did it alright 😄👍 conftest is the way to go! Thanks for being so quick.

Thanks for checking, yes I already had tests but they failed on CI cause of pytorch version, but I did test for both version so we are good to go with this change.

Rocketknight1 (approved):
Yes, this looks good to me now! cc core maintainers @Cyrilvallez @ArthurZucker just in case, though, since this is touching a lot of core files

tl;dr for reviewers the API in PyTorch for enabling TensorFloat32 changed in `2.9`, and using the old API now triggers a `UserWarning`. This patch makes a helper function `enable_tf32()` that calls either the old or new API depending on the Torch version, so it works on all versions without warnings.

PyTorch docs here: https://docs.pytorch.org/docs/2.9/notes/cuda.html#tensorfloat-32-tf32-on-ampere-and-later-devices

src/transformers/utils/import_utils.py
```diff
@@ -503,6 +502,27 @@ def is_torch_tf32_available() -> bool:
...
+    For older versions, uses the legacy allow_tf32 flags.
+    Args:
+        enable: Whether to enable TF32 mode
+    """
+    import torch
+
+    pytorch_version = version.parse(get_torch_version())
+    if pytorch_version >= version.parse("2.9.0"):
+        precision_mode = "tf32" if enable else "ieee"
+        torch.backends.cuda.matmul.fp32_precision = precision_mode
```
Rocketknight1: One thing I just noticed in the docs - can we also set `torch.backends.fp32_precision = precision_mode`? The docs are a bit unclear but I think that's the "global" setting.

khushali9: @Rocketknight1 can you point me how I can see my PR changes in the document link. If you are talking about pytorch doc then I saw it.

For torch.backends.fp32_precision=precision_mode

I can make change but I do not see anywhere in  code  torch.backends.allow_32 being used. But I did check pytorch doc and yes torch.backends.fp32_precision is global setting. I can update PR with that. 

One more question this issue is for CUDA , but in our code I do see musa , like this 
torch.backends.mudnn.allow_tf32 , which my code does not touch. Do you want me to update that API as well ? I was not sure about that as in example they only talked about cuda and not musa. But I feel I should make that changes too. let me know what you think.

khushali9:
@Rocketknight1 Can you review it again, addressed changes related to setting global setting and also handling MUSA devices.

src/transformers/utils/import_utils.py
Rocketknight1: Hi @khushali9, I **think** what's happening is that `torch.backends.fp32_precision` is the "master" setting, which sets everything else by default. The lower-level settings like `torch.backends.cuda.matmul.fp32_precision` are only necessary if we want to override the master setting.

Therefore, I think `torch.backends.fp32_precision` is what we should do in future, and hopefully we won't need specific code for `musa` after that. However, the PyTorch documentation is a bit unclear about this. I agree that it's a bit different from our old API, but I think this is what we want the function to do!

khushali9: Yeah I agree @Rocketknight1 , thats why as soon as we find out it is > 2.9.0 I set global setting, but to take care of our code lines I also added those musa and cuda related changes in the block. 

Shall I remove and just keep global setting in >2.9.0 block ? or is this PR good to go ?

Rocketknight1:
@khushali9 I think we can just remove everything else and keep the global `torch.backends.fp32_precision` line in the `torch >= 2.9.0` path. However, it'd be cool if you or someone else could test that that actually behaves the way we want it to!

khushali9:
> @khushali9 I think we can just remove everything else and keep the global `torch.backends.fp32_precision` line in the `torch >= 2.9.0` path. However, it'd be cool if you or someone else could test that that actually behaves the way we want it to!

Sure @Rocketknight1 I will remove those things, and test again.

khushali9:
@Rocketknight1 It is ready with new changes and I tested with 2.10.0, 2.9.0, 2.8.0 with enable equals True and False options.

ArthurZucker (approved):
LGTM thanks 🤗

src/transformers/utils/import_utils.py
```diff
@@ -503,6 +502,29 @@ def is_torch_tf32_available() -> bool:
...
+    For older versions, uses the legacy allow_tf32 flags.
+    Args:
+        enable: Whether to enable TF32 mode
+    """
+    import torch
+
+    pytorch_version = version.parse(get_torch_version())
+    if pytorch_version >= version.parse("2.9.0"):
+        precision_mode = "tf32" if enable else "ieee"
+        torch.backends.fp32_precision = precision_mode
```
khushali9, line 519: @Rocketknight1 here we used global instead of specifics like for an example this
torch.backends.cudnn.conv.fp32_precision = "tf32" , regarding my comment on the issue for which this PR is related.

khushali9: nvm! it is pytorch issue and not transformer.

conftest.py
```diff
@@ -137,11 +138,9 @@ def check_output(self, want, got, optionflags):
 doctest.DocTestParser = HfDocTestParser
 
 if is_torch_available():
-    import torch
-
     # The flag below controls whether to allow TF32 on cuDNN. This flag defaults to True.
     # We set it to `False` for CI. See https://github.com/pytorch/pytorch/issues/157274#issuecomment-3090791615
-    torch.backends.cudnn.allow_tf32 = False
+    enable_tf32(False)
```
ydshieh, line 143: Hi @khushali9 This doesn't seem to enough for our CI. Some tests are failing with a slight larger difference beyond our usual tolerance `1e-5` since this PR. `torch.backends.fp32_precision` is correctly set to `ieee`, and `torch.backends.cudnn.allow_tf32` remains to be `True`.

Either we keep also `torch.backends.cudnn.allow_tf32 = False` here, or we add it to `enable_tf32`. I am not 100% sure the 2nd approach reflects accurately what behavior we want `enable_tf32` to have. WDYK?

ydshieh: Sorry, just saw the original issue mentioning it's deprecated method.

Anyway, it turns out we need to set

torch.backends.cudnn.conv.fp32_precision = "ieee"

for our CI. It's not clear to me why 

torch.backends.fp32_precision = "ieee"

isn't enough.

ydshieh: ok, here is what we have so far

```
(Pdb) torch.backends.fp32_precision
'ieee'
(Pdb) torch.backends.cudnn.conv.fp32_precision
'tf32'
(Pdb) torch.backends.cudnn.rnn.fp32_precision
'tf32'
(Pdb) torch.backends.cudnn.fp32_precision
'ieee'
(Pdb) torch.backends.cuda.matmul.fp32_precision
'ieee'
(Pdb) torch.backends.fp32_precision
'ieee
```

khushali9: Yeah I agree, I. made my early PR commit with torch.backends.cudnn.conv.fp32_precision but from here we understood https://docs.pytorch.org/docs/main/notes/cuda.html#tensorfloat-32-tf32-on-ampere-and-later-devices torch.backends.fp32_precision = "ieee" would be enough so I changed it as per that but yes if now it is not taking it implicitly , we should change it to explicitly.

PR #42431, opened 2025-11-26
rewrite _process_parameter_type in auto_docstring.py to improve usability

Yacklin:
# What does this PR do?
When using kimi model, kimi developers used something like "bool|int" as type hint for parameters in function signature and unfortunately, due to python's nature, format of type hint like "bool|int" is not completely the same as "typing.Union[bool, int]" until python 3.14. One of the biggest differences is when accessing param.annotation.\_\_name\_\_, "bool|int" would raise error that has no attribute "\_\_name\_\_". This error bothers me a lot so i have to fix it. I found a beautiful way to fix it and its backward compatibility could be extended to 3.4 (I just tried it a little bit but you could try versions before 3.4).

Please be advised that param_name and func are unused in this rewrited version. I chose not to remove these two parameters just in case something bad would happen.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@Rocketknight1
(I accidentally deleted the previous PR, here is the new one)


Rocketknight1:
Approved the doc build run! You should get a link to the docs with your PR in about 30 minutes - can you check that everything is working as expected?

Yacklin:
> Approved the doc build run! You should get a link to the docs with your PR in about 30 minutes - can you check that everything is working as expected?

i think yes. Could you please let me know if anything should be changed in my code for this PR? I tested this rewritten fuction several times as well. Also, i am a little bit new to git and github so sometimes i might mess up PR or commits. Sorry for that in advance. @Rocketknight1

Yacklin:
i am wondering why sometimes tests failed but sometimes didn't? And my changes have nothing to do with these failures

Yacklin:
> Approved the doc build run! You should get a link to the docs with your PR in about 30 minutes - can you check that everything is working as expected?

Okay i do find something changed, which i would prefer actually.
current one:
image
updated one:
image

Might be a little bit verbose but i think it's okay

Yacklin:
Maybe i should provide more details about this PR.
When using kimi 48B instruct model, the decorator \@auto_docstring would be applied to some functions in modelling_kimi like this:
image

But, auto_docstring would call _process_parameter_type to go through every single parameter's type hint to call this type hint's annotation.\_\_name\_\_. But before python 3.14, bool|int is not the same as Union[bool, int]. It means type hint like "bool|hint" has no attribute \_\_name\_\_. Then _process_parameter_type would raise errors, which prevents users from using kimi 48B instruct models.
image

Btw, the downstream users of kimi 48B noticed this problem then they changed it but it seems that they didn't notify kimi this problem:

image
(coming from modelling_kimi.py of cerebras/Kimi-Linear-REAP-35B-A3B-Instruct)

My PR could solve this problem. Otherwise, python 3.14 is needed to use kimi 48B instruct model and i don't think transformers users could switch to python 3.14 easily
@Rocketknight1

Rocketknight1 (commented):
Hey @Yacklin, I took a look at the code! I think the bug you're fixing is real, and I think the cause is that we do lots of hacky accesses to things like `annotation.__name__`, like you said. I don't like `if ":" in parameter_str` though - doing these kinds of string manipulations also feels very hacky. It's likely to create bugs in future if anything changes with typing, or if anyone writes an unusual type hint.

Is there a clean way to correctly resolve all type hints into some kind of canonical form, and then just operate on that, rather than relying on string representations and attribute accesses?

Yacklin:
> Hey @Yacklin, I took a look at the code! I think the bug you're fixing is real, and I think the cause is that we do lots of hacky accesses to things like `annotation.__name__`, like you said. I don't like `if ":" in parameter_str` though - doing these kinds of string manipulations also feels very hacky. It's likely to create bugs in future if anything changes with typing, or if anyone writes an unusual type hint.
> 
> Is there a clean way to correctly resolve all type hints into some kind of canonical form, and then just operate on that, rather than relying on string representations and attribute accesses?

Hi @Rocketknight1 , string manipulation is unavoidable in this function because this function would return something like tuple[str, bool]. But i could try to write an implmentation to minimize string manipulation.

Rocketknight1:
Hi @Yacklin, some manipulation is unavoidable at the end, but I think you should always be able to get the type hint from `param.annotation` rather than needing to use `str(param)`. What we really want to do is just grab that annotation in the safest and most Pythonic way possible and turn it into a "canonical" form regardless of whether it was written as `Optional[Union[str, bool]]` or `str | bool | None`. I don't think flattening it to a string is the safest way to do that, and will probably get very messy for nested types!

Yacklin:
> Hi @Yacklin, some manipulation is unavoidable at the end, but I think you should always be able to get the type hint from `param.annotation` rather than needing to use `str(param)`. What we really want to do is just grab that annotation in the safest and most Pythonic way possible and turn it into a "canonical" form regardless of whether it was written as `Optional[Union[str, bool]]` or `str | bool | None`. I don't think flattening it to a string is the safest way to do that, and will probably get very messy for nested types!

Hi @Rocketknight1 ,

This situation is more complex than i thought. Please read my comments and implementations carefully. Let me know what you think.

```py

import inspect
import types
import typing
from typing import get_origin, Union, get_args, Optional

def _process_parameter_type(param, param_name, func):
    """
    Process and format a parameter's type annotation.

    Args:
        param (`inspect.Parameter`): The parameter from the function signature
        param_name (`str`): The name of the parameter
        func (`function`): The function the parameter belongs to
    """
    """
    the minimum version of python that transformers supports is 3.9.0.
    but this is the source of problem.

    if user who is using python 3.9.x tried to access models that are using features coming from types.UnionType, like int|float, which is introduced in python 3.10,
    error would be raised (even before this function is executed). Python developers came up with an idea that would squash int|float into string "int|float", using 'from __future__ import annotations'.
    But this is something models developers should care, transformers developers could only warn models developers to solve this issue in their own model script.

    the best way to solve this issue is to level up the minimum python version from 3.9 to 3.10

    if user is using python 3.9, this error would be raised before this function is hit.

    unfortunately, most models developers didn't pay attention to it.

    If python 3.9 users tried to access models that are using features coming from types.UnionType, like int|float, which is introduced in python 3.10,
    users have to contact models developers to support backward-compatibility. Transformers team has nothing to do with it, unless upgrading min version from
    3.9 to 3.10, through voting or something.
    """

    # this PR only takes python 3.10 into consideration, unfortunately.

    param_annotation = param.annotation
    print(param_annotation)
    default_value = param.default
    origin = get_origin(param_annotation)
    args = get_args(param_annotation)
    optional_flag = False
    type_hint_str = ""

    # Optional[sth], under the hood is equivalent to Union[sth, None]
    # if the type hint is None, I mean something like age:None, it would be considered optional, very few developers would write it though.
    # if the type hint is Optional[None] or Union[None], it would be considered optional, very few developers would write it though.
    if (
        (origin is typing.Union and type(None) in args)
        or default_value is not inspect._empty
        or (origin is types.UnionType and type(None) in args)
        or param_annotation is types.NoneType
        or param_annotation is None
    ):
        optional_flag = True

    """
    okay here is another problem, to get the string representation of type hint int, param.annotation.__name__ has to be accessed BUT
    types.UnionType, like int|str, before python 3.14 has no attribute __name__, to get the string representation of int|str, access param.annotation ONLY.
    """
    param_str: str = str(param)
    if param_annotation is not inspect._empty:
        if "=" in param_str:
            type_hint_str = (
                param_str.split(":", maxsplit=1)[1].split("=", maxsplit=1)[0].strip()
            )
        else:
            type_hint_str = param_str.split(":", maxsplit=1)[1].strip()

    return type_hint_str, optional_flag
```

To make sure python 3.9 users could use features like int|str, which is introduced in python 3.10 correctly, what IDE and PEP suggests is to use "from \_\_future\_\_ import annotations", within from \_\_future\_\_ import annotations, you could see it clearly that type hints would all be converted to strings at Runtime.
I don't mean to but this is what python developers did.
image

cturan:
I can confirm that it fixed the issue and I was able to get it working, thank you.

Yacklin:
> I can confirm that it fixed the issue and I was able to get it working, thank you.

But i need to make reviewers happy about my changes, which is harder than writing a PR

Rocketknight1:
man you're lucky i'm on watch and isn't sgugger anymore or he would have had you guillotined for this level of sass :sweat_smile:

Rocketknight1:
Also, quick comment about your implementation: Python 3.9 is now [end-of-life](https://devguide.python.org/versions/). As a result, we're moving to a minimum supported version of 3.10, and it's not necessary to consider Py3.9 compatibility in new code. It's fine for your PR to assume that everyone will be running 3.10 or newer

Yacklin:
> man you're lucky i'm on watch and isn't sgugger anymore or he would have had you guillotined for this level of sass 😅

Oh Thanks to let me know hahaha

Yacklin:
> Also, quick comment about your implementation: Python 3.9 is now [end-of-life](https://devguide.python.org/versions/). As a result, we're moving to a minimum supported version of 3.10, and it's not necessary to consider Py3.9 compatibility in new code. It's fine for your PR to assume that everyone will be running 3.10 or newer

If you are happy with the implementation i have in the comment, i would put it in my branch later. Thanks to let me know that 3.9 is end of life.

Yacklin:
@Rocketknight1 I think it is good to go. I reviewed the docs for this PR and it looks good.
Also, some tests sometimes failed and sometimes didn't, but i believe you are fixing it right now.

src/transformers/utils/auto_docstring.py
```diff
@@ -1288,29 +1289,35 @@ def _process_parameter_type(param, param_name, func):
...
+    ):
+        optional_flag = True
+    """
+    to get the string representation of type hint int, param.annotation.__name__ has to be accessed BUT
+    types.UnionType, like int|str, before python 3.14 has no attribute __name__, to get the string representation of int|str, access param.annotation ONLY.
+    """
+    if param_annotation is not inspect._empty:
+        param_str: str = str(param)
+        if "=" in param_str:
+            type_hint_str = param_str.split(":", maxsplit=1)[1].split("=", maxsplit=1)[0].strip()
```
Rocketknight1: What do you think about `type_hint_str = getattr(param_annotation, "__name__", str(param_annotation))`? This should avoid the string manipulation on `param_str`, grabs `__name__` when available, and falls back cleanly when it isn't.

Yacklin: sounds good!!!! thanks for inspiration!!!!

Yacklin: I just tested it. It works fine but it would hide some details. For example, if the type hint is Optional[str], it would only print "Optional". Are you happy with it? My method presented here could preserve this details. What do you think? @Rocketknight1

Yacklin: I could give you two options, one is adopting your suggestions, which would hide details of type hint.

```py
if param_annotation is not inspect._empty:
    type_hint_str = getattr(param_annotation, "__name__", str(param_annotation))
```

or
mine could show the details of type hint
```py
if param_annotation is not inspect._empty:
    if (type(param_annotation) is not type) or not hasattr(param_annotation, "__name__"):
        type_hint_str = str(param_annotation)
    else:
        type_hint_str = param_annotation.__name__
```
which one do you prefer? i swear this is the last decision you need to make.

Yacklin:
this failed test shows up again....
image

Yacklin:
@Rocketknight1 Remember to review my new code btw

Yacklin:
@Rocketknight1 this is your last review for this PR i promise!!!

src/transformers/utils/auto_docstring.py
Rocketknight1: Ah, good point! I totally failed to spot that `__name__` would only grab the outer class if it was an `Optional` or `Union`. Two more suggestions, though:

1) `type(param_annotation) is not type` seems a little strange. It probably works but feels fragile! Maybe use `get_origin(type) is not None` instead?

2) The `type_hint_str` might start with `typing`, like `typing.Union(...)` which we probably don't want. Can we do `.removeprefix("typing.")` to clean up the displayed hints even further?

Yacklin: Hi @Rocketknight1 ,

1. if the type hint is basic type hint like int, str, bool or no type hint or Union[None], Optional[None], type(param_annotation) is actually type, program would have to access param_annotation.\_\_name\_\_ to get the cleanest string representation of this type hint. Also, if at the worst case this condition "breaks", another condition "not hasattr(param_annotation, "\_\_name\_\_")" could make sure error won't occur.
2. .removeprefix("typing.") might complicate this implementation. For backward compatibility, i need to add several if else statements to see if it's from typing or types module (maybe Python core dev would remove types or remove typing in the future because these two overlap a lot). I think keeping "typing." won't hurt that much and might provide some clarity at some point in the future.

I use the following script to test, feel free to play around with it

```py
import inspect
import types
from typing import Union, get_args, get_origin, Optional


def test(
    eee,
    aaa: int,
    ttt: Union[str, None],
    zzz: None,
    bbb: Union[None],
    iii: Optional[None],
    kkk: Union[int, str],
    lll: int | str = 5,
    ddd: Optional[str] = "",
    vvv: Union[int, str] = 6,
):
    pass


sig = inspect.signature(test)


def _process_parameter_type(param, param_name, func):
    """
    Process and format a parameter's type annotation.

    Args:
        param (`inspect.Parameter`): The parameter from the function signature
        param_name (`str`): The name of the parameter
        func (`function`): The function the parameter belongs to
    """
    param_annotation = param.annotation
    default_value = param.default
    origin = get_origin(param_annotation)
    args = get_args(param_annotation)
    optional_flag = False
    type_hint_str = ""
    if (
        (origin is Union and type(None) in args)
        or default_value is not inspect._empty
        or (origin is types.UnionType and type(None) in args)
        or param_annotation is types.NoneType
        or param_annotation is None
    ):
        optional_flag = True
    if param_annotation is not inspect._empty:

        if (type(param_annotation) is not type) or not hasattr(param_annotation, "__name__"):
            type_hint_str = str(param_annotation)
        else:
            type_hint_str = param_annotation.__name__

    return type_hint_str, optional_flag


for p_name, p in sig.parameters.items():
    print(_process_parameter_type(p, None, None))

```
corresponding output:
image

I personally need this bug to be fixed to continue my work so this is why I urge you to review. I can't come up any better ones without further inspiration btw.

Yacklin: @Rocketknight1

Yacklin: @Rocketknight1

Rocketknight1: Understood! Let me make some edits today and then I'll merge it, just to speed things up. Thank you for the PR!

Yacklin: Yeah folks have been waiting for it for a long while. Thanks for your review btw!!

Yacklin: Hi @Rocketknight1 , I just tested it "get_origin(param_annotation) is not None" and it outputs the same thing as mine did. Probably no need to change it.

Yacklin: You could build the docs for this PR again and let me review it if you like. I am always able to help with it!!!

Yacklin:
Man you changed my implementation significantly that i can't even understand lol

Yacklin:
@Rocketknight1 , i just tested your implementation and found it weird and wrong. Please revert your changes
image

src/transformers/utils/auto_docstring.py
```diff
@@ -1290,26 +1290,30 @@ def _process_parameter_type(param, param_name, func):
...
-                print(
-                    f"[ERROR] {param_type} for {param_name} of {func.__qualname__} in file {func.__code__.co_filename} has an invalid type"
-                )
-        if "ForwardRef" in param_type:
-            param_type = re.sub(r"ForwardRef\('([\w.]+)'\)", r"\1", param_type)
-        if "Optional" in param_type:
-            param_type = re.sub(r"Optional\[(.*?)\]", r"\1", param_type)
-            optional = True
+    if param.annotation == inspect.Parameter.empty:
+        return "", False
```
Yacklin, line 1293: Why don't you reuse variable optional, False instead?

src/transformers/utils/auto_docstring.py
```diff
@@ -1290,26 +1290,30 @@ def _process_parameter_type(param, param_name, func):
...
+    elif get_origin(param.annotation) is Union:
```
Yacklin: Did you consider types.UnionType?

src/transformers/utils/auto_docstring.py
```diff
@@ -1290,26 +1290,30 @@ def _process_parameter_type(param, param_name, func):
...
+    for subtype in subtypes:
+        if subtype is type(None):
+            optional = True
+            out_str.append("None")
+            continue
+        subtype = str(subtype)
+        if "typing" in subtype:
+            subtype = "".join(subtype.split("typing.")).replace("transformers.", "~")
+        elif hasattr(subtype, "__module__") and hasattr(subtype, "__name__"):
+            subtype = f"{subtype.__module__.replace('transformers.', '~').replace('builtins', '')}.{subtype.__name__}".removeprefix(
```
Yacklin: I just read the docs of removeprefix. It sounds good and yeah we could add it in.

Yacklin (commented):
I suggest to not merge PR with your current modifications

Yacklin:
@Rocketknight1 Man why don't you build upon my latest implementation? No need to build it again, if you read all my comments carefully!

Rocketknight1:
@Yacklin you gotta calm down lol. I'm the watcher for this entire repo so I have like 50 issues a day and this isn't a priority one. I'll make a fix and hopefully improve the docstring generation in general, but you need to let go of the laptop and touch grass while I do instead of pinging me every two hours!

Yacklin:
> @Yacklin you gotta calm down lol. I'm the watcher for this entire repo so I have like 50 issues a day and this isn't a priority one. I'll make a fix and hopefully improve the docstring generation in general, but you need to let go of the laptop and touch grass while I do instead of pinging me every two hours!

Sorry i just felt surprised by your implementation but do let me know if you ran into troubles while updating this PR

Rocketknight1:
cc @yacklin it seems to be giving the same outputs as the old system now, but cleaned up a lot, and working correctly for `|` union types (I made a test in GPT-J for that, but we can revert before merging).

Want to take a look and tell me if you're happy with it, and then I'll ping a core maintainer if so?

src/transformers/utils/auto_docstring.py
```diff
@@ -1290,28 +1291,34 @@ def _process_parameter_type(param, param_name, func):
...
-                print(
-                    f"[ERROR] {param_type} for {param_name} of {func.__qualname__} in file {func.__code__.co_filename} has an invalid type"
-                )
-        if "ForwardRef" in param_type:
-            param_type = re.sub(r"ForwardRef\('([\w.]+)'\)", r"\1", param_type)
-        if "Optional" in param_type:
-            param_type = re.sub(r"Optional\[(.*?)\]", r"\1", param_type)
+    if param.annotation == inspect.Parameter.empty:
+        return "", False
+    # This is, astonishingly, the right way to do it: https://docs.python.org/3/library/typing.html#typing.Union
```
Yacklin, line 1296: You are right. This is best approach and i ignored it lol

Yacklin (commented):
Hi @Rocketknight1 , I just tested it locally and i noticed that if the type hint is None, like age: None, it will mark it non-optional. It's an edge case but good to let you know.

Yacklin:
Also, if the parameter has default value, it should be considered optional. It seems that your implementation didn't include this. I am wondering whether this function is supposed to consider default value or not?

Rocketknight1:
@Yacklin I think that behaviour for type hints like `age: None` is fine - that type hint should almost never appear, since it indicates that the only acceptable argument value is `None`. The much more common case is something like `int | none` or `Union[int, None]` or `Optional[int]`, which all mean the same thing, and I think it handles that correctly, right?

Yacklin:
> @Yacklin I think that behaviour for type hints like `age: None` is fine - that type hint should almost never appear, since it indicates that the only acceptable argument value is `None`. The much more common case is something like `int | none` or `Union[int, None]` or `Optional[int]`, which all mean the same thing, and I think it handles that correctly, right?

@Rocketknight1 , A quick question: Should default value be considered in this function? I mean if there is a default value for a parameter, this parameter should be optional. If other functions are designed to handle this, simply ignore what i highlighted here.
I just tested it and it seems that the my idea of implementation is slightly different from yours. Since you are a watcher, i will take yours.

Yacklin:
maybe we could revert changes on gptj since this PR doesn't do anything with it.

Rocketknight1:
Yeah, those were mostly for testing. Will revert the GPT-J changes, keep the rest and merge!

Rocketknight1 (approved):

Yacklin:
I just realized that when the type hint of a parameter is None, the only acceptable argument is None for this parameter. And it should be marked as non-optional. But since nobody would annotate a parameter with None, the if statement i wrote to handle None is useless. You could remove that line, if necessary. Anyway, thanks for editing!

Yacklin:
@Rocketknight1 , also, the way this function handles default value seems to be not perfect (my bad). If a parameter has no type hint but has default value, it would be considered non-optional. Maybe it would introduce some problems in the future.

PR #42433, opened 2025-11-26
perf: Remove implicit CPU-GPU syncs due to implicit .item() call

Dhruv88:
# What does this PR do?
Remove unnecssary cudaStreamSynchronize by implicit call to .item()
Motivation mentioned in issue tagged below.

Fixes #42422

Made changes as proposed in the issue.

## Heads up
When I ran the integration tests two of them were failing:
test_llama_3_1_hard
test_model_7b_logits_bf16

The issue was the actual output was not matching the expected output. I am using an A100 GPU. 
For first case actual output was matching the expected output of rocm and not the cuda one.
In second case it was:
tensor([[-6.5081, -4.1175, -4.9761, -3.1678,  0.8199, -3.0029,  1.2809, -3.3309]] which was again differs a bit from the expected cuda outputs.

However, these tests are failing even without my change. Also, the output with and without my change is same. So, my belief is that my changes are not causing the failure.

Another point is that similar code appears across multiple files so it this one is as per expectation I can make changes in the other files as well. Please let me know how I should proceed.

@Rocketknight1 @ArthurZucker @Cyrilvallez

### DIFF
--- a/src/transformers/models/apertus/modeling_apertus.py
+++ b/src/transformers/models/apertus/modeling_apertus.py
@@ -392,8 +392,8 @@ class ApertusModel(ApertusPreTrainedModel):
 
         if cache_position is None:
             past_seen_tokens = past_key_values.get_seq_length() if past_key_values is not None else 0
-            cache_position: torch.Tensor = torch.arange(
-                past_seen_tokens, past_seen_tokens + inputs_embeds.shape[1], device=inputs_embeds.device
+            cache_position: torch.Tensor = (
+                torch.arange(inputs_embeds.shape[1], device=inputs_embeds.device) + past_seen_tokens
             )
 
         if position_ids is None:
--- a/src/transformers/models/arcee/modeling_arcee.py
+++ b/src/transformers/models/arcee/modeling_arcee.py
@@ -397,8 +397,8 @@ class ArceeModel(ArceePreTrainedModel):
 
         if cache_position is None:
             past_seen_tokens = past_key_values.get_seq_length() if past_key_values is not None else 0
-            cache_position: torch.Tensor = torch.arange(
-                past_seen_tokens, past_seen_tokens + inputs_embeds.shape[1], device=inputs_embeds.device
+            cache_position: torch.Tensor = (
+                torch.arange(inputs_embeds.shape[1], device=inputs_embeds.device) + past_seen_tokens
             )
 
         if position_ids is None:
--- a/src/transformers/models/aria/modeling_aria.py
+++ b/src/transformers/models/aria/modeling_aria.py
@@ -726,8 +726,8 @@ class AriaTextModel(AriaTextPreTrainedModel):
 
         if cache_position is None:
             past_seen_tokens = past_key_values.get_seq_length() if past_key_values is not None else 0
-            cache_position: torch.Tensor = torch.arange(
-                past_seen_tokens, past_seen_tokens + inputs_embeds.shape[1], device=inputs_embeds.device
+            cache_position: torch.Tensor = (
+                torch.arange(inputs_embeds.shape[1], device=inputs_embeds.device) + past_seen_tokens
             )
 
         if position_ids is None:
--- a/src/transformers/models/bitnet/modeling_bitnet.py
+++ b/src/transformers/models/bitnet/modeling_bitnet.py
@@ -396,8 +396,8 @@ class BitNetModel(BitNetPreTrainedModel):
 
         if cache_position is None:
             past_seen_tokens = past_key_values.get_seq_length() if past_key_values is not None else 0
-            cache_position: torch.Tensor = torch.arange(
-                past_seen_tokens, past_seen_tokens + inputs_embeds.shape[1], device=inputs_embeds.device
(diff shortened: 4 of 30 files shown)


Rocketknight1:
Yes, those CI errors are not your fault! However, the `check_repository_failure` issue means that your change might need to be propagated to other files. Run `make fix-copies` and other models that are copying llama code will also update.

Dhruv88:
@Rocketknight1 Among the tests that failed, two of them look to be due to worker timeout or crash so rerunning should work. The third one is for `modeling_colqwen2.py`. However, no code in this file has been changed in my PR so it is unrelated to my changes. The test name is `test_load_save_without_tied_weights` which comes from `test_modeling_common.py`. 

Anyways, after some investigation here is my understanding and possible fix for the issue.

The test sets `config.tie_word_embeddings=False`. However, there is no such key config. What actually should be set is `config.vlm_config.tie_word_embeddings=False`. Doing this instead passes the test since it correctly unties the weight in model.vlm.  

Please let me know if my understanding is right. If yes, then I can try to make a quick fix by ensuring the correct key is set.

ArthurZucker (commented):
Very nice! Can you make sur compile with reduce overheads works as expected still? That's my only concern, otherwise LGTM!

Rocketknight1:
Colqwen2 issue should be fixed on main too, so a rebase should make that go away

Dhruv88:
> Very nice! Can you make sur compile with reduce overheads works as expected still? That's my only concern, otherwise LGTM!

@ArthurZucker. I tried the below minimal code taken from official documentation [here](https://huggingface.co/docs/transformers/main/en/llm_optims?static-kv=2.+StaticCache).
I made minor changes to it so that cache_position is not passed and the path with changed code is taken. 
Putting the snippet in this [gist](https://gist.github.com/Dhruv88/42a140976b65f0b3248338f7e71d3cbe) for reference.

Note that the `cache_position` is not passed to `compiled_model` forward pass and the compiled `decode_one_tokens`.

With the updated code it runs and generates the ouput below:
```
['Simply put, the theory of relativity states that 1) the laws of physics are the same for all observers in uniform motion relative to one another and 2) the speed of light is always constant, regardless of the motion of the observer or the!', 'My favorite all time favorite condiment is ketchup. I love it on burgers, fries, scrambled eggs, and even as a dip for chicken tenders. I have tried many different brands of ketchup over the years, but my go-to is always!']
```

However, with the earlier code the torch.compile fails with a graph break:
```
  File "/home/dd/transformers/test.py", line 53, in <module>
    next_token = decode_one_tokens(model, next_token.clone(), None, past_key_values)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dd/miniconda3/envs/transformers/lib/python3.12/site-packages/torch/_dynamo/eval_frame.py", line 841, in compile_wrapper
    raise e.with_traceback(None) from e.__cause__  # User compiler error
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch._dynamo.exc.Unsupported: Data dependent operator
  Explanation: Operator `aten._local_scalar_dense.default` has a non-Tensor output whose value is dependent on the data of Tensor inputs.
  Hint: Enable tracing of data-dependent output operators with `torch._dynamo.config.capture_scalar_outputs = True`

  Developer debug context: aten._local_scalar_dense.default

 For more details about this graph break, please visit: https://meta-pytorch.github.io/compile-graph-break-site/gb/gb0033.html

from user code:
   File "/home/dd/transformers/test.py", line 20, in decode_one_tokens
    logits = model(
  File "/home/dd/transformers/src/transformers/utils/generic.py", line 764, in wrapper
    output = func(self, *args, **kwargs)
  File "/home/dd/transformers/src/transformers/models/llama/modeling_llama.py", line 489, in forward
    outputs: BaseModelOutputWithPast = self.model(
  File "/home/dd/transformers/src/transformers/utils/generic.py", line 919, in wrapper
    outputs = func(self, *args, **kwargs)
  File "/home/dd/transformers/src/transformers/models/llama/modeling_llama.py", line 404, in forward
    cache_position: torch.Tensor = torch.arange(

Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"
```

To ensure that this code path is taken I also put a breakpoint before the line in the uncompiled version and ran in debugging mode. Can confirm that it indeed executes that line.

Let me know if you want me to check for anything else!

ArthurZucker:
Sounds great!

Dhruv88:
@ArthurZucker @Rocketknight1 I rebased a couple of times but the test_torch and test_tokenization always seem to fail due to worker crash and the test_processor due to timeout. Maybe just rerunning the failed tests might help but not sure how to do that. How can I understand the issue here?

ArthurZucker:
Don't worry these are unrelated!

ArthurZucker:
Thanks for the PR!

PR #42404, opened 2025-11-25
Add Flash Attention 4 (CuTe DSL) Support

sambhavnoobcoder:
## Problem Statement

Flash Attention 4 represents a significant architectural shift in the flash-attention package:

1. **Different import path**: FA4 uses `flash_attn.cute` submodule instead of the main `flash_attn` package
2. **API incompatibility**: FA4's `flash_attn_varlen_func` has a different signature - it does NOT accept `max_seqlen_q` and `max_seqlen_k` parameters (calculates them internally from `cu_seqlens`)
3. **New parameters**: FA4 introduces new optimization parameters like `learnable_sink`, `num_splits`, and `pack_gqa`
4. **Removed parameters**: FA4 removes `dropout_p` and `alibi_slopes` support
5. **Hardware requirements**: FA4 requires CUDA compute capability SM 8.0+ (Ampere or newer), with optimal performance on SM 9.0+ (Hopper/Blackwell)

Without explicit FA4 support, users cannot leverage these improvements even when they have compatible hardware and the flash-attn package with CuTe DSL installed.

## Solution Design

The solution maintains full backward compatibility while adding FA4 support through:

### 1. Detection Layer
Added `is_flash_attn_4_available()` function that checks:
- CUDA availability
- flash-attn package installation
- Presence of `flash_attn.cute` submodule
- GPU compute capability >= SM 8.0

### 2. Priority-Based Auto-Selection
When `attn_implementation=None`, the selection order is:
```
FA4 > FA3 > FA2 > SDPA > Eager
```

FA4 gets highest priority on compatible hardware for optimal performance.

### 3. Runtime Introspection
Created `_is_using_fa4()` helper that uses function signature inspection to detect FA4 vs FA2/FA3 at runtime. This enables conditional code paths without hardcoded version checks.

### 4. Conditional Varlen Calls
Modified two critical call sites in `_flash_attention_forward()` to conditionally pass parameters:
- **FA4 path**: Calls without `max_seqlen_q` and `max_seqlen_k` (calculates internally)
- **FA2/FA3 path**: Calls with explicit max_seqlen parameters (required)

### 5. Parameter Support
Extended `_process_flash_attention_kwargs()` to handle FA4-specific parameters, with automatic filtering based on introspection to maintain compatibility across versions.

### 6. Registration
Registered `flash_attention_4` in `AttentionInterface._global_mapping` to enable explicit selection via `attn_implementation="flash_attention_4"`.

### Core Changes

**Detection and Import**
[...]

PR #42435, opened 2025-11-26
:rotating_light: [`FA4`] Initial support

vasqu:
## :rotating_light: Breaking change
- FA2 is only supported from version `2.3.3` and on

This is due to the fact that this is older than 2+ years (we deprecate torch in 2 year cycles for example) as well as it giving fairly high maintenance burden.

## Related issues and PRs
Fixes #42405
Closes #42404 as it has a lot of unnecessary logic and tests alongside it

## Testing
Sanity Testing
```bash
RUN_SLOW=1 pytest tests/models/llama/test_modeling_llama.py -k flash
```
- [x] Passes all flash attention 4 tests

First quick numbers (hopper)
```text
# No attention mask (base fa)
# RUN_SLOW=1 pytest -s tests/generation/test_flash_attention_parity.py
Latency:
    With FA2: 381.5204345703125
    With FA3: 362.461669921875
    With FA4: 373.788427734375

# With attention mask (varlen fa)
Latency:
    With FA2: 509.337646484375
    With FA3: 476.020654296875
    With FA4: 476.72578125
```

NOTE: FA4 is optimized for blackwell, these are quick numbers on hopper --> it's faster than FA2 but around the same on varlen, between the other FAs on non-varlen


src/transformers/modeling_utils.py
```diff
@@ -1715,6 +1717,86 @@ def _flash_attn_3_can_dispatch(self, is_init_check: bool = False) -> bool:
...
+                f" the model is hosted, on its model hub page: https://huggingface.co/{self.config._name_or_path}/discussions/new"
+                " or in the Transformers GitHub repo: https://github.com/huggingface/transformers/issues/new"
+            )
+
+        if not is_flash_attn_4_available():
+            preface = "FlashAttention4 has been toggled on, but it cannot be used due to the following error:"
+
+            if (
+                importlib.util.find_spec("flash_attn") is None
+                or "flash-attn-cute" in PACKAGE_DISTRIBUTION_MAPPING["flash_attn"]
```
vasqu: should be not in

vasqu:
FA4 support cc @stas00 if you wanna play around with this PR. It's pretty much ready, just not convinced by the numbers but I also don't have quick access to a blackwell GPU (at least today :D)

sfc-gh-sbekman (commented):
Thank you for working on this, Anton. Going to try it out.

To make it easier to try your PR please add to the OP how to install FA4, since it's non-trivial to find.
```
git clone https://github.com/Dao-AILab/flash-attention/
cd flash-attention
cd flash_attn/cute
uv build --wheel . -v --no-build-isolation --out-dir flash-attention/wheels
uv pip install flash-attention/wheels/flash_attn_cute*.whl --prerelease=allow
```

sfc-gh-sbekman:
OK, gave it a test ride using your PR and the above comment's install of FA4 on B200.

I did a quick test with Llama-8b and the integration worked smoothly but the tflops performance is much worse than FA2 - 2-5x slower. Not sure if it's an issue with integration or the FA4 code or the pytorch version - most likely the upstream since the integration is just a wrapper

I tried pt-2.9.1-cu130 and pt-nightly-cu130 - same outcome

edit: `sdpa` in pt-nightly, which supposedly backported FA4, is about 3x faster than FA4 on its own using the same llama-8b - since they both should be using the same code, perhaps there is an issue with the integration?

vasqu:
Thanks for checking this out and all the pointers @sfc-gh-sbekman :heart: 

> To make it easier to try your PR please add to the OP how to install FA4, since it's non-trivial to find.

For sure, I'll add some docs for FA4 before release. Maybe also FA3 in a different PR.

> I did a quick test with Llama-8b and the integration worked smoothly but the tflops performance is much worse than FA2 - 2-5x slower. Not sure if it's an issue with integration or the FA4 code or the pytorch version - most likely the upstream since the integration is just a wrapper

Shoot, so it wasn't an GPU arch issue... This is weird

> `sdpa` in pt-nightly, which supposedly backported FA4, is about 3x faster than FA4 on its own using the same llama-8b - since they both should be using the same code, perhaps there is an issue with the integration?

Do you have a code snippet? There are so many edge cases with sdpa that it maybe is not even entering the FA backend path? Could be quickly checked by restricting the backend usage on SDPA with their context manager 
```python
with torch.nn.attention.sdpa_kernel([torch.nn.attention.SDPBackend.FLASH_ATTENTION]):
    pass  # do your thing here
```
I'm also unsure how FA4 is integrated in SDPA? Do we need to use a flag there? I remember that cudnn backend needed special treatment

stas00:
> Shoot, so it wasn't an GPU arch issue... This is weird

Did you mean that you too have observed a similar slowdown?

> Do you have a code snippet? 

I was just using https://github.com/snowflakedb/ArcticTraining/ normal SFT training recipe where I tried different attention mechanisms. Just normal fwd/bwd/step - nothing special added.

> I'm also unsure how FA4 is integrated in SDPA? Do we need to use a flag there? I remember that cudnn backend needed special treatment

They copied/adapted the FA4 kernels see: https://github.com/huggingface/transformers/pull/42435 - you'd need pt nightly for that to work.

vasqu:
> Did you mean that you too have observed a similar slowdown?

I just did some quick numbers on inference, see the test I noted down in the PR description. I used an H100 there and as you can see it's slower (not on the same magnitude as in your samples - would say it's a mixture of model size / context size)

> I was just using https://github.com/snowflakedb/ArcticTraining/ normal SFT training recipe where I tried different attention mechanisms. Just normal fwd/bwd/step - nothing special added.

Gotcha, I will try to separate our implementation from the base fn of the FA library to see if our wrappers cause this or maybe some perf regression happened sometime else.

> They copied/adapted the FA4 kernels see: https://github.com/huggingface/transformers/pull/42435 - you'd need pt nightly for that to work.

Wrong link? My assumption / hunch was that maybe
1. Even nightly might not use FA4 per default and sticks to FA2 per default, i.e. might need some extra flags to enable that specific backend. But that's just my feeling, need to look into it.
2. Our implementations has some issues where attention masks are created even when it is not needed (full (causal) attention). If a mask is passed to SDPA, then the FA backend can never be entered per their restrictions. So I thought, maybe, we have this case (SDPA with xformers faster than FA4 - xformers is not so bad on short contexts <2k).

stas00:
My apologies, here is the correct link https://github.com/pytorch/pytorch/pull/167348

sfc-gh-sbekman:
Some useful updates from talking to Tri:

- FA4 is supposed to replace FA2 and FA3 and would work with A/H/B archs
- FA4 varlen support is planned in a few weeks time

edixiong:
Hi @sfc-gh-sbekman @vasqu Thanks for contributing! I think the varlen is supported. Do you mind testing FA4 again?

sfc-gh-sbekman:
just tried with the `main` version of fa - `bwd` doesn't work:

```
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/torch/utils/checkpoint.py", line 325, in backward
[rank3]:     torch.autograd.backward(outputs_with_grad, args_with_grad)
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/torch/autograd/__init__.py", line 364, in backward
[rank3]:     _engine_run_backward(
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/torch/autograd/graph.py", line 865, in _engine_run_backward
[rank3]:     return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[rank3]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/torch/autograd/function.py", line 317, in apply
[rank3]:     return user_fn(self, *args)
[rank3]:            ^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/interface.py", line 1385, in backward
[rank3]:     dq, dk, dv = _flash_attn_bwd(
[rank3]:                  ^^^^^^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/interface.py", line 1059, in _flash_attn_bwd
[rank3]:     _flash_attn_bwd.compile_cache[compile_key] = cute.compile(
[rank3]:                                                  ^^^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/cute_dsl_utils.py", line 118, in cute_compile_patched
[rank3]:     output = cute_compile_og(*args, **kwargs)
[rank3]:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_sm100.py", line 802, in __call__
[rank3]:     ).launch(
[rank3]:   ^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_sm100.py", line 1213, in kernel
[rank3]:     if warp_idx >= self.compute_warp_ids[0] and warp_idx <= self.compute_warp_ids[-1]:
[rank3]: ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_sm100.py", line 1215, in then_block_16
[rank3]:     self.compute_loop(
[rank3]: ^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_sm100.py", line 2062, in compute_loop
[rank3]:     while work_tile.is_valid_tile:
[rank3]: ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_sm100.py", line 2317, in if_region_4
[rank3]:     if process_tile:
[rank3]: ^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_sm100.py", line 2339, in then_block_5
[rank3]:     consumer_state_dKV = self.epilogue_dK_or_dV_tma(
[rank3]:                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_sm100.py", line 2877, in epilogue_dK_or_dV_tma
[rank3]:     tdKVtdKV_t2r = self.split_wg(tdKVtdKV_t2r_p, wg_idx, num_wg)[None, None, 0, 0]
[rank3]:                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_sm100.py", line 1834, in split_wg
[rank3]:     t = cute.logical_divide(
[rank3]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/nvidia_cutlass_dsl/python_packages/cutlass/_mlir/dialects/_cute_ops_gen.py", line 1805, in __init__
[rank3]:     super().__init__(self.build_generic(attributes=attributes, operands=operands, successors=_ods_successors, regions=regions, loc=loc, ip=ip))
[rank3]:                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank3]: ValueError: Operation creation failed
loc("t = cute.logical_divide("("/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_sm100.py":1834:16)): error: failed to perform a valid division of '!cute.layout<"(((32,32),1),1,1,1):(((1,65536),0),0,0,0)">' by #cute.tile<"[1024:1;1:0;1:0;0:1]">

```

sfc-gh-sbekman:
Tried again with today's FA4 and a new error is reported on H200 w/ varlen.
```
[rank6]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/torch/autograd/function.py", line 311, in apply
[rank6]:     return user_fn(self, *args)
[rank6]:            ^^^^^^^^^^^^^^^^^^^^
[rank6]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/interface.py", line 1395, in backward
[rank6]:     dq, dk, dv = _flash_attn_bwd(
[rank6]:                  ^^^^^^^^^^^^^^^^
[rank6]:   File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/interface.py", line 617, in _flash_attn_bwd
[rank6]:     assert not is_varlen, "varlen backward is not yet supported on sm90"
[rank6]:            ^^^^^^^^^^^^^
[rank6]: AssertionError: varlen backward is not yet supported on sm90
```

non-varlen works, but haven't measured the performance comparison

edixiong:
Thanks for checking!

edixiong:
Seems like FA4 will be published soon: https://github.com/Dao-AILab/flash-attention/pull/2282
I will ping back later when they release the package.

vasqu:
Thanks for keeping us in the loop :saluting_face: will check it out then when we got the release

stas00:
Feels like the announcement was premature, I can't get it to even load:

```
$ uv pip install fa4
Using Python 3.12.12 environment at: /home/yak/miniconda3/envs/dev
Resolved 37 packages in 56ms
Prepared 1 package in 52ms
Uninstalled 1 package in 0.46ms
Installed 3 packages in 11ms
 - cuda-python==12.9.4
 + cuda-python==13.0.3
 + nvidia-cutlass-dsl==4.4.1
 + nvidia-cutlass-dsl-libs-base==4.4.1


$ python -c "from flash_attn.cute import flash_attn_func"
  File "<string>", line 1, in <module>
  File "/home/yak/miniconda3/envs/dev/lib/python3.12/site-packages/flash_attn/cute/__init__.py", line 10, in <module>
    import cutlass.cute as cute
ModuleNotFoundError: No module named 'cutlass.cute'
```

Does anybody have a better luck?

stas00:
OK, @mayank31398 helped me to find `cutlass.cute` - NVIDIA hid those packages from python - no clue why - instead of putting them into `site-packages/cutlass` it puts them under `site-packages/nvidia_cutlass_dsl/python_packages/cutlass` ?!?!?!

So I had to do:
```
export PYTHONPATH=$CONDA_PREFIX/lib/python3.12/site-packages/nvidia_cutlass_dsl/python_packages/
```
to load `cutlass.cute`

stas00:
ok, so now the integration code fails to load fa4 so back to you @vasqu:

```
[rank0]:   File "/code/users/stas/github/transformers/src/transformers/modeling_utils.py", line 1255, in __init__
[rank0]:     self.config._attn_implementation_internal = self._check_and_adjust_attn_implementation(
[rank0]:                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/code/users/stas/github/transformers/src/transformers/modeling_utils.py", line 1838, in _check_and_adjust_attn_implementation
[rank0]:     applicable_attn_implementation = self.get_correct_attn_implementation(
[rank0]:                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/code/users/stas/github/transformers/src/transformers/modeling_utils.py", line 1870, in get_correct_attn_implementation
[rank0]:     self._flash_attn_can_dispatch(flash_attn_version=4, is_init_check=is_init_check)
[rank0]:   File "/code/users/stas/github/transformers/src/transformers/modeling_utils.py", line 1658, in _flash_attn_can_dispatch
[rank0]:     self._flash_attn_4_import_error()
[rank0]:   File "/code/users/stas/github/transformers/src/transformers/modeling_utils.py", line 1614, in _flash_attn_4_import_error
[rank0]:     raise ImportError(f"{preface} the package flash_attn (under cute) seems to be not installed.")
[rank0]: ImportError: FlashAttention4 has been toggled on, but it cannot be used due to the following error: the package flash_attn (under cute) seems to be not installed.

```

this needs to be fixed. edit: proposed a fix in the next set of comments.

src/transformers/modeling_utils.py
```diff
@@ -1598,116 +1583,135 @@ def _flash_attn_2_can_dispatch(self, is_init_check: bool = False) -> bool:
...
+
+    def _flash_attn_4_import_error(self):
+        if not is_flash_attn_4_available():
+            preface = "FlashAttention4 has been toggled on, but it cannot be used due to the following error:"
+
+            if (
+                importlib.util.find_spec("flash_attn") is None
+                or "flash-attn-cute" not in PACKAGE_DISTRIBUTION_MAPPING["flash_attn"]
+            ):
+                raise ImportError(f"{preface} the package flash_attn (under cute) seems to be not installed.")
```
stas00:
```suggestion
            if importlib.util.find_spec("flash_attn.cute") is None:
                raise ImportError(f"{preface} the package flash_attn (cute) seems to be not installed.")
```

src/transformers/modeling_utils.py
```diff
@@ -1598,116 +1583,135 @@ def _flash_attn_2_can_dispatch(self, is_init_check: bool = False) -> bool:
...
+                major, _ = torch.cuda.get_device_capability()
+                if major < 9:
                     raise ValueError(
-                        "You are attempting to use Flash Attention 2 with a model not initialized on GPU and with no GPU available. "
-                        "This is not supported yet. Please make sure to have access to a GPU and either initialise the model on a GPU by passing a device_map "
-                        "or initialising the model on CPU and then moving it to GPU."
+                        f"{preface} Flash Attention 4 requires compute capability >= 9.0, but found {torch.cuda.get_device_capability()} with compute capability {major}.0."
                     )
+                else:
+                    raise ImportError(f"{preface} Flash Attention 4 is not available.")
```
stas00:
```suggestion
```

stas00 (commented):
these are needed to make it work with the current fa4 as of this morning

vasqu:
Thanks a lot, will try to take a stab tomorrow :saluting_face:

stas00:
FWIW, I run a benchmark on a B300 comparing FA2 to FA4 and I see only a ~3% improvement end-to-end, which makes no sense - it should be much higher. 

Either something isn't right, or FA4 isn't quite there yet, I'm not sure.

This use case uses varlen.

edit: my colleague clocked an almost 2x speed up end-to-end with FA4 over FA2 on B200 - different workload. So that's a great sign. Perhaps his wasn't varlen.

vasqu:
Ok, so I did not go as far as I wanted to because I had torch 2.10 issues in regards to cublass (for anyone interested, pin `nvidia-cublas-cu12==12.9.1.4` re [issue](https://github.com/pytorch/pytorch/issues/174949))

I got FA4 working with my commit 776a1af358ce1d187e927caee228c7a16f393d8c and checked the parity test where I fell into a rabbit hole:
- #41427 changed the loading logic and truly enabled properly loading different FA versions in the same process, i.e. the same FA was run twice
- Fixed loading already shows diffs within FA3 and FA2 

My initial numbers (still hopper, will check with blackwell later on if I can) show that it's around the same time as FA3. Iirc, this should be varlen as well.

Edit: another note `head dim == 256` does not work and results in a tvm-ffi error which is not so obvious

src/transformers/modeling_utils.py
```diff
@@ -1616,10 +1616,7 @@ def _flash_attn_4_import_error(self):
         if not is_flash_attn_4_available():
             preface = "FlashAttention4 has been toggled on, but it cannot be used due to the following error:"
 
-            if (
-                importlib.util.find_spec("flash_attn") is None
-                or "flash-attn-cute" not in PACKAGE_DISTRIBUTION_MAPPING["flash_attn"]
-            ):
+            if importlib.util.find_spec("flash_attn") is None or importlib.util.find_spec("flash_attn.cute") is None:
```
sfc-gh-sbekman: I don't think this is correct, as it'll find FA2 and succeed, no?

was there something wrong with the fixes I proposed yesterday? You ignored those and wrote a different version w/o saying why they didn't work for you

vasqu: We shouldn't have entered that path as the import shouldn't fail so I fixed the import utils first. This here only happens if we know that fa4 is requested so it doesn't really matter in relation to fa2. The or is to first check the base package and then the underlying fa4 cute

sfc-gh-sbekman: oh of course, because it's `is None` - let's change this to `A is not None and B is not None` - would be more intuitive to read, no?

we want `if A and B` - easier to read than `if not A or not B`

sfc-gh-sbekman: I rewrote my comments - as they made no sense. All is good.

vasqu: I should have communicated the changes a bit more as well, no worries. Will check a bit more tomorrow probably

vasqu:
@sfc-gh-sbekman So it was pretty much ready already, I added some small details especially on the parity test. These deviation are somewhat expected and was already happening on FA3 (but went undetected for quite a while).

I added support for attention sinks so people may run GPT Oss which produces fairly sane output, similar to FA3 (vllm version).

I will revisit with some blackwell gpu to properly benchmark, hopper just really ain't the target device group (at least so far). Probably also adding some docs but thankfully the install is easy

jiosephlee:
> I added support for attention sinks so people may run GPT Oss which produces fairly sane output, similar to FA3 (vllm version).

Butting in with a naive question but does FA4 include bwd kernels? afaik, bwd kernels for sink attention are not supported in the main release of FA2/3

vasqu:
Iirc, attention sinks are only in form of PRs available for fa2/3. On a quick glance, fa4 only support the fwd for the sinks, the bwd is ignored.

You could quickly check it by just creating random tensors, run some arbitrary loss and check the grads of your tensors. Can do so tomorrow or so if no one has checked until then

The more general stuff has a bwd version

src/transformers/modeling_flash_attention_utils.py
```diff
@@ -564,9 +575,28 @@ def _process_flash_attention_kwargs(
...
+    # There is a limitation of the flash attention API, as the function `flash_attn_varlen_func`
+    # may require `max_length_q`, `max_length_k` to be passed as `int` and not `torch.Tensor`.
+    #
+    # You can either set
+    #   - Env: `TORCHDYNAMO_CAPTURE_SCALAR_OUTPUTS=1`
+    #   - Before compiling: `torch._dynamo.config.capture_scalar_outputs = True`
+    # to allow torch compile to handle scalar outputs in those cases.
+    if supports_mapping["max_seqlen_q"] and max_seqlen_q is not None:
+        if not isinstance(max_seqlen_q, int) and is_tracing(max_seqlen_q):
+            max_seqlen_q = max_seqlen_q.item()
```
Cyrilvallez, line 654: Why do we also check `is_tracing` here? I believe it would make it fail for previous FA version and no compile

src/transformers/modeling_utils.py
```diff
@@ -1607,116 +1592,132 @@ def _flash_attn_2_can_dispatch(self, is_init_check: bool = False) -> bool:
...
                 else:
+                    raise ImportError(f"{preface} Flash Attention 3 is not available.")
+            else:
+                raise ValueError(
+                    f"{preface} Flash Attention 3 is not available on CPU. Please make sure torch can access a CUDA device."
+                )
+
+    def _flash_attn_4_import_error(self):
+        if not is_flash_attn_4_available():
+            preface = "FlashAttention4 has been toggled on, but it cannot be used due to the following error:"
```
Cyrilvallez: Any way we can take the opportunity to make this slightly simpler instead of more or less duplicating 3 times for each fa?

src/transformers/modeling_utils.py
```diff
@@ -1877,9 +1878,9 @@ def _check_and_adjust_attn_implementation(
                 # raise the proper exception for requested flash attention
                 if requested_original_flash_attn:
                     if attn_implementation.endswith("2"):
-                        self._flash_attn_2_can_dispatch()
+                        self._flash_attn_can_dispatch(flash_attn_version=2, is_init_check=is_init_check)
                     else:
-                        self._flash_attn_3_can_dispatch()
+                        self._flash_attn_can_dispatch(flash_attn_version=3, is_init_check=is_init_check)
```
Cyrilvallez: SHould'nt we have a new path for fa4 here?

Cyrilvallez (commented):
Nice! Just a few comments!

src/transformers/modeling_flash_attention_utils.py
vasqu: This doesn't fail for any FA version. In non-compile environments, FA can handle this and we avoid device syncs with this. Only compile fails, because the fake ops and the underlying signature forces base int which dynamo cannot reason about if it is a scalar int tensor instead.

Initially, this change was specifically for FA4 which did not have max_seqlen_q/k but at some point it was added back with the same signature issue. This way we boost non-compile forward at least

src/transformers/modeling_utils.py
vasqu: Not really, I already refactored as much as I could into https://github.com/vasqu/transformers/blob/65912b2c02fe2f0653eb9710dc32b84aa209ae56/src/transformers/modeling_utils.py#L1635

There are subtle differences that cannot be ignored and these are the `_flash_attn_x_import_error` ones
- FA2 has very specific versioning stuff along kernels support for more devices like xpu
- FA3 has only support for ampere and hopper
- FA4 has only support for hopper and blackwell

src/transformers/modeling_utils.py
vasqu: There is no FA4 hf kernel yet. It needs support for cute and tvm ffi first. Iirc, cute is done but tvm still needs work

src/transformers/modeling_utils.py
vasqu: Let me check if I can simplify a bit more tho

sfc-gh-sbekman: If I remember correctly the plan is to make fa4 work for any gpu arch, so there is only one package to maintain. But perhaps this has changed recently - or more likely the older archs are just not a priority.

e.g. FA4 on hopper still doesn't support varlen

vasqu: That's a good point, I need to write some docs for compatibility tbh. It may not be super obvious atm

vasqu: Ok I refactored a bit in d07749f4c6c11d0d4726d5d6a9ca272e76424d34

It will force a lower bound on FA2 tho. Imo, it doesn't make sense to support these very old versions which atp are older than 2 years as it introduces additional checks. Wdyt about this refactored version

src/transformers/modeling_utils.py
```diff
@@ -1547,176 +1551,183 @@ def can_generate(cls) -> bool:
...
+                "cuda_min_major_version": 8,  # Ampere
+            },
+            4: {
+                "flash_attn_version": 4,
+                "general_availability_check": is_flash_attn_4_available,
+                "pkg_availability_check": lambda *args, **kwargs: importlib.util.find_spec("flash_attn") is not None
+                and importlib.util.find_spec("flash_attn.cute") is not None,
+                "supported_devices": ((is_torch_cuda_available, "cuda"),),
+                "cuda_min_major_version": 9,  # Hopper
+            },
```
ArthurZucker: we could bring this outside this scope!

ArthurZucker: mayne in fa2 utils

tests/generation/test_flash_attention_parity.py
```diff
@@ -74,68 +81,106 @@ def _benchmark_generation(self, model, inputs, n_warmup=3, n_runs=5):
 
         return start_time.elapsed_time(end_time) / n_runs
 
-    @pytest.mark.flash_attn_3_test
     @require_torch_gpu
+    @pytest.mark.flash_attn_test
+    @pytest.mark.flash_attn_3_test
+    @pytest.mark.flash_attn_4_test
```
ArthurZucker:
```suggestion
    @pytest.mark.all_flash_attn_tes
```
or 2 is the only not supported ahha

ArthurZucker (approved):
Ty!

imstevenpmwork:
Hey @vasqu  👋 Just a quick heads-up that the API changes in this PR (shipped in the 5.4.0 release) introduced a breaking change for us.

Over at [lerobot](https://github.com/huggingface/lerobot), we were relying on `is_flash_attn_greater_or_equal_2_10`, which looks like it got removed during the `import_utils.py` refactor. Just wanted to flag this to save some debugging time in case anyone else is suddenly staring at a [red CI](https://github.com/huggingface/lerobot/actions/runs/23629232685)!

Cheers!

Second thought:
Maybe we can add it into the Release Notes. Users should move from `is_flash_attn_greater_or_equal_2_10` to `is_flash_attn_greater_or_equal("2_10")`

ArthurZucker:
@vasqu let's add BC unless we had a deprecation cycle

vasqu:
Check out #45061, mb

PR #42443, opened 2025-11-27
remove reference to TF models from docs

zucchini-nlp:
# What does this PR do?

As per title

### DIFF
--- a/src/transformers/models/albert/configuration_albert.py
+++ b/src/transformers/models/albert/configuration_albert.py
@@ -20,7 +20,7 @@ from ...configuration_utils import PreTrainedConfig
 
 class AlbertConfig(PreTrainedConfig):
     r"""
-    This is the configuration class to store the configuration of a [`AlbertModel`] . It is used
+    This is the configuration class to store the configuration of a [`AlbertModel`]. It is used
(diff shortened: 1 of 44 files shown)


src/transformers/models/albert/configuration_albert.py
```diff
@@ -20,7 +20,7 @@
 
 class AlbertConfig(PreTrainedConfig):
     r"""
-    This is the configuration class to store the configuration of a [`AlbertModel`] or a [`TFAlbertModel`]. It is used
+    This is the configuration class to store the configuration of a [`AlbertModel`] . It is used
```
vasqu:
```suggestion
    This is the configuration class to store the configuration of a [`AlbertModel`]. It is used
```
Super big nit, seeing this elsewhere too

vasqu (approved):
Lgtm, just my nit about the space

src/transformers/models/albert/configuration_albert.py
Rocketknight1: Spaces before punctuation are a sign that a French person has been here :fr:

Rocketknight1 (approved):
Nice cleanup, I totally missed these!

PR #42444, opened 2025-11-27
[Trainer] use output.loss when using liger-kernel

### LINKED ISSUE #42414: `use_liger_kernel` is not compatible with sequence parallel
### System Info

- `transformers` version: 4.57.3 (cherry-pick commit `7e0ea6997411f2633712cec5c475b791efe69785`)
- Platform: Linux-5.4.203-1-tlinux4-0011.spr.0001-x86_64-with-glibc2.38
- Python version: 3.10.19
- Safetensors version: 0.5.3
- Accelerate version: 1.12.0
- DeepSpeed version: 0.18.2
- PyTorch version (accelerator?): 2.6.0+cu124 (CUDA)
- Jax version: not installed
- Using GPU in script?: yes
- GPU type: NVIDIA H800

Accelerate config:

```yaml
compute_environment: LOCAL_MACHINE
debug: false
deepspeed_config:
  deepspeed_hostfile: accelerate_configs/hostfile
  deepspeed_multinode_launcher: pdsh
  gradient_clipping: auto
  offload_optimizer_device: none
  offload_param_device: none
  zero3_init_flag: true
  zero3_save_16bit_model: true
  zero_stage: 3
distributed_type: DEEPSPEED
downcast_bf16: 'no'
enable_cpu_affinity: false
machine_rank: 0
main_process_ip: x.x.x.x
main_process_port: 8005
main_training_function: main
mixed_precision: 'no'
num_machines: 2
num_processes: 16
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
parallelism_config:
  parallelism_config_sp_size: 4
  parallelism_config_dp_replicate_size: 4
  parallelism_config_sp_backend: deepspeed
  parallelism_config_sp_seq_length_is_variable: true
  parallelism_config_sp_attn_implementation: flash_attention_2
```

### Who can help?

@SunMarc 

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

When using trl `SFTTrainer` with `--use_liger_kernel true`, the training loop is broken. The `outputs.logits` is None and `ForCausalLMLoss` method call `logits = logits.float()` which breaks it.

https://github.com/huggingface/transformers/blob/c12dfddba94c04fb6a5f5be4be879af51e354e2f/src/transformers/trainer.py#L3950-L3955

https://github.com/huggingface/transformers/blob/c12dfddba94c04fb6a5f5be4be879af51e354e2f/src/transformers/loss/loss_utils.py#L54

### Expected behavior

Deepspeed SP can work together with liger kernel enabled.

kashif:
# What does this PR do?

Handle loss computation for models using Liger-kernel. 

Fixes #42414

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -3925,7 +3925,7 @@ class Trainer:
 
     def _deepspeed_sp_compute_loss(self, model, inputs, return_outputs, pc):
         """
-        How the loss is computed by Trainer under sequence parallelism with sp_backend=="deepspeed" and sp_size>1.
+        How the loss is computed by the Trainer under sequence parallelism with sp_backend=="deepspeed" and sp_size>1.
         Performs weighted loss aggregation across SP ranks, accounting for varying numbers of valid tokens per rank
         (e.g., when some ranks receive only padding or prompt tokens that are masked with -100).
 
@@ -3945,18 +3945,14 @@ class Trainer:
 
         unwrapped_model = self.accelerator.unwrap_model(model)
 
+        # DeepSpeed SP automatically injects shift_labels into inputs (pre-shifted labels for SP).
+        # The model's forward pass receives shift_labels via **kwargs and passes it to the loss function.
+        # Both standard transformer models and Liger-patched models handle shift_labels correctly,
+        # so we can directly use the computed loss from the model output.
+        # See: https://huggingface.co/docs/accelerate/en/concept_guides/sequence_parallelism
         outputs = model(**inputs)
         shift_labels = inputs["shift_labels"]
-        # When using Liger-kernel, the model already computed the loss and outputs.logits is None
-        if outputs.loss is not None:
-            loss = outputs.loss
-        else:
-            loss = unwrapped_model.loss_function(
-                logits=outputs.logits,
-                labels=None,
-                shift_labels=shift_labels,
-                vocab_size=unwrapped_model.config.vocab_size,
-            )
+        loss = outputs.loss
 
         sp_group = self.accelerator.torch_device_mesh["sp"].get_group()
         sp_world_size = pc.sp_size
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -3943,8 +3943,6 @@ class Trainer:
             The loss of the model along with its output if return_outputs was set to True
         """
 
-        unwrapped_model = self.accelerator.unwrap_model(model)
-
         # DeepSpeed SP automatically injects shift_labels into inputs (pre-shifted labels for SP).
         # The model's forward pass receives shift_labels via **kwargs and passes it to the loss function.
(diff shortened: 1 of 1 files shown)


src/transformers/trainer.py
```diff
@@ -3947,12 +3947,16 @@ def _deepspeed_sp_compute_loss(self, model, inputs, return_outputs, pc):
...
         shift_labels = inputs["shift_labels"]
-        loss = unwrapped_model.loss_function(
-            logits=outputs.logits,
-            labels=None,
-            shift_labels=shift_labels,
-            vocab_size=unwrapped_model.config.vocab_size,
-        )
+        # When using Liger-kernel, the model already computed the loss and outputs.logits is None
+        if outputs.loss is not None:
+            loss = outputs.loss
```
SunMarc: For all models, the loss will be computed unfortunately. So this check will actually already be true. Maybe do a check we are using liger kernel simply. Also, can you also add a comment regarding `shift_labels` that deepspeed automatically inject this key in our inputs, so this is why we need to recompute the loss but for liger kernels, it is not needed as everything is handled. https://huggingface.co/docs/accelerate/en/concept_guides/sequence_parallelism#:~:text=If%20you%20use,tensor%20in%20memory.

Also, we might not even need to do `unwrapped_model.loss_function` as I think the loss should be computed correctly as we have `shift_labels` in inputs. So the fix could potentially just to remove that

SunMarc (commented):
Thanks, left a few ideas to fix this !

src/transformers/trainer.py
kashif: ok yes let me fix

kashif: can you kindly check now

kashif:
@SunMarc  i'll update the accelerate docs based on this change when you approve

SunMarc (approved):
Much cleaner thanks !

zhangwj618:
With the latest code and without liger kernel, I ran into an error caused by outputs.loss being None, as models (e.g. Qwen3ForCausalLM) won't calculate loss if labels is None. @kashif @SunMarc

SunMarc:
> With the latest code and without liger kernel, I ran into an error caused by outputs.loss being None, as models (e.g. Qwen3ForCausalLM) won't calculate loss if labels is None. @kashif @SunMarc

How come there is no labels ?

zhangwj618:
@SunMarc [UlyssesSPDataLoaderAdapter](https://github.com/deepspeedai/DeepSpeed/blob/6eb98aaf03419d1a3a608b0955eb3eae5ad2dc27/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L598) took labels out from inputs while inserting shift_labels

SunMarc:
Indeed @zhangwj618  ... thanks for spotting this ! I will open a PR for a quick fix

PR #42448, opened 2025-11-27
[TP plans] Fix some incorrects TP plans

Cyrilvallez:
# What does this PR do?

As per the title. See [here](https://github.com/huggingface/transformers/pull/41634#discussion_r2565560762) for the source

TODO CYRIL: models modified only apply norm on head_dim!! So should be fine without replication. But then nano_chat needs to remove useless replication, and let's check other models with norms to be sure

cc @vasqu


src/transformers/models/doge/modular_doge.py
```diff
@@ -146,11 +146,6 @@ class DogeConfig(PreTrainedConfig):
         "layers.*.self_attn.v_proj": "colwise",
         "layers.*.self_attn.dt_proj": "rowwise",
         "layers.*.self_attn.o_proj": "rowwise",
-        "layers.*.input_layernorm.weight": "sequence_parallel",
-        "layers.*.input_residual": "sequence_parallel",
-        "layers.*.post_attention_layernorm.weight": "sequence_parallel",
-        "layers.*.post_attention_residual": "sequence_parallel",
-        "norm.weight": "sequence_parallel",
```
vasqu, line 153: SP is super slow so removed it while I was at it

vasqu:
Only a bit weird one is `Cohere`, it has this `use_qk_norm` attribute but it defaults to false and the base models also doesn't use it. If it were to use it, we would need to rep, otherwise not. See https://github.com/huggingface/transformers/blob/352a2e0c120370f7db3b40caa4c60f0efca4e737/src/transformers/models/cohere/modular_cohere.py#L139-L146

Not sure if we just rep either way but left it for now

Cyrilvallez:
Thanks a lot for double checking @vasqu! For cohere, if the main checkpoints do not use the norm, let's not replicate as it's a huge performance drawback! 
Cannot approve my own PR, so just gonna merge it!

PR #42450, opened 2025-11-27
Future annotations break FastAPI

LysandreJik:
Unfortunately, `from __future__ import annotations` breaks pydantic/FastAPI -> everything results in "422: unprocessable entity".

The following is therefore not possible: `"PreTrainedTokenizerFast" | None`, reverting to an `Optional`.

### DIFF
--- a/benchmark_v2/framework/benchmark_runner.py
+++ b/benchmark_v2/framework/benchmark_runner.py
@@ -10,7 +10,6 @@ from datetime import datetime
 from queue import Queue
 from typing import Any
 
-import numpy as np
 import torch
 from datasets import Dataset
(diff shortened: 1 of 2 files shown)


benchmark_v2/framework/benchmark_runner.py
```diff
@@ -10,7 +10,6 @@
 from queue import Queue
 from typing import Any
 
-import numpy as np
```
LysandreJik, line 13: Issue on `main`

Wauplin (approved):
Too bad 😬 We did not introduce future annotations in huggingface_hub either

PR #42452, opened 2025-11-27
examples(android): add FastImageProcessor to avoid ImageProcessor ove…

CodersAcademy006:
This PR fixes a slowdown we hit when using TensorFlow Lite Task’s `ImageProcessor` on full-HD FP32 images (for example, **1920×1080×3**). On large images the ImageProcessor pipeline allocates temporary objects, performs multiple buffer copies, and runs per-pixel work in Java. On our test device that added roughly **150–180 ms** of overhead per frame, pushing total frame time to about **200 ms**, even though the actual model/QNN inference was only **~90 ms**.

To address this I added a small, focused helper called **FastImageProcessor** that skips ImageProcessor for big FP32 frames and does lightweight, reusable preprocessing/postprocessing instead:

- Allocate a direct, native-order `ByteBuffer` / `FloatBuffer` once at construction and reuse it every frame.
- Read pixels using `bitmap.getPixels(IntArray)` once per frame, then convert the IntArray → normalized floats in a tight loop (RGB order, normalized to `[0,1]`).
- Postprocess model output by converting floats → an `IntArray` and calling `bitmap.setPixels(...)` once.
- Reuse buffers and a single output `Bitmap` to avoid per-frame allocations and GC pressure.

In practice this reduces preprocess + postprocess time substantially (we see it drop into the tens of ms on the tested hardware), so the pipeline becomes model-bound again instead of CPU-bound.

### Files added
- `FastImageProcessor.kt` — the small reusable helper class for FP32 pipelines  
- `README.md` — short usage notes and testing instructions

### How to test
Replace the current ImageProcessor usage with `FastImageProcessor` (create one instance, reuse per frame), run your FP32 model (e.g., a super-resolution test), and measure: preprocess → model → postprocess. You should see a large reduction in CPU time and allocations.

**Fixes:** #105217

### DIFF
--- a/examples/android/fast_image_processing/FastImageProcessor.kt
+++ b/examples/android/fast_image_processing/FastImageProcessor.kt
@@ -1,0 +1,111 @@
+package examples.android.fast_image_processing
+
+import android.graphics.Bitmap
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+import java.nio.FloatBuffer
+
+/**
+ * FastImageProcessor — small helper to preprocess and postprocess full-HD FP32 images
+ * without using TensorFlow Lite Task's ImageProcessor (which allocates per call and is
+ * expensive for ~2M pixels). This keeps reusable direct ByteBuffer/FloatBuffer,
+ * IntArray buffers and Bitmaps to avoid per-frame allocation and GC.
+ *
+ * Usage:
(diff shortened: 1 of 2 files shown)


CodersAcademy006:
Got tensorflow in this, my bad.

PR #42453, opened 2025-11-27
Add SDPA and FlashAttention support to T5

### LINKED ISSUE #26350: Community contribution: Adding Flash Attention 2 support for more architectures
### Feature request

Flash Attention 2 is a library that provides attention operation kernels for faster and more memory efficient inference and training: https://github.com/Dao-AILab/flash-attention

Screenshot 2023-09-22 at 17 49 18

Let's try to add Flash Attention 2 support for more architectures! Currently supported architectures are

- [x] Llama
- [x] Falcon

It would be great to add the support for more architectures such as

- [x] Bark
- [x] Bart
- [ ] BERT | @sorenmc 
- [x] DistilBERT
- [ ] GPT-2
- [ ] GPT-J
- [x] GPTBigCode (Starcoder) | @susnato 
- [x] GPT-neo
- [x] GPT-neo-x | @younesbelkada #26463
- [x] OPT | @susnato #26414 
- [x] Llava
- [x] VipLlava
- [x] mBART
- [x] Mistral
- [x] Mixtral
- [ ] MPT | @rajveer43 
- [ ] T5
- [ ] Persimmon | @jeromeku 
- [x] Phi
- [x] Whisper
- [x] Qwen2

... and many more

Adding this feature would require to follow the same protocol as in https://github.com/huggingface/transformers/pull/25598 
 . First create a new module inside the corresponding modeling file termed as `xxxFlashAttention` that inherits from `xxxAttention` and override the foward method to use the public methods from `flash-attn`. Make sure to have access to a GPU that supports Flash Attention 2. 

Given the slight challenge of the issue, labelling it as a good second issue!

If you are interested to take up the challenge, comment below with the architecture name you want to integrate and open a PR!

Once you open a PR, feel free to ping @LysandreJik @ArthurZucker @amyeroberts @younesbelkada @fxmarty @SunMarc @pacman100 for a review

### Motivation

Making LLMs more memory efficient and faster ! 

### Your contribution

Reviewing PRs and possibly adding the support for more models

DuyguA:
I made some changes to the T5 modeling file to support new attention interface. I made a bit of rearrangements to employ `position_bias` correctly into the attention mask.

Fixes #26350

A note though, I made a `make fix-copies` , however it broke several related models such as `longt5` and `mt5`. Somehow fix script didn't copy over the imports, couldn't grab the attention code correctly hence I skipped that part. If applicable we can merge this PR + I can work on related models in another PR or I'm happy to take some hints to make the script work properly.

## Before submitting
- [ x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ x] Did you make sure to update the documentation with your changes? Here are the
- [ x] Did you write any new necessary tests?

@ArthurZucker @Cyrilvallez @vasqu


vasqu (commented):
Sorry to be so strict about this but T5 is not a good candidate for flash attention / sdpa. The reason is that the relative attention bias has to be modeled there and as of now, it's not possible with base flash attention (might be possible with sdpa but needs proper mask preparation). tl;dr: It will only support eager attention in the end

We can still refactor this to have the attention interface-like implementation but only for eager in the end (i.e. `_supports_sdpa/flash_attn` remain False). Wdyt?

DuyguA:
> Sorry to be so strict about this but T5 is not a good candidate for flash attention / sdpa. The reason is that the relative attention bias has to be modeled there and as of now, it's not possible with base flash attention (might be possible with sdpa but needs proper mask preparation). tl;dr: It will only support eager attention in the end
> 
> We can still refactor this to have the attention interface-like implementation but only for eager in the end (i.e. `_supports_sdpa/flash_attn` remain False). Wdyt?

Sounds reasonable  to me!

DuyguA:
Heys again @vasqu  , I made the changes for restricting only eager attention. Model tests are passing, only repo consistency checks fail as I mentioned above. PR is ready for merge 😊

src/transformers/models/t5/modeling_t5.py
```diff
@@ -157,6 +161,35 @@ def forward(self, hidden_states):
         return hidden_states
 
 
+def eager_attention_forward(
```
vasqu, line 166: I would rather have the relative position bias within here, see #38301 or more specifically https://github.com/huggingface/transformers/blob/1c3188fb648c4fa8065356d512ffd7d8ba119ed8/src/transformers/models/bert/modeling_bert.py#L121-L176 (no longer on main but should give you the idea how this should look like)

src/transformers/models/t5/modeling_t5.py
```diff
@@ -182,6 +216,8 @@ def __init__(
                 "when creating this class."
             )
 
+        self.scaling = self.d_model**-0.5
```
vasqu: For completeness, we should have the `is_causal` flag here, you can look into Bart for this - i.e. encoder = False, decoder = False if self attn or True if cross attn.

src/transformers/models/t5/modeling_t5.py
```diff
@@ -319,39 +355,46 @@ def forward(
...
-        scores += position_bias_masked
-
-        # (batch_size, n_heads, seq_length, key_length)
-        attn_weights = nn.functional.softmax(scores.float(), dim=-1).type_as(scores)
-        attn_weights = nn.functional.dropout(attn_weights, p=self.dropout, training=self.training)
+        attention_interface: Callable = eager_attention_forward
+        if self.config._attn_implementation != "eager":
+            logger.warning_once(
+                "T5 uses relative position bias; SDPA/FlashAttention not supported, fall back to eager."
+            )
```
vasqu: This should never happen as we don't support anything other than eager. I would even raise an error here if anything

src/transformers/models/t5/modeling_t5.py
```diff
@@ -262,16 +298,16 @@ def compute_bias(self, query_length, key_length, device=None, cache_position=Non
...
+        hidden_states: torch.FloatTensor,
+        key_value_states: Optional[torch.FloatTensor] = None,
+        past_key_values: Optional[torch.FloatTensor] = None,
+        attention_mask: Optional[torch.FloatTensor] = None,
+        position_bias: Optional[torch.FloatTensor] = None,
+        query_length: Optional[torch.LongTensor] = None,
+        output_attentions: Optional[bool] = False,
+        cache_position: Optional[torch.LongTensor] = None,
+        **kwargs: Unpack[FlashAttentionKwargs],
+    ) -> tuple[torch.Tensor, Optional[torch.Tensor], Optional[tuple[torch.Tensor]]]:
```
vasqu: Let's not rename here, this would break BC. The type annotations are fine by itself.

src/transformers/models/t5/modeling_t5.py
```diff
@@ -547,6 +590,10 @@ class T5PreTrainedModel(PreTrainedModel):
     config: T5Config
     base_model_prefix = "transformer"
     supports_gradient_checkpointing = True
+    _supports_attention_backend = True
```
vasqu, line 586: Not supported - kwargs are not used everywhere so far and enc-dec will need another look

src/transformers/models/t5/modeling_t5.py
```diff
@@ -319,39 +355,46 @@ def forward(
...
+            output_attentions=output_attentions,
+            **kwargs,
+        )
 
-        attn_output = attn_output.transpose(1, 2).contiguous()
-        attn_output = attn_output.view(batch_size, -1, self.inner_dim)
+        attn_output = attn_output.view(batch_size, -1, self.inner_dim).contiguous()
         attn_output = self.o(attn_output)
 
         outputs = (attn_output, position_bias)
```
vasqu: Would be nice if we could refactor this along in this PR, we have an `Outputrecorder` which can handle collecting the weights. We no longer need to explicitly have the kwargs then. You can take a look at other model like Llama or t5gemma2 which do this. In essence, you need decorators (check_model_input, can_return_tuple) and the respective flag `_can_record_outputs`.

vasqu (commented):
Some initial comments. Would be nice if we could go further to include the recorder and avoid unnecessary code along `output_xxx`.

src/transformers/models/t5/modeling_t5.py
DuyguA: Fair, I made the changes 😊

src/transformers/models/t5/modeling_t5.py
DuyguA: Added!

src/transformers/models/t5/modeling_t5.py
DuyguA: Changed it into an error!

src/transformers/models/t5/modeling_t5.py
DuyguA: OK, went back to the old name `mask`

src/transformers/models/t5/modeling_t5.py
DuyguA: I replaced more `**kwargs` to the places where attention is called.

src/transformers/models/t5/modeling_t5.py
DuyguA: Why not, I added `OutputRecorder` to pick up the weights along `_can_record_outputs`.

I decorated  `T5Block.forward` with `check_model_inputs`, I think strategically best place is there.

`can_return_tuple`s are added too!

DuyguA:
Heys @vasqu  , thanks for your detailed review and suggestions. I made the changes, please have a newer look 😊  I also run some rounds of `T5ForConditionalGeneration.generate` on CPU and GPU with `t5-small` and `t5-base` to double check the functionality. I examined encoder outputs separately again to check attention implementation, all looks good.

src/transformers/models/t5/modeling_t5.py
```diff
@@ -157,15 +163,110 @@ def forward(self, hidden_states):
...
+
 class T5Attention(nn.Module):
     def __init__(
         self,
         config: T5Config,
         has_relative_attention_bias=False,
         layer_idx: Optional[int] = None,
+        is_causal: bool = False,
     ):
         super().__init__()
```
vasqu, line 203: Let's add a `self.scale` even if it is 1, i.e. the factor we scale `qk^T` before applying softmax

src/transformers/models/t5/modeling_t5.py
```diff
@@ -192,96 +293,22 @@ def __init__(
 
         self.gradient_checkpointing = False
 
-    @staticmethod
-    def _relative_position_bucket(relative_position, bidirectional=True, num_buckets=32, max_distance=128):
```
vasqu: We can still keep this, we have access to `self` via `module` in the eager forward

src/transformers/models/t5/modeling_t5.py
```diff
@@ -192,96 +293,22 @@ def __init__(
...
-        """
-        Self-attention (if key_value_states is None) or attention over source sentence (provided by key_value_states).
-        """
-        # Input is (batch_size, seq_length, dim)
-        # Mask is (batch_size, 1, 1, key_length) (non-causal encoder) or (batch_size, 1, seq_length, key_length) (causal decoder)
+        hidden_states: torch.FloatTensor,
+        mask: Optional[torch.FloatTensor] = None,
+        key_value_states: Optional[torch.FloatTensor] = None,
+        past_key_values: Optional[torch.FloatTensor] = None,
+        position_bias: Optional[torch.FloatTensor] = None,
```
vasqu:
```suggestion
        position_bias: Optional[torch.FloatTensor] = None,
        past_key_values: Optional[torch.FloatTensor] = None,
```
May seem trivial but changing the order is breaking itself

src/transformers/models/t5/modeling_t5.py
```diff
@@ -192,96 +293,22 @@ def __init__(
...
+        query_length: Optional[torch.LongTensor] = None,
+        output_attentions: Optional[bool] = False,
```
vasqu: Let's use `check_mode_inputs` instead, you can take look into `t5gemma2` for this. The key is to properly define `_can_record_outputs`

src/transformers/models/t5/modeling_t5.py
```diff
@@ -318,44 +344,56 @@ def forward(
...
-            real_seq_length = query_length if query_length is not None else cache_position[-1] + 1
-            if not self.has_relative_attention_bias:
-                position_bias = torch.zeros(
-                    (1, self.n_heads, seq_length, key_length), device=scores.device, dtype=scores.dtype
-                )
-                if self.gradient_checkpointing and self.training:
-                    position_bias.requires_grad = True
+        # relative context position computations
+        relative_context_positions = None
+        if self.has_relative_attention_bias:
```
vasqu: I would move this logic to the eager forward

src/transformers/models/t5/modeling_t5.py
```diff
@@ -318,44 +344,56 @@ def forward(
...
-        attn_weights = nn.functional.softmax(scores.float(), dim=-1).type_as(scores)
-        attn_weights = nn.functional.dropout(attn_weights, p=self.dropout, training=self.training)
-
-        attn_output = torch.matmul(attn_weights, value_states)
+        attention_interface: Callable = eager_attention_forward
+        if self.config._attn_implementation != "eager":
+            raise ValueError(
+                "T5Attention only supports attn_implementation='eager' due to relative position bias."
+                "Got: "
+                f"{self.config._attn_implementation} instead. "
```
vasqu:
```suggestion
                f"Got `{self.config._attn_implementation}` instead. "
```

src/transformers/models/t5/modeling_t5.py
```diff
@@ -318,44 +344,56 @@ def forward(
...
+            relative_attention_max_distance=self.relative_attention_max_distance,
+            seq_length=seq_length,
+            relative_context_positions=relative_context_positions,
+            output_attentions=output_attentions,
+            **kwargs,
+        )
 
-        attn_output = attn_output.transpose(1, 2).contiguous()
-        attn_output = attn_output.view(batch_size, -1, self.inner_dim)
+        attn_output = attn_output.view(batch_size, -1, self.inner_dim).contiguous()
```
vasqu:
```suggestion
        attn_output = attn_output.view(batch_size, seq_length, -1).contiguous()
```
Can you follow Bart more closely? Same for the view operations after the projection

src/transformers/models/t5/modeling_t5.py
```diff
@@ -547,11 +595,21 @@ class T5PreTrainedModel(PreTrainedModel):
...
     _can_compile_fullgraph = True
 
     _no_split_modules = ["T5Block"]
     _keep_in_fp32_modules = ["wo"]
 
+    _can_record_outputs = {
+        "hidden_states": [OutputRecorder(T5Block, index=0, layer_name="block")],
+        "attentions": [OutputRecorder(T5LayerSelfAttention, index=-1, layer_name="self_attn")],
+        "cross_attentions": [OutputRecorder(T5LayerCrossAttention, index=-1, layer_name="cross_attn")],
+    }
```
vasqu, line 599: You will need to properly remove all `output_xxx` for everything, see T5Gemma2. For different parts of the model it will likely need to be overriden

src/transformers/models/t5/modeling_t5.py
vasqu: Sorry I have to revert back a bit. I initially thought it would be like Bert but T5 integrates its bias directly into the mask. 

Imo, we can directly use the eager forward of e.g. Bert and before calculate the bias as is. SDPA should also be supportable this ways. So 
- `eager` from Bert (current)
- Calculate bias as before
  - If we have SDPA, we have boolean mask which we need to convert see https://github.com/huggingface/transformers/blob/5b710c75426639aff3348589fd9095b5f229cb87/src/transformers/masking_utils.py#L529-L531
- The forward of the normal attention can follow Bart closely except we calculate and add the bias to the mask.

src/transformers/models/t5/modeling_t5.py
```diff
@@ -318,44 +344,56 @@ def forward(
...
-        position_bias_masked = position_bias
-        scores += position_bias_masked
-
-        # (batch_size, n_heads, seq_length, key_length)
-        attn_weights = nn.functional.softmax(scores.float(), dim=-1).type_as(scores)
-        attn_weights = nn.functional.dropout(attn_weights, p=self.dropout, training=self.training)
-
-        attn_output = torch.matmul(attn_weights, value_states)
+        attention_interface: Callable = eager_attention_forward
+        if self.config._attn_implementation != "eager":
```
vasqu, line 389: We might be able to use SDPA, see my other comment

src/transformers/models/t5/modeling_t5.py
```diff
@@ -318,44 +344,56 @@ def forward(
...
+            has_relative_attention_bias=self.has_relative_attention_bias,
+            is_decoder=self.is_decoder,
+            relative_attention_bias=getattr(self, "relative_attention_bias", None),
+            relative_attention_num_buckets=self.relative_attention_num_buckets,
+            relative_attention_max_distance=self.relative_attention_max_distance,
+            seq_length=seq_length,
+            relative_context_positions=relative_context_positions,
+            output_attentions=output_attentions,
+            **kwargs,
+        )
```
vasqu: Sorry about all this, let's revert and try to integrate the bias into the mask --> base interface call

src/transformers/models/t5/modeling_t5.py
```diff
@@ -490,6 +537,7 @@ def forward(
                 query_length=cache_position[-1] + 1,
                 use_cache=use_cache,
                 output_attentions=output_attentions,
+                **kwargs,
```
vasqu, line 535: Need to type annotate all of them as well with `Unpack[TransformersKwargs]`

src/transformers/models/t5/modeling_t5.py
```diff
@@ -547,11 +595,21 @@ class T5PreTrainedModel(PreTrainedModel):
     config: T5Config
     base_model_prefix = "transformer"
     supports_gradient_checkpointing = True
+    _supports_attention_backend = True
+    _supports_flash_attn = False
+    _supports_sdpa = False
```
vasqu: See my other comments, since the bias is directly added to the mask, we can do the same for SDPA (might need a bool->float conversion tho)

vasqu (commented):
Sorry about this but after taking a closer look then it seems that T5 directly integrates the relative position bias into the mask. We can make use of that to support sdpa as well! We will possibly need to convert the bool to float mask

The implementation should then look close to Bart with the exception that we add the position bias to the mask before calling the interface. If the position bias is given, then that acts as a mask directly it seems (to double check).

Make sure to run integration tests that it works as expected, e.g. `RUN_SLOW=1 pytest tests/models/t5/test_modeling_t5.py -k "integration"`

vasqu:
run-slow: t5

vasqu:
@DuyguA I've refactored myself because it involves quite a few things and I also had to backpaddle a bit on what I said before. Now everything works for T5 (+ it supports SDPA). However, we need to now fix the other broken tests that relied on the code of T5 by copying from it or using it in some other manner.

I will leave it for now. Would be nice if you could continue from here or I pick it up at some other time. It should at least provide a good basis

DuyguA:
> @DuyguA I've refactored myself because it involves quite a few things and I also had to backpaddle a bit on what I said before. Now everything works for T5 (+ it supports SDPA). However, we need to now fix the other broken tests that relied on the code of T5 by copying from it or using it in some other manner.
> 
> I will leave it for now. Would be nice if you could continue from here or I pick it up at some other time. It should at least provide a good basis

Great, thanks @vasqu . I'll take it from here, hope to finish in couple of days.

PR #42465, opened 2025-11-27
Add SDPA support for PatchTST model

Furkan-rgb:
## What does this PR do?

Adds SDPA (Scaled Dot Product Attention) support for the PatchTST model.

### Changes:
- Added `PatchTSTSdpaAttention` class using `torch.nn.functional.scaled_dot_product_attention`
- Integrated attention class selection in `PatchTSTEncoderLayer` based on `config._attn_implementation`
- Added `_supports_sdpa = True` to `PatchTSTPreTrainedModel`
- Fixed `test_modeling_patchtst.py` `_prepare_for_class` method for proper dynamic batch size handling

### Testing:
All 67 PatchTST tests pass (103 skipped as expected):
```
pytest tests/models/patchtst/test_modeling_patchtst.py
```

### Notes:
- The SDPA implementation falls back to eager attention when `output_attentions=True` since SDPA doesn't return attention weights
- Uses the standard attention implementation pattern from other models in transformers


kashif (approved):
looks good thanks

Furkan-rgb:
Seems like the documentation build is failing due to improperly closed </div> tag during the build. Unrelated to my changes.

vasqu (commented):
Sorry, we no longer support implementing seperate Attention classes for each attention flavor. There are some older models which haven't been refactored yet for this case. 

Please take a look at models like Albert or Bert (a bit messy because it has additional enc-dec logic) which already implement this. The essence is to have 
- One Attention class
- Specific attributes like `is_causal`, `scaling`, `num_attention_heads` to directly reuse across flavors
- Generic code around everything around the core attn mechanism, e.g. the projections and views/reshaping
- The interface (`ALL_ATTENTION_FUNCTIONS`) that calls the underlying flavor for us

Furkan-rgb:
> Sorry, we no longer support implementing seperate Attention classes for each attention flavor. There are some older models which haven't been refactored yet for this case.
> 
> Please take a look at models like Albert or Bert (a bit messy because it has additional enc-dec logic) which already implement this. The essence is to have
> 
> * One Attention class
> * Specific attributes like `is_causal`, `scaling`, `num_attention_heads` to directly reuse across flavors
> * Generic code around everything around the core attn mechanism, e.g. the projections and views/reshaping
> * The interface (`ALL_ATTENTION_FUNCTIONS`) that calls the underlying flavor for us

Thanks for the feedback. Took a look and updated accordingly. Much cleaner now. The existing PatchTSTAttention already uses ALL_ATTENTION_FUNCTIONS.

kashif:
@Furkan-rgb I fixed the SLOW tests and explicitly testing with sdpa

vasqu:
run-slow: patchtst

vasqu:
run-slow: patchtst

vasqu:
run-slow: patchtst

src/transformers/models/patchtst/modeling_patchtst.py
```diff
@@ -571,7 +575,15 @@ def _init_weights(self, module: nn.Module):
...
-            init.copy_(module.position_enc, module._init_pe(self.config, num_patches))
+            position_enc = module._init_pe(self.config, num_patches)
+            if is_deepspeed_zero3_enabled():
+                import deepspeed
+
+                with deepspeed.zero.GatheredParameters(module.position_enc, modifier_rank=None):
+                    if module.position_enc.numel() > 0:
+                        init.copy_(module.position_enc, position_enc)
+            else:
+                init.copy_(module.position_enc, position_enc)
```
vasqu, line 586: Not exactly against this but why was this added to this PR? Can we move this to a separate PR? cc @kashif

vasqu (approved):
Just a small comment about the deepspeed addition

I was free to update some more to add flex attn and enable some other tests. SDPA is enabled by default, so no need to set the attn implementation here. We should probably update the title tho ~ something along `Update supported atttns for PatchTST`

src/transformers/models/patchtst/modeling_patchtst.py
kashif: the slow test were failing without this change...

vasqu: Gotcha, thx for adding then

I will probably merge this PR tomorrow then, we cutting last PRs for v5 rn

vasqu:
Sorry for the delays, merged now :hugs:

PR #42456, opened 2025-11-27
Fix MoE for V5

3outeille:

PR #42468, opened 2025-11-28
Fix broken models due to modeling/weight mapping

Cyrilvallez:
# What does this PR do?

As per the title. A few models got broken in https://github.com/huggingface/transformers/pull/41580, but it was silent issues on the fast tests.
Mostly discovered in https://github.com/huggingface/transformers/pull/42396, because the new tests added were not passing for those models, due to the models and not the logic of that PR.

### DIFF
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -322,15 +322,15 @@ class WeightTransform:
             # This is ugly but needed for reverse mapping of Qwen2.5!
             if r"(?!\.(language_model|visual))" in pattern:
                 pattern = pattern.replace(r"(?!\.(language_model|visual))", "")
-            # Allow capturing groups in patterns, i.e. to add a prefix to all keys (e.g. timm_wrapper)
+            # Allow capturing groups in patterns, i.e. to add/remove a prefix to all keys (e.g. timm_wrapper, sam3)
             if r"(.+)" in pattern:
-                pattern = pattern.replace(r"(.+)", "")
+                pattern = pattern.replace(r"(.+)", r"\1")
             self.target_patterns[i] = pattern
 
-        # We also need to check capturing groups in the sources during reverse mapping (e.g. timm_wrapper)
+        # We also need to check capturing groups in the sources during reverse mapping (e.g. timm_wrapper, sam3)
         for i, pattern in enumerate(self.source_patterns):
             if r"\1" in pattern:
-                pattern = pattern.replace(r"\1", "")
+                pattern = pattern.replace(r"\1", r"(.+)")
             self.source_patterns[i] = pattern
 
     def add_tensor(self, target_key: str, source_key: str, source_pattern: str, future: Future):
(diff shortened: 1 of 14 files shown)


src/transformers/core_model_loading.py
```diff
@@ -647,17 +650,16 @@ def rename_source_key(
     Rename a source key given all the renaming and weight conversion patterns we have. Also takes care of adding/removing
     the base model prefix during loading if necesary.
     """
-    # 1. apply all renamings
-    renamed_key = rename_alternation.sub(lambda m: repl(m, rename_by_group), source_key).replace("\\", "")
+    # 1. apply all renamings (we need to replace only the first match of the alternation if multiple matches, so count=1)
+    renamed_key = rename_alternation.sub(lambda m: repl(m, rename_by_group), source_key, count=1)
```
Cyrilvallez: Super important to have `count=1` here, otherwise the way `re.sub` works is that every time it finds a match, it replace, and then continues with the end of the regex... So it's basically a for loop over all patterns of the alternation replacing each one which is unwanted. 
The guard
```python
if len(matched_groups) > 1:
        raise ValueError(
            "only a single match should happen, your regex patterns are tangled: "
            f"groups matched = {matched_groups} for the patternsL {repl_map.keys()}"
        )
```
inside `repl` is in fact never reached because it would match only 1 pattern at a time, but do it several time potentially...
So the prefix addition of sam3 would be matched even if another of the branches was taken before which is unwanted

src/transformers/models/lfm2_moe/modeling_lfm2_moe.py
```diff
@@ -178,7 +176,7 @@ def forward(
             _, token_idx = torch.where(expert_mask[expert_idx])
             current_state = hidden_states[token_idx]
             gate, up = nn.functional.linear(current_state, self.gate_up_proj[expert_idx]).chunk(2, dim=-1)
-            current_hidden_states = self.act_fn(gate) * up
+            current_hidden_states = gate * up
```
ArthurZucker: that looks weird but okay

src/transformers/core_model_loading.py
ArthurZucker: as discussed offline let's iterate

3outeille:
actually needing this for https://github.com/huggingface/transformers/pull/42456 for jamba + lfm2

src/transformers/models/lfm2_moe/modeling_lfm2_moe.py
Cyrilvallez: Yeah went a bit too fast there was a hardcoded silu haha

src/transformers/models/sam3_tracker/modeling_sam3_tracker.py
```diff
@@ -768,7 +768,7 @@ class Sam3TrackerModel(Sam3TrackerPreTrainedModel):
         "occlusion_spatial_embedding_parameter",
     ]
     _checkpoint_conversion_mapping = {
-        "tracker_model.": "",
+        r"tracker_model.(.+)": r"\1",  # the regex allows to remove the prefix, and add it back in revert mode
```
ArthurZucker, line 771: good I like it!

ArthurZucker (approved):
is it missing a test that can help prevent this?

Cyrilvallez:
Yes will enforce it with ast later for sure! Along with the `post_init`

PR #42472, opened 2025-11-28
Grouped beam search from config params

zucchini-nlp:
# What does this PR do?

Ensures that we can fallback to hub-based group generation from generation args, even if the custom generate is not passed. And, I updated the version in deprecation cycle. I don't think we will break it for v5 yet

```python

from transformers import BartForConditionalGeneration, GenerationConfig
import torch

model_id = "hf-internal-testing/tiny-random-bart"
model = BartForConditionalGeneration.from_pretrained(model_id)
gen_config = GenerationConfig(
    num_beams=4,
    do_sample=False,
    max_new_tokens=10,
    min_new_tokens=10,
    num_beam_groups=2,
    diversity_penalty=0.0001,
    use_cache=True,
)

input_ids = torch.tensor([[model.config.bos_token_id]])
out = model.generate(input_ids, generation_config=gen_config, trust_remote_code=True)

```


zucchini-nlp:
Test failures not related

src/transformers/generation/utils.py
```diff
@@ -2240,7 +2240,7 @@ def _get_deprecated_gen_repo(
         logger.warning_once(
             f"{generation_mode.name.replace('_', ' ').title()} was moved to a `custom_generate` repo: https://hf.co/{repo}. "
             f"To prevent loss of backward compatibility, add `custom_generate='{repo}'` "
-            "to your `generate` call before v4.62.0."
+            "to your `generate` call before v5.2.0."
```
ArthurZucker: not sure I understand, do we patch 4.57?

ArthurZucker (approved):

src/transformers/generation/utils.py
zucchini-nlp: oh right, this commit doesn't go in patch! I will revert it

Meant to update the end-version in v5 branch and give users time to adapt to `custom_generate`

ArthurZucker (approved):

PR #42474, opened 2025-11-28
Fix a bug where compile_config gets deleted

remi-or:
Currently,  `to_dict` method of `GenerationConfig` always deletes the `compile_config` attribute. This is because when we call `save_pretrained` on a `GenerationConfig` object, we don't want to save the `CompileConfig` object.
But this create an issue where `compile_config` is deleted even when we load the model. For instance, this snippet:

```python
import torch
from transformers import AutoModelForCausalLM, CompileConfig, GenerationConfig

compile_config = CompileConfig(fullgraph=True)
generation_config = GenerationConfig(compile_config=compile_config)
model = AutoModelForCausalLM.from_pretrained("meta-llama/Llama-3.1-8B-Instruct", dtype=torch.bfloat16, generation_config=generation_config)

print(f"After model loading:\n{model.generation_config = }")
model.generation_config.save_pretrained("test.json")

reloaded_generation_config = GenerationConfig.from_pretrained("test.json")
print(f"After reloading:\n{reloaded_generation_config = }")

```

Running on `main` produces:
```
After model loading: model.generation_config = GenerationConfig {}
After reloading from file: reloaded_generation_config = GenerationConfig {}
```

Running after this PR:
```
After model loading: model.generation_config = GenerationConfig {
  "compile_config": {
    "backend": "inductor",
    "dynamic": null,
    "fullgraph": true,
    "mode": "reduce-overhead",
    "options": null
  }
}
After reloading from file: reloaded_generation_config = GenerationConfig {}
```

Thus we get the desired behavior of not saving the `compile_config` without deleting it as soon as we load the model.

### DIFF
--- a/src/transformers/generation/configuration_utils.py
+++ b/src/transformers/generation/configuration_utils.py
@@ -755,7 +755,7 @@ class GenerationConfig(PushToHubMixin):
 
         output_config_file = os.path.join(save_directory, config_file_name)
 
-        self.to_json_file(output_config_file, use_diff=True)
+        self.to_json_file(output_config_file, use_diff=True, pop_keys=["compile_config"])
         logger.info(f"Configuration saved in {output_config_file}")
 
         if push_to_hub:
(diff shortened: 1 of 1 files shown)


remi-or:
The fails in `tests_tokenization` and `tests_torch` seem unrelated. `test_processors` fails with the same error as on main.

ArthurZucker (approved):
`keys_to_pop` is a bit better

PR #42479, opened 2025-11-28
Fix minor bugs in ConversionOps and WeightConverter

SunMarc:
# What does this PR do?

This fixes a couple of minor details in ConversionOps and WeightConverter. We still need to pass `full_layer_name` as we won't be able to get the module otherwise. Also dim is not always defined, so the __repr__ was failing.

cc @Cyrilvallez for vis

### DIFF
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -105,7 +105,10 @@ class ConversionOps:
     """Base class for weight conversion operations."""
 
     def __repr__(self):
-        return f"{self.__class__.__name__}(dim={self.dim})"
+        if hasattr(self, "dim"):
+            return f"{self.__class__.__name__}(dim={self.dim})"
+        else:
+            return f"{self.__class__.__name__}"
 
     @abstractmethod
     def convert(
@@ -435,6 +438,7 @@ class WeightConverter(WeightTransform):
                     source_patterns=self.source_patterns,
                     target_patterns=self.target_patterns,
                     # Additional kwargs, ususally not used
+                    full_layer_name=layer_name,
                     model=model,
                     config=config,
                     missing_keys=missing_keys,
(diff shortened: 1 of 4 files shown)


MekkCyber (approved):
Thanks

src/transformers/core_model_loading.py
```diff
@@ -435,6 +438,7 @@ def convert(
                     source_patterns=self.source_patterns,
                     target_patterns=self.target_patterns,
                     # Additional kwargs, ususally not used
+                    full_layer_name=layer_name,
```
Cyrilvallez, line 441: Why do we need to add it back here? It's already passed for QuantizedOps, and should not be needed for base ops

Cyrilvallez (approved):
Nice thanks!

src/transformers/core_model_loading.py
MekkCyber: It's needed for the deserilization which is a conversion ops

SunMarc: We have some quantized ops in base ops also. Those are for loading quantized checkpoints. You can check `Bnb4bitDeserialize` for example. For the QuantizeOps, it is only used when we want to quantize a model e.g `Bnb4bitQuantize`

src/transformers/models/gpt2/tokenization_gpt2.py
```diff
@@ -101,7 +101,7 @@ def __init__(
         bos_token="<|endoftext|>",
         eos_token="<|endoftext|>",
         pad_token=None,
-        add_prefix_space=True,
+        add_prefix_space=False,
```
ArthurZucker, line 104: let's fix in a different PR please

ArthurZucker (approved):

src/transformers/models/gpt2/tokenization_gpt2.py
SunMarc: done

ArthurZucker (approved):

src/transformers/core_model_loading.py
Cyrilvallez: Oh ok, I thought that all quantized related were in the `quantization_ops`! Thanks for explaining!

PR #42483, opened 2025-11-28
remove tokenizer warning

SunMarc:
# What does this PR do?

This PR fixes a warning that shouldn't appear in a specific case when calling AutoTokenizer.from_pretrained.  `config_tokenizer_class` and `cls.__name__` might not the same due to the alias we created.

```
# Backward-compatible alias: allow referring to TokenizersBackend as PreTrainedTokenizerFast
PreTrainedTokenizerFast = TokenizersBackend
```

So `config_tokenizer_class` ends up being `PreTrainedTokenizerFast` and when checking the class name `cls.__name__` we get `TokenizersBackend` . 

The warning is not needed anymore, so let's remove it entirely.

## Reproduction

```python
>>> from transformers import AutoTokenizer
>>> tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-3.2-3B-Instruct")
```

Warning: 

```
The tokenizer class you load from this checkpoint is 'PreTrainedTokenizerFast'.
The class this function is called from is 'TokenizersBackend'.
```

### DIFF
--- a/src/transformers/tokenization_utils_base.py
+++ b/src/transformers/tokenization_utils_base.py
@@ -1904,7 +1904,10 @@ class PreTrainedTokenizerBase(PushToHubMixin):
                     config_tokenizer_class = TOKENIZER_MAPPING_NAMES.get(model_type)
 
         if config_tokenizer_class is not None:
-            if cls.__name__.replace("Fast", "") != config_tokenizer_class.replace("Fast", ""):
+            if config_tokenizer_class == "PreTrainedTokenizerFast" and cls.__name__== "TokenizersBackend":
+                # those two class are the same since we created an alias but the naming didn't change
+                pass
+            elif cls.__name__.replace("Fast", "") != config_tokenizer_class.replace("Fast", ""):
                 logger.warning(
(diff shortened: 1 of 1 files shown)


ArthurZucker (approved):
feel free to remove it entirely its no longer valid

src/transformers/tokenization_utils_base.py
```diff
@@ -1827,16 +1827,14 @@ def _from_pretrained(
         if tokenizer_config_file is not None:
             with open(tokenizer_config_file, encoding="utf-8") as tokenizer_config_handle:
                 init_kwargs = json.load(tokenizer_config_handle)
-            # First attempt. We get tokenizer_class from tokenizer_config to check mismatch between tokenizers.
-            config_tokenizer_class = init_kwargs.get("tokenizer_class")
+            # used in the past to check if the tokenizer class matches the class in the repo
             init_kwargs.pop("tokenizer_class", None)
```
SunMarc, line 1831: left a small comment here to understand why we pop `tokenizer_class`. Maybe we don't even need to pop it ?

PR #42484, opened 2025-11-28
more tts pipeline exampel

Deep-unlearning:
Following this: https://github.com/huggingface/transformers/pull/42473

Here are more example for TTS pipeline

@ebezzam @eustlb

### DIFF
--- a/docs/source/en/tasks/text-to-speech.md
+++ b/docs/source/en/tasks/text-to-speech.md
@@ -24,14 +24,13 @@ languages and for multiple speakers. Several text-to-speech models are currently
 
 You can easily generate audio using the `"text-to-audio"` pipeline (or its alias - `"text-to-speech"`). Some models, like Dia,
 can also be conditioned to generate non-verbal communications such as laughing, sighing and crying, or even add music.
-Here's an example of how you would use the `"text-to-speech"` pipeline with Dia:
+Here's an example of how you would use the `"text-to-speech"` pipeline with CSM:
 
-```py
+```python
 >>> from transformers import pipeline
 
->>> pipe = pipeline("text-to-speech", model="nari-labs/Dia-1.6B-0626")
->>> text = "[S1] (clears throat) Hello! How are you? [S2] I'm good, thanks! How about you?"
->>> output = pipe(text)
+>>> pipe = pipeline("text-to-audio", model="sesame/csm-1b")
+>>> output = pipe("Hello from Sesame.")
 ```
 
 Here's a code snippet you can use to listen to the resulting audio in a notebook:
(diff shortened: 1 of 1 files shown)


docs/source/en/tasks/text-to-speech.md
```diff
@@ -24,24 +24,68 @@ languages and for multiple speakers. Several text-to-speech models are currently
 
 You can easily generate audio using the `"text-to-audio"` pipeline (or its alias - `"text-to-speech"`). Some models, like Dia,
 can also be conditioned to generate non-verbal communications such as laughing, sighing and crying, or even add music.
-Here's an example of how you would use the `"text-to-speech"` pipeline with Dia:
+Here's an example of how you would use the `"text-to-speech"` pipeline with CSM:
```
vasqu: Isn't the order a bit confusing? Would leave one model or the other but not both.

I.e. `Some models, like Dia,...` ... `Here's an example ... with CSM`

docs/source/en/tasks/text-to-speech.md
```diff
@@ -24,24 +24,68 @@ languages and for multiple speakers. Several text-to-speech models are currently
...
+>>> pipe = pipeline("text-to-audio", model="sesame/csm-1b")
+>>> output = pipe("Hello from Sesame.")
+```
+
+Here's a code snippet you can use to listen to the resulting audio in a notebook:
+
+```python
+>>> from IPython.display import Audio
+>>> Audio(output["audio"], rate=output["sampling_rate"])
+```
```
vasqu, line 40: We have the same code below, would double check that we don't repeat ourselves

vasqu (commented):
Just some smaller comments that confused me :sweat_smile:

docs/source/en/tasks/text-to-speech.md
ebezzam: Agreed. I would remove the line just before to only mention CSM, and also add a link to CSM: https://huggingface.co/sesame/csm-1b

docs/source/en/tasks/text-to-speech.md
```diff
@@ -24,24 +24,68 @@ languages and for multiple speakers. Several text-to-speech models are currently
...
 >>> output = pipe(text)
 ```
 
-Here's a code snippet you can use to listen to the resulting audio in a notebook:
+```python
+>>> from IPython.display import Audio
+>>> Audio(output["audio"], rate=output["sampling_rate"])
+```
+
+You can also do voice cloning with CSM:
```
ebezzam: I would move this CSM example just after the "Hello from Sesame" example, to do something like @vasqu mentioned (all CSM examples together, and then Dia). And introduce it like so:

"By default, CSM uses a random voice. You can do voice cloning by providing a reference audio as part of a chat template dictionary:"

docs/source/en/tasks/text-to-speech.md
```diff
@@ -24,24 +24,68 @@ languages and for multiple speakers. Several text-to-speech models are currently
...
 >>> pipe = pipeline("text-to-speech", model="nari-labs/Dia-1.6B-0626")
 >>> text = "[S1] (clears throat) Hello! How are you? [S2] I'm good, thanks! How about you?"
 >>> output = pipe(text)
 ```
 
-Here's a code snippet you can use to listen to the resulting audio in a notebook:
+```python
+>>> from IPython.display import Audio
+>>> Audio(output["audio"], rate=output["sampling_rate"])
+```
```
ebezzam: Move above line here like so:

---

Some models, like [Dia](https://huggingface.co/nari-labs/Dia-1.6B-0626), can also be conditioned to generate non-verbal communications such as laughing, sighing and crying, or even add music. Below is such an example:

```
EXAMPLE
```
 
Note that Dia also accepts speaker tags such as `[S1]` and `[S2]` to generate a conversation between unique voices.

docs/source/en/tasks/text-to-speech.md
```diff
@@ -24,24 +24,68 @@ languages and for multiple speakers. Several text-to-speech models are currently
...
+...     },
+...     {"role": "0", "content": [{"type": "text", "text": "How much money can you spend?"}]},
+... ]
+>>> output = pipe(conversation)
+```
 
 ```python
 >>> from IPython.display import Audio
 >>> Audio(output["audio"], rate=output["sampling_rate"])
 ```
```
ebezzam, line 75: As @vasqu mentioned, we can remove duplicated snippets for running in a notebook

ebezzam (commented):
Thanks @Deep-unlearning for updating the examples! Some re-ordering will help for clarity!

docs/source/en/tasks/text-to-speech.md
Deep-unlearning: Made it less confusing!

docs/source/en/tasks/text-to-speech.md
Deep-unlearning: I've removed the duplicates!

docs/source/en/tasks/text-to-speech.md
Deep-unlearning: Thanks for the suggestions, i've changed the order of the examples

docs/source/en/tasks/text-to-speech.md
Deep-unlearning: Thanks for the suggestions, i've added it!

vasqu (approved):
Thanks, looking sharp now 

Would leave a final look to @ebezzam since you have more exp on the pipeline side now ;)

ebezzam (approved):
Thanks @Deep-unlearning, looks good!

PR #42487, opened 2025-11-28
[`CI`] Fix copies

vasqu:
Previous PR merged without syncing with main --> new model --> out of sync

### DIFF
--- a/src/transformers/models/nanochat/modeling_nanochat.py
+++ b/src/transformers/models/nanochat/modeling_nanochat.py
@@ -30,6 +30,7 @@ from ... import initialization as init
 from ...activations import ACT2FN
 from ...cache_utils import Cache, DynamicCache
 from ...generation import GenerationMixin
+from ...integrations import use_kernel_func_from_hub
 from ...masking_utils import create_causal_mask
 from ...modeling_layers import GradientCheckpointingLayer
 from ...modeling_outputs import BaseModelOutputWithPast, CausalLMOutputWithPast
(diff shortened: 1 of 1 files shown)


ydshieh (approved):
ok, but see comment

src/transformers/models/nanochat/modeling_nanochat.py
```diff
@@ -218,6 +220,7 @@ def __init__(self, config: NanoChatConfig, layer_idx: int):
         self.o_proj = nn.Linear(
             config.num_attention_heads * self.head_dim, config.hidden_size, bias=config.attention_bias
         )
+        self.rotary_fn = apply_rotary_pos_emb
```
ydshieh, line 223: if you are sure this should be here (i mean I understand the sync , but does this `self.rotary_fn = apply_rotary_pos_emb` makes sense for this model. If so, good. if not, maybe should not be a sync and the modular file should be adjusted

vasqu: Afaik, it's due to kernels needing this but let's double check @MekkCyber

ArthurZucker: I think we wanted it outside not using self.xxxx

ArthurZucker (commented):
ty tho, let's move it out!

ArthurZucker:
If possible

src/transformers/models/nanochat/modeling_nanochat.py
vasqu: https://github.com/huggingface/transformers/pull/41147#discussion_r2572190793

Looks like it's not possible? Otherwise, we need to update all models. But it's getting late, let's resolve this next week with a fresh mind 😄

kaixuanliu:
Hi @ArthurZucker , As this [comment](https://github.com/huggingface/transformers/pull/41147#discussion_r2572190793): curent design for `use_kernel_func_from_hub` decorator need bind the function to a Module. And I think this design is excatly what we want: we need to control the kernelize replacement for function by `use_kernels` param in `from_pretrained` as well. Hence I want to keep this `self.xxx` if possible.

ArthurZucker:
Right my bad!

Issue #42489, opened 2025-11-28
PEFT training with gradient checkpointing fails


qgallouedec:
### System Info

- `transformers` version: 5.0.0.dev0
- PEFT : 0.18.0
- Platform: Linux-5.15.0-1048-aws-x86_64-with-glibc2.31
- Python version: 3.12.12
- Huggingface_hub version: 1.1.5
- Safetensors version: 0.7.0
- Accelerate version: 1.12.0.dev0
- Accelerate config:    not found
- DeepSpeed version: 0.18.2
- PyTorch version (accelerator?): 2.9.0+cu128 (CUDA)
- Using distributed or parallel set-up in script?: No
- Using GPU in script?: Yes
- GPU type: NVIDIA H100 80GB HBM3

### Who can help?

@BenjaminBossan @SunMarc 

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

```python
import numpy as np
from datasets import Dataset
from peft import LoraConfig, get_peft_model
from transformers import AutoModelForCausalLM, Trainer, TrainingArguments

data = np.random.randint(0, 1000, (16, 64)).tolist()
dataset = Dataset.from_dict({"input_ids": data, "labels": data})
model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen3-0.6B", dtype="auto", device_map="auto")
model = get_peft_model(model, LoraConfig())
trainer = Trainer(
    model=model,
    args=TrainingArguments(gradient_checkpointing=True),
    train_dataset=dataset,
)
trainer.train()
```

```
Loading weights: 100%|██████████████████████████████████████████████████████| 311/311 [00:00<00:00, 1662.75it/s, Materializing param=lm_head.weight]
The model is already on multiple devices. Skipping the move to device specified in `args`.
  0%|                                                                                                                         | 0/6 [00:00<?, ?it/s]/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/torch/utils/checkpoint.py:85: UserWarning: None of the inputs have requires_grad=True. Gradients will be None
  warnings.warn(
Traceback (most recent call last):
  File "/fsx/qgallouedec/trl/sandbox/prepare_model_fp32_lora_gc_on.py", line 23, in <module>
[...]

PR #42494, opened 2025-11-29
Fix PEFT gradient checkpointing with frozen input embeddings

Aaraviitkgp:
error: #42489
I have added a fallback check in gradient_checkpointing_enable, if the PEFT flag is missing, the code now checks if the input embeddings are frozen. If they are, it explicitly calls enable_input_require_grads().

@qgallouedec


Aaraviitkgp:
@qgallouedec Th test_processors and test_tokenization error seem to have nothing to do with pr, is there a issue in ci pipeline, it happened before in my other pr aswell.

PR #42496, opened 2025-11-29
feat: allow CP with trainer

kmehant:
# What does this PR do?

1. does batch padding to cp_size * 2 with shift_label calculation (we can omit this if you think this has to be outside the train loop)
2. adds cp accelerate context
3. loss calculation for cp and cp + dp cases

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

- trainer: @SunMarc

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -2517,9 +2517,45 @@ class Trainer:
                         and self.accelerator.distributed_type != DistributedType.DEEPSPEED
                         else contextlib.nullcontext
                     )
-                    with context():
-                        tr_loss_step = self.training_step(model, inputs, num_items_in_batch)
 
+                    cp_context = contextlib.nullcontext()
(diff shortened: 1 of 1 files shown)


src/transformers/trainer.py
```diff
@@ -2517,9 +2517,45 @@ def _inner_training_loop(
...
+                                inputs["position_ids"],
+                            ],
+                            buffer_seq_dims=[1, 1, 1, 1],
+                            no_restore_buffers={
+                                inputs["input_ids"],
+                                inputs["shift_labels"],
+                                inputs["labels"],
+                                inputs["position_ids"],
+                            },
+                        )
```
SunMarc, line 2542: we have `_prepare_context_parallel_inputs` in `training_step` ! If things needs to be changed, it's better to do it there

SunMarc (commented):
Left a comment !

PR #42497, opened 2025-11-30
Add `skip_post_init` flag

### LINKED ISSUE #42418: Custom nn.Parameter initialization in PreTrainedModel subclasses is overwritten by post_init()/from_pretrained() causing NaNs/Zeros
### System Info

- Platform: Linux-4.18.0-147.mt20200626.413.el8_1.x86_64-x86_64-with-glibc2.35
- Python version: 3.10.14
- Safetensors version: 0.6.2
- Accelerate version: 1.11.0
- DeepSpeed version: 0.18.2
- PyTorch version (accelerator?): 2.7.1+cu126 (CUDA)
- Jax version: not installed
- Using GPU in script?: No
- GPU type: NVIDIA A100-SXM4-80GB

### Who can help?

@Cyrilvallez @zucchini-nlp 

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

```python
import numpy as np
import os
import random
import torch
import torch.nn as nn

from transformers import Qwen3VLForConditionalGeneration


def seed_everything(TORCH_SEED):
    random.seed(TORCH_SEED)
    os.environ["PYTHONHASHSEED"] = str(TORCH_SEED)
    np.random.seed(TORCH_SEED)
    torch.manual_seed(TORCH_SEED)
    torch.cuda.manual_seed(TORCH_SEED)
    torch.cuda.manual_seed_all(TORCH_SEED)
    torch.backends.cudnn.deterministic = True
    torch.backends.cudnn.benchmark = False


seed_everything(66)


class TestModel1(Qwen3VLForConditionalGeneration):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.action_head = nn.Linear(1024, 7)
        self.positional_embedding = nn.Parameter(torch.randn(16, 1152))
        self.post_init()


class TestModel2(nn.Module):
    def __init__(self, *args, model_path, **kwargs):
        super().__init__(*args, **kwargs)
        self.model = Qwen3VLForConditionalGeneration.from_pretrained(model_path)
        self.action_head = nn.Linear(1024, 7)
        self.positional_embedding = nn.Parameter(torch.randn(16, 1152))


test_model1 = TestModel1.from_pretrained("Qwen/Qwen3-VL-4B-Instruct")
test_model2 = TestModel2(model_path="Qwen/Qwen3-VL-4B-Instruct")
print(test_model1.positional_embedding)
print(test_model1.positional_embedding.mean(), test_model1.positional_embedding.std())
print(test_model2.positional_embedding)
print(test_model2.positional_embedding.mean(), test_model2.positional_embedding.std())
````

### Expected behavior

When subclassing a model (inheriting from PreTrainedModel, e.g., Qwen3VLForConditionalGeneration, LlamaForCausalLM) to add custom learnable parameters, user-defined initialization in __init__ is often silently overwritten.

This occurs because from_pretrained (or the end of __init__) triggers self.post_init(), which recursively calls _init_weights. This mechanism re-initializes all parameters, ignoring the explicit initialization code provided by the user in __init__.

In the specific case of Qwen3-VL (and potentially others), this re-initialization results in NaNs or Zeros, rendering the model unusable without manual intervention.

Steps to reproduce The following script demonstrates the issue. Note: I used torch.randn for the custom parameter initialization. While I understand that torch.randn samples from a standard normal distribution and does not guarantee an exact sample mean of 0 and std of 1, it should result in valid float values. The observed NaNs/Zeros confirm that this initialization is being discarded and replaced by a faulty internal initialization logic.

### PR DESCRIPTION
**Pull request by Aravind-11**
# What does this PR do?

This PR adds a `skip_post_init` parameter to `PreTrainedModel.from_pretrained()` to allow users to skip the post-initialization step that reinitializes model weights. This is essential for users who subclass models with custom parameters and want to preserve their initialization.

When users subclass HuggingFace models (e.g., `Qwen2VLForConditionalGeneration`) and add custom parameters initialized in their `__init__()`, the initialization gets silently overwritten by `post_init()` which calls `_init_weights()` .

Fixes #42418 

## Before submitting
      Pull Request section?
      to it if that's the case.
- [ ] Did you write any new necessary tests? Let me know if additional tests are needed. 

## Who can review?

@ArthurZucker @Cyrilvallez

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -1215,6 +1215,7 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
 
     def __init__(self, config: PreTrainedConfig, *inputs, **kwargs):
         super().__init__()
+        self._skip_post_init = kwargs.pop("skip_post_init", False)
         if not isinstance(config, PreTrainedConfig):
             raise TypeError(
                 f"Parameter config in `{self.__class__.__name__}(config)` should be an instance of class "
@@ -1258,12 +1259,14 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
 
         self._no_split_modules = self._no_split_modules or []
         _CAN_RECORD_REGISTRY[str(self.__class__)] = self._can_record_outputs  # added for executorch support only
-
+        
     def post_init(self):
         """
         A method executed at the end of each Transformer model initialization, to execute code that needs the model's
         modules properly initialized (such as weight initialization).
         """
+        if getattr(self, "_skip_post_init", False):
+            return
         # Attach the different parallel plans and tied weight keys to the top-most model, so that everything is
         # easily available
         self._tp_plan, self._ep_plan, self._pp_plan = {}, {}, {}
@@ -3518,6 +3521,7 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         revision: str = "main",
         use_safetensors: Optional[bool] = True,
         weights_only: bool = True,
+        skip_post_init: bool = False,
         **kwargs,
     ) -> SpecificPreTrainedModelType:
         r"""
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by ArthurZucker — COMMENTED** [review]
Ty tho I think we want to be a bit more careful in designing this! 
We need tests that's for sure, but yes we need a good api to make sure users are not supprised. 

For example -> based on the classes that are in the public init only. 
cc @Cyrilvallez wdyt

**Comment by Cyrilvallez** [comment]
Hey @Aravind-11 ! The thing is that `post_init` is responsible for more than weight initialization, so we cannot simply skip it. Also, adding yet another flag is not a good idea IMO. Custom models should instead do
```python
def _init_weights(self, module): ...
```
to skip weight initialization, instead of 
```python
self.skip_post_init = True
```
It's not more verbose, and avoid adding a flag and missing the rest of the `post_init` operations.

However, we could indeed think of something so that it's the default from remote code models. But in general we cannot be sure if people will rely on it or not, so a bit hard to decide. What's sure it that this used to be the default, so could make sense to go back to it for remote code

Aravind-11:
That makes a lot of sense @Cyrilvallez ! Thank you for clarifying!!

PR #42501, opened 2025-12-01
Add afmoe to models __init__.py

alyosha-swamy:


ArthurZucker (approved):
Was in release, need ty!

Issue #42460, opened 2025-11-27
Loading a checkpoint with tied weights results in `ValueError: This checkpoint seem corrupted.`


tomaarsen:
### System Info

- `transformers` version: 5.0.0.dev0
- Platform: Windows-10-10.0.26100-SP0
- Python version: 3.11.13
- Huggingface_hub version: 1.1.5
- Safetensors version: 0.6.2
- Accelerate version: 1.11.0
- Accelerate config:    not found
- DeepSpeed version: not installed
- PyTorch version (accelerator?): 2.9.0+cu126 (CUDA)
- Using distributed or parallel set-up in script?: No
- Using GPU in script?: No
- GPU type: NVIDIA GeForce RTX 3090

### Who can help?

@Cyrilvallez @molbap 

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction
```python
from transformers import AutoModel

model_name = "sentence-transformers/sentence-t5-base"
model = AutoModel.from_pretrained("sentence-transformers/sentence-t5-base")
print(type(model))
```
Currently, the behaviour on `main` is:
```
Traceback (most recent call last):
  File "c:\code\transformers\demo_sentence_t5.py", line 5, in <module>
    model = AutoModel.from_pretrained("sentence-transformers/sentence-t5-base")
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\code\transformers\src\transformers\models\auto\auto_factory.py", line 373, in from_pretrained
    return model_class.from_pretrained(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\code\transformers\src\transformers\modeling_utils.py", line 275, in _wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "C:\code\transformers\src\transformers\modeling_utils.py", line 3977, in from_pretrained
    model, missing_keys, unexpected_keys, mismatched_keys, offload_index, error_msgs = cls._load_pretrained_model(
                                                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\code\transformers\src\transformers\modeling_utils.py", line 4163, in _load_pretrained_model
[...]

Issue #42313, opened 2025-11-20
[v5, Mistral models broken] Embeddings are automatically tied


patrickvonplaten:
### System Info

...

### Who can help?

```py
import torch
from transformers import Mistral3ForConditionalGeneration, AutoTokenizer

model_id = "mistralai/Mistral-Small-3.2-24B-Instruct-2506"

# Load tokenizer
tokenizer = AutoTokenizer.from_pretrained(model_id)

# Load model
model = Mistral3ForConditionalGeneration.from_pretrained(
    model_id,
    torch_dtype=torch.bfloat16,
    device_map="auto",
).eval()

# Chat template input
messages = [
    {
        "role": "user",
        "content": [
            {
                "type": "text",
                "text": "Hello",
            }
        ],
    }
]

# Tokenize
tokenized = tokenizer.apply_chat_template(messages, return_dict=True)
input_ids = torch.tensor(tokenized["input_ids"], device="cuda").unsqueeze(0)
attention_mask = torch.tensor(tokenized["attention_mask"], device="cuda").unsqueeze(0)

# Generate
with torch.inference_mode():
    output = model.generate(
        input_ids=input_ids,
        attention_mask=attention_mask,
        max_new_tokens=128,
    )[0]

# Decode
decoded_output = tokenizer.decode(output, skip_special_tokens=True)
print(decoded_output)
```

is broken an current main

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

execute the code

### Expected behavior

To give non gibberish

**Note**: Only happens on v5

patrickvonplaten:
Can be fixed by setting `tie_word_embeddings=False` for the `config.json`

molbap:
Acknowledged - current bug is on `main` branch (4.57 dev), to fix before v5 release.

ArthurZucker:
Yeah, we need to change the default to `False` in individual config. It was always `True` but never enforced. Sorry about that

molbap:
This should be solved thanks to https://github.com/huggingface/transformers/pull/42507. Tested it, just FYI this also is needed to make it work https://github.com/huggingface/transformers/pull/42997 (there's a bug in the automapping). 

On our radar in any case for the full v5.0 release!

molbap:
[...]

PR #42507, opened 2025-12-01
[loading] `tie_weights` is finally symmetric

Cyrilvallez:
# What does this PR do?

As per the title. This makes `tie_weights` work fully symmetrically, even in cases where one weight is tied between more than 2 keys.
This will hopefully end the pain of all models checkpoints having the saved weight in the "wrong" key.

This fixes several opened issues: https://github.com/huggingface/transformers/issues/42313 (though we still have an issue with the tokenizer here apparently), https://github.com/huggingface/transformers/issues/42460 and I think others that I lost track of.

Note: for `facebook/bart-large` bart checkpoint, all 3 tied weight keys are explicitly present in the safetensors checkpoint. So according to this PR, they are not tied anymore (with a warning), whereas they would be anyway before. IMO this is fine as if they are in the checkpoints, they were not tied when saved. We could introduce a value check if both source and targets are present and still tie if everything is equal, but I think this is too expensive and too niche to add for now. We may consider doing it if we notice this is more frequent so those `bart` checkpoints

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -2283,6 +2283,9 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         If `recompute_mapping=True`, it will re-check all internal submodels and their config to determine the params
         that need to be tied. This is the default when `model.tie_weights()` is called on its own, outside of
         `__init__`, and `from_pretrained`, in case the config values were changed somewhere.
+
+        Note that during `from_pretrained`, this function will work symmetrically, i.e. if the `source` is a missing
+        weight, but the `target` is present, tying will still work in reverse.
         """
         # In this case, the keys stored in `all_tied_weights_keys` are already correct
         if not recompute_mapping:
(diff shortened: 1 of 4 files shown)


src/transformers/modeling_utils.py
```diff
@@ -2283,37 +2283,74 @@ def tie_weights(self, missing_keys: Optional[set[str]] = None, recompute_mapping
...
+                # checkpoint -> let's raise a warning and do nothing
+                if source_is_there and target_is_there:
+                    logger.warning(
+                        f"The tied weights mapping and config for this model specifies to tie {source_param_name} to "
+                        f"{target_param_name}, but both are present in the checkpoints, so we will NOT tie them. "
+                        "You should update the config with `tie_word_embeddings=False` to silence this warning"
                     )
+                    # Skip to next iteration
+                    continue
+                # We're missing the source but we have the target -> simply reverse both
```
molbap:
```suggestion
                # We're missing the source but we have the target -> we swap them, tying the parameter that exists
```

src/transformers/modeling_utils.py
```diff
@@ -2283,37 +2283,75 @@ def tie_weights(self, missing_keys: Optional[set[str]] = None, recompute_mapping
         If `recompute_mapping=True`, it will re-check all internal submodels and their config to determine the params
         that need to be tied. This is the default when `model.tie_weights()` is called on its own, outside of
         `__init__`, and `from_pretrained`, in case the config values were changed somewhere.
+
+        Note that during `from_pretrained`, this function will work symmetrically, i.e. if the `source` is a missing
+        weight, but the `target` is present, tying will still work in reverse.
```
molbap: mostly phrasing/clarity
```suggestion
        Note that during `from_pretrained`, tying is *symmetric*: if the
        mapping says "tie target -> source" but `source` is missing in the checkpoint while `target` exists,
        we *swap* source and target so we can still tie everything to the parameter that actually exists.
```

src/transformers/modeling_utils.py
```diff
@@ -2283,37 +2283,75 @@ def tie_weights(self, missing_keys: Optional[set[str]] = None, recompute_mapping
...
-                name = target_param_name
-                parent = self
-            setattr(parent, name, source_param)
-            self._adjust_bias(parent, source_param)
+        tied_keys = list(tied_keys.items())
+        for i, (target_param_name, source_param_name) in enumerate(tied_keys):
+            # If both source and target are missing later, we may need to tie 2 weights so we use a list here
+            target_param_names = [target_param_name]
+
+            # This is `from_pretrained` -> let's check symmetrically
```
molbap:
```suggestion
```

src/transformers/modeling_utils.py
```diff
@@ -2283,37 +2283,75 @@ def tie_weights(self, missing_keys: Optional[set[str]] = None, recompute_mapping
...
-                parent_name, name = target_param_name.rsplit(".", 1)
-                parent = self.get_submodule(parent_name)
-            else:
-                name = target_param_name
-                parent = self
-            setattr(parent, name, source_param)
-            self._adjust_bias(parent, source_param)
+        tied_keys = list(tied_keys.items())
+        for i, (target_param_name, source_param_name) in enumerate(tied_keys):
+            # If both source and target are missing later, we may need to tie 2 weights so we use a list here
```
molbap: If I understand the logic?

```suggestion
            # Usually we tie a single target to a single source, but when both are missing we may later tie
            # multiple names to a third "backup" parameter that is present in the checkpoint.
```

src/transformers/modeling_utils.py
```diff
@@ -2283,37 +2283,75 @@ def tie_weights(self, missing_keys: Optional[set[str]] = None, recompute_mapping
...
                     )
+                    # Skip to next iteration
+                    continue
+                # We're missing the source but we have the target -> simply reverse both
+                elif not source_is_there and target_is_there:
+                    target_param_name, source_param_name = source_param_name, target_param_name
+                    target_param_names = [target_param_name]
+                # Both are missing -> check other keys in case more than 2 keys are tied to the same weight
+                elif not source_is_there and not target_is_there:
+                    for target_backup, source_backup in tied_keys[i + 1 :]:
```
molbap, line 2325: maybe dumb question, why do we only search forward here? is there some ordering implied?

molbap (commented):
Read through the new logic, thanks! I suggested some changes in the comments for clarity mostly, logic seems sound except the ordering in case of missing both keys but that's it!

src/transformers/modeling_utils.py
Cyrilvallez: Since they all have the same `source`, and we tie both source and target if we find it somewhere at the beginning when both are missinf, all later iterations will not enter this branch anymore as the source is now present 👌 But you're definitely right to ask, not necessarily trivial!

molbap: aah got it. Thanks! Yeah all live modifications of iterables are hard to track haha

src/transformers/modeling_utils.py
```diff
@@ -2283,37 +2283,80 @@ def tie_weights(self, missing_keys: Optional[set[str]] = None, recompute_mapping
...
-                elif target_is_there:
-                    raise ValueError(
-                        f"This checkpoint seem corrupted. The tied weights mapping for this model specifies to tie "
-                        f"{source_param_name} (which should be present and is not), to {target_param_name} (which is "
-                        f"present)."
+                # Both are already present -> it means the config is wrong and do not reflect the actual
+                # checkpoint -> let's raise a warning and do nothing
+                if source_is_there and target_is_there:
+                    logger.warning(
+                        f"The tied weights mapping and config for this model specifies to tie {source_param_name} to "
```
ArthurZucker, line 2313: very nice and actionnable

src/transformers/modeling_utils.py
```diff
@@ -2283,37 +2283,80 @@ def tie_weights(self, missing_keys: Optional[set[str]] = None, recompute_mapping
...
+                    parent = self.get_submodule(parent_name)
+                else:
+                    name = target_param_name
+                    parent = self
+                # Tie the weights
+                setattr(parent, name, source_param)
+                self._adjust_bias(parent, source_param)
+                # Remove from missing if necesary
+                if missing_keys is not None:
+                    missing_keys.discard(target_param_name)
```
ArthurZucker, line 2362: can be isolated but LGTM

tests/utils/test_modeling_utils.py
```diff
@@ -1372,6 +1394,83 @@ def test_tied_weights_reload(self):
...
+                        tmp_dir, output_loading_info=True
+                    )
+                    # Assert no missing keys
+                    self.assertSetEqual(load_info["missing_keys"], set())
+                    # It's still the same weight
+                    self.assertIs(new_model.linear.weight, new_model.linear_2.weight)
+                    self.assertIs(new_model.linear.weight, new_model.linear_3.weight)
+
+                    # Make sure both state dict are the same
+                    compare_state_dicts(model.state_dict(), new_model.state_dict())
```
ArthurZucker, line 1476: can you add message helpers please 🙏🏻

ArthurZucker (approved):
Thanks, let's make sure the marian and etc models are fixed !

PR #41173, opened 2025-09-26
Rope for Qwen2--5-vl

zucchini-nlp:
# What does this PR do?

Attempt to fix https://github.com/huggingface/transformers/issues/41093. I believe that the `is_prefill()` logic had edge cases which were caught in the linked issue. Let's remove it since the position ids are also prepared in `prepare_inputs_for_generation`.

PR #42508, opened 2025-12-01
Fix Qwen-VL family with prompt tuning

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/pull/41173#issuecomment-3586424032.

TL;DR: PEFT pop position ids and cache position when doing prompt tuning, and expects that they will be reconstructed in forward using past cache length. However Qwen assumes that `cache_position == None` is always a prefill stage which will not be correct with PEFT

This PR changes the is-prefill check in qwen forward pass to be more explicit and rely on cache length. Checked with the below script

cc @BenjaminBossan . Btw, there are some more issues if we install "rc0" and run PEFT. I guess one of the breaking PR affected it, lmk if you need help on that as well

```python
import torch
from transformers import AutoTokenizer, Qwen2_5_VLForConditionalGeneration
from peft import get_peft_model, PromptTuningConfig, PromptTuningInit, TaskType

max_new_tokens = 40
model_id =  "Qwen/Qwen2.5-VL-3B-Instruct"
text = "Discuss the most important work by Mary Shelley."
inputs = tokenizer(text, return_tensors="pt").to(0)

model = Qwen2_5_VLForConditionalGeneration.from_pretrained(model_id, device_map=0)
prompt_tune_config = PromptTuningConfig(
    task_type=TaskType.CAUSAL_LM,
    prompt_tuning_init=PromptTuningInit.RANDOM,
    num_virtual_tokens=5,
)
model = get_peft_model(model, prompt_tune_config)
model.prompt_encoder.default.embedding.weight.data.zero_()  # make peft almost no-op

torch.manual_seed(0)
with torch.no_grad():
    generated_ids = model.generate(
        # seq len should be 5 virtual + 9 normal tokens
        **inputs,
        max_new_tokens=max_new_tokens,
        # use_cache=False,  # without cache, it works
    )
print(tokenizer.batch_decode(generated_ids, skip_special_tokens=True)[0])

```

### DIFF
--- a/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py
@@ -1207,7 +1207,7 @@ class Qwen3VLModel(Qwen3VLPreTrainedModel):
                     input_ids,
                     image_grid_thw,
                     video_grid_thw,
-                    attention_mask=attention_mask_tensor,
+                    attention_mask=attention_mask,
                 )
                 self.rope_deltas = rope_deltas
             # then use the prev pre-calculated rope-deltas to get the correct position ids
--- a/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py
@@ -1213,11 +1213,7 @@ class Qwen3VLModel(Qwen3VLPreTrainedModel):
             # then use the prev pre-calculated rope-deltas to get the correct position ids
             else:
                 batch_size, seq_length, _ = inputs_embeds.shape
-                delta = (
-                    (cache_position[0] + self.rope_deltas).to(inputs_embeds.device)
-                    if cache_position is not None
-                    else 0
-                )
+                delta = (past_key_values_length + self.rope_deltas).to(inputs_embeds.device)
                 position_ids = torch.arange(seq_length, device=inputs_embeds.device)
                 position_ids = position_ids.view(1, -1).expand(batch_size, -1)
                 if cache_position is not None:  # otherwise `deltas` is an int `0`
--- a/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py
@@ -38,7 +38,7 @@ from ...modeling_outputs import BaseModelOutputWithPast, ModelOutput
 from ...modeling_rope_utils import ROPE_INIT_FUNCTIONS, dynamic_rope_update
 from ...modeling_utils import ALL_ATTENTION_FUNCTIONS, PreTrainedModel
 from ...processing_utils import Unpack
-from ...utils import TransformersKwargs, auto_docstring, is_torchdynamo_compiling
+from ...utils import TransformersKwargs, auto_docstring
 from ...utils.generic import check_model_inputs
 from .configuration_qwen3_vl import Qwen3VLConfig, Qwen3VLTextConfig, Qwen3VLVisionConfig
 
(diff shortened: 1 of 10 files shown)


BenjaminBossan:
Thanks @zucchini-nlp this indeed solves the issue. The generations with cache enabled are still bad compared to cache being disabled, but that was already true before #41173 so it must be something else. 

```
With cache:
Discuss the most important work by Mary Shelley. Mary Shelley,,, and, and, and and and and the novel Shelley W., and and and, and the, and Mary Shelley, and and Mary Shelley Shelley, and Mary Shelley,
```

```
Without cache:
Discuss the most important work by Mary Shelley. Mary Shelley was a British author who wrote "Frankenstein" in 1818. She was born in 1797 and died in 1851. She was a
```

> Btw, there are some more issues if we install "rc0" and run PEFT. I guess one of the breaking PR affected it, lmk if you need help on that as well

We have a backlog of a couple of things that are broken since then, but maybe not everything. What do you mean specifically?

zucchini-nlp:
> The generations with cache enabled are still bad compared to cache being disabled,

Hmm, I ran the script with 4.57.3 and got a normal generation. Smth might have been broken after that in peft

> We have a backlog of a couple of things that are broken since then, but maybe not everything. What do you mean specifically?

Nothing specifically, just saw a few issues which I think are related to recent breaking changes I introduced in v5. If peft might requires fixes or changes on transformers side re these, lmk

zucchini-nlp:
run-slow: qwen2_5_omni, qwen2_5_vl, qwen2_vl, qwen3_omni_moe, qwen3_vl, qwen3_vl_moe

BenjaminBossan:
> Hmm, I ran the script with 4.57.3 and got a normal generation. Smth might have been broken after that in peft

Hmm, how? Did you apply your patch from this PR to 4.57.3? Also, what PEFT version did you use? I tried a few PEFT versions back and each time, the generation was just gibberish.

> Nothing specifically, just saw a few issues which I think are related to recent breaking changes I introduced in v5. If peft might requires fixes or changes on transformers side re these, lmk

Thanks for the offer, I'll ping you if something comes up. The remaining issues I'm aware of are related to weight conversion or tied weights/gradient checkpointing.

src/transformers/models/qwen3_vl/modeling_qwen3_vl.py
```diff
@@ -1449,8 +1428,33 @@ def prepare_inputs_for_generation(
...
+                self.model.rope_deltas = rope_deltas
+            # then use the prev pre-calculated rope-deltas to get the correct position ids
+            elif "position_ids" in model_inputs:
+                batch_size, seq_length = model_inputs["position_ids"].shape
+                device = model_inputs["position_ids"].device
+                position_ids = torch.arange(seq_length, device=device)
+                position_ids = position_ids.view(1, 1, -1).expand(3, batch_size, -1)
+                delta = cache_position[0] + self.model.rope_deltas
+                delta = delta.repeat_interleave(batch_size // delta.shape[0], dim=0)
+                vision_positions = position_ids + delta.expand_as(position_ids)
```
molbap, line 1449: seems ok, so aligned with qwen2- 2.5?

molbap (approved):
Thanks for fixing! Trusting you on the harmonisation between qwen family members.

src/transformers/models/qwen3_vl/modeling_qwen3_vl.py
zucchini-nlp: yep, mostly copy from qwen2-vl. Modular is overwriting this class, so we're not copying automatically apparently

zucchini-nlp:
> Hmm, how? Did you apply your patch from this PR to 4.57.3? Also, what PEFT version did you use? I tried a few PEFT versions back and each time, the generation was just gibberish.

@BenjaminBossan oh you're right, my mind made sense of gibberish and assumed it was normal haha. The issue is again related to popping cache position because the fallback was initialize position ids of all zeros. I fixed it in my last commit to rely on cache length as it is equivalent to `cache_position[0]`

zucchini-nlp:
run-slow: qwen2_5_omni, qwen2_5_vl, qwen2_vl, qwen3_omni_moe, qwen3_vl, qwen3_vl_moe

BenjaminBossan:
> The issue is again related to popping cache position because the fallback was initialize position ids of all zeros. I fixed it in my last commit to rely on cache length as it is equivalent to `cache_position[0]`

Oh really nice, now I get the same generations with and without cache, thank you so so much!

PR #40617, opened 2025-09-02
Fix flaky `JambaModelTest.test_load_balancing_loss`

ydshieh:
# What does this PR do?

Fix this flaky test which has

```bash
>       torch.testing.assert_close(result.aux_loss.cpu(), torch.tensor(2, dtype=torch.float32), rtol=1e-2, atol=1e-2)
E       AssertionError: Scalars are not close!
```

Running 10000 times and all pass.

Also makes this much much faster by reducing model size and padded sequence length: running the same tests 20 times

    20 passed, 2 warnings in 11.35s v.s. 20 passed, 2 warnings in 75.72s (0:01:15) 

### Full log

```
___________________ JambaModelTest.test_load_balancing_loss ____________________
[gw3] linux -- Python 3.9.23 /usr/local/bin/python3

self = <tests.models.jamba.test_modeling_jamba.JambaModelTest testMethod=test_load_balancing_loss>

    def test_load_balancing_loss(self):
        r"""
        Let's make sure we can actually compute the loss and do a backward on it.
        """
        config, input_dict = self.model_tester.prepare_config_and_inputs_for_common()
        config.num_labels = 3
        config.num_experts = 16
        config.output_router_logits = True
        input_ids = input_dict["input_ids"]
        attention_mask = input_ids.ne(config.pad_token_id).to(torch_device)
        model = JambaForCausalLM(config)
        model.to(torch_device)
        model.eval()
        result = model(input_ids, attention_mask=attention_mask)
        bs, seqlen = input_ids.shape
        self.assertEqual(result.router_logits[0].shape, (bs * seqlen, config.num_experts))
>       torch.testing.assert_close(result.aux_loss.cpu(), torch.tensor(2, dtype=torch.float32), rtol=1e-2, atol=1e-2)
E       AssertionError: Scalars are not close!
E       
E       Expected 2.0 but got 2.0415196418762207.
E       Absolute difference: 0.0415196418762207 (up to 0.01 allowed)
E       Relative difference: 0.02075982093811035 (up to 0.01 allowed)
```

https://app.circleci.com/pipelines/github/huggingface/transformers/144153/workflows/ce26cb2f-f3b1-45c5-bca9-60868311cd7f/jobs/1905517

PR #42509, opened 2025-12-01
another way of fix for #42400

ydshieh:
# What does this PR do?

This was originally #40617, which IMO a better approach than #42400 (despite not ideal neither).

With current form, the failure ratio is 0.1% instead of 1%. We can add `@is_flaky(max_attempts=2)` as in #40617 however.


ydshieh:
@Rocketknight1 if you are happy with this PR, I will apply the same changes to other places.

Rocketknight1:
This works too! I think we should add `is_flaky` and **also** `set_seed` though. All of those will reduce the chance of rare flakiness further, so we might as well do them all together!

ydshieh:
I will apply the changes, but refrain myself of adding `set_seed`. I understand it works, but this is not an integration test that has use randomness to perform generation (or forward) then compare against some predefined expected outputs.

Setting seed for such non-integration tests isn't really a good approach (and just playing some lottery).

Unfortunately, I don't have bandwidth for now to see if there is more intrinsic reason behind the failures (that is related to randomness of weights I agree).

However, with the changes in this PR + is_flaky, it works well (last PR: no failures in 3000 runs)

PR #42511, opened 2025-12-01
[quantization] Dequant fp8 when cuda or xpu not available

MekkCyber:
# What does this PR do?

Instead of erroring when no cuda or xpu device is available, we dequantize the weights to make it possible to do inference on cpu

### DIFF
--- a/src/transformers/integrations/finegrained_fp8.py
+++ b/src/transformers/integrations/finegrained_fp8.py
@@ -15,7 +15,7 @@
 
 import re
 from collections.abc import Sequence
-from typing import Any
+from typing import Any, Optional
 
 from ..core_model_loading import ConversionOps
 from ..utils import is_accelerate_available, is_torch_accelerator_available, is_torch_available, logging
@@ -497,6 +497,7 @@ def _replace_with_fp8_linear(
     quantization_config=None,
     has_been_replaced=False,
 ):
+
     iterator = list(model.named_parameters()).copy()
     for name, empty_tensor in iterator:
         current_key_name = name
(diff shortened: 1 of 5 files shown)


src/transformers/quantizers/quantizer_finegrained_fp8.py
```diff
@@ -231,3 +238,18 @@ def get_quantize_ops(self):
...
+
+        if self.pre_quantized and self.quantization_config.dequantize:
+            return [
+                # either use the dollar sign, or permute the source patterns to start matching against the scales first
+                WeightConverter(
+                    source_patterns=["weight$", "weight_scale_inv"],
+                    target_patterns="weight",
+                    operations=[Fp8Dequantize(self)],
+                )
+            ]
```
ArthurZucker, line 254: perfect!

ArthurZucker (approved):
Just missing a fast test!! Since this is targeted at CPU

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -652,41 +654,45 @@ def convert(self, input_dict: torch.Tensor, **kwargs) -> dict[str, torch.Tensor]
...
+        missing_keys=None,
+        **kwargs,
+    ) -> dict[str, torch.Tensor]:
+        if len(input_dict) != 2:
+            # in case of no scales, the weights are not quantized, so we return the weights as is
+            return {
+                full_layer_name: input_dict["weight$"][0]
+                if isinstance(input_dict["weight$"], list)
+                else input_dict["weight$"]
+            }
```
SunMarc, line 674: I think you can just simply return `input_dict`. I did that for bnb.

SunMarc (approved):
Nice PR, thanks a lot for adding this !

PR #40994, opened 2025-09-19
Check model inputs - hidden states

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/40964

In most vision models the `output.hidden_states` are the hiddens right after encoder blocks, i.e. before the last layernorm. Therefore for these models `output.hidden_states != output.last_hidden_state`

Currently `check_model_inputs` assumes that last hidden state is the correct one to return which is true for language models only. This PR adds a kwarg for `check_model_inputs` which decides whether to replace last hidden state or not

TBH, i think the way it is done in LMs is the ultimate correct version and we probably need to "break" vision models. But I can't think of a way to obtain pre-norm last hidden states which are needed for some VLMs

PR #42512, opened 2025-12-01
Handle decorator with optional arguments better

hmellor:
In https://github.com/huggingface/transformers/pull/40994 this decorator was changed so that, without arguments, it had to be used like

```py
@check_model_inputs()
def ...
```

instead of 

```py
@check_model_inputs
def ...
```

This PR fixes that so that custom models which use this decorator can continue to work.


zucchini-nlp (approved):
Thanks, totally forgot about custom models when adding it

ArthurZucker (approved):
Yep I think I spotted that in my review, should not have slipped through!

PR #42517, opened 2025-12-01
Add backward compatibility for methods which have been moved to `RotaryEmbeddingConfigMixin`

hmellor:
This function was moved in https://github.com/huggingface/transformers/pull/42255. Many custom code models import this function, which is arguably good practise as it shows they are leveraging our first party utilities.

This PR reinstates the function by calling the methods which now live in `RotaryEmbeddingConfigMixin` so that these custom code models will continue to work with Transformers v5.


zucchini-nlp (approved):
Thanks

PR #42518, opened 2025-12-01
[FP8] Support for static FP8

patrickvonplaten:
...

### DIFF
--- a/src/transformers/integrations/finegrained_fp8.py
+++ b/src/transformers/integrations/finegrained_fp8.py
@@ -338,7 +338,7 @@ class FP8Linear(nn.Linear):
 
         self.activation_scheme = activation_scheme
         if self.activation_scheme == "static":
-            self.register_parameter("act_scale", None)
+            self.register_parameter("activation_scale", None)
 
 
         if bias:
--- a/src/transformers/integrations/finegrained_fp8.py
+++ b/src/transformers/integrations/finegrained_fp8.py
@@ -337,6 +337,9 @@ class FP8Linear(nn.Linear):
         self.block_size = block_size
 
         self.activation_scheme = activation_scheme
+        if self.activation_scheme == "static":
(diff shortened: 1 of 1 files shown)


src/transformers/integrations/finegrained_fp8.py
```diff
@@ -337,6 +337,9 @@ def __init__(
         self.block_size = block_size
 
         self.activation_scheme = activation_scheme
+        if self.activation_scheme == "static":
+            self.register_parameter("act_scale", None)
+
```
SunMarc, line 342: Is it possible to create the parameter with the right shape and `dtype` as we do for `weight_scale_inv` ?

SunMarc (approved):
Looks good ! Thanks for adding this, just a nit

src/transformers/integrations/finegrained_fp8.py
MekkCyber: also it would be nice to call it `activation_scale` if possible

MekkCyber (commented):
Thanks a lot for adding this

ArthurZucker (approved):
Let's go! 🚀

src/transformers/integrations/finegrained_fp8.py
patrickvonplaten: Totally fine, can you change directly?

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -337,6 +337,9 @@ def __init__(
         self.block_size = block_size
 
         self.activation_scheme = activation_scheme
+        if self.activation_scheme == "static":
+            self.register_parameter("act_scale", None)
```
SunMarc:
```suggestion
            self.register_parameter("activation_scale", None)
```

PR #42476, opened 2025-11-28
Fixed convert_batch_to_list_format staticmethod function call

Sai-Suraj-27:
# What does this PR do?
Fixes these [failing tests](https://github.com/huggingface/transformers/actions/runs/19758977591/job/56616632584#step:14:5132)

image

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?
@ArthurZucker @zucchini-nlp @itazap

PR #42522, opened 2025-12-01
Fix interactions between require_read_token and staticmethod

Rocketknight1:
As @Sai-Suraj-27 pointed out in #42476, there's something weird going on with static methods in some tests. The cause seems to be `require_read_token` having a faulty check for `staticmethod` which obliterates args in some cases. Trying this to see if it helps!

### DIFF
--- a/src/transformers/testing_utils.py
+++ b/src/transformers/testing_utils.py
@@ -638,6 +638,9 @@ def require_read_token(test_case):
                 if getattr(attr, "__require_read_token__", False):
                     continue
                 wrapped = require_read_token(attr)
+                if isinstance(inspect.getattr_static(test_case, attr_name), staticmethod):
+                    # Don't accidentally bind staticmethods to `self`
+                    wrapped = staticmethod(wrapped)
                 setattr(test_case, attr_name, wrapped)
         return test_case
     else:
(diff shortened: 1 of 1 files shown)


Rocketknight1:
Explanation for reviewers because this one is quite tricky:

The `require_read_token` decorator decorates a test or function. However, in some cases, the function is a method with a `staticmethod` decorator. The problem arises when we apply `@require_read_token` **after** `@staticmethod`. In this case, the decorator returns a non-static function, and so we get lots of argument confusion because a `self` argument is inserted where we don't want it.

The old code fixed this with the hack `if "staticmethod" in inspect.getsource(test_case).strip():`, but this breaks when the test case contains that string **anywhere**. The new decorator code correctly checks whether the method we're grabbing is static, and if so, returns a similarly-static output.

Rocketknight1:
cc @ydshieh since this mostly affects tests!

ydshieh:
Thank you. I added such so hacky due to my limited knowledge of python, sorry.

With this PR, it seems works with or without running with the ` HF_HUB_READ_TOKEN`, so ✅ .

You are so good in python

ydshieh (approved):

ydshieh:
and thanks @Sai-Suraj-27 for identifying this, you are so good at this!

Rocketknight1:
I had to do some reading on this one, the interactions between staticmethods and wrappers were confusing for me as well at first :sweat_smile:

PR #42523, opened 2025-12-01
Fix tied_weights for vlms

Cyrilvallez:
# What does this PR do?

Minimum fix from https://github.com/huggingface/transformers/pull/42385


ArthurZucker (approved):
OK I think bart and seamless m4t and marian were the most proeminant to test

PR #42524, opened 2025-12-01
avoid `afmoe` job crashing

ydshieh:
# What does this PR do?

Avoid `afmoe` job crashing like

https://app.circleci.com/pipelines/github/huggingface/transformers/155470/workflows/1bff2732-c69b-4d34-b2ab-ffc45a80b559/jobs/2044143


vasqu (approved):
Quick solution, if you have time can you check if we can remove the `__init__` like in llama https://github.com/huggingface/transformers/blob/6316a9e176d22f8f09d44ef72ec0aaa2ce7b8780/tests/models/llama/test_modeling_llama.py#L47-L49

ArthurZucker (approved):

PR #42528, opened 2025-12-01
small fix tokenizer regex patch

ArthurZucker:
# What does this PR do?
Were issues

### DIFF
--- a/src/transformers/tokenization_utils_tokenizers.py
+++ b/src/transformers/tokenization_utils_tokenizers.py
@@ -1089,7 +1089,9 @@ class TokenizersBackend(PreTrainedTokenizerBase):
                     return True
             return False
 
-        if pretrained_model_name_or_path is not None and (is_local or is_base_mistral(pretrained_model_name_or_path)):
+        if pretrained_model_name_or_path is not None and (
+            is_local or (not is_local and is_base_mistral(pretrained_model_name_or_path))
+        ):
             _config_file = cached_file(
                 pretrained_model_name_or_path,
                 "config.json",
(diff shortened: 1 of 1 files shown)


src/transformers/tokenization_utils_tokenizers.py
```diff
@@ -219,6 +219,7 @@ def __init__(self, *args, **kwargs):
 
         # Optionally patches mistral tokenizers with wrong regex
         if vocab_size > 100000 and getattr(self._tokenizer, "pre_tokenizer", None) is not None:
+            kwargs.pop("tokenizer", None)
```
vasqu, line 222: We dont need the kwargs in the patch fn, how about we change the signature and remove the **kwargs instead? This was how it was before

ArthurZucker: Yeah but I had to re-introduce them IDR why but the ci was complaining post refactor

winglian (approved):

src/transformers/tokenization_utils_tokenizers.py
```diff
@@ -1089,7 +1090,12 @@ def is_base_mistral(model_id: str) -> bool:
                     return True
             return False
 
-        if pretrained_model_name_or_path is not None and (is_local or is_base_mistral(pretrained_model_name_or_path)):
+        if is_offline_mode():
+            is_local = True
+
+        if pretrained_model_name_or_path is not None and (
+            is_local or (not is_local and is_base_mistral(pretrained_model_name_or_path))
+        ):
```
Killusions, line 1098: @ArthurZucker Can we get a backport of this fix, specifically:

```python
if is_offline_mode():
            is_local = True
```

and

```python
if pretrained_model_name_or_path is not None and (
            is_local or (not is_local and is_base_mistral(pretrained_model_name_or_path))
        ):
```

Without this offline loading is broken (also for non-mistral models).

Killusions: @ArthurZucker Created a minimal backport PR for these lines #42880

ArthurZucker: ok!

PR #42529, opened 2025-12-01
[torchao] safetensors

liangel-02:
**Summary**
updating safetensors logic to support all tensor subclasses generically 

**Test**
verified integration with vllm 
unit tests pass: `python tests/quantization/torchao_integration/test_torchao.py -k TorchAoSafeSerializationTest`

### DIFF
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -48,14 +48,36 @@ if TYPE_CHECKING:
 logger = logging.get_logger(__name__)
 
 
+def extract_concrete_key(key: str, pattern: str, pattern_regex: re.Pattern) -> str:
+    match = pattern_regex.match(key)
+    if not match:
+        return pattern
+
+    groups = match.groups()
+    wildcard_count = pattern.count("*")
+
+    if wildcard_count == 0:
+        return pattern
+    elif wildcard_count == 1:
+        return pattern.replace("*", groups[0])
+    else:
+        parts = pattern.split("*")
+        result = "*".join(parts[1:])
+        for i, captured in enumerate(groups[1:], start=0):
+            result = result.replace("*", str(captured), 1)
+        return result
+
+
 def build_glob_alternation(
     globs: list[Union[WeightRenaming, WeightConverter, str]],
-) -> tuple[re.Pattern, dict[str, str], dict[str, str]]:
+) -> tuple[re.Pattern, dict[str, str], dict[str, str], dict[str, re.Pattern]]:
     """
     Build a single alternation regex with one named group per glob.
+    Returns also a dict mapping group names to their compiled regex patterns for extracting matched parts.
     """
     src_group_to_glob: dict[str, str] = {}
     tgt_group_to_glob: dict[str, str] = {}
+    group_to_pattern: dict[str, re.Pattern] = {}
     branches: list[str] = []
     i = 0
     for glob in globs:
@@ -64,9 +86,12 @@ def build_glob_alternation(
                 group_name = f"g{i}"
                 src_group_to_glob[group_name] = src
                 i += 1
+                # Convert the glob pattern to a regex with capture groups for wildcards
+                pattern_with_captures = src.replace("*", r"(.*?)")
+                group_to_pattern[group_name] = re.compile(f"^{pattern_with_captures}$")
                 body = src.replace("*", r".*")
                 branches.append(f"(?P<{group_name}>{body})")
-                tgt_group_to_glob[group_name] = glob.target_patterns[0]  # we index witht the first target
+                tgt_group_to_glob[group_name] = glob.target_keys[0] if isinstance(glob.target_keys, list) else glob.target_keys
         else:
             group_name = f"g{i}"
             src_group_to_glob[group_name] = glob
@@ -77,7 +102,7 @@ def build_glob_alternation(
             tgt_group_to_glob[group_name] = glob
 
     alternation = re.compile("|".join(branches))
-    return alternation, src_group_to_glob, tgt_group_to_glob
+    return alternation, src_group_to_glob, tgt_group_to_glob, group_to_pattern
 
 
 class ConversionOps:
@@ -444,6 +469,7 @@ class WeightConverter(WeightTransform):
             )
 
         collected_tensors = self.collected_tensors
+
         for op in self.operations:
             with log_to_misc(layer_name, misc, (collected_tensors, layer_name), op):
                 collected_tensors = op.convert(
@@ -532,7 +558,7 @@ def log_to_misc(
     try:
         yield
     except Exception as e:
-
+        print(f"error: {e}")
         def _format_op_name(curr_op: Union[list[ConversionOps], ConversionOps, None]) -> Optional[str]:
             if curr_op is None:
                 return None
@@ -547,6 +573,7 @@ def log_to_misc(
         if isinstance(extras, tuple) and len(extras) == 2:
             values, target_keys = extras
             descriptor = f"{op_name} " if op_name else ""
+            # print(values)
             misc[first_target_key] = (
                 f"{e}\nError: {descriptor}on tensors destined for {target_keys}. Ckpt contains: {len(values)}"
             )
@@ -595,7 +622,7 @@ def set_param_for_module(
                         param_value = param_value.to_local()
                 if param_name not in module_obj._buffers:
                     param_value = torch.nn.Parameter(param_value, requires_grad=param_value.is_floating_point())
-
+            print(f"removing {target_name} from missing keys")
             # Remove from missing keys (it's either mismatched, or all good)
             missing_keys.discard(target_name)
             if ref is not None and ref.shape != param_value.shape and hf_quantizer is None:
@@ -768,6 +795,9 @@ def convert_and_load_state_dict_in_model(
     ```
 
     """
+    print('in convert and load state dict')
+    print(f"model state_dict keys: {model.state_dict().keys()}")
+    print(f"state_dict keys: {state_dict}")
     prefix = model.base_model_prefix
     tp_plan = tp_plan or {}
     device_map = device_map or {"": "cpu"}
@@ -794,9 +824,9 @@ def convert_and_load_state_dict_in_model(
     # build '(?P<g0>.*.*\\.block_sparse_moe\\..*)' and group to source {'g0': '*.block_sparse_moe.'}
     # and target to source {'g0': '*.mlp.'}. This allows us to quickly find which pattern matched.
     if tp_plan != {}:
-        tp_plan_alt, tp_plan_by_group_name, _ = build_glob_alternation(list(tp_plan.keys()))
+        tp_plan_alt, tp_plan_by_group_name, _, _ = build_glob_alternation(list(tp_plan.keys()))
     if dtype_plan != {}:
-        dtype_policy_alt, dtype_policy_by_group_name, _ = build_glob_alternation(list(dtype_plan.keys()))
+        dtype_policy_alt, dtype_policy_by_group_name, _, _ = build_glob_alternation(list(dtype_plan.keys()))
 
     pattern_to_converter = {k: converter for converter in converters for k in converter.source_patterns}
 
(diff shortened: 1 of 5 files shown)


src/transformers/integrations/torchao.py
```diff
@@ -215,56 +215,31 @@ def convert(
...
-        # If it's a bias, no need to do anything special (except removing the ":_data" part of the key, but was
-        # already done) - if it's unsafe-serialized (i.e. not safetensors), not need for anything either
+        # If it's unsafe-serialized (i.e. not safetensors), no need for anything
         if is_unsafe_serialization:
             return {full_layer_name: weight}
         # Sanity check for the new serialization format
-        elif not (TORCHAO_VERSION >= version.parse("0.14.0") and is_metadata_torchao(self.hf_quantizer.metadata)):
-            # print("metadata", self.hf_quantizer.metadata)
+        elif not (TORCHAO_VERSION >= version.parse("0.15.0") and is_metadata_torchao(self.hf_quantizer.metadata)):
             raise ValueError("To use `safetensors` serialization, you should have `torchao>=0.14.0` installed")
```
jerryzh168: nit: 0.14.0 -> 0.15.0 to match code

src/transformers/integrations/torchao.py
```diff
@@ -215,56 +215,31 @@ def convert(
...
-            if isinstance(input_dict["weight:scale"], list):
-                param_data[f"{full_layer_name}:scale"] = input_dict["weight:scale"][0]
-            else:
-                param_data[f"{full_layer_name}:scale"] = input_dict["weight:scale"]
-
-            if is_int_4:
-                if isinstance(input_dict["weight:zero_point"], list):
-                    param_data[f"{full_layer_name}:zero_point"] = input_dict["weight:zero_point"][0]
+            for suffix in input_dict.keys():
+                if isinstance(input_dict[suffix], list):
```
jerryzh168: can you add some comment? when is this a list?

src/transformers/integrations/torchao.py
```diff
@@ -215,56 +215,31 @@ def convert(
...
         if is_unsafe_serialization:
             return {full_layer_name: weight}
         # Sanity check for the new serialization format
-        elif not (TORCHAO_VERSION >= version.parse("0.14.0") and is_metadata_torchao(self.hf_quantizer.metadata)):
-            # print("metadata", self.hf_quantizer.metadata)
+        elif not (TORCHAO_VERSION >= version.parse("0.15.0") and is_metadata_torchao(self.hf_quantizer.metadata)):
             raise ValueError("To use `safetensors` serialization, you should have `torchao>=0.14.0` installed")
 
-        new_param = unflatten_tensor_state_dict(param_data, self.hf_quantizer.metadata)[full_layer_name]
+        unflattened_state_dict, _ = unflatten_tensor_state_dict(param_data, self.hf_quantizer.metadata)
```
jerryzh168: should we assert the second result to be empty here?

liangel-02:
cc @SunMarc @MekkCyber

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -91,6 +92,11 @@ def _linear_extra_repr(self):
     SUPPORTED_SAFE_SERIALIZATION_CONFIGS = [
         torchao.quantization.Float8WeightOnlyConfig,
         torchao.quantization.Float8DynamicActivationFloat8WeightConfig,
+        torchao.quantization.Int4WeightOnlyConfig,
+        torchao.quantization.IntxWeightOnlyConfig,
+        torchao.quantization.Int8DynamicActivationIntxWeightConfig,
+        torchao.quantization.ModuleFqnToConfig,
```
jerryzh168: nit: maybe add `torchao.quantization.FqnToConfig` as well, that's going to replace `ModuleFqnToConfig` I think

Also after the PR, can we turn on safe tensor serialization by default for all non-prototype configs now, or how far are we from that?

tests/quantization/torchao_integration/test_torchao.py
```diff
@@ -676,7 +676,7 @@ def test_autoquant(self):
 
 
 @require_torchao_version_greater_or_equal("0.11.0")
-@slow
+# @slow
 class TorchAoSerializationTest(unittest.TestCase):
```
MekkCyber:
```suggestion
@slow
class TorchAoSerializationTest(unittest.TestCase):
```

src/transformers/core_model_loading.py
```diff
@@ -860,7 +887,15 @@ def convert_and_load_state_dict_in_model(
...
                 future = spawn_materialize(thread_pool, tensor, param_device, _dtype)
 
-            mapping.add_tensor(renamed_key, original_key, source_pattern, future)
+            concrete_source_pattern = source_pattern
+            if isinstance(mapping, WeightConverter) and source_pattern is not None and "*" in source_pattern:
+                pattern_with_captures = source_pattern.replace("*", r"(.*?)")
+                pattern_regex = re.compile(f"^{pattern_with_captures}$")
+                concrete_source_pattern = extract_concrete_key_from_regex_pattern(
+                    original_key, source_pattern, pattern_regex
+                )
```
MekkCyber: why do we need this ? could you explain a bit more why we need to change the regex handling here and in the other parts of the code?

MekkCyber (commented):
Thanks a lot for working on this

src/transformers/core_model_loading.py
liangel-02: for the torchao `WeightConverter`, the original code has hardcoded tensor data components (ie. 
`weight_qdata`, `weight_scale`) mapped to the consolidated `weight`. however, these components change depending on the config used, with some also being optional. for max generality, i wanted to use wildcard matching (`*_weight_*` --> `*weight*`).

but with the current regex handling, the code was renaming the key with the literal "`*weight*`" rather than matching the regex. the changes i have extracts the prefix and uses that for source/target. lmk if this is an ok approach/if theres a better solution

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -241,6 +234,8 @@ def update_unexpected_keys(self, model, unexpected_keys: list[str]) -> list[str]
         return [k for k in unexpected_keys if not any(k.endswith(x) for x in self.full_ao_keys)]
 
     def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **kwargs) -> bool:
+        if "_weight_" in param_name:
+            return True
```
jerryzh168: is this going to work in all cases?

liangel-02:
@MekkCyber to avoid adding torchao specific logic to the core model loader and unblock landing, i reverted the changes to the regex logic and added all other tensor data names. in the future, maybe we can implement a strategy that generalizes this matching in a better way?

cc @jerryzh168

src/transformers/integrations/torchao.py
```diff
@@ -215,56 +215,46 @@ def convert(
...
+                "_weight_qdata": torch.Tensor,
+                "_weight_scale": torch.Tensor,
+            }
+            full_layer_name: "model.layers.0.self_attn.k_proj"
+
+            Given this, we reconstruct a Float8Tensor instance using the qdata and scale
+            and return it as a dictionary with the full_layer_name as the key and the recovered
+            Float8Tensor instance as the value.
+        """
+        is_unsafe_serialization = "_weight_" not in list(input_dict.keys())[0]
```
jerryzh168: this is using the same logic, we can probably make this more robust as well

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -249,8 +244,6 @@ def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **
         # check if the param_name is not in self.modules_to_not_convert
         if any(key + "." in param_name or key == param_name for key in self.modules_to_not_convert):
             return False
-        elif any(param_name.endswith(f":{x}") for x in self.full_ao_keys):
```
jerryzh168, line 252: for this one, maybe just change it to `(f"_{x} for x in self.full_ao_keys)"` to be safe, at least with this we are confident it's correct. if there are no better general way to detect safetensors

liangel-02: i think this is achieving the same thing as L237, i think we only need one check

src/transformers/quantizers/quantizer_torchao.py
liangel-02: i think i can remove this actually

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -548,15 +536,18 @@ def get_weight_conversions(self):
...
+                    # TODO: incr flexibility by generalizing the source patterns to match the format of "_weight_"
+                    # note that the matching logic is greedy, so for ex, if _weight_scale is before _weight_scale_and_zero in this list, it will match _weight_scale always (this is incorrect)
+                    # thus, the order of source_patterns is intentional
+                    source_patterns=[
+                        "_weight_qdata",
+                        "_weight_scale_and_zero",
+                        "_weight_scale",
+                        "_weight_zero_point",
+                        "_weight_act_pre_scale",
+                    ],
```
MekkCyber, line 548: Yes it is greedy but to match for example `_weight_scale` and not `_weight_scale_and_zero` you can just do something like `_weight_scale$`, but ordering the keys works as well 👍

src/transformers/modeling_utils.py
```diff
@@ -2104,7 +2104,6 @@ def set_decoder(self, decoder):
         possible_module_names = ["language_model", "text_model", "decoder"]
         for name in possible_module_names:
             if hasattr(self, name):
-                print(name)
                 setattr(self, name, decoder)
```
MekkCyber, line 2108: Thanks for catching this

src/transformers/integrations/torchao.py
```diff
@@ -210,61 +210,52 @@ def __init__(self, hf_quantizer):
...
                 weight = input_dict["weight"][0]
             else:
                 weight = input_dict["weight"]
         else:
-            if isinstance(input_dict["weight:qdata"], list):
-                param_data[f"{full_layer_name}:qdata"] = input_dict["weight:qdata"][0]
-            else:
-                param_data[f"{full_layer_name}:qdata"] = input_dict["weight:qdata"]
+            for suffix in input_dict.keys():
+                assert len(input_dict[suffix]) == 1
```
MekkCyber: let's do an if/else and raise an error to follow the same pattern we use in transformers, but i'm not sure if it's necessary since if we have the `suffix` in `input_dict` it means we collected at least a tensor for that suffix

src/transformers/integrations/torchao.py
```diff
@@ -210,61 +210,52 @@ def __init__(self, hf_quantizer):
...
+        """
+        Consolidates tensor subclass components before reconstructing the object
+
+        For example:
+            input_dict: {
+                "_weight_qdata": torch.Tensor,
+                "_weight_scale": torch.Tensor,
+            }
+            full_layer_name: "model.layers.0.self_attn.k_proj"
+
```
MekkCyber:
```suggestion
            full_layer_name: "model.layers.0.self_attn.k_proj.weight"

```

MekkCyber (approved):
Very nice alternative solution ! Thanks a lot

src/transformers/integrations/torchao.py
```diff
@@ -210,61 +210,52 @@ def __init__(self, hf_quantizer):
...
+            Given this, we reconstruct a Float8Tensor instance using the qdata and scale
+            and return it as a dictionary with the full_layer_name as the key and the recovered
+            Float8Tensor instance as the value.
+        """
+        is_unsafe_serialization = list(input_dict.keys())[0] not in source_patterns
 
         param_data = {}
+        layer_name = ".".join(full_layer_name.split(".")[:-1])
```
SunMarc: Let's use rsplit as we do for other methods 
```suggestion
        layer_name = full_layer_name.rsplit(".", 1)[0]
```

src/transformers/integrations/torchao.py
liangel-02: will add an if/else! we wanted to add this assert since there also shouldn't be more than one tensor for a component (qdata, scale, etc)

liangel-02:
@MekkCyber @SunMarc can you help me merge this, thanks!

ArthurZucker (approved):
LGTM

PR #42531, opened 2025-12-01
Fix Qwen3OmniMoE weight init

Rocketknight1:
In #42456 we renamed `self.router` to `self.gate` in some layers but the `_init_weights` for `Qwen3OmniMoE` still refers to router. This PR fixes it!


ArthurZucker (approved):
please update the modular

Rocketknight1:
today is not a smart day for me

PR #42535, opened 2025-12-01
Fix ernie moe

vasqu:
As per title

### DIFF
--- a/src/transformers/models/ernie4_5_moe/modeling_ernie4_5_moe.py
+++ b/src/transformers/models/ernie4_5_moe/modeling_ernie4_5_moe.py
@@ -318,51 +318,41 @@ class Ernie4_5_MoeStatics(nn.Module):
 
 
 class Ernie4_5_MoeExperts(nn.Module):
+    """Collection of expert weights stored as 3D tensors."""
+
     def __init__(self, config):
         super().__init__()
         self.num_experts = config.moe_num_experts
         self.hidden_dim = config.hidden_size
         self.intermediate_dim = config.moe_intermediate_size
-        self.use_bias = config.use_bias
+        self.gate_up_proj = nn.Parameter(torch.empty(self.num_experts, 2 * self.intermediate_dim, self.hidden_dim))
+        self.down_proj = nn.Parameter(torch.empty(self.num_experts, self.hidden_dim, self.intermediate_dim))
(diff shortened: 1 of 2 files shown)


src/transformers/models/ernie4_5_moe/modeling_ernie4_5_moe.py
```diff
@@ -318,51 +318,41 @@ def forward(self, hidden_states):
...
         self.act_fn = ACT2FN[config.hidden_act]
 
-        self.gate_up_proj = nn.Parameter(torch.zeros(self.num_experts, 2 * self.intermediate_dim, self.hidden_dim))
-        self.down_proj = nn.Parameter(torch.zeros(self.num_experts, self.hidden_dim, self.intermediate_dim))
-        if self.use_bias:
-            self.gate_up_proj_bias = nn.Parameter(torch.zeros(self.num_experts, 2 * self.intermediate_dim))
-            self.down_proj_bias = nn.Parameter(torch.zeros(self.num_experts, self.hidden_dim))
-        else:
-            self.gate_up_proj_bias = None
-            self.down_proj_bias = None
```
vasqu, line 336: That bias was not necessary neither use a bias

src/transformers/models/ernie4_5_moe/modeling_ernie4_5_moe.py
```diff
@@ -383,14 +373,14 @@ def forward(self, hidden_states: torch.Tensor) -> tuple[torch.Tensor, torch.Tens
...
-                routing_weights.sum(dim=-1, keepdim=True), min=self.norm_min
+            router_logits = F.softmax(router_logits, dim=1, dtype=torch.float)
+            router_top_value, router_indices = torch.topk(self.moe_statics(router_logits), self.top_k, dim=-1)
+            router_top_value = router_top_value / torch.clamp(
+                router_top_value.sum(dim=-1, keepdim=True), min=self.norm_min
             )
-        routing_weights = routing_weights.to(router_logits.dtype)
-        return routing_weights, selected_experts
+            router_scores = router_top_value
+        router_scores = router_scores.to(hidden_states.dtype)
```
vasqu, line 382: This dtype casting is necessary, because we need half here (which would only be in the og states

ArthurZucker (approved):
Ok let's go

PR #42536, opened 2025-12-01
extend FA2 and other cases to XPU, 

yao-matrix:
we expect all model cases except CUDAGraph specific, CUDA compute capability specific and FA3 specific can run XPU. For FA3, we are developing.

@ydshieh, pls help review, thx very much.


src/transformers/models/mimi/modeling_mimi.py
```diff
@@ -931,6 +931,7 @@ def forward(
 
 MIMI_ATTENTION_CLASSES = {
     "eager": MimiAttention,
+    "kernels-community/flash-attn2": MimiFlashAttention2,
```
ydshieh: could you explain this part?

tests/generation/test_continuous_batching.py
```diff
@@ -315,36 +314,47 @@ def test_continuous_batching_parity_gemma_sdpa(self) -> None:
...
+    @require_torch_accelerator
     @require_kernels
     @slow
     def test_continuous_batching_parity_qwen_flash(self) -> None:
-        expected_outputs = {}
+        expected_outputs = Expectations({
+            ("xpu", None): {
+                "req_1":  " 3.5 bolts.\n\nLet's break it down step by step:\n\n- Blue fiber: 2 bolts\n- White fiber: half of 2 bolts = 1 bolt\n\nTotal = ",
+            },
+        }).get_expectation()  # fmt: skip
```
ydshieh, line 354: i need to check why this was `{}` before, but thank you.

ydshieh (approved):
Thank you! LGTM, but has one question

ydshieh:
@yao-matrix don't forget 

https://github.com/huggingface/transformers/pull/42536#discussion_r2580336110

🙏

src/transformers/models/mimi/modeling_mimi.py
yao-matrix: @ydshieh , sure. in latest design, when users set `attn_implementation == "flash_attention_2"`, there will be 2 branches:
1) if `flash_attn` package is available, it will go directly to use it
2) else, do not fail as before, but use kernels instead, in this case, the `attn_implementation` will be updated to `"kernels-community/flash-attn2"`, as in [code](https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L1779) here

For XPU, we go with the kernels path in transformers for FA support, so we need this key. 

Thx very much.

yao-matrix:
> @yao-matrix don't forget
> 
> [#42536 (comment)](https://github.com/huggingface/transformers/pull/42536#discussion_r2580336110)
> 
> 🙏

Yes, done, thx very much for your always support, :).

src/transformers/models/mimi/modeling_mimi.py
vasqu: I'd rather not do this, even tho you are correct here. We should rather refactor mimi here with the attention interface and not have these manual registrations. We could infinitely extend these edge cases in the future to FA3 etc which makes this not scalable (without using/refactoring to the interface).

ydshieh: @yao-matrix 

Let's revert this line 🙏 . We can skip the relevant FA tests if necessary.

yao-matrix: done, I am not skipping this case, let's figure out a grace solution later, thx.

yao-matrix:
@ydshieh , pls take a second look, thx very much.

tests/generation/test_continuous_batching.py
```diff
@@ -315,36 +314,47 @@ def test_continuous_batching_parity_gemma_sdpa(self) -> None:
...
+            },
         }).get_expectation()  # fmt: skip
         self._continuous_batching_parity("google/gemma-2-2b-it", "paged|flash_attention_2", expected_outputs)
 
-    @require_torch_gpu
+    @require_torch_accelerator
     @require_kernels
     @slow
     def test_continuous_batching_parity_qwen_flash(self) -> None:
-        expected_outputs = {}
```
ydshieh, line 344: @remi-or Could you add code comment and/or docstring for `_continuous_batching_parity` to explain what 

> expected_outputs = {}

is for?

remi-or: Sure! Was actually thinking of doing a test-oriented PR.

PR #42538, opened 2025-12-01
[docs] optimizations quickstart

stevhliu:
Adds an overview/quickstart of available Transformers optimization techniques to provide a clear and centralized place where they're all documented and help users select the right optimization to increase speed or reduce memory footprint

### DIFF
--- a/docs/source/en/optimization_overview.md
+++ b/docs/source/en/optimization_overview.md
@@ -27,9 +27,9 @@ Use the table below to pick an optimization technique.
 |---|:---:|:---:|
 | [Compilation](#compilation) | ✅ | |
 | [Attention backends](#attention-backends) | ✅ | ✅ |
-| [Kernels](#kernels) | ✅ | |
+| [Kernels](#kernels) | ✅ | ✅ |
 | [Quantization](#quantization) | ✅ | ✅ |
-| [Caching](#caching) | ✅ | |
+| [Caching](#caching) | ✅ | ✅ |
 | [Parallelism](#parallelism) | ✅ | |
 | [Continuous batching](#continuous-batching) | ✅ | |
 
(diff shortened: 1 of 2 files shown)


docs/source/en/optimization_overview.md
```diff
@@ -0,0 +1,185 @@
...
+
+| Technique | Speed | Memory |
+|---|:---:|:---:|
+| [Compilation](#compilation) | ✅ | |
+| [Attention backends](#attention-backends) | ✅ | ✅ |
+| [Kernels](#kernels) | ✅ | |
+| [Quantization](#quantization) | ✅ | ✅ |
+| [Caching](#caching) | ✅ | |
+| [Parallelism](#parallelism) | ✅ | |
+| [Continuous batching](#continuous-batching) | ✅ | |
```
Cyrilvallez, line 34: This one is tricky as kernels can usually lead to less memory usage as well! And for caching, we have some tricks to use less as well, e.g. offloading

docs/source/en/optimization_overview.md
```diff
@@ -0,0 +1,185 @@
...
+[torch.compile](./perf_torch_compile) reduces Python overhead, fuses operations, and creates kernels tuned for your shapes and hardware. The first run warms it up and subsequent runs use the faster compiled path.
+
+Call `torch.compile()` on a model to enable it.
+
+```py
+import torch
+from transformers import AutoModelForCausalLM
+
+model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen3-0.6B")
+compiled_model = torch.compile(model)
```
Cyrilvallez: Recommended way is NOT to compile the model outside like this! Otherwise we keep recompiling all the time. Instead, use it directly inside `generate` with a compileable cache, e.g. `cache_implementation="static"` triggers compile

docs/source/en/optimization_overview.md
```diff
@@ -0,0 +1,185 @@
...
+import torch
+from transformers import AutoTokenizer, AutoModelForCausalLM
+
+tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen3-0.6B")
+model = AutoModelForCausalLM.from_pretrained(
+    "Qwen/Qwen3-0.6B", attn_implementation="kernels-community/flash-attn2"
+)
+inputs = tokenizer("The Le Décret Pain states that a baguette must,", return_tensors="pt")
+outputs = model.generate(**inputs, do_sample=False, max_new_tokens=50, cache_implementation="static")
+```
```
Cyrilvallez: Maybe talk about offloading instead `cache_implementation="offloaded"` for reduced memory usage? Cause this one is more related to torch.compile IMO

Cyrilvallez (commented):
Very nice! Sorry for the delay, added some comments!

Cyrilvallez (approved):
Nice! cc @LysandreJik as well if you want to have a look/have more comments!

docs/source/en/optimization_overview.md
```diff
@@ -0,0 +1,191 @@
...
+rendered properly in your Markdown viewer.
+
+-->
+
+# Overview
+
+Transformers provides multiple inference optimization techniques to make models fast, affordable, and accessible. Options include alternative attention mechanisms for reduced memory traffic, code compilation for faster execution, and optimized kernels for throughput. Combine these techniques for maximum performance.
+
+> [!NOTE]
+> Memory and speed are closely related but not the same. Shrinking your memory footprint makes a model "faster" because there is less data to move around. Pure speed optimizations don't always reduce memory and sometimes increase usage. Choose the appropriate optimization based on your use case and hardware.
```
LysandreJik, line 22: Cool note

LysandreJik (approved):
Super good start! Love it. Let's get it in and iterate over it then

cc @SunMarc as well, related to some topics we discussed recently.

docs/source/en/optimization_overview.md
```diff
@@ -0,0 +1,191 @@
...
+```py
+from transformers import AutoModelForCausalLM
+from transformers.distributed.configuration_utils import DistributedConfig
+
+distributed_config = DistributedConfig(enable_expert_parallel=True)
+model = AutoModelForCausalLM.from_pretrained(
+    "openai/gpt-oss-120b",
+    distributed_config=distributed_config,
+)
+```
```
SunMarc: Not sure if we want to promote this API for now as it is not that used for now. Since we are refactoring moe and more, maybe it is a good time to fix how `distributed_config` will work (tp, ep, pp) ? cc @ArthurZucker @3outeille. Right now, the only model that is using this feature is llama4.

SunMarc (approved):
Thanks that's really nice ! Just left a minor comment

docs/source/en/optimization_overview.md
stevhliu: sounds good, i can update docs for expert parallelism in #42409 once the API is more stable

PR #42545, opened 2025-12-02
[Quantization] fix dequant when block size is none & static quantization

MekkCyber:
# What does this PR do?

This pr fixes the dequantization when no block size is provided by assuming it's a per tensor quantization. 
It also fixes the input scales not being expected in the dequantization case.


ArthurZucker (approved):
Niceee! Let's add a bit of doc on the new method !

SunMarc (approved):
Thanks !

PR #42547, opened 2025-12-02
Fixed paged|FA2 kernel loading logic and UT.

YangKai0616:
# What does this PR do?

**1. [General]** When running the test case `tests/generation/test_continuous_batching.py::ContinuousBatchingTest::test_continuous_batching_parity_qwen_flash`, it reports error as follows: 
```
                flash_attn_func = getattr(kernel, "flash_attn_func", None)
                flash_attn_varlen_func = getattr(kernel, "flash_attn_varlen_func", None)
                if flash_attn_varlen_func is None:
>                   raise ValueError(
                        f"Could not find the currently requested flash attention implementation at `{implementation}`."
                        "Make sure that you request a valid kernel from the hub, e.g. `kernels-community/flash-attn2`."
                    )
E                   ValueError: Could not find the currently requested flash attention implementation at `flash_attention_2`.Make sure that you request a valid kernel from the hub, e.g. `kernels-community/flash-attn2`.

src/transformers/modeling_flash_attention_utils.py:115: ValueError
```
The root cause is a kernel loading failure of `paged|flash_attention_2`. This PR fixes the issue.

**2. [XPU]** The above test case reports error `E   AssertionError: Test request_id = 'req_1' failed, no expected output was provided.....` after successfully running on both XPU and CUDA. However, on the XPU side, the test case can stably pass by using `require_deterministic_for_xpu`.

For enabling the related tests on XPU, please refer to PR https://github.com/huggingface/transformers/pull/42536.


YangKai0616:
@vasqu , please help review, thanks!

src/transformers/modeling_utils.py
```diff
@@ -1764,9 +1764,12 @@ def _check_and_adjust_attn_implementation(
         """
         applicable_attn_implementation = attn_implementation
 
+        is_paged = attn_implementation is not None and attn_implementation.startswith("paged|")
+        actual_attn_impl = attn_implementation.split("|", 1)[1] if is_paged else attn_implementation
```
vasqu:
```suggestion
        actual_attn_implementation = attn_implementation.split("|", 1)[1] if is_paged else attn_implementation
```
nit

src/transformers/modeling_utils.py
```diff
@@ -1775,7 +1778,7 @@ def _check_and_adjust_attn_implementation(
             and is_kernels_available()
             and not is_torch_npu_available()
         ):
-            if attn_implementation.endswith("2"):
+            if actual_attn_impl.endswith("2"):
```
vasqu: We could check after the if else, if it is a paged version and directly prepend. 
```python
            if attn_implementation.endswith("2"):
                applicable_attn_implementation = "kernels-community/flash-attn2"
                if is_torch_xpu_available():
                    # On XPU, kernels library is the native implementation
                    # Disabling this flag to avoid giving wrong fallbacks on errors and warnings
                    requested_original_flash_attn = False
            else:
                applicable_attn_implementation = "kernels-community/vllm-flash-attn3"

            if is_paged:
                applicable_attn_implementation = f"paged|{applicable_attn_implementation}"
```

This will force us to properly register the kernel attention then but avoids loading the kernel 2 times (as I suspect currently it would load it again within the CB API because it's not registered with "paged"). So, in essence a bit down below after checking if it is a kernel, we can do
```python
        if is_kernel(applicable_attn_implementation):
            try:
                # preload flash attention here to allow compile with fullgraph
                if is_paged:
                   lazy_import_paged_flash_attention(applicable_attn_implementation)
                else:
                   lazy_import_flash_attention(applicable_attn_implementation)
```
It's only really important for kernels, because they need to register with the correct attention wrapper function (which we do with the paged import version).

vasqu (commented):
I would like to extend this a bit to allow general CB support for kernels without 2x loading the kernel. 

The idea is to modify the fallback to include the `paged|` prefix directly + load the kernel properly via `lazy_import_paged_flash_attention`. This should allow us to use the prefix version directly as well, e.g. `attn_implementation="paged|kernels-community/flash-attn2"`

src/transformers/modeling_utils.py
YangKai0616: Done.

src/transformers/modeling_utils.py
YangKai0616: Done

src/transformers/modeling_utils.py
```diff
@@ -1764,9 +1764,12 @@ def _check_and_adjust_attn_implementation(
         """
         applicable_attn_implementation = attn_implementation
 
+        is_paged = attn_implementation is not None and attn_implementation.startswith("paged|")
+        actual_attn_implementation = attn_implementation.split("|", 1)[1] if is_paged else attn_implementation
```
vasqu: On second thought, I'd like to avoid introducing another name here, i.e. we would have `attn_implementation`, `actual_attn_implementation`, and `applicable_attn_implementation`.

I'll try to go a bit through below for those changes

src/transformers/modeling_utils.py
```diff
@@ -1764,9 +1764,12 @@ def _check_and_adjust_attn_implementation(
...
+
         # If FA not installed, do not fail but use kernels instead
-        requested_original_flash_attn = attn_implementation is not None and (
-            attn_implementation == "flash_attention_2" or attn_implementation == "flash_attention_3"
+        requested_original_flash_attn = actual_attn_implementation is not None and (
+            actual_attn_implementation == "flash_attention_2" or actual_attn_implementation == "flash_attention_3"
         )
```
vasqu:
```suggestion
        requested_original_flash_attn = actual_attn_implementation is not None and (
            attn_implementation.removeprefix("paged|") == "flash_attention_2" or attn_implementation.removeprefix("paged|") == "flash_attention_3"
        )
```
This way, we can use the old implementation. Doesnt matter if it had the prefix or not

src/transformers/modeling_utils.py
```diff
@@ -1775,7 +1778,7 @@ def _check_and_adjust_attn_implementation(
             and is_kernels_available()
             and not is_torch_npu_available()
         ):
-            if attn_implementation.endswith("2"):
+            if actual_attn_implementation.endswith("2"):
```
vasqu: Can be reverted as we only care about the last symbol, not anything before (as we checked that already)

src/transformers/modeling_utils.py
```diff
@@ -1798,7 +1807,7 @@ def _check_and_adjust_attn_implementation(
             except Exception as e:
                 # raise the proper exception for requested flash attention
                 if requested_original_flash_attn:
-                    if attn_implementation.endswith("2"):
+                    if actual_attn_implementation.endswith("2"):
```
vasqu: Same here, can be reverted

vasqu (commented):
Let's simplify a bit, introducing another variable will be even more confusing on second thought. We can avoid that

src/transformers/modeling_utils.py
YangKai0616: I have reverted here to `attn_implementation` and deleted `actual_attn_implementation`.

src/transformers/modeling_utils.py
YangKai0616: Done.

src/transformers/modeling_utils.py
YangKai0616: Done.

src/transformers/modeling_utils.py
YangKai0616: Done.

YangKai0616:
Strange, the CI failing example `tests/models/mvp/test_modeling_mvp.py::MvpHeadTests::test_generate_beam_search` can pass on both CUDA (A100) and XPU for me (with torch 2.9).

vasqu:
Yup, not sure what's going on. It's an annoying flaky test. Added a new commit to simplify something, let's see if CI passes this time otherwise I'll try to get a core maintainer to merge if necessary.

vasqu (approved):
Forgot to approve but lgtm now

YangKai0616:
Hey @vasqu , when I run the command `pytest -rA tests/test_tokenization_mistral_common.py::TestMistralCommonBackend::test_all_special_ids tests/test_tokenization_mistral_common.py::TestMistralCommonBackend::test_all_special_tokens tests/test_tokenization_mistral_common.py::TestMistralCommonBackend::test_apply_chat_template_basic` on A100 and XPU, I can’t reproduce the CI error. Could you able to reproduce it?

ydshieh:
merge as requested by @vasqu 

(irrelevant failures)

PR #42549, opened 2025-12-02
Fix the FA2 logic in the longcat_flash model

YangKai0616:
# What does this PR do?

FA2 does not support MLA (i.e., cases where the dimensions of Q, K, and V heads are inconsistent), so skip this test.


YangKai0616:
@vasqu , please help review. Thanks!

tests/models/longcat_flash/test_modeling_longcat_flash.py
```diff
@@ -297,6 +297,13 @@ def test_flash_attn_2_fp32_ln(self):
             if not model_class._supports_flash_attn:
                 self.skipTest(f"{model_class.__name__} does not support Flash Attention 2")
             config, inputs_dict = self.model_tester.prepare_config_and_inputs_for_common()
+
+            if config.qk_head_dim != config.v_head_dim:
+                self.skipTest(
+                    reason="Flash Attention 2 requires qk_head_dim == v_head_dim, but got "
+                    f"qk_head_dim={config.qk_head_dim}, v_head_dim={config.v_head_dim}"
+                )
```
vasqu: Will this not skip all tests here? I doubt that the classes will have different head dims.

I'd rather we properly adjust the sizes than to skip - should always be the last resort.

YangKai0616: Yes, this test only involves this single `model_class`. After digging deeper into `LongcatFlashForCausalLM`, I found that it already implements the padding pre-processing for FA2 internally. However, for the fallback FA2 path(kernels), it failed to correctly match the parameter naming. This PR updates that part with the fix. Please review it again!

src/transformers/models/longcat_flash/modular_longcat_flash.py
```diff
@@ -215,7 +215,10 @@ def forward(
             cache_kwargs = {"sin": sin, "cos": cos, "cache_position": cache_position}
             key_states, value_states = past_key_values.update(key_states, value_states, self.layer_idx, cache_kwargs)
 
-        if self.config._attn_implementation == "flash_attention_2" and self.qk_head_dim != self.v_head_dim:
+        uses_flash_attention_2 = (
+            "flash" in self.config._attn_implementation and self.config._attn_implementation.endswith("2")
+        )
+        if uses_flash_attention_2 and self.qk_head_dim != self.v_head_dim:
```
vasqu:
```suggestion
        if "flash" in self.config._attn_implementation and self.qk_head_dim != self.v_head_dim:
```
I think we should generalize this here to check for all flavors. FA3 etc would face the same issue

src/transformers/models/longcat_flash/modular_longcat_flash.py
```diff
@@ -233,7 +236,7 @@ def forward(
             **kwargs,
         )
 
-        if self.config._attn_implementation == "flash_attention_2" and self.qk_head_dim != self.v_head_dim:
+        if uses_flash_attention_2 and self.qk_head_dim != self.v_head_dim:
```
vasqu:
```suggestion
        if "flash" in self.config._attn_implementation and self.qk_head_dim != self.v_head_dim:
```
Same here then

vasqu (commented):
Thx for iterating! I think we can generalize this some more to include all attentions? They will probably face similar issues

src/transformers/models/longcat_flash/modular_longcat_flash.py
YangKai0616: Great! Done.

src/transformers/models/longcat_flash/modular_longcat_flash.py
YangKai0616: Done.

vasqu (approved):
Perfect, let's merge

PR #42554, opened 2025-12-02
Refactor: Remove redundant else in get_activation

leaderofARS:
Hi! I was reading through 
activations.py
 to better understand how the library handles activation mappings, and I noticed a small opportunity to clean up the 
get_activation
 function.

Since the if block returns early, the else block isn't actually needed. I've removed it to flatten the logic a bit. I know it's a tiny change, but I figured it was worth polishing up to keep the code readable.


Rocketknight1:
Hey! Can you run `pip install -e .[quality]` followed by `make style` to get the code quality tests to pass?

PR #42557, opened 2025-12-02
Fix eetq quanto quant methods

SunMarc:
# What does this PR do?

This PR fixes EETQ and quanto integration ! We switched to kernels for eetq and we simplified a bit quanto. All tests are passing.


SunMarc:
@bot /style

src/transformers/integrations/eetq.py
```diff
@@ -25,91 +27,104 @@
...
+
+
+class EetqLinearMMFunction(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, x, weight, scales, bias=None):
+        # The forward pass can use ctx.
+        ctx.save_for_backward(x, weight, scales, bias)
+        output = eetq_kernels_hub.w8_a16_gemm(x, weight, scales)
+        output = output + bias if bias is not None else output
+        return output
```
MekkCyber, line 57: very nice to have the linear class here !

src/transformers/integrations/eetq.py
```diff
@@ -25,91 +27,104 @@
...
+
+class EetqLinear(nn.Module):
+    def __init__(self, in_features, out_features, dtype=torch.int8, bias=False):
+        super().__init__()
+        self.register_buffer("weight", torch.zeros((in_features, out_features), dtype=dtype))
+        if bias:
+            self.register_buffer("bias", torch.zeros((out_features), dtype=torch.float16))
+        else:
+            self.bias = None
+        self.register_buffer("weight_scales", torch.zeros((out_features), dtype=torch.float16))
```
MekkCyber: maybe we can use register them as parameters, since we are moving away from using buffers

src/transformers/integrations/quanto.py
```diff
@@ -12,21 +12,51 @@
...
+        _load_parameter_into_model(model, full_layer_name, value)
+        module, _ = get_module_from_name(model, full_layer_name)
+        module.freeze()
+        module.weight.requires_grad = False
+
+        # need to discard some missing keys we already updated the module in freeze. For the weight, we still pass it
+        module_name = full_layer_name.rsplit(".", 1)[0]
+        missing_keys.discard(f"{module_name}.input_scale")
+        missing_keys.discard(f"{module_name}.output_scale")
+        return {full_layer_name: module.weight}
```
MekkCyber: not sure why `_load_paramter_into_model` don't set the weights

src/transformers/integrations/quanto.py
```diff
@@ -35,64 +65,50 @@ def replace_with_quanto_layers(
...
+            continue
+        with init_empty_weights():
+            new_module = None
+            if isinstance(module, nn.Linear):
+                new_module = QLinear(
+                    in_features=module.in_features,
+                    out_features=module.out_features,
+                    bias=module.bias is not None,
+                    dtype=module.weight.dtype,
+                    weights=w_mapping[quantization_config.weights],
```
MekkCyber, line 93: why are we not using the same trick for the dtype ?

MekkCyber (approved):
Really love how kernels, and the refactor are making the code much more cleaner! Thanks for working on this

src/transformers/integrations/quanto.py
SunMarc: it's setting, just sending them anyway. shouldn't be an issue

src/transformers/integrations/quanto.py
SunMarc: the dtype is cleaning defined there with dtype=module.weight.dtype. We don't need the same trick as we don't need the deal with loading pre-quantized model

src/transformers/integrations/quanto.py
ArthurZucker: lets return empty dict to show its already loaded

ArthurZucker (approved):

src/transformers/integrations/quanto.py
SunMarc: ok done !

PR #42542, opened 2025-12-02
handle get_input_embeddings() on models like gemma3 gracefully

winglian:
# What does this PR do?

#41993 breaks calls to `model.enable_input_require_grads()`

```
stderr: [rank1]:   File "/home/wing/axolotl-dev/src/axolotl/loaders/adapter.py", line 188, in load_adapter
stderr: [rank1]:     model.enable_input_require_grads()
stderr: [rank1]:   File "/home/wing/.venvs/axolotl/lib/python3.11/site-packages/transformers/modeling_utils.py", line 1986, in enable_input_require_grads
stderr: [rank1]:     input_embeddings = module.get_input_embeddings()          
stderr: [rank1]:                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                        
stderr: [rank1]:   File "/home/wing/.venvs/axolotl/lib/python3.11/site-packages/transformers/modeling_utils.py", line 993, in get_input_embeddings
stderr: [rank1]:     raise NotImplementedError(
stderr: [rank1]: NotImplementedError: `get_input_embeddings` not auto‑handled for SiglipVisionTransformer; please override in the subclass.
```

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #42558, opened 2025-12-02
Make gradient-checkpoint enabling tolerant of models without get_input_embeddings

molbap:
# What does this PR do?

Ad title indicates, #42542 and likely a few other models are broken by merged #41993 . This adds an embedding getter and attempts to test the feature with more coverage.

Basically what it does 

 - Stop hard-failing gradient_checkpointing_enable when a model lacks a get_input_embeddings. We now just call enable_input_require_grads, let it attach hooks where it can, and issue a single warning if no embedding module is found.
 - Simplify enable_input_require_grads (and the InternVL/MLCD and a couple more model overrides/adjustments) by making them responsible for the warning.
-  Adds a big test to make sure all of that works (please take a look)

Should help GC for PEFT adapters for many VLMs hopefully (and normal models too)


molbap:
added a test as well - but can't find a clean way around the models for which it is _not_ relevant to have a getter method and _not_ causing as many side-effects. WDYT @zucchini-nlp ? kind of stumped (try/excepting at higher level would always work but hides a lot)

src/transformers/modeling_utils.py
```diff
@@ -961,59 +961,78 @@ class EmbeddingAccessMixin:
...
-        else:
-            base_model = getattr(self, "base_model_prefix", None)
-            if base_model is not None:
-                base_model = getattr(self, base_model, None)
-                if base_model is not None and base_model is not self:
-                    return base_model.get_input_embeddings()
+
+        base_model = getattr(self, "base_model_prefix", None)
+        if base_model is not None:
+            base_model = getattr(self, base_model, None)
```
zucchini-nlp: nit: `self.base_model` property has the same functionality

src/transformers/modeling_utils.py
```diff
@@ -961,59 +961,78 @@ class EmbeddingAccessMixin:
 
     _input_embed_layer = "embed_tokens"  # default layer that holds input embeddings.
 
-    def get_input_embeddings(self) -> nn.Module:
+    def _get_input_embeddings_no_raise(self) -> Optional[nn.Module]:
```
zucchini-nlp: oh interesting, I was assuming the base `get_input_embedding` already returns `None`

zucchini-nlp (commented):
> models for which it is not relevant to have a getter method and not causing as many side-effects

Would this also mean that we can't support correctly PEFT and GC with these models, or do they have a custom way to set grad on the inputs? We could raise an error with a better message saying that models doesn't support unless it has a way to get its input embeddings, wdyt?

src/transformers/modeling_utils.py
molbap: well I ended up in some many little edge cases lol

src/transformers/modeling_utils.py
molbap: true!

molbap:
Yes it's a good idea to raise/inform for downstream users. I reverted a couple things and will update the test so it actually checks that enabling GC works (probably add another test)

src/transformers/utils/import_utils.py
molbap, line 1: This is mostly to fix a broken env situation that can be caused around timm_wrapper (or timm_backbone?) so it protects a few imports

molbap:
I reverted a few models to inner positional embeddings calls as mentioned in https://github.com/huggingface/transformers/pull/38913 . 

Modified a few others models as the test I added (`test_enable_input_require_grads_with_gradient_checkpointing` ) was a bit naive and I was just `continue`-ing, now it's a proper skip if the loss is undefined. 

Hopefully that helps VLMs + GC and does not break adapters

src/transformers/modeling_utils.py
```diff
@@ -1978,14 +1978,18 @@ def make_inputs_require_grads(module, input, output):
...
 
         for module in self.modules():
             if not (isinstance(module, PreTrainedModel) and hasattr(module, "get_input_embeddings")):
                 continue
 
-            input_embeddings = module.get_input_embeddings()
+            try:
+                input_embeddings = module.get_input_embeddings()
+            except NotImplementedError:
+                continue
```
molbap, line 2053: no simple way around this unfortunately

src/transformers/modeling_utils.py
```diff
@@ -1994,11 +1998,18 @@ def make_inputs_require_grads(module, input, output):
...
 
         self._require_grads_hooks = hooks
         if hooks:
             # for BC
             self._require_grads_hook = hooks[0]
+        if not found_embeddings:
+            logger.warning_once(
+                f"{self.__class__.__name__} does not expose input embeddings. Gradients cannot flow back to the token "
+                "embeddings when using adapters or gradient checkpointing. Override `get_input_embeddings` to fully "
+                "support those features."
```
molbap: at least we can warn users!

src/transformers/models/internvl/modeling_internvl.py
```diff
@@ -209,10 +209,9 @@ def forward(self, pixel_values: torch.Tensor) -> torch.Tensor:
             )
 
         embeddings = self.projection(pixel_values)
-        patch_height, patch_width = embeddings.shape[2], embeddings.shape[3]
         embeddings = embeddings.flatten(2).transpose(1, 2)
 
-        return embeddings, (patch_height, patch_width)
+        return embeddings
```
zucchini-nlp, line 215: haha, looks like a left-over from a model shipping and debugging

src/transformers/models/altclip/modeling_altclip.py
```diff
@@ -367,7 +367,6 @@ def feed_forward_chunk(self, attention_output):
         return layer_output
 
 
-# Copied from transformers.models.align.modeling_align.AlignTextEncoder with AlignText->AltRoberta
```
zucchini-nlp, line 370: why we had to delete a copy?

src/transformers/models/mlcd/modeling_mlcd.py
```diff
@@ -333,9 +334,13 @@ def forward(
         residual = hidden_states
 
         hidden_states = self.layer_norm1(hidden_states)
+        rotary_embeddings = position_embeddings
+        if rotary_position_tensor is not None:
+            rotary_embeddings = (rotary_position_tensor.cos(), rotary_position_tensor.sin())
+
```
zucchini-nlp: the `position_embeddings` are already supposed to be present so we don't need the embeddings, isn't it?

src/transformers/modeling_utils.py
zucchini-nlp: nice!

src/transformers/modeling_utils.py
zucchini-nlp: oke, I think with the warning below, it is more explicit

src/transformers/modeling_utils.py
```diff
@@ -2940,7 +2949,10 @@ def gradient_checkpointing_enable(self, gradient_checkpointing_kwargs=None):
                 "Please update to the new format on your modeling file. To use the new format, you need to completely remove the definition of the method `_set_gradient_checkpointing` in your model."
             )
 
-        if getattr(self, "_hf_peft_config_loaded", False):
+        needs_embedding_grads = self.main_input_name == "input_ids"
+        # we use that also to detect whether or not we have to raise if embeddings are missing (the submodel might not have embeddings at all)
+        enable_input_grads = needs_embedding_grads or getattr(self, "_hf_peft_config_loaded", False)
+        if enable_input_grads:
```
zucchini-nlp, line 3015: hmm, for my understanding, why do we always need to enable grads when doing GC training with text models?

tests/test_modeling_common.py
```diff
@@ -906,6 +906,111 @@ def test_peft_gradient_checkpointing_enable_disable(self):
...
+            self.skipTest(reason="ModelTester is not configured to run training tests")
+
+        config, inputs_dict = self.model_tester.prepare_config_and_inputs_for_common()
+        if hasattr(config, "use_cache"):
+            config.use_cache = False
+
+        has_verified_model = False
+
+        for model_class in self.all_model_classes:
+            if not getattr(model_class, "supports_gradient_checkpointing", False):
```
zucchini-nlp, line 934: I see now what you meant earlier, this test has a lot of edge cases

zucchini-nlp (commented):
It is a pity that there are so many edge cases to handle. Raising a warning seems like a good solution to not silently skip exceptions

I think there are some unrelated changes with rope, let's revert those before merging

src/transformers/models/internvl/modeling_internvl.py
molbap: lol yes, not mine, and maybe somebody uses it because it's a public method... but better do that than overriding the grad enabling method

src/transformers/models/altclip/modeling_altclip.py
molbap: because now both are not exposing their embeddings in exactly the same way, but will see if I can fix it!

tests/test_modeling_common.py
molbap: yes, it's a bit clunky to have this bool flag but wasn't seeing a simpler option

src/transformers/models/internvl/modeling_internvl.py
zucchini-nlp: love seeing a fix to a general consistency issue :)

src/transformers/models/mlcd/modeling_mlcd.py
molbap: that one is a little bit more annoying. I'm trying to revert it but the thing is this model should work except that it's a tuple... and gradient checkpointing does not disable grads on tuples, only on tensors _passed as pos args_ . 

so this was a hacky trick (for the use_reentrant case

src/transformers/modeling_utils.py
molbap: we don't always, but we do with reentrant checkpointing. IIUC it's not to actualy use these gradients, it's that torch.utils.checkpoint needs at least one input and one output to actually have gradients, else the checkpointed part will not have a gradient.

src/transformers/models/mlcd/modeling_mlcd.py
zucchini-nlp: weird, we use tuple cos/sin in most LLMs. I'd prefer to skip this model's test instead of fixing by duplicate args, i think it is not used as commonly

zucchini-nlp (approved):
Thanks again for handling all edge cases, was not an easy one

molbap:
run-slow: align, altclip, chinese_clip, clap, clvp, falcon_mamba, fast_vlm, internvl, layoutlm, layoutlmv3, lilt, mamba, markuplm, mlcd, poolformer, siglip

src/transformers/models/mlcd/modeling_mlcd.py
molbap: I did skip it yes because it was being annoying :grin: and yes we use them, but they don't usually carry grads, here (in that particular model) they do

src/transformers/modeling_utils.py
```diff
@@ -1994,11 +1996,18 @@ def make_inputs_require_grads(module, input, output):
...
 
         self._require_grads_hooks = hooks
         if hooks:
             # for BC
             self._require_grads_hook = hooks[0]
+        if not found_embeddings:
+            logger.warning_once(
+                f"{self.__class__.__name__} does not expose input embeddings. Gradients cannot flow back to the token "
+                "embeddings when using adapters or gradient checkpointing. Override `get_input_embeddings` to fully "
+                "support those features."
```
ArthurZucker: either that or sometimes just add a `_input_embedding_layer`

tests/test_modeling_common.py
```diff
@@ -906,6 +906,111 @@ def test_peft_gradient_checkpointing_enable_disable(self):
                         m.gradient_checkpointing, f"Module {n} does not have gradient_checkpointing set to False"
                     )
 
+    def test_enable_input_require_grads(self):
+        config, _ = self.model_tester.prepare_config_and_inputs_for_common()
+        for model_class in self.all_model_classes:
+            model = model_class(copy.deepcopy(config))
+            if not hasattr(model, "get_input_embeddings"):
+                continue
```
ArthurZucker, line 915: why not raise an error instead this way all new models wiil make sure they have this go green before merging?

tests/test_modeling_common.py
```diff
@@ -906,6 +906,111 @@ def test_peft_gradient_checkpointing_enable_disable(self):
...
+            try:
+                model.enable_input_require_grads()
+            except NotImplementedError as error:
+                self.fail(f"enable_input_require_grads raised NotImplementedError for {model_class.__name__}: {error}")
+            finally:
+                model.disable_input_require_grads()
+
+    def test_enable_input_require_grads_with_gradient_checkpointing(self):
+        if not getattr(self.model_tester, "is_training", False):
+            self.skipTest(reason="ModelTester is not configured to run training tests")
```
ArthurZucker, line 925: If this one is `True` by default for all models sg

ArthurZucker (approved):
Kudos very nice PR!

src/transformers/modeling_utils.py
ArthurZucker:
```suggestion
                "support those features, or add the `_input_embedding_layer` attribut with the name of the embedding layer!"
```

tests/test_modeling_common.py
```diff
@@ -906,6 +906,111 @@ def test_peft_gradient_checkpointing_enable_disable(self):
...
+            if loss_tensor is None or not isinstance(loss_tensor, torch.Tensor) or not loss_tensor.requires_grad:
+                model.zero_grad(set_to_none=True)
+                continue
+            loss = loss_tensor.sum()
+            loss.backward()
+
+            grad_after_gc = embedding_param.grad
+            self.assertIsNotNone(
+                grad_after_gc,
+                f"{model_class.__name__} should produce embedding gradients when gradient checkpointing is enabled.",
```
ArthurZucker: if you have an idea of what could cause this to fail, add it!

tests/test_modeling_common.py
```diff
@@ -906,6 +906,111 @@ def test_peft_gradient_checkpointing_enable_disable(self):
...
+                grad_after_gc.abs().sum().item(),
+                0,
+                f"{model_class.__name__} should keep non-zero embedding gradients with gradient checkpointing enabled.",
+            )
+            has_verified_model = True
+
+        if not has_verified_model:
+            self.skipTest(
+                reason="No model with a differentiable loss was available to verify enable_input_require_grads with gradient checkpointing."
+            )
```
ArthurZucker, line 1015: same for each one of these, let's help the use to fix it!

ArthurZucker (approved):
First of all thanks! 
This will also fix some TP recompile issues cc @3outeille on `hidden_states=hidden_states`

src/transformers/modeling_utils.py
ArthurZucker: i mean update the message please

ArthurZucker (approved):
Very nice! small updates and let's merge

tests/test_modeling_common.py
molbap: forgot to answer but: this would currently raise for many existing models

tests/test_modeling_common.py
molbap: yes AFAIK, true for CausalLMTester

molbap:
run-slow: align, altclip, chinese_clip, clap, clvp, falcon_mamba, fast_vlm, internvl, layoutlm, layoutlmv3, lilt, mamba, markuplm, mlcd, poolformer, siglip

molbap:
Comments addressed. Merging and keeping an eye on this :eyes: let's see if something breaks and how

PR #42560, opened 2025-12-02
[Quantization] per tensor quantization kernel

MekkCyber:
# What does this PR do?

Adds a simple kernel for per tensor quantization, where the matmul is done per blocks of 128x128, and the weights scales, and activation scales are expected to be scalars


src/transformers/integrations/finegrained_fp8.py
```diff
@@ -182,16 +255,25 @@ def w8a8_block_fp8_matmul_triton(
     assert len(block_size) == 2
     block_n, block_k = block_size[0], block_size[1]
 
+    # if we have per-tensor quantization, we use 128x128 block size for tiled matmul multiplication
+    if block_n == B.shape[-2] and block_k == B.shape[-1]:
+        block_n = 128
+        block_k = 128
+
```
SunMarc, line 265: it doesn't make sense before to set blocks to something else than None when doing per tensor in the FP8Linear. Can we change that so that we fix it here also ?

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -156,6 +156,79 @@ def _w8a8_block_fp8_matmul(
...
+    # Meta-parameters
+    BLOCK_SIZE_M: tl.constexpr,
+    BLOCK_SIZE_N: tl.constexpr,
+    BLOCK_SIZE_K: tl.constexpr,
+    GROUP_SIZE_M: tl.constexpr,
+):
+    """Triton-accelerated function used to perform linear operations (dot
+    product) on input tensors `A` and `B` with block-wise quantization, and
+    store the result in output tensor `C`.
+    """
```
SunMarc, line 190: update

SunMarc (approved):
Thanks a lot for adding this ! Just a minor comment

ArthurZucker (approved):

PR #42563, opened 2025-12-02
Refactor-tokenization-more

ArthurZucker:
# What does this PR do?
 1. Update `TokenizersBackend` to simplify converting from sentencepiece. 
 2. Add the `model` attribute to guide the conversion. 
 3. Adds the `convert_from_spm` for special cases when the spm is not straight forward. 
4. Simplify init for TokenizersBackend to support back passing vocab file / merges file path.


src/transformers/tokenization_utils_base.py
```diff
@@ -2081,118 +2025,12 @@ def _from_pretrained(
...
-
-        # If tokenizer_file exists and tokenizer has a TokenizersBackend, replace the blank tokenizer with tokenizer.json
-        if tokenizer_file is not None and hasattr(tokenizer, "_tokenizer"):
-            from tokenizers import Tokenizer as TokenizerFast
-
-            tokenizer._tokenizer = TokenizerFast.from_file(tokenizer_file)
-            # Re-run post-initialization if the tokenizer has it
-            if hasattr(tokenizer, "_post_init"):
-                tokenizer._post_init()
-            # If only SPM exists, try to get vocab and merges and init to load a tokenizers-backend
```
tomaarsen, line 2093: I believe this code should not have been removed. Opening a PR now.

PR #42567, opened 2025-12-02
[V5] Return a BatchEncoding dict from apply_chat_template by default again

Rocketknight1:
This is basically PR #41626 again! Some of it got clobbered in the tokenizer refactor, but it's just as good the second time.

### DIFF
--- a/src/transformers/tokenization_utils_base.py
+++ b/src/transformers/tokenization_utils_base.py
@@ -3192,7 +3192,7 @@ class PreTrainedTokenizerBase(PushToHubMixin):
         truncation: bool = False,
         max_length: Optional[int] = None,
         return_tensors: Optional[Union[str, TensorType]] = None,
-        return_dict: bool = False,
+        return_dict: bool = True,
         return_assistant_tokens_mask: bool = False,
         tokenizer_kwargs: Optional[dict[str, Any]] = None,
         **kwargs,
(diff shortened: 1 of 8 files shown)


Rocketknight1:
cc @LysandreJik - this was one of the V5 PRs before, do I need to do anything special with this one, or can we just merge it to `main`?

src/transformers/tokenization_utils_base.py
```diff
@@ -3268,14 +3265,11 @@ def apply_chat_template(
             set, will return a dict of tokenizer outputs instead.
         """
 
-        if return_dict and not tokenize:
-            raise ValueError(
-                "`return_dict=True` is incompatible with `tokenize=False`, because there is no dict "
-                "of tokenizer outputs to return."
-            )
+        if not tokenize:
+            return_dict = False  # dicts are only returned by the tokenizer anyway
```
zucchini-nlp, line 3269: makes me wonder, do we need to support a combination of `tokenize=True, return_dict=False` or can we deprecate/remove `return_dict` over time? Can't think of cases when users want a list of tokens as output

zucchini-nlp (approved):
great, it was already approved once so lgtm 😄

src/transformers/tokenization_utils_base.py
Rocketknight1: Maybe we can get rid of it over time, but I think it's fine as a backward compatibility flag for now!

zucchini-nlp: sure, i meant after v5 + several more minor releases, and if users are fine with it

PR #42571, opened 2025-12-02
[`CI`] Fix copies

vasqu:
As per title

### DIFF
--- a/src/transformers/models/fast_vlm/modeling_fast_vlm.py
+++ b/src/transformers/models/fast_vlm/modeling_fast_vlm.py
@@ -59,7 +59,7 @@ class FastVlmMultiModalProjector(nn.Module):
 class FastVlmPreTrainedModel(PreTrainedModel):
     config: FastVlmConfig
     base_model_prefix = "model"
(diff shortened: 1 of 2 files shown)


src/transformers/models/fast_vlm/modeling_fast_vlm.py
```diff
@@ -59,7 +59,7 @@ def forward(self, image_features):
 class FastVlmPreTrainedModel(PreTrainedModel):
     config: FastVlmConfig
     base_model_prefix = "model"
-    input_modalities = ["image", "text"]
+    input_modalities = ("image", "text")
```
molbap, line 62: thanks! seen that in a couple models I feel, hopefully fixed now?

molbap (approved):
Sounds good to me, I think we can merge

src/transformers/models/fast_vlm/modeling_fast_vlm.py
vasqu: Hope so :cry: I trust `fix-copies` to have caught them

PR #42573, opened 2025-12-02
Fix no grad usage

vasqu:
As per title, you need to "()" for the decorator to work as intended. Not sure about the implications but it certainly does not hurt to enable it and avoid any sort of unexpected memory surge

### DIFF
--- a/src/transformers/generation/continuous_batching/continuous_api.py
+++ b/src/transformers/generation/continuous_batching/continuous_api.py
@@ -623,7 +623,7 @@ class ContinuousBatchProcessor:
         self.scheduler.waiting_requests_order.clear()
 
     @traced
-    @torch.no_grad
+    @torch.no_grad()
     def _generation_step(self, model: nn.Module, logit_processor: LogitsProcessor, do_sample: bool) -> None:
         """Perform a single generation step."""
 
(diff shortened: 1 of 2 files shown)


src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -623,7 +623,7 @@ def fail_all_requests(self, error: Exception) -> None:
         self.scheduler.waiting_requests_order.clear()
 
     @traced
-    @torch.no_grad
+    @torch.no_grad()
```
vasqu, line 626: Unsure about this one but I don't see a reason why we would not close here as well cc @remi-or

ArthurZucker (approved):
Both are equivalent no?

vasqu:
> Both are equivalent no?

Did a sanity check
```python
import torch
x = torch.tensor([1.], requires_grad=True)

def identity(x):
    return x * 1
z = identity(x)
z.requires_grad  # True

@torch.no_grad()
def doubler(x):
    return x * 2
z = doubler(x)
z.requires_grad  # False

@torch.no_grad
def tripler(x):
    return x * 3
z = tripler(x)
z.requires_grad  # False
```

You are correct, they are the same. I was confused about the standards around `()` - guess this is really nitpicky then. Will close this.

PR #42576, opened 2025-12-02
Every model forward() should have **kwargs

Rocketknight1:
As the name suggests, this adds a test to ensure that all model classes have `**kwargs` at the end of their `forward()` methods. It also updates all models that were failing the test.


Rocketknight1:
Closing and moving to a cleaner branch

PR #42584, opened 2025-12-03
[XPU] Fix fp8 UT patch func

YangKai0616:
# What does this PR do?

When testing `tests/quantization/finegrained_fp8/test_fp8.py` on XPU, it reported the following error:
```
FAILED tests/quantization/finegrained_fp8/test_fp8.py::FP8QuantizerTest::test_dequantization_no_accelerator - AssertionError: False is not true
FAILED tests/quantization/finegrained_fp8/test_fp8.py::FP8QuantizerTest::test_dequantize_when_no_accelerator - ValueError: Expected a xpu device, but got: cpu
FAILED tests/quantization/finegrained_fp8/test_fp8.py::FP8QuantizerTest::test_quantizer_validation_no_accelerator - AssertionError: RuntimeError not raised
```
The root cause is that the `_patch_no_accelerator` method does not fully take effect on XPU.

### DIFF
--- a/tests/quantization/finegrained_fp8/test_fp8.py
+++ b/tests/quantization/finegrained_fp8/test_fp8.py
@@ -46,6 +46,7 @@ def _patch_no_accelerator():
         stack.enter_context(patch("torch.cuda.is_available", return_value=False))
         if hasattr(torch, "xpu"):
             stack.enter_context(patch("torch.xpu.is_available", return_value=False))
+            stack.enter_context(patch("transformers.quantizers.quantizer_finegrained_fp8.is_torch_xpu_available", return_value=False))
         yield
 
 
--- a/tests/quantization/finegrained_fp8/test_fp8.py
+++ b/tests/quantization/finegrained_fp8/test_fp8.py
@@ -46,7 +46,9 @@ def _patch_no_accelerator():
         stack.enter_context(patch("torch.cuda.is_available", return_value=False))
         if hasattr(torch, "xpu"):
             stack.enter_context(patch("torch.xpu.is_available", return_value=False))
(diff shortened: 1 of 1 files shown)


YangKai0616:
Hi @MekkCyber , please help review, thanks!

tests/quantization/finegrained_fp8/test_fp8.py
```diff
@@ -46,6 +46,9 @@ def _patch_no_accelerator():
         stack.enter_context(patch("torch.cuda.is_available", return_value=False))
         if hasattr(torch, "xpu"):
             stack.enter_context(patch("torch.xpu.is_available", return_value=False))
+            stack.enter_context(
+                patch("transformers.quantizers.quantizer_finegrained_fp8.is_torch_xpu_available", return_value=False)
+            )
```
MekkCyber, line 51: weird that we need to add it, but I guess it's fine

MekkCyber (approved):
Thanks for the fix @YangKai0616

tests/quantization/finegrained_fp8/test_fp8.py
YangKai0616: It's because `is_torch_xpu_available` is used [here](https://github.com/huggingface/transformers/blob/7266f50bc8c3f071fd2fddb5eeb87280a029eabc/src/transformers/quantizers/quantizer_finegrained_fp8.py#L41). During the `Setup` phase, we called this func once via `cls.quantized_model`, and since it is decorated with `@lru_cache`, even if we enter `_patch_no_accelerator`, the value of `is_torch_xpu_available` read here is still True.

PR #42586, opened 2025-12-03
Fix some models cache initialization

albertvillanova:
Fix Gemma3 cache initialization.

Current code:
```python
if use_cache and past_key_values is None and not self.training:
    past_key_values = DynamicCache(config=self.config)
```

The `not self.training` check prevents cache creation during training. However, GRPO (and similar RL trainers) call `model.generate()` during training to produce completions for scoring. This condition blocks cache creation, resulting in:
  - First forward: `past_key_values=None`, no cache → no error (prefill processes all tokens)
  - Second forward: `past_key_values=None`, `cache_position=[279]` → mask expects 280 tokens but KV has only 1 → **shape mismatch error**

This PR removes the `not self.training` condition. 
```python
if use_cache and past_key_values is None:
    past_key_values = DynamicCache(config=self.config)
```
Now, the cache flow is:
  - First forward:  `past_key_values=None` → DynamicCache(length=279)
  - Second forward: `past_key_values=DynamicCache(279)` → DynamicCache(length=280)

Related issue in TRL:
- https://github.com/huggingface/trl/issues/4593

CC: @Cyrilvallez and @vasqu, who were investigating the underlying issue

### DIFF
--- a/src/transformers/models/cohere2/modular_cohere2.py
+++ b/src/transformers/models/cohere2/modular_cohere2.py
@@ -367,7 +367,7 @@ class Cohere2Model(Gemma2Model):
         if inputs_embeds is None:
             inputs_embeds = self.embed_tokens(input_ids)
 
-        if use_cache and past_key_values:
+        if use_cache and past_key_values is None:
             past_key_values = DynamicCache(config=self.config)
 
         if cache_position is None:
--- a/src/transformers/models/cohere2/modular_cohere2.py
+++ b/src/transformers/models/cohere2/modular_cohere2.py
@@ -22,7 +22,6 @@ import torch.nn as nn
 from ...cache_utils import Cache, DynamicCache
 from ...configuration_utils import PreTrainedConfig, layer_type_validation
 from ...masking_utils import create_causal_mask, create_sliding_window_causal_mask
-from ...modeling_flash_attention_utils import FlashAttentionKwargs
 from ...modeling_outputs import BaseModelOutputWithPast
 from ...modeling_rope_utils import (
     RopeParameters,
@@ -271,7 +270,7 @@ class Cohere2Attention(CohereAttention):
         attention_mask: Optional[torch.Tensor],
         past_key_values: Optional[Cache] = None,
         cache_position: Optional[torch.LongTensor] = None,
-        **kwargs: Unpack[FlashAttentionKwargs],
+        **kwargs: Unpack[TransformersKwargs],
     ) -> tuple[torch.Tensor, Optional[torch.Tensor], Optional[tuple[torch.Tensor]]]:
         input_shape = hidden_states.shape[:-1]
         hidden_shape = (*input_shape, -1, self.head_dim)
@@ -322,7 +321,7 @@ class Cohere2DecoderLayer(CohereDecoderLayer):
         past_key_values: Optional[Cache] = None,
         use_cache: Optional[bool] = False,
         cache_position: Optional[torch.LongTensor] = None,
-        **kwargs: Unpack[FlashAttentionKwargs],
+        **kwargs: Unpack[TransformersKwargs],
     ) -> tuple[torch.FloatTensor, Optional[tuple[torch.FloatTensor, torch.FloatTensor]]]:
         residual = hidden_states
         hidden_states = self.input_layernorm(hidden_states)
--- a/src/transformers/models/cohere2/modular_cohere2.py
+++ b/src/transformers/models/cohere2/modular_cohere2.py
@@ -367,7 +367,7 @@ class Cohere2Model(Gemma2Model):
         if inputs_embeds is None:
             inputs_embeds = self.embed_tokens(input_ids)
 
-        if use_cache and past_key_values is None and not self.training:
+        if use_cache and past_key_values:
             past_key_values = DynamicCache(config=self.config)
 
         if cache_position is None:
--- a/src/transformers/models/gemma3/modular_gemma3.py
+++ b/src/transformers/models/gemma3/modular_gemma3.py
@@ -527,8 +527,6 @@ class Gemma3DecoderLayer(GradientCheckpointingLayer):
         attention_mask: Optional[torch.Tensor] = None,
         position_ids: Optional[torch.LongTensor] = None,
         past_key_values: Optional[Cache] = None,
-        output_attentions: Optional[bool] = False,
-        use_cache: Optional[bool] = False,
         cache_position: Optional[torch.LongTensor] = None,
         **kwargs,
     ) -> tuple[torch.FloatTensor, Optional[tuple[torch.FloatTensor, torch.FloatTensor]]]:
@@ -536,14 +534,12 @@ class Gemma3DecoderLayer(GradientCheckpointingLayer):
 
         hidden_states = self.input_layernorm(hidden_states)
 
-        hidden_states, self_attn_weights = self.self_attn(
+        hidden_states, _ = self.self_attn(
             hidden_states=hidden_states,
             position_embeddings=position_embeddings,
             attention_mask=attention_mask,
             position_ids=position_ids,
             past_key_values=past_key_values,
-            output_attentions=output_attentions,
-            use_cache=use_cache,
             cache_position=cache_position,
             **kwargs,
         )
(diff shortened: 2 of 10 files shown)


Cyrilvallez:
I will check more, but I believe the true issue comes from the line I showed in my comment https://github.com/huggingface/trl/issues/4593#issuecomment-3606035594! Here, you're only treating a symptom I think, not the full problem. Also, you said in the issue that you were not in training mode 👀

Usually, even in the middle of training, you will go back to inference mode to perform evaluation in order to deactivate dropout etc, unless I'm very mistaken

Cyrilvallez:
Actually I was wrong, and this is the correct fix @albertvillanova! 

@vasqu because the input preparation relies on `use_cache` and not on `cache is None` to know whether we'll have a cache or not, there was a silent bug in which the input would be sliced as if a cache was used, but due to the `is training` conditional, no cache was provided. So generation would be successful, but completely wrong, as it would just see the last token in each generation step. The change in `mask_utils` actually uncovered it, because it shown the mismatch between what the size should be and what the size actually is, because we should have a cache but don't.... Before we would just prepare a mask of size 1, which is the input size, but the input size should NOT be that.
It started being an issue since https://github.com/huggingface/transformers/pull/41505 2 months ago... Before that, the Cache would have been there anyway and no problem

@albertvillanova I'm going to push much more changes on this PR, as we should really be more careful about this!
In the future, I'm going to make sure that `generate` relies on whether we actually have a cache or not, and not on `use_cache` which can be decorrelated from the cache existence as it's the case here

vasqu:
> Usually, even in the middle of training, you will go back to inference mode to perform evaluation in order to deactivate dropout etc, unless I'm very mistaken

I think we are lucky here that most newer models do not use dropout so there is no real difference in inference. I agree tho, in general, the model should be set to eval `model.eval()` just to be sure.

> Here, you're only treating a symptom I think, not the full problem. Also, you said in the issue that you were not in training mode 👀

I think this really is the core issue --> We should not cut our inputs at all when we do not use a cache (as indicated by the config flag). However, the default flag is set to to `True`. This forces the cutting of the input, except you explicitly pass `use_cache=False` yourself. That shouldn't be the case at all! 

Some quick pointers what happens:
- https://github.com/huggingface/transformers/blob/75c135d76f6db226d83213c6bb658a474dddb07d/src/transformers/generation/utils.py#L616 We get `use_cache=True` here
- https://github.com/huggingface/transformers/blob/75c135d76f6db226d83213c6bb658a474dddb07d/src/transformers/generation/utils.py#L619-L625 We get into the cache-dependent input prep
- https://github.com/huggingface/transformers/blob/75c135d76f6db226d83213c6bb658a474dddb07d/src/transformers/generation/utils.py#L523 Our input gets cut (even if it shouldn't

The real solution IMO is as @Cyrilvallez said: Remove the default for `use_cache`, e.g. by removing it completely there and let it be handled within the kwargs.

vasqu:
Ok, but then let's do both @Cyrilvallez: This fix along the kwarg being defaulted to `True`. It's not good either way

Cyrilvallez:
Haha we commented at the same time

Yeah agree the kwargs should not be True, but it's not the one at fault here

src/transformers/models/cohere2/modular_cohere2.py
```diff
@@ -367,7 +367,7 @@ def forward(
         if inputs_embeds is None:
             inputs_embeds = self.embed_tokens(input_ids)
 
-        if use_cache and past_key_values is None and not self.training:
+        if use_cache and past_key_values:
```
vasqu:
```suggestion
        if use_cache and past_key_values is None:
```
Careful here

src/transformers/models/fast_vlm/modeling_fast_vlm.py
vasqu, line 1: Should be handled on main iirc

vasqu (commented):
Small intermediate review

src/transformers/models/fast_vlm/modeling_fast_vlm.py
Cyrilvallez: Yep indeed just noticed it and rebased

src/transformers/models/gemma3/modular_gemma3.py
```diff
@@ -934,6 +881,7 @@ class Gemma3ForConditionalGeneration(PaliGemmaForConditionalGeneration):
     # Fix: https://github.com/huggingface/transformers/issues/40564
     accepts_loss_kwargs = False
 
+    @can_return_tuple
```
albertvillanova, line 883: I think, after your changes, multiple `forward` methods cannot return a tuple anymore. Their return signature should be updated as well.

```suggestion
```

src/transformers/models/gemma3n/modeling_gemma3n.py
```diff
@@ -1426,154 +1420,7 @@ def forward(
...
-
-    def forward(self, x):
-        down_proj = self.down_proj(self.act_fn(self.gate_proj(x)) * self.up_proj(x))
-        return down_proj
-
-
-class Gemma3nAttention(nn.Module):
-    """Multi-headed attention from 'Attention Is All You Need' paper"""
-
-    def __init__(self, config: Gemma3nConfig, layer_idx: int):
```
Cyrilvallez, line 1456: Those classes should never have been here... Only the `Text` ones are needed. Because `_can_record_outputs` was not correctly overwritten with the correct `Text` prefix, they were imported...

src/transformers/models/gemma3/modular_gemma3.py
vasqu: Hmm, not sure if I can follow:
- `can_return_tuple` allows users to return a tuple if they pass `return_dict=False`
- Defaults to returning the output class tho

After multiple forwards it still depends on how you call it.

Cyrilvallez: The decorator automatically handle tuple casting if needed!

src/transformers/models/gemma3/modular_gemma3.py
```diff
@@ -527,23 +527,19 @@ def forward(
         attention_mask: Optional[torch.Tensor] = None,
         position_ids: Optional[torch.LongTensor] = None,
         past_key_values: Optional[Cache] = None,
-        output_attentions: Optional[bool] = False,
-        use_cache: Optional[bool] = False,
         cache_position: Optional[torch.LongTensor] = None,
         **kwargs,
```
vasqu: Should we add the `TransformersKwargs` here for typing, just for completeness sake (and elsewhere)

Cyrilvallez:
Pushed the necessary change related to this PR on several models, and took the opportunity to do quite some cleanup at the same time!

src/transformers/models/gemma3n/modeling_gemma3n.py
vasqu: Wow, so the manual collecting was needed because we didn't properly type out out records either way

src/transformers/models/gemma3n/modular_gemma3n.py
```diff
@@ -1976,7 +1972,8 @@ def project_per_layer_inputs(
             dtype=inputs_embeds.dtype, device=per_layer_projection.device
         )
 
-    @can_return_tuple
+    # Last hidden states should be before reprojecting it to stay consistent, so we do not overwrite with the final one
```
vasqu:
```suggestion
    # Last hidden states should be before reprojecting it to stay consistent with layer output
```
Just nit, but felt like shortening it

src/transformers/models/gemma3n/modular_gemma3n.py
```diff
@@ -1987,46 +1984,28 @@ def forward(
...
             past_key_values = DynamicCache(config=self.config)
 
         if cache_position is None:
             past_seen_tokens = past_key_values.get_seq_length() if past_key_values is not None else 0
-            cache_position = torch.arange(
-                past_seen_tokens,
-                past_seen_tokens + inputs_embeds.shape[1],
-                device=inputs_embeds.device,
-            )
+            cache_position = torch.arange(inputs_embeds.shape[1], device=inputs_embeds.device) + past_seen_tokens
```
vasqu, line 2007: Hmm, that was to avoid a device sync iirc? Just fmi

vasqu (approved):
LGTM, just some smaller comments + let's just make sure to check with slow tests that nothing has been significantly broken

src/transformers/models/gemma3n/modular_gemma3n.py
Cyrilvallez: Yeah apparently `arange` sync by calling `.item` if a tensor on-device is passed to it.... 🫠🫠🙃🙃

PR #42587, opened 2025-12-03
fix : cast into floats AFTER all assignments

marconaguib:
# What does this PR do?
If user sets `whole_word_mask=True` and doesn't assign correct values to `mask_replace_prob` and `random_replace_prob`, their values are set automatically to 1 and 0 but are never cast to floats. This may be problematic if, for instance, when running whithout GPUs, because user would get `NotImplementedError: "bernoulli_tensor_cpu_p_" not implemented for 'Long'`.
This PR does the casting *after* handling incorrect probabilities.

## Before submitting
- [X] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
      Pull Request section?
      to it if that's the case.

## Who can review?


Rocketknight1 (approved):
Yes, this is a good catch, thank you!

PR #42588, opened 2025-12-03
Document the /v1/models endpoint

LysandreJik:


docs/source/en/serving.md
```diff
@@ -413,6 +413,17 @@ For example, to enable 4-bit quantization with bitsandbytes, you need to pass ad
 transformers serve --quantization bnb-4bit
 ```
 
+### Available models
+
+The `/v1/models` endpoint scans your local cache for generative models. It checks for LLMs as well as VLMs.
+The easiest way to download models to your cache is with the `hf download` command:
+
+```shell
+hf download <model_id>
```
LysandreJik, line 422: Actually this will download all the files (even "original" files), so there is probably a better way

LysandreJik: Maybe it would make sense to expose a utility that would only download the transformers-specific files

Wauplin: do you mean something like `transformers download` ? If only it existed...  😛 

Image

Wauplin:
```suggestion
transformers download <model_id>
```

docs/source/en/serving.md
```diff
@@ -413,6 +413,17 @@ For example, to enable 4-bit quantization with bitsandbytes, you need to pass ad
 transformers serve --quantization bnb-4bit
 ```
 
+### Available models
+
+The `/v1/models` endpoint scans your local cache for generative models. It checks for LLMs as well as VLMs.
+The easiest way to download models to your cache is with the `hf download` command:
```
Wauplin, line 419:
```suggestion
The easiest way to download models to your cache is with the `transformers download` command:
```

docs/source/en/serving.md
```diff
@@ -413,6 +413,17 @@ For example, to enable 4-bit quantization with bitsandbytes, you need to pass ad
...
+
+```shell
+hf download <model_id>
+```
+
+As long as it's a transformers-compatible LLM/VLM, it will now show up under `/v1/models`
```
Wauplin, line 425:
```suggestion
Once downloaded, the model will show up under `/v1/models`.
```

stevhliu (commented):
Nice! This adds more context than my current docs for `v1/models` [here](https://github.com/huggingface/transformers/pull/42263/files#diff-0d245075e8c2672d792f96d8fd2b695b603c204231f70fca1f3154205b39d6dfR352).

Maybe we can combine the two to also show how to curl the `v1/models` endpoint for a more complete end-to-end example?

PR #42592, opened 2025-12-03
Fix convert_tekken_tokenizer

juliendenize:
# What does this PR do?

Right now the `convert_tekken_tokenizer` does not add `bos_tokens`, `eos_token`  to the special tokens via the `add_special_tokens` method.

This prevents the chat templates that expect `eos_token` and `bos_token` to work properly.

Previously this was working as when saving the tokenizer a `special_tokens_map.json` was created which is no longer the case. Unknown to me why but I'd assume this is due to the V5 refactoring ?

This PR fixes that by adding explicitly these tokens to the tokenizer and when saving they're now stored in `tokenizer_config.json`.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker


src/transformers/integrations/mistral.py
```diff
@@ -106,4 +107,15 @@ def convert_tekken_tokenizer(tokenizer_file: str):
...
+    MAP_SPECAL = {
+        "bos_token": SpecialTokens.bos.value,
+        "eos_token": SpecialTokens.eos.value,
+        "pad_token": SpecialTokens.pad.value,
+        "unk_token": SpecialTokens.unk.value
+    }
+
+    for special_key, special_token in MAP_SPECAL.items():
+        if special_token in all_special:
+            tokenizer.add_special_tokens({special_key: special_token})
```
juliendenize, line 119: this is for backward and forward compatibility

Rocketknight1:
cc @itazap

itazap:
run-slow: ministral3, mistral3

itazap:
Hey! Thanks for the PR, can you please share a short reproducer of the problem (you mentioned in chat templates)? perhaps we'll need to add a test !

ArthurZucker (approved):
LGTM, we are done with the other refactoring / fix on the other side so should be good!

juliendenize:
@itazap just a classic apply_chat_template with the wrong config.

here is a chat template that would fail
https://huggingface.co/mistralai/Ministral-3-3B-Instruct-2512-BF16?chat_template=default

as you can see BOS and EOS tokens are variable.

In Transformers when you apply_chat_template here:
https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_utils_base.py#L3140-L3149

you pass special tokens that are defined by the tokenizer's config.

Because special maps are no longer generated as before, it wasn't put in the tokenizer config so never passed to the rendered so it was outputting empty strings ( not even an error :( )

It's not really a bug but a wrong configuration passed so you could repro if you'd like by removing in ministral's config these lines

https://huggingface.co/mistralai/Ministral-3-3B-Instruct-2512/blob/main/tokenizer_config.json#L1007-L1010
and apply_chat_template wouldn't encode properly eos and bos for Transformers' tokenizer backend
```
import torch

model_id = "mistralai/Ministral-3-3B-Instruct-2512"

tokenizer = MistralCommonBackend.from_pretrained(model_id)
model = Mistral3ForConditionalGeneration.from_pretrained(model_id, device_map="auto")

messages = [
    {
        "role": "user",
        "content": [
            {
                "type": "text",
                "text": "How are you",
            },
        ],
    },
]

tokenized = tokenizer.apply_chat_template(messages, return_tensors="pt", return_dict=True)

print(tokenized["input_ids"]) # outputs <s> as first token if right config else outputs [SYSTEM_PROMPT]
```

ArthurZucker (approved):
Thanks sorry for the late merge

PR #42593, opened 2025-12-03
Use `getattr` in `standardize_rope_params` because `rope_parameters` not always present

hmellor:
Some custom models written for v4 will specify `rope_theta` but not `rope_scaling`/`rope_parameters`. In this case, there was an attribute error when accessing `rope_parameters` in `standardize_rope_params`. This PR fixes that.


zucchini-nlp (approved):
This fn was simplified a lot in the prev PR since we can make a few assumptions about inputs. Looks like it will still work with custom models, as long as they have single rope per model. Thanks for fixing

PR #42595, opened 2025-12-03
Fix Ernie Moe Test

vasqu:
Flaky test: 
- `pytest tests/models/ernie4_5_moe/test_modeling_ernie4_5_moe.py -k test_load_balancing_loss --flake-finder --flake-runs=1000`
- Encounter 0 failures

Hope it's more robust again

### DIFF
--- a/tests/models/ernie4_5_moe/test_modeling_ernie4_5_moe.py
+++ b/tests/models/ernie4_5_moe/test_modeling_ernie4_5_moe.py
@@ -97,7 +97,6 @@ class Ernie4_5_MoeModelTest(CausalLMModelTest, unittest.TestCase):
         config, input_dict = self.model_tester.prepare_config_and_inputs_for_common()
         config.num_labels = 3
         config.num_experts = 3
-        config.expert_interval = 1
         config.output_router_logits = True
         input_ids = input_dict["input_ids"]
         attention_mask = input_ids.ne(config.pad_token_id).to(torch_device)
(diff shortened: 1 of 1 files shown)


tests/models/ernie4_5_moe/test_modeling_ernie4_5_moe.py
```diff
@@ -97,24 +97,28 @@ def test_load_balancing_loss(self):
...
-        attention_mask = input_ids.ne(1).to(torch_device)
+        attention_mask = input_ids.ne(config.pad_token_id).to(torch_device)
         model = Ernie4_5_MoeForCausalLM(config)
         model.to(torch_device)
         model.eval()
         result = model(input_ids, attention_mask=attention_mask)
-        self.assertEqual(result.router_logits[0].shape, (91, config.num_experts))
+        bs, seqlen = input_ids.shape
+        self.assertEqual(result.router_logits[0].shape, (bs * seqlen, config.num_experts))
+        # After #40617, we still have 0.01 % of failure rate here.
```
ydshieh: let's not add `# After #40617` here and below.

it's there for jamba, but no need to propagate to here.

ydshieh (approved):
Thank you.

> @is_flaky(max_attempts=2)

(already here before this PR)

not working well?

vasqu:
I think the core changes are
- Properly propagate padding tokens within the ids https://github.com/huggingface/transformers/blob/eb3c0880a4d2b713b49da7763b129e199608ba56/tests/models/ernie4_5_moe/test_modeling_ernie4_5_moe.py#L115-L120
- And using the config pad token id, not some arbitrary number

I think either or both contributed to slightly different behavior compared to the other moe models

PR #42603, opened 2025-12-03
Every model forward() should have **kwargs

Rocketknight1:
This PR ensures every model inheriting directly or indirectly from PreTrainedModel accepts `**kwargs` in its `forward()` method. We were quite inconsistent about this before, and not accepting them causes issues in some cases. It also adds a test to `check_repo.py` to ensure we stay consistent in future!


ArthurZucker (approved):
perfect 🫡

PR #42614, opened 2025-12-04
Fix incorrect tokenizer __repr__ formatting

### LINKED ISSUE #34437: Incorrect repr string for tokenizer objects
### System Info

transformers 4.46.0
Any OS and python version

### Who can help?

@ArthurZucker @itazap

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

```
tokenizer = AutoTokenizer.from_pretrained('Qwen/Qwen2.5-0.5B-Instruct') 
print(tokenizer)
```

### Expected behavior

The repr of tokenizer objects is incorrectly formatted due to this part of the code:
https://github.com/huggingface/transformers/blob/1d063793318b20654ebb850f48f43e0a247ab7bb/src/transformers/tokenization_utils_base.py#L1684C1-L1692C10

The repr of a Tokenizer object looks like this: `Tokenizer(...), added_tokens_decoder={...}`
Whereas is should look like this: `Tokenizer(..., added_tokens_decoder={...})`
The dict that is the value of the `added_tokens_decoder` attribute should be listed within the parentheses along with the other attributes, not after the closing parenthesis.

The current representation is problematic because having the `added_tokens_decoder` outside the main parenthesized structure breaks the expected flow of representing object attributes, and it's confusing. It suggests that the relationship between the tokenizer parameters and the added tokens decoder is different from what it actually is. Someone reading the string representation could assume it's a separate entity instead of an attribute belonging to the tokenizer.

Lines 1690-1691 should be corrected like this:
```
            f" special_tokens={self.special_tokens_map}, clean_up_tokenization_spaces={self.clean_up_tokenization_spaces}, "
            " added_tokens_decoder={\n\t" + added_tokens_decoder_rep + "\n})"
```

bala-539958:
This PR fixes a formatting issue in the tokenizer __repr__ output where
added_tokens_decoder was displayed outside the main parentheses.
It is now correctly displayed as part of the Tokenizer(...) structure.

Fixes #34437

### DIFF
--- a/src/transformers/tokenization_utils_base.py
+++ b/src/transformers/tokenization_utils_base.py
@@ -1451,7 +1451,7 @@ class PreTrainedTokenizerBase(PushToHubMixin):
             f" vocab_size={self.vocab_size}, model_max_length={self.model_max_length},"
             f" padding_side='{self.padding_side}', truncation_side='{self.truncation_side}',"
             f" special_tokens={self.special_tokens_map},"
-            " added_tokens_decoder={\n\t" + added_tokens_decoder_rep + "\n}\n)"
+            " added_tokens_decoder={\n\t" + added_tokens_decoder_rep + "\n}\n)" 
         )
 
     def __len__(self) -> int:


Rocketknight1:
That issue was closed over a year ago.

PR #42615, opened 2025-12-04
Fix _is_package_available to handle underscore/hyphen equivalence

mertunsall:
## Problem

`is_flash_attn_2_available()` returns `False` when both `sgl-kernel` and `flash-attn` are installed, even though `flash-attn` 2.8.1 works fine.

**Reproduction:**
```python
# Simulates the bug when distributions = ['sgl-kernel', 'flash-attn']
pkg_name = "flash_attn"
distributions = ["sgl-kernel", "flash-attn"]

# Current logic (line 60)
result = pkg_name if pkg_name in distributions else distributions[0]
print(result)  # 'sgl-kernel' ❌ (version 0.3.16, fails >= 2.1.0 check)

# 'flash_attn' not in ['sgl-kernel', 'flash-attn'] because underscore ≠ hyphen
```

## Root Cause

`importlib.metadata.packages_distributions()` returns multiple distributions for `flash_attn`. The check `pkg_name in distributions` fails because `'flash_attn'` (underscore) doesn't match `'flash-attn'` (hyphen), so it falls back to `distributions[0]`.

## Fix

Per PEP 503, underscores and hyphens are equivalent in package names. Normalize before matching:

```python
normalized_pkg_name = pkg_name.replace("_", "-")
if normalized_pkg_name in distributions:
    distribution_name = normalized_pkg_name
elif pkg_name in distributions:
    distribution_name = pkg_name
else:
    distribution_name = distributions[0]
```

Now `flash_attn` → `flash-attn` → correct version `2.8.1` ✓

### DIFF
--- a/src/transformers/utils/import_utils.py
+++ b/src/transformers/utils/import_utils.py
@@ -55,9 +55,15 @@ def _is_package_available(pkg_name: str, return_version: bool = False) -> tuple[
             # importlib.metadata works with the distribution package, which may be different from the import
             # name (e.g. `PIL` is the import name, but `pillow` is the distribution name)
             distributions = PACKAGE_DISTRIBUTION_MAPPING[pkg_name]
-            # In most cases, the packages are well-behaved and both have the same name. If it's not the case, we
-            # pick the first item of the list as best guess (it's almost always a list of length 1 anyway)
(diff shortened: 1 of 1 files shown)


Rocketknight1 (approved):
Yes, this is a good fix! The original version that just picked the first option was obviously shaky when the assumption wasn't met.

PR #42618, opened 2025-12-04
Update revision so that there is a safetensors model

LysandreJik:
The linked revision does not contain a safetensors model


ydshieh (approved):
merci!!

ArthurZucker (approved):

PR #42619, opened 2025-12-04
[loading] Allow loading to happen without threading

Cyrilvallez:
# What does this PR do?

As per the title. This should help a lot in our CI, and for ZeroGPU as well!

### DIFF
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -884,9 +884,9 @@ def convert_and_load_state_dict_in_model(
         else:
             unexpected_keys.add(renamed_key)
 
-    total_entries = len(param_name_to_load)
-    with logging.tqdm(total=total_entries, desc="Loading weights") as pbar:
-        try:
+    try:
+        total_entries = len(param_name_to_load)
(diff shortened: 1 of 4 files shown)


src/transformers/core_model_loading.py
```diff
@@ -906,14 +921,17 @@ def convert_and_load_state_dict_in_model(
...
                 mapping.reset()
+
             except SkipLayer:
                 continue
 
     # Keep the current weight conversion mapping for later saving (in case it was coming directly from the user)
     model._weight_conversions = weight_mapping
-    thread_pool.shutdown(wait=False)
+    if thread_pool is not None:
+        thread_pool.shutdown(wait=False)
```
ArthurZucker: should be a finally as well if someone does a ctrl  + C while loading

ArthurZucker (approved):
Ty, needs probably a test?

src/transformers/core_model_loading.py
Cyrilvallez: Humm, indeed not a bad idea at all! Just added it

PR #42621, opened 2025-12-04
Remove redundant else in activation.py

leaderofARS:
# Remove Redundant Else Block in get_activation()

**Fixes #42621**

This PR removes a redundant `else` block in `get_activation()` in `src/transformers/activations.py`. The `else` was unnecessary because the `if` branch already returned, and keeping it went against the project's preferred early-return style.

## Previous Control Flow

```python
def get_activation(activation_string):
    if activation_string in ACT2FN:
        return ACT2FN[activation_string]
    else:
        raise KeyError(f"function {activation_string} not found in ACT2FN mapping {list(ACT2FN.keys())}")
```

## Updated Control Flow

```python
def get_activation(activation_string):
    if activation_string in ACT2FN:
        return ACT2FN[activation_string]
    raise KeyError(f"function {activation_string} not found in ACT2FN mapping {list(ACT2FN.keys())}")
```

## Summary

- Pure cleanup (no behavior change)
- Simplifies logic and matches HF style
- All style & repo checks pass


leaderofARS:
Hi team!  
All CI checks are passing and the branch is rebased to latest main.  
This is a small cleanup with no functional changes.  
Let me know if anything else is needed!

PR #42597, opened 2025-12-03
test ci training for text model only

3outeille:
- Only 1 failing test (flagging https://github.com/huggingface/transformers/issues/42629) . The test will be skipped for now

image
image
image

PR #42627, opened 2025-12-04
Initialize cache state in RecurrentGemmaRecurrentBlock based on batch…

3outeille:
Fix needed to pass CI training test (generation matching test) in https://github.com/huggingface/transformers/pull/42597


3outeille:
run-slow: recurrent_gemma

PR #42632, opened 2025-12-04
[loading] Correctly load params during offloading & careful memory considerations

Cyrilvallez:
# What does this PR do?

During offloading, we are under heavy cpu memory constraints: we have disk offloading because the cpu/devices cannot hold all parameters. Currently, all parameters would be loaded on cpu async, then converted, then resaved if needed, which results in killing the cpu ram and crashing.

Loading asynchronously means we assume that we can hold all parameters in memory, which is true in general if we want to be able to load the model anyway. But with offloading, this assumption does not hold anymore, as we *know we don't have enough memory* so we offload to disk. For this reason, threading is deactivated when offloading is needed, as we cannot control the memory otherwise.

Moreover, this PR changes the logic when we do not use threading from gathering all tensors on cpu before, then converting them all, to loading them on cpu only when entering the Converter. This is essential, as during offloading, as explained above, if we load everything first, we kill the cpu RAM. 
In terms of logic, everything is sequential in both cases, so it has no impact. We just move from:
- loading everything sequentially -> converting everything sequentially with everything in RAM
to
- for each Converter: load all related params sequentially -> convert -> move on to next Converter

This allows to lift all memory issues when offloading, as only the required tensors are loaded, then we convert, resave the converted param to disk, and move on to the next one.
It's basically only a smart reordering but has not other impact on speed etc

Also, this PR makes sure that we never use more memory than necesary during conversion ops, especially if several are chained, by not keeping any reference to the internal `self.collected_tensors` attribute, and deleting refs to collected tensors in the loading loop.


src/transformers/core_model_loading.py
```diff
@@ -376,6 +372,21 @@ def reverse_transform(self) -> WeightTransform:
         return reverse_transform
 
 
+def collect_tensors(tensors: list[torch.Tensor | Future | Callable]) -> list[torch.Tensor]:
```
ArthurZucker: can be just property of weight converter

ArthurZucker (approved):
Nice! did we miss a fast test?

src/transformers/core_model_loading.py
Cyrilvallez: True, will make it so!

src/transformers/core_model_loading.py
```diff
@@ -389,19 +411,17 @@ def convert(
...
         misc: Optional[MutableMapping[str, str]] = None,
     ):
-        # Collect the tensor if using threading
-        for pattern, futures in self.collected_tensors.items():
-            self.collected_tensors[pattern] = (
-                futures if isinstance(futures[0], torch.Tensor) else [future.result() for future in futures]
-            )
+        # Collect the tensors here - we use a new dictionary to avoid keeping them in memory in the internal
+        # attribute during the whole process
+        collected_tensors = self.materialize_tensors()
```
ArthurZucker, line 416: nice

ArthurZucker (commented):
nice!!!

PR #42636, opened 2025-12-04
[docs] WeightConverter

stevhliu:
Adds docs for `WeightConverter` from #41580

### DIFF
--- a/docs/source/en/weightconverter.md
+++ b/docs/source/en/weightconverter.md
@@ -16,9 +16,9 @@ rendered properly in your Markdown viewer.
 
 # Dynamic weight loading
 
-Transformers provides a set of composable operations in [`WeightConverter`] for defining how to map checkpoint tensors to model tensors. These operations allow you to reshape checkpoint tensors into any expected model format during loading, like fusing the query (Q), key (K), and value (V) layers for example.
+Checkpoints are often serialized in a format that does not match what a model expects at runtime. Quantization and parallelism frequently require reshaping, splitting, or merging tensors into the expected model format instead of loading weights as-is.
 
-This allows Transformers to adapt to new model or weight formats instead of manually adding code for them. It is especially useful for loading mixture-of-experts (MoE) models with different expert representations, handling quantized checkpoints with special tensor layouts, supporting tensor parallelism during loading, or a combination of all of these.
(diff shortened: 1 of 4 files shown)


LysandreJik:
Sound good! You might want to intro the weight converter with some docs about why it's needed and how dynamic weight loading works

I detailed it in the release notes: https://github.com/huggingface/transformers/releases/tag/v5.0.0rc0

docs/source/en/weightconverter.md
```diff
@@ -0,0 +1,103 @@
...
+        PermuteForRope(),
+    ],
+)
+```
+
+## Fast model loading
+
+Loading a model is faster and uses less memory because the loader knows which tensors are required for operations and schedules their materialization lazily.
+
+The loader scans the checkpoint *once* to discover pattern matches and collect tensors. Tensors are collected as `Future` objects and not loaded into memory immediately. They're kept as lazy references until needed to defer memory allocation. Tensor loading is scheduled asynchronously without blocking the GIL.
```
Cyrilvallez: This is not entirely true: as soon as we add the tensor to the `thread_pool` to load it asynchornously, the `thread_pool` is free to load it as soon as it has an available thread to do it. 
So when we enter conversion for a weight, we make sure all needed tensors are materialized by waiting for them if they are not already loaded, but they could very well all be loaded already.
This is why the theoretical memory footprint is `model_size + all_largest_params_needed_in_a_single_conversion` as I mentioned [here](https://github.com/huggingface/transformers/pull/42749#discussion_r2610832732).

This is however somewhat of an "unlucky" outcome: it either means the more memory demanding conversion is the last to happen, or that all other params were already loaded by the `thread_pool` when entering the more demanding conversion.
So in practice, the memory peak can be even lower than that, closer to `model_size`!

Hope this is clear enough, IMO those considerations are interesting to add in the doc as well, as it makes people understand more how it really works!

EDIT: was talking about the "They're kept as lazy references until needed to defer memory allocation." and "
Tensors are materialized once all `Future` objects are collected. Operations are batched together and applied to return the transformed tensors." bits here!

stevhliu: thanks so much for the clarification!

stevhliu:
> You might want to intro the weight converter with some docs about why it's needed and how dynamic weight loading works

reworked the intro to start with motivation and then brief explanation of how it works!

Cyrilvallez (approved):
Hey @stevhliu! Just pushed some changes to (hopefully) clarify a bit and add more details, let me know what you think! Feel free to change a bit the language etc if you think something is not so nice (I'm not a native english speaker 🙃), or maybe use `tip` instead of `note` etc... Feel free to merge then!

stevhliu:
Nice, thanks for adding more concrete examples! I think that'll make it easier to understand.

Also big thanks for iterating with me on this, super cool learning about our new loading system ❤️

ArthurZucker (approved):

PR #42642, opened 2025-12-05
Command-a-vision fix

dongluw:
# What does this PR do?
- fix a bug in the image resize code, which affects performance
before fix:
patch_01

after fix:
patch_1

- add test cases

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


Rocketknight1:
cc @molbap @yonigozlan @zucchini-nlp

molbap:
Hi @dongluw , thanks for the contribution, indeed there seems to be a flip issue. however how do you obtain the test image above? I'm surprised our tests haven't caught this, it should be all wrong, so a reproducer would help

dongluw:
hey @molbap I saved the `stacked_images` to images patch by patch https://github.com/dongluw/transformers/blob/9ef13ef775fe5a05c634fb2705a500ef59f28763/src/transformers/models/cohere2_vision/image_processing_cohere2_vision_fast.py#L228

this issue only affects generation quality if images are of very high/low aspect ratio

full image is 
output_image17

src/transformers/models/cohere2_vision/image_processing_cohere2_vision_fast.py
```diff
@@ -103,7 +103,8 @@ def get_optimal_tiled_canvas(
         # Pick the resolution that required the least upscaling so that it most closely fits the image
         required_scale = np.where(required_scale < 1.0, 10e9, required_scale)
         best_grid = possible_resolutions[np.argmin(required_scale)]
-    return best_grid
+    best_grid_row, best_grid_col = best_grid
+    return best_grid_col, best_grid_row  # revere the order to align with boilerplate code
```
dongluw:
```suggestion
    return best_grid_col, best_grid_row  # revert the order to align with boilerplate code
```

src/transformers/models/cohere2_vision/modular_cohere2_vision.py
```diff
@@ -305,7 +305,8 @@ def get_optimal_tiled_canvas(
         # Pick the resolution that required the least upscaling so that it most closely fits the image
         required_scale = np.where(required_scale < 1.0, 10e9, required_scale)
         best_grid = possible_resolutions[np.argmin(required_scale)]
-    return best_grid
+    best_grid_row, best_grid_col = best_grid
+    return best_grid_col, best_grid_row  # revere the order to align with boilerplate code
```
dongluw:
```suggestion
    return best_grid_col, best_grid_row  # revert the order to align with boilerplate code
```

src/transformers/models/cohere2_vision/image_processing_cohere2_vision_fast.py
zucchini-nlp: the issue is actually here because the grids come in `(w, h)` format but the original size is in `(h, w)` format. We need to swap the original size format

tests/models/cohere2_vision/test_image_processing_cohere2_vision.py
```diff
@@ -190,3 +190,128 @@ def test_call_numpy_4_channels(self):
...
             self.assertEqual(tuple(encoded_images.shape), (70, 4, 30, 30))
+
+    def test_crop_to_patches_aspect_ratio(self):
+        """Test that row/column ordering is correct when cropping non-square images to patches.
+
+        This test verifies that patches can be stitched back to reconstruct the original image,
+        which validates that the row/column ordering in get_optimal_tiled_canvas is correct.
+        If row/column are swapped, the image would be resized to wrong dimensions and patches
+        would not match the original content.
+        """
```
zucchini-nlp, line 201: thanks for adding a test!

zucchini-nlp (commented):
hey, nice catch! I was adapting the processing from GotOCR and misplaced the sizes for (h, w), commented it below. I think we need to fix where aspect ratios are computed

src/transformers/models/cohere2_vision/image_processing_cohere2_vision_fast.py
dongluw: the (h, w) order of original_image_size is derived from the input https://github.com/dongluw/transformers/blob/cfef59b0d012002cea6ee16e7b68d2e9af0a4f44/src/transformers/models/cohere2_vision/image_processing_cohere2_vision_fast.py#L167-L169

IIUC it would make more sense to change the the function call above, since the input has order `(original_height, original_width)` while the output expects `num_columns, num_rows`, which is flipped

If you can point me to where this part is generated from, I can try to make a change there instead

dongluw: seems the function call is generated from `GotOcr2ImageProcessorFast`
https://github.com/dongluw/transformers/blob/cfef59b0d012002cea6ee16e7b68d2e9af0a4f44/src/transformers/models/got_ocr2/image_processing_got_ocr2_fast.py#L93-L95

however modifying this class would probably affect other models, so I think we can keep the `(h, w)` order in this pr. 

btw the grids is list of symmetric tuples that doesn't assume specific dim order, the fix of flipping dim order at the return statement still needs to be there IMO

zucchini-nlp: yeah, the grids do not assume specific order, The issue is that we need to choose one layout and follow it for consistency. And since the naming suggests that (w, h) layout is used in grids, I prefer to keep it consistent. Currently grids assume (w, h) and the number of columns also assume that layout is (w, h). The issue is in original size not following the same format and thus messing with aspect ratios

So for `original_size = np.stack([image_width, image_height])` looks an easier approach to me, instead of having to rename more variables for general consistency

dongluw: okay improved the naming

src/transformers/models/cohere2_vision/modular_cohere2_vision.py
```diff
@@ -295,8 +295,9 @@ def get_optimal_tiled_canvas(
     patch_size_height, patch_size_width = target_tile_size  # (height == width)
 
     candidate_resolutions = np.array(possible_resolutions) * patch_size_height
-    original_size = np.stack([image_height, image_width])
-    required_scales = candidate_resolutions / original_size
+    # tiles following (width, height) order to align with aspect ratio convention
+    tile_size = np.stack([image_width, image_height])
+    required_scales = candidate_resolutions / tile_size
```
zucchini-nlp, line 300: great, thanks for explicitly commenting out

zucchini-nlp (approved):
LGTM, thanks for iterating

PR #42644, opened 2025-12-05
qwen2 tok regression

itazap:
handle when merges are a list format - happens when loading from json


src/transformers/models/qwen2/tokenization_qwen2.py
```diff
@@ -63,7 +63,10 @@ def __init__(
             self._vocab = {
                 "<|endoftext|>": 0,
             }
-        self._merges = merges if merges is not None else generate_merges(self._vocab)
+        if merges is not None:
+            self._merges = [tuple(merge) if isinstance(merge, list) else merge for merge in merges]
+        else:
+            self._merges = generate_merges(self._vocab)
```
ArthurZucker, line 69: 42563 fixes it!

PR #42648, opened 2025-12-05
Lasr model

eustlb:


ArthurZucker (approved):
Reviewed offline, you can use utils/modular_utils to find closest attentio class (rope but not kv cache)

ArthurZucker:
Always assume it already exists

LysandreJik (approved):
Merging, failing tests unrelated

PR #42651, opened 2025-12-05
Only default `rope_parameters` to empty `dict` if there is something to put in it

hmellor:
`set(rope_parameters.keys()).issubset(self.layer_types)` will evaluate to `True` if `rope_parameters` is an empty `dict`.

This is a problem if `standardise_rope_params` is called on a config for a model which do not have RoPE because it will trigger `Case 2`. This method might be called on models which do not have RoPE in third party libraries (e.g. vLLM) in order to "upgrade" configs of custom models they use.

This PR makes `standardise_rope_params` harmless to call on configs with no RoPE at all.


src/transformers/modeling_rope_utils.py
```diff
@@ -654,13 +654,19 @@ def standardize_rope_params(self):
...
+        # If `rope_parameters` is not defined, but either `rope_theta` or `partial_rotary_factor` are, create it
+        if rope_parameters is None and (rope_theta is not None or partial_rotary_factor is not None):
+            rope_parameters = {}
 
+        # Case 0: no RoPE params defined
+        if rope_parameters is None:
+            return
         # Case 1: RoPE param keys do not intersect with possible `layer_types` -> one global dict
-        if getattr(self, "layer_types", None) is None or not set(rope_parameters.keys()).issubset(self.layer_types):
+        elif getattr(self, "layer_types", None) is None or not set(rope_parameters.keys()).issubset(self.layer_types):
```
zucchini-nlp: If the custom model has no rope parameters (i.e. it is `{}`), we will get triggered by the original issue of evaluting to `True`, no?

hmellor: You're right that if `rope_parameters` is explicitly set to `{}` we'll still get here. But if `rope_parameters` is unset, it will remain `None` and this method will exit early with a warning

src/transformers/modeling_rope_utils.py
```diff
@@ -654,13 +654,20 @@ def standardize_rope_params(self):
...
-
+        rope_parameters = getattr(self, "rope_parameters", None)
+        # If `rope_parameters` is not defined, but either `rope_theta` or `partial_rotary_factor` are, create it
+        if rope_parameters is None and (rope_theta is not None or partial_rotary_factor is not None):
+            rope_parameters = {}
+
+        # Case 0: no RoPE params defined
+        if rope_parameters is None:
+            logger.warning("`standardize_rope_params` was called but no RoPE parameters were found.")
+            return
```
zucchini-nlp, line 667: we could also adjust to catch cases when `rope_params = {}` set explicitly:

```python
rope_parameters = getattr(self, "rope_parameters", None) or {}
if not (rope_parameters or rope_theta): # partial_rotary_factor is optional and can't be used if we have no theta
    logger.warning("No rope related args in config found")

```

zucchini-nlp (approved):
Perfect

ArthurZucker (approved):
Thanks, merging!

PR #42654, opened 2025-12-05
[Quantization] Fix FP8 experts replacing

MekkCyber:
# What does this PR do?

In the 3D case for MoEs, `gate_up_proj` and `down_proj` are not modules, just parameters, so the experts weren't replaced as expected


SunMarc (approved):
Thanks !

rogeryoungh:
I have verified that this resolves the FP8Linear issue in #42028 .

PR #32189, opened 2024-07-24
[WIP] - Enable speculative decoding with batch size >1

kamilakesbi:
# What does this PR do?

This PR aims at solving issue #32165. 

I've started adapting code to enable speculative decoding with batch_size >1. I've reused some of the work in former PR #26875. 

# Main steps of the solution:  

When batch size > 1: 

1. Compute the number of similar tokens between candidate tokens and tokens obtained after doing a forward pass with the main model. This results in a tensor `n_matches` with the number of matches for each sequence in the batch. 

2. We keep the matching tokens. For that, we keep all tokens from the output to the main model with a sequence position inferior to `n_matches.max() + 1`. In doing so, we also retain some potential mismatched tokens, which we will deal with in the next steps using padding tokens. The resulting tensor, `input_ids`,  is thus in the form (`batch_size`, `n_matches` + 1).

3. We shift each sequence `i` in `input_ids` by `n_matches.max() - n_matches[i]`. The matching tokens are displaced to the right of `input_ids` and `n_matches.max() - n_matches[i]` padding tokens are added to the left.

5. Left cut: We cut all columns that contain only padding tokens. By design, these columns are to the left of the `input_ids`. In this way we keep the smallest possible `input_ids' that contain all the information needed to continue assisted generation. 

Steps 1 to 4 are the main addition to the the original speculative decoding loop described in detail in [this blog](https://huggingface.co/blog/assisted-generation) to enable assisted generation with BS > 1.

To make this work, we also need to adapt the computation of  the `attention_masks`, `past_key_values` and `position_ids` to take into account the shifted positions of the generated tokens. 

# To do: 

For now, I want to make this work with Whisper using this [snippet](https://github.com/huggingface/transformers/issues/32165). 

I've implemented steps 1 to 4 and adapted the computation of the `attention_masks` and `past_key_values` to handle the new padding tokens. 

I still need to make some adaptations with the `position_ids` to make this work properly. From what I can see: 

- For the main model (here `WhisperForConditionalGeneration`), `position_ids` are inferred directly from the attention_mask as we can see [here](https://github.com/huggingface/transformers/blob/7f5d644e69068825bb5b6e84cdc56b3d3a9bd04f/src/transformers/models/whisper/modeling_whisper.py#L1774). So if we pass the right attention mask to `generate` we should be good. 

[...]

Issue #32165, opened 2024-07-23
Enable speculative decoding with batch size >1


kamilakesbi:
### Feature request

Speculative decoding isn't currently enabled for batch sizes >1. PR #26875 was previously open to add this feature to main, but never merged. As the PR is quite old and has been closed, I'm opening an issue to motivate the addition of this feature to Transformers. 

Two approaches can be implemented to enable speculative decoding with batch size >1: 

1. **First approach**: at each step, we compare the token ids obtained with the assistant model with those obtained with the main model for all sequences in the batch, and roll back to the first incorrect token id in the batch. 

This is a simple approach, but rather naive, since some valid tokens would have to be regenerated during the successive iterations of the assistant model. 

2. **Second approach:** At each step, we dynamically roll back to the first mismatching token ID of each sequence in the batch, correcting with padding tokens and attention masks. 

In this way, at each step, we would keep all the valid tokens and wouldn't need to regenerate them during future iterations of the assistant model. 

The second approach is better, and PR #26875 already implements most of the solution, so we should focus on that one IMO.

## How to reproduce: 

```python
from transformers import pipeline, AutoModelForCausalLM, AutoModelForSpeechSeq2Seq, AutoProcessor
import torch
from datasets import load_dataset

device = "cuda:0" if torch.cuda.is_available() else "cpu"
torch_dtype = torch.float16 if torch.cuda.is_available() else torch.float32

assistant_model_id = "distil-whisper/distil-large-v2"

assistant_model = AutoModelForCausalLM.from_pretrained(
    assistant_model_id, torch_dtype=torch_dtype, low_cpu_mem_usage=True, use_safetensors=True
)
assistant_model.to(device)

model_id = "openai/whisper-large-v2"

model = AutoModelForSpeechSeq2Seq.from_pretrained(
    model_id, torch_dtype=torch_dtype, low_cpu_mem_usage=True, use_safetensors=True
)
model.to(device)

processor = AutoProcessor.from_pretrained(model_id)

pipe = pipeline(
    "automatic-speech-recognition",
    model=model,
    tokenizer=processor.tokenizer,
    feature_extractor=processor.feature_extractor,
    max_new_tokens=128,
    generate_kwargs={"assistant_model": assistant_model},
    torch_dtype=torch_dtype,
    chunk_length_s=15,
    batch_size=4,
    device=device,
)

dataset = load_dataset("distil-whisper/librispeech_long", "default", split="validation")
sample = dataset[0]["audio"]

[...]

Issue #29769, opened 2024-03-21
Support batch_size > 1 in assisted decoding


jiqing-feng:
### Feature request

Support batch_size > 1 in assisted decoding

### Motivation

With this support, we can provide more capability for assisted decoding, including beam search.

### Your contribution

I would like to submit a PR to enable this; I mainly need to cut and pad the past_key_values because each sequence may have a different length in each generation round.

Would like to hear your opinion @gante

gante:
@jiqing-feng yes! That is a great feature :) Bear in mind that you need to add padding between each iteration, and adapt the attention mask/position ids accordingly. Looking forward to the PR

cc @zucchini-nlp

jiqing-feng:
I reconsidered it and found it is complicated and inefficient (the latency depends on the lowest sequence in the batch, and there is also many of padding and other ops, which will bring extra latency and memory). Maybe continuous batching is the correct way.

WindChimeRan:
Batch Speculative Decoding Done Right: https://arxiv.org/pdf/2510.22876
code: https://github.com/eBay/spec_dec

Implemented Unpad-Append-Repad Batch Speculative Decoding for input id, position id and contiguous kv cache

PR #26875, opened 2023-10-17
[WIP][Speculative Decoding] Enable batch size > 1

patrickvonplaten:
# What does this PR do?

This PR enables speculative decoding when batch size is > 1.

This PR has been tested to work correctly for Whisper. You can try it as follows:

```py
from transformers import pipeline, AutoModelForCausalLM, AutoModelForSpeechSeq2Seq, AutoProcessor
import torch
from datasets import load_dataset

device = "cuda:0" if torch.cuda.is_available() else "cpu"
torch_dtype = torch.float16 if torch.cuda.is_available() else torch.float32

assistant_model_id = "distil-whisper/distil-large-v2"

assistant_model = AutoModelForCausalLM.from_pretrained(
    assistant_model_id, torch_dtype=torch_dtype, low_cpu_mem_usage=True, use_safetensors=True
)
assistant_model.to(device)

model_id = "openai/whisper-large-v2"

model = AutoModelForSpeechSeq2Seq.from_pretrained(
    model_id, torch_dtype=torch_dtype, low_cpu_mem_usage=True, use_safetensors=True
)
model.to(device)

processor = AutoProcessor.from_pretrained(model_id)

pipe = pipeline(
    "automatic-speech-recognition",
    model=model,
    tokenizer=processor.tokenizer,
    feature_extractor=processor.feature_extractor,
    max_new_tokens=128,
    generate_kwargs={"assistant_model": assistant_model},
    torch_dtype=torch_dtype,
    chunk_length_s=15,
    batch_size=4,
    device=device,
)

dataset = load_dataset("distil-whisper/librispeech_long", "default", split="validation")
sample = dataset[0]["audio"]

result = pipe(sample)
print(result["text"])
```

## TODOS
- [x] Test that it works correctly for Whisper
- [ ] Test for Llama
- [ ] Write nice tests
- [ ] Clean code
- [ ] Write docs

PR #42655, opened 2025-12-05
New Feature: Enabling Speculative Decoding with Batch Size > 1 (If draft and target model share tokenizer)

YanivDorGalron:
# What does this PR do?
Many discussions focused on speculative decoding with 'batch_size > 1' (#26875, #29769, #32165, #32189), but none were fully implemented. This PR does that.

### **Summary of Changes: Enable Batched Speculative Decoding**

Until now, invoking `generate()` with an `assistant_model` raised a `ValueError` if `batch_size > 1`. This update modifies the codebase to support batched speculative decoding.

#### **Key Features & Limitations**
* **Batch Support:** Enabled `batch_size > 1` for standard speculative decoding.
* **Tokenizer Constraint:** Batched execution is currently **restricted to scenarios where the draft (assistant) and target (main) models share the same tokenizer**.
    * *Universal Assisted Generation (UAG)*—where models use different tokenizers—remains restricted to `batch_size = 1` and will still raise a `ValueError` if used with a batch.

#### **Technical Implementation Details**
Supporting batches requires dealing with "ragged tensors", where different sequences in a batch accept a different number of speculative tokens, leading to misalignment in tensor shapes and KV caches. The implementation addresses this via the following mechanisms:

**1. Assistant Cache Alignment**
* When reusing the assistant model for the next draft generation, the previous KV cache must match the current state of the sequences.
* `align_cache` function was implemented in `candidate_generator.py`. It identify which parts of the old cache match the new inputs and removing the irrelevant keys and values from the cache. This ensures the assistant's cache correctly reflects the accepted tokens.

**2. Repadding Inputs and Main Model Cache**
* After verification, if Batch Item A accepts 2 tokens and Batch Item B accepts 4, the resulting input tensor becomes ragged.
* The rejected tokens and the padding tokens are removed from each item and the entire batch is repadded. The same is happening fo the main model KV cache

**3. Vectorized Verification & Sampling**
* **Greedy Search:** The acceptance logic (checking if candidate tokens match target output) now uses vectorized `cumsum` operations to identify the index of the first mismatch for every item in the batch simultaneously.
* **Sampling:** `_speculative_sampling` has been updated to calculate acceptance probabilities ($p$ vs $q$) and handle token rejection/resampling for the entire batch in parallel.

**4. Dynamic Heuristics (`update_candidate_strategy`)**
* The heuristic that adjusts the number of assistant tokens (`num_assistant_tokens`) now aggregates statistics across the batch (e.g., using the average number of matches) to determine if the draft model step size should increase or decrease.
* For the dynamic confidence threshold (ROC curve calculation), matches and probabilities are collected per batch item, flattened, and then used to update the threshold.

#### **Note on `max_new_tokens`**
When using speculative decoding with `batch_size > 1`, the behavior regarding `max_new_tokens` requires nuance. Because the generation loop serves the entire batch, the process continues until one items satisfy the stopping criteria. However, due to the repadding and varying acceptance rates, other items in the batch may effectively stop generating before reaching the global `max_new_tokens` limit.

### Batch Size = 3: Assistant vs Standard Decoding
WITHOUT assistant_model:  9.2215 seconds
WITH assistant_model:         7.4744 seconds

| Sequence | # Tokens w Assistant | # Tokens w/o Assistant |
|----------|----------------|-----------------|
| 1        | 50             | 50              |
| 2        | 41             | 50              |
| 3        | 40             | 50              |

**I look forward to your response and comments to ensure this can be included in the next version.**

### DIFF
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -3693,8 +3693,8 @@ class GenerationMixin(ContinuousMixin):
                 candidate_kwargs,
                 candidate_input_ids.shape[1],
                 self.config.is_encoder_decoder,
-                candidate_input_ids,
-                pad_token_id,
+                candidate_input_ids if batch_size > 1 else None,
+                pad_token_id if batch_size > 1 else None,
             )
             candidate_kwargs = _prepare_token_type_ids(candidate_kwargs, candidate_input_ids.shape[1])
             if "cache_position" in candidate_kwargs:
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -3666,6 +3666,7 @@ class GenerationMixin(ContinuousMixin):
         is_first_iteration = True  # to preserve the same API in the output as other generation methods
         assistant_ids_in_cache = None
         pad_token_id = generation_config._pad_token_tensor
+        decoder_start_token_tensor = generation_config._decoder_start_token_tensor
         while self._has_unfinished_sequences(this_peer_finished, synced_gpus, device=input_ids.device):
             cur_len = input_ids.shape[1]
 
@@ -3781,7 +3782,12 @@ class GenerationMixin(ContinuousMixin):
 
             # 4.1. Get the valid continuation, after the matching tokens
             input_ids, outputs.past_key_values = repadd_batch_and_fix_cache(
-                input_ids, outputs.past_key_values, valid_tokens_padded, pad_token_id
+                input_ids,
+                outputs.past_key_values,
+                valid_tokens_padded,
+                pad_token_id,
+                self.config.is_encoder_decoder,
+                decoder_start_token_tensor,
             )
             if streamer is not None:
                 streamer.put(
@@ -3945,7 +3951,6 @@ def _speculative_sampling(
     """
     Applies sampling as in the speculative decoding paper (https://huggingface.co/papers/2211.17192, algorithm 1). Returns
     the selected tokens, as well as the number of candidate matches.
-
     NOTE: Unless otherwise stated, the variable names match those in the paper.
     """
     new_candidate_input_ids = candidate_input_ids[:, -candidate_length:]
@@ -4027,7 +4032,9 @@ def _split_model_outputs(outputs, new_outputs, cur_len, added_len, is_decoder_at
     return outputs
 
 
-def repadd_batch_and_fix_cache(input_ids, past_key_values, accepted_tokens_padded, pad_token_id):
+def repadd_batch_and_fix_cache(
+    input_ids, past_key_values, accepted_tokens_padded, pad_token_id, is_encoder_decoder, decoder_start_token_tensor
+):
     """
     params
         input_ids: the input ids of the model (without the candidate tokens). shape: [B, S]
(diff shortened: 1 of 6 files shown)


YanivDorGalron:
Hi!
Before I spend more time resolving the remaining failing tests, I’d like to confirm whether adding batched speculative decoding is something the team is open to including in the library.
If you have any guidance or a preferred direction for how this should integrate with the existing generation code, I’d be happy to align with it.
@zucchini-nlp @gante

Thanks!

zucchini-nlp:
@YanivDorGalron hey! Thanks for the PR

It is a nice-to-have feature for sure. We have been delaying it since an implementation would be complicated unless we can use a `ragged_tensor` natively from torch. As you might have already noticed, there are several edge cases with assisted decoding and we don't want to increase maintaining costs even more

I will definitely take a closer look around this week

src/transformers/generation/utils.py
```diff
@@ -3972,3 +4030,42 @@ def _split_model_outputs(outputs, new_outputs, cur_len, added_len, is_decoder_at
...
+    # we remove the last token because it is the bonus token and it does not appear in the cache.
+    cache_input_ids = next_input_ids.clone()
+    # last non zero token in each row is set to 0 because it is the bonus token and it does not appear in the cache.
+    cache_input_ids[torch.arange(cache_input_ids.shape[0]), (cache_input_ids != pad_token_id).cumsum(1).argmax(1)] = (
+        pad_token_id
+    )
+    padding_mask = cache_input_ids[:, :-1] == pad_token_id
+    past_key_values.compress_and_repad_cache(padding_mask)
+    # 1. Filter out current padding and repad to minimum length.
+    next_input_ids_clean = [row[row != pad_token_id] for row in next_input_ids]
```
zucchini-nlp, line 4065: nice!

zucchini-nlp (commented):
Hey again, sorry for the delay

I get the idea behind and I think it is in a good direction. I liked that we don't pad infinitely and the next input ids are re-padded back. Though the cache logic look more involved as we need to basically re-create a new padded cache each time. I personally prefer to not have more code in `cache_utils`. Since it is creating new cache from an old, we can separate it out to a helper fn

Given the complexities of the feature and since it wasn't being requested much recently, I suggest us to move it to hub via remote code. Here is more info on [custom generate](https://huggingface.co/docs/transformers/en/generation_strategies#custom-generation-methods). We can add it in transformers-community org and communicate to users. If the feature gets traction and interest from community, we will consider adding it in core library. 

That will help us to make sure that the feature is shipped fast bypassing the review cycle, we can get feedback from community fast. Also it helps to not put maintenance burden for a rarely-used feature. LMK if that sounds good to you and I can help with making a custom generate repo

YanivDorGalron:
Thanks for your response,  that sounds great.

If you can create a repo under `transformers-community` and add me as a collaborator, I’m happy to take ownership of the implementation. In the meantime, I’ll start refactoring the cache logic into a helper and prepare a minimal structure so we can plug it in once the repo is ready.

Looking forward to it!

zucchini-nlp:
@YanivDorGalron sure, what is you HF handle?

YanivDorGalron:
My HF handle is `yaniv-galron`

> @YanivDorGalron sure, what is you HF handle?

zucchini-nlp:
I will ask @ArthurZucker to add, just noticed I am not admin 😄 

Most maintainer are now in winter vacations so feel free in the meantime experiment with a repo under your own profile/org.

WindChimeRan:
Relevant paper using unpad-append-repad for batch speculative decoding: https://arxiv.org/pdf/2510.22876

The code is open sourced. 

May I ask what's the current progress for this PR?

YanivDorGalron:
> Relevant paper using unpad-append-repad for batch speculative decoding: https://arxiv.org/pdf/2510.22876
> 
> The code is open sourced.
> 
> May I ask what's the current progress for this PR?

Hey @WindChimeRan, we have decided to move this code to a custom generate which we will publish on the hub.
This is still not done and your help will be appreciated.
@zucchini-nlp @ArthurZucker can you create a repo under transformers-community and add me as a collaborator?

WindChimeRan:
> > Relevant paper using unpad-append-repad for batch speculative decoding: https://arxiv.org/pdf/2510.22876
> > The code is open sourced.
> > May I ask what's the current progress for this PR?
> 
> Hey @WindChimeRan, we have decided to move this code to a custom generate which we will publish on the hub. This is still not done and your help will be appreciated. @zucchini-nlp @ArthurZucker can you create a repo under transformers-community and add me as a collaborator?

@YanivDorGalron Sure! I'm glad to help!

PR #42660, opened 2025-12-05
[`mRope`] Fix warnings

vasqu:
As per title, `mrope_section` was the keyword, not `mrope` leading to incorrect warnings


zucchini-nlp (approved):
Oops, right, must have been a typo

PR #42663, opened 2025-12-05
FIX Error when trying to load non-LoRA PEFT

BenjaminBossan:
# What does this PR do?

This PR fixes a bug that prevented non-LoRA PEFT adapters to be loaded into a transformers model. It was just a very simple logical error where a check was performed generally when it actually only is relevant for hotswapping. A test for this has been added.

Additionally, also testing if a non-LoRA adapter can be added to a transformers model. This was not broken but still lacked test coverage.

## Before submitting
      Pull Request section?
      to it if that's the case.


BenjaminBossan:
Not sure who would be best suited to review this from among the transformers devs, pinging @Cyrilvallez because you reviewed the original hotswap PR.

Personally, I'd consider this a regression introduced by #41297 so ideally it could be part of the v5 release :)

src/transformers/integrations/peft.py
```diff
@@ -279,7 +279,7 @@ def load_adapter(
             )
             peft_config.inference_mode = not is_trainable
 
-        if peft_config.peft_type != PeftType.LORA:
+        if hotswap and (peft_config.peft_type != PeftType.LORA):
```
Cyrilvallez: Actually, we can fully remove this check no? It's already checked above L206

Cyrilvallez (commented):
Looks like the check is already performed above no? But I still think we should probably update the default value of `hotswap` in the function?? See also my comment here https://github.com/huggingface/transformers/pull/41297#discussion_r2436599593

src/transformers/integrations/peft.py
BenjaminBossan: Indeed, great catch. I removed it and confirmed that the corresponding test still passes. Please review again. Failing `test_tokenization_mistral_common.py` seem to be unrelated.

Cyrilvallez (approved):
Alright, LGTM then! Will merge as the failing test is unrelated

BenjaminBossan:
> Alright, LGTM then! Will merge as the failing test is unrelated

Fantastic, thanks. Anything I should do to ensure that this is included in v5?

Cyrilvallez:
Nop, will be automatically in the next rc1 release!

PR #42666, opened 2025-12-05
Remove Neptune integration references and deprecate `NeptuneCallback`

qgallouedec:
image


src/transformers/integrations/integration_utils.py
```diff
@@ -1455,6 +1456,10 @@ def __init__(self):
 class NeptuneCallback(TrainerCallback):
```
tomaarsen, line 1456: I'm not very familiar with the modern `transformers` deprecation approach, but do we have a nice decorator here that can inform IDE's that objects are deprecated, perhaps?
image

src/transformers/integrations/integration_utils.py
```diff
@@ -1500,6 +1505,11 @@ def __init__(
         log_checkpoints: str | None = None,
         **neptune_run_kwargs,
     ):
+        warnings.warn(
+            "The NeptuneCallback is deprecated and will be removed in a future version of Transformers. We recommend "
+            "using other supported experiment tracking integrations.",
+            FutureWarning,
+        )
```
SunMarc, line 1512: Thanks ! We can keep this deprecation message for now as it still works but I've checked that their services will go down on after a 3‑month transition period ending on March 5, 2026. We can remove it after this date !

SunMarc (approved):
LGTM ! Thanks for the PR

PR #42686, opened 2025-12-07
Fix: Skip weight initialization for quantized int8 models

AnjaliiD:
This PR fixes the RuntimeError encountered when loading quantized models (int8 weights). 

The issue occurs because PyTorch’s `normal_()` cannot operate on integer tensors. 

Changes:
- Skip weight initialization for quantized models in `_load_pretrained_model()`
- Print a small info message when skipping

Fixes #42574

- @SunMarc
- @MekkCyber


SunMarc:
A lot of refactor happened since the issue was posted. I don't think this is relevant anymore, so i'm closing this !

PR #42688, opened 2025-12-07
update gemma3 exepectations and switch to dynamic cache

Abdennacer-Badaoui:
When using the ROCm platform, some tests that process multimodal inputs with a static cache produce incomplete outputs. They generate only one token and then immediately output the EOS token, because the logits become NaN.

It appears that the issue comes from `index_copy_` operations in `cache_utils.py`, which fail silently. Replacing these operations with slice assignment, which is more reliable, makes some of these tests pass. It may also be related to the model being compiled while using a static cache. However, changing the cache system for a single model failing on a single hardware platform is probably not a good idea.

For now, I switched to the dynamic cache, which produces correct outputs.

### DIFF
--- a/src/transformers/cache_utils.py
+++ b/src/transformers/cache_utils.py
@@ -340,11 +340,18 @@ class StaticLayer(CacheLayerMixin):
                 self.values.index_copy_(2, cache_position, value_states)
                 return self.keys, self.values
             except NotImplementedError:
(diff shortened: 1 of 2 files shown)


remi-or:
run-slow: gemma3

remi-or (approved):
LGTM!

PR #42696, opened 2025-12-08
Feature/parallelize model download clean

RituAddepalli:
Parallelize Multi-File Downloads in cached_files()

This PR improves the cached_files() function in hub.py by parallelizing multi-file downloads (such as sharded model weights). The change is fully backward-compatible and does not alter any existing logic or file-resolution behavior.

**🔍 Before**

Single file → hf_hub_download()

Multiple files → snapshot_download() (downloads sequentially)

**🚀 After (This PR)**

Single file → still uses hf_hub_download()

Multiple files → each file is downloaded in parallel using:

concurrent.futures.ThreadPoolExecutor

Target concurrency: min(8, number_of_files)

This improves download performance for multi-file models while keeping all caching and path resolution identical to before.
### 🧩 Implementation (Core Snippet)
if len(full_filenames) == 1:
    return [hf_hub_download(...)]

def download_file(file):
    return hf_hub_download(
        path_or_repo_id,
        file,
        subfolder=None if len(subfolder) == 0 else subfolder,
        repo_type=repo_type,
        revision=revision,
        cache_dir=cache_dir,
        user_agent=user_agent,
        force_download=force_download,
        proxies=proxies,
        token=token,
        local_files_only=local_files_only,
    )

with ThreadPoolExecutor(max_workers=min(8, len(full_filenames))) as executor:
    resolved_files = list(executor.map(download_file, full_filenames))

return resolved_files
### 📈 Impact

✔️ No functional change to single-file downloads

✔️ Multi-file downloads are now notably faster

✔️ Compatible with all existing cache behaviors

✔️ Does not affect snapshot structure or weight loading

✔️ Minimal, isolated change, easy to maintain and extend

This PR provides a clean foundation for future parallel weight‐loading improvements.
### 🧪 Tests

Local manual tests performed:

**Test 1: Parallel download of 3 model files**

Verified that:

All file downloads start at the same timestamp

Downloads complete faster than sequential behavior

All files correctly appear in the HF cache structure

Sample output:
Screenshot (466)

**Test 2: Parallel download of 3 full models**

Models: bert-base-uncased, distilbert-base-uncased, roberta-base

All resolved files match correct HF cache layout

No interference or cross-pollution between model caches

sample output:
Screenshot (457)

### 🔄 Other Changes

- Fixed circular import in hub.py
- Updated logging imports
- Applied ruff formatting fixes (hub.py, hf_logging.py)
- Rebased on top of the latest huggingface/main

### 🙏 Request for Review
@huggingface/transformers-maintainers
@CyrilVallez

Happy to update or revise anything based on feedback!

### DIFF
--- a/.circleci/create_circleci_config.py
+++ b/.circleci/create_circleci_config.py
@@ -32,7 +32,7 @@ COMMON_ENV_VARIABLES = {
     "DISABLE_SAFETENSORS_CONVERSION": True,
 }
 # Disable the use of {"s": None} as the output is way too long, causing the navigation on CircleCI impractical
-COMMON_PYTEST_OPTIONS = {"max-worker-restart": 0, "vvv": None, "rsfE":None}
+COMMON_PYTEST_OPTIONS = {"max-worker-restart": 0, "vvv": None, "rsfE": None}
 DEFAULT_DOCKER_IMAGE = [{"image": "cimg/python:3.8.12"}]
 
 # Strings that commonly appear in the output of flaky tests when they fail. These are used with `pytest-rerunfailures`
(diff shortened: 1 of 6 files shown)


Cyrilvallez:
@RituAddepalli It's the 2nd time you tag me on almost the same PR without understanding at all what you're doing and letting AI do it all. If you keep doing this, we will find ourselves force to restrict you from our repos

PR #42697, opened 2025-12-08
batched and grouped experts implementations

IlyasMoutawwakil:
# What does this PR do?

I have started experimenting with pure pytorch MoE implementations following the [HF exporters PR](https://github.com/huggingface/transformers/pull/41992) while trying to find a traceable/exportable variant for onnx/openvino.

In this PR I copy the `attn_implementation` API into a similar `experts_implementation` API, and added two new implementations:
- `batched_mm` (the exportable one) which uses `torch.bmm`, is fastest on single batch size / small inputs.
- `grouped_mm` (the pytorch custom kernel one) inspired from torchtitan's moe imp (using `torch._grouped_mm`), which is generally fastest.

# benchmark
An initial benchmark shows promising results on (A100), I know that the `torch._grouped_mm` uses bfloat16 or something under the hood, so these might not be apples to apples (i'm still looking for more references on this function and how to use it "equivalently")

# MoE Implementations Benchmark

Benchmark script: [bench.py](https://github.com/user-attachments/files/24125816/bench.py)

It uses qwen2_moe ("Qwen/Qwen1.5-MoE-A2.7B", bfloat16) where latency and memory are for the forward pass / prefill

TLDR; for very small inputs batched_mm can be extremely fast and even faster with compilation, for bigger inputs grouped_mm is unbeatable but it doesn't seem to get much faster with torch compilation.

| Batch Size | Seq Length | Torch Compile              | Implementation | Mean Latency (ms) | Median Latency (ms) | P90 Latency (ms) | Peak Mem (MB) |
| ---------- | ---------- | -------------------------- | -------------- | ----------------- | ------------------- | ---------------- | ------------- |
| 1          | 16         | False                      | eager          | 271.80            | 272.94              | 295.34           | 27324.65      |
| 1          | 16         | True                       | eager          | 351.86            | 351.64              | 384.64           | 27329.29      |
| 1          | 16         | max-autotune-no-cudagraphs | eager          | 352.52            | 352.15              | 382.79           | 27329.29      |
| 1          | 16         | False                      | batched_mm     | 52.03             | 52.07               | 52.67            | 28382.50      |
| 1          | 16         | True                       | batched_mm     | 53.04             | 53.04               | 53.11            | 28029.63      |
| 1          | 16         | max-autotune-no-cudagraphs | batched_mm     | 23.87             | 23.86               | 24.02            | 27329.29      |
| 1          | 16         | False                      | grouped_mm     | 64.27             | 64.09               | 65.49            | 27329.29      |
| 1          | 16         | True                       | grouped_mm     | 59.45             | 59.52               | 60.99            | 27329.29      |
| 1          | 16         | max-autotune-no-cudagraphs | grouped_mm     | 59.61             | 59.55               | 60.89            | 27329.29      |
| 1          | 128        | False                      | eager          | 471.73            | 472.65              | 487.97           | 27396.46      |
| 1          | 128        | True                       | eager          | 637.32            | 613.70              | 845.01           | 27429.82      |
| 1          | 128        | max-autotune-no-cudagraphs | eager          | 620.21            | 619.35              | 657.74           | 27429.82      |
| 1          | 128        | False                      | batched_mm     | 316.67            | 316.94              | 317.92           | 35854.56      |
| 1          | 128        | True                       | batched_mm     | 370.29            | 370.29              | 370.57           | 33031.64      |
| 1          | 128        | max-autotune-no-cudagraphs | batched_mm     | 151.87            | 150.38              | 158.01           | 27429.82      |
| 1          | 128        | False                      | grouped_mm     | 78.50             | 78.53               | 80.00            | 27429.82      |
| 1          | 128        | True                       | grouped_mm     | 72.95             | 72.99               | 74.60            | 27429.82      |
| 1          | 128        | max-autotune-no-cudagraphs | grouped_mm     | 72.71             | 72.89               | 73.55            | 27429.82      |
| 4          | 16         | False                      | eager          | 431.87            | 433.38              | 448.01           | 27391.57      |
| 4          | 16         | True                       | eager          | 566.63            | 569.74              | 598.98           | 27372.12      |
| 4          | 16         | max-autotune-no-cudagraphs | eager          | 563.13            | 567.79              | 588.25           | 27372.12      |
| 4          | 16         | False                      | batched_mm     | 163.41            | 163.38              | 164.84           | 31585.54      |
| 4          | 16         | True                       | batched_mm     | 189.18            | 189.08              | 189.79           | 30173.45      |
| 4          | 16         | max-autotune-no-cudagraphs | batched_mm     | 79.15             | 79.10               | 79.74            | 27372.11      |
| 4          | 16         | False                      | grouped_mm     | 75.23             | 75.18               | 76.74            | 27372.11      |
| 4          | 16         | True                       | grouped_mm     | 70.35             | 70.40               | 71.71            | 27372.12      |
| 4          | 16         | max-autotune-no-cudagraphs | grouped_mm     | 70.26             | 70.43               | 71.32            | 27372.12      |
| 4          | 128        | False                      | eager          | 526.88            | 522.75              | 570.01           | 27632.62      |
| 4          | 128        | True                       | eager          | 678.18            | 677.54              | 690.97           | 27762.46      |
| 4          | 128        | max-autotune-no-cudagraphs | eager          | 676.22            | 677.07              | 681.91           | 27762.45      |
| 4          | 128        | False                      | batched_mm     | 1235.25           | 1235.33             | 1237.90          | 61465.85      |
| 4          | 128        | True                       | batched_mm     | 1505.00           | 1503.31             | 1536.10          | 50174.26      |
| 4          | 128        | max-autotune-no-cudagraphs | batched_mm     | 572.37            | 570.81              | 589.74           | 27762.45      |
| 4          | 128        | False                      | grouped_mm     | 80.95             | 81.06               | 81.70            | 27762.45      |
| 4          | 128        | True                       | grouped_mm     | 79.67             | 79.69               | 80.54            | 27762.45      |
| 4          | 128        | max-autotune-no-cudagraphs | grouped_mm     | 83.29             | 79.83               | 111.83           | 27762.46      |

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


src/transformers/integrations/moe.py
```diff
@@ -0,0 +1,205 @@
...
+
+    # tokens per expert -> offsets for grouped_mm (int32)
+    num_tokens_per_expert = torch.bincount(expert_ids_g, minlength=num_experts)
+    offsets = torch.cumsum(num_tokens_per_expert, dim=0, dtype=torch.int32)
+
+    # Important: torch._grouped_mm requires mat_a.dtype == out_dtype when out_dtype is provided.
+    # Important: torch._grouped_mm requires mat_a and mat_b to have strides that are multiples of 16
+    # still can't find a reference for this constraint but I had models failing if not respected
+    mat_a_up = current_states_g
+    mat_b_up = gate_up_proj.transpose(-2, -1)
```
ArthurZucker: if you transpose then maybe this one can be done in the online weight conversion

src/transformers/integrations/moe.py
```diff
@@ -0,0 +1,205 @@
...
+
+    if mat_a_up.stride(1) % 16 != 0:
+        mat_a_up = _make_stride_multiple_of(mat_a_up, 1, 16)
+    if mat_b_up.stride(1) % 16 != 0:
+        mat_b_up = _make_stride_multiple_of(mat_b_up, 1, 16)
```
ArthurZucker: bit weird to do this on each forward are we not wasting time there?

src/transformers/integrations/moe.py
```diff
@@ -0,0 +1,205 @@
...
+
+    gate_up_out = torch._grouped_mm(mat_a_up, mat_b_up, offs=offsets).to(current_states_g.dtype)
+
+    gate, up = gate_up_out.chunk(2, dim=-1)
+    hidden_after_activation = act_fn(gate) * up  # (S, intermediate_dim)
+
+    # --- Down projection per expert (grouped_mm) ---
+    mat_a_down = hidden_after_activation
+    mat_b_down = down_proj.transpose(-2, -1)
```
ArthurZucker: same here, the way we want v5 is to have "perfect" weights with the weight converter -> this can be don ein the weight converter

src/transformers/integrations/moe.py
```diff
@@ -0,0 +1,205 @@
...
+    selected_down = down_proj[expert_ids]  # (S, hidden_dim, intermediate_dim)
+    out_per_sample = torch.bmm(selected_down, hidden_after_activation.unsqueeze(-1)).squeeze(-1)  # (S, hidden_dim)
+
+    # Apply routing weights and cast to output dtype
+    out_per_sample = out_per_sample * sample_weights.unsqueeze(-1).to(out_per_sample.dtype)  # (S, hidden_dim)
+
+    # Accumulate results back to the final_hidden_states using original token indices
+    final_hidden_states.index_add_(0, token_idx, out_per_sample.to(final_hidden_states.dtype))
+
+    return final_hidden_states
```
ArthurZucker, line 123: I like this one more because it does not have the additional transpose. 
How does it compare for different num experts setups?

src/transformers/integrations/moe.py
ArthurZucker, line 1: missing licence. 

If we go down that road, which I like TBH, we should also add kernels supports + try at least for FP8 to see how this would work. 

Because the other solution is to use `use_hf_hub_kernel` decorator as well but looks more cumbersome. So following FA2 we want to support kernels from the hub in this as well + quantization

This does look nice, for the bench can you add compile cases as well please? 

Also we need to make sure this works with TP / EP

src/transformers/integrations/moe.py
```diff
@@ -0,0 +1,205 @@
...
+def _make_stride_multiple_of(t: torch.Tensor, dim: int, multiple: int):
+    stride = t.stride(dim)
+    if stride % multiple == 0:
+        return t
+    elem_size = t.element_size()
+    align_elems = max(1, multiple // elem_size)
+    k = t.shape[dim]
+    pad = (-k) % align_elems
+    return _pad_dim_end(t, dim, pad)
+
```
ArthurZucker: any of these should in some way update the weights -> caching the changes we do as long as they are small might need us to have access to the state

src/transformers/integrations/moe.py
IlyasMoutawwakil: the same transposition is done in eager (inside the torch.nn.functional.linear), it's also a no-op in the case of strided tensors

src/transformers/integrations/moe.py
IlyasMoutawwakil: it's actually rare that we hit this part of the code, it only happens on the tiny models / configs used for testing, for example with deepsek_v2, we have moe_intermediate_size=12 which errors with bf16 and fp16 because 12x2 bytes = 24 not divisible by 16, but works fine with fp32 because 12x4 bytes = 48 divisible by 16.
I added this fix in general to force the 16 bytes stride but we can also change the moe intermediate state to 16 and get the tests passing.

IlyasMoutawwakil: moe_intermediate_size=8 also wroks (since bf16/fp16 will bring the 2 multiplier)
wdyt ? I think we should just change the moe intermediate size, in these tests, no need to force our matrices to be compliant with the kernel since it will be the case in all real world model configs

src/transformers/integrations/moe.py
IlyasMoutawwakil: it's rare for this to be triggered, the grouped mm kernel is very particular about the layout of its inputs https://github.com/pytorch/pytorch/blob/23761d4f8149aaa16649c5494e57d53f192cf0f2/aten/src/ATen/native/GroupedMMUtils.h#L19 see my reply below for more details

src/transformers/integrations/moe.py
IlyasMoutawwakil: I went ahead and removed the need for these by modifying the config of the testing models

stevhliu:
subscribing so i can add some docs for this once it's ready :)

src/transformers/integrations/moe.py
IlyasMoutawwakil: I have a commit locally with the transposition removed from both eager (using matmul instead of linear) and grouped_mm. I reran the same benchmark above but can't say for sure if it's faster

# Experts Implementations Benchmark Results

| Batch Size | Seq Length | Torch Compile              | Implementation | Mean Latency (ms) | Median Latency (ms) | P90 Latency (ms) | Peak Mem (MB) |
| ---------- | ---------- | -------------------------- | -------------- | ----------------- | ------------------- | ---------------- | ------------- |
| 1          | 16         | False                      | eager          | 264.84            | 264.32              | 288.75           | 27324.65      |
| 1          | 16         | True                       | eager          | 350.00            | 349.41              | 376.13           | 27329.29      |
| 1          | 16         | max-autotune-no-cudagraphs | eager          | 341.95            | 342.61              | 374.87           | 27329.29      |
| 1          | 16         | False                      | batched_mm     | 51.97             | 51.99               | 52.64            | 28382.50      |
| 1          | 16         | True                       | batched_mm     | 53.10             | 53.07               | 53.39            | 28029.63      |
| 1          | 16         | max-autotune-no-cudagraphs | batched_mm     | 23.48             | 23.51               | 23.54            | 27329.29      |
| 1          | 16         | False                      | grouped_mm     | 63.18             | 63.08               | 64.82            | 27329.29      |
| 1          | 16         | True                       | grouped_mm     | 59.43             | 59.41               | 60.98            | 27329.29      |
| 1          | 16         | max-autotune-no-cudagraphs | grouped_mm     | 60.68             | 60.76               | 62.36            | 27329.29      |
| 1          | 128        | False                      | eager          | 490.27            | 488.27              | 504.51           | 27396.46      |
| 1          | 128        | True                       | eager          | 671.43            | 637.76              | 1008.31          | 27429.82      |
| 1          | 128        | max-autotune-no-cudagraphs | eager          | 618.78            | 620.04              | 639.69           | 27429.82      |
| 1          | 128        | False                      | batched_mm     | 316.88            | 317.25              | 317.86           | 35854.56      |
| 1          | 128        | True                       | batched_mm     | 370.47            | 370.36              | 371.28           | 33031.64      |
| 1          | 128        | max-autotune-no-cudagraphs | batched_mm     | 152.58            | 150.86              | 159.43           | 27429.82      |
| 1          | 128        | False                      | grouped_mm     | 77.71             | 77.88               | 78.76            | 27429.82      |
| 1          | 128        | True                       | grouped_mm     | 72.99             | 73.06               | 74.21            | 27429.82      |
| 1          | 128        | max-autotune-no-cudagraphs | grouped_mm     | 72.67             | 72.94               | 73.60            | 27429.82      |
| 4          | 16         | False                      | eager          | 433.87            | 431.03              | 455.82           | 27391.57      |
| 4          | 16         | True                       | eager          | 569.67            | 571.82              | 586.81           | 27372.12      |
| 4          | 16         | max-autotune-no-cudagraphs | eager          | 553.63            | 557.50              | 579.73           | 27372.12      |
| 4          | 16         | False                      | batched_mm     | 164.00            | 164.10              | 164.77           | 31585.54      |
| 4          | 16         | True                       | batched_mm     | 189.24            | 189.19              | 189.62           | 30173.45      |
| 4          | 16         | max-autotune-no-cudagraphs | batched_mm     | 79.59             | 79.45               | 80.01            | 27372.11      |
| 4          | 16         | False                      | grouped_mm     | 75.40             | 75.04               | 78.37            | 27372.11      |
| 4          | 16         | True                       | grouped_mm     | 69.93             | 70.10               | 71.27            | 27372.12      |
| 4          | 16         | max-autotune-no-cudagraphs | grouped_mm     | 69.79             | 69.97               | 71.58            | 27372.12      |
| 4          | 128        | False                      | eager          | 524.48            | 520.53              | 561.31           | 27632.62      |
| 4          | 128        | True                       | eager          | 702.86            | 702.79              | 716.28           | 27762.46      |
| 4          | 128        | max-autotune-no-cudagraphs | eager          | 687.21            | 682.97              | 716.89           | 27762.45      |
| 4          | 128        | False                      | batched_mm     | 1236.74           | 1236.75             | 1239.48          | 61465.86      |
| 4          | 128        | True                       | batched_mm     | 1469.06           | 1468.82             | 1470.12          | 50174.26      |
| 4          | 128        | max-autotune-no-cudagraphs | batched_mm     | 570.72            | 570.03              | 576.75           | 27762.45      |
| 4          | 128        | False                      | grouped_mm     | 81.61             | 81.54               | 82.90            | 27762.45      |
| 4          | 128        | True                       | grouped_mm     | 79.41             | 79.44               | 79.85            | 27762.46      |
| 4          | 128        | max-autotune-no-cudagraphs | grouped_mm     | 79.44             | 79.50               | 80.12            | 27762.45      |

src/transformers/configuration_utils.py
```diff
@@ -413,6 +416,28 @@ def _attn_implementation(self, value: str | dict | None):
                 )
                 subconfig._attn_implementation = sub_implementation
 
+    @property
+    def _experts_implementation(self):
```
ArthurZucker, line 424:
```suggestion
    def experts_implementation(self):
```

src/transformers/configuration_utils.py
```diff
@@ -413,6 +416,28 @@ def _attn_implementation(self, value: str | dict | None):
...
+        return self._experts_implementation_internal
+
+    @_experts_implementation.setter
+    def _experts_implementation(self, value: str | dict | None):
```
ArthurZucker, line 428:
```suggestion
    def experts_implementation(self, value: str | dict | None):
```

src/transformers/configuration_utils.py
```diff
@@ -413,6 +416,28 @@ def _attn_implementation(self, value: str | dict | None):
                 )
                 subconfig._attn_implementation = sub_implementation
 
+    @property
+    def _experts_implementation(self):
+        return self._experts_implementation_internal
+
+    @_experts_implementation.setter
```
ArthurZucker, line 427:
```suggestion
    @experts_implementation.setter
```

src/transformers/configuration_utils.py
```diff
@@ -413,6 +416,28 @@ def _attn_implementation(self, value: str | dict | None):
...
+    def _experts_implementation(self, value: str | dict | None):
+        """We set it recursively on the sub-configs as well"""
+        # Set if for current config
+        current_moe = getattr(self, "_experts_implementation", None)
+        experts_implementation = value if not isinstance(value, dict) else value.get("", current_moe)
+        self._experts_implementation_internal = experts_implementation
```
ArthurZucker, line 433: since we want this public the above 3 comments. 
The `internal` should only just be for from pretrained / from config so logic can probably be simplified

tests/models/deepseek_v2/test_modeling_deepseek_v2.py
```diff
@@ -57,6 +57,7 @@ class DeepseekV2ModelTest(CausalLMModelTest, unittest.TestCase):
     test_all_params_have_gradient = False
     model_tester_class = DeepseekV2ModelTester
     model_split_percents = [0.5, 0.7, 0.8]
+    has_experts = True
```
ArthurZucker: we could check "experts" in "".join(config.dictg())

src/transformers/integrations/moe.py
```diff
@@ -0,0 +1,186 @@
...
+    original_forward = experts_class.forward
+
+    def __init__(self, config):
+        original_init(self, config)
+        self.config = config
+
+    def forward(self, *args, **kwargs):
+        experts_forward = original_forward
+
+        if self.config._experts_implementation != "eager":
```
ArthurZucker, line 233:
```suggestion
        if self.config.experts_implementation != "eager":
```

src/transformers/integrations/moe.py
```diff
@@ -0,0 +1,186 @@
...
+            experts_forward = ALL_EXPERTS_FUNCTIONS[self.config._experts_implementation]
+
+        return experts_forward(self, *args, **kwargs)
```
ArthurZucker, line 236: nice this means hot swap will always be applied / taken into account

ArthurZucker (approved):
Ok very nice. 
- we need a big doc page just for this + the benches on the PR and doc shows how to run them

It seems fine do de-corelate from the `@use_hf_hub _kernel` as this is EXPERT specific so yes let's go. 

Let's document what this supports (compile). 
Let's find a default for the impolementation to not be eager pls!

Let's make sure we have a big test eager matches gmm test please! 

Make sure setting - reseting the implementation properly updates the foward

tests/models/deepseek_v2/test_modeling_deepseek_v2.py
IlyasMoutawwakil: Doesn't that mean always true ? because _experts_implementation will always be in the config (even for non-moe models) I can remove _experts_implementation from PretrainedConfig and only add it to supported MoEs if that's what you mean

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# Experts Matrix Multiplication Backends
```
stevhliu:
```suggestion
# Experts backends
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+
+All Mixture-of-Experts (MoE) implementations perform the same high-level computation: for each token, a router selects $k$ experts, then the token hidden state is projected through the selected experts' parameters and aggregated with routing weights. The difference between experts backends is *how* those expert matrix multiplications / projections are executed.
```
stevhliu:
```suggestion
All Mixture-of-Experts (MoE) implementations perform the same high-level computation. For each token, a router selects *k* experts. The token hidden state is then projected through the selected experts' parameters and aggregated with routing weights. The difference between experts backends is *how* those expert matrix multiplications execute.
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Set an experts backend
+
+Use the `experts_implementation` argument in [`~PreTrainedModel.from_pretrained`] to instantiate a model with a specific experts backend.
+
+```py
+from transformers import AutoModelForCausalLM
+
+model = AutoModelForCausalLM.from_pretrained(
+
```
stevhliu:
```suggestion
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+    "Qwen/Qwen1.5-MoE-A2.7B",
+    dtype="bfloat16",
+    experts_implementation="batched_mm",
+)
+```
+
+Switch between experts backends at runtime (without reloading the model) using [`~PreTrainedModel.set_experts_implementation`].
```
stevhliu:
```suggestion
Switch between experts backends at runtime without reloading the model using [`~PreTrainedModel.set_experts_implementation`].
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+
+```py
+model.set_experts_implementation("eager")
+```
+
+### Backbone-specific experts backend
```
stevhliu:
```suggestion
## Backbone-specific experts backend
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# Experts Matrix Multiplication Backends
+
+All Mixture-of-Experts (MoE) implementations perform the same high-level computation: for each token, a router selects $k$ experts, then the token hidden state is projected through the selected experts' parameters and aggregated with routing weights. The difference between experts backends is *how* those expert matrix multiplications / projections are executed.
+
+The [`ExpertsInterface`] provides optimized experts implementations. It decouples the experts implementation from the model implementation to simplify experimentation with different functions. Add new backends easily with this consistent interface.
```
stevhliu:
```suggestion
The [`ExpertsInterface`] provides optimized experts backends. It decouples the experts implementation from the model code to simplify experimentation with different functions. Add new backends through the same interface.
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+
+Multimodal models can have multiple sub-configs (for example, different backbones). You can set a different experts backend per sub-config by passing a `dict` to `experts_implementation` at load time.
+
+Keys in the mapping must match sub-config names.
+
+```py
+from transformers import AutoModelForImageTextToText
+
+experts_implementation_per_backbone = {
+
```
stevhliu:
```suggestion
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+    "text_config": "grouped_mm",
+    "vision_config": "eager",
+}
+
+model = AutoModelForImageTextToText.from_pretrained(
+
```
stevhliu:
```suggestion
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+    "Qwen/Qwen3-VL-Moe",
+    experts_implementation=experts_implementation_per_backbone,
+)
+```
+
+Set the experts backend globally with an empty key.
+
+```py
+model = AutoModelForCausalLM.from_pretrained(
+
```
stevhliu:
```suggestion
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+    "Qwen/Qwen1.5-MoE-A2.7B",
+    experts_implementation={"": "batched_mm"},
+)
+```
+
+## `torch.compile` support
```
stevhliu:
```suggestion
## torch.compile
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+
+All three backends (`"eager"`, `"batched_mm"`, `"grouped_mm"`) support `torch.compile`.
```
stevhliu:
```suggestion
All three backends (`"eager"`, `"batched_mm"`, `"grouped_mm"`) support [torch.compile](./perf_torch_compile).
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+
+- `"eager"` and `"batched_mm"` work with all `torch.compile` modes.
+- `"grouped_mm"` compiles, but does **not** support compilation modes that rely on CUDA graphs (for example `mode="max-autotune"` or `mode="reduce-overhead"`). If you want to compile `"grouped_mm"`, use a mode that disables CUDA graphs (for example `mode="max-autotune-no-cudagraphs"`) or leave `mode=None`.
```
stevhliu:
```suggestion
- `"grouped_mm"` does **not** support compilation modes, `"max-autotune"` or `"reduce-overhead"`, that rely on CUDA graphs. To compile `"grouped_mm"`, use `"max-autotune-no-cudagraphs"` to disable CUDA graphs or leave `mode=None`.
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+
+```py
+import torch
+from transformers import AutoModelForCausalLM
+
+model = AutoModelForCausalLM.from_pretrained(
+
```
stevhliu:
```suggestion
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+    "Qwen/Qwen1.5-MoE-A2.7B",
+    dtype="bfloat16",
+    experts_implementation="grouped_mm",
+).eval().cuda()
+
+# Works for grouped_mm (no CUDA graphs)
+model.forward = torch.compile(model.forward, mode="max-autotune-no-cudagraphs")
+```
+
+For the best performance we ran a [benchmark](https://github.com/user-attachments/files/24125816/bench.py) of different inputs sizes, with different experts implementations, with and without `torch.compile`. The results are summarized in the table below.
```
stevhliu: I think it'd be cool and look better to present the benchmark in a table like this [one](https://huggingface.co/docs/transformers/perf_torch_compile#benchmarks) :)

```suggestion
## Benchmarks

This [benchmark](https://github.com/user-attachments/files/24125816/bench.py) compares different input sizes and experts implementations with and without `torch.compile`.
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+
+The [`ExpertsInterface`] provides optimized experts implementations. It decouples the experts implementation from the model implementation to simplify experimentation with different functions. Add new backends easily with this consistent interface.
+
+| experts backend | description                                                                                       |
+| --------------- | ------------------------------------------------------------------------------------------------- |
+| `"eager"`       | Reference implementation that loops over active experts and applies projections per-expert.       |
+| `"batched_mm"`  | Uses `torch.bmm` to compute per-(token, expert) projections in a batched way.                     |
+| `"grouped_mm"`  | Uses `torch._grouped_mm` to group tokens by expert and run grouped GEMMs (requires PyTorch 2.9+). |
+
+## Set an experts backend
```
stevhliu:
```suggestion
| `"grouped_mm"`  | Uses `torch._grouped_mm` to group tokens by expert and run grouped GEMMs (requires PyTorch 2.9+). |

`batched_mm` is fastest for very small inputs and compilation speeds it up further. `grouped_mm` performs best for larger inputs.

## Set an experts backend
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+| 4          | 128        | False                      | batched_mm     | 1235.25           | 1235.33             | 1237.90          | 61465.85      |
+| 4          | 128        | True                       | batched_mm     | 1505.00           | 1503.31             | 1536.10          | 50174.26      |
+| 4          | 128        | max-autotune-no-cudagraphs | batched_mm     | 572.37            | 570.81              | 589.74           | 27762.45      |
+| 4          | 128        | False                      | grouped_mm     | 80.95             | 81.06               | 81.70            | 27762.45      |
+| 4          | 128        | True                       | grouped_mm     | 79.67             | 79.69               | 80.54            | 27762.45      |
+| 4          | 128        | max-autotune-no-cudagraphs | grouped_mm     | 83.29             | 79.83               | 111.83           | 27762.46      |
+
+TLDR; For very small inputs batched_mm can be extremely fast and even faster with compilation, otherwise grouped_mm is unbeatable for larger inputs.
```
stevhliu:
```suggestion
```

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,144 @@
...
+# Experts Matrix Multiplication Backends
+
+All Mixture-of-Experts (MoE) implementations perform the same high-level computation: for each token, a router selects $k$ experts, then the token hidden state is projected through the selected experts' parameters and aggregated with routing weights. The difference between experts backends is *how* those expert matrix multiplications / projections are executed.
+
+The [`ExpertsInterface`] provides optimized experts implementations. It decouples the experts implementation from the model implementation to simplify experimentation with different functions. Add new backends easily with this consistent interface.
+
+| experts backend | description                                                                                       |
+| --------------- | ------------------------------------------------------------------------------------------------- |
+| `"eager"`       | Reference implementation that loops over active experts and applies projections per-expert.       |
+| `"batched_mm"`  | Uses `torch.bmm` to compute per-(token, expert) projections in a batched way.                     |
```
stevhliu:
```suggestion
| `"batched_mm"`  | Uses [torch.bmm](https://docs.pytorch.org/docs/stable/generated/torch.bmm.html) to compute per-(token, expert) projections in a batched way.                     |
```

stevhliu (approved):
very nice thanks!

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,142 @@
...
+
+The [`ExpertsInterface`] provides optimized experts backends. It decouples the experts implementation from the model code to simplify experimentation with different functions. Add new backends through the same interface.
+
+| experts backend | description                                                                                       |
+| --------------- | ------------------------------------------------------------------------------------------------- |
+| `"eager"`       | Reference implementation that loops over active experts and applies projections per-expert.       |
+| `"batched_mm"`  | Uses [torch.bmm](https://docs.pytorch.org/docs/stable/generated/torch.bmm.html) to compute per-(token, expert) projections in a batched way.                     |
+| `"grouped_mm"`  | Uses `torch._grouped_mm` to group tokens by expert and run grouped GEMMs (requires PyTorch 2.9+). |
+
+`batched_mm` is fastest for very small inputs and compilation speeds it up further. `grouped_mm` performs best for larger inputs.
```
ArthurZucker, line 28: larger inputs == longer context / more tokens?

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,142 @@
...
+Keys in the mapping must match sub-config names.
+
+```py
+from transformers import AutoModelForImageTextToText
+
+experts_implementation_per_backbone = {
+    "text_config": "grouped_mm",
+    "vision_config": "eager",
+}
+
```
ArthurZucker, line 63: very nice

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+## Benchmarks
+
+This [benchmark](https://github.com/user-attachments/files/24125816/bench.py) compares different input sizes and experts implementations with and without `torch.compile`.
+
+| Batch Size | Seq Length | Torch Compile              | Implementation | Mean Latency (ms) | Median Latency (ms) | P90 Latency (ms) | Peak Mem (MB) |
+| ---------- | ---------- | -------------------------- | -------------- | ----------------- | ------------------- | ---------------- | ------------- |
+| 1          | 16         | False                      | eager          | 271.80            | 272.94              | 295.34           | 27324.65      |
+| 1          | 16         | True                       | eager          | 351.86            | 351.64              | 384.64           | 27329.29      |
```
ArthurZucker: might be hard to read, colors / bold could help for important numbers! (no1 etc)

src/transformers/models/jetmoe/modeling_jetmoe.py
```diff
@@ -583,6 +583,7 @@ class JetMoePreTrainedModel(PreTrainedModel):
         "hidden_states": JetMoeDecoderLayer,
         "attentions": OutputRecorder(JetMoeAttention, index=1),
     }
+    _supports_grouped_mm = True
```
ArthurZucker: I don't think we need this, we can just check if decorator is used in the file / imported

tests/test_modeling_common.py
```diff
@@ -509,6 +509,92 @@ def _can_output_attn(model):
...
+
+            model.set_experts_implementation("batched_mm")
+            self.assertEqual(model.config._experts_implementation, "batched_mm")
+            outputs_batched_mm = model(**prepared_inputs)
+            outputs_batched_mm = _get_output_tensors(outputs_batched_mm)
+
+            model.set_experts_implementation("grouped_mm")
+            self.assertEqual(model.config._experts_implementation, "grouped_mm")
+            outputs_grouped_mm = model(**prepared_inputs)
+            outputs_grouped_mm = _get_output_tensors(outputs_grouped_mm)
```
ArthurZucker: nice!

tests/test_modeling_common.py
```diff
@@ -509,6 +509,92 @@ def _can_output_attn(model):
...
+            outputs_eager = _get_output_tensors(outputs_eager)
+
+            model.set_experts_implementation("batched_mm")
+            self.assertEqual(model.config._experts_implementation, "batched_mm")
+            outputs_batched_mm = model(**prepared_inputs)
+            outputs_batched_mm = _get_output_tensors(outputs_batched_mm)
+
+            model.set_experts_implementation("grouped_mm")
+            self.assertEqual(model.config._experts_implementation, "grouped_mm")
+            outputs_grouped_mm = model(**prepared_inputs)
```
ArthurZucker: Imo we need to check here as well that the callable is different because tthe setter could work, but not do anything and if this is the only test, it will not fai

ArthurZucker (approved):

docs/source/en/experts_interface.md
IlyasMoutawwakil: yes larger inputs = larger batch_size * sequence_length, because grouped_mm dominate (1, 128), (4,16), (4, 128)

docs/source/en/experts_interface.md
IlyasMoutawwakil: great idea ! will add it !

src/transformers/models/jetmoe/modeling_jetmoe.py
IlyasMoutawwakil: sometimes the decorator is not there, for example with deepseek_vl and other multimodal models that might reuse the text-only model

IlyasMoutawwakil: okay I just found out deepseek_vl creates a llama so yes for now simply checking the code is enough to know whether model supports setting experts imp !

tests/test_modeling_common.py
IlyasMoutawwakil: added

src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
```diff
@@ -453,7 +454,7 @@ class DeepseekV2PreTrainedModel(PreTrainedModel):
     _supports_flash_attn = True
     _supports_sdpa = True
     _supports_flex_attn = True
-    _can_compile_fullgraph = False
+    _can_compile_fullgraph = False  # MoE models don't work with torch.compile (`torch.where(condition)` not supported)
```
IlyasMoutawwakil: @ArthurZucker I added the same comment as the one I saw in most MoEs, but it shouldn't be the case anymore, should I set these to True now ? or keep them False for a future PR

src/transformers/integrations/moe.py
```diff
@@ -0,0 +1,249 @@
...
+            experts_forward = ALL_EXPERTS_FUNCTIONS[self.config._experts_implementation]
+
+        if (
+            (self.gate_up_proj.dtype != torch.bfloat16 or self.down_proj.dtype != torch.bfloat16)
+            and self.config._experts_implementation == "grouped_mm"
+            and is_torchdynamo_compiling()
+        ):
+            # grouped_mm currently only supports bfloat16 when being compiled with torchdynamo
+            # we dispatch to batched_mm as it behaves better than eager in this case
+            experts_forward = batched_mm_experts_forward
```
IlyasMoutawwakil: @ArthurZucker wdyt

src/transformers/integrations/moe.py
```diff
@@ -0,0 +1,249 @@
...
+) -> torch.Tensor:
+    device = hidden_states.device
+    num_top_k = top_k_index.size(-1)
+    num_tokens = hidden_states.size(0)
+    num_experts = self.gate_up_proj.size(0)
+    final_hidden_states = torch.zeros_like(hidden_states)
+
+    # Flatten top_k_index to get expert_ids per selected sample
+    expert_ids = top_k_index.reshape(-1)
+    token_idx = torch.arange(num_tokens, device=device).unsqueeze(1).expand(-1, num_top_k).reshape(-1)
```
ArthurZucker, line 79:
```suggestion
    token_idx = (final_hidden_states+1).cumsum().unsqueeze(1).expand(-1, num_top_k).reshape(-1)
```
I think is more cudagraph compat?

src/transformers/integrations/moe.py
ArthurZucker: sounds good to me but needs to be documented a bit more!!

src/transformers/modeling_utils.py
```diff
@@ -1932,6 +1992,17 @@ def _can_set_attn_implementation(cls) -> bool:
...
             return True
 
+    @classmethod
+    def _can_set_experts_implementation(cls) -> bool:
+        """Detect whether the class supports setting its experts implementation dynamically. It is an ugly check based on
+        opening the file, but avoids maintaining yet another property flag.
+        """
+        class_file = sys.modules[cls.__module__].__file__
+        with open(class_file, "r") as f:
+            code = f.read()
```
ArthurZucker, line 1959: actually sorry, IDK how fast this is having the explicit arg might be better!

src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
ArthurZucker: yeah. it might depend on the impl no?

ArthurZucker (approved):
🚀

src/transformers/modeling_utils.py
IlyasMoutawwakil: it's rare for it to be called so I guess it's fine, the counter argument for the explicit arg is that we might forget to add it and the model will never be tested with grouped_mm, with this one we have a single source of truth.

src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
IlyasMoutawwakil: yes but since we default to grouped_mm, can we set this flag to is_grouped_mm_available() ?

IlyasMoutawwakil: when switched to True locally, all tests (including compilation+fullgraph) passed

src/transformers/integrations/moe.py
IlyasMoutawwakil: done in https://github.com/huggingface/transformers/pull/42697/commits/a4447fb380bc27e58b00c722efa1e246b05bfada

IlyasMoutawwakil:
I updated the branch and the latest ernie_4_5_vl_moe

docs/source/en/experts_interface.md
```diff
@@ -0,0 +1,168 @@
...
+| -------------- | ------------------------------------ | -------------------------------- | ---------------- |
+| `grouped_mm`   | `None`, `max-autotune-no-cudagraphs` | `bfloat16`                       | Yes              |
+| `batched_mm`   | all                                  | `bfloat16`, `float16`, `float32` | Yes              |
+| `eager`        | all                                  | `bfloat16`, `float16`, `float32` | No               |
+
+Notes:
+
+- The `grouped_mm` experts backend currently only supports `bfloat16` when compiled with `torch.compile`. Additionally, it is not compatible with CUDA graphs, so you must use `mode=None` or `mode="max-autotune-no-cudagraphs"` when compiling.
+- The `eager` experts backend uses a data-dependent operation to find which experts are used in a forward pass. This operation is not compatible with full graph compilation (`fullgraph=True`).
+- When using `float16` or `float32` with `grouped_mm`, the model will automatically fall back to `batched_mm` when compiled.
```
IlyasMoutawwakil: @stevhliu

src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
ArthurZucker: yes sure!

src/transformers/cache_utils.py
```diff
@@ -281,16 +281,18 @@ def lazy_initialization(self, key_states: torch.Tensor):
         i.e. `mode="reduce-overhead"` is known to fail). But it will in general work correctly, and prefill should
         not be compiled anyway for performances!
         """
-        self.max_batch_size, self.num_heads, _, self.head_dim = key_states.shape
         self.dtype, self.device = key_states.dtype, key_states.device
+        self.max_batch_size, self.num_heads = key_states.shape[:2]
+        self.v_head_dim = value_states.shape[-1]
+        self.k_head_dim = key_states.shape[-1]
```
IlyasMoutawwakil, line 287: had to add value_states to the signature since some models like deepseek have different k/v head_dim

src/transformers/generation/utils.py
```diff
@@ -2178,8 +2178,10 @@ def _valid_auto_compile_criteria(self, model_kwargs: dict[str, Any], generation_
                 "will be skipped."
             )
 
-            # Finally: if we can compile, disable tokenizers parallelism and check for FA2 + static cache
+        if can_compile:
```
IlyasMoutawwakil, line 2181: seemed to me like this if was missing from the logic

src/transformers/generation/utils.py
```diff
@@ -2190,6 +2192,15 @@ def _valid_auto_compile_criteria(self, model_kwargs: dict[str, Any], generation_
...
                     generation_config.compile_config.fullgraph = False
 
+            # If we use grouped_mm and dtype different than bfloat16, we fallback to batched_mm
+            if self.config._experts_implementation == "grouped_mm":
+                if self.dtype != torch.bfloat16:
+                    logger.warning_once(
+                        "torch._grouped_mm currently only supports bfloat16 when being compiled with torch.compile. "
+                        "Falling back to batched_mm implementation for compilation."
+                    )
+                    self.set_experts_implementation("batched_mm")
```
IlyasMoutawwakil, line 2202: preferred falling back to batched_mm here only when we are optimizing the model for compiled generation.
falling back during the forward pass seemed like very implicit behavior that might go against user intention.

ArthurZucker:
Kudos!

PR #42700, opened 2025-12-08
[kernels] make the module declaration implicit with decorator

MekkCyber:
# What does this PR do?

instead of setting the module explicitely and not using it, this pr simply uses a decorator to set the attribute which is cleaner

### DIFF
--- a/src/transformers/integrations/hub_kernels.py
+++ b/src/transformers/integrations/hub_kernels.py
@@ -372,7 +372,7 @@ def lazy_load_kernel(kernel_name: str, mapping: dict[str, ModuleType | None] = _
 def use_kernelized_func(module_names: list[Callable] | Callable):
     """
     This decorator attaches the target function as an attribute of the module.
-    The function must already be decorated with @use_kernel_func_from_hub; 
+    The function must already be decorated with @use_kernel_func_from_hub
     this decorator then wraps it as an nn.Module internally.
     When kernelize is later applied to the full model, the function can be accessed as a regular module attribute and kernelized just like any other layer.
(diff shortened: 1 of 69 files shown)


Cyrilvallez (commented):
Very nice, we keep the decorator-only principle of `kernels`!
As discussed offline, let's call the decorator `use_kernelized_func` or something similar though, to make it more explicit that it's kernels! And no need to provide "rotary_fn" explicitement, only the func name!

src/transformers/integrations/hub_kernels.py
```diff
@@ -369,6 +369,22 @@ def lazy_load_kernel(kernel_name: str, mapping: dict[str, ModuleType | None] = _
     return mapping[kernel_name]
 
 
+def use_kernelized_func(module_names: list[Callable]):
+    def decorator(cls):
```
ArthurZucker: let's document this black magic please 🤣

src/transformers/models/afmoe/modeling_afmoe.py
```diff
@@ -338,6 +338,7 @@ def eager_attention_forward(
     return attn_output, attn_weights
 
 
+@use_kernelized_func([apply_rotary_pos_emb])
```
ArthurZucker:
```suggestion
@use_kernelized_func(apply_rotary_pos_emb)
```
let's support non list and cast to list just QOL

ArthurZucker (approved):
Much better ty!

src/transformers/integrations/hub_kernels.py
MekkCyber: Done

PR #42701, opened 2025-12-08
Fix PEFT integration with new weight loader

Cyrilvallez:
# What does this PR do?

As per the title. The `VLMS` list should not be duplicated in `peft.py` at it's not getting updated! Also fix it when it's not coming from `from_pretrained`!


BenjaminBossan (approved):
Nice deduplication, makes sense.

PR #42702, opened 2025-12-08
🚨 Generation config defaults are now `None`

zucchini-nlp:
# What does this PR do?

As per title, must have been done long time ago but could't break BC. The current impl breaks BC only half-way, i.e. the generation loop is not affected and will keep using the old defaults. The biggest difference is for users to init, access, modify, etc. the model's generation config directly:

```python
# `0` before this PR, `None` after the PR
print(model.generation_config.no_repeat_ngram_size)
```


src/transformers/generation/candidate_generator.py
```diff
@@ -117,11 +117,16 @@ def __init__(
...
         self.assistant_model = assistant_model
-        self.num_assistant_tokens = assistant_model.generation_config.num_assistant_tokens
-        self.assistant_confidence_threshold = assistant_model.generation_config.assistant_confidence_threshold
+
+        # Prepare the generation config by updating with default values if not already set by users
+        self.assistant_generation_config = copy.deepcopy(assistant_model.generation_config)
+        global_defaults = self.assistant_generation_config._get_default_generation_params()
+        _ = self.assistant_generation_config.update(**global_defaults, defaults_only=True)
+        self.num_assistant_tokens = self.assistant_generation_config.num_assistant_tokens
+        self.assistant_confidence_threshold = self.assistant_generation_config.assistant_confidence_threshold
```
zucchini-nlp, line 126: since the config is no longer initialized with default values, we need to update with defaults manually. Otherwise it can be too breaking

src/transformers/generation/configuration_utils.py
```diff
@@ -38,6 +37,7 @@
 
 
 if TYPE_CHECKING:
+    from ..configuration_utils import PreTrainedConfig
```
zucchini-nlp, line 40: solves circular import

src/transformers/generation/configuration_utils.py
```diff
@@ -337,38 +337,36 @@ class GenerationConfig(PushToHubMixin):
...
-        self.min_length = kwargs.pop("min_length", 0)
+        self.min_length = kwargs.pop("min_length", None)
         self.min_new_tokens = kwargs.pop("min_new_tokens", None)
-        self.early_stopping = kwargs.pop("early_stopping", False)
+        self.early_stopping = kwargs.pop("early_stopping", None)
         self.max_time = kwargs.pop("max_time", None)
         self.stop_strings = kwargs.pop("stop_strings", None)
 
         # Parameters that control the generation strategy used
         self.do_sample = kwargs.pop("do_sample", False)
```
zucchini-nlp, line 349: a few boolean values are kept as is. Unlike `temperature/topK/etc`, these attributes cannot be set to `None`. It is either True or False, and a third option makes little sense

src/transformers/generation/watermarking.py
```diff
@@ -16,20 +16,23 @@
...
+from ..configuration_utils import PreTrainedConfig
 from ..modeling_utils import PreTrainedModel
 from ..utils import ModelOutput, logging
-from .configuration_utils import PreTrainedConfig, WatermarkingConfig
 from .logits_process import SynthIDTextWatermarkLogitsProcessor, WatermarkLogitsProcessor
 
 
+if TYPE_CHECKING:
+    from .configuration_utils import WatermarkingConfig
+
```
zucchini-nlp, line 35: another circular import between PreTrainedConfig and GenerationConfig

src/transformers/configuration_utils.py
```diff
@@ -1063,58 +1062,17 @@ def register_for_auto_class(cls, auto_class="AutoConfig"):
...
-                    continue
-                non_default_generation_parameters[parameter_name] = getattr(self_decoder_config, parameter_name)
-
-        return non_default_generation_parameters
+        generation_params = {}
+        default_config = self.__class__().to_dict() if not self.has_no_defaults_at_init else {}
+        for key in GenerationConfig._get_default_generation_params().keys():
+            if hasattr(self, key) and getattr(self, key) is not None and key not in default_config:
+                generation_params[key] = getattr(self, key)
+
```
zucchini-nlp, line 1074: could have been simplified because we no longer have any generation params in `model.config`

tests/models/reformer/test_modeling_reformer.py
```diff
@@ -408,11 +408,11 @@ def create_and_check_reformer_model_fp16_forward(self, config, input_ids, input_
     def create_and_check_reformer_model_generate(self, config, input_ids, input_mask, choice_labels):
         config.is_decoder = True
         config.lsh_num_chunks_after = 0
-        config.bos_token_id = 0
-        config.eos_token_id = None
-        config.max_length = 20
```
zucchini-nlp, line 413: generation params should not be assigned in model's config!

zucchini-nlp:
Ready for review!

src/transformers/configuration_utils.py
```diff
@@ -449,13 +450,11 @@ def save_pretrained(self, save_directory: str | os.PathLike, push_to_hub: bool =
...
-        if len(non_default_generation_parameters) > 0:
+        generation_parameters = self._get_generation_parameters()
+        if len(generation_parameters) > 0:
             raise ValueError(
-                "Some non-default generation parameters are set in the model config. These should go into either a) "
-                "`model.generation_config` (as opposed to `model.config`); OR b) a GenerationConfig file "
-                "(https://huggingface.co/docs/transformers/generation_strategies#save-a-custom-decoding-strategy-with-your-model)."
-                f"\nNon-default generation parameters: {str(non_default_generation_parameters)}",
+                "Some generation parameters are set in the model config. These should go into `model.generation_config`"
+                f"as opposed to `model.config. \nGeneration parameters found: {str(generation_parameters)}",
```
vasqu:
```suggestion
                f"as opposed to `model.config`. \nGeneration parameters found: {str(generation_parameters)}",
```
Super nit

src/transformers/configuration_utils.py
vasqu: Just for my understanding, generation config is top level either way. Hence, we no longer need to discern for submodels etc in composite models (which could have possibly different config values here)

src/transformers/generation/candidate_generator.py
```diff
@@ -117,11 +117,16 @@ def __init__(
 
         # Prepare the assistant and the starting number of candidate tokens
         self.assistant_model = assistant_model
-        self.num_assistant_tokens = assistant_model.generation_config.num_assistant_tokens
-        self.assistant_confidence_threshold = assistant_model.generation_config.assistant_confidence_threshold
+
+        # Prepare the generation config by updating with default values if not already set by users
+        self.assistant_generation_config = copy.deepcopy(assistant_model.generation_config)
+        global_defaults = self.assistant_generation_config._get_default_generation_params()
+        _ = self.assistant_generation_config.update(**global_defaults, defaults_only=True)
```
```suggestion
        self.assistant_generation_config.update(**global_defaults, defaults_only=True)
```
Nit, just a personal preference

src/transformers/generation/configuration_utils.py
vasqu: If it's optional bool, we can still have `None` - would be equal to default `False` then. But I see the reason semantically, it should not be optional and we prefer explicit boolean values. 

Should we update the typing then? I think we mark it as optional for these

src/transformers/generation/configuration_utils.py
```diff
@@ -381,20 +379,12 @@ def __init__(self, **kwargs):
         self.token_healing = kwargs.pop("token_healing", False)
         self.guidance_scale = kwargs.pop("guidance_scale", None)
 
-        watermarking_config = kwargs.pop("watermarking_config", None)
-        if watermarking_config is None:
-            self.watermarking_config = None
-        elif isinstance(watermarking_config, BaseWatermarkingConfig):
-            self.watermarking_config = watermarking_config
-        else:
-            self.watermarking_config = WatermarkingConfig.from_dict(watermarking_config)
```
vasqu, line 390: That was a bit confusing, but seeing it's handled later? Just super big nit, but it would be nice if we could keep somewhat the same order as the one of the docstring

src/transformers/generation/configuration_utils.py
```diff
@@ -488,8 +482,8 @@ def get_generation_mode(self, assistant_model: Optional["PreTrainedModel"] = Non
         # property and part of the `__repr__`
         if self.constraints is not None or self.force_words_ids is not None:
             generation_mode = GenerationMode.CONSTRAINED_BEAM_SEARCH
-        elif self.num_beams == 1:
-            if self.do_sample is False:
+        elif self.num_beams is None or self.num_beams == 1:
+            if self.do_sample is not True:
```
```suggestion
            if not self.do_sample:
```
Any reason we do explicit `is`

src/transformers/generation/configuration_utils.py
```diff
@@ -537,6 +531,45 @@ def get_generation_mode(self, assistant_model: Optional["PreTrainedModel"] = Non
                 )
         return generation_mode
 
+    @staticmethod
+    def _get_default_generation_params() -> dict[str, Any]:
```
vasqu, line 535: Much more reasonable location of default values :pray:

src/transformers/generation/configuration_utils.py
```diff
@@ -381,20 +379,12 @@ def __init__(self, **kwargs):
...
-        watermarking_config = kwargs.pop("watermarking_config", None)
-        if watermarking_config is None:
-            self.watermarking_config = None
-        elif isinstance(watermarking_config, BaseWatermarkingConfig):
-            self.watermarking_config = watermarking_config
-        else:
-            self.watermarking_config = WatermarkingConfig.from_dict(watermarking_config)
-
         # Parameters that define the output variables of `generate`
         self.num_return_sequences = kwargs.pop("num_return_sequences", 1)
```
vasqu, line 387: Not optional then?

src/transformers/generation/configuration_utils.py
```diff
@@ -1101,7 +1133,7 @@ def to_json_file(
             writer.write(self.to_json_string(use_diff=use_diff, keys_to_pop=keys_to_pop))
 
     @classmethod
-    def from_model_config(cls, model_config: PreTrainedConfig | dict) -> "GenerationConfig":
+    def from_model_config(cls, model_config: Union["PreTrainedConfig", dict]) -> "GenerationConfig":
```
vasqu, line 1136:
```suggestion
    def from_model_config(cls, model_config: "PreTrainedConfig" | dict) -> "GenerationConfig":
```
Does this not work?

src/transformers/generation/utils.py
```diff
@@ -1760,101 +1758,50 @@ def _prepare_generated_length(
...
         **kwargs: Any,
     ) -> tuple[GenerationConfig, dict]:
         """
         Prepares the base generation config, then applies any generation configuration options from kwargs. This
         function handles retrocompatibility with respect to configuration files.
         """
         # parameterization priority:
-        # kwargs > non-global default values in `generation_config` > `model.generation_config` > GenerationConfig()
+        # user-defined kwargs or `generation_config` > `self.generation_config` > global default values
+        # NOTE: doesn't make sense to allow kwargs and `generation_config`. Might be strict and make them mutually exclusive?
```
vasqu: Just so that I'm in the loop it's essentially
- kwargs as in `model.generate(..., do_sample=False)`
- generation config as in `model.generate(..., generation_config=GenerationConfig(do_sample=False))`
- self.generation_config as in the associated json saved along
- defaults from the static method

In this case, yes we should be strict here IMO. It does not make sense to have something along `model.generate(..., do_sample=False, generation_config=GenerationConfig(do_sample=True))` --> ambiguity is bad here and would rely on the user to know what takes precedence

src/transformers/generation/utils.py
```diff
@@ -1760,101 +1758,50 @@ def _prepare_generated_length(
...
-            if generation_config.cache_implementation == "hybrid":
-                generation_config.cache_implementation = None
+            # Users may modify `model.config` to control generation. This is a legacy behavior and is not supported anymore
+            if len(self.config._get_generation_parameters()) > 0:
+                raise ValueError(
+                    "You have modified the pretrained model configuration to control generation "
+                    f" We detected the following values set - {self.config._get_generation_parameters()}. "
+                    " This strategy to control generation is not supported anymore. Please use and modify `model.generation_config` "
+                    " (see https://huggingface.co/docs/transformers/generation_strategies#default-text-generation-configuration )",
+                )
```
vasqu, line 1780:
```suggestion
            if len(config_generation_parameters := self.config._get_generation_parameters()) > 0:
                raise ValueError(
                    "You have modified the pretrained model configuration to control generation. "
                    f"We detected the following values set - {config_generation_parameters}. "
                    "This strategy to control generation is not supported anymore. Please use and modify `model.generation_config` "
                    "(see https://huggingface.co/docs/transformers/generation_strategies#default-text-generation-configuration )",
                )
```
Don't mind the `:=` too much (personal preference) but the whitespaces are doubled if I see it correctly.

src/transformers/generation/utils.py
```diff
@@ -1760,101 +1758,50 @@ def _prepare_generated_length(
...
-                if generation_config.decoder_start_token_id is None:
-                    generation_config.decoder_start_token_id = self.generation_config.decoder_start_token_id
-
-        # Finally, apply any passed kwargs
+        # First set values from the loaded `self.generation_config`, then set default values (BC)
+        # Do not update any values that aren't `None`, i.e. if set by users explicitly and passed
+        # to `generate()`. Thus the `defaults_only=True` is used
+        global_defaults = self.generation_config._get_default_generation_params()
+        _ = generation_config.update(**self.generation_config.to_dict(), defaults_only=True)
+        _ = generation_config.update(**global_defaults, defaults_only=True)
```
vasqu:
```suggestion
        generation_config.update(**self.generation_config.to_dict(), defaults_only=True)
        generation_config.update(**global_defaults, defaults_only=True)
```

src/transformers/generation/utils.py
```diff
@@ -2495,9 +2436,15 @@ def generate(
             streamer,
         )
 
-        generation_config, model_kwargs = self._prepare_generation_config(
-            generation_config, use_model_defaults, **kwargs
+        # Check length values before updating the config with defaults. We'll use it later in (# 6)
```
vasqu:
```suggestion
        # Check length values before updating the config with defaults. We'll use it later to define the final min/max length (# 6)
```
Let's keep the comment a bit more explicit, helps to follow along more easily

src/transformers/generation/watermarking.py
```diff
@@ -120,13 +123,13 @@ class WatermarkDetector:
 
     def __init__(
         self,
-        model_config: PreTrainedConfig,
+        model_config: "PreTrainedConfig",
         device: str,
-        watermarking_config: WatermarkingConfig | dict,
+        watermarking_config: Optional["WatermarkingConfig"],
```
vasqu:
```suggestion
        watermarking_config: "WatermarkingConfig" | dict | None,
```
Can be a dict, no? `if not isinstance(watermarking_config, dict)` also kinda implies this imo

src/transformers/models/whisper/generation_whisper.py
vasqu, line 1: Very important model, let's double check that the slow tests pass (the same as on main)

tests/models/vision_encoder_decoder/test_modeling_vision_encoder_decoder.py
```diff
@@ -1105,6 +1108,7 @@ def check_encoder_decoder_model_generate(self, config, decoder_config, pixel_val
         if hasattr(enc_dec_model.generation_config, "eos_token_id"):
             enc_dec_model.generation_config.eos_token_id = None
         enc_dec_model.to(torch_device)
+        enc_dec_model.generation_config.max_length = 20
```
vasqu, line 1111: `generation_config.max_length = 20` appears fairly often, what's the default value or more like could we avoid this, i.e. avoid setting this in all kind of tests.

tests/utils/test_cache_utils.py
```diff
@@ -467,10 +469,13 @@ def test_data_parallel_dynamic_cache(self):
 
         # Check that the caches are the same
         for layer_idx in range(len(no_parallelism_cache)):
-            for kv_idx in range(2):  # 0 = key, 1 = value
-                torch.testing.assert_close(
-                    actual=parallelism_cache[layer_idx][kv_idx], expected=no_parallelism_cache[layer_idx][kv_idx]
-                )
```
vasqu, line 473: That was BC behavior?

vasqu (commented):
Let's still add an :rotating_light: even if it's not completely breaking, I rather be safe than sorry here. We never know 

First round of comments, my biggest issue would be the kwargs vs generation config passing. But you also left a note there.

src/transformers/generation/configuration_utils.py
zucchini-nlp: yep, I think updating the typing is better than allowing `None`. I doubt that users were setting `None` anyway

src/transformers/configuration_utils.py
zucchini-nlp: yeah, that and also because these lines are more of a workaround for old models (e.g. bart). New models don't have any generation params in model config anyway, we don't allow it for quite a long time

src/transformers/generation/configuration_utils.py
zucchini-nlp: nope, when the PreTrainedConfig is imported only for type checking, ruff will complain about "str and a type cannot be used together"

src/transformers/generation/watermarking.py
zucchini-nlp: right, will update without `|`

src/transformers/generation/configuration_utils.py
zucchini-nlp: ig it was because `do_sample` could technically be `None`. Personally can't think of any situation except when user sets it explicitly, left as is just in case we have `do_sample=None` configs on the hub.

I think we can break it and nudge users to not set `None`, will update the checks and warning messages

src/transformers/generation/configuration_utils.py
zucchini-nlp: optional, my bad. There should be a `self.watermarking_config = kwargs.pop("watermarking_config", None)` which got lost

tests/utils/test_cache_utils.py
zucchini-nlp: the test was broken on main and never fixed 💀  CI complained about it so I fixed this one and skipped offloaded cache tests with a FIXME

tests/models/vision_encoder_decoder/test_modeling_vision_encoder_decoder.py
zucchini-nlp: it is because we re-use below and pass it directly to generation (`max_length=enc_dec_model.generation_config.max_length,`). If not set explicitly, then we're passing `None` which means we're relying on default length. The default is 20 max new tokens if not passed by users, which is not the same

src/transformers/generation/utils.py
zucchini-nlp: Yep, correct! Also agree with the second point which I prefer to do in another PR. I noticed that it's a common pattern for tests and most probably users also do it

We would need a proper deprecation cycle instead of breaking it

zucchini-nlp:
run-slow: bart, csm, dia, encoder_decoder, musicgen, rag, reformer, speech_encoder_decoder, vision_encoder_decoder, whisper

zucchini-nlp:
@vasqu requesting another review :)

src/transformers/configuration_utils.py
vasqu: Gotcha, makes sense to me

src/transformers/generation/configuration_utils.py
vasqu: Looks like it was missed here?

src/transformers/generation/utils.py
vasqu: Fair enough but maybe we can break for v5? Not super important but it gives us a good opportunity to do so. 

Either way, let's upgrade this to a TODO (as well).

tests/generation/test_configuration_utils.py
```diff
@@ -252,8 +252,9 @@ def test_refuse_to_save(self):
             with self.assertRaises(ValueError) as exc:
                 config.save_pretrained(tmp_dir)
             self.assertTrue("Fix these issues to save the configuration." in str(exc.exception))
+            print(exc.exception)
```
vasqu:
```suggestion
```
:p

tests/utils/test_cache_utils.py
vasqu: oof, makes sense

vasqu (approved):
LGTM, just a few last nits

zucchini-nlp:
Will merge when CI is fixed on main

albertvillanova:
Hi,

First of all, thanks for your reactivity and addressing the underlying issue:
- #42762

On the other hand, I did a quick pass over the generation code, and I think there is a subtle semantic point worth double-checking: None is not always just "unset", it can also mean "disable this behavior".

Concretely, `top_k=None` disables top-k filtering entirely and allows all tokens. Because of that, we now have two different concepts that can both be represented as "None":
- Use the global/default value for top_k (e.g. 50)
- Explicitly disable top-k filtering (top_k=None or 0) even if the model's generation config or the global defaults say otherwise

Therefore, a situation can occur during training where:
- the user-provided `generation_config` has `top_k=None` (intending to disable top-k filtering),
- the model’s own `generation_config` has a non-None value for `top_k`,
- merging logic currently preserves the model’s value instead of respecting the explicit `None`.

In that scenario, `None` is not just "unset"; it is a meaningful instruction ("don't apply top-k filtering"). If so, the merge semantics may need refinement to avoid unintentionally re-enabling filtering (e.g. by using a sentinel value instead).

I'll continue digging into this, but flagging it early so we can discuss it before merging.

zucchini-nlp:
@albertvillanova I think if `top_k=None` in the generation config it is the same as if users did not pass any `top_k` in kwargs. A value is not set to any value (`None`) does not specifically mean that the users is requesting to not use it, so the users would need to explicitly unset as `top_k=0` if model has saved a different value.

Unfortunately we have no way to 100% know what users wants when they set values to `None` in current code. The only way would be for us to not update generation config with model's defaults if users pass `my_generation_config`. But that will be much more breaking and will require users to always create a custom config from `model.generation_config`

For ex, if everyone prepared custom configs as below, we can fix your issue. I'm afraid it's not the case for most users
```python
my_generation_config = model.generation_config
my_generation_config.top_k = None
model.generate(inputs, generation_config=my_generation_config)

```

vasqu:
I think we can merge? @zucchini-nlp

src/transformers/generation/utils.py
```diff
@@ -1760,101 +1758,50 @@ def _prepare_generated_length(
...
-            # - `use_model_defaults`: let's fallback ALL default values to the model's generation config
-            # - otherwise: legacy behavior, let's just make sure we have the tokens defined
-            model_base_version = version.parse(version.parse(self.generation_config.transformers_version).base_version)
-            if use_model_defaults is True or (
-                use_model_defaults is None and model_base_version >= version.parse("4.50.0")
-            ):
-                modified_values = {}
-                global_default_generation_config = GenerationConfig()
-                model_generation_config = self.generation_config
-                # we iterate over the model's generation config: it may hold custom keys, which we'll want to copy
```
ebezzam, line 1821: @vasqu, @zucchini-nlp are custom keys no longer copied over? 

Seems like only the ones [here](https://github.com/huggingface/transformers/blob/d14d99eadf41fd6702522eaea2f17b346fc5800a/src/transformers/generation/configuration_utils.py#L535) are defaulted at [this line](https://github.com/huggingface/transformers/blob/d14d99eadf41fd6702522eaea2f17b346fc5800a/src/transformers/generation/utils.py#L1794), which wouldn't copy custom keys in `self.generation_config` anymore

src/transformers/generation/utils.py
```diff
@@ -1760,101 +1758,50 @@ def _prepare_generated_length(
...
-                    generation_config.decoder_start_token_id = self.generation_config.decoder_start_token_id
-
-        # Finally, apply any passed kwargs
+        # First set values from the loaded `self.generation_config`, then set default values (BC)
+        # Do not update any values that aren't `None`, i.e. if set by users explicitly and passed
+        # to `generate()`. Thus the `defaults_only=True` is used
+        global_defaults = self.generation_config._get_default_generation_params()
+        generation_config.update(**self.generation_config.to_dict(), defaults_only=True)
+        generation_config.update(**global_defaults, defaults_only=True)
+
```
ebezzam, line 1793: something like this just after?

```python
# add custom keys not in global defaults
for key, value in self.generation_config.to_dict().items():
    if not hasattr(generation_config, key):
        setattr(generation_config, key, value)
```

PR #42703, opened 2025-12-08
Fix tests quantization 

SunMarc:
# What does this PR do?

This PR fixes some errors in our quantization tests. Let's have a green ci for 2026. The only two quantization method left would be bnb and autoawq but those will be fixed in follow-up PRs


SunMarc:
@bot /style

MekkCyber (approved):
Thanks for taking care of this

PR #42704, opened 2025-12-08
fix tp

SunMarc:
# What does this PR do?

This PR fixes the initialization of `parallelism_config` when using tp. Fixes https://github.com/huggingface/transformers/issues/42661


SunMarc:
@bot /style

MekkCyber (approved):
Thanks

quic-meetkuma:
Why below line was added?

Requires accelerate>1.12.0 to use Tensor Parallelism.

SunMarc:
`ParallelismConfig` comes from a recent version of accelerate

PR #42715, opened 2025-12-08
[Devstral] Make sure FP8 conversion works correctly

patrickvonplaten:
Improve script for FP8 conversion.
Tested on all relevant checkpoints (both Devstral 2 & Ministral 3B Reasoning & 14B - FP8)


src/transformers/models/ministral3/convert_ministral3_weights_to_hf.py
```diff
@@ -173,11 +181,25 @@ def convert_config(original_config: dict, max_position_embeddings: int = 262144)
...
+                "modules_to_not_convert": ["model.vision_tower", "model.multi_modal_projector", "lm_head"],
+                "quant_method": "fp8",
+                "weight_block_size": None,
+            }
+            kwargs["quantization_config"] = AutoQuantizationConfig.from_dict(quantization_config)
+        return kwargs
 
     # No vision
     if original_vision_config is None:
+        new_text_config = Ministral3Config(**new_text_config_kwargs, **get_maybe_quant_config())
```
SunMarc, line 199: let's just add the quantization config in `new_config` and not potentially in the `new_text_config`. In the `config.json` that you converted, everything looks good

SunMarc (approved):
Thanks ! LGTM, just a nit

MekkCyber (approved):
Thank you

ArthurZucker (approved):

src/transformers/models/ministral3/convert_ministral3_weights_to_hf.py
```diff
@@ -218,7 +229,8 @@ def convert_and_write_model(input_dir: str, output_dir: str, max_position_embedd
     """Convert the model and save it (this implicitly save the config as well)."""
     params = read_json(os.path.join(input_dir, "params.json"))
 
-    config = convert_config(params, max_position_embeddings)
+    is_vision = params.get("vision_encoder") is not None
```
NikeNano, line 232: Would it make sense to default to "False" instead?

src/transformers/models/ministral3/convert_ministral3_weights_to_hf.py
```diff
@@ -173,11 +181,25 @@ def convert_config(original_config: dict, max_position_embeddings: int = 262144)
...
+                "weight_block_size": None,
+            }
+            kwargs["quantization_config"] = AutoQuantizationConfig.from_dict(quantization_config)
+        return kwargs
 
     # No vision
     if original_vision_config is None:
+        new_text_config = Ministral3Config(**new_text_config_kwargs, **get_maybe_quant_config())
         return new_text_config
+    else:
```
NikeNano, line 201: Would it make sense to remove the else since we do an early return if the condition evaluates to true?

PR #42718, opened 2025-12-08
Fixes for the failures of AMD CI

remi-or:
This PR fixes the following:
- multiple devices errors for `blt`, `efficientnet`, `ernie` and `mimi`
- added a synchronization to fix a `test_multi_gpu_data_parallel_forward` that failed with `align`
- removed the requirement that `timm` should be version `<=.19` because there are some new models that are not supported by this version. The real requirement should have been version `>.19` (https://github.com/huggingface/transformers/pull/39640)

This reduces the number of fails on the AMD CI by a lot (~150 I think)

### DIFF
--- a/src/transformers/models/mimi/modeling_mimi.py
+++ b/src/transformers/models/mimi/modeling_mimi.py
@@ -1380,7 +1380,7 @@ class MimiPreTrainedModel(PreTrainedModel):
     main_input_name = "input_values"
     input_modalities = "audio"
     supports_gradient_checkpointing = True
-    _no_split_modules = ["MimiDecoderLayer"]
+    _no_split_modules = ["MimiResidualVectorQuantizer", "MimiTransformerLayer"]
     _skip_keys_device_placement = "past_key_values"
     _supports_flash_attn = True
     _supports_sdpa = True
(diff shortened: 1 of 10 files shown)


src/transformers/models/mimi/modeling_mimi.py
```diff
@@ -1379,7 +1379,7 @@ class MimiPreTrainedModel(PreTrainedModel):
     main_input_name = "input_values"
     input_modalities = "audio"
     supports_gradient_checkpointing = True
-    _no_split_modules = ["MimiDecoderLayer"]
+    _no_split_modules = ["MimiResidualVectorQuantizer", "MimiTransformerLayer"]
```
Cyrilvallez, line 1383: Yes! I was gonna say about that: most of the time, device mismatch inside modeling code means that this attribute is not set correctly, or that a `nn.Parameter` was used outside of one of those modules, and `accelerate` does not track it correctly...

remi-or: did the same for other modules that could be fixed thish way (not blt)

remi-or:
run-slow: align, blt, efficientnet, ernie, mimi

Cyrilvallez (approved):
My guy 🚀🚀🚀🚀🚀

PR #42732, opened 2025-12-09
fix(Qwen3VLProcessor): fix incorrect handling of return_metadata

yiguanxianyu:
This PR fixes the following:

- ensures video metadata is correctly returned in `Qwen3VLProcessor` when `return_metadata=True` during `processor.apply_chat_template()`.

### DIFF
--- a/src/transformers/models/qwen3_vl/modular_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/modular_qwen3_vl.py
@@ -1441,11 +1441,8 @@ class Qwen3VLProcessor(Qwen2VLProcessor):
         if videos is not None:
             videos_inputs = self.video_processor(videos=videos, **output_kwargs["videos_kwargs"])
             video_grid_thw = videos_inputs["video_grid_thw"]
-            # If user has not requested video metadata, pop it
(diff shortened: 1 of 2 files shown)


Rocketknight1:
cc @zucchini-nlp @molbap @yonigozlan

zucchini-nlp (commented):
Oh, interesting! I didn't know that the `BatchFeature` doesn't work well with non array-like structure. I don't think the current solution works well, it pushed us to update all models and the output type is different depending on the metadata

I recommend to fix it in `def as_tensor()` fn in `BatchFeature`, so that it doesn't try to convert everything to torch/np and instead checks the type of the input, whether it can be converted

zucchini-nlp:
@yonigozlan recently added a cool feature for `BatchFeature` to skip converting inputs to tensors. He'll work on adding support for non array-like inputs next :)

Thanks for pointing out the issue @yiguanxianyu

yiguanxianyu:
> @yonigozlan recently added a cool feature for `BatchFeature` to skip converting inputs to tensors. He'll work on adding support for non array-like inputs next :)
> 
> Thanks for pointing out the issue @yiguanxianyu

Thanks!

PR #42735, opened 2025-12-09
Fix AutoTokenizer fallback to TokenizersBackend when tokenizer class is unknown

Abdennacer-Badaoui:
When `AutoTokenizer.from_pretrained()` cannot determine the tokenizer class from config, it now attempts to load using `TokenizersBackend` directly before falling back to SentencePiece. This enables loading tokenizers that have `tokenizer.json` files (like Bloom models) even when the tokenizer class cannot be determined from the model configuration.

This fix will make all the failing tests in `test_pipelines_text_generation` pass, and possibly some others as well.


ArthurZucker:
Hey! if you can rebase / check this should have been fixed AFAIK

ArthurZucker:
but thanks for giving it a look

Abdennacer-Badaoui:
Yes, that's working now! 👌🏻

PR #42736, opened 2025-12-09
gpt-oss is not working with flash-attention

pushkar-hue:
When initializing `gpt-oss` model with `attn_implementation="flash_attention_2"` or `"flash_attention_3"` would result in silent failures and garbage generation output as reported in #42533.

`gpt-oss` models rely on attention sinks which are not yet implemented for the `flash_attention` as suggested the safest path is to strictly block unsupported attention backends rather than failing silently or assuming a fallback.

@vasqu can you see if this is what you had in mind?


src/transformers/models/gpt_oss/modular_gpt_oss.py
```diff
@@ -383,6 +383,17 @@ def _init_weights(self, module):
 class GptOssModel(MixtralModel):
     _no_split_modules = ["GptOssDecoderLayer"]
 
+    def __init__(self, config: GptOssConfig):
+        super().__init__(config)
+
+        if config._attn_implementation in ["flash_attention_2", "flash_attention_3"]:
```
vasqu: 2 things that might not work as expected
- This only checks during init, but we can also update afterwards with e.g. `set_attn_implementation`. We probably cannot avoid checking during runtime (e,g, forward).
- There are more FA implementations especially with the kernels ones. Afaik, we only have the vllm kernel working so we should check along `if "flash" in config._attn_implementation and config._attn_implementation != "vllm-kernel": raise ValueError(...)` (don't remember the name of the kernel).

src/transformers/models/gpt_oss/modeling_gpt_oss.py
vasqu, line 1: This is weird, there is a lot of changes like `__init__ -> _init_` which shouldn't happen. Maybe some ruff version mismatch? It's weird

vasqu (commented):
Added some comments, I have some concerns on the check + somehow there are a lot of changes in the modeling that shouldn't be there (e.g. `__ changes to _`)

src/transformers/models/gpt_oss/modeling_gpt_oss.py
pushkar-hue: @vasqu Thank you so much for your patience. There was some issue with modeling file from my side. I have updated the conditions to allow `kernels-community/vllm-flash-attn3`, I believe this is the one you were referring to. I have also added a check inside `forward` to catch any runtime configuration changes. Let me know if I'm missing something.

vasqu (commented):
2 smaller things:
- Check the CI 
  - `make style` and `make fix-copies` should help there
  - You can also generate the modeling file from modular via `python utils/modular_model_converter gpt_oss` 
- Can you add a small test so that we can check that correctly raises/not raises an error as expected

pushkar-hue:
@vasqu I have fixed the styling issues and added the test case as requested.

However, the test currently fails in CI because `flash_attn` is not installed in the environment. This causes the library to raise an `ImportError` before it reaches my new `ValueError` check.

tests/models/gpt_oss/test_modeling_gpt_oss.py
```diff
@@ -62,6 +62,23 @@ class GptOssModelTest(CausalLMModelTest, unittest.TestCase):
     model_split_percents = [0.5, 0.6]
     model_tester_class = GptOssModelTester
 
+    def test_initialization_raises_error_for_flash_attn(self):
```
vasqu, line 73: We have decorators for non-standard library, here you should use `@require_kernels` and `@require_flash_attn`

tests/models/gpt_oss/test_modeling_gpt_oss.py
```diff
@@ -62,6 +62,23 @@ class GptOssModelTest(CausalLMModelTest, unittest.TestCase):
...
+        """
+        Tests that initializing the model with unsupported Flash Attention implementations raises a ValueError,
+        but allows the specific vllm kernel.
+        """
+
+        config = self.model_tester.get_config()
+        config._attn_implementation = "flash_attention_2"
+        with self.assertRaisesRegex(ValueError, "GPT-OSS models do not support"):
+            GptOssModel(config)
```
vasqu: This checks init but not runtime

tests/models/gpt_oss/test_modeling_gpt_oss.py
```diff
@@ -62,6 +62,23 @@ class GptOssModelTest(CausalLMModelTest, unittest.TestCase):
...
+
+        config._attn_implementation = "kernels-community/vllm-flash-attn3"
+        try:
+            GptOssModel(config)
+        except ValueError:
+            self.fail("GptOssModel raised ValueError unexpectedly with vllm-flash-attn3!")
```
vasqu: I don't think we need this extra try catch, we should expect no error either way

vasqu (commented):
Small comments on the test, the first comment is about the error you see (which is due to fa not being installed on the small CI here)

pushkar-hue:
Can you check if the tests look good now? let me know if there's anything else I need to change. Thanks for your suggestions.

tests/models/gpt_oss/test_modeling_gpt_oss.py
```diff
@@ -62,6 +66,40 @@ class GptOssModelTest(CausalLMModelTest, unittest.TestCase):
     model_split_percents = [0.5, 0.6]
     model_tester_class = GptOssModelTester
 
+    @require_kernels
+    @require_flash_attn
+    @pytest.mark.flash_attn_test
```
vasqu, line 71: Just added 2 things, 1 here to mark it as FA test

tests/models/gpt_oss/test_modeling_gpt_oss.py
```diff
@@ -62,6 +66,40 @@ class GptOssModelTest(CausalLMModelTest, unittest.TestCase):
...
+        """
+
+        config, inputs_dict = self.model_tester.prepare_config_and_inputs_for_common()
+
+        config._attn_implementation = "flash_attention_2"
+        with self.assertRaisesRegex(ValueError, "GPT-OSS model does not support"):
+            model = GptOssModel(config)
+
+        config._attn_implementation = "kernels-community/vllm-flash-attn3"
+        model = GptOssModel(config).to(device=torch_device, dtype=torch.bfloat16)
```
vasqu, line 83: And here added the dtype casting to half precision (otherwise it won't run because FA only works with half precision)

vasqu (approved):
Thx for iterating just added some last nits to make the test run

src/transformers/models/gpt_oss/modular_gpt_oss.py
```diff
@@ -354,8 +354,7 @@ def forward(
 class GptOssPreTrainedModel(LlamaPreTrainedModel):
     _keep_in_fp32_modules = ["post_attention_layernorm", "input_layernorm", "norm"]
     _supports_sdpa = False
-    _supports_flash_attention = False
-    _supports_flex_attention = False
+    _supports_flex_attn = False
```
vasqu, line 357: Was wrongly typed :eyes:

src/transformers/models/gpt_oss/modular_gpt_oss.py
```diff
@@ -381,6 +380,19 @@ def _init_weights(self, module):
 class GptOssModel(MixtralModel):
     _no_split_modules = ["GptOssDecoderLayer"]
 
+    def __init__(self, config: GptOssConfig):
+        super().__init__(config)
+
+        if (
+            "flash" in config._attn_implementation
+            and config._attn_implementation.removeprefix("paged|") != "kernels-community/vllm-flash-attn3"
```
vasqu: "paged|" in case we have anything CB related

vasqu:
Gpt oss pretty important so wanna keep @ArthurZucker in the loop if this is ok

src/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -481,6 +479,16 @@ def __init__(self, config: GptOssConfig):
...
+        if (
+            "flash" in config._attn_implementation
+            and config._attn_implementation.removeprefix("paged|") != "kernels-community/vllm-flash-attn3"
+        ):
+            raise ValueError(
+                f"GPT-OSS model does not support the specified "
+                f"flash attention implementation: {config._attn_implementation}. "
+                f"Only 'kernels-community/vllm-flash-attn3' is supported."
+            )
+
```
ArthurZucker: it's not bad, but I think we could also put this in the `config` of GPTOss directly? As a property for the `attn_implementation`

ArthurZucker (approved):
Sounds good, I think documenting could also be fine, along with having a new attr in the config "default. attn ipmlementation" to make sure we chose this one for this model (cases like this are bound to happen)

src/transformers/models/gpt_oss/modeling_gpt_oss.py
pushkar-hue: Thanks for your suggestion. I agree it would be cleaner to move this check into `config`. Since this PR is approved would you prefer to merge it as it is, or should I proceed to refactor it? I'm happy to do it either way!

vasqu: Let's try to do it directly here

pushkar-hue:
@vasqu @ArthurZucker I have moved the validation inside config as suggested. Can see if this looks good? please let me know if there's anything else I need to do or if i'm missing something.

src/transformers/models/gpt_oss/configuration_gpt_oss.py
```diff
@@ -117,5 +117,22 @@ def __init__(
             **kwargs,
         )
 
+    def __setattr__(self, key, value):
```
vasqu, line 120: I had to overwrite the setter instead of the property, see the reasons below - tldr is that we have a method that sets the internal attribute which would not be considered with that.

vasqu:
@pushkar-hue sorry for the delays, we were off for the holidays, I pushed a fix to make this work with all of our functions, see https://github.com/huggingface/transformers/pull/42736#discussion_r2662432027

Merging now and thx for the contribution :hugs:

PR #42747, opened 2025-12-09
Only call `torch.autocast` if it will have an effect

hmellor:
Adds a context manager to replace direct use of `torch.autocast` that only autocasts if:
    
- `autocast` is already enabled in this context 
- Or this call to `maybe_autocast` has `enabled=True`

This prevents `autocast` being added to the graph when it is effectively a no-op.
Which makes graph splitting in `torch.compile` more flexible as it removes the requirement that partition IDs be monotonically increasing.


ArthurZucker (approved):
Perfect thanks for pushing this!

PR #42748, opened 2025-12-09
[docs] Chat content patterns

stevhliu:
Adds a guide to centralize all the different `content` patterns available for chat models

### DIFF
--- a/docs/source/en/chat_content_patterns.md
+++ b/docs/source/en/chat_content_patterns.md
@@ -1,0 +1,167 @@
+<!--Copyright 2025 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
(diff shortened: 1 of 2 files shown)


docs/source/en/chat_content_patterns.md
```diff
@@ -0,0 +1,167 @@
...
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# Chat content patterns
+
+Chat models expect conversations as a list of dictionaries. Each dictionary uses `role` and `content` keys. The `content` key holds the user message passed to the model. Large language models accept text and multimodal models combine text with images, videos, and audio.
```
zucchini-nlp: do you think `tools` would count as another content type?

docs/source/en/chat_content_patterns.md
```diff
@@ -0,0 +1,167 @@
...
+
+Pass your message to the `"text"` key.
+
+```py
+message = [
+    {
+        "role": "user",
+        "content": [{"type": "text", "text": "Explain the French Bread Law."}]
+    }
+]
```
zucchini-nlp, line 46: should we mention that it can be a simple str for LLMs or is this doc intended to be multimodal?

zucchini-nlp (approved):
Nice, I feel like the doc focuses on multimodal chats currently. Maybe we can extend with other patterns from LLMs such as tools, or we can move it inside multimodal docs page. Up to you

docs/source/en/chat_content_patterns.md
stevhliu: ooh yes definitely!

docs/source/en/chat_content_patterns.md
stevhliu: mentioned strings as the first option! the doc is supposed to cover all the content patterns. i think there are just more options for multimodal models than llms 😃

PR #42749, opened 2025-12-09
[docs] Update shard size

stevhliu:
Updates docs to reflect increased shard size in #42734

### DIFF
--- a/docs/source/en/models.md
+++ b/docs/source/en/models.md
@@ -117,7 +117,7 @@ Transformers reduces some of these memory-related challenges with fast initializ
 Parameters load in parallel and peak memory only depends on model size. Set `max_shard_size` in [`~PreTrainedModel.save_pretrained`] to control the threshold.
 
 > [!NOTE]
-> Memory usage for models requiring dynamic weight conversion depends on the model size and the size of the largest parameters in a single conversion. This typically applies to mixture-of-experts (MoE) models where the memory usage is the model size plus the number of experts on one layer.
+> Memory usage for models requiring dynamic weight conversion depends on the model size and the size of the largest parameters in a single conversion. This typically applies to mixture-of-experts (MoE) models where the memory usage is the model size plus the number of experts on one layer. Refer to the [dynamic weight loader](./weightconverter#fast-model-loading) guide to learn more about how models are loaded.
 
 [`~PreTrainedModel.save_pretrained`] also creates an index file mapping parameter names to their shard files. The index contains two keys, `metadata` and `weight_map`.
 
--- a/docs/source/en/models.md
+++ b/docs/source/en/models.md
@@ -117,7 +117,7 @@ Transformers reduces some of these memory-related challenges with fast initializ
 Parameters load in parallel and peak memory only depends on model size. Use `max_shard_size` in [`~PreTrainedModel.save_pretrained`] to set the maximum checkpoint size before sharding.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


docs/source/en/models.md
```diff
@@ -112,64 +112,31 @@ Transformers reduces some of these memory-related challenges with fast initializ
 
 ### Sharded checkpoints
 
-The [`~PreTrainedModel.save_pretrained`] method automatically shards checkpoints larger than 10GB.
+[`~PreTrainedModel.save_pretrained`] automatically shards checkpoints larger than 50GB. This keeps shard counts low for large models and simplifies file management without significantly slowing load times.
```
Cyrilvallez: Let's not make people see potential ghosts here: there is literally no slowing down at all!

```suggestion
[`~PreTrainedModel.save_pretrained`] automatically shards checkpoints larger than 50GB. This keeps shard counts low for large models and simplifies file management without slowing load times.
```

docs/source/en/models.md
```diff
@@ -112,64 +112,31 @@ Transformers reduces some of these memory-related challenges with fast initializ
...
 
-Each shard is loaded sequentially after the previous shard is loaded, limiting memory usage to only the model size and the largest shard size.
+Shards load sequentially and memory usage is limited to the model size plus the largest shard. Set `max_shard_size` in [`~PreTrainedModel.save_pretrained`] to control the threshold.
```
Cyrilvallez: This is not true anymore!
Parameters are loaded in parallel by default now, (can be deactivated by setting `HF_DEACTIVATE_ASYNC_LOAD` env variable), and the memory usage is strictly constrained to the model size, EXCEPT if the model needs on-the-fly weight conversions (i.e. most MoE models), in which case the memory peak is `model_size + largest_params_needed_in_a_single_conversion`, i.e. for the MoE models, `model_size + experts_on_one_layer`

Cyrilvallez (commented):
Nice, thanks for being quick on this!! It's indeed a very important aspect!

docs/source/en/models.md
stevhliu: Ah thanks, I forgot about this! I think it'd also be nice to link to https://github.com/huggingface/transformers/pull/42636, which touches on this a bit at the [end](https://moon-ci-docs.huggingface.co/docs/transformers/pr_42636/en/weightconverter#fast-model-loading), for more details. wdyt?

Cyrilvallez: Definitely!

docs/source/en/models.md
```diff
@@ -112,64 +112,34 @@ Transformers reduces some of these memory-related challenges with fast initializ
 
 ### Sharded checkpoints
 
-The [`~PreTrainedModel.save_pretrained`] method automatically shards checkpoints larger than 10GB.
+[`~PreTrainedModel.save_pretrained`] automatically shards checkpoints larger than 50GB. This keeps shard counts low for large models and simplifies file management.
 
-Each shard is loaded sequentially after the previous shard is loaded, limiting memory usage to only the model size and the largest shard size.
+Parameters load in parallel and peak memory only depends on model size. Set `max_shard_size` in [`~PreTrainedModel.save_pretrained`] to control the threshold.
```
Cyrilvallez: What threshold are we talking about here?

Cyrilvallez (approved):
Nice! Thanks a lot! Just left a final comment

docs/source/en/models.md
stevhliu: threshold refers to the maximum checkpoint size before sharding. i updated it so its more clear what "threshold" is :)

PR #42766, opened 2025-12-10
[saving] Simplify general logic

Cyrilvallez:
# What does this PR do?

As per the title


ArthurZucker (approved):
yep nice simplification ty

PR #42770, opened 2025-12-10
Override Transformers defaults by GGUF defaults

a4lg:
# What does this PR do?

In some models, GGUF uses default or fixed values different from this library.
To integrate GGUF-based models without additional configuration (and/or individual `config.json`), we need some kind of compatibility layer (like tensor processors in `src/transformers/modeling_gguf_pytorch_utils.py` while loading/converting GGUF-based model tensors).

This commit provides additional mapping to provide GGUF-specific default values to initialize parameters in this library.

Currently, only fixed `norm_topk_prob` value of Qwen3 MoE (`True`) is defined because:

1.  It differs from the default value of this library (`False`) and
2.  If this parameter is incorrectly set, it results in almost completely garbled output (see vllm-project/vllm#30118 for examples).

I'm sure that Qwen3 MoE + GGUF issue should be addressed somehow but I'm not sure whether this PR is the right way to fix the issue.
So, feel free to leave a review!

I could not check all parameters but at least I checked important parameters for MoE models with GGUF support in this library.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@CyrilVallez @SunMarc @MekkCyber

### DIFF
--- a/src/transformers/integrations/ggml.py
+++ b/src/transformers/integrations/ggml.py
@@ -313,10 +313,11 @@ GGUF_TOKENIZER_MAPPING = {
     },
 }
 
+# We only need to set here the parameters that default to different values between transformers and llamacpp.
 GGUF_CONFIG_DEFAULTS_MAPPING = {
     "qwen3_moe": {
         # NOTE: Qwen3MoeConfig defaults to false but llama.cpp needs this to be true.
-        # See: https://github.com/ggml-org/llama.cpp/blob/b7345/src/models/qwen3moe.cpp#L85-L96
+        # See: https://github.com/ggml-org/llama.cpp/blob/17f7f4baad8b3a716ee139da7bb56ae984e8c0fa/src/models/qwen3moe.cpp#L85-L96
         #      (the parameter right after LLM_FFN_SILU corresponds to norm_topk_prob)
         "norm_topk_prob": True,
     },
--- a/src/transformers/integrations/ggml.py
+++ b/src/transformers/integrations/ggml.py
@@ -313,6 +313,15 @@ GGUF_TOKENIZER_MAPPING = {
     },
 }
 
+GGUF_CONFIG_DEFAULTS_MAPPING = {
+    "qwen3_moe": {
+        # NOTE: Qwen3MoeConfig defaults to false but llama.cpp needs this to be true.
+        # See: https://github.com/ggml-org/llama.cpp/blob/b7345/src/models/qwen3moe.cpp#L85-L96
+        #      (the parameter right after LLM_FFN_SILU corresponds to norm_topk_prob)
+        "norm_topk_prob": True,
+    },
+}
+
 
 def _gguf_parse_value(_value, data_type):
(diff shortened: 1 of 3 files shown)


src/transformers/integrations/ggml.py
```diff
@@ -313,6 +313,12 @@
     },
 }
 
+GGUF_CONFIG_DEFAULTS_MAPPING = {
+    "qwen3_moe": {
+        "norm_topk_prob": True,
+    },
+}
```
MekkCyber, line 324: could you add a comment to specify that we only need to set this if the default values differ between `transformers` and `llamacpp`

MekkCyber (commented):
Thanks for looking into this @a4lg !

src/transformers/integrations/ggml.py
a4lg: Okay.  Like the comment which looks like this?

```python
# NOTE: Qwen3MoeConfig defaults to false but llama.cpp needs this to be true.
# See: https://github.com/ggml-org/llama.cpp/blob/b7345/src/models/qwen3moe.cpp#L85-L96
#      (the parameter right after LLM_FFN_SILU corresponds to norm_topk_prob)
"norm_topk_prob": True,
```

src/transformers/integrations/ggml.py
```diff
@@ -313,6 +313,15 @@
...
 }
 
+GGUF_CONFIG_DEFAULTS_MAPPING = {
+    "qwen3_moe": {
+        # NOTE: Qwen3MoeConfig defaults to false but llama.cpp needs this to be true.
+        # See: https://github.com/ggml-org/llama.cpp/blob/b7345/src/models/qwen3moe.cpp#L85-L96
+        #      (the parameter right after LLM_FFN_SILU corresponds to norm_topk_prob)
+        "norm_topk_prob": True,
+    },
+}
```
MekkCyber, line 324:
```suggestion
# We only need to set here the parameters that default to different values between transformers and llamacpp.
GGUF_CONFIG_DEFAULTS_MAPPING = {
    "qwen3_moe": {
        # NOTE: Qwen3MoeConfig defaults to false but llama.cpp needs this to be true.
        # See: https://github.com/ggml-org/llama.cpp/blob/17f7f4baad8b3a716ee139da7bb56ae984e8c0fa/src/models/qwen3moe.cpp#L85-L96
        #      (the parameter right after LLM_FFN_SILU corresponds to norm_topk_prob)
        "norm_topk_prob": True,
    },
}
```

MekkCyber: sounds good, i just added a comment about the use of the whole dict, and added a permanent link instead

MekkCyber (approved):
Thank you for iterating

src/transformers/integrations/ggml.py
a4lg: Feels right and applied!

SunMarc (approved):
Thanks !

PR #42775, opened 2025-12-10
[Quantization] Fix Static FP8 Quantization

### LINKED ISSUE #42746: [Devstral 24B] FP8 is currently not working correctly
### System Info

Latest transformers "main".

### Who can help?

@SunMarc @MekkCyber 

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

If you run this code snippet: https://huggingface.co/mistralai/Devstral-Small-2-24B-Instruct-2512#transformers **with** `dequantize=False`, you will notice that it has infinite repetition issues.

If however you run the model from this PR: https://github.com/huggingface/transformers/pull/42715, you can see that everything works fine which leads to the conclusion that something funky is going on with the activation scales (maybe they give `inf` values somewhere?).

The same activation scales work for vLLM: https://huggingface.co/mistralai/Devstral-Small-2-24B-Instruct-2512#vllm-recommended so there is probably something we can do inside transformers to fix it?

### Expected behavior

That FP8 works correctly

MekkCyber:
# What does this PR do?

Fix https://github.com/huggingface/transformers/issues/42746, before downcasting to FP8 we should clamp values to the right range to avoid having NaNs

### DIFF
--- a/src/transformers/integrations/finegrained_fp8.py
+++ b/src/transformers/integrations/finegrained_fp8.py
@@ -486,7 +486,7 @@ class FP8Linear(nn.Linear):
             if self.bias is not None:
                 output = output + self.bias
 
-            output = torch.nan_to_num(output, nan=0.0)
+#            output = torch.nan_to_num(output, nan=0.0)
             return output.to(dtype=input.dtype)
 
 
--- a/src/transformers/integrations/finegrained_fp8.py
+++ b/src/transformers/integrations/finegrained_fp8.py
@@ -466,17 +466,22 @@ class FP8Linear(nn.Linear):
                     qinput, scale = act_quant(input, self.block_size[1])
                 elif self.activation_scheme == "static":
                     scale = self.activation_scale.to(torch.float32)
-                    qinput = (input / scale).to(torch.float8_e4m3fn)
+                    qinput = (input / scale)
                 else:
                     raise NotImplementedError("Not supported")
-                output = w8a8_block_fp8_matmul_triton(
-                    qinput,
-                    weight,
-                    scale,
-                    scale_inv,
-                    self.block_size,
-                    output_dtype=input.dtype,
-                )
+
+                if self.activation_scheme == "static":
+                    # TODO: fix that so we don't have to upcast to bfloat16
+                    output = F.linear(qinput.to(torch.bfloat16), weight.to(torch.bfloat16), self.bias) * scale * scale_inv
+                else:
+                    output = w8a8_block_fp8_matmul_triton(
+                        qinput,
+                        weight,
+                        scale,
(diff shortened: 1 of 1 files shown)


src/transformers/integrations/finegrained_fp8.py
```diff
@@ -466,17 +466,24 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
...
-                    scale_inv,
-                    self.block_size,
-                    output_dtype=input.dtype,
-                )
+
+                if self.activation_scheme == "static":
+                    # TODO: fix that so we don't have to upcast to bfloat16
+                    output = (
+                        F.linear(qinput.to(torch.bfloat16), weight.to(torch.bfloat16), self.bias) * scale * scale_inv
+                    )
```
SunMarc:
```suggestion
                if self.activation_scheme == "static":
                    # TODO: fix that so we don't have to upcast to bfloat16
                    output = (
                        F.linear(qinput.to(input.dtype), weight.to(input.dtype), self.bias) * scale * scale_inv
                    )
```

SunMarc (approved):
Thanks, let's do that for now

SunMarc:
@bot style

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -466,17 +466,24 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
...
-                    qinput,
-                    weight,
-                    scale,
-                    scale_inv,
-                    self.block_size,
-                    output_dtype=input.dtype,
-                )
+
+                if self.activation_scheme == "static":
+                    # TODO: fix that so we don't have to upcast to bfloat16
```
patrickvonplaten: This will slow down all ministral instruct and devstral 123B model however no? We should maybe at least add a warning that FP8 is not yet fully supported or else a bunch of people will get a silent slow down no?

MekkCyber: Should be fixed now

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -483,6 +485,7 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
             torch_accelerator_module.synchronize()
             if self.bias is not None:
                 output = output + self.bias
+
             output = torch.nan_to_num(output, nan=0.0)
```
SunMarc: let's remove or comment the `nan_to_num`, this is confusing I think. In this case, it gave a somewhat correct output but not the intended value 
```suggestion
#            output = torch.nan_to_num(output, nan=0.0)
```

SunMarc:
@bot /style

PR #42776, opened 2025-12-10
fix awq 

SunMarc:
# What does this PR do?

This PR simplifies a bit the AWQ integration. cc @Qubitium 
Also, I fixed some issues with gptq tests as the hub tests are failiing due to them

### DIFF
--- a/src/transformers/quantizers/quantizer_awq.py
+++ b/src/transformers/quantizers/quantizer_awq.py
@@ -18,11 +18,10 @@ from packaging import version
 
 from .base import HfQuantizer
 
-
 if TYPE_CHECKING:
     from ..modeling_utils import PreTrainedModel
 
-from ..utils import is_accelerate_available, is_gptqmodel_available, is_torch_available, logging
+from ..utils import is_accelerate_available, is_gptqmodel_available, is_llm_awq_available , is_torch_available, logging
 from ..utils.quantization_config import AwqBackend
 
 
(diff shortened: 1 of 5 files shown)


Qubitium:
@SunMarc There is an issue where gpt-qmodel will not select the correct kernel for llm-awq in this PR due to the config compat diff between autoawq and llm-awq. We are checking to see if gpt-qmodel needs to be patched only, or requires small changes in HF as well.

src/transformers/quantizers/quantizer_awq.py
```diff
@@ -43,9 +43,11 @@ class AwqQuantizer(HfQuantizer):
     def __init__(self, quantization_config, **kwargs):
         super().__init__(quantization_config, **kwargs)
 
-    def validate_environment(self, device_map, **kwargs):
+    def validate_environment(self, **kwargs):
         if not is_gptqmodel_available():
-            raise ImportError("Loading an AWQ quantized model requires gptqmodel library (`pip install gptqmodel`)")
+            raise ImportError(
+                "Loading an AWQ quantized model requires gptqmodel. Please install it with `pip install gptqmodel`"
```
MekkCyber, line 49: does this work or we should direct them to the github repo?

MekkCyber (approved):
Niice

Qubitium:
@SunMarc @MekkCyber gpt-qmodel and autoawq does not support support llm-awq quantized models so this pr will break llm-awq loading.

After checking, llm-awq stores qweight/scales differently and inference kernel is also different. autoawq also does not support llm-awq from a quick scan.

As such gpt-qmodel only supports autoawq compat and does not yet support llm-awq format/kernel. We can add this support in a future gpt-qmodel update but until then, we still need the if/else branching code for llm-awq

SunMarc:
> @SunMarc @MekkCyber gpt-qmodel and autoawq does not support support llm-awq quantized models so this pr will break llm-awq loading.
> 
> After checking, llm-awq stores qweight/scales differently and inference kernel is also different. autoawq also does not support llm-awq from a quick scan.
> 
> As such gpt-qmodel only supports autoawq compat and does not yet support llm-awq format/kernel. We can add this support in a future gpt-qmodel update but until then, we still need the if/else branching code for llm-awq

Well, I think it fine to remove llm-awq support. The code was added before as we hoped that the llm-awq will be maintained in the same way as auto-awq but clearly it was abandoned as well. Most models on the hub don't use this format so let's take this chance to remove it entirely. WDTY ? We can still add it back if users complain but I don't think this will happen.

Qubitium:
> Well, I think it fine to remove llm-awq support. The code was added before as we hoped that the llm-awq will be maintained in the same way as auto-awq but clearly it was abandoned as well. Most models on the hub don't use this format so let's take this chance to remove it entirely. WDTY ? We can still add it back if users complain but I don't think this will happen. 

If llm-awq compat is not an issue, then this pr is good to go. I am in agreement here. 

We will add llm-awq inference support to gpt-qmodel sometime next week so the og awq model support is retained for academic purposes as I still see most papers using reference og gptq and awq code rather than more modern frameworks.

PR #42784, opened 2025-12-10
Preprocessing fixes and more tests for LFM2-VL

ankke:
Fixes: 
* no image tokens for small images when `use_thumbnail` set to False
* repeated text part of the prompt for multi-image and multi-turn input

Adds several missing tests to catch bugs like above earlier.


Rocketknight1:
cc @zucchini-nlp @molbap @yonigozlan

zucchini-nlp:
run-slow: lfm2_vl

zucchini-nlp (approved):
Super extensive tests, thanks!

zucchini-nlp:
Tests are failing on main branch so I can't yet merge until they are fixed, let's wait

PR #42799, opened 2025-12-11
Add requires_backends to the main init

LysandreJik:
Fixes https://github.com/huggingface/transformers/issues/42778


ArthurZucker (approved):
Then it needs to be documented no? in the `tools` section

PR #42801, opened 2025-12-11
[`Padding-Free Attention`] Fix packed FA attention with pos ids only 

vasqu:
We pass position ids as kwargs which are picked up by FA to be used as padding-free variant. This is already tested within the fa tests (padding-free) - the issue is that FA had so many issues, it was not so easy to see when things like these slip through (especially on older models).

Also updated the Transformers kwargs to include position ids for better typing.


src/transformers/models/gpt_oss/modular_gpt_oss.py
```diff
@@ -299,7 +298,6 @@ def forward(
             dropout=0.0 if not self.training else self.attention_dropout,
             scaling=self.scaling,
             sliding_window=self.sliding_window,
-            position_ids=position_ids,
```
vasqu, line 302: Could keep it but thought aligning makes more sense

ArthurZucker (approved):
well def missing a test!

vasqu:
Tests are already here: https://github.com/huggingface/transformers/blob/0af2381f7bfae0569b876dd40eefda1503da927f/tests/generation/test_utils.py#L2082-L2112

We cannot avoid having this as FA test because only FA depends on this :cry: the other do this via mask creation while FA needs the positions ids itself

PR #42802, opened 2025-12-11
fix error: 'BlockMask' object has no attribute 'dtype' for lasr model

kaixuanliu:
It returns error `AttributeError: 'BlockMask' object has no attribute 'dtype'` when running test case:
`pytest -rA tests/models/lasr/test_modeling_lasr.py::LasrEncoderModelTest::test_flex_attention_with_grads`. This PR will fix this bug.
@eustlb @ebezzam @vasqu pls help review


vasqu (commented):
I'd rather we disable flex attention support over here https://github.com/huggingface/transformers/blob/86a93150842428d1b62e5172000a86a2bcd1d53e/src/transformers/models/lasr/modeling_lasr.py#L428 (via modular)

The reason is that while it runs, it fundamentally does something wrong --> The padding is simply ignored. We shouldn't introduce faulty behavior IMO but cc @eustlb

yao-matrix:
@kaixuanliu , could you disable flex_attention for lasr?

kaixuanliu:
Sure, have updated the code.  @vasqu plss help review again.

src/transformers/models/lasr/modular_lasr.py
```diff
@@ -422,6 +422,8 @@ def forward(
 
 
 class LasrPreTrainedModel(ParakeetPreTrainedModel):
+    _supports_flex_attn = False
```
vasqu, line 430: Let's add a small comment why

vasqu (approved):
LGTM, just a small nit

src/transformers/models/lasr/modular_lasr.py
kaixuanliu: Done.

yao-matrix:
@vasqu , seems the ci error not related to this PR, could you pls take a second look and merge if it's OK, thx very much.

src/transformers/models/lasr/modular_lasr.py
vasqu: Ah, I meant the comment to be more specific, i.e. `padding is incompatible with flex attention as the resulting mask cannot be used to apply padding`

kaixuanliu: ok, updated.

vasqu:
@yao-matrix @kaixuanliu Thx for iterating, merging now :hugs:

PR #42805, opened 2025-12-11
Default auto 🚨 🚨 

ArthurZucker:
# What does this PR do?
Superseeds #34919


src/transformers/modeling_utils.py
```diff
@@ -724,16 +732,26 @@ def _get_dtype(
                     if is_sharded and "dtype" in sharded_metadata:
                         dtype = sharded_metadata["dtype"]
                     elif state_dict is not None:
-                        dtype = get_state_dict_dtype(state_dict)
+                        dtype = get_state_dict_dtype(state_dict, getattr(config, "dtype", None))
```
Cyrilvallez: Why do we need to change this? They are both already in branches when we know that `config.dtype` does not exist anyway

src/transformers/modeling_utils.py
```diff
@@ -269,16 +269,24 @@ def get_torch_context_manager_or_global_device():
     return device_in_context
 
 
-def get_state_dict_dtype(state_dict):
+def get_state_dict_dtype(state_dict, config_dtype: Optional[torch.dtype] = None):
```
Cyrilvallez: I don't think this function needs to be changed, see previous comment

src/transformers/modeling_utils.py
```diff
@@ -1221,6 +1239,14 @@ def __init__(self, config: PreTrainedConfig, *inputs, **kwargs):
...
             )
         self.config = config
+        if getattr(self.config, "dtype", None) is None:
+            default_dtype = torch.get_default_dtype()
+            self.config.dtype = default_dtype
+            for sub_config_key in self.config.sub_configs:
+                if (sub_config := getattr(self.config, sub_config_key)) is not None and getattr(
+                    sub_config, "dtype", None
+                ) is None:
+                    sub_config.dtype = default_dtype
```
Cyrilvallez: Do we need to write it at `__init__`? In any case, no need to do it on all subconfigs, as all submodels run `__init__` during the whole `__init__` process (so writing it only to `self.config` is enough, they will each do it)

Cyrilvallez (commented):
Very happy so break this and have `auto` by default!!

Cyrilvallez (approved):
Let's gooooo the nice defaults 🚀🚨

BenjaminBossan:
I understand that this is a big change that is intended to break some stuff. E.g. for opt-125m, the auto dtype is now float16 instead of float32. However, this PR also makes it so that the `dtype` attribute is no longer updated when converting the model. Is that also intended?

```python
import torch

model_id = "facebook/opt-125m"

# previous commit (8a2a83d574fd461697a29410a36737ed112f8ba7)
# this passes
assert model.dtype == torch.float32
model.half()
assert {p.dtype for p in model.parameters()} == {torch.float16}
assert model.dtype == torch.float16, f"not float16, got {model.dtype} instead"  # passes

# after this commit (6217adc6c8f0be7b5374e6a46129ad2214e4c6ed)
assert model.dtype == torch.float16  # <= used to be float32
model.float()
assert {p.dtype for p in model.parameters()} == {torch.float32}
assert model.dtype == torch.float32, f"not float32, got {model.dtype} instead"  # fails
# AssertionError: not float32, got torch.float16 instead
```

Cyrilvallez:
Ha damn, I just merged https://github.com/huggingface/transformers/pull/42825 but seeing this now... Will open a new one, it indeed looks like relying on the config is a bad idea in those cases!

PR #42806, opened 2025-12-11
Remove duplicated processor class from config

zucchini-nlp:
# What does this PR do?

As per title, adding/adjusting tests and then ready for review


src/transformers/processing_utils.py
```diff
@@ -743,7 +732,6 @@ def cast_array_to_list(dictionary):
             )
         }
         output = cast_array_to_list(output)
-        output = save_public_processor_class(output)
```
yonigozlan, line 746: Does that mean we're not saving "processor_class" even at the general level?

tests/test_processing_common.py
```diff
@@ -428,6 +428,11 @@ def test_processor_from_and_save_pretrained(self):
 
                     # tokenizer repr contains model-path from where we loaded
                     if "tokenizer" not in attribute:
+                        # We don't store/load `_processor_class` for subprocessors.
+                        # The `_processor_class` is saved once per config, at general level
+                        self.assertFalse(hasattr(attribute_second, "_processor_class"))
+                        self.assertFalse(hasattr(attribute_first, "_processor_class"))
```
yonigozlan, line 434: Should we test for "processor_class" also?

yonigozlan (approved):
I just had a few questions, but other than that LGTM for me !

src/transformers/processing_utils.py
zucchini-nlp: ah no, in general level the next line add it `output['processor_class'] = self.__class__.__name__`

PR #42807, opened 2025-12-11
Raise conversion errors after loading

Cyrilvallez:
# What does this PR do?

As per the title. As discussed offline, it's very dangerous in general to NOT re-raise them, as it would make any bug in Converters invisible to us in tests (and to users not looking properly at logs)

### DIFF
--- a/src/transformers/utils/loading_report.py
+++ b/src/transformers/utils/loading_report.py
@@ -170,6 +170,7 @@ def log_state_dict_report(
     color_enabled = bool(color and sys.stdout.isatty())
     ansi = ANSI(color_enabled)
 
+    # Re-raise errors early if needed
     if error_msgs:
         error_msg = "\n\t".join(error_msgs)
         if "size mismatch" in error_msg:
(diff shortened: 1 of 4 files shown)


src/transformers/utils/loading_report.py
```diff
@@ -204,13 +212,6 @@ def log_state_dict_report(
             )
             rows.append(data)
 
-    if misc:
-        for k, v in update_key_name(misc).items():
-            status = "MISC"
-            status = _color(status, "purple", ansi)
-            _details = v[:term_w]
-            rows.append([k, status, _details])
-
```
ArthurZucker, line 213: We still want them in the report

src/transformers/utils/loading_report.py
```diff
@@ -179,6 +179,14 @@ def log_state_dict_report(
...
         raise RuntimeError(f"Error(s) in loading state_dict for {model.__class__.__name__}:\n\t{error_msg}")
 
+    if misc:
+        misc_error_msg = ""
+        for k, v in update_key_name(misc).items():
+            misc_error_msg += f"\n{k}: {v}"
+        raise RuntimeError(
+            f"We encountered the following issues during automatic conversion of the weights:{misc_error_msg}"
+        )
+
```
ArthurZucker: this should be after the full report maybe? The report is a `at a glance` idea of what went wrong

src/transformers/utils/loading_report.py
```diff
@@ -230,14 +231,12 @@ def log_state_dict_report(
...
+        tips += f"\n- {_color('MISMATCH', 'yellow', ansi) + ansi['italic']}\t:ckpt weights were loaded, but they did not match the original empty weight shapes."
     tips += f"{ansi['reset']}"
 
+    # Log the report as warning
     logger.warning(prelude + table + tips)
     if not ignore_mismatched_sizes and mismatched_keys:
         raise RuntimeError(
             "You set `ignore_mismatched_sizes` to `False`, thus raising an error. For details look at the above report!"
         )
-    return prelude + table + tips
```
ArthurZucker, line 243: this is useful for debugging!

ArthurZucker (approved):

src/transformers/utils/loading_report.py
Cyrilvallez: IMO the report is broken in those cases anyway, as we have `missing_keys` that are missing just because the conversion errored out no?

PR #42808, opened 2025-12-11
Automatic release script

LysandreJik:
Automatic releases to pypi when a tag is pushed; the workflow will still need to be validated by a core maintainer.


.github/workflows/release.yml
```diff
@@ -0,0 +1,63 @@
...
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - 'v*-release'
+
+env:
+  PYPI_TOKEN: ${{ secrets.PYPI_TOKEN_DIST }}
+
```
ArthurZucker:
```suggestion
```

ArthurZucker (approved):

PR #42809, opened 2025-12-11
refactor + robusts tests for Tensor Parallel 

3outeille:
image

- distributed training CI (https://github.com/huggingface/transformers/pull/42765) needs FSDP to be refactor.  We need same FSDP degree to compare training loss properly (https://github.com/pytorch/torchtitan/blob/main/docs/converging.md#guidelines) 
- Referencing https://github.com/huggingface/transformers/pull/43157 because it uses it and hasn't been review


src/transformers/core_model_loading.py
```diff
@@ -621,33 +621,36 @@ def set_param_for_module(
...
+    
+    # super important otherwise _init_weight will re-init the param
+    param_value._is_hf_initialized = True
+    # local_rowise/colwise case
+    if is_local_tensor and isinstance(param_value, DTensor):
+        param_value = param_value.to_local()
+
+    if param_name not in module_obj._buffers:
+        param_value = torch.nn.Parameter(param_value, requires_grad=param_value.is_floating_point())
+
```
ArthurZucker: makes more sense perf!

src/transformers/integrations/tensor_parallel.py
```diff
@@ -758,6 +758,15 @@ def partition_tensor(self, param, empty_param, param_type, param_casting_dtype,
         return nn.Parameter(parameter, requires_grad=parameter.is_floating_point())
 
 
+class LocalColwiseParallel(ColwiseParallel):
```
ArthurZucker: nice

ArthurZucker (commented):
🚀 starting to look good!

src/transformers/models/mixtral/configuration_mixtral.py
```diff
@@ -120,8 +120,7 @@ class MixtralConfig(PreTrainedConfig):
         "layers.*.self_attn.k_proj": "colwise",
         "layers.*.self_attn.v_proj": "colwise",
         "layers.*.self_attn.o_proj": "rowwise",
-        "layers.*.mlp.gate": "ep_router",  # we need to replicate here to correctly route experts
```
3outeille, line 120: In TP, all ranks have all experts but just sharded weights on each GPU. If  we use`RouterParallel`, we will find ourselves in a case where we mask an expert that is needed (because EP assume we have a full expert in a GPU, not every experts but sharded in all GPUs ), thus missing its output, thus having partial output, thus incorrect `all_reduce` that happens at `local_rowise` in `down_proj`

src/transformers/models/mixtral/configuration_mixtral.py
```diff
@@ -120,8 +120,7 @@ class MixtralConfig(PreTrainedConfig):
...
-        "layers.*.mlp.experts.gate_up_proj": "local_colwise",
+        "layers.*.mlp.experts.gate_up_proj": "local_packed_colwise",
```
3outeille: `gate_up` is packed so we need to use a local version of `packed_colwise`

3outeille:
referencing https://github.com/huggingface/transformers/pull/43003

ArthurZucker (approved):

src/transformers/integrations/tensor_parallel.py
```diff
@@ -512,10 +512,19 @@ def _prepare_input_fn(input_layouts, desired_input_layouts, mod, inputs, device_
...
         if isinstance(outputs, torch.Tensor):
-            dist.all_reduce(outputs, op=dist.ReduceOp.SUM, async_op=False)
+            outputs = DTensor.from_local(outputs, device_mesh, [Partial()], run_check=False)
+            outputs = outputs.redistribute(placements=(Replicate(),), async_op=False)
+            if use_local_output:
+                outputs = outputs.to_local()
         else:
-            dist.all_reduce(outputs[0], op=dist.ReduceOp.SUM, async_op=False)
+            output_tensor = outputs[0]
+            output_tensor = DTensor.from_local(output_tensor, device_mesh, [Partial()], run_check=False)
```
ArthurZucker: this looks costly

src/transformers/quantizers/quantizer_compressed_tensors.py
```diff
@@ -84,13 +84,15 @@ def _process_model_after_weight_loading(self, model, **kwargs):
...
+    # NOTE: TP plan override for compressed tensors removed - unsupported styles were used.
+    # TODO: Implement proper TP support for compressed tensors quantization
     def update_tp_plan(self, config):
         additional_plan = {
-            "layers.*.feed_forward.experts.*.gate_proj.weight": "local_colwise",
-            "layers.*.feed_forward.experts.*.gate_proj.weight_scale": "local_colwise",
-            "layers.*.feed_forward.experts.*.up_proj.weight": "local_colwise",
-            "layers.*.feed_forward.experts.*.up_proj.weight_scale": "local_colwise",
-            "layers.*.feed_forward.experts.*.down_proj.weight": "local_rowwise",
+            "layers.*.feed_forward.experts.*.gate_proj.weight": "lcolwise",
```
3outeille: typo

src/transformers/core_model_loading.py
```diff
@@ -915,21 +913,21 @@ def set_param_for_module(
...
-                        param_value,
-                        distributed_operation.device_mesh,
-                        getattr(distributed_operation, "shard", Replicate()),
-                        run_check=False,
-                        shape=ref.size(),
-                        stride=ref.stride(),
-                    )
+                device_mesh = distributed_operation.device_mesh
+                target_device = torch.device(f"{device_mesh.device_type}:{device_mesh.get_local_rank()}")
+                param_value = param_value.to(target_device)
```
ArthurZucker: this is a mistake, what takes the mos amount of time is literally this operation: non asycn move to device. The worst part is you are also most probably getting the full tensor on all devices.

src/transformers/core_model_loading.py
```diff
@@ -915,21 +913,21 @@ def set_param_for_module(
...
             if param_name not in module_obj._buffers:
                 param_value = torch.nn.Parameter(param_value, requires_grad=param_value.is_floating_point())
 
         # Remove from missing keys (it's either mismatched, or all good)
         missing_keys.discard(target_name)
-        if ref is not None and ref.shape != param_value.shape and hf_quantizer is None:
+        # Skip shape check when tensor parallel sharding is applied (shape is intentionally different)
```
ArthurZucker: as we discussed, we cannot do this. You need to expose a function to check shapes

src/transformers/integrations/tensor_parallel.py
```diff
@@ -424,6 +426,208 @@ def get_tensor_shard(param, empty_param, device_mesh, rank, dim, tensor_idx: int
...
+def _split_along_last_dim(x, world_size):
+    """Split tensor along last dimension into world_size chunks."""
+    return torch.chunk(x, world_size, dim=-1)
+
+
+# =============================================================================
+# Distributed Communication Primitives
+# =============================================================================
+#
+# Naming convention:
```
ArthurZucker, line 438: nice I like that!

src/transformers/integrations/tensor_parallel.py
```diff
@@ -424,6 +426,208 @@ def get_tensor_shard(param, empty_param, device_mesh, rank, dim, tensor_idx: int
...
+    def forward(ctx, x, device_mesh):
+        ctx.device_mesh = device_mesh
+        return x
+
+    @staticmethod
+    @torch._dynamo.disable
+    # NOTE(3outeille):
+    # Newer versions of PyTorch has torch.library.register_autograd in https://github.com/pytorch/pytorch/blob/8bcedd6e6029cce5f3a3731dd59be4941414c731/torch/distributed/_functional_collectives.py#L630
+    # that fix the warning "autograd kernel was not registered to the Autograd key(s) but we are trying to backprop through it"
+    # but we are forced to disable dynamo for now to make it compile-compatible for now
```
ArthurZucker: I suppose its not full graph comptible? This is a big loss because backward benefits the most from full compile AFAIK

src/transformers/integrations/tensor_parallel.py
```diff
@@ -1076,38 +1054,93 @@ def _prepare_output_fn(output_layouts, use_local_output, mod, outputs, device_me
             )
         num_local_experts = mod.num_experts // ep_size
         router_logits, router_scores, router_indices = outputs
+        router_scores = torch.zeros_like(router_logits).scatter_(1, router_indices, router_scores)
```
ArthurZucker, line 1011: perfect

src/transformers/integrations/tensor_parallel.py
```diff
@@ -1116,69 +1149,148 @@ class ParallelInterface(GeneralInterface):
...
+    plan_to_bias_dim = {
+        "colwise": -1,
+        "colwise_gather_output": -1,
+        "packed_colwise": -1,
+        "rowwise": None,
+        "rowwise_split_input": None,
+        "packed_rowwise": None,
+        "embedding_rowwise": None,
+        "sequence_parallel": None,
+    }
```
ArthurZucker: these need to be external, people are supposed to register custom TensorParallel stuff, as such it should probably be `ALL_PARALLEL.plan_to_bias_dim` something like that, gets updated

tests/tensor_parallel/test_tensor_parallel.py
```diff
@@ -212,20 +234,22 @@ def test_tp_plan_none_handling(self):
...
 # ====== TEST FUNCTIONS ======
-def _test_model_dense_forward_impl(rank, mode):
+def _test_model_dense_forward_impl(rank, mode, dtype=torch.float32):
     """Implementation for comparing TP and non-TP model outputs."""
-    model_id = "JackFram/llama-68m"
+    model_id = "hf-internal-testing/tiny-random-LlamaForCausalLM"
 
-    # Ensure same random seed for reproducibility
     torch.manual_seed(0)
+    torch.cuda.manual_seed_all(0)
```
ArthurZucker: are we not CPU running this? Should import from transformers set seed to be good

tests/tensor_parallel/test_tensor_parallel.py
```diff
@@ -466,13 +582,411 @@ def test_model_dense_save(self):
...
+        result_dir = f"{tmp_dir}/tp"
+    else:
+        kwargs = {}
+        result_dir = f"{tmp_dir}/nontp"
+
+    model = AutoModelForCausalLM.from_pretrained(model_id, dtype="auto", **kwargs)
+    model.save_pretrained(result_dir)
+
+
+class TestTensorParallelMoeBase(TestCasePlus):
```
ArthurZucker: very very nice

ArthurZucker: you could use parametrized? But I don't really mind it like that

ArthurZucker (commented):
Great work! 
My main concern is on the dynamic weight loading and the TP materialize tthat does not take the device. Might be in the sharding method? but in anycase we should never have to move the param in a non async manner (in set param)

src/transformers/integrations/tensor_parallel.py
```diff
@@ -424,6 +426,208 @@ def get_tensor_shard(param, empty_param, device_mesh, rank, dim, tensor_idx: int
...
+class _AllReduceBackward(torch.autograd.Function):
+    """Identity forward, all-reduce backward. Used before colwise layers (f in Megatron)."""
+
+    @staticmethod
+    def forward(ctx, x, device_mesh):
+        ctx.device_mesh = device_mesh
+        return x
+
+    @staticmethod
+    @torch._dynamo.disable
```
Cyrilvallez: Note that this decorator is simply a graph break, so breaks whenever compiling full graph!! This is a huge limitation, not sure if it's such a good idea

src/transformers/integrations/tensor_parallel.py
```diff
@@ -434,557 +638,335 @@ def distribute_module(
...
-    def partition_tensor(self, param: torch.Tensor, dtype, to_contiguous: bool):
+        self,
+        param,
+        param_type=None,
+        param_casting_dtype=None,
+        to_contiguous=None,
+        rank=None,
+        device_mesh=None,
+        tensor_idx=None,
+    ):
```
Cyrilvallez: We should not need all these arguments I believe, see the current signature which is much simpler - also, let's keep simpler names as iscurrently the case, i.e. `param_casting_dtype` -> `dtype`

src/transformers/integrations/tensor_parallel.py
```diff
@@ -434,557 +638,335 @@ def distribute_module(
...
+        device_mesh = self.device_mesh
+        empty_param = self.empty_param
+        rank = self.rank
+        if param_type == "bias":
+            parameter = get_tensor_shard(param, empty_param, device_mesh, rank, -1, tensor_idx)
+        else:
+            parameter = get_tensor_shard(param, empty_param, device_mesh, rank, -2, tensor_idx)
+        parameter = parameter.to(param_casting_dtype)
+        return parameter, None
```
Cyrilvallez: Same here, looks like you are behind what's on main and trying to revert to previous version! `shard_tensor` functions were simplified quite a bit!

src/transformers/integrations/tensor_parallel.py
```diff
@@ -434,557 +638,335 @@ def distribute_module(
...
+    ):
+        device_mesh = device_mesh or self.device_mesh
+        empty_param = self.empty_param
+        rank = rank if rank is not None else self.rank
+        if param_type == "bias":
+            parameter = param[...]
+        else:
+            parameter = get_tensor_shard(param, empty_param, device_mesh, rank, -1, tensor_idx=tensor_idx)
+        parameter = parameter.to(param_casting_dtype)
+        return parameter, None
```
Cyrilvallez: Same here, need to be simplified and synced with main! Another example of why it was simplified, it returns None everywhere

src/transformers/core_model_loading.py
```diff
@@ -826,13 +823,14 @@ def _job():
 
 
 def spawn_tp_materialize(
-    thread_pool: ThreadPoolExecutor | None, tensor: torch.Tensor, sharding_method, tensor_idx, device=None, dtype=None
+    thread_pool: ThreadPoolExecutor | None, tensor: torch.Tensor, sharding_method, tensor_idx, dtype=None
```
Cyrilvallez: Why remove the device here? It should be moved as early as possible

src/transformers/core_model_loading.py
```diff
@@ -915,21 +913,21 @@ def set_param_for_module(
...
-                if getattr(distributed_operation, "use_dtensor", False):
-                    param_value = DTensor.from_local(
-                        param_value,
-                        distributed_operation.device_mesh,
-                        getattr(distributed_operation, "shard", Replicate()),
-                        run_check=False,
-                        shape=ref.size(),
-                        stride=ref.stride(),
-                    )
+                device_mesh = distributed_operation.device_mesh
```
Cyrilvallez: Same as Arthur, this is too late to move to device, let's keep it early just after it was splitted!

Cyrilvallez (commented):
Nice! 
However, looks like you are basing your work on a quite older version of the most important files!! I simplified all the `shard_tensor` functions recently, as well as moving to device much earlier for big wins, and you are reverting those! Extremely important to sync with main on those aspects!
Also, `disable` is a decorator that forces a graph break, so by definition is very dangerous as compiling fullgraph will crash immediately (as the graph break is thus not possible)

vasqu:
A little offside: Can you also add the plan for ernie vl moe? It might be more of a special case since it has two sets of Moes :eyes:

tests/tensor_parallel/test_tensor_parallel.py
3outeille: i thought there was argument to not use `pytest.parametrize` . I am down changing it to parameterize

tests/tensor_parallel/test_tensor_parallel.py
3outeille: fixed

src/transformers/integrations/tensor_parallel.py
3outeille: moved it inside ALL_PARALLEL_STYLES (ParallelInterface)

src/transformers/integrations/tensor_parallel.py
3outeille: done

src/transformers/core_model_loading.py
3outeille: see [d6cb145](https://github.com/huggingface/transformers/pull/42809/commits/d6cb145416cf0eee60f615336071eecc77657dc7)

src/transformers/core_model_loading.py
3outeille: see [d6cb145](https://github.com/huggingface/transformers/pull/42809/commits/d6cb145416cf0eee60f615336071eecc77657dc7)

src/transformers/core_model_loading.py
3outeille: see [d6cb145](https://github.com/huggingface/transformers/pull/42809/commits/d6cb145416cf0eee60f615336071eecc77657dc7)

3outeille:
run-slow: afmoe, apertus, arcee, aria, bamba, cohere, cohere2, cwm, dbrx, deepseek_v2, deepseek_v3, diffllama, doge, dots1, emu3, ernie4_5

3outeille:
run-slow: afmoe, apertus, arcee, aria, bamba, cohere, cohere2, cwm, dbrx, deepseek_v2, deepseek_v3, diffllama, doge, dots1, emu3, ernie4_5

src/transformers/core_model_loading.py
```diff
@@ -886,23 +883,20 @@ def set_param_for_module(
...
                 param_value = torch.nn.Parameter(param_value, requires_grad=param_value.is_floating_point())
 
         # Remove from missing keys (it's either mismatched, or all good)
         missing_keys.discard(target_name)
-        if ref is not None and ref.shape != param_value.shape and hf_quantizer is None:
-            mismatch_keys.add((target_name, param_value.shape, ref.shape))
+
+        # Determine expected shape: for TP, use sharded shape; otherwise, use full shape
+        if distributed_operation is not None:
+            expected_shape = torch.Size(distributed_operation.get_expected_sharded_shape(tuple(ref.shape), param_name))
```
ArthurZucker: nice, let's hide as much as possible to only have to call `distributed_operation.get_expected_sharded_shape(ref.shape)`

src/transformers/integrations/tensor_parallel.py
```diff
@@ -434,224 +633,183 @@ def distribute_module(
...
     """
-    This class is used to isolate computation in a TP layer from the rest of the world.
-    Parameters need to be LOCAL, so not dtensors
+    Row-wise parallel: weight is sharded on dim -1 (input features).
+    Forward: input (optionally split) -> output partial -> all-reduce to replicate.
+
+    Args:
+        split_input: If True, splits replicated input before matmul. Use when input
+                     comes from a non-parallelizable operation (chunk/slice).
+                     Default False (expects pre-sharded input from colwise layer).
```
ArthurZucker, line 740: Okay I am not certain here, QOL might be better to always split input if shape is not what we expect (which dtensor did on its own()

src/transformers/integrations/tensor_parallel.py
```diff
@@ -660,331 +818,144 @@ def shard_tensor(
...
-        # 1. to replicate -> allreduce
-        # 2. to shard -> reduce_scatter
-        if outputs.placements != output_layouts:
-            outputs = outputs.redistribute(placements=output_layouts, async_op=True)
-        outputs = outputs.to_local()  # otherwise the `+=` op will gather
-        if hasattr(mod, "_bias"):
-            outputs = outputs + mod._bias
-        # back to local tensor if use_local_output is True
-        return outputs
+        return parameter.to(device=device, dtype=dtype)
```
ArthurZucker, line 885: I think `get_tensor_shard` should use with device already and thus we should not need to move here no?

src/transformers/integrations/tensor_parallel.py
```diff
@@ -1116,69 +1122,152 @@ class ParallelInterface(GeneralInterface):
...
+    plan_to_weight_dim: dict[str, int | None] = {
+        "colwise": -2,
+        "colwise_gather_output": -2,
+        "packed_colwise": -2,
+        "rowwise": -1,
+        "rowwise_split_input": -1,
+        "packed_rowwise": -1,
+        "embedding_rowwise": 0,
+        "sequence_parallel": None,
+    }
```
ArthurZucker, line 1102: can we add repetition versions as well? colwise rep for example

ArthurZucker (approved):
Very nice!
Thanks for taking all the feedback into account! Let's goooooo

src/transformers/integrations/tensor_parallel.py
```diff
@@ -434,224 +633,183 @@ def distribute_module(
...
+        return parameter.to(device=device, dtype=dtype)
 
     def prepare_module_tp(self, module: nn.Module, device_mesh) -> nn.Module:
         distribute_module(
             module,
             device_mesh,
-            partial(self._prepare_input_fn, None, None),
-            partial(self._prepare_output_fn, None, None),
+            self._prepare_input_fn,
+            self._prepare_output_fn,
```
Cyrilvallez: We can fully drop this function as it's the same as the parent!

src/transformers/integrations/tensor_parallel.py
```diff
@@ -434,224 +633,183 @@ def distribute_module(
...
 
     def prepare_module_tp(self, module: nn.Module, device_mesh) -> nn.Module:
+        module._distribute_module_applied = True
         distribute_module(
             module,
             device_mesh,
-            partial(self._prepare_input_fn, None, None),
-            partial(self._prepare_output_fn, None, None),
+            self._prepare_input_fn,
+            self._prepare_output_fn,
```
Cyrilvallez: Let's just call `super()` after here if we just want to add the internal attr compared to parent!

src/transformers/integrations/tensor_parallel.py
Cyrilvallez: IMO maybe very slightly more clear to keep `get_tensor_shard` only for the splitting and then move `device`/`dtype` here to avoid overloading the other, but no strong opinions at all, both work

src/transformers/integrations/tensor_parallel.py
```diff
@@ -660,331 +818,144 @@ def shard_tensor(
...
-            outputs = outputs.redistribute(placements=output_layouts, async_op=True)
-        outputs = outputs.to_local()  # otherwise the `+=` op will gather
-        if hasattr(mod, "_bias"):
-            outputs = outputs + mod._bias
-        # back to local tensor if use_local_output is True
-        return outputs
+        return parameter.to(device=device, dtype=dtype)
 
     def prepare_module_tp(self, module: nn.Module, device_mesh) -> nn.Module:
         module._distribute_module_applied = True
```
Cyrilvallez: Same and seem valid everywhere, we don't need to redefine everywhere!

src/transformers/integrations/tensor_parallel.py
Cyrilvallez: I think it's now called `gather_output` instead of `rep`, is my understanding correct @3outeille?

Cyrilvallez (approved):
All right! Much better version, thanks a lot for taking all the changes into account! Great work! 🤗🚀
Hard to make sure everything will work perfectly by simply reading distributed code though - could you notably confirm that chaining several "matching" layers, such as a MLP which contains 2 colwise and 1 rowwise will only incur a communication at the very end, after the final rowwise?
Otherwise, main comment is to remove all the redundant `prepare_module_tp` as they are all similar!

src/transformers/integrations/tensor_parallel.py
ArthurZucker: Ah IDK, we need a simple gatther ouptut as well

3outeille: it's  `colwise_gather_output` now

3outeille:
> could you notably confirm that chaining several "matching" layers, such as a MLP which contains 2 colwise and 1 rowwise will only incur a communication at the very end, after the final rowwise?

yes !

PR #42812, opened 2025-12-11
Fix deepspeed sp loss due to missing labels 

SunMarc:
# What does this PR do?

This PR fixes `_deepspeed_sp_compute_loss`. [UlyssesSPDataLoaderAdapter](https://github.com/deepspeedai/DeepSpeed/blob/6eb98aaf03419d1a3a608b0955eb3eae5ad2dc27/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L598) took `labels` out from `inputs` while inserting `shift_labels` so it doesn't compute the `loss` at all since we didn't pass `labels` ...

Fixes https://github.com/huggingface/transformers/pull/42444#issuecomment-3639642523


kashif:
currently looking at deepspeed side

kashif (approved):

PR #42813, opened 2025-12-11
update deprecation msg for `warmup_ratio`

SunMarc:
# What does this PR do?

This PR updates the deprecation msg for `warmup_ratio` to have it less confusing.

### DIFF
--- a/src/transformers/training_args.py
+++ b/src/transformers/training_args.py
@@ -860,7 +860,7 @@ class TrainingArguments:
     warmup_ratio: float | None = field(
         default=None,
         metadata={
-            "help": "This argument is deprecated and will be removed in v5. Use `warmup_steps` instead as it also works with float values."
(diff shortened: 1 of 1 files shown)


MekkCyber (approved):
Thanks

src/transformers/training_args.py
```diff
@@ -860,7 +860,7 @@ class TrainingArguments:
     warmup_ratio: float | None = field(
         default=None,
         metadata={
-            "help": "This argument is deprecated and will be removed in v5. Use `warmup_steps` instead as it also works with float values."
+            "help": "This argument is deprecated and will be removed in v5.2 . Use `warmup_steps` instead as it also works with float values."
```
BenjaminBossan:
```suggestion
            "help": "This argument is deprecated and will be removed in v5.2. Use `warmup_steps` instead as it also works with float values."
```

PR #42816, opened 2025-12-11
validate tokenizer components

itazap:
check if the tokenizer object in `tokenizer.json` that is being loaded is actually matching that of mapped tokenizer in TOKENIZER_MAPPINGS in `tokenization_auto.py`


src/transformers/tokenization_utils_tokenizers.py
```diff
@@ -54,6 +57,73 @@
 TOKENIZER_CONFIG_FILE = "tokenizer_config.json"
 TIKTOKEN_VOCAB_FILE = "tokenizer.model"
 
+
+def _validate_tokenizer_components(tokenizer_class, tokenizer_json_path):
```
ArthurZucker:
```suggestion
def _validate_tokenizer_components(tokenizer_instance tokenizer_json_path):
```
will do ourselves a favor IMO to have an instance

src/transformers/tokenization_utils_tokenizers.py
```diff
@@ -54,6 +57,73 @@
...
+        
+        expected_components = {
+            "normalizer": _get_component_type(tokenizer.normalizer),
+            "pre_tokenizer": _get_component_type(tokenizer.pre_tokenizer),
+            "decoder": _get_component_type(tokenizer.decoder),
+            "model": _get_component_type(tokenizer.model),
+        }
+        
+        # Compare and warn on mismatches
+        mismatches = []
```
ArthurZucker, line 103:
```suggestion
        mismatches = set(expected_components) - set(json_components)
```

src/transformers/tokenization_utils_tokenizers.py
```diff
@@ -54,6 +57,73 @@
...
+        for name in ["normalizer", "pre_tokenizer", "decoder", "model"]:
+            json_val = json_components[name]
+            expected_val = expected_components[name]
+            if json_val != expected_val:
+                mismatches.append(f"{name}: expected {expected_val}, found {json_val}")
```
ArthurZucker, line 108:
```suggestion
```

src/transformers/tokenization_utils_tokenizers.py
```diff
@@ -240,6 +311,7 @@ def __init__(self, *args, **kwargs):
         if tokenizer_object is not None:
             fast_tokenizer = copy.deepcopy(tokenizer_object)
         elif fast_tokenizer_file is not None and os.path.isfile(fast_tokenizer_file):
+            _validate_tokenizer_components(self.__class__, fast_tokenizer_file)
```
ArthurZucker: No validate should go after INIT, most probably in the `convert_to_native_format` function

ArthurZucker (commented):
ty

src/transformers/tokenization_utils_tokenizers.py
itazap: how do we check if they match?

ArthurZucker: with m y above:
```python
 set(expected_components) - set(json_components)
```

PR #42818, opened 2025-12-11
Update `param_element_size`

SunMarc:
# What does this PR do?

This PR fixes how we calculate the param size for quantized models. This should be simpler to hack around !

I added some tests to check that it works for the following methods `bnb`, `finegrained_fp8`, `torchao`, `mxfp4`, `quanto`. 

cc @Cyrilvallez as you were concerned at some point.


src/transformers/modeling_utils.py
```diff
@@ -4411,8 +4411,7 @@ def _move_missing_keys_from_meta_to_cpu(
             # Buffers are not initialized on the meta device, so we still need this check to avoid overwriting them
             if param.device == torch.device("meta"):
                 value = torch.empty_like(param, dtype=dtype, device="cpu")
-                if not is_quantized or not hf_quantizer.param_needs_quantization(self, key):
-                    _load_parameter_into_model(self, key, value)
```
SunMarc, line 4402: it should be fine to remove the condition no ? cc @Cyrilvallez

src/transformers/quantizers/base.py
```diff
@@ -118,33 +118,7 @@ def update_device_map(self, device_map: dict[str, Any] | None) -> dict[str, Any]
...
-            mapping = {
-                torch.int8: 1,
-                CustomDtype.INT4: 0.5,
-                CustomDtype.FP8: 1,
-                CustomDtype.INT2: 0.25,
-            }
-            # The value passed is actually not used when the method is overridden
-            if (custom_dtype := self.adjust_target_dtype(torch.float16)) in mapping:
-                return mapping[custom_dtype]
         return param.element_size()
```
MekkCyber, line 112:
```suggestion
    def param_element_size(self, model: "PreTrainedModel", param_name: str, param: "torch.Tensor") -> float:
        "Return the element size (in bytes) for `param_name`."
        return param.element_size()
```

MekkCyber: just adding the docstring back

MekkCyber (approved):
Awesome 🧼

src/transformers/modeling_utils.py
Cyrilvallez: Well, yes and no... Previously there was a 
```python
else:
    hf_quantizer.create_quantized_param(...)
```
in case the missing weight needed to be quantized (we quantized the new random weight).

I'm not sure why it was removed, but it should probably be the best way no?

src/transformers/modeling_utils.py
```diff
@@ -4593,21 +4592,18 @@ def caching_allocator_warmup(model: PreTrainedModel, expanded_device_map: dict,
...
-        if hf_quantizer is not None:
-            param_name = hf_quantizer.get_param_name(param_name)
-
         try:
             param = model.get_parameter_or_buffer(param_name)
         except AttributeError:
-            # TODO: for now let's skip if we can't find the parameters
-            if hf_quantizer is not None:
-                continue
             raise AttributeError(f"Parameter {param_name} not found in model")
```
Cyrilvallez: Let's remove the try/except block if we don't do anything else than re-raising!

src/transformers/quantizers/quantizer_bnb_4bit.py
```diff
@@ -51,15 +51,6 @@ class Bnb4BitHfQuantizer(HfQuantizer):
...
         super().__init__(quantization_config, **kwargs)
 
-        # This describes the additional items that are saved on the state dict (on the params themselves)
-        self.bnb_keys = [
-            f"quant_state.bitsandbytes__{self.quantization_config.bnb_4bit_quant_type}",
-            "absmax",
-            "quant_map",
-        ]
-        if self.quantization_config.bnb_4bit_use_double_quant:
-            self.bnb_keys.extend(["nested_absmax", "nested_quant_map"])
```
Cyrilvallez, line 61: Not needed anymore since new loading?

Cyrilvallez (commented):
SUPER NICE! 🤗 Thanks a lot for doing that! High time we finally have a nice and clean way to pre-allocate with quantization! 🚀
Just added a few comments, then we can merge!

src/transformers/modeling_utils.py
SunMarc: yes !

src/transformers/quantizers/quantizer_bnb_4bit.py
SunMarc: not needed anymore !

src/transformers/modeling_utils.py
SunMarc: We removed `create_quantized_param` as we didn't need to anymore since we have the quantize ops xD. Let's see in another how to deal with that in another PR cc @MekkCyber but I don't think this urgent.

src/transformers/modeling_utils.py
```diff
@@ -4520,6 +4521,41 @@ def is_accelerator_device(device: Union[str, int, torch.device]) -> bool:
...
+    tied_param_names = model.all_tied_weights_keys.keys()
+
+    tp_plan = getattr(model, "_tp_plan", []) or []
+    tp_plan_regex = (
+        re.compile("|".join([re.escape(plan) for plan in tp_plan]))
+        if _torch_distributed_available and torch.distributed.is_initialized()
+        else None
+    )
+
+    modules_sizes, _ = compute_module_sizes(model, hf_quantizer, only_modules=False)
```
SunMarc: using this instead cc @Cyrilvallez if that's fine

src/transformers/modeling_utils.py
```diff
@@ -4036,7 +4036,9 @@ def _load_pretrained_model(
         }
 
         # Model's definition arriving here is final (TP hooks added, quantized layers replaces)
-        expected_keys = list(model.state_dict().keys())
+        # We need parameters + buffers here, as state_dict does not count non-persistent buffers which are taking space
+        expected_keys = [name for name, _ in model.named_parameters()] + [name for name, _ in model.named_buffers()]
+
```
SunMarc: there were some differences because of this

tests/quantization/bnb/test_mixed_int8.py
```diff
@@ -428,6 +428,55 @@ def test_int8_from_pretrained(self):
...
+            ]
+
+        for name, module in model.named_modules():
+            if isinstance(module, bnb.nn.Linear8bitLt):
+                # from 16 bits to 8 bits
+                assert int(model_size[f"{name}.weight"] // 2) == int(quantized_model_size[f"{name}.weight"])
+
+        # check that we get the same value, as we use `compute_module_sizes` in `get_total_byte_count`
+        assert total_byte_count == model_size[""]
+        assert quantized_total_byte_count == quantized_model_size[""]
```
SunMarc, line 467: here i check that we have the same result

Cyrilvallez:
Will have a last look next Monday!

src/transformers/modeling_utils.py
```diff
@@ -4036,7 +4036,9 @@ def _load_pretrained_model(
         }
 
         # Model's definition arriving here is final (TP hooks added, quantized layers replaces)
-        expected_keys = list(model.state_dict().keys())
+        # We need parameters + buffers here, as state_dict does not count non-persistent buffers which are taking space
+        expected_keys = [name for name, _ in model.named_parameters()] + [name for name, _ in model.named_buffers()]
```
Cyrilvallez: Humm, the fact here is that non-persistent buffers are NOT loaded with the other params (because they are non-persistent of course), so it is not needed to account for them when allocating memory before loading
Thus I believe we should revert this change

src/transformers/modeling_utils.py
```diff
@@ -4520,6 +4521,41 @@ def is_accelerator_device(device: Union[str, int, torch.device]) -> bool:
...
+
+    tp_plan = getattr(model, "_tp_plan", []) or []
+    tp_plan_regex = (
+        re.compile("|".join([re.escape(plan) for plan in tp_plan]))
+        if _torch_distributed_available and torch.distributed.is_initialized()
+        else None
+    )
+
+    modules_sizes, _ = compute_module_sizes(model, hf_quantizer, only_modules=False)
+    for param_name, device in accelerator_device_map.items():
```
Cyrilvallez: Humm, here we iterate twice over all params for no reason... Better to go back to the old loop and mimic what's being done in `compute_module_sizes` by using `dtype_size = hf_quantizer.param_element_size(model, name, param)` if we have a quantizer!

SunMarc: fair, sounds good !

src/transformers/modeling_utils.py
SunMarc: indeed xD I was only trying to match the numbers and didn't think too far. In any case, the tests I wrote should still be valid !

src/transformers/modeling_utils.py
```diff
@@ -4520,6 +4520,46 @@ def is_accelerator_device(device: Union[str, int, torch.device]) -> bool:
...
+            dtype_size = param.element_size()
+
+        param_byte_count = param.numel() * dtype_size
+
+        if tp_plan_regex is not None:
+            generic_name = re.sub(r"\.\d+\.", ".*.", param_name)
+            param_byte_count //= torch.distributed.get_world_size() if tp_plan_regex.search(generic_name) else 1
+
+        total_byte_count[device] += param_byte_count
+    return total_byte_count
```
Cyrilvallez, line 4579: nit: if you don't mind, I find that it's easier to follow if everything is inside the same function in this case, so IMO I would put it back in `caching_allocator_warmup`
No super strong opinions here though, so if you think you'll ever need it elsewhere we can keep separate

Cyrilvallez (approved):
LGTM! Left a final comment but that's it!
Feel free to merge after the conflicts on quantizers have been resolved 🤗
Thanks again for this, will make everything much much smoother! 🤗

src/transformers/modeling_utils.py
SunMarc: I had a do this  as I need to to test that we get the correct allocation ;D

PR #42819, opened 2025-12-11
[kernels] Centralize kernels tests

MekkCyber:
# What does this PR do?

Merges the two kernel tests files into one file, and remove redundant tests


SunMarc (approved):
Ty

PR #42820, opened 2025-12-11
[`T5Gemma2`] Fix bidirectional mask for encoder

vasqu:
Wrong `is_causal` flag causes unexpected behavior where when we shortcut mask creation (all True) we trigger SDPA's `is_causal=True` behavior.


ArthurZucker (approved):
thanks!

PR #42824, opened 2025-12-11
Fix torch only support for fast Processors

Blaizzy:
# What does this PR do?

Adds MLX tensor support to `BatchFeature` and removes the PyTorch-only restriction in fast image processing utilities.

**Changes:**
- Added `is_mlx_array` export to `transformers/utils/__init__.py`
- Implemented MLX tensor conversion logic in `BatchFeature.convert_to_tensors()` (`feature_extraction_utils.py`)
- Removed the hardcoded validation that restricted fast image processors to PyTorch tensors only (`image_processing_utils_fast.py`)

This enables using `return_tensors="mlx"` with vision models and fast image processors on Apple Silicon, which previously raised:
```
ValueError: Only returning PyTorch tensors is currently supported.
```

## Before submitting
- [x] Did you read the [[contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request)](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [[forum](https://discuss.huggingface.co/)](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
      [[documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs)](https://github.com/huggingface/transformers/tree/main/docs), and
      [[here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation)](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).

## Who can review?

- @yonigozlan @molbap (vision models / image processing)
- @ArthurZucker (library / model loading)


Rocketknight1:
cc @yonigozlan @molbap but not sure who to tag for MLX!

Blaizzy:
@ArthurZucker has some context as well

ArthurZucker (approved):
LGTM, @yonigozlan is it fine for you as well?

src/transformers/feature_extraction_utils.py
```diff
@@ -141,6 +143,30 @@ def as_tensor(value):
...
+                    elif (
+                        isinstance(value[0], (list, tuple))
+                        and len(value[0]) > 0
+                        and isinstance(value[0][0], np.ndarray)
+                    ):
+                        value = np.array(value)
+                if isinstance(value, np.ndarray):
+                    return mx.array(value)
+                else:
+                    return mx.array(value)
```
yonigozlan:
```suggestion
                else:
                    return mx.array(value)
```
This should be equivalent I think?

tests/models/pop2piano/test_feature_extraction_pop2piano.py
```diff
@@ -50,6 +50,9 @@ class Pop2PianoFeatureExtractionTester:
     def __init__(
```
yonigozlan, line 50: Not sure if these modifications are related to this PR or if they were added here by mistake?

yonigozlan (approved):
Hello @Blaizzy ! Sorry for the delay. I'm ok with merging this if it can improve compatibility with mlx! We don't have any mlx image processor in the library for now, but I'm guessing this is to use with custom image processors?
As discussed with Arthur, we'll move towards having different backends for image processors, so we could add one for mlx. However image processors are quite different across models, so we would need to have custom ones for each backend and models, as making image processing operation not library dependent seems difficult. Anyway these are issues for another PR ;)

Blaizzy:
> We don't have any mlx image processor in the library for now, but I'm guessing this is to use with custom image processors

Thank you very much @yonigozlan! Yes, that will be the path I'll take in the meantime. Because for majority of models they can return mlx tensors from transformers but a few use torch and torch vision explicitly, that forces torch as dep.

tests/models/pop2piano/test_feature_extraction_pop2piano.py
Blaizzy: They were a mistake 😅. I was following the formatter feedback.

src/transformers/feature_extraction_utils.py
Blaizzy: Yes, it is! Thanks for catching it

Blaizzy:
```suggestion
                return mx.array(value)
```

src/transformers/feature_extraction_utils.py
```diff
@@ -141,6 +143,28 @@ def as_tensor(value):
...
+                    if isinstance(value[0], np.ndarray):
+                        value = np.array(value)
+                    elif (
+                        isinstance(value[0], (list, tuple))
+                        and len(value[0]) > 0
+                        and isinstance(value[0][0], np.ndarray)
+                    ):
+                        value = np.array(value)
+                else:
+                    return mx.array(value)
```
Blaizzy:
```suggestion
                return mx.array(value)
```

yonigozlan:
Hey @Blaizzy ! I fixed  some tiny formatting errors and moved the test to `test_feature_extraction_utils.py‎`. It would be great to get this merged soon as we're making progress on the PR to introduce image processing backends ;) https://github.com/huggingface/transformers/pull/43514

However, I can't use mlx on my machine, and we don't have mlx support in our CI. Would you mind running the test (`RUN_SLOW=1 python3 -m pytest tests/utils/test_feature_extraction_utils.py -k test_batch_feature_mlx_conversion`) to make sure it passes?
Thanks!

ArthurZucker (approved):

PR #42825, opened 2025-12-11
Simplify dtype instantiation

Cyrilvallez:
# What does this PR do?

As per the title. Cleanu-up a bit after https://github.com/huggingface/transformers/pull/42805 and in general

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -696,22 +696,19 @@ def _get_resolved_checkpoint_files(
 
 
 def _get_dtype(
-    cls,
     dtype: Optional[Union[str, torch.dtype, dict]],
     checkpoint_files: Optional[list[str]],
     config: PreTrainedConfig,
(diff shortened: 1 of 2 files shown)


ArthurZucker (commented):
```python

    @property
    def dtype(self) -> torch.dtype:
        """
        `torch.dtype`: The dtype of the module (assuming that all the module parameters have the same dtype).
        """
        dtype = self._dtype or next(param.dtype for param in self.parameters() if param.is_floating_point())
        if isinstance(dtype, str):
            if hasattr(torch, dtype):
                dtype = getattr(torch, dtype)
        return dtype
```
you did not change this one either no

Cyrilvallez:
I did, it's [here](https://github.com/huggingface/transformers/pull/42825/files#diff-6b72b98c4c2dcfc6cc606843917733f5d858374fbc22a735ff483bbc0c1e63eaL801-R810) haha

ArthurZucker (approved):

PR #42827, opened 2025-12-12
[docs] optimization cleanup

stevhliu:
cleans up the Optimization section a bit to remove redundant content

- removes the hardware specific sections (GPU/CPU) since many of the same optimizations in these docs already exist
- give Optimum it's own section in the docs for more visibility and combines GPU/CPU usage for it
- give assisted decoding it's own section in the docs for more visibility as well and it's also not really a decoding strategy like the other search/sampling methods are
- also removes the [Optimizing inference](https://huggingface.co/docs/transformers/llm_optims) doc since it also has redundant content in it and will be replaced by the newer overview in #42538


Cyrilvallez (approved):
LGTM!

PR #42830, opened 2025-12-12
Fix speccht5_tts pipeline

jiqing-feng:
Fix #42792

### DIFF
--- a/src/transformers/pipelines/text_to_audio.py
+++ b/src/transformers/pipelines/text_to_audio.py
@@ -118,7 +118,7 @@ class TextToAudioPipeline(Pipeline):
             )
 
         if self.model.config.model_type in ["musicgen", "speecht5"]:
-            # MusicGen expect to use the tokenizer
+            # MusicGen and SpeechT5 expect to use their tokenizer instead
             self.processor = None
 
         self.sampling_rate = sampling_rate
--- a/src/transformers/pipelines/text_to_audio.py
+++ b/src/transformers/pipelines/text_to_audio.py
@@ -117,7 +117,7 @@ class TextToAudioPipeline(Pipeline):
                 else vocoder
             )
 
-        if self.model.config.model_type in ["musicgen"]:
+        if self.model.config.model_type in ["musicgen", "speecht5"]:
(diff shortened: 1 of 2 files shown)


src/transformers/pipelines/text_to_audio.py
```diff
@@ -117,7 +117,7 @@ def __init__(self, *args, vocoder=None, sampling_rate=None, **kwargs):
                 else vocoder
             )
 
-        if self.model.config.model_type in ["musicgen"]:
+        if self.model.config.model_type in ["musicgen", "speecht5"]:
             # MusicGen expect to use the tokenizer
```
vasqu:
```suggestion
        if self.model.config.model_type in ["musicgen", "speecht5"]:
            # MusicGen and SpeechT5 expect to use their tokenizer instead
```

tests/pipelines/test_pipelines_text_to_audio.py
```diff
@@ -40,6 +41,38 @@ class TextToAudioPipelineTests(unittest.TestCase):
     model_mapping = MODEL_FOR_TEXT_TO_WAVEFORM_MAPPING
     # for now only test text_to_waveform and not text_to_spectrogram
 
+    @require_torch
+    def test_small_speecht5_pt(self):
```
vasqu, line 45: Not for this PR but might be nice to refactor the tests a bit to unify what we can :eyes:

vasqu (approved):
LGTM, just that nit

src/transformers/pipelines/text_to_audio.py
ebezzam: or perhaps changing the comment to:
```
# models that expect to use their tokenizer instead
```
in case other models come up 😅

ebezzam (approved):
Thanks @jiqing-feng for the PR!

And just to confirm that your example from #42792 now works with this fix?

src/transformers/pipelines/text_to_audio.py
vasqu: Yup fair enough

PR #42838, opened 2025-12-12
Do not rely on config for inferring model dtype

Cyrilvallez:
# What does this PR do?

As per the title. Back to what it was before https://github.com/huggingface/transformers/pull/42805. Relying on the config is an issue in some cases, see https://github.com/huggingface/transformers/pull/42805#issuecomment-3646172797


BenjaminBossan:
Thanks

tomaarsen:
Thanks for the fix, this also e.g. fixes `model.half()` saving as `"dtype": "float32"` via `model.save_pretrained(...)` in `5.0.0rc01`, which broke some tests in Sentence Transformers.

PR #42842, opened 2025-12-12
Pass in position_ids into attention for GPT2

jackzhxng:
# What does this PR do?

Pass in `position_ids` so that the custom attention interface implementations have access to it.

## Who can review?

@ArthurZucker @Cyrilvallez

### DIFF
--- a/src/transformers/models/gpt2/modeling_gpt2.py
+++ b/src/transformers/models/gpt2/modeling_gpt2.py
@@ -710,6 +710,7 @@ class GPT2Model(GPT2PreTrainedModel):
                 encoder_attention_mask=encoder_attention_mask,
                 use_cache=use_cache,
                 output_attentions=output_attentions,
+                position_ids=position_ids,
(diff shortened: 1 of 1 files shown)


vasqu (approved):
We have a test for this but it is skipped with GPT2 currently, i.e. check if paddingfree is the same as padding which relies exactly on this. I have a PR for this over here #42848

The logic is sound tho, depending on the other PR, it might make more sense to add an explicit test here

jackzhxng:
Sounds good, lmk if you want the test here / feel free to merge

vasqu:
Could you add a small test here directly? My other PR is going a bit beyond what I had originally planned

ArthurZucker (commented):
Indeed let's add a test!

vasqu:
#42848 is now merged and includes the passing of the pos ids (along testing) so I'm closing this one - feel free to reopen or tag me if it does not work as expected

PR #42853, opened 2025-12-12
[docs] serving

stevhliu:
splits the serving docs from #42263


docs/source/en/serving.md
```diff
@@ -364,67 +336,192 @@ ResponseCompletedEvent(response=Response(id='resp_req_0', created_at=1754060400.
...
+
+## Port forwarding
+
+The `transformers serve` server supports port forwarding. This lets you serve models from a remote server. Make sure you have ssh access from your device to the server. Run the following command on your device to set up port forwarding.
+
+```bash
+ssh -N -f -L 8000:localhost:8000 your_server_account@your_server_IP -p port_to_ssh_into_your_server
+```
+
+## Serving optimizations
```
stevhliu: i think we can create a separate doc for the serving optimizations, wdyt @LysandreJik?

LysandreJik (commented):
I'd encourage this section to remain split, if possible -> I think we'll want to showcase quite a few integrations and it'll be simpler for users to click on the integration they want rather than seeing a dump of possible integrations.

I'm quite down for a separate page for CLI optimizations though, as it goes with that spirit :)

docs/source/en/serving.md
LysandreJik: Yes!

LysandreJik (approved):
Thank you @stevhliu, let's merge it!

PR #42864, opened 2025-12-14
fix: Initialize ApertusMLP's xielu activation using `torch_dtype`

wasertech:
# What does this PR do?

I cannot infer this model architecture on Turing. This fixes the issue at the root.

See #42371, https://github.com/vllm-project/vllm/issues/29349, https://github.com/vllm-project/vllm/pull/30635 and https://huggingface.co/swiss-ai/Apertus-8B-Instruct-2509/discussions/21 for context.

Fixes `RuntimeError: expected m1 and m2 to have the same dtype, but got: float != c10::Half`

Issue reproduction:

```py

import torch
from torch import nn
from transformers.models.apertus.configuration_apertus import ApertusConfig
from transformers.models.apertus.modeling_apertus import ApertusMLP
import sys

def test_apertus_mlp_crash():
    print(f"Python executable: {sys.executable}")
    print("Testing ApertusMLP crash on float16...")
    
    # 1. Setup Config with float16 and xielu
    config = ApertusConfig(
        hidden_size=64,
        intermediate_size=128,
        hidden_act="xielu",
        torch_dtype="float16" # Simulating the runner or user choice
    )
    
    # 2. Instantiate MLP
    # We might need to manually set default dtype if the model doesn't pick it up from config immediately
    # usage in vllm often sets the torch default dtype or casts the model
    torch.set_default_dtype(torch.float16) # Simulating "float16 hardware" env
    
    try:
        mlp = ApertusMLP(config)
        
        # Check act_fn dtype
        if hasattr(mlp.act_fn, 'beta'):
             print(f"XIELU beta dtype: {mlp.act_fn.beta.dtype}")
        
        # 3. Create Input
        x = torch.randn(1, 64, dtype=torch.float16)
        
        # 4. Forward Pass
        output = mlp(x)
        print("Forward pass successful!")
        print(f"Output dtype: {output.dtype}")
        
    except RuntimeError as e:
        print("\nCaught expected RuntimeError:")
        print(e)
    except Exception as e:
        print(f"\nCaught unexpected exception: {type(e)}")
        print(e)
    finally:
         torch.set_default_dtype(torch.float32) # Cleanup

if __name__ == "__main__":
    test_apertus_mlp_crash()

```

Output Without this patch:
```
Python executable: /home/waser/Projets/Transformers/transformers/venv/bin/python
Testing ApertusMLP crash on float16...
CUDA-fused xIELU not available (No module named 'xielu') – falling back to a Python version.
For CUDA xIELU (experimental), `pip install git+https://github.com/nickjbrowning/XIELU`
XIELU beta dtype: torch.bfloat16

Caught expected RuntimeError:
expected m1 and m2 to have the same dtype, but got: float != c10::Half
```

Output with this patch:
```
Python executable: /home/waser/Projets/Transformers/transformers/venv/bin/python
Testing ApertusMLP crash on float16...
CUDA-fused xIELU not available (No module named 'xielu') – falling back to a Python version.
For CUDA xIELU (experimental), `pip install git+https://github.com/nickjbrowning/XIELU`
`torch_dtype` is deprecated! Use `dtype` instead!
XIELU beta dtype: torch.float16
Forward pass successful!
Output dtype: torch.float16
```

Big thanks to everyone who helped to shape this patch into light.

### DIFF
--- a/src/transformers/models/apertus/modeling_apertus.py
+++ b/src/transformers/models/apertus/modeling_apertus.py
@@ -49,6 +49,10 @@ class ApertusMLP(nn.Module):
         self.up_proj = nn.Linear(self.hidden_size, self.intermediate_size, bias=False)
         self.down_proj = nn.Linear(self.intermediate_size, self.hidden_size, bias=False)
         self.act_fn = ACT2FN[config.hidden_act]
+        if config.hidden_act == "xielu":
+            from ...activations import ACT2CLS
+
+            self.act_fn = ACT2CLS["xielu"](dtype=config.dtype)
 
     def forward(self, x):
         return self.down_proj(self.act_fn(self.up_proj(x)))
--- a/src/transformers/models/apertus/modeling_apertus.py
+++ b/src/transformers/models/apertus/modeling_apertus.py
@@ -25,7 +25,7 @@ from typing import Optional, Union
 import torch
 from torch import nn
 
-from ...activations import ACT2FN
+from ...activations import ACT2CLS, ACT2FN
 from ...cache_utils import Cache, DynamicCache
 from ...generation import GenerationMixin
 from ...integrations import use_kernel_forward_from_hub, use_kernel_func_from_hub, use_kernelized_func
@@ -50,8 +50,6 @@ class ApertusMLP(nn.Module):
         self.down_proj = nn.Linear(self.intermediate_size, self.hidden_size, bias=False)
         self.act_fn = ACT2FN[config.hidden_act]
         if config.hidden_act == "xielu":
-            from ...activations import ACT2CLS
-
             self.act_fn = ACT2CLS["xielu"](dtype=config.dtype)
 
     def forward(self, x):
--- a/src/transformers/models/apertus/modular_apertus.py
+++ b/src/transformers/models/apertus/modular_apertus.py
@@ -192,9 +192,13 @@ class ApertusConfig(PreTrainedConfig):
 
 class ApertusMLP(NemotronMLP):
     def __init__(self, config):
-        super().__init__()
+        super().__init__(config)
         self.up_proj = nn.Linear(self.hidden_size, self.intermediate_size, bias=False)
         self.down_proj = nn.Linear(self.intermediate_size, self.hidden_size, bias=False)
+        if config.hidden_act == "xielu":
+            from ...activations import ACT2CLS
+
+            self.act_fn = ACT2CLS["xielu"](dtype=config.dtype)
 
 
 class ApertusRMSNorm(LlamaRMSNorm):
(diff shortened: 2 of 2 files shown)


tests/models/apertus/test_modeling_apertus.py
wasertech, line 1: It may be a bit overkill to do a full forward pass in f16 just for this architecture... I don't need it; even if it's nice to have, the cost is too great for the return. I can remove it. I'm not even sure it would need to be replaced with another test honestly.

wasertech (commented):
Let me remove the test and I'm happy with this PR.

Rocketknight1:
run-slow: apertus

Rocketknight1:
The test runner is very slow so I checked this out locally and the slow tests seem fine, so I'm happy to approve this!

wasertech:
Imagine adding another test for f16 precision crazy ahaha thanks!

Rocketknight1 (approved):
LGTM! One nit: Can we move the import to the top of the file? We're already importing `ACT2FN` and `ACT2CLS` is in the same file, so there should be no performance penalty.

Rocketknight1:
Thanks for the quick iteration, merging!

tests/models/apertus/test_modeling_apertus.py
wasertech: Thinking about this a little more, we can simply make sure the correct dtype is passed to the hidden layers. Should be enough and we technically don't need a GPU to do this test.
```py
@require_torch
class ApertusMLPTest(unittest.TestCase):
    def test_apertus_mlp_float16_execution(self):
        """
        Tests that ApertusMLP can execute in float16 without crashing due to dtype mismatch
        in the XIELU activation function.
        """
        import torch

        from transformers import ApertusConfig
        from transformers.models.apertus.modeling_apertus import ApertusMLP

        config = ApertusConfig(hidden_size=64, intermediate_size=128, hidden_act="xielu", dtype="float16")

        # Force float16 execution environment
        torch.set_default_dtype(torch.float16)
        try:
            mlp = ApertusMLP(config)

            # Verify act_fn is using correct dtype parameters if applicable
            # XIELU parameters should be float16
            if hasattr(mlp.act_fn, "beta"):
                self.assertEqual(mlp.act_fn.beta.dtype, torch.float16)

            # x = torch.randn(1, 64, dtype=torch.float16)
            # output = mlp(x)

            # self.assertEqual(output.dtype, torch.float16)
        finally:
            torch.set_default_dtype(torch.float32)
```
But it's highly specific to this arch. and the implication is that some kind of similar test would need to be done tailored for each arch. and since the the CI test takes already 8h to complete, I'm still not convinced it's a good idea.

PR #42865, opened 2025-12-14
Raise explicit error when FP8 is requested via from_config

dikshyantacharya:
### Summary
`from_config()` currently ignores FP8 intent specified via `quantization_config`, silently falling back to FP32.
This PR makes the behavior explicit by raising a clear error instead.

### What this PR does
- Raises `NotImplementedError` when FP8 is requested via `from_config()`
- Adds a regression test to lock in the behavior

### Why
- Prevents silent misconfiguration
- Improves API correctness
- Establishes a safe foundation for future FP8 support

### Scope
This PR does **not** enable FP8 kernels or backend support.
It focuses solely on correctness and explicit behavior.


src/transformers/modeling_utils.py
```diff
@@ -1306,6 +1306,13 @@ def __init__(self, config: PreTrainedConfig, *inputs, **kwargs):
                 f"`model = {self.__class__.__name__}.from_pretrained(PRETRAINED_MODEL_NAME)`"
             )
         self.config = config
+        quant_config = getattr(config, "quantization_config", None)
+        if quant_config is not None:
+            if quant_config.get("quant_method") == "fp8":
+                raise NotImplementedError(
+                    "FP8 via `from_config()` is not yet supported. "
+                    "FP8 models must be created via `from_pretrained()` with an FP8-capable backend."
+                )
```
MekkCyber: let's make it general and not only for fp8

tests/quantization/config/test_from_config.py
```diff
@@ -0,0 +1,11 @@
...
+
+from transformers import AutoConfig, AutoModel
+
+
+def test_fp8_from_config_raises():
+    config = AutoConfig.from_pretrained("gpt2")
+    config.quantization_config = {"quant_method": "fp8"}
+
+    with pytest.raises(NotImplementedError, match="FP8 via"):
+        AutoModel.from_config(config)
```
MekkCyber, line 14: could you add the test to the quantization tests instead ?

MekkCyber (commented):
Thanks for the PR!

PR #42867, opened 2025-12-15
fix tests/models/edgetam_video/test_modeling_edgetam_video.py

sywangyi:
and tests/models/sam2_video/test_modeling_sam2_video.py crash

fix crash like
FAILED tests/models/edgetam_video/test_modeling_edgetam_video.py::EdgeTamVideoModelIntegrationTest::test_inference_mask_generation_video_multi_objects_multi_points - AttributeError: `positional_embedding` is not an nn.Parameter

### DIFF
--- a/src/transformers/models/edgetam_video/modeling_edgetam_video.py
+++ b/src/transformers/models/edgetam_video/modeling_edgetam_video.py
@@ -1976,9 +1976,7 @@ class EdgeTamVideoModel(EdgeTamVideoPreTrainedModel):
     input_modalities = ("video", "text")
     _can_record_outputs = {"mask_decoder_attentions": OutputRecorder(EdgeTamVideoTwoWayAttentionBlock, index=2)}
     _keys_to_ignore_on_load_unexpected = []
-    _tied_weights_keys = {
-        "prompt_encoder.shared_embedding.positional_embedding": "shared_image_embedding.positional_embedding"
-    }
+    _tied_weights_keys = {}
     # need to be ignored, as it's a buffer and will not be correctly detected as tied weight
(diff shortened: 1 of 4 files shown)


sywangyi:
@SunMarc please help review.
wo the fix the tests/models/sam2_video/test_modeling_sam2_video.py and  tests/models/edgetam_video/test_modeling_edgetam_video.py all fail

SunMarc (commented):
Thanks for checking, while this might fix the tests, I'm sure this will probably lead to another issues as . cc @molbap @ArthurZucker 

The comment below is probably connected to this issue: 
>     # need to be ignored, as it's a buffer and will not be correctly detected as tied weight
>     _keys_to_ignore_on_load_missing = ["prompt_encoder.shared_embedding.positional_embedding"]

sywangyi:
> Thanks for checking, while this might fix the tests, I'm sure this will probably lead to another issues as . cc @molbap @ArthurZucker
> 
> The comment below is probably connected to this issue:
> 
> > ```
> > # need to be ignored, as it's a buffer and will not be correctly detected as tied weight
> > _keys_to_ignore_on_load_missing = ["prompt_encoder.shared_embedding.positional_embedding"]
> > ```

I see https://github.com/huggingface/transformers/blob/64c12fdf5f6a35704f2932db533c76973b1e9a30/src/transformers/models/sam_hq/modular_sam_hq.py#L445， it only contains _keys_to_ignore_on_load_missing wo tied weight

sywangyi:
same with https://github.com/huggingface/transformers/pull/42840, so could be closed

SunMarc:
Thanks !

PR #42873, opened 2025-12-15
Enforce call to `post_init` and fix all of them

Cyrilvallez:
# What does this PR do?

As per the title. It is high-time to enforce a check in repo consistency for it.


Makefile
```diff
@@ -45,7 +45,7 @@ repo-consistency:
 	python utils/check_modular_conversion.py
 	python utils/check_dummies.py
 	python utils/check_repo.py
-	python utils/check_init_weights_data.py
+	python utils/check_modeling_structure.py
```
ArthurZucker, line 48: time to re-order as well, and make them || 
```suggestion
	python utils/check_modeling_structure.py ||
```
this way you get many errors instead of one

utils/check_modeling_structure.py
```diff
@@ -0,0 +1,144 @@
...
+        except Exception as exc:
+            violations.append(f"{file_path}: failed to parse ({exc}).")
+            continue
+
+        for node in ast.walk(tree):
+            violations = check_init_weights(node, violations, file_path)
+            violations = check_post_init(node, violations, file_path)
+
+    if violations:
+        print("\n".join(violations), file=sys.stderr)
```
ArthurZucker, line 145: qol let's ask AI to add colors to focus on which violations for which models / which line etc please

ArthurZucker (approved):
Ty

PR #42876, opened 2025-12-15
Document tensor parallelism configuration with Trainer

### LINKED ISSUE #41141: Need a concise example of Tensor Parallelism training using Trainer/SFTTrainer.
### Feature request

I have checked the code and there are few places which talk about TP. I saw from_pretrained method for model contains tp_plan and device_mesh. I also checked that the TrainingArgument can take parallelism_config which defines the TP/CP plan along with FSDP. However, I am not able to successfully stitch things together to make the only TP based training work. Please help.

Ref: 
- https://github.com/huggingface/transformers/blob/main/examples/3D_parallel.py

### Motivation

Need to enable only TP based training, but no tutorial or example is available.

### Your contribution

Given proper understanding and proper guidance, I can come up with clean example and documentation for the same.

Likhita-17:
This PR adds a concise documentation section explaining how to configure tensor parallelism (TP) when training with `Trainer`.

It clarifies that TP must be managed by Accelerate rather than during model loading, and explains why `device_map="auto"` should not be used with distributed training. A minimal, runnable TP-only Trainer example is included.

Fixes #41141

## Before submitting

## Who can review?

@stevhliu

### DIFF
--- a/docs/source/en/perf_train_gpu_many.md
+++ b/docs/source/en/perf_train_gpu_many.md
@@ -93,6 +93,116 @@ Tensor parallelism is effective for training large models that don't fit into th
 
 Refer to the [Tensor parallelism](./perf_infer_gpu_multi) guide to learn how to use it for inference.
 
+
+### Tensor parallelism for training with Trainer
+
+This section focuses on configuring tensor parallelism for training
+using Trainer.
+
+When training with `Trainer`, tensor parallelism must be configured and
+managed by  Accelerate rather than during model loading.
+
+**Important**
+
+Do **not** load the model with `device_map="auto"` when using `Trainer`
+in distributed training.  
+The `device_map="auto"` option pre-shards the model for inference and
+conflicts with Accelerate-managed training, which can lead to runtime
+errors.
+
+#### Minimal TP-only Trainer example
+
+The following example demonstrates tensor-parallel training using
+`Trainer` and `ParallelismConfig`, without combining data parallelism or
+other parallelism strategies.
+
+```python
+import os
+import torch
+from datasets import Dataset
+from transformers import (
+    AutoModelForCausalLM,
+    AutoTokenizer,
+    TrainingArguments,
+    Trainer,
+    DataCollatorForLanguageModeling,
+)
+from accelerate import ParallelismConfig
+
+# Number of processes launched with accelerate
+tp_size = int(os.environ.get("WORLD_SIZE", 1))
+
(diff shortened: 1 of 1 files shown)


stevhliu (commented):
thanks!

- i would put this under the Trainer optimization section [here](https://huggingface.co/docs/transformers/trainer#optimizations)
- add a link to tensor parallelism [docs](https://huggingface.co/docs/transformers/perf_infer_gpu_multi)
- i think it's possible to just set `tp_plan="auto"` in `from_pretrained` without explicitly using `ParallelismConfig`. i believe Trainer creates this for you [here](https://github.com/huggingface/transformers/blob/c7e5b749a6392ea2f42fea983af41f825b0bc78d/src/transformers/trainer.py#L5098) right @SunMarc? 🙏

SunMarc:
> i think it's possible to just set tp_plan="auto" in from_pretrained without explicitly using ParallelismConfig. i believe Trainer creates this for you [here](https://github.com/huggingface/transformers/blob/c7e5b749a6392ea2f42fea983af41f825b0bc78d/src/transformers/trainer.py#L5098) right @SunMarc? 🙏

Yeah you can pass tp_plan="auto" which will shard the model on the gpus that you have and Trainer will create the correct ParallelismConfig. You can also pass a device_mesh created from ParallelismConfig when initializing the model. As for `device_map="auto"`, if you run it with `torchrun` it will also enable tp when loading the model.

KrishnaKarunya-2K6:
Thanks very much for the review and constructive feedback!

I’m happy to make the suggested adjustments:

- Move the TP configuration section under the existing Trainer optimization area as recommended.
- Add a link to the tensor parallelism docs for additional context.
- Clarify that in many cases, users can simply pass `tp_plan="auto"` in `from_pretrained()` and Trainer will create the correct parallelism configuration for TP-only training.

If it’s helpful, I can update the doc text to reflect these points — just let me know if there’s a preferred location/heading for the Trainer optimization section.

Thanks again for guiding this consolidated effort — I’m glad the content has come together nicely and addresses the original issue (#41141)!

stevhliu:
sounds good, you can add a "tensor parallelism" section after [this](https://huggingface.co/docs/transformers/main/en/trainer#torchcompile)

KrishnaKarunya-2K6:
Thanks for the guidance! I’ve added the Tensor Parallelism section immediately after the `torchcompile` heading. Please let me know if you'd like any refinements or additional detail.

PR #42877, opened 2025-12-15
[CB] Allow block sharing in hybrid models

remi-or:
## Summary
This PR increases the granularity of KV cache sharing. Previously, sharing was only enabled for full-attention only models. With this PR, sharing is enabled for any model that has full attention layers. It can still be disabled with a flag. 
This PR paves the way for parallel decoding, by making it more efficient in hybrid-architectures.

## Performance 
| Attention | Version | Generated tokens | Duration (s) | Throughput (tok/s) |
|------------------------|----------------|--------------------------|------------------|--------------------------|
| Flash attention 3 | This PR         | 113149 | 16.61 | 6811.56 |
| Flash attention 3 | Main branch  | 112599 | 16.73       | 6729.27           |
| Flash attention 2 |  This PR        | 113670 | 24.83 | 4578.74 |
| Flash attention 2 | Main branch  | 112822 | 24.61       | 4584.74          |
| SDPA                  | This PR         | 112970 | 84.32 | 1339.76 |
| SDPA                  | Main branch  | 113254  | 82.49       | 1373.00         |

## Tests
No new failures for the CB tests. 

## Sanity check
Ran the command `python examples/pytorch/continuous_batching.py --samples 20 --add-prefix --compile --compare` and outputs were nearly the same and made sense.


stevhliu:
Looks cool, let me know if I can help add some docs for it [here](https://huggingface.co/docs/transformers/continuous_batching)!

remi-or:
@stevhliu I am pushing a lot of new features to CB right now, will let you know when things settle down so we can do a big push on docs! Thanks!

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -220,19 +221,20 @@ def __init__(
...
                 cm = SlidingAttentionCacheAllocator(i, self.block_size, config.sliding_window)
             else:
                 raise ValueError(f"Invalid group type: {group_type}")
             self.group_cache_managers.append(cm)
 
-        # We only use prefix sharing if the whole model has only full attention layers
-        self.use_prefix_sharing = allow_prefix_sharing and group_types == ["full_attention"]
-        self._block_manager = BlockManager(num_blocks, self.block_size, self.use_prefix_sharing)
+        # We only use prefix sharing if the whole model has only full attention layers and block sharing is allowed
+        self.use_prefix_sharing = allow_block_sharing and group_types == ["full_attention"]
```
ArthurZucker, line 236: no rename on the self attr?

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -352,7 +354,8 @@ def search_prefix_match(self, request_id: str, prompt_ids: list[int]) -> int:
         allocated_blocks = []
         for b in range(len(prompt_ids) // self.block_size):
             tokens = prompt_ids[b * self.block_size : (b + 1) * self.block_size]
-            current_hash = self._block_manager.compute_hash(current_hash, tokens)
+            # Prefix sharing is only supported when there is only one full attention layer group, so group_id=0.
```
ArthurZucker, line 357: is the comment still valid? thought this PR allowed different groups to exist, and only acts on 1

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -369,18 +372,19 @@ def search_prefix_match(self, request_id: str, prompt_ids: list[int]) -> int:
         self._total_prefix_length += prefix_length
         return prefix_length
 
-    def mark_blocks_as_complete(self, state: RequestState) -> None:
-        """Marks the blocks that have been computed in the forward pass as complete. If prefix sharing is off, this is
-        a no-op."""
-        num_complete_blocks = 0 if not self.use_prefix_sharing else self.blocks_to_complete.pop(state.request_id)
+    def mark_shareable_blocks_as_complete(self, state: RequestState) -> None:
+        """Marks the shareable blocks that have been computed in the forward pass as complete. If block sharing is off,
```
ArthurZucker:
```suggestion
        """Marks the shareable blocks that have been computed in the forward pass as complete (meaning it contains cache for tokens that are already processed, vs empty cache for futur new tokens). If block sharing is off,
```

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -369,18 +372,19 @@ def search_prefix_match(self, request_id: str, prompt_ids: list[int]) -> int:
...
         if num_complete_blocks == 0:
             return None
-        cm = self.group_cache_managers[0]  # if prefix sharing is on, there is only one group
-        self._block_manager.mark_blocks_as_complete(
-            num_complete_blocks=num_complete_blocks,
-            allocated_blocks=cm.block_table[state.request_id],
-            prompt_ids=(state.initial_tokens + state.generated_tokens),
-        )
+        for cm in self.group_cache_managers:
+            if cm.uses_block_sharing:
```
ArthurZucker, line 384: ok

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -31,29 +31,31 @@ def reverse_enumerate(xs: list[T]) -> Iterator[tuple[int, T]]:
         index -= 1
 
 
-class Block:
+class Block:  # TODO: rename to ShareableBlock and update the docs
```
ArthurZucker, line 34: SlidingBlock(Block) are not sharable

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -102,17 +104,20 @@ def has_enough_free_blocks(self, n_blocks: int) -> bool:
...
+        If the (shareable) flag is set to True, a Block object is created to keep track of the block, with the
+        (last_block_id) to indicate the last block id in the sequence, also named the parent block. If the manager
+        cannot find enough free blocks, it returns None."""
         if not self.has_enough_free_blocks(n_blocks):
             return None
         allocated_block_ids = [self._uninit_block_ids.popleft() for _ in range(n_blocks)]
-        # If we use prefix caching, we keep track of the allocated blocks as partial blocks
-        if self._use_prefix_sharing:
+        # If the block is shareable, we keep track of the allocated blocks as partial blocks
+        if shareable:
```
ArthurZucker, line 118: in general with `transformers` we don't want if else, we want 2 classes, 1 for sharable, on for non sharable. Splitting the logic by class usually scales better

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -137,23 +142,23 @@ def decrease_ref_count(self, block_id: int) -> None:
...
         blocks queue. Otherwise, their new state depends on whether they are complete."""
-        if self._use_prefix_sharing:
+        if shareable:
             for block_id in blocks:
                 self.decrease_ref_count(block_id)
         else:
             self._uninit_block_ids.extend(blocks)
 
-    def mark_blocks_as_complete(
+    def mark_shareable_blocks_as_complete(
```
ArthurZucker, line 154: only sharable blocks can be marked as complete and only sharable blocks need this logic (same as above comment) how can we better "encapsulate" this?

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -780,7 +780,8 @@ def __init__(
         self.profile = getattr(generation_config, "profile", False)  # TODO: not supported yet
         self.manual_eviction = manual_eviction
         self.batch_processor: ContinuousBatchProcessor | None = None
-        self._allow_prefix_sharing = allow_prefix_sharing
+        self._allow_block_sharing = allow_block_sharing
+        self._use_prefix_sharing = allow_block_sharing  # approximation until the cache is created
```
ArthurZucker, line 784: kinda a weird to have 2 that do the same thing

tests/generation/test_continuous_batching.py
```diff
@@ -402,52 +404,103 @@ def test_prefix_sharing(self) -> None:
...
+
+            # Second request, which should reuse the same blocks for the full attention layer groups
+            request_id = manager.add_request(inputs, max_new_tokens=32)
             chunk_with_reuse = next(manager.request_id_iter(request_id))
 
             # There should only still be two blocks in the hash table because of block reuse
             self.assertEqual(
                 len(hash_table),
-                2,
+                2 * num_fa,
```
ArthurZucker, line 463: actually a nice in depth test sir, lgtm

ArthurZucker (approved):
sounds good! The perf don't show diff, they should for models that were hybrid no?

src/transformers/generation/continuous_batching/cache.py
remi-or: Good question! `prefix_sharing` is only possible if block sharing (which is more of a memory optimization) is enabled AND the model has no sliding window layers: if there are any, they will create sliding window groups with no shareable blocks, hence no prefix sharing.

src/transformers/generation/continuous_batching/cache.py
remi-or: Yes, because prefix sharing is still only activated if there is only one group -- cf comment above

src/transformers/generation/continuous_batching/cache.py
remi-or: Not entirely, but I can see the confusion. Will adapt

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: Yes, so they wont create this kind of object! Hence the proposed name change.

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: The thing is, for non shareable blocks, we dont need the block object at all: we just want the physical block_id. Hence no need to create a python object that will never be used and keep track of it. So less overhead!

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: I guess this will make more sense when Block is renamed to ShareableBlock. In that case, I think the function name makes sense. The issue is I renammed one but not the other imo

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: They don't! The block sharing boolean allows hybrid models to to block sharing. The use_prefix_sharing bool is updated after we know if the model is hybrid or not. Hence the comment about the approximation!

tests/generation/test_continuous_batching.py
remi-or: Thanks sir!

remi-or:
> sounds good! The perf don't show diff, they should for models that were hybrid no?

Not really sadly, this is more of a memory optimization. It will ensure the parallel decoding wont eat too much memory with hybrid models. The perf table is more here as a sanity check to make sure we did not incur any major overhead with these changes.

src/transformers/generation/continuous_batching/cache_manager.py
ArthurZucker: ok got it !

src/transformers/generation/continuous_batching/cache_manager.py
ArthurZucker: ok!

src/transformers/generation/continuous_batching/cache.py
ArthurZucker: but then to mark as complete we loop on the groups no?

remi-or: Yes because marking a block as complete is useful in the context of block sharing, which can happen in an hybrid model. But here we are in the context of prefix sharing, which is more restrictive, so we know there is only one group. Maybe I am missing something here

PR #42879, opened 2025-12-15
[CI] Fixing some AMD failures

Abdennacer-Badaoui:
This PR fixes several failures on AMD for Qwen2, Qwen2.5-Omni, and Qwen2.5-VL.


src/transformers/models/qwen2_5_omni/modeling_qwen2_5_omni.py
```diff
@@ -2388,7 +2388,7 @@ def forward(
                 self.rope_deltas = rope_deltas
 
             else:
-                batch_size, seq_length, _ = inputs_embeds.shape
+                batch_size, seq_length = input_ids.shape
```
remi-or: I think `input_ids` are optional, as is `input_embeds` -- is there an argument that is not that we could infer those dimensions with? Otherwise I think one of the two inputs_` needs to not be `None` , so we can get the dims from whichever is not `None`

src/transformers/models/qwen2_5_omni/modular_qwen2_5_omni.py
```diff
@@ -2588,7 +2588,7 @@ def forward(
                 self.rope_deltas = rope_deltas
 
             else:
-                batch_size, seq_length, _ = inputs_embeds.shape
+                batch_size, seq_length = input_ids.shape
```
remi-or: same as above

tests/models/qwen2/test_modeling_qwen2.py
```diff
@@ -78,11 +78,10 @@ def test_model_450m_logits(self):
         with torch.no_grad():
             out = model(input_ids).logits.float().cpu()
         # Expected mean on dim = -1
-        EXPECTED_MEAN = torch.tensor([[-1.9537, -1.6193, -1.4123, -1.4673, -1.8511, -1.9309, -1.9826, -2.1776]])
+        EXPECTED_MEAN = torch.tensor([[-2.2121, -1.6335, -1.4816, -1.5035, -1.9110, -1.8979, -1.9682, -2.1980]])
```
remi-or, line 81: I will run the slow test to see if this does not break things on CUDA, otherwise you should add Expectations

remi-or (commented):
A comment (repeated twice) to address + I will run the slow tests

remi-or:
run-slow: qwen2, qwen2_5_omni, qwen2_5_vl

src/transformers/models/qwen2_5_omni/modeling_qwen2_5_omni.py
Abdennacer-Badaoui: Okay, let's do that. We'll try to get the dimensions from input_embeds. If they are None (as was the case for some failing tests), we will then try to infer them from input_ids.

tests/models/qwen2/test_modeling_qwen2.py
Abdennacer-Badaoui: Yes. I remember checking the Nvidia CI, they were the same expectation outputs. But let's see for these exepected means.

remi-or:
run-slow: qwen2, qwen2_5_omni, qwen2_5_vl

remi-or (approved):

PR #42880, opened 2025-12-15
fix: make mistral base check conditional to fix offline loading

Killusions:
Backport of a small part of #42528, the model_info call currently breaks offline loading

⚒️ with ❤️ by @siemens

### DIFF
--- a/src/transformers/tokenization_utils_base.py
+++ b/src/transformers/tokenization_utils_base.py
@@ -2435,7 +2435,12 @@ class PreTrainedTokenizerBase(SpecialTokensMixin, PushToHubMixin):
                     return True
             return False
 
-        if _is_local or is_base_mistral(pretrained_model_name_or_path):
+        if is_offline_mode():
+            _is_local = True
+
+        if pretrained_model_name_or_path is not None and (
+            _is_local or (not _is_local and is_base_mistral(pretrained_model_name_or_path))
+        ):
             _config_file = cached_file(
                 pretrained_model_name_or_path,
                 "config.json",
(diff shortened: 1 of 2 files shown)


src/transformers/tokenization_utils_base.py
```diff
@@ -2435,7 +2435,12 @@ def is_base_mistral(model_id: str) -> bool:
                     return True
             return False
 
-        if _is_local or is_base_mistral(pretrained_model_name_or_path):
+        if is_offline_mode():
+            is_local = True
+
+        if pretrained_model_name_or_path is not None and (
+            is_local or (not is_local and is_base_mistral(pretrained_model_name_or_path))
+        ):
```
vasqu, line 2443:
```suggestion
        if is_offline_mode():
            _is_local = True

        if pretrained_model_name_or_path is not None and (
            _is_local or (not _is_local and is_base_mistral(pretrained_model_name_or_path))
        ):
```
The signature is different from v4 to current v5

vasqu (commented):
Can you add a test? This should fail in general already because the local var `_is_local` vs `is_local`

Killusions:
@vasqu I'm fine with doing these changes and adding a test, but should we really deviate from the upstream/main branch in a backport? See https://github.com/huggingface/transformers/pull/42528

Killusions:
And if we add a test we should also merge it to main

Killusions:
@vasqu Ah I think I misunderstood you, fixed

src/transformers/tokenization_utils_base.py
Killusions: Fixed.

Killusions:
@vasqu Also added a test

vasqu:
cc @ArthurZucker @itazap

Killusions:
@vasqu @ArthurZucker @itazap 

What's the timeline for this? Is there any change you still require before merging this?

Killusions:
Bump @vasqu @ArthurZucker @itazap

ArthurZucker:
Hey! I'll see with @vasqu

PR #42882, opened 2025-12-15
Fix dtype quantizer

SunMarc:
# What does this PR do?

This PR fixes code related to quantizer since `dtype=auto` default created some minor issues in our CI. We clean a bit of code related to dtype and fixes the tests as some of ours tests depends on the dtype of the model, which before where always on `torch.float32` if not specified.

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -282,7 +282,6 @@ def get_state_dict_dtype(state_dict):
     """
     Returns the first found floating dtype in `state_dict` if there is one, otherwise returns the first dtype.
     """
-    print(state_dict)
     for t in state_dict.values():
         if t.is_floating_point():
             return t.dtype
@@ -323,15 +322,12 @@ def load_state_dict(
     """
     # Use safetensors if possible
     if checkpoint_file.endswith(".safetensors"):
-        print(checkpoint_file)
         with safe_open(checkpoint_file, framework="pt") as f:
             state_dict = {}
             for k in f.keys():
                 if map_location == "meta":
                     _slice = f.get_slice(k)
                     k_dtype = _slice.get_dtype()
-                    print("getting dtype")
-                    print(k_dtype)
                     if k_dtype in str_to_torch_dtype:
                         dtype = str_to_torch_dtype[k_dtype]
                     else:
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -1437,7 +1437,7 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
     def pp_plan(self, plan: dict[str, tuple[str, str]]):
         self._pp_plan = plan
 
-    def dequantize(self):
+    def dequantize(self, dtype=None):
         """
         Potentially dequantize the model in case it has been quantized by a quantization method that support
         dequantization.
@@ -1447,7 +1447,7 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         if hf_quantizer is None:
             raise ValueError("You need to first quantize your model in order to dequantize it")
 
-        return hf_quantizer.dequantize(self)
+        return hf_quantizer.dequantize(self, dtype=dtype)
 
     def _backward_compatibility_gradient_checkpointing(self):
         if self.supports_gradient_checkpointing and getattr(self.config, "gradient_checkpointing", False):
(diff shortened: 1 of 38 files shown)


src/transformers/modeling_utils.py
```diff
@@ -3937,22 +3943,18 @@ def from_pretrained(
...
 
+            if hf_quantizer is not None:  # replace module with quantized modules (does not touch weights)
+                hf_quantizer.preprocess_model(
+                    model=model,
+                    dtype=dtype,
+                    device_map=device_map,
+                    checkpoint_files=checkpoint_files,
+                    use_kernels=use_kernels,
+                )
+
```
SunMarc, line 3954: we need to pre_process the model under the context manager. otherwise, we don't init some tensors with the right dtype

src/transformers/quantizers/base.py
```diff
@@ -198,9 +198,8 @@ def preprocess_model(self, model: "PreTrainedModel", config, dtype=None, checkpo
         # once the weights have been quantized
         # Note that once you have loaded a quantized model, you can't change its dtype so this will
         # remain a single source of truth
-        original_dtype = dtype if dtype is not None else torch.get_default_dtype()
-        config._pre_quantization_dtype = original_dtype
-        _assign_original_dtype(model, original_dtype)
+        model.config._pre_quantization_dtype = dtype
+        _assign_original_dtype(model, dtype)
```
MekkCyber: is this still needed ? shouldn't we be able to access the model dtype using `model.config.dtype` as we did in `dequantize` ?

src/transformers/quantizers/quantizer_aqlm.py
```diff
@@ -28,7 +28,7 @@
 
 
 if is_torch_available():
-    import torch
+    pass
```
MekkCyber: to remove

src/transformers/quantizers/quantizer_auto_round.py
```diff
@@ -24,7 +24,7 @@
 
 
 if is_torch_available():
-    import torch
+    pass
```
MekkCyber: to remove as well

src/transformers/quantizers/quantizer_torchao.py
```diff
@@ -134,22 +134,11 @@ def validate_environment(self, *args, **kwargs):
 
     def update_dtype(self, dtype):
         if self.quantization_config.quant_type == "int4_weight_only":
-            if dtype is not None and dtype != torch.bfloat16:
+            if dtype != torch.bfloat16:
                 logger.warning_once(
```
MekkCyber, line 138: not sure why in some cases we are raising errors in case the dtypes are not supported, and in other cases we simply override the dtype

src/transformers/modeling_utils.py
MekkCyber: but the context is needed only when no dtype is set, but for quantized linears I think we set the correct dtypes or am i wrong ?

src/transformers/modeling_utils.py
```diff
@@ -3999,7 +4001,7 @@ def from_pretrained(
 
         if hf_quantizer is not None:
             model.hf_quantizer = hf_quantizer
-            hf_quantizer.postprocess_model(model, config=config)  # usually a no-op but sometimes needed
+            hf_quantizer.postprocess_model(model)  # usually a no-op but sometimes needed
```
MekkCyber:
```suggestion
            hf_quantizer.postprocess_model(model)  # usually a no-op but sometimes needed, e.g to remove the quant config when dequantizing 
```

MekkCyber (commented):
Amazing ! Thanks for cleaning the dtype mess 🔥

src/transformers/modeling_utils.py
SunMarc: even when it is set, we need it. It changes the default torch dtype in the context manager. I didn't want to recreate a context manager. I put that there because some tests where failing because of that. The quantized model was not the same in term of memory.

src/transformers/quantizers/quantizer_aqlm.py
SunMarc: indeed

src/transformers/quantizers/quantizer_auto_round.py
SunMarc: thx

src/transformers/quantizers/quantizer_torchao.py
SunMarc: indeed, warning is better

src/transformers/quantizers/base.py
SunMarc: indeed, now that we are dtype is officially used, it should be possible. Made the change !

MekkCyber (approved):
Incredible, thanks a lot!

PR #42883, opened 2025-12-15
[docs] dtype

stevhliu:
update the model loading docs to clarify that `"auto"` is the default now in `from_pretrained`


Cyrilvallez (approved):
Very nice, thanks!

PR #42887, opened 2025-12-15
[Quantization] [Compressed Tensors] Support Transforms, Fix Tests

kylesayrs:
## Purpose ##
* Support loading models with online transforms applied via Compressed Tensors (LLM Compressor)
* Fix tests which referenced deleted models

## Background ##
Transforms are extra weights added to a model which improve accuracy recovery from quantization. These extra weights are required to be shared in order to reduce memory requirements of the model.

## Prerequisites ##
* Transformers: #42935
* Compressed Tensors: [[Transforms] Support loading transforms in transformers](https://github.com/vllm-project/compressed-tensors/pull/528)

## Changes ##
* Require a minimum compressed tensors version of `0.13.1` (to support transform features)
* Apply transforms to the model before weight loading
  * `apply_transform_config` implements `_update_transforms_tied_weights`, which leverages @Cyrilvallez 's refactored `tie_weights` functionality!

Implemented in CT's `apply_transform_config`
```python3
def _update_transforms_tied_weights(model: torch.nn.Module):
    """
    This function updates the `_tied_weights_keys` and `all_tied_weights_keys`
    attributes of the given model with transform weights.

    This function is needed because transformers only knows which weights are shared
    via the `_tied_weights_keys` attributes. These attributes are used to tie
    weights after the model has loaded.

    CompressedTensors does not enforce a particular weight is the source weight :.
    We rely on correctness of the following mapping in PreTrainedModel.tie_weights():
    ```
    B -> A
    C -> A
    D -> A

    Where any of A,B,C,D might be the loaded source weight
    ```
    This property is tested by `test_modeling_utils::BaseModelWithMultipleTiedWeights`
    """
```

Example `_tied_weights_keys`:
```python3
"model.layers.1.self_attn_.q_proj.v_input.weight": "model.layers.0.self_attn_.q_proj.v_input.weight",
"model.layers.2.self_attn_.q_proj.v_input.weight": "model.layers.0.self_attn_.q_proj.v_input.weight",
"model.layers.3.self_attn_.q_proj.v_input.weight": "model.layers.0.self_attn_.q_proj.v_input.weight",
...
"model.layers.1.self_attn_.q_proj.u_output.weight": "model.layers.0.self_attn_.q_proj.u_output.weight",
"model.layers.2.self_attn_.q_proj.u_output.weight": "model.layers.0.self_attn_.q_proj.u_output.weight",
"model.layers.3.self_attn_.q_proj.u_output.weight": "model.layers.0.self_attn_.q_proj.u_output.weight",
```

## Testing ##
* Regression tested using `CompressedTensorsTest`, added an online quip-style transformed model for testing
  * Perplexity results match expectation

## Suggested Reviewers ##
@SunMarc @Cyrilvallez @Rocketknight1


Rocketknight1:

kylesayrs:
`make fix-copies` does not fix the CI 🥲

src/transformers/quantizers/quantizer_compressed_tensors.py
```diff
@@ -112,3 +128,26 @@ def is_qat_trainable(self) -> bool:
     def is_serializable(self) -> bool:
         """Models quantized using compressed tensors can be saved to disk"""
         return True
+
+    def _update_transforms_tied_weights(self, model: "PreTrainedModel"):
+        from compressed_tensors.transform import TransformBase
+
+        # create mapping: tied_ptr -> key
```
SunMarc: can you add a docstring here to explain why this is needed ?

src/transformers/utils/quantization_config.py
```diff
@@ -1094,25 +1094,16 @@ class CompressedTensorsConfig(QuantizationConfigMixin):
...
+            config_groups (`dict[str, typing.Union[ForwardRef('QuantizationScheme'), list[str]]] | None`, *optional*): <fill_docstring>
+            format (`str`, *optional*, defaults to `"dense"`): <fill_docstring>
+            quantization_status (`QuantizationStatus`, *optional*, defaults to `"initialized"`): <fill_docstring>
+            kv_cache_scheme (`Optional`, *optional*): <fill_docstring>
+            global_compression_ratio (`float | None`, *optional*): <fill_docstring>
+            ignore (`list[str] | None`, *optional*): <fill_docstring>
+            sparsity_config (`dict[str, typing.Any] | None`, *optional*): <fill_docstring>
+            transform_config (`Optional`, *optional*): <fill_docstring>
+            quant_method (`str`, *optional*, defaults to `"compressed-tensors"`): <fill_docstring>
+            run_compressed (`bool`, *optional*, defaults to `True`): <fill_docstring>
```
SunMarc: don't forget to fill it

src/transformers/utils/quantization_config.py
```diff
@@ -1124,23 +1115,25 @@ def __init__(
         global_compression_ratio: float | None = None,
         ignore: list[str] | None = None,
         sparsity_config: dict[str, Any] | None = None,
+        transform_config: Optional[dict[str, Any]] = None,
```
SunMarc:
```suggestion
        transform_config: dict[str, Any] | None = None,
```

tests/quantization/compressed_tensors_integration/test_compressed_tensors.py
```diff
@@ -98,9 +82,13 @@ def _test_quantized_model(self, model_name: str, expected_output: list):
...
-        outputs = tokenizer.batch_decode(generated_ids)
+        labels = inputs["input_ids"]
+        with torch.no_grad():
+            outputs = quantized_model(**inputs, labels=labels)
 
-        self.assertIsNotNone(outputs)
-        self.assertIn(outputs[0], expected_output)
+        # check perplexity
+        perplexity = torch.exp(outputs.loss)
+        self.assertLessEqual(perplexity, expected_perplexity)
```
SunMarc, line 94: Happy to check for perplexity ! From your experience, do you think it is suitable to catch regression, even when it is a minor one ? I'm thinking about switching to potentially this as this might solves issues when we are checking for different hardware.

SunMarc (commented):
Thanks, overall LGTM ! Just a few nits

src/transformers/utils/import_utils.py
```diff
@@ -1028,7 +1028,8 @@ def is_qutlass_available():
 
 @lru_cache
 def is_compressed_tensors_available() -> bool:
-    return _is_package_available("compressed_tensors")
+    is_available, ct_version = _is_package_available("compressed_tensors", return_version=True)
+    return is_available and version.parse(ct_version) >= version.parse("0.11.0")
```
MekkCyber: could you use the same pattern we use for packages with minimal versions required (like `accelerate` for example)

src/transformers/quantizers/quantizer_compressed_tensors.py
MekkCyber: Yes, a docstring is needed here. We should also consider a way to avoid this loop, since iterating over all parameters might be a bit costly.

MekkCyber (commented):
Thank you for the fix @kylesayrs

tests/quantization/compressed_tensors_integration/test_compressed_tensors.py
kylesayrs: LLM Compressor uses weekly GSM8K tests to catch regressions, but from the experience of my team members, they've found that wikitext perplexity is even more reliable than GSM8K. vLLM uses perplexity for small tests like these.

So yes, I suspect that perplexity is a good metric, but  the perplexity thresholds in this test are very loose.

src/transformers/quantizers/quantizer_compressed_tensors.py
kylesayrs: Technically we only have to iterate through every module. Each `TransformBase` only has 1-2 parameters. I agree that this could still be costly for large models, but there already exist many other places in `transformers` code where we iterate through all modules in the model.

src/transformers/utils/quantization_config.py
kylesayrs: Oof, thanks for catching

kylesayrs:
@SunMarc @MekkCyber

Recent changes from #42882 ended up breaking this PR as implemented. This is because calling `preprocess` within the `init_empty_weights` context creates meta tensors which are no longer identical. I actually think that this is fine and cleaner behavior, but it does mean that this PR needs to be refactored.

I'm moving `_update_transforms_tied_weights` into `apply_transform_config` and refactoring the function to work around `init_empty_weights` (that's done [here](https://github.com/vllm-project/compressed-tensors/pull/528)). Once CT releases, I'll reopen this PR so we can start supporting transforms in transformers :)

Thanks

Cyrilvallez:
Thanks @kylesayrs! As discussed offline, indeed it won't matter which weight is actually present in the checkpoint for tying them! Letting you with @SunMarc and @MekkCyber for the quantization part!

SunMarc:
> Recent changes from https://github.com/huggingface/transformers/pull/42882 ended up breaking this PR as implemented. This is because calling preprocess within the init_empty_weights context creates meta tensors which are no longer identical. I actually think that this is fine and cleaner behavior, but it does mean that this PR needs to be refactored.

Oh indeed, as most quants methods calling preprocess is actually calling `with_init_empty_weights` again, I didn't predict that it will create this kind of issue. Let us know if you manage to fix it ! If there are really no way and it's too breaking, we can revert this. But as you said, I think this behavior is probably better.

PR #42874, opened 2025-12-15
Fix nativ tok

itazap:
fixes https://github.com/huggingface/transformers/issues/42796

Fixes incorrect tokenization for Ministral-3 models by preserving and restoring the ByteLevel decoder from tokenizer.json when loading tokenizers. Also custom `init` methods were overwriting the ByteLevel decoder with a Metaspace decoder.

Happens because these models reference a `LlamaTokenizerFast` in their `tokenizer_config.json` on the hub. and LlamaTokenizer's init overrides tokenizer components

PR #42894, opened 2025-12-16
use `TokenizersBackend`

ArthurZucker:
# What does this PR do?
Fixes a bunch of issues (including https://github.com/huggingface/transformers/pull/42874)

With this PR, we now kinda enforce that **if there is a specific `python` path to be used** (meaning there is a `XXXXTokenizer` class that has python code whatever the backend) the tokenizer's saved `tokenizer_class` needs to match the class mapped in `tokenization_auto` to the `model_type`. 

- If there is no `config.json` we are just gonna use the serialized `tokenizer_class` but in most cases its gonna produce gibberish outputs because many many models on the hub use say `LlamaTokenizer` when the tokenizer they actually need is completely different.
- If there is a `config.json`, we extract the `model_type`, and check if `TOKENIZER_MAPPING[model_type]` matches the serialized `tokenizer_class`. 
    - if NO: we just use `TokenizersBackend`: we assume its not intended + its not a special python path + its a recent model -> we use `TokenizersBackend`. Falling back upon failure to convert to the `tokenizer_class`.
    - if YES: its safe to say this is intended, we have a good match / map we can enforce the `tokenizer` architecture.
- If it is remote code, `trust_remote_code=True` we always use the class people want

We cannot rely on the `tokenizer_class`, and we should not rely on it, but we rely on the model type. 

Also 90% of the tokenizers on the hub don't need special code, and are just supported OOB by `tokenizers` that the motivations behind this decision. 

We cannot fix ALL `tokenizers_config.json` on the hub + we actually don't want to. We want to just read the `tokenizer.json` in general using `TokenizersBackend`. 

## What matters, breaking changes 🔴 🔴 🔴 🔴 🔴 🔴 
1. `TokenizersBackend` no longer default to having `token_type_ids`. If you want them, set the flag.

2. 
Given:
```python 
tok = AutoTokenizer.from_pretrained("Qwen/Qwen2-7B")
tok.push_to_hub("ArthurZ/MyTokenizer") 
```
This
```python
tok = AutoTokenizer.from_pretrained("Qwen/Qwen2-7B")
assert tok.__class__ == Qwen2Tokenizer
```

still works, because if there is no `config.json` we do fallback to the serialized `tokenizer_class` in the `tokenizer_config.json`. But this class, for MOST of the models out there is WRONG! 😉


src/transformers/models/auto/tokenization_auto.py
```diff
@@ -336,16 +331,16 @@
         ("tvp", "BertTokenizer" if is_tokenizers_available() else None),
         ("udop", "UdopTokenizer" if is_tokenizers_available() else None),
         ("umt5", "T5Tokenizer" if is_tokenizers_available() else None),
-        ("video_llava", "LlamaTokenizer" if is_tokenizers_available() else None),
+        ("video_llava", "TokenizersBackend" if is_tokenizers_available() else None),
```
itazap: @ArthurZucker Just wondering if the `LlamaTokenizer` for all of these was causing issues?

ArthurZucker: yep! Because we now enforce Llama if it is mapped to Llama, all of these were assumed to have Llama like pre tokenizer but they really don't!

ArthurZucker: we should / could just remove them entirely from the mapping! defaulting to `TokenizersBackend`!

itazap (approved):

PR #42896, opened 2025-12-16
Overwrite `get_decoder()` in AudioLLMs

zucchini-nlp:
# What does this PR do?

As per title, audio models need to call LM's `get_decoder()` because the model has no `base model`


zucchini-nlp:
@bot /style

molbap (approved):
Sounds good - these are the only two in this case?

zucchini-nlp:
Yes, also there is Qwen2-Audio but it already has the method overriden

zucchini-nlp:
@bot /style

zucchini-nlp:
Voxtral had it overriden already 🤦🏻  I misread the GH comment, so it's only GraniteSpeech

PR #39585, opened 2025-07-22
[`Ernie 4.5`] Ernie VL models

vasqu:
Continuation of #39228 for the VL models

Current inference script for testing (torch 2.9.1):
```python
import requests
from PIL import Image

from transformers import AutoConfig, AutoModelForImageTextToText, AutoProcessor


model_path = "/raid/anton/code/forks/transformers/AntonV/ErnieVL"
processor = AutoProcessor.from_pretrained(model_path)

config = AutoConfig.from_pretrained(model_path)
model = AutoModelForImageTextToText.from_pretrained(
    "baidu/ERNIE-4.5-VL-28B-A3B-PT",
    config=config,
    device_map="auto",
    dtype="auto",
)

messages = [
    {
        "role": "user",
        "content": [
            {"type": "text", "text": "Only use English during your responses and describe the following image."},
            {"type": "image"},
        ]
    },
]
text = processor.apply_chat_template(messages, tokenize=False, add_generation_prompt=True)
image = Image.open(requests.get("https://paddlenlp.bj.bcebos.com/datasets/paddlemix/demo_images/example1.jpg", stream=True).raw)
inputs = processor(text=[text], images=[image], return_tensors="pt").to(model.device)

generated_ids = model.generate(
    **inputs,
    max_new_tokens=64,
    do_sample=False,
)
print(processor.decode(generated_ids[0][len(inputs['input_ids'][0]):]))
```
Output: 
`The image features a scenic view of rolling mountains under a clear sky. In the foreground, there is a person sitting on a hillside, facing away from the camera. The person is wrapped in a colorful, striped blanket. The landscape is dotted with vibrant pink flowers, adding a pop of color to the`

Left TODOs:
- [x] TP
  - [ ] EP to add later after its been fixed

PR #42903, opened 2025-12-16
[`Tokenizers`] Change treatment of special tokens

vasqu:
Related to #39585 where we try to make some previously made special tokens to non-special tokens


ArthurZucker (approved):
We can IMO add a small self contained test!

PR #42905, opened 2025-12-16
 Sam: Perception Encoder Audiovisual

eustlb:
# What does this PR do?


ArthurZucker (approved):
We still have a few todos post merge but huge work @eustlb GG! 🚀 
What a big model!

PR #42356, opened 2025-11-24
Fix GPT-OSS TP IndexError and unwrapping DTensor

akshan-main:
# [GPT-OSS] Fix Tensor Parallelism IndexError and DTensor casting

## What does this PR do?

This PR fixes two specific issues preventing **GPT-OSS** models from training with Tensor Parallelism (TP) and FSDP, as reported in #41819.

**The changes are:**

1.  **Fix `IndexError` in TP Hooks:**
    * **Bug:** The `tensor_parallel` hooks in transformers expect the input tensor (hidden states) to be passed as the first positional argument (`args[0]`). The `GptOssDecoderLayer` was previously passing `hidden_states` as a keyword argument, causing the hook to fail with `IndexError: tuple index out of range`.
    * **Fix:** Updated `GptOssDecoderLayer.forward` to pass `hidden_states` as the first positional argument.

2.  **Fix DTensor Casting in Eager Attention:**
    * **Bug:** When TP is enabled, `module.sinks` is wrapped as a `DTensor`. The `eager_attention_forward` function attempts to `torch.cat` this with `attn_weights` (a local tensor), causing a crash.
    * **Fix:** Added a check to detect if `sinks` is a `DTensor` and unwrap it before concatenation.

**Status:**
I have applied these changes to `modular_gpt_oss.py` and ran `make fix-copies`

Fixes #41819

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [x] Was this discussed/approved via a Github issue or the [forum](https://github.com/huggingface/transformers/issues/41819)? Please add a link
      to it if that's the case. (Linked above)
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

@3outeille 
@ArthurZucker

PR #42906, opened 2025-12-16
fix `Dtensor` and `tensor` mismatch

3outeille:
# Bug

`local_rowise` or `local_colwise` is calling `RowiseParallel(use_dtensor=False)` (resp. `ColwiseParallel(use_dtensor=False)`. Issue was first noticed in https://github.com/huggingface/transformers/pull/42356 , quoting 

> we would like to not have Dtensor logic in the modeling. For example, sinks are supposed to use local_rowwise (cf [main/src/transformers/models/gpt_oss/configuration_gpt_oss.py#L41](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt_oss/configuration_gpt_oss.py?rgh-link-date=2025-11-24T12%3A36%3A48Z#L41)) which is supposed to not return a Dtensor (cf [main/src/transformers/integrations/tensor_parallel.py#L1171](https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/tensor_parallel.py?rgh-link-date=2025-11-24T12%3A36%3A48Z#L1171)) but somehow doesnt work

# Fix

```python
def convert_and_load_state_dict_in_model(
	#...
	tp_layer = ALL_PARALLEL_STYLES[model.tp_plan[matched_tp_pattern]].__class__
```

was creating the bug because `.__class__` will reuse the class default `use_dtensor` value. Thus, overwritting the value we specifed in `local_rowise/colwise`.

The fix makes sure to properly use the `use_dtensor` value and thus no more `Dtensor` and `tensor` mismatch


src/transformers/core_model_loading.py
```diff
@@ -895,10 +895,12 @@ def convert_and_load_state_dict_in_model(
                 if matched_tp_pattern := tp_plan_alt.search(renamed_key):
                     matched_tp_pattern = tp_plan_by_group_name[matched_tp_pattern.lastgroup]
                     if getattr(mapping, "distributed_operation", None) is None:
-                        tp_layer = ALL_PARALLEL_STYLES[model.tp_plan[matched_tp_pattern]].__class__
+                        tp_layer_instance = ALL_PARALLEL_STYLES[model.tp_plan[matched_tp_pattern]]
+                        tp_layer = tp_layer_instance.__class__
                         mapping.distributed_operation = tp_layer(
                             device_mesh=device_mesh, rank=device_map[""].index, empty_param=empty_param.clone()
                         )
+                        mapping.distributed_operation.use_dtensor = tp_layer_instance.use_dtensor
```
ArthurZucker: a nice hack, tho if we can come up with a better fix lets try to avoid that please! 
the kwargs should only be 
```
device_mesh=device_mesh, rank=device_map[""].index, empty_param=empty_param.clone()
```
for init, the rest should not be kwargs of init more like hardcoded for that "type". 
If you see what I mean here we should only get the class and init it -> local_colwise should get its stuff

3outeille: Do you suggest instead ?:
```diff
mapping.distributed_operation = tp_layer(
              device_mesh=device_mesh, rank=device_map[""].index, empty_param=empty_param.clone(),
+			  use_dtensor=use_dtensor
          )
- mapping.distributed_operation.use_dtensor = tp_layer_instance.use_dtensor
```

3outeille: done

src/transformers/integrations/tensor_parallel.py
```diff
@@ -1129,7 +1156,7 @@ def _prepare_output_fn(output_layouts, use_local_output, mod, outputs, device_me
         router_indices = router_indices.masked_fill(
             router_indices == -1, num_local_experts
         )  # masking class for one hot
-        return router_scores, router_indices
+        return router_logits, router_scores, router_indices
```
ArthurZucker, line 1159: unsure this works for all models but let's see!

ArthurZucker (approved):
Looks good to me otherwise, the test will come later AFAIK with your PR on fast distributed tests

src/transformers/integrations/tensor_parallel.py
3outeille: yeah I need to fix the Expert parallel anyway so we will see

Issue #42898, opened 2025-12-16
`clean_up_tokenization_spaces` behavior changes in v5


apaniukov:
### System Info

- `transformers` version: 4.57.3/5.0.0rc1
- Platform: Linux-5.15.0-56-generic-x86_64-with-glibc2.35
- Python version: 3.10.12

### Who can help?

@ArthurZucker and @itazap

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

Code sample:
```python
from transformers import AutoTokenizer, __version__ as hf_version

checkpoint = "google/gemma-2-9b-it"
hf_tokenizer = AutoTokenizer.from_pretrained(checkpoint)

text_input = "beginning , and"    
ids = hf_tokenizer(text_input, return_tensors="np").input_ids
hf_detokenized = hf_tokenizer.decode(ids, skip_special_tokens=True, clean_up_tokenization_spaces=True)[0]
print(f"{hf_version}: `{hf_detokenized}`")
```
Outputs:
```
4.57.3:   `beginning, and`
5.0.0rc1: `beginning , and`
```

`TokenizersBackend` class (ex `PreTrainedTokenizerFast`) has no default `clean_up_tokenization` method implementation, which previously was in the `PreTrainedTokenizerBase`. This changes behavior for tokenizers that has no provide it's own method, like `GemmaTokenizer`.

`PythonBackend` class provides default behavior in the `._decode` method, but not default `clean_up_tokenization` implementation, which might also be a problem.

### Expected behavior

1. `clean_up_tokenization_spaces` flag in the `.decode` method works as in v4 version
2. No changes in detokenization behavior for tokenizers that has `self.clean_up_tokenization_spaces=True`

ArthurZucker:
see linked PR we are probably gonna allow that!

PR #42916, opened 2025-12-17
Fix: Apply clean_up_tokenization_spaces in TokenizersBackend._decode

ghost:
# What does this PR do?

This PR fixes a regression in v5.0.0rc1 where the `_decode` method in `TokenizerBackend` was not respecting the `clean_up_tokenization_spaces` paramter, causing unwanted spaces to appear before punctuation in decoded output.

**Reproduction:**

```python
tokenizer = AutoTokenizer.from_pretrained('mlx-community/Llama-3.2-1B-Instruct-4bit')
text = tokenizer.decode([128000, 64, 1174, 65])
print(text)
```

**Behavior:**

- v4.57.3: <|begin_of_text|>a,b ✅
- v5.0.0rc1 (before fix): <|begin_of_text|>a ,b ❌ (extra space before comma)
- v5.0.0rc1 (after fix): <|begin_of_text|>a,b ✅

## Solution

Added the missing `clean_up_tokenization_spaces` logic to `TokenizersBackend._decode()` method. When enabled (default behavior), it removes extra spaces before punctuation using regex pattern matching.

Fixes #42913

## Who can review?

@ArthurZucker @itazap

### DIFF
--- a/src/transformers/tokenization_utils_tokenizers.py
+++ b/src/transformers/tokenization_utils_tokenizers.py
@@ -900,8 +900,21 @@ class TokenizersBackend(PreTrainedTokenizerBase):
 
         if isinstance(token_ids, int):
             token_ids = [token_ids]
-        return self._tokenizer.decode(token_ids, skip_special_tokens=skip_special_tokens)
 
+        text = self._tokenizer.decode(token_ids, skip_special_tokens=skip_special_tokens)
+
+        clean_up_tokenization_spaces = (
+            clean_up_tokenization_spaces 
+            if clean_up_tokenization_spaces is not None 
+            else self.clean_up_tokenization_spaces
+        )
+    
+        if clean_up_tokenization_spaces:
+            import re
+            text = re.sub(r'\s+([?.!,"])', r'\1', text)
+    
+        return text
+    
     def _save_pretrained(
         self,
         save_directory: str | os.PathLike,


ArthurZucker (commented):
Hey! is there a motivation to have that? We removed it because its unintuitive, and can be done by the user itself outside. do you have a specific usecase in mind?

ghost:
Hi @ArthurZucker!

Thanks for the response! I realize I should have clarified this before submitting the PR - apologize for that.

**What I observed:**
- The `clean_up_tokenization_spaces` param exists in `decode()` and defaults to `True`
- But in `TokenizersBackend._decode()`, it's not being applied
- This causes: `decode([128000, 64, 1174, 65])` → `'a ,b'` instead of `'a,b'`

**My question:**
Was removing this cleanup intentional for v5.0.0?

If yes, I can close this PR. If it was unintended, Im happy to adjust the implementation based on your guidance.

Thanks!

apaniukov:
> Hey! is there a motivation to have that? We removed it because its unintuitive, and can be done by the user itself outside. do you have a specific usecase in mind?

@ArthurZucker the issue is that in v4 `clean_up_tokenization_spaces` flag worked for all tokenizers, and in v5 it works only when the particular tokenizer implementation defined `clean_up_tokenization` method.
1) It is inconsistent with `PythonBackend` tokenizer, where there is default implementation in `_decode` method.
2) `clean_up_tokenization_spaces` flag in `decode` method now works for some tokenizers, and stop working for other.
3) (Some) Tokenizers that has `self.clean_up_tokenization_spaces=True` change behavior in v5, like `Llama` form the issue.

Removing the behavior and keeping the `clean_up_tokenization_spaces` flag in the `decode` is unintuitive, because we need to check does it _really_ work for each tokenizer we use.

I created the issue about it earlier: https://github.com/huggingface/transformers/issues/42898

itazap:
yes it is intentional that we move away from using `clean_up_tokenization_spaces` in v5, that is why `TokenizersBackend` `_decode` does not use it (and doesn't need to by default). If it is necessary it is part of the _decode function like for Luke: 

`https://github.com/huggingface/transformers/blob/c8bc4dea4c24037cafa38a10ffd0d583b35441a2/src/transformers/models/luke/tokenization_luke.py#L420`

so the extra space in the original example is expected !

PR #42823, opened 2025-12-11
Add an integration test for LASR using pipe and chunked decoding

kho:
# What does this PR do?

- Update the goldens for `test_model_integration_batched`
- Add an integration test for LASR using pipe and chunked decoding.

This is separated from https://github.com/huggingface/transformers/pull/42720 per request from @eustlb.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [x] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [x] Did you write any new necessary tests?

PR #42920, opened 2025-12-17
Fix audio pipelines

zucchini-nlp:
# What does this PR do?

Attribute `_processor_class` was deleted from subprocessors but there are still places where we try to access it. This PR deletes  `_processor_class` everywhere and fixes failing audio pipeline tests


src/transformers/pipelines/__init__.py
```diff
@@ -986,12 +986,13 @@ def pipeline(
                 feature_extractor = AutoFeatureExtractor.from_pretrained(
                     feature_extractor, _from_pipeline=task, **hub_kwargs, **model_kwargs
                 )
+                config_dict, _ = FeatureExtractionMixin.get_feature_extractor_dict(
+                    pretrained_model_name_or_path,
```
pcuenca:
```suggestion
                    pretrained_model_name_or_path or model_name,
```

otherwise `pipeline("automatic-speech-recognition")` (without specifying a model) would fail, as `pretrained_model_name_or_path` will be `None`.

pcuenca (approved):
Loos good to me with my suggestion!

Tested on:

- Default pipeline (no model specified)
- [google/medasr](https://huggingface.co/google/medasr), recently released
- [patrickvonplaten/wav2vec2-base-100h-with-lm](https://huggingface.co/patrickvonplaten/wav2vec2-base-100h-with-lm), whose processor class is `Wav2Vec2ProcessorWithLM` and follows a slightly different code path.

pcuenca:
As commented in Slack, I would support merging this with my suggestion (and iterating later, if we need to) because it fixes the ASR pipelines in `main` and allows #42823 to proceed.

cc @eustlb too.

PR #42921, opened 2025-12-17
[CB] Support the `num_return_sequences` argument

remi-or:
## Summary
This PR adds the options to fork requests during continuous batching, which duplicates the request and uses as much as possible the existing cache. This is then leveraged to make the `num_return_sequences` argument available in CB. 
This PR enables parallel decoding, which will be useful for RL workflows.

## Performance 
| Samples | Attention         | Add Prefix | Source  | Duration (s) | Generated tokens | Throughput (tok/s) |
|---------|-------------------|------------|---------|--------------|------------------|--------------------|
| 100     | flash_attention_2 | False      | With PR | 6.82         | 17480            | 2563.24            |
| 100     | flash_attention_2 | False      | On main | 6.91         | 17698            | 2562.81            |
| 100     | sdpa                    | False      | With PR | 21.77        | 17637            | 810.33             |
| 100     | sdpa                    | False      | On main | 21.81        | 17234            | 790.03             |
| 500     | flash_attention_3 | True       | With PR | 16.4         | 113054           | 6895.14            |
| 500     | flash_attention_3 | True       | On main | 16.73        | 112333           | 6715.63            |
* `--compile` is always on, thus the number of tokens varies between runs because the compiled kernels do

## Tests
Tests pass, including the one added to test the feature.

## Sanity check
Looks good


src/transformers/generation/continuous_batching/cache.py
```diff
@@ -386,6 +386,28 @@ def mark_shareable_blocks_as_complete(self, state: RequestState) -> None:
                     prompt_ids=(state.initial_tokens + state.generated_tokens),
                 )
 
+    def copy_cache(self, source_blocks: list[int], forked_blocks: list[int]) -> None:
```
ArthurZucker, line 391: very interesting, I would assume we want to delay requests that are getting forked to the next batch to do this async (I might be wrong).

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -386,6 +386,28 @@ def mark_shareable_blocks_as_complete(self, state: RequestState) -> None:
...
+        """Copy the cache from the source blocks to the forked blocks."""
+        source_blocks = torch.tensor(source_blocks, device=self.device, dtype=torch.int32)
+        forked_blocks = torch.tensor(forked_blocks, device=self.device, dtype=torch.int32)
```
ArthurZucker, line 394: this is allocating memeory + might need a sync (tensor of list -> sync cpu GPU) we wanna avoid that

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -386,6 +386,28 @@ def mark_shareable_blocks_as_complete(self, state: RequestState) -> None:
...
+        for key_cache, value_cache in zip(self.key_cache, self.value_cache):
+            key_cache = key_cache.view(-1, self.block_size, self.num_key_value_heads, self.head_dim)
+            value_cache = value_cache.view(-1, self.block_size, self.num_key_value_heads, self.head_dim)
+            key_cache[forked_blocks] = key_cache[source_blocks]
+            value_cache[forked_blocks] = value_cache[source_blocks]
+            # FIXME: should be one copy for al CMs with only the changing blocks
+            # FIXME: even once per fork batch
```
ArthurZucker: copy should be async as well (async=True)

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -386,6 +386,28 @@ def mark_shareable_blocks_as_complete(self, state: RequestState) -> None:
...
+
+    def fork_request(self, state: RequestState, new_request_id: str) -> RequestState:
+        """Fork a request into a new request. The new request is created by copying the state and updating the
+        request_id."""
+        new_state = state.fork(new_request_id)
+        for cm in self.group_cache_managers:
+            source_blocks, forked_blocks = cm.fork_blocks(state.request_id, new_state.request_id, self._block_manager)
+            self.copy_cache(source_blocks, forked_blocks)
```
ArthurZucker: same here we should "schedule" the copy. Remember we are in Python and the gil is killing us

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -123,6 +123,27 @@ def get_free_blocks(
         # In both cases, we return the allocated block ids
         return allocated_block_ids
 
+    def fork_blocks(self, source_blocks: list[int], shareable: bool, group_id: int) -> list[int] | None:
+        """Fork a given list of (source_blocks) into a new list of forked_blocks. If the blocks are (shareable), we
+        reference the existing blocks when they are complete. Otherwise, we allocate new blocks if possible. The
+        (group_id) of the layer group the blocks belong to is also needed."""
```
ArthurZucker: need do that shows in / out

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -594,6 +600,16 @@ def update_batch(self) -> None:
...
 
+        # If some requests need to be forked, we do it now
+        while self.scheduler._requests_to_fork:
+            state = self.scheduler._requests_to_fork.pop()
+            num_children = state.num_children
+            state.num_children = 0
+            for i in range(num_children):
+                # FIXME: if fork cant be done, create a new pending request without forking
+                new_request = self.cache.fork_request(state, f"{state.request_id}__child#{i}")
+                self.scheduler.active_requests[new_request.request_id] = new_request
```
ArthurZucker: same here, we should make that async IMO (new status "FORKING" -> wait until forked? IDK but we need to bench a tad bit

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -760,29 +776,35 @@ def __init__(
...
+        self.batch_processor: ContinuousBatchProcessor | None = None
         self._generation_thread = None
         self._request_counter = 0
         self._request_lock = threading.Lock()
-        self.model.generation_config.top_p = None
+
+        # Generation config related arguments
+        generation_config = model.generation_config if generation_config is None else generation_config
+        self.generation_config = generation_config
+        self.log_prob_generation = getattr(generation_config, "log_prob_generation", False)
```
ArthurZucker, line 811: another PR for this maybe?

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -929,6 +952,7 @@ def add_request(
         state = RequestState(
             request_id=request_id,
             initial_tokens=list(input_ids),
+            num_children=self.num_return_sequences - 1,
```
ArthurZucker, line 966: name is weird but ok in the context of forking etc

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -1226,24 +1250,26 @@ def generate_batch(
...
+            block=True,
+            timeout=5,
+        )
+        logging_cm = logging_redirect_tqdm([logger])
+        pbar_cm = tqdm(
+            total=num_requests,
+            disable=(not progress_bar),
+            desc=f"Solving {num_requests} requests",
+            unit="request",
+        )
```
ArthurZucker, line 1281: you can create a get cm func?

src/transformers/generation/continuous_batching/requests.py
```diff
@@ -227,3 +230,25 @@ def to_generation_output(self):
             error=self.error,
             timestamps=self.timestamps,
         )
+
+    def fork(self, new_request_id: str) -> "RequestState":
+        """Fork the request into a new request with the same state expect for request_id, created_time and lifespan."""
+        return RequestState(
```
ArthurZucker: there has to be a better way to do this no? deepcopy?

src/transformers/generation/continuous_batching/cache.py
remi-or: I tried playing around with this and I was surprised this is the fastest alternative, which makes no sense to me. Will leave a TODO to deep dive later

src/transformers/generation/continuous_batching/cache.py
remi-or: cf. above

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: For the log probs? Sure, I will add it to the todo list

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: not sure I understand this -- what would it return? It seems `self.continuous_batching_context_manager(...)` is the get cm

src/transformers/generation/continuous_batching/cache.py
remi-or: On the opposite, to maximize prefix sharing, you want to schedule those request asap. But there might be something to the idea that we can do much of the cpu-side of forking in the async. The issue is that there will always be a copy of the cache, hence GPU intervenes, but maybe it can be done in a side stream.
I think the best compromise is to add the feature now and later, when we get to CPU asynchronous-ness, we can add the FORKING status to let the scheduler know we need those requests to not be scheduled -- until the cache has been copied.

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: Not sure what you means by this sorry!

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: this has changed to be done in batch, which without asynchronous mode is the best we can do for CPU side.

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: yes thats what i was going for

src/transformers/generation/continuous_batching/requests.py
remi-or: I will macro benchmark the two and it the deepcopy is on par / better we will move to that, agreed it's better looking

src/transformers/generation/continuous_batching/cache.py
remi-or: Since the copy is device to device, I think the best we can do for now is one copy as is the case right now. Plus, I think pytorch works in async in that case, ie. CPU operations continue after the copy is launched

src/transformers/generation/continuous_batching/requests.py
remi-or: Changed to deepcopy!

src/transformers/generation/continuous_batching/cache_manager.py
ArthurZucker: sorry I mean doc should help us more, showing example with input output!

src/transformers/generation/continuous_batching/cache.py
ArthurZucker: okayyy

src/transformers/generation/continuous_batching/cache.py
ArthurZucker: yep makes sense!

src/transformers/generation/continuous_batching/cache.py
ArthurZucker: Got it thanks for checking!

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -243,6 +282,38 @@ def get_write_indices(self, request_id: str, past_length: int, query_length: int
     def get_seqlens_k(self, request_id: str, past_length: int, query_length: int) -> tuple[str, int]:
         """Returns the attention type of the cache allocator and the key sequence length for the given request_id."""
 
+    def fork_blocks(
+        self, parent_request_id: str, children_request_ids: list[str], block_manager: BlockManager
+    ) -> tuple[list[int], list[int]]:
+        """Forks the cache blocks of a (parent_request_id) to a list of (children_request_ids). To manage the blocks,
+        the (block_manager) is used. When forking, the child's block are either shared with the parent, or they need to
```
ArthurZucker, line 306:
```suggestion
        the block_manager is used. When forking, the child's block are either shared with the parent, or they need to
```

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -243,6 +282,38 @@ def get_write_indices(self, request_id: str, past_length: int, query_length: int
...
+            num_forks=len(children_request_ids),
+            shareable=self.uses_block_sharing,
+            group_id=self._index,
+        )
+        if list_forked_blocks is None:
+            raise ValueError(f"Failed to fork blocks for request {parent_request_id}")
+
+        # Update the block table for all children requests
+        for children_request_id, forked_blocks in zip(children_request_ids, list_forked_blocks):
+            self.block_table[children_request_id] = forked_blocks
```
ArthurZucker, line 329: it feels like we should not need to iterate twice on the `list_forked_blocks` if `block_manager.fork_blocks` updates the block table on the fly. But emcapsulation might be better this way?

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -594,6 +600,29 @@ def update_batch(self) -> None:
...
+        # If some requests need to be forked, we do it now
+        copy_source, copy_destination = [], []
+        while self.scheduler._requests_to_fork:
+            # Get the number of children and reset it so it's not forked again
+            state = self.scheduler._requests_to_fork.pop()
+            num_children = state.num_children
+            state.num_children = 0
+            # Create the new request
+            new_request_ids = [f"{state.request_id}__child#{i}" for i in range(num_children)]
+            new_requests = [state.fork(new_request_id) for new_request_id in new_request_ids]
```
ArthurZucker:
```suggestion
            for new_request_id in new_request_ids:
                self.scheduler.active_requests[new_request.request_id] = state.fork(new_request_id)
```

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -594,6 +600,29 @@ def update_batch(self) -> None:
...
+            # Fork the cache
+            copy_src, copy_dst = self.cache.fork_request(state.request_id, new_request_ids)
+            copy_source.extend(copy_src)
+            copy_destination.extend(copy_dst)
+            # Add the new requests to the scheduler
+            for new_request in new_requests:
+                self.scheduler.active_requests[new_request.request_id] = new_request
```
ArthurZucker:
```suggestion
```

src/transformers/generation/continuous_batching/continuous_api.py
ArthurZucker: Yeah sorry I don't remember what I wanted here LGTM

src/transformers/generation/continuous_batching/requests.py
```diff
@@ -227,3 +231,11 @@ def to_generation_output(self):
             error=self.error,
             timestamps=self.timestamps,
         )
+
+    def fork(self, new_request_id: str) -> "RequestState":
+        """Fork the request into a new request with the same state expect for request_id, created_time and lifespan."""
+        new = deepcopy(self)
```
ArthurZucker: I did not check, IDT we need to deep copy everything (like `initial_tokens`, and etc can be shared refs and we should actually ensure that)

ArthurZucker (approved):
LGTM! `def fork(self, new_request_id: str) -> "RequestState":` should be optimized as much as possible,

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: Added an ascii table

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: I use parenthesis to denote arguments of the function, would be weird to change convention midway

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: You are right, moving the check to inside the loop. Thanks!

remi-or: Oh my bad, i mistook your point. We iterate twice on that list indeed, because `block_manager.fork_blocks` does not update the block table directly. I tried making it handle that part and it led to a messy function that was not all that readable. imo it's best to leave it clear for now and re-visit it if it's a hotspotfor optimization, which it might be? then again it's just an additional iteration on a small list.

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: Good idea, done

src/transformers/generation/continuous_batching/requests.py
remi-or: Ok, makes sense. I deepcopied the attributes that are mutable and changed during generation

src/transformers/generation/continuous_batching/cache_manager.py
ArthurZucker: yeah yeah, I am just trying to avoid any extra for looop when we can, from reading the code it appeared to be removable, but if its not no worries!

src/transformers/generation/continuous_batching/cache_manager.py
ArthurZucker: ah ok did not know!

PR #42922, opened 2025-12-17
Load generation config from nested configs

### LINKED ISSUE #42794: `decoder_start_token_id` or `bos_token_id` has to be defined for encoder-decoder generation.
### System Info

latest transformers

### Who can help?

@zucchini-nlp 

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

```python
import torch

pipe = pipeline(
    "document-question-answering",
    model="naver-clova-ix/donut-base-finetuned-docvqa",
    dtype=torch.float16,
)

image = "https://huggingface.co/spaces/impira/docquery/resolve/2359223c1837a7587402bda0f2643382a6eefeab/invoice.png"
question = "What is the invoice number?"

result = pipe(image=image, question=question)
print(result)
```

error:
```
  File "/home/jiqingfe/transformers/test_dqa.py", line 13, in <module>
    result = pipe(image=image, question=question)
  File "/home/jiqingfe/transformers/src/transformers/pipelines/document_question_answering.py", line 310, in __call__
    return super().__call__(inputs, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jiqingfe/transformers/src/transformers/pipelines/base.py", line 1278, in __call__
    return next(
           ^^^^^
  File "/home/jiqingfe/transformers/src/transformers/pipelines/pt_utils.py", line 126, in __next__
    item = next(self.iterator)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/jiqingfe/transformers/src/transformers/pipelines/pt_utils.py", line 271, in __next__
    processed = self.infer(next(self.iterator), **self.params)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jiqingfe/transformers/src/transformers/pipelines/base.py", line 1185, in forward
    model_outputs = self._forward(model_inputs, **forward_params)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jiqingfe/transformers/src/transformers/pipelines/document_question_answering.py", line 468, in _forward
    model_outputs = self.model.generate(**model_inputs, **generate_kwargs)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/miniforge3/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/jiqingfe/transformers/src/transformers/generation/utils.py", line 2551, in generate
    self._prepare_special_tokens(generation_config, kwargs_has_attention_mask, device=device)
  File "/home/jiqingfe/transformers/src/transformers/generation/utils.py", line 2145, in _prepare_special_tokens
    raise ValueError(
ValueError: `decoder_start_token_id` or `bos_token_id` has to be defined for encoder-decoder generation.
```

### Expected behavior

Cannot locate which PR caused this regression because too many errors recently. The transformers 4.57.3 works well on the script.

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/42794. The issue was that model's generation params are saved in text config, but we couldn't call `config.get_text_config()` on a dict object. We can delete the second `load_pretrained` because each model gets a default generation config from model config in `__init__`

We could technically get the text config from dict, but I don't see a reason for adding more code if we can delete redundant blocks


src/transformers/generation/utils.py
```diff
@@ -407,17 +407,11 @@ def adjust_generation_fn(
                     **repo_loading_kwargs,
                 )
             except OSError:
+                # `self` already has a generation config created from model config so we can simply raise a warning
                 logger.info(
                     "Generation config file not found, using a generation config created from the model config."
                 )
```
vasqu: Super nit: But it's an info, not a warning (or do we want to increase this to a warning)

tests/generation/test_utils.py
```diff
@@ -4401,7 +4400,7 @@ def test_generate_vision2text_conditioning(self):
         """Test that `decoder_input_ids` can be used to condition the generation in vision-to-text models"""
         pixel_values = floats_tensor((2, 3, 30, 30))
         conditioning_input = torch.tensor([[10], [10]])  # this should be the 2nd output token, after the BOS token
-        model = AutoModelForVision2Seq.from_pretrained(
+        model = AutoModelForImageTextToText.from_pretrained(
```
vasqu, line 4403: Looks unrelated? Nothing to change just noticed

src/transformers/generation/utils.py
```diff
@@ -407,17 +407,11 @@ def adjust_generation_fn(
...
-                self.generation_config = GenerationConfig.from_pretrained(
-                    pretrained_model_name_or_path,
-                    config_file_name="config.json",
-                    _from_auto=from_auto_class,
-                    _from_pipeline=from_pipeline,
-                    _from_model_config=True,
-                    **repo_loading_kwargs,
-                )
```
vasqu, line 420: Just to double check with https://github.com/huggingface/transformers/pull/41695#discussion_r2534719205

Do you remember which models might've been affected here. Would be nice to double check

vasqu (approved):
Nice thx for the test just nits and let's double-check possibly affected models (with run-slow or so)

src/transformers/generation/utils.py
zucchini-nlp: oh, didn't notice! Nah, I think we don't need a warning, it's a common situation when no generation config is saved

tests/generation/test_utils.py
zucchini-nlp: the mapping was deprecated in favor of `AutoModelForImageTextToText`, so just making sure we don't use deprecated stuff in tests

src/transformers/generation/utils.py
zucchini-nlp: Yeah, it is bart, t5 and similar models. I will trigger slow tests

zucchini-nlp:
run-slow: t5, bart, mbart, umt5

zucchini-nlp:
super weird, I am getting a failure with these test cases even on main branch. Will try also by ssh-ing to runners

edit: i am blind, i was using an incorrect python env and thus checking against this same branch 🤦🏻

zucchini-nlp:
run-slow: t5, bart, mbart, umt5, llama, gemma3, mistral

zucchini-nlp:
run-slow: t5, bart, mbart, umt5, llama, gemma3, mistral

zucchini-nlp:
Didn't see that slow tests were triggered. All pass, merging!

PR #42926, opened 2025-12-17
Fp8 dq

SunMarc:
# What does this PR do?

This PR extend the dq when users don't have compute capability >= 8.9 for fp8.


src/transformers/quantizers/quantizer_finegrained_fp8.py
```diff
@@ -29,27 +29,27 @@ def validate_environment(self, *args, **kwargs):
         if not is_accelerate_available():
             raise ImportError("Loading an FP8 quantized model requires accelerate (`pip install accelerate`)")
 
-        if self.quantization_config.dequantize:
-            return
-
         if not torch.cuda.is_available() and not is_torch_xpu_available():
-            if self.pre_quantized and not self.quantization_config.dequantize:
+            if self.pre_quantized:
```
MekkCyber, line 36: yes I agree we should remove the dequantization condition from the first `if` statement, but I think we should add an other `if` statement inside to only show the warning if the user is not dequantizing

MekkCyber (approved):
Makes sense!

src/transformers/quantizers/quantizer_finegrained_fp8.py
SunMarc: oh yeah indeed, I changed a bit the logic

PR #42928, opened 2025-12-17
fix(tvp): add missing type_vocab_size parameter to TvpConfig

majiayu000:
## Summary
- Fixes #42925
- The `type_vocab_size` parameter was missing from `TvpConfig`, causing `AttributeError` when instantiating `TvpModel`
- Added the parameter with default value of 2, which matches all TVP models on Hugging Face Hub

## Test plan
- [x] Existing tests should pass (this adds a default value that matches existing models)


yonigozlan (approved):
Hello @majiayu000 ! Thanks for contributing, looks good to me. Strange that this issue wasn't raised before

majiayu000:
Glad I could help.

PR #42929, opened 2025-12-17
Stop ContinuousBatchingManager.request_id_iter after terminal outputs

antznette1:
## Summary 
ContinuousBatchingManager.request_id_iter() could keep polling after a request had already reached a terminal state. This PR makes the iterator stop once it yields a terminal GenerationOutput (`FINISHED` or `FAILED`), and also exit when the generation thread is no longer alive and there are no more results to drain.

## Changes
- Stop iterating after emitting `RequestStatus.FINISHED` or `RequestStatus.FAILED` for the request.
- Add a unit test that verifies request_id_iter() terminates after a terminal output, even if the manager thread is still alive.

### DIFF
--- a/src/transformers/generation/continuous_batching/continuous_api.py
+++ b/src/transformers/generation/continuous_batching/continuous_api.py
@@ -997,12 +997,19 @@ class ContinuousBatchingManager:
     def request_id_iter(self, request_id: str) -> Generator[GenerationOutput]:
         """Iterate over results matching a specific request id as they become available."""
         request_cancelled = False
-        while self._generation_thread is not None and self._generation_thread.is_alive() and not request_cancelled:
-            result = self.get_result(request_id=request_id, timeout=0.1)
+        while not request_cancelled:
+            thread_alive = self._generation_thread is not None and self._generation_thread.is_alive()
+            timeout = 0.1 if thread_alive else 0
+
+            result = self.get_result(request_id=request_id, timeout=timeout)
             if result is not None:
                 yield result
+                if result.status in {RequestStatus.FINISHED, RequestStatus.FAILED}:
+                    return
             if self.batch_processor is not None:
                 request_cancelled = self.batch_processor.scheduler.request_is_cancelled(request_id)
+            if not thread_alive and result is None:
+                return
 
     @traced
     def _generation_step(self) -> None:
--- a/tests/generation/test_continuous_batching.py
+++ b/tests/generation/test_continuous_batching.py
@@ -15,6 +15,7 @@
 import gc
 import itertools
 import unittest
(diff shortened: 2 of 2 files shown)


pythongiant (commented):
I feel like the requeue logic in get_result() is a bug waiting to happen.

I feel like something like this would make more sense, right?

```python
def request_id_iter(self, request_id: str):
    while True:
        if not self.is_running() and self.output_queue.empty():
            return
        result = self.get_result(request_id=request_id, timeout=0.1)
        if result is not None:
            yield result
            if result.status in {RequestStatus.FINISHED, RequestStatus.FAILED}:
                return
  ```

**Comment by antznette1** [comment]
> I feel like the requeue logic in get_result() is a bug waiting to happen.
> 
> I feel like something like this would make more sense, right?
> 
> ```python
> def request_id_iter(self, request_id: str):
>     while True:
>         if not self.is_running() and self.output_queue.empty():
>             return
>         result = self.get_result(request_id=request_id, timeout=0.1)
>         if result is not None:
>             yield result
>             if result.status in {RequestStatus.FINISHED, RequestStatus.FAILED}:
>                 return
> ```

Thanks for the suggestion — I agree the current “requeue” behavior in `get_result()` is a bit fragile (it can cause churn / unfairness if the queue has multiple request_ids and a consumer keeps asking for one specific id).
My change was intentionally scoped to make `request_id_iter()` terminate correctly once a terminal `GenerationOutput` is observed, without touching the broader queue semantics. But your proposed loop makes sense and is cleaner than my current `thread_alive/timeout=0` handling.
I’m happy to update `request_id_iter()` to follow the pattern you suggested:
```
def request_id_iter(self, request_id: str):
    while True:
        if not self.is_running() and self.output_queue.empty():
            return
        result = self.get_result(request_id=request_id, timeout=0.1)
        if result is not None:
            yield result
            if result.status in {RequestStatus.FINISHED, RequestStatus.FAILED}:
                return
```

Separately (or in a follow-up PR if you prefer), I can also take a look at improving/removing the requeue logic in `get_result()` itself (e.g. a per-request buffer or a different retrieval API), since that’s likely the more fundamental fix.

Let me know if you’d like me to adjust `request_id_iter()` in this PR and keep any `get_result()` redesign for a follow-up.

McPatate:
This modifies behaviour in a breaking way, we haven't documented but the iterator is mostly used in scenarios where you never want to stop inference. If you really want to kill the iterator for now you'll have to call `ContinuousBatchingManager.stop` in a separate thread, which will stop iteration.
Your current implementation, IIUC, exits at the first finished request, which completely defeats the purpose of using continuous batching.

PR #42932, opened 2025-12-17
[docs] inference engines

stevhliu:
starts off a new section dedicated to our integrations with inference engines


docs/source/en/community_integrations/sglang.md
```diff
@@ -0,0 +1,52 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# SGLang
```
stevhliu, line 17: maybe @SunMarc can take a look at the SGLang section please 🙂

SunMarc (approved):
Nice, thanks for updating this !

docs/source/en/community_integrations/vllm.md
```diff
@@ -0,0 +1,47 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# vLLM
```
stevhliu, line 17: @hmellor, would you mind taking a look at the vLLM section please? 🙂

docs/source/en/community_integrations/vllm.md
```diff
@@ -0,0 +1,47 @@
...
+
+Pass `--model-impl transformers` to the `vllm serve` command for online serving.
+
+```bash
+vllm serve meta-llama/Llama-3.2-1B \
+    --task generate \
+    --model-impl transformers
+```
+
+vLLM fetches `config.json` from the Hub or your Hugging Face cache and reads `model_type` to load the config class. If the model type isn't in vLLM's internal registry, it falls back to [`AutoConfig.from_pretrained`]. With `model_impl="transformers"`, vLLM passes the config to [`~AutoModelForCausalLM.from_pretrained`] to instantiate the model instead of using vLLM's native model classes.
```
hmellor: vLLM's config registry is a bit of a hack for when custom models don't write compatible config classes. I wouldn't say this is the primary way that configs are loaded.

It'd say it's more like:

We  use `AutoConfig.from_pretrained` to load the config, then from there we check `architectures` against the internal model registry to know which vLLM model class to load.

Then if the model isn't found in the internal registry we fall back to `AutoModel.from_config` to load the model with the Transformers modelling backend.

You can use `model_impl="transformers"` to go bypass the vLLM model registry and go straight to Transformers.

docs/source/en/community_integrations/vllm.md
```diff
@@ -0,0 +1,47 @@
...
+
+[`AutoTokenizer.from_pretrained`] loads tokenizer files. Some tokenizer internals are cached to reduce repeated overhead during inference. Model weights download from the Hub in safetensors format.
+
+Transformers handles the model and forward pass. vLLM provides the high-throughput serving stack.
```
hmellor: Kind of, most of the modules (moe/attention interface/linear/rms norm/vocab embedding/lm head) are replaced with vLLM's modules which is what gives us access to vLLM's high-throughput serving and fancy parallelisation features

docs/source/en/community_integrations/vllm.md
```diff
@@ -0,0 +1,47 @@
...
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# vLLM
+
+[vLLM](https://github.com/vllm-project/vllm) is a high-throughput inference engine for serving LLMs at scale. It continuously batches requests and keeps KV cache memory compact with PagedAttention.
+
+Set `model_impl="transformers"` to load a Transformers modeling backend.
```
hmellor:
```suggestion
Set `model_impl="transformers"` to load a model using the Transformers modeling backend.
```

hmellor (approved):
LGTM, thanks for this!

LysandreJik (approved):
Thank you Steven!

PR #42935, opened 2025-12-17
[Tests] Fix CompressedTensors tests

kylesayrs:
## Purpose ##
* Fix failing compressed tensors tests which referenced deleted models

## Changes ##
* Refactor compressed tensors tests to check for perplexity, rather than exact output matches
* Update model stubs to use up-to-date models

## Testing ##
* Newly refactored tests pass locally

### Suggested Reviewers ###
@SunMarc @MekkCyber I've broken out the testing changes from #42887. The main functionality of that original PR is now pending a CT release, and we can land that later.

### DIFF
--- a/tests/quantization/compressed_tensors_integration/test_compressed_tensors.py
+++ b/tests/quantization/compressed_tensors_integration/test_compressed_tensors.py
@@ -1,15 +1,8 @@
 import gc
 import unittest
-from unittest import skip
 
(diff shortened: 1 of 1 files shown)


MekkCyber (approved):
Okay let's do instead in the meantime, thank you

SunMarc (approved):
Thanks !

Issue #42890, opened 2025-12-16
tests/models/sam_hq/test_modeling_sam_hq.py::SamHQModelIntegrationTest may fail since a lot of cases are lack of set_seed()


sywangyi:
### System Info

transformers built from latest main.

### Who can help?

@ydshieh 
similar issue may occur in  tests/models/sam/test_modeling_sam.py

see https://github.com/sywangyi/transformers/blob/main/src/transformers/models/sam_hq/modeling_sam_hq.py#L1077
 randn() is used in positional_embedding, so you need to set_seed() before modeling loading to guarantee the stable output. or else accuracy check will fail.

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

pytest tests/models/sam_hq/test_modeling_sam_hq.py::SamHQModelIntegrationTest

### Expected behavior

all pass in cuda.
actual: 11 failed, 3 passed,

Rocketknight1:
I think this is a general issue - any test that doesn't load a pretrained model will randomly initialize weights, and so there will be test variability. This is the root of a lot of flaky tests for us.

However, I can confirm the SamHQ slow tests are failing for me too - if you can fix it by setting a seed, we'd welcome a PR!

engmohamedsalah:
I'm working on this issue and have opened PR #43211 to fix the flaky SAM-HQ integration tests.

**Root Cause**: The `positional_embedding` is randomly initialized with `torch.randn()` and excluded from checkpoint loading, causing non-deterministic test results.

**Solution**: Added `set_seed(0)` in the `setUp()` method of `SamHQModelIntegrationTest` to ensure reproducible tests.

The fix follows KISS and DRY principles with minimal code changes (8 lines).

Rocketknight1:
After a little more investigation I think the issue might actually be with how we load the positional embedding - we only need to set a seed because of how it's being randomly reinitialized. cc @nielsrogge @molbap @yonigozlan, see #43133 - should that positional embedding be shared?

engmohamedsalah:
I've identified and fixed the root cause! This was a **weight loading bug** introduced in commit `0b369802cf`.

## The Bug

[...]

PR #42938, opened 2025-12-18
make tests/models/sam_hq/test_modeling_sam_hq.py::SamHQModelIntegrati…

sywangyi:
…onTest pass in cuda and xpu

@Rocketknight1


sywangyi:
fix https://github.com/huggingface/transformers/issues/42890

Rocketknight1:
run-slow: sam_hq

Rocketknight1:
cc @ydshieh! I confirmed that the SamHQ slow tests were failing locally for me before this

yao-matrix:
@ydshieh , could you pls help review this, thx very much.

yao-matrix:
@ydshieh , happy new year! could you help take a look at this PR? Thx very much.

ydshieh:
Hi. Thank you for diving into this.

The checkpoint

https://huggingface.co/syscv-community/sam-hq-vit-base/blob/main/model.safetensors

has the weight `shared_image_embedding.positional_embedding`, so it's not very clear why we need to set seed, as the model is loaded from that checkpoint

> model = SamHQModel.from_pretrained("syscv-community/sam-hq-vit-base")

Do you have any idea what actually causes the issue? Maybe we don't need seed, and the original failure is just the value was wrong  but it's not really changing on each run? cc @Rocketknight1 as you mentioned this particular

https://github.com/huggingface/transformers/issues/42890#issuecomment-3660799382

Anyway, it's too much time consuming to change again, so I am happy to approve and merge as it, but just add a comment to the seed to say it's not really necessary but just for convenience (if it turns out not necessary).

If possible, let's avoid use anything like

> self.assertTrue(torch.allclose(scores[0][0][-1], torch.tensor(0.4482), atol=2e-4))

but `torch.testing.assert_close` instead. But if you have no bandwidth, it's fine.

sywangyi:
see https://github.com/huggingface/transformers/issues/42890 description. 

ydshieh:
@sywangyi I understand randn() is used in positional_embedding. But in the tests, we are not creating a model from scratch but loading from a mode checkpoint 

> model = SamHQModel.from_pretrained("syscv-community/sam-hq-vit-base")

which has the weight 

> shared_image_embedding.positional_embedding

so eventually, the loaded model has the same weight across run.

Unless there is any other random operation involved,  we shouldn't have the results being different across runs.

So my question is:

- Do you observe the results being different across different multiple runs.
  - if this is the case, the root cause might not be the   randn() is used in positional_embedding 
- Or the results are consistent, but just not matching the expected values on the main branch.

sywangyi:
it's "prompt_encoder.shared_embedding.positional_embedding" [https://github.com/huggingface/transformers/blob/main/src/transformers/models/sam_hq/modeling_sam_hq.py#L1238]
this param is random, however, I found set_seed also could not guarantee it's fixed. so maybe we need to fix the param.

change seed value now could make the result being different

> @sywangyi I understand randn() is used in positional_embedding. But in the tests, we are not creating a model from scratch but loading from a mode checkpoint
> 
> > model = SamHQModel.from_pretrained("syscv-community/sam-hq-vit-base")
> 
> which has the weight
> 
> > shared_image_embedding.positional_embedding
> 
> so eventually, the loaded model has the same weight across run.
> 
> Unless there is any other random operation involved, we shouldn't have the results being different across runs.
> 
> So my question is:
> 
> * Do you observe the results being different across different multiple runs.
>   
>   * if this is the case, the root cause might not be the   randn() is used in positional_embedding
> * Or the results are consistent, but just not matching the expected values on the main branch.

ydshieh:
Hi @sywangyi 

> it's "prompt_encoder.shared_embedding.positional_embedding"

Thank you a lot for identifying this ! Great detective work 🙏 .

Let me ping a team member to see what's going on with this missing weight on the hub repository, or it's designed to be so.
I will come back to you as soon as possible.

ydshieh:
I verified and this is more involved which requires a fix in the modeling code (not simply the test file).

We will try to work on this next week 🙏

sywangyi:
PR in modeling has been merged

PR #42940, opened 2025-12-18
[Quantization] Misc tests fixes

MekkCyber:
# What does this PR do?

Fixing various Quantization tests

### DIFF
--- a/src/transformers/conversion_mapping.py
+++ b/src/transformers/conversion_mapping.py
@@ -142,12 +142,12 @@ def _build_checkpoint_conversion_mapping():
     if hasattr(torch.nn.utils.parametrizations, "weight_norm"):
         mapping["legacy"] += [
             WeightRenaming(
-                source_patterns="weight_g",
-                target_patterns="parametrizations.weight.original0",
+                source_patterns=".weight_g$",
+                target_patterns=".parametrizations.weight.original0",
             ),
             WeightRenaming(
-                source_patterns="weight_v",
-                target_patterns="parametrizations.weight.original1",
+                source_patterns=".weight_v$",
+                target_patterns=".parametrizations.weight.original1",
             ),
         ]
     else:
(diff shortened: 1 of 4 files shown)


src/transformers/core_model_loading.py
```diff
@@ -342,7 +342,8 @@ def rename_source_key(self, source_key: str) -> tuple[str, str | None]:
         matching_group_name = next(name for name, val in match_object.groupdict().items() if val is not None)
         source_pattern_that_matched = self.source_patterns[int(matching_group_name[1:])]
         # If we matched, we always replace with the first target pattern, in case we have several (one to many transform)
-        replacement = self.target_patterns[0]
```
MekkCyber, line 345: in reverse ops, I think we should remove the $ and ^ characters if used as regex patterns in source_patterns (which become target patterns during saving) @Cyrilvallez

src/transformers/conversion_mapping.py
```diff
@@ -142,12 +142,12 @@ def _build_checkpoint_conversion_mapping():
     if hasattr(torch.nn.utils.parametrizations, "weight_norm"):
         mapping["legacy"] += [
             WeightRenaming(
-                source_patterns="weight_g",
-                target_patterns="parametrizations.weight.original0",
+                source_patterns=".weight_g$",
+                target_patterns=".parametrizations.weight.original0",
```
MekkCyber, line 146: we need this for `fp-quant` since it has a param called `weight_global_scale` that will get replaced if we don't use a `$` delimiter in the regex

SunMarc (approved):

src/transformers/core_model_loading.py
Cyrilvallez: I see! The `^` is already being removed! The correct place to do it is https://github.com/huggingface/transformers/blob/main/src/transformers/core_model_loading.py#L302-L316 those lines 🤗 So we keep all those transforms at the same location

MekkCyber: oh indeed thanks! updated that 👍

Issue #38066, opened 2025-05-10
`AutoModel.from_pretrained(...)` (with explicit `device_map` unset) fails under `with torch.device("meta")` with PyTorch 2.6.0 and 2.7.0


vadimkantorov:
```python
# from torch.nn.attention.flex_attention import BlockMask, flex_attention
from transformers import AutoModel
import torch

with torch.device('meta'):
    AutoModel.from_pretrained('Qwen/Qwen2.5-0.5B', trust_remote_code=True)
````

I found this code in the wild in https://github.com/Open-Reasoner-Zero/Open-Reasoner-Zero/blob/f6d1ec77ce2ce18f3d925a1014c9e4d6b4ad3072/orz/ppo/actors.py#L745-L746 (linked issue https://github.com/Open-Reasoner-Zero/Open-Reasoner-Zero/issues/71)

fails with:
```
Sliding Window Attention is enabled but not implemented for `sdpa`; unexpected results may be encountered.
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
[<ipython-input-1-00ba4c43be18>](https://localhost:8080/#) in <cell line: 0>()
      4 
      5 with torch.device('meta'):
----> 6     AutoModel.from_pretrained('Qwen/Qwen2.5-0.5B', trust_remote_code=True)

6 frames
[/usr/local/lib/python3.11/dist-packages/transformers/models/auto/auto_factory.py](https://localhost:8080/#) in from_pretrained(cls, pretrained_model_name_or_path, *model_args, **kwargs)
    569             if model_class.config_class == config.sub_configs.get("text_config", None):
    570                 config = config.get_text_config()
--> 571             return model_class.from_pretrained(
    572                 pretrained_model_name_or_path, *model_args, config=config, **hub_kwargs, **kwargs
    573             )

[/usr/local/lib/python3.11/dist-packages/transformers/modeling_utils.py](https://localhost:8080/#) in _wrapper(*args, **kwargs)
    277         old_dtype = torch.get_default_dtype()
    278         try:
--> 279             return func(*args, **kwargs)
    280         finally:
    281             torch.set_default_dtype(old_dtype)

[...]

PR #42941, opened 2025-12-18
[loading] Really initialize on meta device for huge perf gains

Cyrilvallez:
# What does this PR do?

Follow-up to https://github.com/huggingface/transformers/pull/42309 to really leverage meta device loading. Gives crazy speedups for loading some models, e.g. *about 2.5x on gpt-oss 20b* and *about 3x on the 120b version*

## The issue at hand

Currently, during loading we initialize the model on meta device before loading weights, thanks to `init_empty_weights` from `accelerate`.
However, this context manager has BIG drawbacks:
- everything (every parameter and buffer) is first materialized on cpu before being moved to meta device
- this is extremely inefficient of course, as we only want them on meta -> it wastes time and memory
- all buffers stay on cpu (even the persistent ones, that we are loading again after anyway, so don't need to be there...)

For some models, e.g. gpt-oss, we have the following during loading:

Screenshot 2025-12-19 at 12 11 30

Note how most of the loading time is BEFORE the actual loading of the weights (`_load_pretrained` call), just to initialize parameters that should be on meta device anyway.... 

## What this PR is doing

This PR completely removes `init_empty_weights` in favor of `torch.device("meta")` to really start with a model on meta device, without first putting them on cpu.
We are free to do so since I've merged https://github.com/huggingface/transformers/pull/42309 yesterday, to correctly handle re-initialization of the non-persistent buffers which are put on meta device as well.

The performance gains are immense, the same benchmark as before for gpt-oss shows it:

Screenshot 2025-12-19 at 12 16 00

and we can now see that most of the time in `from_pretrained` is used for the actual weight loading (`_load_pretrained` call), as it should be.

Raw numbers for the following simple benchmark script (from which the above traces are from) on our cluster:

```python
import torch
import time
from viztracer import VizTracer

device = 0

tracer = VizTracer()
tracer.start()

torch.cuda.synchronize()
t0 = time.time()
model = AutoModelForCausalLM.from_pretrained(model_id, device_map=device, dtype=torch.float16)
torch.cuda.synchronize()
dt = time.time() - t0
print(f"Took {dt:.2f} s")

tracer.stop()
tracer.save(output_file="../trace.json")
```

are the following:

BEFORE THIS PR -> ~9.1s
ON THIS PR -> ~3.7s

*Which means a speedup of about 2.5x*.

For the 120B gpt-oss version, we have:

BEFORE THIS PR -> ~32.4s
ON THIS PR -> ~10.7s

or a *speedup of about 3x*

### DIFF
--- a/src/transformers/integrations/aqlm.py
+++ b/src/transformers/integrations/aqlm.py
@@ -14,13 +14,11 @@
 "AQLM (Additive Quantization of Language Model) integration file"
 
 from ..quantizers.quantizers_utils import should_convert_module
-from ..utils import is_accelerate_available, is_torch_available, logging
+from ..utils import is_torch_available, logging
 
-
-if is_accelerate_available():
-    from accelerate import init_empty_weights
 
 if is_torch_available():
+    import torch
     import torch.nn as nn
 
 logger = logging.get_logger(__name__)
@@ -46,7 +44,7 @@ def replace_with_aqlm_linear(model, modules_to_not_convert: list[str] | None = N
     for module_name, module in model.named_modules():
         if not should_convert_module(module_name, modules_to_not_convert):
             continue
-        with init_empty_weights():
+        with torch.device("meta"):
             if isinstance(module, nn.Linear):
                 new_module = QuantizedLinear(
                     module.in_features,
--- a/src/transformers/integrations/awq.py
+++ b/src/transformers/integrations/awq.py
@@ -16,11 +16,8 @@
 from typing import Optional, Union
 
 from ..quantizers.quantizers_utils import should_convert_module
-from ..utils import is_accelerate_available, is_torch_available, logging
+from ..utils import is_torch_available, logging
 
-
-if is_accelerate_available():
-    pass
 
 if is_torch_available():
     import torch
--- a/src/transformers/integrations/awq.py
+++ b/src/transformers/integrations/awq.py
@@ -20,7 +20,7 @@ from ..utils import is_accelerate_available, is_torch_available, logging
 
 
 if is_accelerate_available():
-    from accelerate import init_empty_weights
+    pass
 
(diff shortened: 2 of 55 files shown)


src/transformers/integrations/awq.py
```diff
@@ -97,7 +94,7 @@ def replace_with_awq_linear(
     for module_name, module in model.named_modules():
         if not should_convert_module(module_name, modules_to_not_convert):
             continue
-        with init_empty_weights():
+        with torch.device("meta"):
```
Cyrilvallez, line 97: cc @SunMarc for this change, do you know if this switch on all the quantizers's `replace_with_xxx` is fine? Basically, the only difference is if the quantized layer registers some buffers, they would now be on meta as well. I checked for bnb and there it seems to be alright at least (no buffers)

src/transformers/integrations/aqlm.py
```diff
@@ -14,13 +14,11 @@
 "AQLM (Additive Quantization of Language Model) integration file"
 
 from ..quantizers.quantizers_utils import should_convert_module
-from ..utils import is_accelerate_available, is_torch_available, logging
+from ..utils import is_torch_available, logging
 
 
-if is_accelerate_available():
-    from accelerate import init_empty_weights
```
ArthurZucker, line 21: 💌

ArthurZucker (approved):
An early Christmas gift for everyone

src/transformers/integrations/awq.py
SunMarc: as long as it's not a non persistant buffer, it should be fine !

vadimkantorov: Could be related minor issue:
- https://github.com/huggingface/transformers/issues/38066

PR #42942, opened 2025-12-18
Fix result retrieval starvation and terminate request-scoped iteration on completion

### LINKED ISSUE #42943: Continuous batching: output queue requeue starvation and request-scoped iterator does not terminate on completion
### **Description**

There are two related correctness issues in the continuous batching result consumption logic that can lead to unfairness and non-terminating iterators under concurrent workloads.

#### 1. Starvation and incorrect timeout handling in `get_result`

`ContinuousBatchingManager.get_result` currently retrieves a single item from the shared output queue and immediately re-queues it if the `request_id` does not match, returning `None` afterward. Under concurrent requests, this can lead to:

* Starvation when mismatched outputs are repeatedly re-queued
* Timeout semantics that are not respected, since re-queueing returns early instead of continuing to search within the remaining timeout
* Unfair consumption behavior that depends on queue ordering rather than request progress

This behavior is observable when multiple streaming requests are active and results are interleaved in the output queue.

#### 2. `request_id_iter` does not terminate after normal completion

`request_id_iter` currently exits only when a request is cancelled or the generation thread terminates. For requests that complete normally, the iterator continues polling indefinitely after the final `FINISHED` output has been yielded.

This can result in:

* Infinite iteration loops for request-scoped consumers
* Unexpected blocking behavior in streaming-style usage
* Reliance on caller-side logic to manually stop iteration

The iterator should terminate once a terminal `FINISHED` output is observed for the given request.

---

### **Expected behavior**

* `get_result` should fairly search for a matching result within the specified timeout without starvation or early return.
* `request_id_iter` should stop iterating once the request reaches a terminal finished state, in addition to cancellation or thread termination.

---

### **Proposed fix** 

A minimal, backward-compatible fix (#42942) can:

* Defer re-queuing mismatched outputs until a matching result is found or the timeout expires
* Explicitly terminate `request_id_iter` when a `GenerationOutput` reports a finished state

This preserves existing APIs, streaming semantics, and benchmarking behavior while fixing the correctness issues.

---

### **Environment**

* Transformers version: main
* Feature: continuous batching
* Device: CPU / CUDA (independent of backend)

---

### **Additional context**

These issues are easiest to reproduce with multiple concurrent streaming requests sharing a single `ContinuousBatchingManager`.

pythongiant:
# What does this PR do?

This PR fixes two correctness issues in the continuous batching result consumption logic.

First, it improves `ContinuousBatchingManager.get_result` to avoid starvation and incorrect timeout behavior when multiple requests share a single output queue. Instead of immediately re-queuing mismatched results and returning, the updated logic temporarily defers non-matching outputs while continuing to search for a matching one, reinserting deferred results afterward. This preserves FIFO ordering, ensures fair consumption across requests, and respects the specified timeout under concurrent access.

Second, it fixes `request_id_iter` so that iteration stops when a request reaches a terminal `FINISHED` state, rather than only stopping on cancellation. This prevents infinite iteration after normal request completion and aligns the iterator’s behavior with expected streaming semantics.

These changes are local, backward-compatible, and do not alter the public API or benchmarking behavior.

Fixes #42943 

## Before submitting
- [x] Did you read the contributor guideline, Pull Request section?
- [x] Was this discussed/approved via a Github issue or the forum?

## Who can review?

@remi-or @ArthurZucker @McPatate

### DIFF
--- a/src/transformers/generation/continuous_batching/continuous_api.py
+++ b/src/transformers/generation/continuous_batching/continuous_api.py
@@ -964,28 +964,37 @@ class ContinuousBatchingManager:
         if self.batch_processor is not None:
             self.batch_processor.scheduler.set_request_cancellation(request_id)
 
-    # TODO:handle benchmarking properly when updating / fixing the requeue logic
     def get_result(self, request_id: str | None = None, timeout: float | None = None) -> GenerationOutput | None:
-        """Retrieve one result from the output queue.
-
-        Args:
-            timeout: Maximum time to wait for a result
-
-        Returns:
-            Optional[GenerationOutput]: The result data or None if timeout
-        """
+        # Fast exit: no thread + no pending output
         if self._generation_thread is None and self.output_queue.empty():
             return None
+    
+        deadline = None if timeout is None else perf_counter() + timeout
+        deferred: list[GenerationOutput] = []
+    
         try:
-            result = self.output_queue.get(block=True, timeout=timeout)
-            # NOTE: requeue logic here
-            if request_id is not None and result.request_id != request_id:
-                self.output_queue.put(result)
-                return None
-            return result
-        except queue.Empty:
-            return None
-
+            while True:
+                remaining = None if deadline is None else max(0.0, deadline - perf_counter())
+                if remaining == 0.0:
+                    return None
+    
+                try:
+                    result = self.output_queue.get(timeout=remaining)
+                except queue.Empty:
+                    return None
+    
+                # Match found
+                if request_id is None or result.request_id == request_id:
+                    return result
+    
+                # Defer mismatched result instead of immediately requeuing
+                deferred.append(result)
+    
+        finally:
+            # Reinsert deferred results preserving order
+            for item in deferred:
+                self.output_queue.put(item)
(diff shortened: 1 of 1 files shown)


McPatate:
This isn't how we plan to solve the outstanding issue, although I do appreciate you taking a look.

Given we're in a multi threaded context, I was thinking of holding a hashmap of running requests with their req_id as key, and calling get result would wait til completion or timeout. But this may need some careful attention to avoid breaking things.

Feel free to refactor in a direction that looks more like how we intend to address the problem, otherwise we'll close the PR.

pythongiant:
> This isn't how we plan to solve the outstanding issue, although I do appreciate you taking a look.
> 
> Given we're in a multi threaded context, I was thinking of holding a hashmap of running requests with their req_id as key, and calling get result would wait til completion or timeout. But this may need some careful attention to avoid breaking things.
> 
> Feel free to refactor in a direction that looks more like how we intend to address the problem, otherwise we'll close the PR.

Instead of pushing everything directly into output_queue maybe we could route through a small dispatcher? would that fit the context more?

PR #42945, opened 2025-12-18
[Quantization] Removing misleading int8 quantization in Finegrained FP8

MekkCyber:
# What does this PR do?

instead of falling to int8 quantization, which we don't do for now since we validate that the environment has a `8.9` compute capability at least, we just ask the user to use an other quantization method


SunMarc (approved):
Yes let's do that

PR #42946, opened 2025-12-18
fix concat order

eustlb:
# What does this PR do?

Had one in two chances, picked the wrong order 😅
here's[ the original code for reference](https://github.com/facebookresearch/perception_models/blob/3e352cca660658d4b5c90f42a7808b11469e4c66/core/audio_visual_encoder/pe.py#L436-L447)


ArthurZucker (approved):
ty, fast integration tests would help!

PR #42953, opened 2025-12-19
[kernels] Fix failling tests

MekkCyber:
# What does this PR do?

After moving the use of `get_kernel` from `kernels` to using a wrapper around it from `hub_kernels.py`, we need to adapt the tests to override `get_kernel` inside `integrations/hub_kernels.py`


SunMarc (approved):
Thx

PR #42957, opened 2025-12-19
Do not use global variable, and improve context manager coverage

Cyrilvallez:
# What does this PR do?

As per the title

### DIFF
--- a/src/transformers/initialization.py
+++ b/src/transformers/initialization.py
@@ -206,3 +206,40 @@ def guard_torch_init_functions():
         for module, functions in originals.items():
             for func_name, func in functions.items():
                 setattr(module, func_name, func)
+
+
+@contextmanager
+def no_init_weights():
+    """
+    Disable weight initialization both at the torch-level, and at the transformers-level (`init_weights`).
+    This is used to speed-up initializing an empty model with deepspeed, as we do not initialize the model on meta device
+    with deepspeed, but we still don't need to run expensive weight initializations as we are loading params afterwards.
+    """
+    from .modeling_utils import PreTrainedModel
+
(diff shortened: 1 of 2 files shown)


pranjaldoshi96 (approved):

Cyrilvallez:
Hi @pranjaldoshi96, I noticed you keep randomly approving pull requests without any reason. We do not tolerate this kind of behavior simply to try to maximize your github activity or something, as it can give false signal to members (PRs appear as approved - even though your approval is meaningless). This is my first and only warning, if I see you doing it again I will be forced to ban you from the repo

BenjaminBossan:
@Cyrilvallez I'm wondering if it would make sense to re-import `no_init_weights` into `modeling_utils.py` for backwards compatibility (with potential deprecation).

Cyrilvallez:
Humm, we could import it explicitly in `modeling_utils` so that the object is still reachable as before - would that work for you @BenjaminBossan? But be careful about it, we want to move away from it, and the only reason it still exists is due to deepspeed when model instantiation does not happen on meta device yet (but hopefully in the future)

BenjaminBossan:
> would that work for you

Actually, I don't directly need it, I just saw that the latest gptqmodel version fails because of this (they have a fix ready) and was concerned that there might be other code out there that depends on this function. If it is intended to be removed completely, I'd say it would be nice to have a deprecation warning and to add the re-import.

Cyrilvallez:
I did not add any deprecation cycle, as this function is originally coming from accelerate (and was only copy/pasted in Transformers directly for convenience, and recently), so I assumed anyone that wanted to use it/import it would do it directly from accelerate

PR #42958, opened 2025-12-19
[`Generation`] Fix default overwrite for non-`None` defaults

vasqu:
Followup to #42702 where defaults are now either `None` or a boolean value. 

The logic was dependent on the values being `None` which ignored overriding other values, e.g. `do_sample`. The fails due to this can be seen in a few integration tests, e.g. gpt2.

### DIFF
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -1805,7 +1805,7 @@ class GenerationMixin(ContinuousMixin):
                 for k, v in self.generation_config.to_dict().items()
                 if isinstance(v, bool)
                 and hasattr(default_generation_config, k)
-                and getattr(generation_config, k) == getattr(default_generation_config, k)
+                and getattr(generation_config, k, None) == getattr(default_generation_config, k)
             }
         )
 
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -1788,6 +1788,10 @@ class GenerationMixin(ContinuousMixin):
         # with ``strict=True``. deepcopy can only be processed if ``strict=False``.
         generation_config = copy.deepcopy(generation_config)
 
+        # Due to some values being boolean and not `None`, we need additional logic to overwrite
+        # them explicitly (`defaults_only=False`)
+        generation_config.update(**{k: v for k, v in self.generation_config.to_dict().items() if isinstance(v, bool)})
+
         # First set values from the loaded `self.generation_config`, then set default values (BC)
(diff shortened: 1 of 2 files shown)


vasqu:
run-slow: gpt2,whisper,qwen2_5_vl,glm4v,biogpt

ArthurZucker (approved):
We need a small fast test 🫡

src/transformers/generation/utils.py
```diff
@@ -1788,6 +1788,10 @@ def _prepare_generation_config(
         # with ``strict=True``. deepcopy can only be processed if ``strict=False``.
         generation_config = copy.deepcopy(generation_config)
 
+        # Due to some values being boolean and not `None`, we need additional logic to overwrite
+        # them explicitly (`defaults_only=False`)
+        generation_config.update(**{k: v for k, v in self.generation_config.to_dict().items() if isinstance(v, bool)})
```
molbap: so we don't need to restrict that update? to something like

```python
default_gc = GenerationConfig()


bool_defaults = {k: v for k, v in self.generation_config.to_dict().items() if isinstance(v, bool)
    and hasattr(generation_config, k)
    and hasattr(default_gc, k)
    and getattr(generation_config, k) == getattr(default_gc, k)
}

generation_config.update(**bool_defaults)

```

just to be sure I understand :grin:

vasqu: Yup, it's correct I didnt notice that we can pass a generation config directly there as well (e.g. as in assisted decoding)

vasqu:
run-slow: whisper,gemma3n,gpt2,moshi,bert

vasqu:
run-slow: whisper,gemma3n,gpt2,moshi,bert

src/transformers/generation/utils.py
```diff
@@ -1795,6 +1795,20 @@ def _prepare_generation_config(
...
+        default_generation_config = GenerationConfig()
+        generation_config.update(
+            **{
+                k: v
+                for k, v in self.generation_config.to_dict().items()
+                if isinstance(v, bool)
+                and hasattr(default_generation_config, k)
+                and getattr(generation_config, k, None) == getattr(default_generation_config, k)
+            }
+        )
```
zucchini-nlp, line 1810: sorry for late review. I think we're back to the original issue where users pass `do_sample=False` in their custom config and the model has `do_sample=True` saved on the hub. And since `False` is the default value, we won't update it here and keep `True`

In that case we'd need to set the booleans to `None` by default as well, I dont' see a cleaner way than defining `None` as non-set

PR #42467, opened 2025-11-28
Fixes StaticCache Crashes 

i3hz:
# What does this PR do?

Fixes #42454 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.
@zucchini-nlp  @Rocketknight1 @mobicham 

Benchmarking script  - 
```python
import torch
import time
import numpy as np
from transformers import WhisperForConditionalGeneration
from transformers.cache_utils import StaticCache, EncoderDecoderCache


MODEL_ID = "openai/whisper-tiny"
DTYPE = torch.float16 if torch.cuda.is_available() else torch.float32
DEVICE = "cuda" if torch.cuda.is_available() else "cpu"

MAX_BATCH_SIZE = 64
TEST_BATCHES = [1, 8, 32, 64]

SEQ_LEN = 128
WARMUP = 10
REPEATS = 50


def load_model():
    model = WhisperForConditionalGeneration.from_pretrained(
        MODEL_ID,
        torch_dtype=DTYPE,
        attn_implementation="sdpa",
    ).to(DEVICE)
    model.eval()
    return model


def run_benchmark(model, batch_size, cache_cap, tag):
    decoder = model.model.decoder
    cache_len = SEQ_LEN + 10
    enc_len = SEQ_LEN

    self_cache = StaticCache(
        config=decoder.config,
        max_batch_size=cache_cap,
        max_cache_len=cache_len,
        device=DEVICE,
        dtype=DTYPE,
    )
    cross_cache = StaticCache(
        config=decoder.config,
        max_batch_size=cache_cap,
        max_cache_len=enc_len,
        device=DEVICE,
        dtype=DTYPE,
    )
[...]

PR #42960, opened 2025-12-19
Fix StaticCache crash when batch size changes between runs

lukepayyapilli:
## Summary
Fix `StaticCache` crash when batch size changes between inference runs, even when the new batch size is smaller than `max_batch_size`.

## Problem
When using `StaticCache` with different batch sizes across multiple inference runs (after calling `reset()`), it would crash with:
```
RuntimeError: index_copy_(): Source/destination tensor must have same slice shapes.
Destination slice shape: 4 20 64 at dimension 2 and source slice shape: 8 20 64 at dimension 0.
```

This is problematic for users who want to reuse a pre-allocated cache with CUDA graphs / `torch.compile`, but need to handle variable batch sizes.

## Solution
- Use tensor slicing instead of `index_copy_` to update only the relevant batch positions
- Return only the batch slice that was actually used

## Changes
- `StaticLayer.update()`: Use slicing to support variable batch sizes
- `StaticSlidingWindowLayer.update()`: Same fix for sliding window variant
- Added regression test `test_static_cache_variable_batch_size`

Fixes #42454

### DIFF
--- a/src/transformers/cache_utils.py
+++ b/src/transformers/cache_utils.py
@@ -332,15 +332,15 @@ class StaticLayer(CacheLayerMixin):
             cache_position if cache_position is not None else torch.arange(key_states.shape[-2], device=self.device)
         )
 
-        # Update the cache
-        try:
-            self.keys.index_copy_(2, cache_position, key_states)
-            self.values.index_copy_(2, cache_position, value_states)
-        except NotImplementedError:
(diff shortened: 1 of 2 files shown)


lukepayyapilli:
Closing as duplicate of #42467. I didn't check for existing PRs before submitting - apologies for the noise\! The fix in #42467 is more comprehensive and handles edge cases with torch.compile and encoder-decoder models.

PR #42963, opened 2025-12-19
[`Tests`] Fix inputs placement

vasqu:
As per title, we init our model on CPU but the inputs are not guaranteed to be on CPU as well. Forcing CPU here

Related failure: https://github.com/huggingface/transformers/actions/runs/20375489555/job/58553450832


zucchini-nlp (approved):
Facing the same issue when running tests locally, thanks for the fix

PR #37590, opened 2025-04-17
:rotating_light: :rotating_light: Inherited CausalLM Tests

Rocketknight1:
This is an experimental PR to see if we can inherit tests for causal language models, which should reduce the amount of boilerplate model authors need to write by a lot.

Right now it's **very** experimental and will only touch a couple of models! Expect lots of failures for a while.

Update: It was stable and working for ~20 models, so I'm going to merge it at this point and migrate other models to it over time.

PR #42964, opened 2025-12-19
Add shared VLM tests

Rocketknight1:
EDIT: Ready for review!

This is a base test and tester class for VLMs, similar to the existing [Causal LM tests](https://github.com/huggingface/transformers/pull/37590).

The main difficulty here is that VLMs are more variable than pure LLMs. In particular, VLMs have variations in how many image tokens they add, where they place them in the sequence, and how they handle things like attention masks for image tokens. 

This is handled by the `Processor` class, but we generally don't load processor classes in modeling tests. Because we don't have processors, VLM model tester classes usually need to override a couple of base methods, so that they can generate test inputs that work for their VLM. I've tried to keep these methods as small and simple as possible.

I've reparented four existing popular VLM classes onto the new framework, and it seems to work well for all of them (Gemma3, Llava-Next, Qwen3VL and Qwen3VL-MoE). We can reparent more in future, or just leave older tests alone and only use this for new model additions.

Also, I asked Claude to try some parts of this and it **really** liked `kwargs.setdefault()` instead of the long `__init__()` methods that we usually have for tester classes. I think it's cleaner too, but if we want to go back to the old method I can revert that.


tests/vlm_tester.py
```diff
@@ -0,0 +1,372 @@
...
+        [
+            (-1,),
+            ([-1],),
+            ([-1, -2],),
+        ],
+    )
+
+    @unittest.skip(reason="VLMs cannot pass input_embeds without input_ids")
+    def test_inputs_embeds(self):
+        pass
```
molbap: Isn't that possible now?

molbap (commented):
Thanks for starting this work! Subscribed to the PR :eyes: (left a random comment)

tests/vlm_tester.py
Rocketknight1: This was just copied from an existing class test! I'll try enabling it

Rocketknight1: Enabled it and it works, thanks for pointing it out! Resolving this.

Rocketknight1:
This should be ready for review now! cc @zucchini-nlp for VLMs plus core maintainers @ArthurZucker @Cyrilvallez @vasqu

zucchini-nlp:
Didn't yet have a look, just want to share one thought before I forget it. I am seeing a lot of models in `utils/check_repo.py` in the following format. Would be great to do smth about this as well and clean-up the list of skipped tests, maybe as a subsequent PR

```
"Glm4vTextModel",  # Building part of bigger (tested) model
        "Glm4vMoeTextModel",  # Building part of bigger (tested) model
        "GlmImageTextModel",  # Building part of bigger (tested) model
        "GlmOcrTextModel",  # Building part of bigger (tested) model
```

tests/vlm_tester.py
```diff
@@ -0,0 +1,418 @@
+# Copyright 2025 HuggingFace Inc.
```
zucchini-nlp: super nit: 2026 😄

tests/vlm_tester.py
```diff
@@ -0,0 +1,418 @@
...
+        mamba_d_conv=4,
+        mamba_expand=2,
+        mamba_chunk_size=16,
+        image_size=8,
+        patch_size=4,
+        num_channels=3,
+        projection_dim=32,
+        projector_hidden_act="gelu",
+        ignore_index=-100,
+        image_token_index=0,
```
zucchini-nlp: image token id can't be same as `PAD` . In general we gotta be careful with these special tokens, some models have a ton of special multimodal tokens

Also, let's call it `image_token_id` to be consistent with config naming format

tests/vlm_tester.py
```diff
@@ -0,0 +1,418 @@
...
+    def seq_length(self):
+        return self._base_seq_length + self.num_image_tokens
+
+    def create_pixel_values(self):
+        # Override to 5D for patch-based models
+        return floats_tensor([self.batch_size, self.num_channels, self.image_size, self.image_size], scale=1.0)
+
+    def create_attention_mask(self, input_ids):
+        # Override for bidirectional attention models like Gemma3
+        return torch.tril(torch.ones_like(input_ids).to(torch_device))
```
zucchini-nlp, line 157: should we do left padding because VLMs are always generative? Also I'm not sure if random mask would work for all models, if it doesn't agree with tokens in `input_ids`. Models can complain if a placeholder image token id is masked (e.g. qwen-vl)

tests/vlm_tester.py
```diff
@@ -0,0 +1,418 @@
...
+
+    def place_image_tokens(self, input_ids, config):
+        # Override if the image tokens shouldn't be placed at the start of the test sequence
+        image_token_index = getattr(config, "image_token_index", self.image_token_index)
```
zucchini-nlp: ig this diff was done in 2025 😄  Config attr is now called `image_token_id` for all models

tests/vlm_tester.py
```diff
@@ -0,0 +1,418 @@
...
+        # Clear any accidental image tokens first
+        input_ids = input_ids.clone()
+        input_ids[input_ids == image_token_index] = self.bos_token_id
+        # Place image tokens at the start
+        input_ids[:, : self.num_image_tokens] = image_token_index
+        return input_ids
```
zucchini-nlp, line 167: maybe we could decompose it into following and let testers to override those:

1. `self.one_image_as_string` which is a placeholder for one image
2. `self.mm_special_tokens` which are image/video and any other special tokens. We'll use it to set `input_ids.isin(mm_special_tokens)` to `BOS/PAD`

tests/vlm_tester.py
```diff
@@ -0,0 +1,418 @@
...
+
+    def get_additional_inputs(self, config, input_ids, pixel_values):
+        # Override for model-specific inputs like LlavaNext's image_sizes
+        return {}
+
+    # End of overridable methods/properties
+
+    def prepare_config_and_inputs(self):
```
zucchini-nlp: tbh I don't know if we use `prepare_config_and_inputs` anywhere. The current outputs will not be usable in VLMs because it has no placeholder tokens. So should we move `self.place_image_tokens` in here?

tests/vlm_tester.py
```diff
@@ -0,0 +1,418 @@
...
+
+    @property
+    def vision_config_args(self):
+        return list(signature(self.vision_config_class.__init__).parameters.keys())
+
+    def get_config(self):
+        kwargs = {}
+        attribute_map = getattr(self.config_class, "attribute_map", {})
+        model_name_to_common_name = {v: k for k, v in attribute_map.items()}
+        for k in self.config_args + self.forced_config_args:
```
zucchini-nlp, line 230: I wonder why we need `config.pad_token_id` if we have `config.text_config.pad_token_id`

tests/vlm_tester.py
```diff
@@ -0,0 +1,418 @@
...
+        for model_class in self.all_model_classes:
+            model = model_class(config).to(torch_device)
+            model.eval()
+            curr_input_dict = copy.deepcopy(input_dict)
+            _ = model(**curr_input_dict)  # successful forward with no modifications
+
+            # Test 1: remove one image but leave the image token in text
+            curr_input_dict["pixel_values"] = curr_input_dict["pixel_values"][-1:, ...]
+            if "image_sizes" in curr_input_dict:
+                curr_input_dict["image_sizes"] = curr_input_dict["image_sizes"][-1:, ...]
```
zucchini-nlp, line 330: maybe we could loop over `self.additional_inputs.keys()` for all inputs

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
```diff
@@ -16,186 +16,127 @@
...
-        patch_size = config.vision_config.patch_size
-        temporal_patch_size = config.vision_config.temporal_patch_size
-        pixel_values = floats_tensor(
+class Qwen3VLVisionText2TextModelTester(VLMModelTester):
+    if is_torch_available():
+        base_model_class = Qwen3VLModel
+        config_class = Qwen3VLConfig
+        text_config_class = Qwen3VLTextConfig
+        vision_config_class = Qwen3VLVisionConfig
+        conditional_generation_class = Qwen3VLForConditionalGeneration
```
zucchini-nlp: curious, why we need to check `is_torch_available` in qwen but not in llava?

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
```diff
@@ -16,186 +16,127 @@
...
+
+    # Qwen3 VL-specific configuration
+    image_token_id = 3
+    video_token_id = 4
+    vision_start_token_id = 5
+    vision_end_token_id = 6
```
zucchini-nlp: btw, what is the diff between setting them as cls attr or in `__init__`?

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
```diff
@@ -16,186 +16,127 @@
...
+        self.vision_hidden_act = "gelu_pytorch_tanh"
+        self.out_hidden_size = self.hidden_size
+        self.vision_hidden_size = self.hidden_size
+        self.vision_intermediate_size = self.hidden_size
+        self.num_heads = 4
+        self.spatial_merge_size = 1
+        self.temporal_patch_size = 2
+        self.num_position_embeddings = 16
+        self.deepstack_visual_indexes = [0, 1]
+        self.rope_theta = 10000
```
zucchini-nlp: shouldn't be needed, no? Rope theta can be moved inside `rope_parameters` as a new field

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
```diff
@@ -16,186 +16,127 @@
...
         input_ids[input_ids == self.image_token_id] = self.pad_token_id
         input_ids[input_ids == self.vision_start_token_id] = self.pad_token_id
-        input_ids[:, self.num_image_tokens] = self.image_token_id
-        input_ids[:, self.num_image_tokens - 1] = self.vision_start_token_id
-        inputs_dict = {
-            "pixel_values": pixel_values,
+        # Place image tokens with vision_start_token_id prefix
+        input_ids[:, self._base_num_image_tokens] = self.image_token_id
+        input_ids[:, self._base_num_image_tokens - 1] = self.vision_start_token_id
+        return input_ids
```
zucchini-nlp, line 108: looks weird, we're placing one image token at `index=_base_num_image_tokens`. Most prob the tester needs one image token anyway, can we make it clear with `input_ids[:, :self._base_num_image_tokens]` though?

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
```diff
@@ -16,186 +16,127 @@
...
+    def get_additional_inputs(self, config, input_ids, pixel_values):
+        return {
             "image_grid_thw": torch.tensor([[1, 1, 1]] * self.batch_size, device=torch_device),
-            "input_ids": input_ids,
-            "attention_mask": attention_mask,
         }
-        return config, inputs_dict
+
+    def get_config(self):
+        # Qwen3VLConfig expects text_config and vision_config as dicts, not config objects
```
zucchini-nlp, line 116: this is interesting, I have updated all configs to accept dicts and `PreTrainedConfig` as args in the past. New models seem to be missing it 😢  IMO we need to fix the config class instead

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
zucchini-nlp, line 235: this is another point I've been wanting to work on, and discussed with Joao in the past

So we have models which can accept text, text+image, text+video, text+image+video as Qwen-VL. We also have models with `audio` added to this combination as in Gemma3n. In our tests though we always check a certain preset combination input args, mostly it is "text+image"

What I wanted to do is to allow us prepare each modality args with separate fn, and at least run forward passes with different combinations. Models can behave differently depending on what args it got. Leaving it here as a thought, I might work on it in the future if no-one implements it first :)

tests/models/qwen3_vl_moe/test_modeling_qwen3_vl_moe.py
```diff
@@ -25,6 +25,7 @@
     Qwen3VLMoeModel,
     is_torch_available,
 )
+from transformers.models.qwen3_vl_moe.configuration_qwen3_vl_moe import Qwen3VLMoeTextConfig, Qwen3VLMoeVisionConfig
```
zucchini-nlp, line 28: same comments as in qwen3-vl

zucchini-nlp (commented):
Super happy to have a VLM tester, thanks for working on it! I left a couple comments about what we can do next to improve on this

Overall LGTM, just a few questions here and there

tests/vlm_tester.py
Rocketknight1: Fixed!

tests/vlm_tester.py
Rocketknight1: Fixed!

tests/vlm_tester.py
Rocketknight1: Fixed! And yes it was, I started it before Christmas and then resumed it this week lol

tests/vlm_tester.py
Rocketknight1: Yes, good point! This was copied originally from Gemma3 (I think I used Gemma3?) and then modified to make it more general, where we had both `prepare_config_and_inputs` and `prepare_config_and_inputs_from_common`. But actually, all of the tests (including in the mixins) just call `prepare_config_and_inputs_for_common`, so it makes a lot more sense to just have a single overridable method.

Rocketknight1: Replaced everything with `prepare_config_and_inputs_for_common`

tests/vlm_tester.py
Rocketknight1: This is mostly paranoia (because I wasn't sure which config downstream models would read it from, and I wanted it to work for all models)

tests/vlm_tester.py
Rocketknight1: Hmm, is it always true that the `additional_inputs` have `num_images` as their first dimension? I think this would break for e.g. `token_type_ids`

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
Rocketknight1: Good catch! This was just caused by me copying from multiple original tests. I removed the `if` block because `@require_torch` handles it

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
Rocketknight1: No difference, some of this is Claude's fault :sweat_smile:

I think the right structure is:

- Lots of default args in the `__init__` of the base class
- The only named kwargs in the model class are attributes that don't exist in the base `__init__`
- Model-specific **overrides** of base class defaults are all in `**kwargs` and handled with `kwargs.setdefault()`
- Call `super().__init__()` after the `kwargs.setdefault()` block but before setting model-specific args, so that if the base class adds a new kwarg it doesn't break any existing model classes that already had it

WDYT? I can standardize them all to that

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
Rocketknight1: Done!

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
Rocketknight1: Yeah, this was another case of stuff getting copied around and edited and then not making sense! The tester does need one image token (because we only pass one patch in the tests), but I can just make the indices `-2` and -1` instead of `num_image_tokens`, which is unnecessary

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
Rocketknight1: I'll do a config fix on Monday and see if we can drop some of these overrides!

tests/models/qwen3_vl_moe/test_modeling_qwen3_vl_moe.py
Rocketknight1: Fixed!

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
Rocketknight1: Definitely a cool idea, but trying to do it at the same time as the inheritance refactor feels dangerous, yeah. I'll refactor to VLM shared tests first, and then we can think about a multi-modality shared tester?

zucchini-nlp: sure thing, just wanted to share it here as it's related! I've been thinking about it for a while and never had a chance to check if it's feasible

tests/vlm_tester.py
zucchini-nlp: ohh right, the shapes of inputs are going to be different from model to model 😢  So we have additional inputs for those with "num_images as their first dimension" and a separate attribute for inputs with "batch_size as their first dimension"?

Rocketknight1:
@zucchini-nlp I dropped this to focus on the CI, but I got back to it this week! I think it's in good shape, and I added `deepseek-vl` as well, so it's got pretty broad coverage of several model families now. Do you want to take one final review and then I'll ping core maintainers?

src/transformers/models/qwen3_5/modeling_qwen3_5.py
```diff
@@ -2018,8 +2018,8 @@ def _prepare_position_ids_for_generation(self, inputs_tensor, model_kwargs):
         if (cache := model_kwargs.get("past_key_values")) is not None:
             past_length = cache.get_seq_length()
         if past_length != 0 and self.model.rope_deltas is not None:
-            text_positions += self.model.rope_deltas
-            return text_positions
+            # No need to add rope_deltas because the positions already account for padding
+            return text_positions[None, ...].expand(4, -1, -1)
```
zucchini-nlp: The super call will account for padding, but it doesn't know account for vision tokens. Vision tokens usually don't grow incrementally by one, so we still need the deltas here. Maybe there's another reason the test was failing or we can try to rebase, in case it was fixed on `main`

tests/models/gemma3/test_modeling_gemma3.py
```diff
@@ -246,115 +247,47 @@ def test_model_rope_scaling_frequencies(self):
...
-            [
-                self.batch_size,
-                self.vision_config["num_channels"],
-                self.vision_config["image_size"],
-                self.vision_config["image_size"],
-            ]
-        )
-        config = self.get_config()
+class Gemma3Vision2TextModelTester(VLMModelTester):
+    if is_torch_available():
```
zucchini-nlp: do we need `is_torch_available` in the tester? Currently it's used in gemma3 and not used in deepseek

tests/models/gemma3/test_modeling_gemma3.py
```diff
@@ -246,115 +247,47 @@ def test_model_rope_scaling_frequencies(self):
...
-        input_ids[input_ids == config.image_token_index] = self.pad_token_id
-        input_ids[:, :1] = config.image_token_index
+    def create_attention_mask(self, input_ids):
+        # Gemma3 uses padding mask for bidirectional attention on image tokens
+        return input_ids.ne(self.pad_token_id).to(torch_device)
 
+    def get_additional_inputs(self, config, input_ids, pixel_values):
+        # Gemma3 requires specific token_type_ids for bidirectional attention on image tokens
         token_type_ids = torch.zeros_like(input_ids)
         token_type_ids[input_ids == config.image_token_index] = 1
```
zucchini-nlp: let's use `image_token_id` please, so the old format `index` gets deprecated at some point

tests/models/gemma3/test_modeling_gemma3.py
```diff
@@ -246,115 +247,47 @@ def test_model_rope_scaling_frequencies(self):
...
-            Gemma3ForConditionalGeneration,
-            Gemma3ForSequenceClassification,
-        )
-        if is_torch_available()
-        else ()
-    )
-    all_generative_model_classes = (Gemma3ForConditionalGeneration,) if is_torch_available() else ()
+class Gemma3Vision2TextModelTest(VLMModelTest, unittest.TestCase):
+    model_tester_class = Gemma3Vision2TextModelTester
+    config_tester = ConfigTester
```
zucchini-nlp: do we need to explicitly state `ConfigTester` to run common config tests? Maybe run it by default always?

tests/models/llava_next/test_modeling_llava_next.py
```diff
@@ -58,245 +53,68 @@
...
-        self.num_image_tokens = 24
-        self.seq_length = seq_length + self.num_image_tokens
-        self.encoder_seq_length = self.seq_length
+class LlavaNextVisionText2TextModelTester(VLMModelTester):
+    base_model_class = LlavaNextModel
+    config_class = LlavaNextConfig
+    conditional_generation_class = LlavaNextForConditionalGeneration
+    text_config_class = LlamaConfig
+    vision_config_class = CLIPVisionConfig
+    all_model_classes = (LlavaNextModel, LlavaNextForConditionalGeneration)
```
zucchini-nlp: for `all_model_classes`, ig we can infer it from existing `base/conditional/sequence_clf` classes. And let users to override only if the model is non-standard

Also applied to gemma3

tests/models/llava_next/test_modeling_llava_next.py
```diff
@@ -58,245 +53,68 @@
...
+
+    @property
+    def num_image_tokens(self):
+        """
+        Calculate num_image_tokens based on LlavaNext's pack_image_features logic:
+        - base_image_feature: (image_size/patch_size)^2 tokens
+        - grid patches: height * (width + 1) tokens (includes newline token)
+        """
+        # Use vocab_size - 1 for image_token_index to avoid conflicts with pad_token_id (0)
+        self.image_token_index = self.vocab_size - 1
```
zucchini-nlp: this should be set in `__init__`, no? Also I think using `vocab_size-1` will cause issues when testing `resize_embeddings`. That is why I usually set special mm tokens to the next index after pad/bos/eos

tests/models/llava_next/test_modeling_llava_next.py
```diff
@@ -58,245 +53,68 @@
...
-        {
-            "image-text-to-text": LlavaNextForConditionalGeneration,
-            "any-to-any": LlavaNextForConditionalGeneration,
-        }
-        if is_torch_available()
-        else {}
-    )
-    # Llava-NeXT merges batch_size and num_patches in the first output dimension
+    model_tester_class = LlavaNextVisionText2TextModelTester
+    test_resize_embeddings = False
```
zucchini-nlp: oh i see, the test is skipped. I don't think skipping is a good idea, we need to make sure that VLMs know how to get/set text embedding layer and tie weights if needed

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
zucchini-nlp: Yeah, that sounds good to me. And if the `kwarg` is super specific to a certain model, we can also hardcode it in `__init__`. I am quite worried about cls attributes, because we might manually change values in tests and never re-set it back

Saving everything in instance attr seems to be better to me

tests/vlm_tester.py
```diff
@@ -0,0 +1,406 @@
...
+    text_config_class = None
+    vision_config_class = None
+    conditional_generation_class = None
+    sequence_classification_class = None
+    # These attributes are required after the initialization phase of the tester.
+    _required_attributes = ("base_model_class", "config_class", "conditional_generation_class")
+
+    # Arguments that should be passed to the config class even if not in its signature
+    forced_config_args = ["pad_token_id"]
+
```
zucchini-nlp, line 50: i am actually thinking now that not forcing `pad_token_id` will help us to encourage users to set correct config defaults. If the config needs an attr, it should be in the `config.__init__` with a default value

Would be great to not force and instead fix model configs!

tests/vlm_tester.py
```diff
@@ -0,0 +1,406 @@
...
+        return config, inputs_dict
+
+    @property
+    def config_args(self):
+        return list(signature(self.config_class.__init__).parameters.keys())
+
+    @property
+    def text_config_args(self):
+        args = list(signature(self.text_config_class.__init__).parameters.keys())
+        for token_arg in ["pad_token_id", "bos_token_id", "eos_token_id"]:  # Not always explicitly in the sig
```
zucchini-nlp, line 217: same here, if the token id is not in signature but the model forward fails without the attr, we need to fix config class itself

zucchini-nlp (commented):
Great, lgtm! I think there are some left-over claude code that needs to be cleaned, for example using `torch_available` in tester or setting tester attr as class vs instance attr

tests/models/gemma3/test_modeling_gemma3.py
Rocketknight1: Yes, good point! We can probably skip it here - a lot of this was created by copying from different files, which is why it gets a bit confused.

tests/models/gemma3/test_modeling_gemma3.py
Rocketknight1: Fixed!

tests/models/gemma3/test_modeling_gemma3.py
Rocketknight1: Yeah this was either Claude or dead code copied from somewhere. It had no effect, thankfully, because the attribute was overwritten by `__init__` anyway

tests/models/llava_next/test_modeling_llava_next.py
Rocketknight1: Yeah, this was actually overriding a property! Not sure how it ever got in there, probably due to copied code again

tests/models/llava_next/test_modeling_llava_next.py
Rocketknight1: I'll remove that skip, hopefully things aren't failing now!

src/transformers/testing_utils.py
```diff
@@ -216,6 +216,14 @@
     "token_classification_class": "ForTokenClassification",
 }
 
+# Used in VLMModelTester (and related classes/methods) to infer the common model classes from the base model class
+_VLM_COMMON_MODEL_NAMES_MAP = {
```
tarekziade, line 220: I don't see this used in `check_repo.py` - (which uses `_COMMON_MODEL_NAMES_MAP`). don't we need to merge the maps to avoid a KeyError to make it future proof?

tests/models/llava_next/test_modeling_llava_next.py
```diff
@@ -58,245 +53,66 @@
...
-            # one image and two image tokens raise an error
-            with self.assertRaisesRegex(ValueError, "Image features and image tokens do not match"):
-                _ = model(input_ids=input_ids, pixel_values=pixel_values, image_sizes=image_sizes)
-
-            # two images and two image tokens don't raise an error
-            pixel_values = torch.cat([pixel_values, pixel_values], dim=0)
-            image_sizes = torch.cat([image_sizes, image_sizes], dim=0)
-            _ = model(input_ids=input_ids, pixel_values=pixel_values, image_sizes=image_sizes)
-
-    def test_odd_sized_image(self):
```
tarekziade, line 252: will this one be replaced?

tests/models/llava_next/test_modeling_llava_next.py
```diff
@@ -58,245 +53,66 @@
...
-            model(**inputs_dict)
-
-    @parameterized.expand(
-        [
-            (-1,),
-            ([-1],),
-            ([-1, -2],),
-        ],
-    )
-    def test_vision_feature_layers(self, vision_feature_layer):
```
tarekziade, line 281: same question

tarekziade:
Great work on this! just noticed a small issue on `check_repo` and also wondered about the test coverage loss on llavanext as the new tests don't seem to keep some of those edge cases

tests/vlm_tester.py
Rocketknight1: I changed the `__init__` args to all be set with `kwargs.setdefault()`. The `forced_config_args` line is copied from the `CausalLMTester`, so we probably shouldn't try to totally fix it here. Instead I think we should merge it with a TODO (and the same TODO in `CausalLMTester`), and make a follow-up PR to fix all the `config` classes and remove that line in both.

tests/vlm_tester.py
Rocketknight1: Same as above - I think we should merge with a TODO because the scope extends beyond this PR, then do a follow-up cleanup PR afterwards!

src/transformers/testing_utils.py
Rocketknight1: Yes, good point! I'll update `check_repo`

tests/models/llava_next/test_modeling_llava_next.py
Rocketknight1: For now, probably not! A lot of specific models have unusual tests that are not common in other VLMs (this test is only seen in llava and llava-derived models I think). I don't want to add all of them to the base class, and I think it's okay to lose coverage on things that are unlikely to regress!

Rocketknight1:
After another [detour to patch some blocking flaky tests](https://github.com/huggingface/transformers/pull/44340), this should be ready for re-review @zucchini-nlp!

tests/models/gemma3/test_modeling_gemma3.py
```diff
@@ -246,115 +246,47 @@ def test_model_rope_scaling_frequencies(self):
...
+        input_ids = super().place_image_tokens(input_ids, config)
+        # Gemma3 uses a padding-based attention mask, so we must ensure no random
+        # input_ids are pad tokens otherwise those positions get masked out and
+        # cause flaky failures
+        input_ids[input_ids == self.pad_token_id] = self.bos_token_id
+        return input_ids
+
+    def create_attention_mask(self, input_ids):
+        # Gemma3 uses padding mask for bidirectional attention on image tokens
+        return input_ids.ne(self.pad_token_id).to(torch_device)
```
zucchini-nlp: does this mean we are always testing with full-mask, i.e. no padding? If yes, I'm worried that it might result in missing out on some cases to test

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
```diff
@@ -45,167 +43,104 @@
...
+        kwargs.setdefault("num_attention_heads", 4)
+        kwargs.setdefault("num_key_value_heads", 2)
+        kwargs.setdefault("head_dim", 8)
+        super().__init__(parent, **kwargs)
+
+        # Qwen3 VL-specific vision config attributes (computed from instance state)
+        self.depth = 2
+        self.vision_hidden_act = "gelu_pytorch_tanh"
+        self.out_hidden_size = self.hidden_size
+        self.vision_hidden_size = self.hidden_size
```
zucchini-nlp: just curious, what is the difference between `setdefault` and force-setting attributes after super?

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
zucchini-nlp: I just checked the config, and it is indeed not working with `Config` objects 😢  I will fix and add a test for that in separate PR, no idea how many more models are affected

tests/vlm_tester.py
```diff
@@ -0,0 +1,353 @@
...
+    def create_attention_mask(self, input_ids):
+        # Override for bidirectional attention models like Gemma3
+        return torch.tril(torch.ones_like(input_ids).to(torch_device))
+
+    def place_image_tokens(self, input_ids, config):
+        # Override if the image tokens shouldn't be placed at the start of the test sequence
+        image_token_id = getattr(config, "image_token_id", self.image_token_id)
+        # Clear any accidental image tokens first
+        input_ids = input_ids.clone()
+        input_ids[input_ids == image_token_id] = self.bos_token_id
```
zucchini-nlp, line 164: we might improve on this in the next PR by keeping a list of `special_multimodal_token_ids` and setting all of them to PAD/BOS. It will help to generalize better imo

zucchini-nlp (approved):
Great, thanks for cleaning it up! I think we can ask for core maintainer's review and merge

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
Rocketknight1: Ugh, I thought I'd fixed all of those. I'll use `setdefault` for those too!

They're mostly equivalent but setting attributes makes it impossible to override them by passing a kwarg. I think we should just use always use `setdefault`!

tests/vlm_tester.py
Rocketknight1: Yeah, that might be a good idea! It'd be cool to refactor the base class so we don't need as many overrides, but I agree that we probably want to merge at this point to avoid merge conflicts and then work from there.

Rocketknight1:
cc core maintainers @arthurzucker @cyrilvallez @vasqu this should be ready for final review by now!

Cyrilvallez (approved):
Hey @Rocketknight1!
Sorry for the late review on this! I maybe don't have as much of a good mental map as @zucchini-nlp for vlms tests, so trusting you all as I see you iterated on this PR previously.
What I think is very important though, if you could confirm:
- for VLMs we need to be very careful as what we allow as random tokens when we create random inputs, as img token, video tokens etc are not so unlikely to randomly appear if we use small vocabs (e.g. usually 100), and it can lead to flakyness in the tests as the model will (correctly) raise errors or crash
- Do we have (roughly) the same test coverage? I.e. sometimes tests can be activated/skipped based on obscure class-level flags in the tests, is this refactor using a general base class influencing on this?

If it's all good on thos aspects, feel free to merge!

Rocketknight1:
run-slow: deepseek_vl, gemma3, llava_next, qwen3_vl, qwen3_vl_moe

Rocketknight1:
@cyrilvallez no problem! I had some class-specific guards to prevent special tokens being generated, but you're right, it makes sense to do this in the base methods, so that we eliminate any potential future source of flaky tests too. I added that change!

Also, test coverage should be roughly equivalent. The tester is based on the Llava and Gemma3 model testers, so the test coverage for other models will not be exactly the same afterwards, but it should still cover the important bits because those models, especially Gemma3, were well-tested.

Rocketknight1:
run-slow: deepseek_vl, gemma3, llava_next, qwen3_vl, qwen3_vl_moe

Rocketknight1:
Slow tests were good except `test_torch_export` for `llava_next`, this was skipped on `main` and accidentally unskipped in this PR! I've fixed that, so we should be ready to merge.

tests/vlm_tester.py
```diff
@@ -0,0 +1,364 @@
...
+        kwargs.setdefault("intermediate_size", 37)
+        kwargs.setdefault("hidden_act", "gelu")
+        kwargs.setdefault("hidden_dropout_prob", 0.1)
+        kwargs.setdefault("attention_probs_dropout_prob", 0.1)
+        kwargs.setdefault("max_position_embeddings", 512)
+        kwargs.setdefault("type_vocab_size", 16)
+        kwargs.setdefault("type_sequence_label_size", 2)
+        kwargs.setdefault("initializer_range", 0.02)
+        kwargs.setdefault("num_labels", 3)
+        kwargs.setdefault("num_choices", 4)
```
ArthurZucker, line 96:
```suggestion
DEFAULTS = {
    "batch_size": 3,
    "is_training": True,
    "use_input_mask": True,
    "use_token_type_ids": False,
    "use_labels": True,
    "vocab_size": 99,
    "hidden_size": 32,
    "num_hidden_layers": 2,
    "num_attention_heads": 2,
    "num_key_value_heads": 2,
    "intermediate_size": 37,
    "hidden_act": "gelu",
    "hidden_dropout_prob": 0.1,
    "attention_probs_dropout_prob": 0.1,
    "max_position_embeddings": 512,
    "type_vocab_size": 16,
    "type_sequence_label_size": 2,
    "initializer_range": 0.02,
    "num_labels": 3,
    "num_choices": 4,
}

for k, v in DEFAULTS.items():
    kwargs.setdefault(k, v)
```

tests/vlm_tester.py
ArthurZucker: would it makes sense to have a "test_inputs" that's explicit with tokens and etc and that is used for both integration tests and the vlm tests? or its not generatl enough?

ArthurZucker (commented):
Small nits but good otherwise! 
I'm wondering if we can have explicit input string <image> examples and force users to just overwrite this ? but if it all works by guessing its fine!

tests/vlm_tester.py
Rocketknight1: I actually kind of prefer the repeated `kwargs.setdefault()`! It's more verbose, but it's actually fewer LOC than the other solution, and it makes it more immediately clear to new users what's happening

It's a very small nit though

Rocketknight1:
@ArthurZucker yeah, that could be an interesting follow-up refactor! I'd prefer not to put it in this PR, because it's already large and touching a lot of files, and altering all of the integration tests will make it a real nightmare to test/merge it! Definitely makes sense to check it out in a subsequent PR, though

PR #42965, opened 2025-12-19
Feature/add max eval steps

### LINKED ISSUE #31561: Add argument to set number of eval steps in Trainer

priyankaz-ml:
Summary:
This PR introduces a new optional argument max_eval_steps to TrainingArguments and integrates it into the Trainer evaluation loop. When set, evaluation will stop after the specified number of batches, even if the dataset has more batches.

Changes:

Added max_eval_steps to TrainingArguments with documentation.

Updated Trainer.evaluation_loop() to break early when max_eval_steps is reached.

Updated evaluate() docstring to mention the new behavior.

Added inline comments for traceability (#31561).

Motivation:
Large evaluation datasets can be time-consuming. This feature allows users to limit evaluation runtime for quicker experiments and debugging.

Issue Reference: Fixes #31561

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -4471,6 +4471,10 @@ class Trainer:
 
         # Main evaluation loop
         for step, inputs in enumerate(dataloader):
+            # Stop evaluation early if max_eval_steps is set — added as part of issue #31561 fix by @priyankaz-ml
+            if self.args.max_eval_steps is not None and step >= self.args.max_eval_steps:
+             \ break
+
             # Update the observed num examples
             observed_batch_size = find_batch_size(inputs)
             if observed_batch_size is not None:
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -4231,44 +4231,24 @@ class Trainer:
         ignore_keys: list[str] | None = None,
         metric_key_prefix: str = "eval",
     ) -> dict[str, float]:
-        """
-        Run evaluation and returns metrics.
+       """
+    Run evaluation and returns metrics.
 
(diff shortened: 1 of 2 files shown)


priyankaz-ml:
Hi! This PR adds max_eval_steps to limit evaluation batches (fixes #31561). Let me know if anything needs adjusting — happy to revise!

PR #42970, opened 2025-12-20
Fix issue loading ASR pipelines

pcuenca:
Introduced in #42806

Before this PR, the following failed:

```py
pipe = pipeline("automatic-speech-recognition")
```

### DIFF
--- a/src/transformers/pipelines/__init__.py
+++ b/src/transformers/pipelines/__init__.py
@@ -989,9 +989,10 @@ def pipeline(
                     feature_extractor, _from_pipeline=task, **hub_kwargs, **model_kwargs
                 )
 
+                feature_extractor_type = feature_extractor.to_dict().get("feature_extractor_type", None)
                 if (
-                    feature_extractor._processor_class
-                    and feature_extractor._processor_class.endswith("WithLM")
+                    feature_extractor_type
+                    and feature_extractor_type.endswith("WithLM")
                     and isinstance(model_name, str)
                 ):
                     try:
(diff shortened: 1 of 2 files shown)


src/transformers/pipelines/__init__.py
```diff
@@ -989,9 +989,10 @@ def pipeline(
                     feature_extractor, _from_pipeline=task, **hub_kwargs, **model_kwargs
                 )
 
+                feature_extractor_type = feature_extractor.to_dict().get("feature_extractor_type", None)
                 if (
-                    feature_extractor._processor_class
-                    and feature_extractor._processor_class.endswith("WithLM")
+                    feature_extractor_type
+                    and feature_extractor_type.endswith("WithLM")
```
zucchini-nlp, line 995: I think the feature extractor is a simple `Wav2Vec2FeatureExtractor` so this might not work. I found only one processor with LM in hte codebase tbh

zucchini-nlp (commented):
Yeah, I forgot to check if `_processor_class` is used in other places. Not sure if the current fix works because feature extractor type is not the same as processor class. I also had a PR in https://github.com/huggingface/transformers/pull/42920, just waiting for a review 😄

src/transformers/pipelines/__init__.py
pcuenca: Yeah, that was my question too, I wasn't sure how many cases would be impacted. Closing this in favor of yours then.

pcuenca:
Closing in favor of #42920.

Issue #42641, opened 2025-12-05
Cannot inference llava-next with transformers==4.57.1 on dtype="auto" bug


rebel-seinpark:
### System Info

```
- `transformers` version: 4.57.1
- Platform: Linux-5.15.0-161-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- Huggingface_hub version: 0.35.3
- Safetensors version: 0.6.2
- Accelerate version: 1.10.1
- Accelerate config:    not found
- DeepSpeed version: not installed
- PyTorch version (accelerator?): 2.8.0+cpu (NA)
- Tensorflow version (GPU?): 2.18.0 (False)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
- Using distributed or parallel set-up in script?: <fill in>
```

### Who can help?

@zucchini-nlp 

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

```
from transformers import LlavaNextProcessor, LlavaNextForConditionalGeneration
import torch
from PIL import Image
import requests

processor = LlavaNextProcessor.from_pretrained("llava-hf/llava-v1.6-mistral-7b-hf")

model = LlavaNextForConditionalGeneration.from_pretrained("llava-hf/llava-v1.6-mistral-7b-hf", dtype="auto", low_cpu_mem_usage=True) 

# prepare image and text prompt, using the appropriate prompt template
url = "https://github.com/haotian-liu/LLaVA/blob/1a91fc274d7c35a9b50b3cb29c4247ae5837ce39/images/llava_v1_5_radar.jpg?raw=true"
image = Image.open(requests.get(url, stream=True).raw)

# Define a chat history and use `apply_chat_template` to get correctly formatted prompt
# Each value in "content" has to be a list of dicts with types ("text", "image") 
conversation = [
    {

      "role": "user",
      "content": [
          {"type": "text", "text": "What is shown in this image?"},
          {"type": "image"},
        ],
    },
]
prompt = processor.apply_chat_template(conversation, add_generation_prompt=True)

inputs = processor(images=image, text=prompt, return_tensors="pt")

# autoregressively complete prompt
output = model.generate(**inputs, max_new_tokens=100)

print(processor.decode(output[0], skip_special_tokens=True))
```

[...]

PR #42990, opened 2025-12-22
Deprecate dtype per sub config

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/42968, https://github.com/huggingface/transformers/issues/42641

As discussed, this PR deprecated different dtypes per backbone and falls back to the main dtype if a user is requesting different dtypes. 

There is no clean way to support different dtypes when the backbone can have a task head. We either cast hidden states to task dtype in all model files, or we need a workaround to infer task head and force dtype on it when loading. So the better solution is to deprecate dict dtypes, apparently it is not used much


zucchini-nlp:
run-slow: llava, llava_next, llava_next_video, llava_onevision

tests/utils/test_modeling_utils.py
```diff
@@ -551,6 +551,8 @@ def test_model_from_config_dtype_composite(self):
         """
         Test that from_pretrained works with dtype being as a dict per each sub-config in composite config
         Tiny-Llava has saved auto dtype as `torch.float32` for all modules.
+        Note, this is a deprecated feature and we fallbacl to main dtype in all cases below. This tests checks
```
albertvillanova: Nit:
```suggestion
        Note, this is a deprecated feature and we fallback to main dtype in all cases below. This test checks
```

src/transformers/modeling_utils.py
```diff
@@ -868,6 +865,14 @@ def _get_dtype(
                 sub_dtype = getattr(torch, sub_dtype) if isinstance(sub_dtype, str) else sub_dtype
             else:
                 sub_dtype = main_dtype
+            if main_dtype != sub_dtype:
+                logger.warning_once(
+                    f"The dtype for {sub_config_key}={sub_dtype} is different from the main dtype={main_dtype}. "
+                    "Setting different dtypes per backbone model might cause device errors downstream, setting the "
+                    f"dtype={main_dtype} for all modules. Note, using different dtypes per module is deprecated and will "
+                    "be removed in future versions."
+                )
```
qgallouedec: It means that in the future it will raise an exception, right?

zucchini-nlp: yes, after a few released we can switch to an error

rebel-seinpark: It means that if a user sets `dtype="auto"`, all sub modules will be forcibly cast to the specified `main_dtype`?

src/transformers/modeling_utils.py
```diff
@@ -865,6 +865,14 @@ def _get_dtype(
...
```
Cyrilvallez: This will launch the warning for all models where the subconfigs saved a different dtype. Is it something frequent that would launch it for a lot of models, or is it more niche?

Cyrilvallez (commented):
If we do that, then we need to plainly remove dtype being passed as a dict as well!

src/transformers/modeling_utils.py
rebel-seinpark: @Cyrilvallez 
I think this might be more frequent than expected, especially for multimodal models(BLIP, ColPali,...). I’ve already encountered [this issue](https://github.com/huggingface/transformers/issues/42641#issuecomment-3686224305) with several models, such as:
- ColQwen2: https://huggingface.co/vidore/colqwen2-v1.0-hf
- Idefics3: https://huggingface.co/HuggingFaceM4/Idefics3-8B-Llama3

zucchini-nlp: oh, we also could check `Noneness` to reduce the warnings. It's not common for models to have different dtypes saved per backbone, the above ones for ex have the same dtype duplicated many times

Cyrilvallez (approved):
Alright, if there is not many models with different dtypes in subconfigs, it's indeed better to only show the warning when explicitly passing a dict, and silently skipping otherwise!

Issue #42954, opened 2025-12-19
XiaoMi MiMo


ArthurZucker:
### Model description

- https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash/tree/main

### Open source status

- [x] The model implementation is available
- [x] The model weights are available

ghost:
Hi @ArthurZucker , I would love to take this on and contribute the implementation for **MiMo-V2-Flash** to `transformers`.

I've reviewed the model's technical details, specifically the **Hybrid Sliding Window Attention** and the **MoE structure** (309B params/15B active). Given the large weight size, I plan to start by:

1. Setting up the model architecture using the **Modular Transformers** approach.
2. Verifying the implementation locally with a **scaled-down dummy configuration** to ensure tensor shapes are correct without needing to download the full weights.
3. Working on a conversion script for the weights once the architecture is stable.

I'll open a Draft PR shortly to track my progress. Are there any specific existing models you recommend I use as a reference for the Hybrid Attention implementation?

Aaraviitkgp:
@Aznix07 Is it ok if we work together on this ?

PR #42995, opened 2025-12-22
implemented mimo model

Aaraviitkgp:
Issue : https://github.com/huggingface/transformers/issues/42954

I have implemented mimo model,
I have just changed rope_type as it was default which was causing error so now it is none by default also changed and have made some changes in class MiMoV2FlashRotaryEmbedding, added files for mimo integration.

@ArthurZucker


Aaraviitkgp:
@ArthurZucker The setup and quality workflow seems to fail, I am not able to understand why it is coming.

Liuweixiong0118:
@Aaraviitkgp  hello, i found your transformers github, https://github.com/Aaraviitkgp/transformers, have four mimo-v2-flash implemented branch. 
which one is the trainable final version of mimo-v2-flash. 
can i use the origin repo https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash model to train?
Thank you for sharing and look forward to your reply

PR #42997, opened 2025-12-22
Do not return a tuple in mistral tokenizer Automapping

molbap:
# What does this PR do?

As per title. Currently seems to break `from_pretrained` IIUC

repro:

```python
model_id = "mistralai/Mistral-Small-3.2-24B-Instruct-2506"

print(tokenizer("about mistral"))
```

breaks on `main`

cc @itazap @ArthurZucker


ArthurZucker (approved):
ty can you add a test it should have never failed this way!

molbap:
Added a basic test just to check that we indeed map to strings/valid class names!

PR #42998, opened 2025-12-22
[`fix:`]: initialize `ModernBertUnpaddedRotaryEmbedding` when loading from a checkpoint

emapco:
# What does this PR do?

 This fix improves accuracy and prevents existing ModernBert models from yielding NaN tensors. Previously, when a pretrained  modernbert model is loaded, the `ModernBertUnpaddedRotaryEmbedding`'s `inv_freq` non-persistent buffer is not initialized resulting in `NaN` loss and logits.

Without fix: 
- `ModernBertForSequenceClassification`  `ModernBertModelIntegrationTest.test_inference_sequence_classification_flash_attention_2` rarely returns NaN tensors but instead fails due to inaccurate logits
- `ModernBertModelIntegrationTest.test_inference_sequence_classification_flash_attention_2_modernbert_base` always fails due to NaN logits.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker @Cyrilvallez


emapco:
I should add that I used the `flash-attn` package that I installed via `uv pip install flash-attn --no-build-isolation` with `torch==2.9` and `cuda12.8`.

Cyrilvallez:
Fixed in https://github.com/huggingface/transformers/pull/43115!

PR #43000, opened 2025-12-22
Generation config boolean defaults

zucchini-nlp:
# What does this PR do?

Continues on https://github.com/huggingface/transformers/pull/42958 and makes sure the opposite case, when config has a non-default value saved but the user wants to set it back to a default boolean, is also covered


zucchini-nlp:
run-slow: llama, gemma2, qwen2, bart, t5, whisper

src/transformers/configuration_utils.py
```diff
@@ -1116,11 +1116,15 @@ def register_for_auto_class(cls, auto_class="AutoConfig"):
...
-        Gets the non-default generation parameters on the PreTrainedConfig instance
+        Checks if there are generation parameters in `PreTrainedConfig` instance. Note that
+        we should not save generation params in PreTrainedConfig, and we will raise error
+        if there are any.
         """
         generation_params = {}
         default_config = self.__class__().to_dict() if not self.has_no_defaults_at_init else {}
         for key in GenerationConfig._get_default_generation_params().keys():
+            if key == "use_cache":
+                continue  # common key for most models
```
zucchini-nlp, line 1127: btw, could we remove `use_cache` as a key from config and instead keep it only in `generation_config`? Imo it isn't really a model attribute

src/transformers/trainer.py
```diff
@@ -751,8 +751,8 @@ def __init__(
         self._created_lr_scheduler = False
 
         # Set use_cache for the model
-        if getattr(self.model, "config", None) is not None:
-            self.model.config.use_cache = self.args.use_cache
+        if getattr(self.model, "generation_config", None) is not None:
+            self.model.generation_config.use_cache = self.args.use_cache
```
zucchini-nlp: model config might not contain `use_cache` unless it is in config's own init. We don't assign generation attributes anymore in base config class 

So checking `generation_config` is more reliable

src/transformers/configuration_utils.py
vasqu: Hmm unsure. I understand the point but that would be hugely breaking for a lot of users and external libraries - on the other hand we could include this for v5 if you feel motivated :thinking:

src/transformers/generation/configuration_utils.py
```diff
@@ -346,11 +346,11 @@ def __init__(self, **kwargs):
         self.stop_strings = kwargs.pop("stop_strings", None)
 
         # Parameters that control the generation strategy used
-        self.do_sample = kwargs.pop("do_sample", False)
+        self.do_sample = kwargs.pop("do_sample", None)
```
vasqu, line 353: Don't forget to update the respective docstrings

src/transformers/generation/configuration_utils.py
```diff
@@ -483,7 +483,7 @@ def get_generation_mode(self, assistant_model: Optional["PreTrainedModel"] = Non
         if self.constraints is not None or self.force_words_ids is not None:
             generation_mode = GenerationMode.CONSTRAINED_BEAM_SEARCH
         elif self.num_beams is None or self.num_beams == 1:
-            if not self.do_sample:
+            if self.do_sample is not True:
```
vasqu, line 490:
```suggestion
            if self.do_sample is False:
```
super nit, seen this multiple times

src/transformers/generation/configuration_utils.py
```diff
@@ -537,6 +537,7 @@ def _get_default_generation_params() -> dict[str, Any]:
             "max_length": 20,
             "min_length": 0,
             "do_sample": False,
+            "use_cache": True,
```
vasqu, line 544: We should maybe explicitly mention this function in the init of this generation config. While we default to `None` everywhere now, it is a bit misleading in the init because we expect some sort of overwrite with this function (at least).

src/transformers/models/clvp/modeling_clvp.py
vasqu, line 1: Are these changes really relevant? Seems unrelated to me

src/transformers/trainer.py
vasqu: What's the current logic here then? kwarg > config > generation config? Seems like another case of bad relations here :sob: 

It does feel a bit out of place tho to me. I'd like to keep this PR strictly to the default values, this semi progresses the point of removing reliance on `use_cache` in the config. On another note, we use the `check_model_inputs` decorator with the config `use_cache` so it might be hard to disentangle this completely

vasqu (commented):
Left a few comments, I'd be pro changing the `use_cache` changes (at least for now) - see my last comment.

Also do we still need https://github.com/huggingface/transformers/blob/a616b914451ffd3bb3553874e67af7fe47e8adeb/src/transformers/generation/utils.py#L1800-L1812
I added this only due to these not being none within their defaults, but now we should update them properly without that workaround, no? 

Last note, semi-unrelated, let's also keep track of https://github.com/huggingface/transformers/pull/42702/files#r2635806800 in another PR cc @ebezzam

src/transformers/generation/configuration_utils.py
Cyrilvallez: Actually, it should be `if not self.do_sample` haha

src/transformers/generation/configuration_utils.py
```diff
@@ -368,28 +368,28 @@ def __init__(self, **kwargs):
...
-        self.output_hidden_states = kwargs.pop("output_hidden_states", False)
-        self.output_scores = kwargs.pop("output_scores", False)
-        self.output_logits = kwargs.pop("output_logits", False)
-        self.return_dict_in_generate = kwargs.pop("return_dict_in_generate", False)
+        self.num_return_sequences = kwargs.pop("num_return_sequences", None)
+        self.output_attentions = kwargs.pop("output_attentions", None)
+        self.output_hidden_states = kwargs.pop("output_hidden_states", None)
+        self.output_scores = kwargs.pop("output_scores", None)
+        self.output_logits = kwargs.pop("output_logits", None)
+        self.return_dict_in_generate = kwargs.pop("return_dict_in_generate", None)
```
Cyrilvallez, line 396: Sorry, I'm a bit lost into the motivation the PR and making those values to None (which is usually not a super good idea if they are boolean by essence) - could you give me a bit of context on the PR?

src/transformers/configuration_utils.py
zucchini-nlp: it will definitely break stuff 😿 Just a thought for now

src/transformers/trainer.py
zucchini-nlp: we are nor removing `use_cache` yet in this PR. I hacked by the above check that doesn't include `use_cache` in generation params

I will revert it back, forgot to do so after testing some stuff

zucchini-nlp: The logic stays as is, so yes, kwarg > config > generation config

src/transformers/generation/configuration_utils.py
zucchini-nlp: `if not do_sample` will result in false positives when do_sample is `None` 😢  Gotta be explicitly checking for not True, since the value of `None` is same as having `False`, i.e. the new default value

src/transformers/generation/configuration_utils.py
zucchini-nlp: These defaults are used in `generate()` loop for BC only so most ppl should not see them after init. I will mention in the docs somewhere to amek it clear

src/transformers/models/clvp/modeling_clvp.py
zucchini-nlp: some tests were failing without it because the re-loaded model has a incorrect config type. And we couldn't find generation keys in it
It was luckily passing in main branch

src/transformers/configuration_utils.py
zucchini-nlp: so, this is a workaround we have to not throw errors that `model config contains generation params`. We won't throw errors anyway if `use_cache` is in model config signature but unfortunately many don't have it documented

vasqu: Gotcha, maybe at some point we can :eyes:

src/transformers/generation/configuration_utils.py
```diff
@@ -101,6 +101,10 @@ class GenerationConfig(PushToHubMixin):
 
     </Tip>
 
+    Note: the configuration field that are still `None` will be overriden by `GenerationConfig._get_default_generation_params()`
+    during the generation loop. If you want to use different values for these fiels, make sure to expllicitly set then in the
+    generation config.
```
```suggestion
    during the generation loop. If you want to use different values for these fields, make sure to explicitly set them in the
    generation config.
```

vasqu: That note is super important imo to understand for an outsider thx :pray:

src/transformers/generation/configuration_utils.py
vasqu: Can we add a small comment then, that's such a weird case lol.

src/transformers/models/clvp/modeling_clvp.py
vasqu: Ok, then let's keep it, not wrong was just surprised

vasqu (approved):
LGTM, could you just double check if we still need https://github.com/huggingface/transformers/blob/a616b914451ffd3bb3553874e67af7fe47e8adeb/src/transformers/generation/utils.py#L1800-L1812

Now that these values are defaulting to `None` as well, I suspect we might not need it anymore

zucchini-nlp:
> LGTM, could you just double check if we still need

my bad, didn't see the comment on it. Not needed!

src/transformers/generation/configuration_utils.py
zucchini-nlp: The motivation was to fix a ton of edge cases when creating a generation config. In the past we had ugly workaround to allow users to change a generation param back to default if different value was saved on the hub (for ex https://github.com/huggingface/transformers/issues/42762).

That was fixed in https://github.com/huggingface/transformers/pull/42702 and I deleted a lot of code, and defined a clear priority list, Next step will be to disallow users to pass both: a generation config and kwargs. Instead ask to put all kwargs in config. The below is accepted rn and is a bad practice imo

```python
conf = GenerationConfig(do_sample=False)
text = model.generate(input_ids, generation_config=conf, max_new_tokens=128, temperature=1.6)
```

PR #43006, opened 2025-12-22
Fix deepspeed + quantization

SunMarc:
# What does this PR do?

This PR continues to try fixing some tests on deepspeed and quantization side.

One big issue was that we were not moving the missing weights correctly, making the initialization unsuccessful. Another one was that the checkpoint where not loaded correctly due to the prefix not being fixed.


src/transformers/modeling_utils.py
```diff
@@ -4373,12 +4373,27 @@ def _move_missing_keys_from_meta_to_cpu(
...
+        for key, cbuffer in self.named_non_persistent_buffers():
+            if is_deepspeed_zero3_enabled() and not is_quantized:
+                import deepspeed
+
+                with deepspeed.zero.GatheredParameters([buffer], modifier_rank=0):
+                    # needed for the sharding
+                    param.data.copy_(torch.empty_like(param, device=param.device))
+            else:
+                value = torch.empty_like(buffer, device="cpu")
+                _load_parameter_into_model(self, key, value)
```
SunMarc, line 4396: need to correctly move the missing weights

src/transformers/integrations/deepspeed.py
```diff
@@ -304,7 +304,15 @@ def _load_state_dict_into_zero3_model(model_to_load, state_dict):
...
-    missing_keys = set(model_to_load.state_dict().keys())
+    meta_model_state_dict = model_to_load.state_dict()
+    missing_keys = set(meta_model_state_dict.keys())
+
+    prefix_model = getattr(model_to_load, "base_model_prefix", None)
+    # take care of the care where in the checkpoint we don't have the prefix
+    state_dict = {
+        (f"{prefix_model}.{k}" if meta_model_state_dict.get(f"{prefix_model}.{k}") is not None else k): v
+        for k, v in state_dict.items()
+    }
```
SunMarc, line 315: deal with prefix

src/transformers/integrations/deepspeed.py
```diff
@@ -304,7 +304,15 @@ def _load_state_dict_into_zero3_model(model_to_load, state_dict):
         state_dict._metadata = metadata
 
     error_msgs = []
-    missing_keys = set(model_to_load.state_dict().keys())
+    meta_model_state_dict = model_to_load.state_dict()
+    missing_keys = set(meta_model_state_dict.keys())
+
+    prefix_model = getattr(model_to_load, "base_model_prefix", None)
+    # take care of the care where in the checkpoint we don't have the prefix
```
MekkCyber:
```suggestion
    # take care of the case where in the checkpoint we don't have the prefix
```

src/transformers/modeling_utils.py
```diff
@@ -4373,12 +4373,27 @@ def _move_missing_keys_from_meta_to_cpu(
...
             param = self.get_parameter_or_buffer(key)
-            value = torch.empty_like(param, device="cpu")
-            _load_parameter_into_model(self, key, value)
+            if is_deepspeed_zero3_enabled() and not is_quantized:
+                import deepspeed
+
+                with deepspeed.zero.GatheredParameters([param], modifier_rank=0):
+                    # needed for the sharding
+                    param.data.copy_(torch.empty_like(param, device=param.device))
+                    print(param)
```
MekkCyber:
```suggestion
```

MekkCyber (approved):
Thanks for fixing

PR #43008, opened 2025-12-22
Torch distrib smallfix

molbap:
# What does this PR do?

Seems that the TP/dist integ refactor broke a method to count bytes, which is in from_pretrained. cc @3outeille as well if needed (I'm setting a 1 world size by default without initializing dist :shrug: )

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -4574,7 +4574,11 @@ def get_total_byte_count(
 
         if len(tp_plan) > 0:
             is_part_of_plan = _get_parameter_tp_plan(param_name, tp_plan, is_weight=True) is not None
-            param_byte_count //= torch.distributed.get_world_size() if is_part_of_plan else 1
+            if is_part_of_plan:
+                world_size = 1
+                if torch.distributed.is_available() and torch.distributed.is_initialized():
+                    world_size = max(torch.distributed.get_world_size(), 1)
+                param_byte_count //= world_size
 
         total_byte_count[device] += param_byte_count
     return total_byte_count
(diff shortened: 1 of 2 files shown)


molbap:
Referencing https://github.com/huggingface/transformers/pull/43003 and https://github.com/huggingface/transformers/pull/42809 for reference

src/transformers/modeling_utils.py
```diff
@@ -4574,7 +4574,11 @@ def get_total_byte_count(
 
         if len(tp_plan) > 0:
             is_part_of_plan = _get_parameter_tp_plan(param_name, tp_plan, is_weight=True) is not None
-            param_byte_count //= torch.distributed.get_world_size() if is_part_of_plan else 1
+            if is_part_of_plan:
+                world_size = 1
+                if torch.distributed.is_available() and torch.distributed.is_initialized():
+                    world_size = max(torch.distributed.get_world_size(), 1)
+                param_byte_count //= world_size
```
Cyrilvallez: Oh yeah I messed up the `tp_plan` always exist not just when initialized haha
Ca we simply do 
```python
tp_plan = model._tp_plan if everything initialized else {}
```
at the beginning?
Would be simpler IMO!

Cyrilvallez (approved):
LGTM, thanks for fixing! Let's just simplify a bit!

src/transformers/modeling_utils.py
molbap: yes true, we only need tp_plan if dist is init anyways, will do that and merge!

molbap: done!

PR #43015, opened 2025-12-23
FIX: TF32 warning (#43012)

### LINKED ISSUE #43012: Compiling a bfloat16 model triggers float32 precision PyTorch warning
### System Info

5.0.0rc1, 2.9.1

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

```
/.../.venv/lib/python3.12/site-packages/torch/_inductor/compile_fx.py:312: UserWarning: TensorFloat32 tensor cores for float32 matrix multiplication available but not enabled. Consider setting `torch.set_float32_matmul_precision('high')` for better performance.
  warnings.warn(
```

```python
model = transformers.AutoModelForCausalLM.from_pretrained("PleIAs/Monad", attn_implementation = 'sdpa', dtype = torch.bfloat16)
model = torch.compile(model, mode = 'reduce_overhead')
model(...)
```

Is this warning because the model still does some fp32 matmul?

### Expected behavior

no warning when bfloat16 model is used

shantanugupta2004:
# What does this PR do?
This PR replaces the matrix multiplication operator (@) with broadcasting element-wise multiplication (*) in the RotaryEmbedding implementation for several major models (Llama, Mistral, Mixtral, Qwen2, Gemma, Gemma2).
When compiling a model with torch.compile in bfloat16, the RoPE frequency calculation (which is intentionally kept in float32 for precision) triggers a UserWarning regarding TensorFloat32 (TF32) if it's not enabled.
Since the shapes involved in this specific operation [batch, dim/2, 1] and [batch, 1, seq_len] result in an outer product, using @ is mathematically equivalent to * with broadcasting. However, using * avoids the "matrix multiplication" code path in the compiler, effectively silencing the false-positive warning and potentially offering a minor performance optimization by avoiding a full GEMM call for a simple outer product.

Fixes #43012 

## Before submitting
      Pull Request section?
- [x] Was this discussed/approved via a Github issue [link](https://github.com/huggingface/transformers/issues/43012)? 

## Who can review?

Anyone in the community is free to review the PR once the tests have passed.

### DIFF
--- a/src/transformers/models/afmoe/modeling_afmoe.py
+++ b/src/transformers/models/afmoe/modeling_afmoe.py
@@ -99,7 +99,7 @@ class AfmoeRotaryEmbedding(nn.Module):
 
         device_type = x.device.type if isinstance(x.device.type, str) and x.device.type != "mps" else "cpu"
         with maybe_autocast(device_type=device_type, enabled=False):  # Force float32
-            freqs = (inv_freq_expanded.float() @ position_ids_expanded.float()).transpose(1, 2)
+            freqs = (inv_freq_expanded.float() * position_ids_expanded.float()).transpose(1, 2)
             emb = torch.cat((freqs, freqs), dim=-1)
             cos = emb.cos() * self.attention_scaling
             sin = emb.sin() * self.attention_scaling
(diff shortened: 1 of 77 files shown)


ArthurZucker (commented):
This does not sound bad but at the same time this is very specific, we want users to know that full precision could me necessary

Rocketknight1:
I did some checking and the output of `@` or `*` is completely identical in any precision, which makes me suspect the same kernel is being called in both cases. This means there should be no change in performance, but I think it's fine to use `*` to suppress the TF32 warning no? I think it will probably annoy/confuse users who are running the model in `bfloat16` and don't understand why `float32` matmuls are happening. cc @ArthurZucker

PR #43021, opened 2025-12-23
Move missing weights and non-persistent buffers to correct device earlier

Cyrilvallez:
# What does this PR do?

As per the title. Better to send everthing directly to the correct device, rather than cpu, then initialize, then move to device

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -4356,7 +4356,7 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         # will be re-initialized for nothing (which can be quite long)
         for key in missing_keys - self.all_tied_weights_keys.keys():
             param = self.get_parameter_or_buffer(key)
-            param_device = get_device(device_map, key)
+            param_device = get_device(device_map, key, valid_torch_device=True)
             value = torch.empty_like(param, device=param_device)
             # For TP, we may need to shard the param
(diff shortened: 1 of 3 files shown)


src/transformers/modeling_utils.py
```diff
@@ -4345,13 +4323,23 @@ def get_compiled_call(self, compile_config: Optional[CompileConfig]) -> Callable
...
+        """Move the missing keys (keys that are part of the model parameters, but were NOT found in the loaded state dicts)
+        back from meta device to their device according to the `device_map` if any, else cpu. Takes care of sharding those
+        missing parameters if `device_mesh` is provided, i.e. we are using TP.
+        All non-persistent buffers are also moved back to the correct device (they are not part of the state_dict, but are
+        not missing either).
         """
         is_quantized = hf_quantizer is not None
+        # This is the only case where we do not initialize the model on meta device, so we don't have to do anything here
+        if is_deepspeed_zero3_enabled() and not is_quantized:
+            return
```
Cyrilvallez, line 4342: cc @SunMarc, could you make extra sure that simply doing this is enough? It should be as it's technically useless to reset the internal param, as it was not on meta before anyway - but let's make sure as you made that change yesterday

SunMarc (approved):
Nice, thanks ! This is much better

RatStar811 (changes requested):
Pull request

RatStar811 (approved):

PR #42952, opened 2025-12-19
Add FA2 fallback (kernel) support to the modernbert model

YangKai0616:
# What does this PR do?

1. Add **FA2** fallback (kernel) support to the `modernbert` model.
Related UT: 
```
FAILED tests/models/pe_audio/test_modeling_pe_audio.py::PeAudioModelTest::test_cpu_offload - KeyError: 'kernels-community/flash-attn2'
FAILED tests/models/pe_audio/test_modeling_pe_audio.py::PeAudioModelTest::test_determinism - KeyError: 'kernels-community/flash-attn2'
FAILED tests/models/pe_audio/test_modeling_pe_audio.py::PeAudioModelTest::test_disk_offload_bin - KeyError: 'kernels-community/flash-attn2'
FAILED tests/models/pe_audio/test_modeling_pe_audio.py::PeAudioModelTest::test_disk_offload_safetensors - KeyError: 'kernels-community/flash-attn2'
FAILED tests/models/pe_audio/test_modeling_pe_audio.py::PeAudioModelTest::test_model - KeyError: 'kernels-community/flash-attn2'
FAILED tests/models/pe_audio/test_modeling_pe_audio.py::PeAudioModelTest::test_model_outputs_equivalence - KeyError: 'kernels-community/flash-attn2'
FAILED tests/models/pe_audio/test_modeling_pe_audio.py::PeAudioModelTest::test_model_parallelism - KeyError: 'kernels-community/flash-attn2'
```

2. When FA2 is available, `tests/models/pe_audio/test_modeling_pe_audio.py::PeAudioModelTest::test_init_weights_can_init_buffers` will report an error 
```AssertionError: False is not true : The following buffers are not properly handled in `_init_weights()` (the model should be able to reinitialize them correctly if t...```.

Root cause: The `inv_freq` buffer in `ModernBertUnpaddedRotaryEmbedding` (inherited from `flash_attn's 
RotaryEmbedding`) was not properly initialized when loading from meta device. Added 
reinitialization logic in `_init_weights()` to compute and copy correct `inv_freq` values.

PR #43030, opened 2025-12-24
[Model] Refactor modernbert with the attention interface

YangKai0616:
# What does this PR do?

This PR:

1. Refactors `ModernBERT's Flash Attention` implementation using the built-in interfaces from `Transformers`.
2. Fixes some UT issues in the `tests/models/modernbert/test_modeling_modernbert.py` and `tests/models/pe_video/test_modeling_pe_video.py` test files. 

Background PR: https://github.com/huggingface/transformers/pull/42952.


YangKai0616:
Hey @vasqu , could you please review it again? Thank you very much!

vasqu:
Nice thx a lot @YangKai0616, I'll take a look tomorrow, just coming back from the holidays :hugs:

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -557,126 +337,26 @@ def compute_default_rope_parameters(
 
 def eager_attention_forward(
```
vasqu: Can we use the eager attention of bert or similar models? E.g. you can simply import the function here in the modular file and it will be automatically inherited to the modeling file

At least we can adopt the scaling default from bert

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -704,74 +384,91 @@ def __init__(self, config: ModernBertConfig, layer_id: Optional[int] = None):
...
         self.all_head_size = self.head_dim * self.num_heads
+        self.scaling = self.head_dim**-0.5
         self.Wqkv = nn.Linear(config.hidden_size, 3 * self.all_head_size, bias=config.attention_bias)
-        layer_type = config.layer_types[layer_id]
 
-        if layer_id % config.global_attn_every_n_layers != 0:
-            self.local_attention = (config.local_attention // 2, config.local_attention // 2)
-            max_position_embeddings = config.local_attention
+        if layer_idx % config.global_attn_every_n_layers != 0:
+            self.sliding_window = config.local_attention
```
vasqu: Pretty sure we need to add a +1 here, see https://github.com/huggingface/transformers/blob/828c039e08dd91af7ce453567ab1da3370df1f04/src/transformers/modeling_flash_attention_utils.py#L545-L550

Might need to update the comment there as well

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -704,74 +384,91 @@ def __init__(self, config: ModernBertConfig, layer_id: Optional[int] = None):
         self.num_heads = config.num_attention_heads
         self.head_dim = config.hidden_size // config.num_attention_heads
         self.all_head_size = self.head_dim * self.num_heads
+        self.scaling = self.head_dim**-0.5
         self.Wqkv = nn.Linear(config.hidden_size, 3 * self.all_head_size, bias=config.attention_bias)
```
vasqu: There are a lot of unused attributes now
- `deterministic_flash_attn` (we allow `deterministic` as kwarg or use the `FLASH_ATTENTION_DETERMINISTIC` env)
- `num_heads` shouldn't be used in general because in multi GPU cases with TP we reduce the num heads by a factor of the available gpus, let's update the views in the forward as well (adding a comment there in  sec)
- `self.all_head_size` has no usage outside of the linear init, we can just directly put it there imo

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -704,74 +384,91 @@ def __init__(self, config: ModernBertConfig, layer_id: Optional[int] = None):
...
 
         self.Wo = nn.Linear(config.hidden_size, config.hidden_size, bias=config.attention_bias)
         self.out_drop = nn.Dropout(config.attention_dropout) if config.attention_dropout > 0.0 else nn.Identity()
 
     def forward(
         self,
         hidden_states: torch.Tensor,
-        position_embeddings: Optional[torch.Tensor] = None,
+        attention_mask: Optional[torch.Tensor] = None,
+        sliding_window_mask: Optional[torch.Tensor] = None,
```
vasqu: We shouldn't have separate masks here, just one that is properly constructed like in the decoder models.
Ig we are missing a proper mask creation function that handles bidirectional sliding window.

I can take a look lmk, otherwise you can take a close look at the original function https://github.com/huggingface/transformers/blob/828c039e08dd91af7ce453567ab1da3370df1f04/src/transformers/masking_utils.py#L873
with modifications like here
https://github.com/huggingface/transformers/blob/828c039e08dd91af7ce453567ab1da3370df1f04/src/transformers/masking_utils.py#L1011-L1013
https://github.com/huggingface/transformers/blob/828c039e08dd91af7ce453567ab1da3370df1f04/src/transformers/masking_utils.py#L1016
and copy basically this https://github.com/huggingface/transformers/blob/828c039e08dd91af7ce453567ab1da3370df1f04/src/transformers/masking_utils.py#L115-L119
with the proper functions (i.e. bidirectional mask + a new sliding window fn)

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -704,74 +384,91 @@ def __init__(self, config: ModernBertConfig, layer_id: Optional[int] = None):
...
+        position_embeddings: Optional[tuple[torch.Tensor, torch.Tensor]] = None,
         output_attentions: Optional[bool] = False,
         **kwargs,
-    ) -> torch.Tensor:
+    ) -> tuple[torch.Tensor, ...]:
+        input_shape = hidden_states.shape[:-1]
+        original_dtype = hidden_states.dtype
+
         qkv = self.Wqkv(hidden_states)
+        qkv = qkv.view(*input_shape, 3, self.num_heads, self.head_dim)
```
vasqu:
```suggestion
        qkv = qkv.view(*input_shape, 3, -1, self.head_dim)
```
mentioned it before, if there is a case for TP at some point, this is more robust

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -704,74 +384,91 @@ def __init__(self, config: ModernBertConfig, layer_id: Optional[int] = None):
...
+        # Transpose to (batch, num_heads, seq_len, head_dim) for attention interface
+        query_states = query_states.transpose(1, 2)
+        key_states = key_states.transpose(1, 2)
+        value_states = value_states.transpose(1, 2)
+
+        cos, sin = position_embeddings
+        query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin, unsqueeze_dim=1)
 
-        attn_outputs = MODERNBERT_ATTENTION_FUNCTION[self.config._attn_implementation](
+        effective_attention_mask = sliding_window_mask if self.sliding_window is not None else attention_mask
```
vasqu: Yea as mentioned with mask creation we should follow something like this https://github.com/huggingface/transformers/blob/828c039e08dd91af7ce453567ab1da3370df1f04/src/transformers/models/gemma3/modeling_gemma3.py#L552-L573
where we properly prepare the correct mask outside

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -704,74 +384,91 @@ def __init__(self, config: ModernBertConfig, layer_id: Optional[int] = None):
...
+            attention_interface = ALL_ATTENTION_FUNCTIONS[self.config._attn_implementation]
+
+        # Flash Attention 2 only supports fp16 and bf16 - convert if needed
+        if "flash" in self.config._attn_implementation and query_states.dtype not in (
+            torch.float16,
+            torch.bfloat16,
+        ):
+            query_states = query_states.to(torch.bfloat16)
+            key_states = key_states.to(torch.bfloat16)
+            value_states = value_states.to(torch.bfloat16)
```
vasqu: We already handle this ourselves, e.g. see https://github.com/huggingface/transformers/blob/828c039e08dd91af7ce453567ab1da3370df1f04/src/transformers/integrations/flash_attention.py#L12

So we can remove this imo

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -704,74 +384,91 @@ def __init__(self, config: ModernBertConfig, layer_id: Optional[int] = None):
...
-            position_embeddings=position_embeddings,
-            output_attentions=output_attentions,
+            query_states,
+            key_states,
+            value_states,
+            effective_attention_mask,
+            dropout=self.attention_dropout if self.training else 0.0,
+            scaling=self.scaling,
+            sliding_window=self.sliding_window,
+            is_causal=self.is_causal,
```
vasqu:
```suggestion
```
We rely on the module's attribute. There are very special cases like clip but that's an exception

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -704,74 +384,91 @@ def __init__(self, config: ModernBertConfig, layer_id: Optional[int] = None):
...
         )
-        hidden_states = attn_outputs[0]
-        hidden_states = self.out_drop(self.Wo(hidden_states))
 
-        return (hidden_states,) + attn_outputs[1:]  # add attentions if outputted
+        attn_output = attn_output.to(original_dtype)
+        attn_output = attn_output.reshape(*input_shape, -1).contiguous()
+        hidden_states = self.out_drop(self.Wo(attn_output))
+
+        return (hidden_states, attn_weights) if output_attentions else (hidden_states,)
```
vasqu: It feels super basic, we might be able to inherit from something like gemma3 IMO. It would need another round of reactors with `check_model_inputs` that collects the weights if needed

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -782,19 +479,13 @@ def forward(
         hidden_states: torch.Tensor,
         attention_mask: Optional[torch.Tensor] = None,
         sliding_window_mask: Optional[torch.Tensor] = None,
-        position_ids: Optional[torch.LongTensor] = None,
-        cu_seqlens: Optional[torch.Tensor] = None,
-        max_seqlen: Optional[int] = None,
         position_embeddings: Optional[torch.Tensor] = None,
         output_attentions: Optional[bool] = False,
```
vasqu: Ideally, we would have **kwargs here for packed training/inference, you can take look at bert for example

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -1018,35 +694,19 @@ def forward(
...
-                    inputs_embeds, indices, cu_seqlens, max_seqlen, *_ = _unpad_modernbert_input(
-                        inputs=inputs_embeds, attention_mask=attention_mask
-                    )
-            if position_ids is None:
-                position_ids = indices.unsqueeze(0)
-        else:
-            if position_ids is None:
-                position_ids = torch.arange(seq_len, device=device).unsqueeze(0)
+        if position_ids is None:
+            position_ids = torch.arange(seq_len, device=device).unsqueeze(0).expand(batch_size, -1)
```
vasqu: Do we really need the expand?

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -1018,35 +694,19 @@ def forward(
...
 
+        if "flash" not in self.config._attn_implementation:
             attention_mask, sliding_window_mask = self._update_attention_mask(
                 attention_mask, output_attentions=output_attentions
             )
```
vasqu: Reiterating the mask creation we should be able to use https://github.com/huggingface/transformers/blob/828c039e08dd91af7ce453567ab1da3370df1f04/src/transformers/masking_utils.py#L873
or more specifically like I mentioned in the other comment where we create a new fn for sliding window

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -1064,9 +724,6 @@ def forward(
                 hidden_states,
                 attention_mask=attention_mask,
                 sliding_window_mask=sliding_window_mask,
-                position_ids=position_ids,
-                cu_seqlens=cu_seqlens,
-                max_seqlen=max_seqlen,
                 position_embeddings=position_embeddings[encoder_layer.attention_type],
                 output_attentions=output_attentions,
```
vasqu: We will need kwargs to support packed attn, mentioned it already somewhere. Avoiding repeating it now

tests/models/modernbert/test_modeling_modernbert.py
```diff
@@ -157,11 +158,13 @@ def get_config(self):
             if test_name in ("test_retain_grad_hidden_states_attentions", "test_inputs_embeds_matches_input_ids"):
                 config.reference_compile = False
             # Some tests require attentions to be outputted, in that case we'll set the attention implementation to eager
-            # as the others don't support outputted attentions
+            # as the others don't support outputted attentions.
+            # test_all_tensors_are_parameter_or_buffer runs on CPU but ModernBERT defaults to flash_attention_2 (GPU only).
```
vasqu: Why not override the test and force sdpa or similar?

tests/models/pe_audio/test_modeling_pe_audio.py
vasqu, line 1: Slightly unrelated? Let's focus on modernbert only first - same for pe video

vasqu (commented):
Much needed! I left some comments, especially the masking can be further improved along possibly adding `check_model_inputs` decorator with the usage of `**kwargs`. 

Lmk if I can help out there, the masking is pretty core so it might be a bit too much

src/transformers/models/modernbert/modular_modernbert.py
YangKai0616: After removal, the tests need to ensure that the model weights are in bf16 format, otherwise it will report `RuntimeError: FlashAttention only support fp16 and bf16 data type`. As a result, I have modified some of the unit test (UT) code accordingly. I'm not entirely sure if this approach is appropriate.
Could you please share your thoughts on this?

YangKai0616:
Hey @vasqu, I've updated the code based on your comments and removed the unrelated UT changes. Please take another look. Thanks!

src/transformers/masking_utils.py
```diff
@@ -117,6 +117,25 @@ def sliding_window_causal_mask_function(sliding_window: int) -> Callable:
...
 
 
+def sliding_window_bidirectional_overlay(sliding_window: int) -> Callable:
+    """
+    This is an overlay depicting a bidirectional sliding window pattern.
+    Tokens can attend to other tokens within `sliding_window` positions in both directions.
+    """
+
+    def inner_mask(batch_idx: int, head_idx: int, q_idx: int, kv_idx: int) -> bool:
+        return (q_idx - sliding_window < kv_idx) & (kv_idx < q_idx + sliding_window + 1)
```
vasqu: Let's align with gemma3, this was done in a rush but it's time to do it properly https://github.com/huggingface/transformers/blob/59b9cb06f4a19c6d6601cb3e96c018f170ff6808/src/transformers/models/gemma3/modeling_gemma3.py#L488-L491

src/transformers/masking_utils.py
```diff
@@ -1058,6 +1077,81 @@ def create_sliding_window_causal_mask(
...
+        input_embeds (`torch.Tensor`):
+            The input embeddings of shape (batch_size, query_length, hidden_dim). This is used only to infer the
+            batch size, query length and dtype.
+        attention_mask (`torch.Tensor`, optional):
+            The 2D attention mask corresponding to padded tokens of shape (batch_size, kv_length).
+            It can also be an already prepared 4D mask, in which case it is returned as-is.
+        or_mask_function (`Callable`, optional):
+            An optional mask function to combine with the sliding window mask function (by doing the union of both).
+        and_mask_function (`Callable`, optional):
+            An optional mask function to combine with the sliding window mask function (by doing the intersection of both).
```
vasqu: Let's copy the descriptions from the bidirectional version (minus the encoder_hidden_state) https://github.com/huggingface/transformers/blob/59b9cb06f4a19c6d6601cb3e96c018f170ff6808/src/transformers/masking_utils.py#L883-L898

src/transformers/masking_utils.py
```diff
@@ -1058,6 +1077,81 @@ def create_sliding_window_causal_mask(
...
+    """
+    cache_position = torch.arange(input_embeds.shape[1], device=input_embeds.device, dtype=torch.long)
```
vasqu, line 1155:
```suggestion
    # Due to the logic surrounding `cache_position` in inferring query-related information, we
    # construct a dummy tensor imitating initial positions
    cache_position = torch.arange(input_embeds.shape[1], device=input_embeds.device, dtype=torch.long)
```
same here let's follow closely

src/transformers/masking_utils.py
```diff
@@ -1058,6 +1077,81 @@ def create_sliding_window_causal_mask(
...
+
+    early_exit, attention_mask, _, kv_length, kv_offset = _preprocess_mask_arguments(
```
vasqu, line 1158:
```suggestion
    # We ignore a few irrelevant arguments at the end as we do not have a (growing) cache here
    early_exit, attention_mask, _, kv_length, kv_offset = _preprocess_mask_arguments(
```

src/transformers/masking_utils.py
```diff
@@ -1058,6 +1077,81 @@ def create_sliding_window_causal_mask(
...
+        kv_length=kv_length,
+        kv_offset=kv_offset,
+        mask_function=mask_factory_function,
+        attention_mask=attention_mask,
+        allow_is_causal_skip=False,
+        allow_is_bidirectional_skip=allow_is_bidirectional_skip,
+        local_size=sliding_window,
+        dtype=dtype,
+        config=config,
+        use_vmap=use_vmap,
```
vasqu: Let's add the comments as per https://github.com/huggingface/transformers/blob/59b9cb06f4a19c6d6601cb3e96c018f170ff6808/src/transformers/masking_utils.py#L950-L952
and
https://github.com/huggingface/transformers/blob/59b9cb06f4a19c6d6601cb3e96c018f170ff6808/src/transformers/masking_utils.py#L1053

vasqu: It looks like we need to update https://github.com/huggingface/transformers/blob/59b9cb06f4a19c6d6601cb3e96c018f170ff6808/src/transformers/masking_utils.py#L256
Similarly to https://github.com/huggingface/transformers/blob/59b9cb06f4a19c6d6601cb3e96c018f170ff6808/src/transformers/masking_utils.py#L247
Since we now have to check for the window size there as well

src/transformers/models/modernbert/configuration_modernbert.py
```diff
@@ -80,8 +80,8 @@ class ModernBertConfig(PreTrainedConfig):
             Dictionary containing the configuration parameters for the RoPE embeddings. The dictionary should contain
             a value for `rope_theta` and optionally parameters used for scaling in case you want to use RoPE
             with longer `max_position_embeddings`.
-        local_attention (`int`, *optional*, defaults to 128):
-            The window size for local attention.
+        sliding_window (`int`, *optional*, defaults to 128):
+            The window size for sliding window attention.
```
vasqu: This will be breaking, let's add an `attribute_map` instead, e.g. https://github.com/huggingface/transformers/blob/59b9cb06f4a19c6d6601cb3e96c018f170ff6808/src/transformers/models/ernie4_5_moe/configuration_ernie4_5_moe.py#L115

src/transformers/models/modernbert/configuration_modernbert.py
```diff
@@ -99,8 +99,6 @@ class ModernBertConfig(PreTrainedConfig):
             Whether to use bias in the classifier.
         classifier_activation (`str`, *optional*, defaults to `"gelu"`):
             The activation function for the classifier.
-        deterministic_flash_attn (`bool`, *optional*, defaults to `False`):
-            Whether to use deterministic flash attention. If `False`, inference will be faster but not deterministic.
```
vasqu, line 102: Hmm, maybe we can set the env variable based on this instead? Ignore in modeling but adjust this based on if this attribute is set

tests/models/modernbert/test_modeling_modernbert.py
```diff
@@ -264,6 +267,19 @@ class ModernBertModelTest(ModelTesterMixin, PipelineTesterMixin, unittest.TestCa
...
+        # Flash Attention requires bf16/fp16, set default dtype to bfloat16 for all tests
+        if is_torch_available():
+            cls._original_dtype = torch.get_default_dtype()
+            torch.set_default_dtype(torch.bfloat16)
+
+    @classmethod
+    def tearDownClass(cls):
+        # Restore original default dtype
+        if is_torch_available() and hasattr(cls, "_original_dtype"):
+            torch.set_default_dtype(cls._original_dtype)
```
vasqu: Good idea but maybe it makes more sense to force eager/sdpa for all these tests instead? Ig this is due to defaulting to FA?

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -108,8 +100,8 @@ class ModernBertConfig(PreTrainedConfig):
             Dictionary containing the configuration parameters for the RoPE embeddings. The dictionary should contain
             a value for `rope_theta` and optionally parameters used for scaling in case you want to use RoPE
             with longer `max_position_embeddings`.
-        local_attention (`int`, *optional*, defaults to 128):
-            The window size for local attention.
+        sliding_window (`int`, *optional*, defaults to 128):
```
vasqu: Ah crap I commented on the configuration file, should ofc be done in modular

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -689,89 +342,84 @@ class ModernBertAttention(nn.Module):
...
-        self.all_head_size = self.head_dim * self.num_heads
-        self.Wqkv = nn.Linear(config.hidden_size, 3 * self.all_head_size, bias=config.attention_bias)
-        layer_type = config.layer_types[layer_id]
+        self.Wqkv = nn.Linear(config.hidden_size, 3 * config.hidden_size, bias=config.attention_bias)
 
-        if layer_id % config.global_attn_every_n_layers != 0:
-            self.local_attention = (config.local_attention // 2, config.local_attention // 2)
-            max_position_embeddings = config.local_attention
+        if layer_idx % config.global_attn_every_n_layers != 0:
+            self.sliding_window = config.sliding_window + 1
```
vasqu, line 416: Let's add a quick comment for the +1, might seem weird otherwise

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -689,89 +342,84 @@ class ModernBertAttention(nn.Module):
...
         self.Wo = nn.Linear(config.hidden_size, config.hidden_size, bias=config.attention_bias)
         self.out_drop = nn.Dropout(config.attention_dropout) if config.attention_dropout > 0.0 else nn.Identity()
 
     def forward(
         self,
         hidden_states: torch.Tensor,
-        position_embeddings: torch.Tensor | None = None,
-        output_attentions: bool | None = False,
+        attention_mask: torch.Tensor | None = None,
+        position_embeddings: tuple[torch.Tensor, torch.Tensor] | None = None,
```
vasqu:
```suggestion
        position_embeddings: tuple[torch.Tensor, torch.Tensor] | None = None,
        attention_mask: torch.Tensor | None = None,
```
nit: let's keep the order, this will be breaking enough already

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -781,32 +429,24 @@ def forward(
...
         attention_mask: torch.Tensor | None = None,
-        sliding_window_mask: torch.Tensor | None = None,
-        position_ids: torch.LongTensor | None = None,
-        cu_seqlens: torch.Tensor | None = None,
-        max_seqlen: int | None = None,
         position_embeddings: torch.Tensor | None = None,
-        output_attentions: bool | None = False,
-    ) -> torch.Tensor:
-        attn_outputs = self.attn(
+        **kwargs,
```
vasqu: Let's type the kwargs e.g. https://github.com/huggingface/transformers/blob/59b9cb06f4a19c6d6601cb3e96c018f170ff6808/src/transformers/models/gemma3/modeling_gemma3.py#L531
(was the quickest I could find)

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -781,32 +429,24 @@ def forward(
...
+    ) -> tuple[torch.Tensor, torch.Tensor | None]:
+        attn_output, attn_weights = self.attn(
```
vasqu:
```suggestion
        attn_output, _ = self.attn(
```
I think we can ignore this and directly record from the attention class, wdyt?

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -818,6 +458,12 @@ class ModernBertPreTrainedModel(PreTrainedModel):
     _supports_flash_attn = True
     _supports_sdpa = True
     _supports_flex_attn = False
```
vasqu: Not super important but imo this should work now when we use our mask creation fn

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -970,175 +611,75 @@ def get_input_embeddings(self):
...
-            attention_mask, sliding_window_mask = self._update_attention_mask(
-                attention_mask, output_attentions=output_attentions
-            )
+        if position_ids is None:
+            position_ids = torch.arange(seq_len, device=device).unsqueeze(0)
 
         hidden_states = self.embeddings(input_ids=input_ids, inputs_embeds=inputs_embeds)
+
+        if not isinstance(attention_mask_mapping := attention_mask, dict):
+            attention_mask_mapping = self._update_attention_mask(hidden_states, attention_mask)
```
vasqu: Let's remove the separate fn `_update_attention_mask`. I think it's clearer to just directly do this right there

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -1189,75 +730,22 @@ def forward(
...
         inputs_embeds: torch.Tensor | None = None,
         labels: torch.Tensor | None = None,
-        indices: torch.Tensor | None = None,
-        cu_seqlens: torch.Tensor | None = None,
-        max_seqlen: int | None = None,
-        batch_size: int | None = None,
-        seq_len: int | None = None,
-        output_attentions: bool | None = None,
-        output_hidden_states: bool | None = None,
         return_dict: bool | None = None,
```
vasqu: We can remove `return_dict` for every wrapper class around the base model, using the `@can_return_tuple` decorator. 

Note that 
- a) We can make this with `return_dict=True` this way, e.g. https://github.com/huggingface/transformers/blob/59b9cb06f4a19c6d6601cb3e96c018f170ff6808/src/transformers/models/bert/modeling_bert.py#L1406 (allows the return type to be always consistent)
- b) Remove the logic around that kwarg

Just as general pointer, bert is a good e.g. https://github.com/huggingface/transformers/blob/59b9cb06f4a19c6d6601cb3e96c018f170ff6808/src/transformers/models/bert/modeling_bert.py#L942

vasqu (commented):
We are getting there :D another round, these are less critical IMO - the general direction is already pretty good

YangKai0616:
I’ve modified the code according to your suggestions. I’ll find some time tomorrow to test the related unit tests. If everything looks good, we can continue with the review. Thank you for your patience!

src/transformers/models/modernbert/modular_modernbert.py
YangKai0616: Yes, you're right !

src/transformers/models/modernbert/modular_modernbert.py
YangKai0616: It seems that the original implementation didn't have a consistent order to begin with.
But keeping it as it is has always been the safe choice👍.

tests/models/modernbert/test_modeling_modernbert.py
YangKai0616: Yes, ModernBERT uses Flash Attention by default. I believe we should proceed with the testing under this premise.
What are your thoughts on this?

vasqu:
@YangKai0616 just ping me again when you think it's ready for another review :saluting_face:

YangKai0616:
The CI error seems unrelated to this PR. I was able to test it successfully on my local CUDA and XPU. @vasqu let's review it again, thank you !

src/transformers/models/modernbert/configuration_modernbert.py
vasqu: We need to revert that name back to `local_attention`, otherwise we break BC with already uploaded files. The attribute should allow us to do something along `config.sliding_window  # points to local_attention`

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -221,11 +212,12 @@ def __init__(
...
         self.classifier_bias = classifier_bias
         self.classifier_activation = classifier_activation
-        self.deterministic_flash_attn = deterministic_flash_attn
         self.sparse_prediction = sparse_prediction
         self.sparse_pred_ignore_index = sparse_pred_ignore_index
         self.reference_compile = reference_compile
-        self.repad_logits_with_grad = repad_logits_with_grad
+
+        if deterministic_flash_attn:
+            os.environ["FLASH_ATTENTION_DETERMINISTIC"] = "1"
```
vasqu: Hmm, imo we should still have this influence during some script, e.g. if the user sets it again at some point `config.deterministic_flash_attn = False`. So we should keep it as an attribute.

Imo, there are two options:
- As a property and on a custom setter we influence the env variable
- We do it as previously (attr in the attention module) and pass `deterministic=...` to the attn interface as we have this kwarg https://github.com/huggingface/transformers/blob/159219a6601ab87787b75cdc270b3b2438a9f0cf/src/transformers/modeling_flash_attention_utils.py#L580

Sorry I was the first to say to remove it

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -689,89 +351,87 @@ class ModernBertAttention(nn.Module):
...
             )
 
         self.attention_dropout = config.attention_dropout
-        self.deterministic_flash_attn = config.deterministic_flash_attn
-        self.num_heads = config.num_attention_heads
         self.head_dim = config.hidden_size // config.num_attention_heads
-        self.all_head_size = self.head_dim * self.num_heads
-        self.Wqkv = nn.Linear(config.hidden_size, 3 * self.all_head_size, bias=config.attention_bias)
-        layer_type = config.layer_types[layer_id]
+        self.Wqkv = nn.Linear(config.hidden_size, 3 * config.hidden_size, bias=config.attention_bias)
```
vasqu:
```suggestion
        self.Wqkv = nn.Linear(config.hidden_size, 3 * self.head_dim * config.num_attention_heads, bias=config.attention_bias)
```
`self.all_head_size = self.head_dim * self.num_heads` so we have a different dimension just in case

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -689,89 +351,87 @@ class ModernBertAttention(nn.Module):
...
 
-        if layer_id % config.global_attn_every_n_layers != 0:
-            self.local_attention = (config.local_attention // 2, config.local_attention // 2)
-            max_position_embeddings = config.local_attention
+        if layer_idx % config.global_attn_every_n_layers != 0:
+            # +1 is needed because flash attention sets inclusive boundaries (see modeling_flash_attention_utils.py)
+            # i.e., window_size=(w-1, w-1) attends to 2*w-1 tokens total. To get a total window of 2*sliding_window+1,
```
vasqu:
```suggestion
            # i.e., window_size=(w-1, w-1) attends to 2*(w-1)+1=2*w-1 tokens total. To get a total window of 2*sliding_window+1,
```
Let's add the intermediate step for clarification?

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -970,175 +623,64 @@ def get_input_embeddings(self):
...
-            else:
-                batch_size, seq_len = input_ids.shape[:2]
+        if inputs_embeds is not None:
+            batch_size, seq_len = inputs_embeds.shape[:2]
+        else:
+            batch_size, seq_len = input_ids.shape[:2]
         device = input_ids.device if input_ids is not None else inputs_embeds.device
 
         if attention_mask is None:
             attention_mask = torch.ones((batch_size, seq_len), device=device, dtype=torch.bool)
```
vasqu:
```
if attention_mask is None:    
    attention_mask = torch.ones((batch_size, seq_len), device=device, dtype=torch.bool)
```
We should not need to create an attention mask at all :thinking: The mask creation should handle this correctly

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -970,175 +623,64 @@ def get_input_embeddings(self):
...
+        if not isinstance(attention_mask_mapping := attention_mask, dict):
+            mask_kwargs = {
+                "config": self.config,
+                "input_embeds": hidden_states,
+                "attention_mask": attention_mask,
+            }
+            attention_mask_mapping = {
+                "full_attention": create_bidirectional_mask(**mask_kwargs),
+                "sliding_attention": create_bidirectional_sliding_window_mask(**mask_kwargs),
+            }
```
vasqu, line 645: Nice!

test_log.txt
vasqu, line 1: Oops :D

tests/models/modernbert/test_modeling_modernbert.py
vasqu: Our CIs are running mostly on images that do not have flash attention installed (because the tests create a lot of noise, very flaky)

It would be better to have the same behavior everywhere by forcing to load the model with `attn_implementation="sdpa"`. This way we are not dependent on FA and we don't need these workarounds

tests/models/modernbert/test_modeling_modernbert.py
```diff
@@ -507,7 +644,7 @@ def flash_attn_inference_equivalence(
                 f"Model architecture does not support {attn_implementation}, or setting its attention dynamically"
             )
 
-    @require_torch_gpu  # modernbert contains triton code which cannot run on CPU, so we only test on GPU
+    @require_torch_accelerator  # modernbert contains triton code which cannot run on CPU, so we only test on GPU
```
vasqu: Adding the comment here but meant it moreso to ask whether you double checked the integration tests whether they work as previously and if the fa forward is similar.

tests/test_modeling_common.py
```diff
@@ -2720,7 +2720,7 @@ def check_device_map_is_respected(self, model, device_map):
     @require_accelerate
     @mark.accelerate_tests
     @require_torch_accelerator
-    def test_disk_offload_bin(self):
+    def test_disk_offload_bin(self, rtol=1e-5, atol=1e-5):
```
vasqu: Yea like I said previously let's force sdpa attention instead (on the tests)

tests/models/pe_video/test_modeling_pe_video.py
```diff
@@ -355,6 +355,12 @@ def test_batching_equivalence(self):
     def test_can_init_all_missing_weights(self):
         pass
 
+    @unittest.skip(
+        reason="TimmWrapperForImageClassification does not support an attention implementation through torch.nn.functional.scaled_dot_product_attention yet"
+    )
+    def test_can_set_attention_dynamically_composite_model(self):
+        pass
```
vasqu: Can we move these to a different PR?

vasqu (commented):
Another round, looks very good! I have a feeling that we also need to update the docs to demonstrate how to do padding-free inference/training via the data collator, e.g. similar to llama https://gist.github.com/vasqu/c264d14268fbc7f1eb725e09dfb87331. This should result to similar results to the previously manual padding/unpadding approach they did

My last comments surround 
- testing in sdpa (even if FA is the default), it causes too much changes and the values will be more flaky in bf16 than fp32
- integration tests working as expected

vasqu:
Ready for review @YangKai0616? Will take a look tomorrow then

YangKai0616:
> Ready for review @YangKai0616? Will take a look tomorrow then

Current version:
1. Can be tested via **test_modernbert.py**
2. Short sequences perform the same as before, but long sequences involving **sliding window** have some slight errors at the moment. I'm currently working on fixing this.

tests/models/pe_video/test_modeling_pe_video.py
YangKai0616: Sure, but this will cause the current CI to fail.👀

YangKai0616:
Hey @vasqu , we can review it again!
In my local tests, the current version of **FA2/sdpa/eager** can achieve results that are almost identical to those of the main branch.
This change mainly modifies the data format of **FA2 rotary (.float())** and fixes the **sliding_window** calculation to match the implementation in the main branch.

Hmm... it seems there are some issues with the CI. I can't reproduce these failing cases locally(CUDA and XPU).

src/transformers/integrations/hub_kernels.py
```diff
@@ -210,7 +210,12 @@ def use_kernel_func_from_hub(func_name: str):
                 Mode.INFERENCE: FuncRepository(
                     repo_id="kernels-community/rotary", func_name="apply_rotary_transformers"
                 )
-            }
+            },
+            "cuda": {
+                Mode.INFERENCE: FuncRepository(
+                    repo_id="kernels-community/rotary", func_name="apply_rotary_transformers"
+                )
+            },
```
vasqu, line 218: So we need the kernels to have similar outputs? Makes sense iirc those are the same FA kernels

src/transformers/masking_utils.py
```diff
@@ -117,6 +117,26 @@ def sliding_window_causal_mask_function(sliding_window: int) -> Callable:
...
 
+def sliding_window_bidirectional_overlay(sliding_window: int) -> Callable:
+    """
+    This is an overlay depicting a bidirectional sliding window pattern.
+    """
+
+    def inner_mask(batch_idx: int, head_idx: int, q_idx: int, kv_idx: int) -> bool:
+        """A token can attend to any other token if their absolute distance is within
+        the (inclusive) sliding window size (distance <= sliding_window)."""
+        return abs(q_idx - kv_idx) <= sliding_window
```
vasqu, line 128: I remember it being `<`, did I lead you to the wrong formula :cry: my bad sorry

src/transformers/models/modernbert/configuration_modernbert.py
```diff
@@ -258,5 +253,15 @@ def to_dict(self):
...
 
+    @property
+    def sliding_window(self):
+        """Half-window size: `local_attention` is the total window, so we divide by 2."""
+        return self.local_attention // 2
+
+    @sliding_window.setter
+    def sliding_window(self, value):
+        """Set sliding_window by updating local_attention to 2 * value."""
+        self.local_attention = value * 2
```
vasqu, line 282: Had to double check so for `local_attention = 128`, we attend to `64` tokens to the left and right + the token in the middle.

I thought about adjusting the init instead but that would be too breaking IMO. This property solution is a good solution.

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -704,77 +373,91 @@ def __init__(self, config: ModernBertConfig, layer_id: int | None = None):
...
 
-        if layer_id % config.global_attn_every_n_layers != 0:
-            self.local_attention = (config.local_attention // 2, config.local_attention // 2)
-            max_position_embeddings = config.local_attention
+        if layer_idx % config.global_attn_every_n_layers != 0:
+            # config.sliding_window = local_attention // 2 (half-window size, e.g. 64 for local_attention=128)
+            # +1 is needed because flash attention sets inclusive boundaries (see modeling_flash_attention_utils.py)
+            # i.e., window_size=(w-1, w-1) attends to 2*(w-1)+1 = 2*w-1 tokens total.
+            # To get a total window of 2*sliding_window+1, we need to pass sliding_window+1 here
+            # so it becomes window_size=(sliding_window, sliding_window).
```
vasqu: I feel like we are doubling the explanation a bit

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -704,77 +373,91 @@ def __init__(self, config: ModernBertConfig, layer_id: int | None = None):
...
-            qkv = qkv.view(-1, 3, self.num_heads, self.head_dim)
+            original_dtype = query_states.dtype
+            query_states = query_states.float()
+            key_states = key_states.float()
+            query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin, unsqueeze_dim=1)
+            query_states = query_states.to(original_dtype)
+            key_states = key_states.to(original_dtype)
         else:
-            qkv = qkv.view(bs, -1, 3, self.num_heads, self.head_dim)
+            query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin, unsqueeze_dim=1)
```
vasqu: I don't like this difference between FA and eager/sdpa/flex. Rope shouldn't differ, let's remove this difference (seems more like an initial implementation mistake)

This float conversion belongs to `apply_rotary_pos_emb` https://github.com/huggingface/transformers/blob/80f6f6d6a07a6c744f8e0d28cce8ee75d72f942b/src/transformers/models/modernbert/modeling_modernbert.py#L194
You can overwrite it by importing from a different model that does it (modular will import the correct function and overrites it, might need to add something for ruff to ignore the unused import tho), e.g. like here https://github.com/huggingface/transformers/blob/80f6f6d6a07a6c744f8e0d28cce8ee75d72f942b/src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py#L183
(this should fit, lmk if not)

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -704,77 +373,91 @@ def __init__(self, config: ModernBertConfig, layer_id: int | None = None):
...
+            deterministic=self.deterministic_flash_attn,
             **kwargs,
         )
-        hidden_states = attn_outputs[0]
-        hidden_states = self.out_drop(self.Wo(hidden_states))
 
-        return (hidden_states,) + attn_outputs[1:]  # add attentions if outputted
+        attn_output = attn_output.reshape(*input_shape, -1).contiguous()
+        attn_output = self.out_drop(self.Wo(attn_output))
+        return attn_output, attn_weights
```
vasqu, line 464: This is so much clearn :heart:

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -1466,60 +915,30 @@ def __init__(self, config: ModernBertConfig):
...
-            Cumulative sequence lengths of the input sequences. Used to index the unpadded tensors.
-        max_seqlen (`int`, *optional*):
-            Maximum sequence length in the batch excluding padding tokens. Used to unpad input_ids and pad output tensors.
-        batch_size (`int`, *optional*):
-            Batch size of the input sequences. Used to pad the output tensors.
-        seq_len (`int`, *optional*):
-            Sequence length of the input sequences including padding tokens. Used to pad the output tensors.
         """
-        return_dict = return_dict if return_dict is not None else self.config.use_return_dict
         self._maybe_set_compile()
```
vasqu: Do we still even need this? We should support native `torch.compile` now that we do not so much custom stuff anymore

utils/check_config_attributes.py
```diff
@@ -77,6 +77,7 @@
     "GPTNeoXConfig": ["rotary_emb_base"],
     "ShieldGemma2Config": ["mm_tokens_per_image", "vision_config"],
     "Llama4VisionConfig": ["multi_modal_projector_bias", "norm_eps"],
+    "ModernBertConfig": ["local_attention"],
```
vasqu: I think this changed on main? Might need to double-check after merge

tests/models/modernbert/test_modeling_modernbert.py
```diff
@@ -507,6 +516,288 @@ def flash_attn_inference_equivalence(
                 f"Model architecture does not support {attn_implementation}, or setting its attention dynamically"
             )
 
+    # Override tests that use from_pretrained to ensure SDPA attention is used instead of FlashAttention
+    def test_save_load(self):
```
vasqu: Just to double check, would all these work in either case (without any changes)? Meaning if I have flash attn installed (defaulting to fa) or not installed (defaulting to sdpa), both work out of the box? In that case, we can IMO remove these overrides - the core is that we default to SDPA in most cases so that we minimize the friction between different environments

Meaning all tests below as well

vasqu (commented):
Getting closer to the finish line, this is a huge refactor so being a bit more strict

Can we add something to the modernbert docs to show how to do paddingfree inference/training. I.e. some tips on how to utilize the model optimally. I have a [gist for llama](https://gist.github.com/vasqu/c264d14268fbc7f1eb725e09dfb87331) but should be easy to adapt to modernbert

vasqu:
> Hmm... it seems there are some issues with the CI. I can't reproduce these failing cases locally(CUDA and XPU).

Yea our CI is pretty flaky these days, let's focus on modernbert only for now. 

> This change mainly modifies the data format of FA2 rotary (.float()) and fixes the sliding_window calculation to match the implementation in the main branch.

Do the kernels give exact matches / is there a big diff between kernel vs non-kernel RoPE? We might need to adjust our integration test(s)

src/transformers/integrations/hub_kernels.py
YangKai0616: According to my tests, in the original implementation, the `apply_rotary` from `flash_attn.ops.triton.rotary` [here](https://github.com/huggingface/transformers/blob/24807bfcf4a21286fa2a7e728f381ddaaca7bbc7/src/transformers/models/modernbert/modular_modernbert.py#L49C5-L49C58) and the `apply_rotary_transformers` in the `kernels` produce identical results. However, the `apply_rotary_pos_emb` from `..gemma3.modeling_gemma3` [here](https://github.com/huggingface/transformers/blob/24807bfcf4a21286fa2a7e728f381ddaaca7bbc7/src/transformers/models/modernbert/modular_modernbert.py#L43) only gives the same result as the previous two when `.float()` is called on it.

This is also why I later made a distinction between the RoPE handling for **FA2** and for **sdpa/eager**. That said, if the goal weren’t to forcibly align with the main branch, we could indeed just add `.float()` to the sdpa/eager RoPE implementation as well.

src/transformers/masking_utils.py
YangKai0616: This boundary error is so small that it can only be detected during testing.😀

src/transformers/models/modernbert/modular_modernbert.py
YangKai0616: Deleted.

src/transformers/models/modernbert/modular_modernbert.py
YangKai0616: I struggled with this part for a while, but it seems there are no matching `apply_rotary_pos_emb` and `rotate_half` implementations.
The `rotate_half` in `modeling_ernie4_5_vl_moe.py` (and similar models) is different from what `modernbert` needs, so I rewrote a version of RoPE here.

src/transformers/models/modernbert/modular_modernbert.py
YangKai0616: Removed the relevant code and tested consistency on CUDA.

tests/models/modernbert/test_modeling_modernbert.py
YangKai0616: No, if **FA2** is installed, these tests will fail with error `RuntimeError: FlashAttention only support fp16 and bf16 data type`.
If **FA2** is not installed, it will go through the default **sdpa** path and the tests will pass."

YangKai0616:
> Do the kernels give exact matches / is there a big diff between kernel vs non-kernel RoPE? We might need to adjust our integration test(s)

The kernels are an exact match.

For the torch implementation version, I tested it with the configuration set to `batch_size, num_heads, seq_len, head_dim = 2, 8, 128, 64 dtype = torch.bfloat16`, and the maximum difference is on the order of `e-2`, while the average difference is on the order of `e-3`. After adding `.float()`, the results become completely equal.

YangKai0616:
> Can we add something to the modernbert docs to show how to do paddingfree inference/training. I.e. some tips on how to utilize the model optimally. I have a [gist for llama](https://gist.github.com/vasqu/c264d14268fbc7f1eb725e09dfb87331) but should be easy to adapt to modernbert

Done.

docs/source/en/model_doc/modernbert.md
```diff
@@ -89,6 +89,49 @@ echo -e "Plants create [MASK] through a process known as photosynthesis." | tran
 </hfoption>
 </hfoptions>
 
+## Padding-free inference and training
```
vasqu, line 92: Thanks, this is really nice to have and will align with the previous behavior of the implementation on main

src/transformers/integrations/hub_kernels.py
vasqu: It makes more sense to align the RoPE implementations as the attention implementation should introduce the least amount of difference as possible. Having different RoPE (float vs non-float) alone introduces numerical inaccuracies. Also, knowing that FA2 is the preferred way to use modernbert it speaks that it was an implementation mistake to have the wrong `apply_rotary_pos_emb`

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -558,130 +345,34 @@ def compute_default_rope_parameters(
...
-            attn_mask=attention_mask,
-        )
-        .transpose(1, 2)
-        .contiguous()
-    )
-    attn_output = attn_output.view(bs, -1, dim)
-    return (attn_output,)
-
+@use_kernel_func_from_hub("rotary_pos_emb")
+def apply_rotary_pos_emb(q, k, cos, sin, unsqueeze_dim=1):
```
vasqu, line 359: I think we can just import from another model and modular will resolve it, e.g. `from ..ernie4_5_vl_moe.modeling_ernie4_5_vl_moe import apply_rotary_pos_emb`

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -973,175 +605,59 @@ def get_input_embeddings(self):
...
         if (input_ids is None) ^ (inputs_embeds is not None):
             raise ValueError("You must specify exactly one of input_ids or inputs_embeds")
 
-        all_hidden_states = () if output_hidden_states else None
-        all_self_attentions = () if output_attentions else None
-
-        self._maybe_set_compile()
-
         if input_ids is not None:
             self.warn_if_padding_and_no_attention_mask(input_ids, attention_mask)
```
vasqu: Can be removed as well tbh (little value, expensive to run)

src/transformers/models/modernbert/modular_modernbert.py
```diff
@@ -973,175 +605,59 @@ def get_input_embeddings(self):
...
-                    )
-            if position_ids is None:
-                position_ids = indices.unsqueeze(0)
+        if inputs_embeds is not None:
+            seq_len = inputs_embeds.shape[1]
         else:
-            if position_ids is None:
-                position_ids = torch.arange(seq_len, device=device).unsqueeze(0)
+            seq_len = input_ids.shape[1]
+        device = input_ids.device if input_ids is not None else inputs_embeds.device
```
vasqu: Nit: if the device is already a ternary, we can do the same for seq_len then

src/transformers/models/modernbert_decoder/modeling_modernbert_decoder.py
vasqu, line 1: Can you check repo consistency: https://app.circleci.com/pipelines/github/huggingface/transformers/160475/workflows/798c6c0c-8917-4f67-9e04-3c809ba287ef/jobs/2111507

We changed the command to `make fix-repo` btw (copies doesn't exist anymore)

tests/models/modernbert/test_modeling_modernbert.py
vasqu: Ah that's a shame, can we clarify a bit more with the comment then? I.e. compatibility with FA (dtype) and SDPA as the default attn implementation goes for FA

tests/models/modernbert/test_modeling_modernbert.py
vasqu, line 1: I guess the integration test needs an update then as well? I will change the numbers accordingly then when we resolved all comments

vasqu (approved):
Last few nits left, pre-approving because the core stand now and it's rather small changes that are needed

vasqu:
> The kernels are an exact match. [...] After adding .float(), the results become completely equal.

So we can achieve parity either way now (seeing we use the float version), be it with kernels or the torch version? If the kernels are essential, then we should put it to the docs

tests/models/pe_video/test_modeling_pe_video.py
vasqu: Hmm, can you check with main if this still fails. Mb if we still do need to skip for CI

src/transformers/models/modernbert/modular_modernbert.py
YangKai0616: The `rotate_half` grouping methods of these two models([modernbert](https://github.com/huggingface/transformers/blob/5c773b8a84677192d4a52edc1e2c8823f9c1dcea/src/transformers/models/modernbert/modeling_modernbert.py#L329) and [ernie4_5_vl_moe](https://github.com/huggingface/transformers/blob/5c773b8a84677192d4a52edc1e2c8823f9c1dcea/src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py#L176)) are different, so it seems they can’t be directly swapped while still matching the existing modernbert weights? Also, after searching through all the implementations in the current models, I couldn’t find any RoPE version that can be directly imported and used in the way we expected.

What do you think?👀

YangKai0616:
> So we can achieve parity either way now (seeing we use the float version), be it with kernels or the torch version? If the kernels are essential, then we should put it to the docs

Yes, for correctness, kernels are not essential — the torch version with `.float()` works equally well.

src/transformers/models/modernbert/modular_modernbert.py
YangKai0616: Done.

src/transformers/models/modernbert/modular_modernbert.py
YangKai0616: Done.

tests/models/modernbert/test_modeling_modernbert.py
YangKai0616: The tests in `ModernBertModelIntegrationTest` all pass because the input data for RoPE are all `float32`.

tests/models/modernbert/test_modeling_modernbert.py
YangKai0616: Done.

src/transformers/models/modernbert_decoder/modeling_modernbert_decoder.py
```diff
@@ -450,6 +443,12 @@ def init_weight(module: nn.Module, std: float):
                 init.copy_(getattr(module, f"{layer_type}_inv_freq"), curr_inv_freq)
                 init.copy_(getattr(module, f"{layer_type}_original_inv_freq"), curr_inv_freq)
 
+    def _maybe_set_compile(self):
+        raise AttributeError("No need to inherit!")
+
+    def resize_token_embeddings(self, *args, **kwargs):
+        raise AttributeError("No need to inherit!")
+
```
YangKai0616: Here update will cause test cases `tests/models/modernbert_decoder/test_modeling_modernbert_decoder.py::ModernBertDecoderModelTest::test_resize_embeddings_untied` and `tests/models/modernbert_decoder/test_modeling_modernbert_decoder.py::ModernBertDecoderModelTest::test_resize_tokens_embeddings` to fail, please take note.

Not sure why the person who modified the modular part hasn’t updated this section.

YangKai0616:
I’ve run all the failed CI cases, and the following three are confirmed to fail:
```
tests/models/modernbert_decoder/test_modeling_modernbert_decoder.py::ModernBertDecoderModelTest::test_resize_embeddings_untied
tests/models/modernbert_decoder/test_modeling_modernbert_decoder.py::ModernBertDecoderModelTest::test_resize_tokens_embeddings
tests/models/pe_video/test_modeling_pe_video.py::PeVideoModelTest::test_can_set_attention_dynamically_composite_model
```
Among them, `test_resize_embeddings_untied` and `test_resize_tokens_embeddings` are failing because the changes from before in the modular file weren’t updated in modeling file — these are unrelated to this PR.
`test_can_set_attention_dynamically_composite_model` was mentioned before. We need to skip this example.

src/transformers/models/modernbert/modular_modernbert.py
vasqu: I just noticed that it was named `rotate_half_text` - yea that makes it impossible to properly inherit. We need to keep it as you have done, thanks for checking nonetheless

tests/models/modernbert/test_modeling_modernbert.py
vasqu: Will check with our CI just in case

vasqu: Can we add an fa2 integration test? It makes sense to have this as regression test, especially since we change a lot in this PR

Edit: a8874a0

src/transformers/models/modernbert_decoder/modeling_modernbert_decoder.py
vasqu: Yes, let's fix it here as well - while it is unrelated it is related to modernbert. Probably a proper mistake along some refactors Edit: 1d2101f

vasqu:
If `tests/models/pe_video/test_modeling_pe_video.py::PeVideoModelTest::test_can_set_attention_dynamically_composite_model` is blocking us, I'm pro skipping for now - but let's check with our main (merging in a second + running slow tests to check if they work as expected)

Yea we need to take a look: https://app.circleci.com/pipelines/github/huggingface/transformers/160664/workflows/f3cbb6f6-9b01-470b-90e7-1375771e76f6/jobs/2114314 - do you what changed this behavior? It passes on main

vasqu:
run-slow: modernbert, modernbert_decoder

vasqu:
run-slow: modernbert, modernbert_decoder

vasqu:
Do you have a quick script for testing against main re https://github.com/huggingface/transformers/pull/43030#issuecomment-3772099830 @YangKai0616 

I would verify correctness myself again just to be sure. I've added a few smaller changes, let me know if these work for you (mainly tests). Let's also keep https://github.com/huggingface/transformers/pull/43030#issuecomment-3779348609 in mind

YangKai0616:
> Do you have a quick script for testing against main re [#43030 (comment)](https://github.com/huggingface/transformers/pull/43030#issuecomment-3772099830) @YangKai0616
I would verify correctness myself again just to be sure. 
> 
> I've added a few smaller changes, let me know if these work for you (mainly tests). 
> 
> Let's also keep [#43030 (comment)](https://github.com/huggingface/transformers/pull/43030#issuecomment-3779348609) in mind
1.
```
import torch
from kernels import get_kernel


# ==================== PyTorch ====================
# from transformers.models.gemma3.modeling_gemma3 import apply_rotary_pos_emb

# ==================== PyTorch + .float() ====================
from transformers.models.modernbert.modeling_modernbert import apply_rotary_pos_emb

# ==================== Kernel ====================
rotary_kernel = get_kernel("kernels-community/rotary")
apply_rotary_pos_emb_kernel = rotary_kernel.apply_rotary_transformers

# ================================================

def test_with_proper_rope_format():
    batch_size, num_heads, seq_len, head_dim = 2, 8, 128, 64
    half_dim = head_dim // 2
    device = "cuda"
    dtype = torch.bfloat16
    
    q = torch.randn(batch_size, num_heads, seq_len, head_dim, device=device, dtype=dtype)
    k = torch.randn(batch_size, num_heads, seq_len, head_dim, device=device, dtype=dtype)
    
    cos_half = torch.randn(batch_size, seq_len, half_dim, device=device, dtype=dtype)
    sin_half = torch.randn(batch_size, seq_len, half_dim, device=device, dtype=dtype)
    
    cos_full = torch.cat([cos_half, cos_half], dim=-1)
    sin_full = torch.cat([sin_half, sin_half], dim=-1)
    
    print(f"q shape: {q.shape}")
    print(f"cos_half shape: {cos_half.shape}")
    print(f"cos_full shape: {cos_full.shape}")
    
    q_pytorch, k_pytorch = apply_rotary_pos_emb(
        q.clone(), k.clone(), cos_full.clone(), sin_full.clone(), unsqueeze_dim=1
    )

    q_kernel, k_kernel = apply_rotary_pos_emb_kernel(
        q.clone(), k.clone(), cos_full.clone(), sin_full.clone(), unsqueeze_dim=1
    )

    q_close = torch.allclose(q_pytorch, q_kernel, atol=1e-4, rtol=1e-4)
    k_close = torch.allclose(k_pytorch, k_kernel, atol=1e-4, rtol=1e-4)
    
    q_max_diff = (q_pytorch - q_kernel).abs().max().item()
    k_max_diff = (k_pytorch - k_kernel).abs().max().item()
    q_mean_diff = (q_pytorch - q_kernel).abs().mean().item()
    k_mean_diff = (k_pytorch - k_kernel).abs().mean().item()
    
    print(f"\nQ output match: {q_close}, max difference: {q_max_diff:.2e}")
    print(f"K output match: {k_close}, max difference: {k_max_diff:.2e}")
    print(f"Q mean difference: {q_mean_diff:.2e}")
    print(f"K mean difference: {k_mean_diff:.2e}")
    print(f"Test result: {'Passed ✓' if q_close and k_close else 'Failed ✗'}")

if __name__ == "__main__":
    print("Rotary Position Embedding consistency test")
    print("=" * 60)
    
    if not torch.cuda.is_available():
        print("CUDA not available, exiting")
        exit(1)
    
    print(f"device: {torch.cuda.get_device_name(0)}")
    
    test_with_proper_rope_format()

```

2. `test_inference_masked_lm_flash_attention_2` shows an error on the order of `e-2` on XPU (PVC). I added a ground truth specifically for XPU.

3. Because `_can_set_attn_implementation()` [here](https://github.com/huggingface/transformers/blob/de306e8e14672dd8392b4bd344054a6a18de8613/tests/test_modeling_common.py#L4379) returned `False` before the refactoring, the main branch would skip this test and report `"Parts of this model cannot set attention dynamically"`.
After the refactoring, it now returns `True`, so the test continues to run and causes `TimmWrapper` to error out.
Therefore, this test needs to be explicitly skipped.

Commit https://github.com/huggingface/transformers/pull/43030/commits/587cb4f06bf3c96266047e7fe40bedafd98eb8e0 submitted.

vasqu:
Thanks a lot! I've also verified with the script you gave. Also sanity checked with 
```python
from transformers import AutoModelForMaskedLM, AutoTokenizer


tokenizer = AutoTokenizer.from_pretrained(
    "answerdotai/ModernBERT-base",
)
model = AutoModelForMaskedLM.from_pretrained(
    "answerdotai/ModernBERT-base",
    dtype=torch.float16,
    device_map="auto",
    reference_compile=False,  # only relevant for the old model
)
inputs = tokenizer("Plants create [MASK] through a process known as photosynthesis.", return_tensors="pt").to(model.device)

with torch.no_grad():
    outputs = model(**inputs)
    predictions = outputs.logits

masked_index = torch.where(inputs['input_ids'] == tokenizer.mask_token_id)[1]
predicted_token_id = predictions[0, masked_index].argmax(dim=-1)
predicted_token = tokenizer.decode(predicted_token_id)

print(f"The predicted token is: {predicted_token}")
```
And they gave nearly identical matches (atol/rtol of 1e-8 matched)

> test_inference_masked_lm_flash_attention_2 shows an error on the order of e-2 on XPU (PVC). I added a ground truth specifically for XPU.

Perfect

> Because _can_set_attn_implementation() [here](https://github.com/huggingface/transformers/blob/de306e8e14672dd8392b4bd344054a6a18de8613/tests/test_modeling_common.py#L4379) returned False before the refactoring, the main branch would skip this test and report "Parts of this model cannot set attention dynamically".
After the refactoring, it now returns True, so the test continues to run and causes TimmWrapper to error out.
Therefore, this test needs to be explicitly skipped.

Gotcha, thanks for clarifying

vasqu (approved):
Thanks a lot for iterating and wrapping up the last details. I still want to have a second pair of eyes here but on my side everything looks good! cc @ArthurZucker @tomaarsen

tomaarsen:
Hello!

This is an extremely impressive PR, wow. The structure of the modeling/modular files look very strong as a result of all of this. I ran some downstream tests using Sentence Transformers with https://huggingface.co/nomic-ai/modernbert-embed-base, and there are some very slight differences in scores:

```
sdpa <v5:
{'NanoBEIR_mean_cosine_ndcg@10': 0.5147433277390966,
 'NanoMSMARCO_cosine_ndcg@10': 0.5867981583748159,
 'NanoNFCorpus_cosine_ndcg@10': 0.329223326613997,
 'NanoNQ_cosine_ndcg@10': 0.6282084982284769}
Evaluation took 27.23 seconds

eager <v5:
{'NanoBEIR_mean_cosine_ndcg@10': 0.5123706576330123,
 'NanoMSMARCO_cosine_ndcg@10': 0.5770369329466852,
 'NanoNFCorpus_cosine_ndcg@10': 0.32729203600264306,
 'NanoNQ_cosine_ndcg@10': 0.6327830039497084}
Evaluation took 44.04 seconds

fa2 <v5: I don't have flash_attn installed in the right venv at the moment; goes to show how nice having kernels support is

sdpa pr-43030:
{'NanoBEIR_mean_cosine_ndcg@10': 0.5110831146710934,
 'NanoMSMARCO_cosine_ndcg@10': 0.5828975140199241,
 'NanoNFCorpus_cosine_ndcg@10': 0.32718235725979056,
 'NanoNQ_cosine_ndcg@10': 0.6231694727335656}
Evaluation took 32.49 seconds

eager pr-43030:
{'NanoBEIR_mean_cosine_ndcg@10': 0.5099581159107758,
 'NanoMSMARCO_cosine_ndcg@10': 0.5843932028670678,
 'NanoNFCorpus_cosine_ndcg@10': 0.32340108499481846,
 'NanoNQ_cosine_ndcg@10': 0.622080059870441}
Evaluation took 54.74 seconds

fa2 pr-43030:
{'NanoBEIR_mean_cosine_ndcg@10': 0.5119965071683512,
 'NanoMSMARCO_cosine_ndcg@10': 0.581536709004633,
 'NanoNFCorpus_cosine_ndcg@10': 0.3271691052363954,
 'NanoNQ_cosine_ndcg@10': 0.6272837072640253}
Evaluation took 25.31 seconds
```

When I implemented this architecture, these forms of slight differences also existed between e.g. eager/sdpa/fa2, and it's extremely hard to narrow down the source (and from experience: often enough even with the source known, it's not possible to do anything about it). 

The old implementation automatically 'unpadded' and 'repadded' during inference. I take it this variant doesn't do that? You write about a `DataCollatorWithFlattening` etc., is this required now for this speedup? Is there any way to know whether any given `transformers` models supports this class? It would be valuable for me to know for Sentence Transformers.

Also, this might be an issue on my side, but I wasn't able to get `torch.compile` working:
```
torch._dynamo.exc.Unsupported: Data-dependent branching
  Explanation: Detected data-dependent branching (e.g. `if my_tensor.sum() > 0:`). Dynamo does not support tracing dynamic control flow.
  Hint: This graph break is fundamental - it is unlikely that Dynamo will ever be able to trace through your code. Consider finding a workaround.
  Hint: Use `torch.cond` to express dynamic control flow.
```

In short: you've done some real magic here, and it's (at the least) very similar in accuracy, albeit likely quite a bit slower due to not forcing the unpadding/repadding (iiuc). However, that was never very `transformers`-y, and with this PR it should be much easier to build on top of ModernBERT again.

- Tom Aarsen

vasqu:
> The old implementation automatically 'unpadded' and 'repadded' during inference. I take it this variant doesn't do that? You write about a DataCollatorWithFlattening etc., is this required now for this speedup? Is there any way to know whether any given transformers models supports this class? It would be valuable for me to know for Sentence Transformers.

Pretty much any model that supports attention backend should be able to do this. I refactored Bert for exactly the same reason for example (vLLM needs this as well). Transformers was never meant to manually pad/unpad during the forward, that's what the collator is for and is recommended for causal models like llama for quite a while now - this enables complete compile compatibility with FA. 

Some side note: we manually pad/unpad ourselves already in FA (given a non all true attention mask)

> albeit likely quite a bit slower due to not forcing the unpadding/repadding (iiuc)

Should be fast when properly preparing the data :D 

> Also, this might be an issue on my side, but I wasn't able to get torch.compile working

Can you share a script and env? The example works on my end (FA), also check SDPA with non fullgraph. Note
- The data collator is necessary for clean compile compatibility - attention masks should also work but not full graph and minor graph breaks
- Always compile the forward only
- SDPA does not need the collator, it's only for FA

tomaarsen:
> The data collator is necessary for clean compile compatibility 

I see, I might've gone wrong there. I'll run more tests later. I was testing with Sentence Transformers, which doesn't use the collator.

- Tom Aarsen

YangKai0616:
Hi everyone, any updates? 😀

src/transformers/models/modernbert/configuration_modernbert.py
```diff
@@ -135,6 +136,15 @@ class ModernBertConfig(PreTrainedConfig):
     keys_to_ignore_at_inference = ["past_key_values"]
     default_theta = {"global": 160_000.0, "local": 10_000.0}
 
+    def __setattr__(self, name, value):
+        if name == "reference_compile" and value is not None:
+            logger.warning_once(
+                "The `reference_compile` argument is deprecated and will be removed in a future version. "
```
ArthurZucker:
```suggestion
                "The `reference_compile` argument is deprecated and will be removed in a `transformers v5.2.0"
```
 let's put a proper date. This is ~2 weeks we could do more

src/transformers/models/modernbert/modeling_modernbert.py
```diff
@@ -505,112 +251,95 @@ def __init__(self, config: ModernBertConfig, layer_id: int | None = None):
...
-        self.Wqkv = nn.Linear(config.hidden_size, 3 * self.all_head_size, bias=config.attention_bias)
-        layer_type = config.layer_types[layer_id]
+        self.Wqkv = nn.Linear(
+            config.hidden_size, 3 * self.head_dim * config.num_attention_heads, bias=config.attention_bias
+        )
 
-        if layer_id % config.global_attn_every_n_layers != 0:
-            self.local_attention = (config.local_attention // 2, config.local_attention // 2)
-            max_position_embeddings = config.local_attention
+        if layer_idx % config.global_attn_every_n_layers != 0:
```
ArthurZucker: We should use `config.layer_types` here no?

src/transformers/models/modernbert/modeling_modernbert.py
```diff
@@ -1056,84 +546,24 @@ def get_output_embeddings(self):
...
             inputs_embeds=inputs_embeds,
-            indices=indices,
-            cu_seqlens=cu_seqlens,
-            max_seqlen=max_seqlen,
-            batch_size=batch_size,
-            seq_len=seq_len,
-            output_attentions=output_attentions,
-            output_hidden_states=output_hidden_states,
-            return_dict=return_dict,
+            return_dict=True,
```
ArthurZucker: I don't think you need `return_dict` at all here thanks to `can_return_tuple`

tests/models/modernbert/test_modeling_modernbert.py
ArthurZucker, line 1: great new integration test, I don't think we still need the other custom tests as now the code is quite standard (like disk offload)

ArthurZucker (approved):
@YangKai0616 great great PR! 🤗 
Really aligned with the philosophy and is a clear simplification! Kudos 

Thanks @vasqu for the thorough review. 

Only 1 nit on sliding window settings based off `layer_types` if possible but otherwise LGTM

src/transformers/models/modernbert/configuration_modernbert.py
YangKai0616: Done.

src/transformers/models/modernbert/modeling_modernbert.py
YangKai0616: Done.

src/transformers/models/modernbert/modeling_modernbert.py
YangKai0616: Done.

vasqu:
run-slow: modernbert, modernbert_decoder

vasqu:
Green CI, no regression, huge work @YangKai0616! Thanks a lot for this PR and iterating time and time again :heart: it will be merged after the docs are built

PR #43042, opened 2025-12-26
[SAM3] Fix MPS race condition in add_point_inputs

vykhovanets:
# What does this PR do?

For Apple Sillicon this PR enables SAM2 and SAM3 inference on MPS device.

## Problem
SAM2/SAM3 video inference produces "no object" masks on MPS (all values = -1024), while CPU inference was correct.

## Root Cause
`non_blocking=True` in `add_point_inputs` causes a race condition: the CPU source tensor gets garbage collected before the async copy to MPS completes, corrupting `point_coords` and `point_labels`.

## Fix
Change `non_blocking=True` to `non_blocking=False` in `add_point_inputs`.

For my tests MPS inference is **1.78x faster** on and now produces identical masks to CPU.

Why change is made in SAM2 code: because the sam3 code is generated with [utils/modular_model_converter.py](utils/modular_model_converter.py) and change in sam2 code is propagated there.


vykhovanets:
Hi @SangbumChoi! Can you please look through my PR, it adds MPS support for SAM3

SangbumChoi:
@vykhovanets It seems simple change but one question is why only applying optimization in `add_point_inputs` only?

vykhovanets:
Good question @SangbumChoi! After additional testing it seems like `add_points_inputs` is the only place where this race-condition happens.

I checked, there is also `add_mask_inputs` function, but input_masks are already on the MPS device before its invocation(`process_new_mask_for_video_frame`), that is why `non_blocking=True` in `add_mask_inputs` is a no-op. 

Screenshot 2025-12-26 at 12 37 42

atultw:
Thanks for finding the root cause. I can confirm that before this fix, when using sam3_tracker_video on MPS I always get the same -1024 value in the mask, whereas on CPU I get the right mask. 

After your fix it works perfectly.

atultw (commented):
I tested this on MPS for sam3_tracker_video and the change has the intended effect. However I don't know how it would affect usage on CUDA where it already works (?). Should we only use False for non_blocking if the device is MPS and keep it True otherwise?

vykhovanets:
@atultw as far as I know CUDA can implicitly synchronize, and this race-condition is due to different memory handling by MPS, but we shouldn't introduce additional checks, because `non_blocking=False` hardens execution - that is good for all platforms, and additionally it won't make execution slower because input points are tiny tensors.

yonigozlan (approved):
Hey @vykhovanets ! Thanks for fixing this. Indeed it is low-impact in any case for other devices and if it can fix issues with mps I'm ok with merging this.

PR #43049, opened 2025-12-26
Add multimodal_chat sample

### LINKED ISSUE #43045: Multimodal chat sample
### Feature request

Add a sample covering chat scenario including images, videos or audio.

### Motivation

`AutoModelForCausalLM`'s `use_cache` is barely documented.
Describe a pattern handling the following cases
1. Tokenizer replaces tokens that are already in kv cache with a different token. For example, the model generated 2 tokens with string representations: `a` and `b` and the tokenizer replaces them with a single `a b` token on the next iteration invalidating a part of kv cache
2. Reuse embeddings computed earlier for non text modalities

There's https://github.com/huggingface/transformers/blob/a7f29523361b2cc12e51c1f5133d95f122f6f45c/src/transformers/cli/chat.py but it doesn't cover non text modalities.

### Your contribution

I'm fine to submit a PR. That will help me to learn along the way. But I need guidance how to resolve the issues I described in the motivation section.

Wovchena:
# What does this PR do?
Add a sample covering kv cache management

Fixes https://github.com/huggingface/transformers/issues/43045

## Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/examples/generation/multimodal_chat_sample.py
+++ b/examples/generation/multimodal_chat_sample.py
@@ -1,0 +1,77 @@
+import torch
+
+from transformers import DynamicCache, Qwen3VLForConditionalGeneration, Qwen3VLProcessor, TextStreamer
+
+
+def first_diff_nd(a: torch.Tensor, b: torch.Tensor):
+    """
+    Returns a tuple of indices (i0, i1, ..., ik) for the first element
+    where a != b in row-major order. Returns None if no difference.
+    """
+    if a.shape[1] != b.shape[1]:
+        len = min(a.shape[1], b.shape[1])
+        a = a[:, :len]
+        b = b[:, :len]
+
+    mismatch = a != b
+    ids = mismatch.nonzero(as_tuple=False)
+    if ids.numel() == 0:
+        return a.shape
+    # nonzero() returns coordinates sorted in row-major order, so ids[0] is the first
+    return tuple(int(x.item()) for x in ids[0])
+
+
+def trim_cache(past_key_values, diverged_idx):
+    for keys, values, _sliding_window_tensor in past_key_values:
+        if keys is None:
+            continue
+        keys = keys[:, : diverged_idx[1], :]
+        values = values[:, : diverged_idx[1], :]
+
+
+def main():
+    model = Qwen3VLForConditionalGeneration.from_pretrained(
+        "Qwen/Qwen3-VL-2B-Instruct", dtype="auto", device_map="auto"
+    )
+    processor = Qwen3VLProcessor.from_pretrained("Qwen/Qwen3-VL-2B-Instruct")
+    past_key_values = DynamicCache(config=model.config)
+    input_ids_in_cache = torch.tensor([[]], dtype=torch.int64)
+    messages = [
+        {
+            "role": "user",
+            "content": [
+                {
+                    "type": "image",
+                    "image": "https://qianwen-res.oss-cn-beijing.aliyuncs.com/Qwen-VL/assets/demo.jpeg",
+                },
+                {"type": "text", "text": "Describe this image in 5 words."},
+            ],
+        }
+    ]
+    for prompt in None, "What's the dog's color?":
+        if prompt:
+            messages.append({"role": "user", "content": [{"type": "text", "text": prompt}]})
+        inputs = processor.apply_chat_template(
+            messages, tokenize=True, add_generation_prompt=True, return_dict=True, return_tensors="pt"
+        )
+        diverged_idx = first_diff_nd(inputs["input_ids"], input_ids_in_cache)
+        trim_cache(past_key_values, diverged_idx)
+        input_length = inputs["input_ids"].shape[1]
+        if prompt:
+            del inputs["pixel_values"]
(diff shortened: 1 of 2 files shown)


examples/generation/multimodal_chat_sample.py
Wovchena, line 1: Should the word "sample" be in the file name given it's already in `examples` folder?
There's already `examples/pytorch/text-generation/run_generation.py`. Should they be merged?

examples/generation/multimodal_chat_sample.py
```diff
@@ -0,0 +1,77 @@
...
+                    "image": "https://qianwen-res.oss-cn-beijing.aliyuncs.com/Qwen-VL/assets/demo.jpeg",
+                },
+                {"type": "text", "text": "Describe this image in 5 words."},
+            ],
+        }
+    ]
+    for prompt in None, "What's the dog's color?":
+        if prompt:
+            messages.append({"role": "user", "content": [{"type": "text", "text": prompt}]})
+        inputs = processor.apply_chat_template(
```
Wovchena, line 54: The image is recomputed every time. Can you suggest a plan and appropriate processor's methods to call to avoid that?

examples/generation/multimodal_chat_sample.py
```diff
@@ -0,0 +1,77 @@
...
+            messages, tokenize=True, add_generation_prompt=True, return_dict=True, return_tensors="pt"
+        )
+        diverged_idx = first_diff_nd(inputs["input_ids"], input_ids_in_cache)
+        trim_cache(past_key_values, diverged_idx)
+        input_length = inputs["input_ids"].shape[1]
+        if prompt:
+            del inputs["pixel_values"]
+            del inputs["image_grid_thw"]
```
Wovchena, line 62: If I don't delete the entries, there's a mismatch of the number of image tokens during `generate()`. It seems `past_key_values` arg trims `inputs['input_ids']` removing the image placeholders

Can you recommend a way to handle this?

examples/generation/multimodal_chat_sample.py
CodersAcademy006: I agree that having `sample` in the filename is a bit redundant since this already lives under `examples/`. Renaming it to something like `multimodal_chat.py` would make it more consistent with other examples. As for merging with `run_generation.py`, I’d lean toward keeping this separate. This script demonstrates a multimodal chat workflow (vision + text), which feels conceptually different from generic text generation. Keeping it standalone helps with clarity and discoverability.

examples/generation/multimodal_chat_sample.py
CodersAcademy006: the image does seem to be reprocessed on every iteration. A cleaner approach would be to preprocess the image once before entering the chat loop, cache the resulting vision features or image tokens, and then reuse them while only updating the text inputs each turn. This would avoid unnecessary computation and also make it clearer to readers that the image context is static while the conversation evolves.


i hope this may help.

examples/generation/multimodal_chat_sample.py
CodersAcademy006: The mismatch makes sense given how `past_key_values` trims `input_ids` and assumes a text-only growing sequence. One way to handle this more robustly would be to treat the image tokens as fixed context, either by caching the image features separately from the text cache or by explicitly reinserting the image placeholders before calling generate(). Adding a short comment explaining this constraint in the example would also go a long way in helping users understand why this extra handling is needed in a multimodal setup. 

rest looks normal.

CodersAcademy006 (changes requested):

CodersAcademy006:
One possible way to address both the repeated image recomputation and the `past_key_values` / image-token mismatch is to treat the image as fixed context and cache its features outside the chat loop. The image is preprocessed once, its vision-related tensors are reused for every turn, and only the text inputs grow while updating `past_key_values`. This avoids recomputing vision features on each iteration and prevents image placeholder tokens from being trimmed during generation.

```python
# --- Preprocess image ONCE ---
vision_inputs = processor(
    images=image,
    return_tensors="pt"
).to(device)

# Cache vision-related tensors (keep them static)
cached_vision = {
    k: v for k, v in vision_inputs.items()
    if k != "input_ids"
}

past_key_values = None
generated_text = ""

for user_message in chat_history:
    # --- Text-only processing per turn ---
    text_inputs = processor(
        text=user_message,
        return_tensors="pt"
    ).to(device)

    # --- Merge cached vision + new text ---
    model_inputs = {
        **cached_vision,
        "input_ids": text_inputs["input_ids"],
        "attention_mask": text_inputs["attention_mask"],
        "past_key_values": past_key_values,
    }

    outputs = model.generate(
        **model_inputs,
        max_new_tokens=128,
        use_cache=True,
    )

    past_key_values = outputs.past_key_values
    generated_text += processor.decode(outputs[0], skip_special_tokens=True)
````

If the pattern is adopted, the scope of this PR becomes very clear and strong: it serves as a reference-quality multimodal chat example that demonstrates best practices for image caching and correct use of text-only KV caching in a multimodal setting, remember not to touch the core modules or api's in this pr.


Also this is a simple workaround, any other approach is warmly welcome.

PR #43050, opened 2025-12-26
Fix safe _file_ access in CLIP lazy module

Olexandr88:
Avoid potential KeyError when accessing _file_ in CLIP lazy module.

### DIFF
--- a/src/transformers/models/clip/__init__.py
+++ b/src/transformers/models/clip/__init__.py
@@ -28,5 +28,5 @@ if TYPE_CHECKING:
 else:
     import sys
 
-    _file = globals()["__file__"]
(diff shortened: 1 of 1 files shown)


Rocketknight1:
`__file__` may indeed not be defined (I think when Python is running in an interpreter), but I suspect downstream code will probably crash when `_file = ""` , right? Does this actually fix any issues?

Olexandr88:
> `__file__` may indeed not be defined (I think when Python is running in an interpreter), but I suspect downstream code will probably crash when `_file = ""` , right? Does this actually fix any issues?


Updated to globals().get("__file__") to avoid KeyError without introducing a fake path

PR #43056, opened 2025-12-27
Perf: enable pin_memory in DataLoader for CLM no_trainer example

RinZ27:
This PR enables `pin_memory=True` in the `DataLoader` for the causal language modeling standalone example.

Pinned memory is a PyTorch best practice that improves host-to-device transfer efficiency.

### DIFF
--- a/examples/pytorch/language-modeling/run_clm_no_trainer.py
+++ b/examples/pytorch/language-modeling/run_clm_no_trainer.py
@@ -507,10 +507,17 @@ def main():
 
     # DataLoaders creation:
     train_dataloader = DataLoader(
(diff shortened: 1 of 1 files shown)


ArthurZucker:
thanks! tho we don't really maintain these anymore! @merveenoyan do you want to merge?

RinZ27:
Hi @ArthurZucker, thanks for the response. @merveenoyan, just checking in to see if you have any thoughts on merging this. It's a small change but would be beneficial for anyone still using these examples for reference. Thanks!

RinZ27:
I've squashed the commits into a single clean one and ensured the `pin_memory` configuration properly checks for CUDA availability. The implementation is now streamlined and ready. @merveenoyan, would you mind taking a look for a final review when you have a moment? Thanks!

PR #43057, opened 2025-12-27
test: add batched inference test for ObjectDetectionPipeline

majiayu000:
## Summary

Add test coverage for batched inference in the ObjectDetectionPipeline:
- Tests that using `batch_size` parameter with DataLoader-based batching correctly returns results for all input images
- Verifies that each image in the batch receives its own detection results

This test ensures that the pipeline's batched processing path works correctly, addressing the issue described in #31356.

## Test plan

- [x] New test `test_small_model_pt_batched` passes locally
- [x] Existing `test_small_model_pt` continues to pass
- [ ] CI pipeline tests pass

Fixes #31356


ArthurZucker:
cc @yonigozlan

PR #43058, opened 2025-12-28
Fix ObjectDetectionPipeline batch processing bug

ayushm98:
# What does this PR do?

Fixes #31356 

This PR fixes a bug in the `ObjectDetectionPipeline` where only the first image was processed when using the `batch_size` parameter.

## Problem

When using `batch_size > 1`, the pipeline would batch multiple images together in the forward pass, but the `postprocess` method only processed the first element (`raw_annotations[0]`) instead of iterating through all annotations.

## Solution

Modified the `postprocess` method in `object_detection.py` to:
1. Loop through all `raw_annotations` returned by `post_process_object_detection`
2. Process each annotation individually  
3. Handle both single image and batch cases correctly (returns single list for one image, list of lists for multiple)

## Testing

- Added regression test `test_batch_size_parameter` that verifies all images are processed when using `batch_size` parameter
- Existing tests continue to pass (single image and multi-image without batch_size)

## Who can review?

cc @yonigozlan

### DIFF
--- a/src/transformers/pipelines/object_detection.py
+++ b/src/transformers/pipelines/object_detection.py
@@ -159,23 +159,28 @@ class ObjectDetectionPipeline(Pipeline):
         else:
             # This is a regular ForObjectDetectionModel
             raw_annotations = self.image_processor.post_process_object_detection(model_outputs, threshold, target_size)
-            raw_annotation = raw_annotations[0]
-            scores = raw_annotation["scores"]
-            labels = raw_annotation["labels"]
-            boxes = raw_annotation["boxes"]
 
-            raw_annotation["scores"] = scores.tolist()
-            raw_annotation["labels"] = [self.model.config.id2label[label.item()] for label in labels]
-            raw_annotation["boxes"] = [self._get_bounding_box(box) for box in boxes]
+            # Handle both single image and batch cases
+            annotations = []
+            for raw_annotation in raw_annotations:
+                scores = raw_annotation["scores"]
+                labels = raw_annotation["labels"]
+                boxes = raw_annotation["boxes"]
 
-            # {"scores": [...], ...} --> [{"score":x, ...}, ...]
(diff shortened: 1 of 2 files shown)


yonigozlan (commented):
Hey @ayushm98 ! Given how pipelines work, this shouldn't be an issue as the batching/packing is done independently of the pipeline type (handled by the Pipeline object). Could you provide a code sample to reproduce this issue?

ayushm98:
Hi @yonigozlan! Thanks for taking a look.

The issue is specifically in the `postprocess` method of `ObjectDetectionPipeline`. While you're correct that the Pipeline base class handles batching, the problem occurs when `post_process_object_detection` returns results for multiple images (when batch_size > 1), but the current code only processes the first result with `raw_annotation = raw_annotations[0]`.

Here's a code sample to reproduce:

```python
from transformers import AutoModelForObjectDetection, AutoImageProcessor, ObjectDetectionPipeline

model_id = "hf-internal-testing/tiny-detr-mobilenetsv3"
model = AutoModelForObjectDetection.from_pretrained(model_id)
image_processor = AutoImageProcessor.from_pretrained(model_id)
object_detector = ObjectDetectionPipeline(model=model, image_processor=image_processor)

# Test with 4 images and batch_size=2
images = [
    "http://images.cocodataset.org/val2017/000000039769.jpg",
] * 4

outputs = object_detector(images, threshold=0.0, batch_size=2)
print(f"Number of outputs: {len(outputs)}")  # Should be 4, but may not handle all correctly
```

The fix ensures that when `post_process_object_detection` returns multiple results (one per image in the batch), we iterate through all of them instead of just taking the first one. The test case in the PR (test_batch_size_parameter) validates this behavior.

Let me know if you need any clarification!

yonigozlan:
I get the expected value (4) with the given code to reproduce the issue. Looks like this might be an LLM PR, closing for now

PR #43060, opened 2025-12-28
Fix InvalidVersion exception in is_flash_attn_2_available

ayushm98:
Fixes #42999

## Summary
When  package is not installed or its version cannot be determined,  returns  as the version string. This caused  to raise  exception.

## Changes
Added a check for  before attempting to parse the version, returning  early if the version cannot be determined.

## Test Plan
- Tested with flash_attn not installed - should return False without exception
- Tested with flash_attn 2.x installed - should work as before
- Tested with flash_attn 3.x only installed - should properly skip FA2 check and allow FA3 to be used


vasqu:
Closing as #43004 has been merged to fix this

PR #43061, opened 2025-12-28
SAM3 Video: Remove call to pin_memory if the target device is not CUDA

atultw:
# What does this PR do?
Remove the call to pin_memory if the target device is not CUDA. 

This library has been recommended for people who can't use the main SAM3 repo due to CUDA requirement, for example on MPS: https://huggingface.co/facebook/sam3/discussions/11 

I believe pin_memory is only effective for CUDA type devices. There hasn't been movement on pin_memory fix for MPS since 2022: https://github.com/pytorch/pytorch/issues/86060 

Transformers is the easiest way to run SAM3 on a Mac, and this change would let us properly use SAM3 Video. 

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@yonigozlan

### DIFF
--- a/src/transformers/models/sam3_video/processing_sam3_video.py
+++ b/src/transformers/models/sam3_video/processing_sam3_video.py
@@ -340,7 +340,10 @@ class Sam3VideoProcessor(ProcessorMixin):
 
             # slice those valid entries from the original outputs
             keep_idx = torch.nonzero(keep, as_tuple=True)[0]
-            keep_idx_gpu = keep_idx.pin_memory().to(device=out_binary_masks.device, non_blocking=True)
+            if out_binary_masks.device.type == "cuda":
+                keep_idx_gpu = keep_idx.pin_memory().to(device=out_binary_masks.device, non_blocking=True)
+            else:
+                keep_idx_gpu = keep_idx.to(device=out_binary_masks.device, non_blocking=True)
 
             out_obj_ids = torch.index_select(out_obj_ids, 0, keep_idx)
             out_probs = torch.index_select(out_probs, 0, keep_idx)
--- a/src/transformers/models/sam3_video/processing_sam3_video.py
+++ b/src/transformers/models/sam3_video/processing_sam3_video.py
@@ -340,10 +340,7 @@ class Sam3VideoProcessor(ProcessorMixin):
 
             # slice those valid entries from the original outputs
             keep_idx = torch.nonzero(keep, as_tuple=True)[0]
-            if out_binary_masks.device.type == "cuda":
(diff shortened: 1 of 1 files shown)


CodersAcademy006:
@atultw  The logic is sound and low risk, but it would be good to add a small test covering the non-CUDA path to prevent future regressions, and a brief inline comment explaining why `pin_memory()` is gated to CUDA would improve maintainability.

vykhovanets:
Hi @atultw! I recently made a PR that fixes MPS output for SAM3.
https://github.com/huggingface/transformers/pull/43042 

If fixes the input, actually, so the output is correct :)

As far as I know `pin_memory` is ignored if device is not CUDA, but if it actually makes it work on your machine - great!

atultw:
> As far as I know `pin_memory` is ignored if device is not CUDA

This may be the case if you run everything on CPU, let me test that and report back. 

When running on MPS, it does not appear to be ignored because I get the same error as https://github.com/pytorch/pytorch/issues/86060. Maybe we need to only remove the call if the target device is MPS? Rather than all non-CUDA. 

Edit: I just reviewed your PR and reproduced your issue with sam3_tracker_video on MPS. Thanks for finding that root cause. I did not run into the pin_memory error with sam3_tracker_video. This PR is for sam3_video.

src/transformers/models/sam3_video/processing_sam3_video.py
```diff
@@ -340,7 +340,10 @@ def postprocess_outputs(
 
             # slice those valid entries from the original outputs
             keep_idx = torch.nonzero(keep, as_tuple=True)[0]
-            keep_idx_gpu = keep_idx.pin_memory().to(device=out_binary_masks.device, non_blocking=True)
+            if out_binary_masks.device.type == "cuda":
+                keep_idx_gpu = keep_idx.pin_memory().to(device=out_binary_masks.device, non_blocking=True)
+            else:
+                keep_idx_gpu = keep_idx.to(device=out_binary_masks.device, non_blocking=True)
```
yonigozlan: Let's just remove pin_memory althogether

yonigozlan (commented):
Hey @atultw ! Thanks for raising this issue. We don't really need pin_memory here at all, that's an overlook on my part. We can just remove it altogether and keep only the `to()` part

src/transformers/models/sam3_video/processing_sam3_video.py
atultw: Done

yonigozlan (approved):
Thanks for fixing this @atultw !

PR #43063, opened 2025-12-29
Improve documentation for SegFormer image processor

sindhukodati:
This PR adds clearer documentation for the SegFormer image processor.


stevhliu (commented):
hmm i don't think its worth adding extra explanation for what it does since its pretty self-explanatory

PR #43071, opened 2025-12-30
Separate renaming from merging in Mixtral conversion mapping

hmellor:
This allows the Transformers modeling backend for vLLM to load Mixtral with Transformers v5 because it currently (once https://github.com/vllm-project/vllm/pull/31545 is merged) only supports `WeightRenaming` but not `WeightConverter`


ArthurZucker (approved):
It makes sense now that we know for sure that renaming always happens first! cc @Cyrilvallez fine with you as well?

Cyrilvallez:
Humm, in theory should not be too much of a problem, but tests complain here (`test_reverse_loading_mapping` failing is I believe due to the test itself that checks source patterns, so expected to fail without combining both Renaming and Conversion after this change), but others seem to indicate that conversion may not happen as expected.
Do you want to dive in it @hmellor? Otherwise I can do it, let me know!

hmellor:
Yeah I noticed the test failed but didn't know enough about the new system to figure out why. 

I won't be able to look at it for a few weeks so if you can find time that'd be awesome

hmellor:
Superseded by https://github.com/huggingface/transformers/pull/43071

PR #43077, opened 2026-01-01
compileable=>compilable

cyyever:
# What does this PR do?

This PR fixes compileable to compilable in source code.


Rocketknight1:
This is giving me flashbacks to English class, where my teacher called me up on "aging" being the American spelling, with "ageing" being the UK one. "Compilable" is probably a rare enough word that there is no authority on this - are we sure it's the correct form for both UK and US English?

PR #43084, opened 2026-01-03
Fix Qwen3OmniMoe Talker weight loading and config initialization

HuiyingLi:
# What does this PR do?

This PR fixes several issues preventing Qwen3OmniMoeForConditionalGeneration from loading and running correctly from the `Qwen/Qwen3-Omni-30B-A3B-Instruct` checkpoint.

# Issues Fixed
Issue: https://github.com/huggingface/transformers/issues/43083
1. Incorrect inherited _tied_weights_keys in Talker
Qwen3OmniMoeTalkerForConditionalGeneration inherits from Qwen3MoeForCausalLM, which defines _tied_weights_keys = {"lm_head.weight": "model.embed_tokens.weight"}. However, the Talker model uses codec_head instead of lm_head and doesn't tie weights. This caused loading errors as the loader expected keys that don't exist in the checkpoint. This causes garbled audio output.
Fix: Override _tied_weights_keys = {} in the Talker class.

2. Incorrect _tp_plan and _pp_plan references
The inherited tensor/pipeline parallelism plans reference lm_head, but the Talker uses codec_head.
Fix: Override with _tp_plan = {"codec_head": "colwise_rep"} and _pp_plan = {"codec_head": (["hidden_states"], ["logits"])}.
3. Missing initializer_range in config classes
Qwen3OmniMoeTalkerConfig, Qwen3OmniMoeCode2WavConfig, and Qwen3OmniMoeConfig were missing the initializer_range attribute. This caused AttributeError during _init_weights() calls.
Fix: Add initializer_range attribute to all affected config classes.

## Who can review?

@zucchini-nlp @ArthurZucker @Cyrilvallez

### DIFF
--- a/src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
+++ b/src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
@@ -907,6 +907,7 @@ class Qwen3OmniMoeTalkerConfig(PreTrainedConfig):
         self.audio_start_token_id = audio_start_token_id
         self.vision_start_token_id = vision_start_token_id
         self.speaker_id = speaker_id
+        self.initializer_range = self.text_config.initializer_range
         super().__init__(**kwargs)
 
 
@@ -997,6 +998,7 @@ class Qwen3OmniMoeCode2WavConfig(PreTrainedConfig):
         upsampling_ratios=(2, 2),
         decoder_dim=1536,
         attention_dropout=0.0,
+        initializer_range=0.02,
         **kwargs,
     ):
         self.codebook_size = codebook_size
@@ -1016,6 +1018,7 @@ class Qwen3OmniMoeCode2WavConfig(PreTrainedConfig):
         self.upsampling_ratios = upsampling_ratios
         self.decoder_dim = decoder_dim
         self.attention_dropout = attention_dropout
+        self.initializer_range = initializer_range
         self.rope_parameters = rope_parameters
 
         super().__init__(**kwargs)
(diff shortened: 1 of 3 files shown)


src/transformers/models/qwen3_omni_moe/modeling_qwen3_omni_moe.py
```diff
@@ -3042,9 +3042,9 @@ def get_input_embeddings(self):
 
 @auto_docstring
 class Qwen3OmniMoeTalkerForConditionalGeneration(Qwen3OmniMoeThinkerTextPreTrainedModel, GenerationMixin):
-    _tied_weights_keys = {"lm_head.weight": "model.embed_tokens.weight"}
```
zucchini-nlp, line 3045: I believe the correct way is `codec_head: model.codec_embedding.weight`. It will allow users to tie weights if needed. We just need to make sure that the model is not tying weight, I see that the default is already `tie_word_embeddings=False` in config

src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
```diff
@@ -907,6 +907,7 @@ def __init__(
         self.audio_start_token_id = audio_start_token_id
         self.vision_start_token_id = vision_start_token_id
         self.speaker_id = speaker_id
+        self.initializer_range = self.text_config.initializer_range
```
zucchini-nlp, line 910: since we're using text config's init range in any case, we can instead update `init_weights` to use `config.get_text_config().init_range`

src/transformers/models/qwen3_omni_moe/modeling_qwen3_omni_moe.py
HuiyingLi: Thank you very much @zucchini-nlp ! Fixed.

ArthurZucker (approved):
TY!

PR #43085, opened 2026-01-03
Add async_stopping_criteria flag to reduce GPU-CPU syncs during generation

### LINKED ISSUE #43086: Add async_stopping_criteria flag to reduce GPU-CPU synchronization overhead
### Feature request

Add an `async_stopping_criteria` flag to `GenerationConfig` that performs stopping criteria checks asynchronously on a separate CUDA stream. This reduces GPU-CPU synchronization overhead during autoregressive text generation by allowing the model to continue generating tokens while stopping criteria (EOS detection, max_length, custom criteria) are being evaluated in parallel.

Key implementation details:
- Uses a separate CUDA stream for stopping criteria evaluation
- Employs pinned (page-locked) CPU memory for efficient GPU-CPU communication without explicit synchronization
- Implements batched polling to minimize overhead (only polls for async results every N tokens)
- Gracefully falls back to synchronous behavior on CPU

### Motivation

GPU-CPU synchronization during stopping criteria checks creates a significant bottleneck in autoregressive text generation. Each iteration of the generation loop currently requires:
1. Running the forward pass on GPU
2. Synchronizing to check stopping criteria (e.g., `unfinished_sequences.max() == 0`)
3. Only then proceeding to the next token

This synchronization blocks the GPU while the CPU evaluates stopping conditions. By moving these checks to a separate CUDA stream, we can overlap the stopping criteria evaluation with subsequent forward passes, significantly improving throughput.

**Benchmark Results** (using `utf8-lm-tiny` model, 200 new tokens):

| Mode | Tokens/sec | Speedup |
|------|------------|---------|
| Sync (baseline) | 80.92 | 1.00x |
| Async | 137.06 | **1.69x** |

### Your contribution

I have implemented this feature and submitted a PR: #43085

The implementation includes:
- New `async_stopping_criteria` parameter in `GenerationConfig`
- `AsyncStoppingCriteriaList` class that wraps `StoppingCriteriaList` for async execution
- Integration with the generation loop in `utils.py`
- Comprehensive test coverage (10 tests) for:
  - Async/sync output equivalence
  - Various stopping criteria (EOS, max_length, custom)
  - Different batch sizes
  - CPU fallback behavior

AmitMY:
## Summary
- Adds `async_stopping_criteria` flag to `GenerationConfig` that enables asynchronous stopping criteria checks during generation
- When enabled, stopping criteria are evaluated on a separate CUDA stream, allowing generation to continue while the check runs
- Uses pinned (page-locked) CPU memory with numpy view for efficient GPU-CPU communication without explicit synchronization
- Polls for completion only when the CUDA event signals the async operation is done (no artificial delays)

Fixes #43086

## Motivation
GPU-CPU synchronization during stopping criteria checks creates a bottleneck in autoregressive text generation. Each call to check if generation should stop requires syncing with the GPU to evaluate conditions like EOS token detection. This PR overlaps these checks with generation, reducing idle time.

## Key Optimizations
1. **Async CUDA stream**: Stopping criteria run on a separate stream, overlapping with token generation
2. **Pinned memory + numpy**: Results written to pinned memory and read via numpy view - no PyTorch `.item()` sync needed
3. **Event-based polling**: Only check completion when `event.query()` returns True (no fixed polling interval)
4. **CPU-side max_length**: Length checks done on CPU without any GPU sync

## Benchmark Results

**Stopping criteria overhead (200 checks):**

| Mode | Time | Per Check | Speedup |
|------|------|-----------|---------|
| Sync | 104.65ms | 0.523ms | 1.00x |
| Async | 51.76ms | 0.259ms | **2.02x** |

## Usage
```python
model.generate(
    inputs, 
    max_new_tokens=200, 
    async_stopping_criteria=True
)
```

## Test plan
- [x] Verify correctness: async mode produces identical outputs to sync mode
  - `test_async_sync_equivalence_max_length`
  - `test_async_sync_equivalence_eos_token`
  - `test_async_sync_equivalence_partial_eos`
- [x] Test with various stopping criteria (EOS, max_length, custom criteria)
  - `test_async_sync_equivalence_max_length` - max_length criteria
  - `test_async_sync_equivalence_eos_token` - EOS token criteria
  - `test_async_custom_stopping_criteria` - custom stopping criteria
  - `test_async_multiple_eos_tokens` - multiple EOS token IDs
- [x] Test with different batch sizes
  - `test_async_different_batch_sizes` - batch sizes 1, 2, 4, 8, 16
- [x] Ensure graceful fallback when CUDA is not available
  - `test_async_cpu_fallback` - verifies sync fallback on CPU tensors
- [x] Test wrapper interface compatibility
  - `test_async_wrapper_basic` - `__len__`, `__iter__`, `max_length` property
  - `test_async_legacy_call_interface` - legacy `__call__` interface
  - `test_async_finalize` - cleanup method

### DIFF
--- a/src/transformers/generation/__init__.py
+++ b/src/transformers/generation/__init__.py
@@ -76,6 +76,7 @@ else:
         "WatermarkLogitsProcessor",
     ]
     _import_structure["stopping_criteria"] = [
+        "AsyncStoppingCriteriaList",
         "MaxLengthCriteria",
         "MaxTimeCriteria",
         "ConfidenceCriteria",
@@ -173,6 +174,7 @@ if TYPE_CHECKING:
             WhisperTimeStampLogitsProcessor,
         )
         from .stopping_criteria import (
+            AsyncStoppingCriteriaList,
             ConfidenceCriteria,
             EosTokenCriteria,
             MaxLengthCriteria,
--- a/src/transformers/generation/configuration_utils.py
+++ b/src/transformers/generation/configuration_utils.py
@@ -331,6 +331,11 @@ class GenerationConfig(PushToHubMixin):
             Whether to disable the automatic compilation of the forward pass. Automatic compilation happens when
             specific criteria are met, including using a compilable cache. Please open an issue if you find the
             need to use this flag.
+        async_stopping_criteria (`bool`, defaults to `False`):
+            If set to `True`, stopping criteria checks will be performed asynchronously on a separate CUDA stream,
+            allowing generation to continue while the check runs. This can reduce GPU-CPU synchronization overhead
+            and improve throughput, especially for longer generations. The stopping check result is polled
+            periodically rather than blocking on every token. Only effective on CUDA devices.
     """
 
     extra_output_flags = ("output_attentions", "output_hidden_states", "output_scores", "output_logits")
@@ -414,6 +419,7 @@ class GenerationConfig(PushToHubMixin):
         # Performance
         self.compile_config = kwargs.pop("compile_config", None)
         self.disable_compile = kwargs.pop("disable_compile", False)
+        self.async_stopping_criteria = kwargs.pop("async_stopping_criteria", False)
 
         # Deprecated (moved to the Hub). TODO remove for v5
         self.low_memory = kwargs.pop("low_memory", None)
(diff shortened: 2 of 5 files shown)


ArthurZucker:
cc @remi-or but as we are moving towards just improving continious batching I don't think this makes sense without it being supported int `CB` !

remi-or:
Async for CPU-GPU is super cool, it will be implemented in CB down the line for sure. I don't think the implementation in CB needs to be a blocker for implementation in `generate` though, it would be nice to have the two options improve hand in hand. Especially since Cb support much less than generate because it's less flexible on some things.

AmitMY:
Happy you like it.
I don't really see how CB has any effect here because my main use case is `batch_size=1`.
My main goal is solving #43089 which seems very doable - only three spots where there are explicit syncs.

However, I do not feel comfortable tackling all three spots at once, perhaps there is a generic solution but I could not think about one.

PR #43087, opened 2026-01-03
Make projection LayerNorm configurable in LFM2-VL

ankke:
Make projection LayerNorm configurable in LFM2-VL models.

### DIFF
--- a/src/transformers/models/lfm2_vl/configuration_lfm2_vl.py
+++ b/src/transformers/models/lfm2_vl/configuration_lfm2_vl.py
@@ -44,6 +44,8 @@ class Lfm2VlConfig(PreTrainedConfig):
             The activation function used by the multimodal projector.
         projector_hidden_size (`int`, *optional*, defaults to 2560):
             The hidden size of the multimodal projector.
+        projector_use_layernorm (`bool`, *optional*, defaults to `True`):
+            Whether to use layer norm in the multimodal projector.
         projector_bias (`bool`, *optional*, defaults to `True`):
             Whether to use bias in the multimodal projector.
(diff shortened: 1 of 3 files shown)


src/transformers/models/lfm2_vl/configuration_lfm2_vl.py
```diff
@@ -34,20 +34,22 @@ class Lfm2VlConfig(PreTrainedConfig):
...
+            projector_hidden_act (`str`, *optional*, defaults to `"gelu"`):
+                The activation function used by the multimodal projector.
+            projector_hidden_size (`int`, *optional*, defaults to 2560):
+                The hidden size of the multimodal projector.
+            projector_bias (`bool`, *optional*, defaults to `True`):
+                Whether to use bias in the multimodal projector.
+            projector_use_layernorm (`bool`, *optional*, defaults to `True`):
+                Whether to use layer norm in the multimodal projector.
+            downsample_factor (`int`, *optional*, defaults to 2):
+                The downsample_factor factor of the vision backbone.
```
tuliren: This diff will be more specific if you can replace the 8-space indentation with 4-space.

ankke: this part is auto generated with a `make` command, not sure if this is expected

ankke: re-run and fixed

zucchini-nlp (approved):
lgtm

ArthurZucker:
Hey @ankke we usually never do update a model that was already shipped, v2 means a new modeling / modular file for us as the weights are differennt! We are most probably gonna revert this but happy to add the new model when it is out!

PR #43088, opened 2026-01-03
Skip attention_mask.all() GPU-CPU sync during generation

AmitMY:
I guess there could be cases where people define a custom mask, but I don't know of any.

----

## Summary
- Adds a context variable `_attention_mask_all_true` to track if attention_mask is known to be all-True during generation
- Sets the context variable at the start of `generate()` based on the initial mask state
- Modifies `_ignore_causal_mask_sdpa()` to skip the expensive `.all()` check when the context variable indicates no padding

## Problem
During autoregressive generation, `_ignore_causal_mask_sdpa()` in `masking_utils.py` calls `padding_mask.all()` on every generated token to check if the mask contains any padding. This causes a GPU-CPU synchronization that adds ~2-3ms per token, significantly impacting generation latency.

## Solution
Since during generation we only append ones (1s) to the attention mask, if the mask starts all-True (no padding), it will remain all-True throughout generation. We check this once at the start and use a context variable to skip subsequent `.all()` checks.

## Results
- Reduces `.all()` calls from N+1 to 1 for N generated tokens
- Example: 201 calls → 1 call for 200 tokens generated
- Eliminates per-token GPU-CPU sync overhead (~2-3ms per token)

## Test plan
- [x] Verified outputs are identical with and without optimization
- [x] Verified `.all()` call count reduced from 201 to 1 for 200 tokens
- [ ] Run existing tests

### DIFF
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -41,7 +41,7 @@ from ..dynamic_module_utils import (
 )
 from ..integrations.deepspeed import is_deepspeed_zero3_enabled
 from ..integrations.fsdp import is_fsdp_managed_module
-from ..masking_utils import create_masks_for_generate
+from ..masking_utils import _attention_mask_all_true, create_masks_for_generate
 from ..pytorch_utils import isin_mps_friendly
 from ..tokenization_python import ExtensionsTrie
 from ..utils import (
@@ -2645,15 +2645,23 @@ class GenerationMixin(ContinuousMixin):
         model_kwargs["use_cache"] = generation_config.use_cache
 
         # 9. Call generation mode
-        result = decoding_method(
-            self,
-            input_ids,
-            logits_processor=prepared_logits_processor,
-            stopping_criteria=prepared_stopping_criteria,
-            generation_config=generation_config,
-            **generation_mode_kwargs,
-            **model_kwargs,
-        )
+        # Check if attention_mask is all-True to avoid per-token GPU-CPU sync in masking utils.
+        # During generation, if the mask starts all-True and we only append ones, it stays all-True.
+        attention_mask = model_kwargs.get("attention_mask")
+        mask_all_true = attention_mask is None or bool(attention_mask.all())
+        mask_token = _attention_mask_all_true.set(mask_all_true)
+        try:
+            result = decoding_method(
+                self,
+                input_ids,
+                logits_processor=prepared_logits_processor,
+                stopping_criteria=prepared_stopping_criteria,
+                generation_config=generation_config,
+                **generation_mode_kwargs,
+                **model_kwargs,
+            )
+        finally:
+            _attention_mask_all_true.reset(mask_token)
 
         return result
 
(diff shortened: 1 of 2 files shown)


src/transformers/generation/utils.py
```diff
@@ -2645,15 +2645,23 @@ def generate(
...
-            stopping_criteria=prepared_stopping_criteria,
-            generation_config=generation_config,
-            **generation_mode_kwargs,
-            **model_kwargs,
-        )
+        # Check if attention_mask is all-True to avoid per-token GPU-CPU sync in masking utils.
+        # During generation, if the mask starts all-True and we only append ones, it stays all-True.
+        attention_mask = model_kwargs.get("attention_mask")
+        mask_all_true = attention_mask is None or bool(attention_mask.all())
+        mask_token = _attention_mask_all_true.set(mask_all_true)
```
zucchini-nlp, line 2652: hm, we can also set the mask to `None` and avoid the sync in masking utils

AmitMY: meaning, we check once at the start?
```py
if bool(model_kwargs.get("attention_mask").all()):
  model_kwargs.set("attention_mask", None)
```
(or otherwise, of course, feel free to implement it as you want, sounds like a very minimal change)

zucchini-nlp: yeah, once in the beginning. Also want to cc @vasqu for attn masks, maybe we don't even need to tweak generation loop

vasqu: Only sdpa and flash attention can skip in those cases. Eager and flex attention (I think) will still need their mask to work correctly - at least when we have a causal model

Note that the way we skip mask creation atm is data-flow dependent, meaning this could indeed help in compiled cases:

SDPA https://github.com/huggingface/transformers/blob/136c621c00b2536d1f608ed3c6de59b9897afbb8/src/transformers/masking_utils.py#L212
Flash https://github.com/huggingface/transformers/blob/136c621c00b2536d1f608ed3c6de59b9897afbb8/src/transformers/masking_utils.py#L564-L565

AmitMY: Thanks @vasqu ! I clearly do not understand the complexities for each attention implementation, but I now added a breakdown per implementation: https://github.com/huggingface/transformers/issues/43089#issuecomment-3789187158
`_update_causal_mask` affects `eager` and `sdpa` (63 syncs), then `_ignore_causal_mask_sdpa` affects `sdap` with 64 syncs, and finally `_update_causal_mask` affects flash attention with 64 syncs.

My understanding is that if the attention mask is all true, we could ignore it for all these cases

vasqu: If everything is true, we can ignore for SDPA and FA - they have a flag to enable proper causality without a mask.
However, flex attention and eager attention cannot do that. They require a mask at all times (at least when we have causal models and these are the majority of models). 

So we need to differentiate per attention whether we can avoid these sync
- SDPA + FA: Yes
- Flex + Eager: No

Then we also have a few special models like PaliGemma which have special mixed masks which we would not be able to skip in any case. I assume that should be a small number of models tho

PR #43091, opened 2026-01-03
Fix Qwen3OmniMoe Talker loading and config initialization

Krish2002:
## What does this PR do?

This PR fixes two issues that prevented the model
`Qwen/Qwen3-Omni-30B-A3B-Instruct` from loading correctly with
`AutoModelForMultimodalLM`.

---

### Issue
`Qwen3OmniMoeTalkerForConditionalGeneration` deletes `lm_head` in its `__init__`
method, but it inherits `_tied_weights_keys` from its parent class
(`Qwen3MoeForCausalLM`), which references `lm_head.weight`.

During model loading, `mark_tied_weights_as_initialized()` attempts to access
`lm_head.weight`, resulting in an `AttributeError`.

### Fix
Explicitly set `_tied_weights_keys = {}` for
`Qwen3OmniMoeTalkerForConditionalGeneration`, since this model does not use tied
weights.

This change is implemented in:
- `src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py`

---

### Issue
Several composite config classes were missing the `initializer_range` attribute:
- `Qwen3OmniMoeTalkerConfig`
- `Qwen3OmniMoeCode2WavConfig`
- `Qwen3OmniMoeConfig`

When `_initialize_missing_keys()` runs during model loading, it may call
`_init_weights()` for modules that are not loaded from the checkpoint.
`_init_weights()` expects `self.config.initializer_range`, which caused an
`AttributeError`.

### Fix
Add an `initializer_range` parameter (default: `0.02`) and store it as an
attribute in the affected config classes.

This change is implemented in:
- `src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py`

---

## Breaking changes

None.

---

## Tests

Added a unit-level regression test in:

tests/models/qwen3_omni_moe/test_configuration_and_loading.py

The test verifies that:
- `Qwen3OmniMoeTalkerForConditionalGeneration` has empty tied weight keys.
- `Qwen3OmniMoeTalkerConfig`, `Qwen3OmniMoeCode2WavConfig`, and
  `Qwen3OmniMoeConfig` all define the `initializer_range` attribute.

Additionally, verified locally that the model loads successfully using the
following reproduction script:

```python
from transformers import AutoModelForMultimodalLM
import torch

model = AutoModelForMultimodalLM.from_pretrained(
    "Qwen/Qwen3-Omni-30B-A3B-Instruct",
    torch_dtype=torch.bfloat16,
    attn_implementation="flash_attention_2",
)

### DIFF
--- a/src/transformers/models/lfm2_vl/configuration_lfm2_vl.py
+++ b/src/transformers/models/lfm2_vl/configuration_lfm2_vl.py
@@ -46,6 +46,8 @@ class Lfm2VlConfig(PreTrainedConfig):
             The hidden size of the multimodal projector.
         projector_bias (`bool`, *optional*, defaults to `True`):
             Whether to use bias in the multimodal projector.
+        projector_use_layernorm (`bool`, *optional*, defaults to `True`):
+            Whether to use layernorm in the multimodal projector.
         downsample_factor (`int`, *optional*, defaults to 2):
             The downsample_factor factor of the vision backbone.
     """
@@ -61,6 +63,7 @@ class Lfm2VlConfig(PreTrainedConfig):
         projector_hidden_act="gelu",
         projector_hidden_size=2560,
         projector_bias=True,
+        projector_use_layernorm=True,
         downsample_factor=2,
         **kwargs,
     ):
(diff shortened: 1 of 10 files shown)


Rocketknight1:
cc @zucchini-nlp

zucchini-nlp:
resolved in #43084, closing as duplicate

Issue #43089, opened 2026-01-03
Generation overhead:  many GPU syncs per token + PyTorch dispatch overhead


AmitMY:
# Generation overhead: 3.25 GPU syncs per token + PyTorch dispatch overhead

## System Info

- `transformers` version: 5.0.0.dev0 (main branch)
- Platform: Linux
- Python version: 3.12
- PyTorch version: 2.x with CUDA
- GPU: NVIDIA (tested)

## Who can help?

@gante @zucchini-nlp

## Information

- [x] My own modified scripts

## Tasks

- [x] My own task or dataset (give details below)

## Reproduction

We benchmarked generation overhead using a **tiny model** (hidden_size=16, 1 layer, vocab_size=256) to isolate framework overhead from actual compute.

### Benchmark Script

```python
import time
import warnings
import torch
import torch.nn.functional as F
from transformers import LlamaForCausalLM, LlamaConfig
from transformers.generation.stopping_criteria import StoppingCriteria, StoppingCriteriaList
from transformers.generation.logits_process import LogitsProcessor, LogitsProcessorList

warnings.filterwarnings("ignore")

class AlwaysPassStoppingCriteria(StoppingCriteria):
    def __call__(self, input_ids, scores, **kwargs):
        return torch.zeros(input_ids.shape[0], dtype=torch.bool, device=input_ids.device)

class ExtraSoftmaxLogitsProcessor(LogitsProcessor):
    def __call__(self, input_ids, scores):
        return torch.log(F.softmax(scores, dim=-1) + 1e-10)

# Tiny model to minimize compute, expose overhead
config = LlamaConfig(
    vocab_size=256, hidden_size=16, intermediate_size=16,
    num_hidden_layers=1, num_attention_heads=1, num_key_value_heads=1,
    max_position_embeddings=2048, use_cache=True,
)

for device in ["cpu", "cuda"]:
    model = LlamaForCausalLM(config).to(device).eval()
    input_ids = torch.tensor([[1]], device=device)
    attention_mask = torch.ones_like(input_ids)

    stopping_criteria = StoppingCriteriaList([AlwaysPassStoppingCriteria()])
    logits_processor = LogitsProcessorList([ExtraSoftmaxLogitsProcessor()])

    # Warmup
    with torch.no_grad():
        for _ in range(3):
            model.generate(input_ids, attention_mask=attention_mask,
                          min_new_tokens=64, max_new_tokens=64,
                          stopping_criteria=stopping_criteria,
[...]

PR #43094, opened 2026-01-04
Avoid inline .item() sync in decoder start token check

pythongiant:
# What does this PR do?

This PR refactors the decoder start token check in `_prepare_decoder_input_ids_for_generation` to avoid an inline
`.all().item()` call.

The boolean condition is now computed as a tensor on-device and converted to a Python boolean at a single, explicit synchronization point. This keeps the logic composable with other boolean checks and enables future batching of multiple conditions into a single GPU–CPU sync, in line with the motivation discussed in #43089.

There is no functional or behavioral change. The refactor is purely structural and preserves all existing semantics and special-case handling (e.g. Donut and Whisper models).

## Before submitting
      Pull Request section?
      to it if that's the case.

This change does not require new tests as it preserves existing behavior
and only restructures the boolean check to avoid an inline synchronization.

## Who can review?

This change affects generation utilities.

Tagging:
- @gante
- @zucchini-nlp

### DIFF
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -915,72 +915,73 @@ class GenerationMixin(ContinuousMixin):
         model_kwargs["encoder_outputs"]: ModelOutput = encoder(**encoder_kwargs)  # type: ignore
 
         return model_kwargs
-		
-	def _prepare_decoder_input_ids_for_generation(
-	    self,
-	    batch_size: int,
-	    model_input_name: str,
-	    model_kwargs: dict[str, torch.Tensor],
-	    decoder_start_token_id: torch.Tensor,
-	    device: torch.device | None = None,
-	) -> tuple[torch.LongTensor, dict[str, torch.Tensor]]:
-	    """Prepares `decoder_input_ids` for generation with encoder-decoder models"""
-	    # 1. Check whether the user has defined `decoder_input_ids` manually. To facilitate in terms of input naming,
-	    # we also allow the user to pass it under `input_ids`, if the encoder does not use it as the main input.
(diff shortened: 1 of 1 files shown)


src/transformers/generation/utils.py
```diff
@@ -963,16 +963,22 @@ def _prepare_decoder_input_ids_for_generation(
...
-            decoder_input_ids = torch.cat([decoder_start_token_id, decoder_input_ids], dim=-1)
-            if "decoder_attention_mask" in model_kwargs:
-                decoder_attention_mask = model_kwargs["decoder_attention_mask"]
-                decoder_attention_mask = torch.cat(
-                    (torch.ones_like(decoder_attention_mask)[:, :1], decoder_attention_mask),
-                    dim=-1,
-                )
-                model_kwargs["decoder_attention_mask"] = decoder_attention_mask
-
+        else:
```
zucchini-nlp, line 966: The main change is structural/ stylistic, not a speedup. We're calling `.item()` once per generation in the very beginning, so the overhead is tiny and can be ignored

zucchini-nlp, line 966: The main change is structural/ stylistic, not a speedup. We're calling `.item()` once per generation in the very beginning, so the overhead is tiny and can be ignored

pythongiant: The intent here was to make the boolean check composable with others and avoid inline `.item()` usage. This aligns with the broader cleanup discussed in #43089 essentially to keep boolean logic on-device and batch sync points when multiple conditions exist.

pythongiant: @zucchini-nlp what do you think?

PR #43101, opened 2026-01-04
Add Siglip2Tokenizer to enforce training-time text preprocessing defaults

vaibhav-research:
# What does this PR do?

This PR makes SigLIP2 text preprocessing explicit and consistent with how the model was trained.

It introduces a model specific tokenizer (Siglip2Tokenizer) that wraps the existing GemmaTokenizer while enforcing SigLIP2’s training-time defaults (lowercasing and fixed padding/truncation to 64 tokens). AutoTokenizer and AutoProcessor are updated to use this tokenizer for SigLIP2 checkpoints, so users get the correct behavior automatically without relying on implicit processor logic.

The underlying tokenization and vocabulary remain unchanged. This is a lightweight wrapper that improves correctness, reproducibility, and clarity, especially for text embedding and retrieval use cases.

This addresses the behavior reported in #43054.

Fixes # 43054

## Before submitting
      Pull Request section?
      to it if that's the case.
- [] Did you make sure to update the documentation with your changes? Here are the

## Who can review?
@ArthurZucker @itazap


vaibhav-research:
here are the test results including the one that I added in this PR.

```
pytest -q tests/models/siglip2 -k processing

tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_call_numpy PASSED         [  4%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_call_numpy_4_channels SKIPPED [  8%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_call_pil PASSED           [ 13%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_call_pytorch PASSED       [ 17%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_can_compile_fast_image_processor SKIPPED [ 21%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_cast_dtype_device PASSED  [ 26%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_fast_is_faster_than_slow SKIPPED [ 30%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_image_processor_from_and_save_pretrained PASSED [ 34%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_image_processor_from_dict_with_kwargs PASSED [ 39%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_image_processor_preprocess_arguments PASSED [ 43%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_image_processor_properties PASSED [ 47%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_image_processor_save_load_with_autoimageprocessor PASSED [ 52%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_image_processor_to_json_file PASSED [ 56%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_image_processor_to_json_string PASSED [ 60%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_init_without_params PASSED [ 65%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_is_fast PASSED            [ 69%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_new_models_require_fast_image_processor PASSED [ 73%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_override_instance_attributes_does_not_affect_other_instances SKIPPED [ 78%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_save_load_fast_slow PASSED [ 82%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_save_load_fast_slow_auto PASSED [ 86%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_slow_fast_equivalence PASSED [ 91%]
tests/models/siglip2/test_image_processing_siglip2.py::Siglip2ImageProcessingTest::test_slow_fast_equivalence_batched PASSED [ 95%]
tests/models/siglip2/test_processing_siglip2.py::Siglip2ProcessorTest::test_siglip2_text_padding_length_64 PASSED [100%]

=================================================== warnings summary ====================================================
<frozen importlib._bootstrap>:241
  <frozen importlib._bootstrap>:241: DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute

<frozen importlib._bootstrap>:241
  <frozen importlib._bootstrap>:241: DeprecationWarning: builtin type SwigPyObject has no __module__ attribute

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=============================== 19 passed, 4 skipped, 561 deselected, 2 warnings in 9.15s ===============================

```

vaibhav-research:
I verified locally that SigLIP2 text embeddings behave as expected once the training-time preprocessing is applied (lowercasing + fixed-length padding with max_length=64).

This PR intentionally avoids changing processor defaults and instead documents the correct usage and adds a small test to prevent regressions.

fancyerii (approved):
It's ok.

zucchini-nlp (commented):
I think it should be done on a tokenizer level. Siglip model for example loads a `SiglipTokenizer` class which has a lower case called on inputs before tokenizing

@itazap can we use `SiglipTokenizer` here as well or maybe add a special `Siglip2Tokenizer` class?

vaibhav-research:
> I think it should be done on a tokenizer level. Siglip model for example loads a `SiglipTokenizer` class which has a lower case called on inputs before tokenizing
> 
> @itazap can we use `SiglipTokenizer` here as well or maybe add a special `Siglip2Tokenizer` class?

@zucchini-nlp  Thanks for reviewing this PR
SigLIP2 currently uses GemmaTokenizerFast for the text branch (see Siglip2Processor and convert_siglip2_to_hf.py). So IMHO we can’t directly reuse SiglipTokenizer (different tokenization backend).
To avoid relying on user-side lowercasing, I can add automatic lowercasing in Siglip2Processor.__call__ before passing text to the tokenizer (covers the main multimodal usage). If we strongly prefer tokenizer-level like SigLIP v1, we’d likely need a Siglip2TokenizerFast wrapper around GemmaTokenizerFast + AutoTokenizer mapping changes. Let me know which direction you’d prefer.
cc @ArthurZucker @itazap

ArthurZucker:
>  I can add automatic lowercasing in Siglip2Processor.call before passing text to the tokenizer (covers the main multimodal usage)

this sounds like the most reasonable solution no?

vaibhav-research:
> > I can add automatic lowercasing in Siglip2Processor.call before passing text to the tokenizer (covers the main multimodal usage)
> 
> this sounds like the most reasonable solution no?

thanks a lot for reviewing the PR @ArthurZucker 
Yeah, agreed. Since SigLIP2 currently uses GemmaTokenizerFast, adding lowercasing at the tokenizer level would either affect Gemma globally or require a new tokenizer class + mapping. Implementing it in Siglip2Processor.__call__ keeps the behavior SigLIP2-specific and covers the main multimodal usage. I’ll push a patch + regression test.

vaibhav-research:
> > > I can add automatic lowercasing in Siglip2Processor.call before passing text to the tokenizer (covers the main multimodal usage)
> > 
> > 
> > this sounds like the most reasonable solution no?
> 
> thanks a lot for reviewing the PR @ArthurZucker Yeah, agreed. Since SigLIP2 currently uses GemmaTokenizerFast, adding lowercasing at the tokenizer level would either affect Gemma globally or require a new tokenizer class + mapping. Implementing it in Siglip2Processor.**call** keeps the behavior SigLIP2-specific and covers the main multimodal usage. I’ll push a patch + regression test.

Thanks for the guidance @ArthurZucker , this is now addressed 

I implemented automatic lowercasing and the SigLIP2 default text padding/truncation directly in SiglipProcessor.__call__, gated to SigLIP2-branded checkpoints. This avoids changing GemmaTokenizerFast globally or introducing a new tokenizer class, while ensuring SigLIP2 users get the correct preprocessing by default via AutoProcessor.

I have also  updated the SigLIP2 docs to clarify the preprocessing expectations and note that these defaults are now handled automatically by the processor and added regression tests covering the following:
	•	identical outputs for upper/lower-case inputs
	•	default padding to length 64


please let me know if you’d prefer this logic to live elsewhere, but this seemed like the least intrusive fix.

vaibhav-research:
looking for your inputs here as well @ArthurZucker 
TIA

src/transformers/models/siglip/processing_siglip.py
```diff
@@ -19,6 +19,47 @@
...
+def _is_siglip2_checkpoint(processor) -> bool:
+    """
+    SigLIP2 checkpoints currently ship a SiglipConfig (model_type='siglip') and thus load SiglipProcessor.
+    We detect SigLIP2 primarily via the image processor type/module, with a tokenizer
+    name/path fallback.
+    """
+    # SigLIP2 uses Siglip2ImageProcessor / Siglip2ImageProcessorFast
+    image_processor = getattr(processor, "image_processor", None)
+    if image_processor is not None:
+        mod = getattr(image_processor, "__module__", "") or ""
```
zucchini-nlp: I think we need a Siglip2Tokenizer class in that case because it's not guaranteed that users load a whole processor when they only want to encode text samples

vaibhav-research: Good point @zucchini-nlp,  processor-only fixes won’t cover text-only usage via AutoTokenizer.

 SigLIP2 checkpoints ship:
	•	a SiglipConfig (model_type="siglip") → resolves to SiglipProcessor
	•	tokenizer_class="GemmaTokenizer" in tokenizer_config.json → resolves to GemmaTokenizer
	
siglip2 tokenizer_config.json
```
"model_max_length": 1000000000000000019884624838656,
  "pad_token": "<pad>",
  "padding_side": "right",
  "processor_class": "SiglipProcessor",
  "sp_model_kwargs": {},
  "spaces_between_special_tokens": false,
  "tokenizer_class": "GemmaTokenizer",
  "unk_token": "<unk>",
  "use_default_system_prompt": false
 ```
 
That means SigLIP2-specific text behavior (lowercasing + fixed-length padding to 64) is currently only applied when users go through the processor path. Text-only flows correctly end up using Gemma’s tokenizer, which is expected and correct for Gemma models, but not sufficient for SigLIP2.

We shouldn’t bake SigLIP2 behavior into GemmaTokenizer, since that would silently affect all Gemma-based checkpoints. The clean fix is a dedicated Siglip2Tokenizer wrapper subclassing GemmaTokenizer that applies SigLIP2-specific defaults, and can be explicitly selected (or wired via metadata later).

I’ll add a Siglip2Tokenizer + unit test validating lowercasing and default padding/truncation. I believe that AutoTokenizer won’t pick it up automatically until model metadata is updated, but this PR adds the correct tokenizer abstraction on the library side.

Comment (vaibhav-research) [reply src/transformers/models/siglip/processing_siglip.py]: @zucchini-nlp  I’ve updated the PR accordingly.

I’ve added a Siglip2Tokenizer wrapper (subclassing GemmaTokenizer) to explicitly encode the SigLIP2 text-training assumptions at the tokenizer level as well (lowercasing + default padding/truncation to length 64), and added unit tests covering both processor and tokenizer behavior. 

Also for completeness:  SigLIP2 model repos currently declare "tokenizer_class": "GemmaTokenizer" in tokenizer_config.json(https://huggingface.co/google/siglip2-base-patch16-224/blob/main/tokenizer_config.json#L2017) , so AutoTokenizer.from_pretrained() will continue to return GemmaTokenizer unless that metadata is updated. This PR adds the correct tokenizer implementation on the library side, but switching the config to Siglip2Tokenizer would be required for the wrapper to be picked up automatically.

looking for your feedback.

File: src/transformers/models/siglip/processing_siglip.py [hunk 2]
```diff
@@ -19,10 +19,65 @@
...
+        # Only apply SigLIP2 text normalization for SigLIP2-branded checkpoints
+        if _is_siglip2_checkpoint(self):
+            # SigLIP2 text encoder trained on lowercased text
+            text = _lowercase_text(text)
+
+            # SigLIP2 text encoder trained with padding to 64 tokens
+            # Only set defaults if the user didn't specify them.
+            kwargs.setdefault("padding", "max_length")
+            kwargs.setdefault("truncation", True)
+            kwargs.setdefault("max_length", 64)
```
zucchini-nlp: is this needed now? I think these are saved in tokenizer/processor config, except for lowercase

src/transformers/models/siglip2/tokenization_siglip2.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2024 The HuggingFace Inc. team.
```
zucchini-nlp: nit: 2025

src/transformers/models/siglip2/tokenization_siglip2.py
```diff
@@ -0,0 +1,58 @@
...
+    if x is None:
+        return None
+    if isinstance(x, str):
+        return x.lower()
+    if isinstance(x, (list, tuple)):
+        return [_lowercase_text(i) for i in x]
+    return x
+
+
+class Siglip2Tokenizer(GemmaTokenizer):
```
zucchini-nlp: if we are to inherit from existing tokenizer class, we better place it in modular. Modular will take care of copying the code and making sure that Siglip2 doesn't depend on Gemma at run-time

src/transformers/models/siglip2/tokenization_siglip2.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __call__(self, text=None, text_pair=None, **kwargs):
+        # SigLIP2 training: lowercase text
+        text = _lowercase_text(text)
+        text_pair = _lowercase_text(text_pair)
+
+        # SigLIP2 training: fixed padding/truncation to 64
+        kwargs.setdefault("padding", "max_length")
+        kwargs.setdefault("truncation", True)
+        kwargs.setdefault("max_length", 64)
```
zucchini-nlp: not sure about this. It also has to be in saved config and we can also add default values in `ModelProcessorKwargs`

src/transformers/models/siglip2/tokenization_siglip2.py
```diff
@@ -0,0 +1,58 @@
...
+      - fixed padding/truncation to 64 tokens
+
+    This tokenizer wrapper keeps Gemma's underlying vocab/tokenization,
+    while applying SigLIP2-specific preprocessing defaults.
+    """
+
+    def __call__(self, text=None, text_pair=None, **kwargs):
+        # SigLIP2 training: lowercase text
+        text = _lowercase_text(text)
+        text_pair = _lowercase_text(text_pair)
```
zucchini-nlp: nit: I see that configs on the hub have a field `do_lower_case`, imo we need to lower case if that field is set to `True`

src/transformers/models/siglip/processing_siglip.py
ArthurZucker: > AutoTokenizer.from_pretrained() will continue to return GemmaTokenizer

This is no longer true, AutoTokenizer will check if the serialized class matched the `tokenizer_mappin[model_type]` and as it does not, we can enforce whatever we want

src/transformers/models/siglip/processing_siglip.py
```diff
@@ -19,10 +19,65 @@
...
+        return [_lowercase_text(i) for i in x]
+    return x
+
+
 @auto_docstring
 class SiglipProcessor(ProcessorMixin):
     def __init__(self, image_processor, tokenizer):
         super().__init__(image_processor, tokenizer)
 
+    def __call__(self, images=None, text=None, **kwargs):
```
ArthurZucker: this should not be done here

src/transformers/models/siglip2/tokenization_siglip2.py
ArthurZucker: we never use inheritance in transformers like this. Lowercasing is a normalizer from `normalizers.Lowercase` that just needs to be added to the sequence of normalizers for siglip models.

ArthurZucker (commented):
Hey! I really don't think we should do any of the fixes here, but rather on the hub. 
We can add a siglip2 tokenizer but it should be explicit that it uses a lowercase normalizer, and should be for all Siglip2 / `siglip_2` tokenizer. If its just for 1-2 then it should be updated on the hub directly !

tests/models/siglip2/test_processing_siglip2.py
```diff
@@ -0,0 +1,51 @@
...
+
+import torch
+
+from transformers import AutoProcessor, Siglip2Tokenizer
+from transformers.testing_utils import require_tokenizers, require_torch, slow
+
+
+@require_torch
+@require_tokenizers
+@slow
```
zucchini-nlp: why is it slow?

tests/models/siglip2/test_processing_siglip2.py
```diff
@@ -0,0 +1,51 @@
...
+class Siglip2ProcessorTest(unittest.TestCase):
```
zucchini-nlp: let's inherit from `ProcessorTesterMixin` to get common tests

tests/models/siglip2/test_processing_siglip2.py
```diff
@@ -0,0 +1,51 @@
...
+        a = processor(text=["HELLO WORLD"], return_tensors="pt")
+        b = processor(text=["hello world"], return_tensors="pt")
+        self.assertTrue(torch.equal(a["input_ids"], b["input_ids"]))
+
+
+@require_torch
+@require_tokenizers
+@slow
+class Siglip2TokenizerTest(unittest.TestCase):
+    def test_lowercasing_and_padding_defaults(self):
```
zucchini-nlp: same, not slow and inherit from `TokenizerTesterMixin` to get common tests. Also tokenizer tests are usually in their own separate file

vaibhav-research:
Thanks for detailed review of this @zucchini-nlp @ArthurZucker 
I got a lot of pointers to fix :) 
will address each of these and confirm once done.
thanks again

src/transformers/models/siglip/processing_siglip.py
vaibhav-research: Got it,  thanks for clarifying. I was assuming tokenizer_class in the checkpoint would always “win” and keep routing SigLIP2 to GemmaTokenizer, but I see now AutoTokenizer checks the serialized class against the tokenizer_auto.py mapping for the model type

src/transformers/models/siglip2/tokenization_siglip2.py
vaibhav-research: You’re right @ArthurZucker.  I wasn’t using a backend normalizer. After your review, I updated the implementation to apply lowercasing via normalizers.Lowercase() on the tokenizer backend.

src/transformers/models/siglip/processing_siglip.py
vaibhav-research: thanks for reviewing @zucchini-nlp , it is not needed anymore and hence I removed in my latest commit.

src/transformers/models/siglip2/tokenization_siglip2.py
vaibhav-research: fixed

src/transformers/models/siglip2/tokenization_siglip2.py
vaibhav-research: Agreed @zucchini-nlp.  I moved the tokenizer into the Siglip2 modular file so the generated code is self-contained and avoids any runtime dependency on Gemma.

src/transformers/models/siglip2/tokenization_siglip2.py
vaibhav-research: sure, moved it to ModelProcessorKwargs

src/transformers/models/siglip2/tokenization_siglip2.py
vaibhav-research: got it, updated the code to use do_lower_case field

tests/models/siglip2/test_processing_siglip2.py
vaibhav-research: I was not using the fast tokenizer hence thought of using it but I have updated the test cases based on latest changes

tests/models/siglip2/test_processing_siglip2.py
vaibhav-research: sure, I updated the test cases to inherit tests from ProcessorTesterMixin

tests/models/siglip2/test_processing_siglip2.py
vaibhav-research: sure, I updated the test cases to use TokenizerTesterMixin

src/transformers/models/siglip/processing_siglip.py
vaibhav-research: agreed, removed this in my latest commit

src/transformers/models/siglip/processing_siglip.py
vaibhav-research: fixed the issues raised by both of you. whenever you both have sometime, please review and let me know if missed anything. thanks again for your time and feedback. @ArthurZucker @zucchini-nlp

vaibhav-research:
all the test are passing except the glm_image test which is unrelated to this change. I am not sure why but I tested the same locally in main and  PR branch and both were successful .

```
(transformer) vaibhavpandey@Vaibhavs-MacBook-Pro transformers % git switch main
git reset --hard upstream/main
pytest -q tests/models/glm_image/test_modeling_glm_image.py::GlmImageModelTest::test_prompt_lookup_decoding_matches_greedy_search -vv

git switch siglip2_text_embedding
pytest -q tests/models/glm_image/test_modeling_glm_image.py::GlmImageModelTest::test_prompt_lookup_decoding_matches_greedy_search -vv

Already on 'main'
Your branch is ahead of 'origin/main' by 127 commits.
  (use "git push" to publish your local commits)
HEAD is now at 9eea1b0bd3 [`FA`] Generalize fa config checks and fix flags (#43121)
================================================== test session starts ==================================================
platform darwin -- Python 3.11.9, pytest-8.4.2, pluggy-1.6.0 -- /Users/vaibhavpandey/.pyenv/versions/transformer/bin/python
cachedir: .pytest_cache
hypothesis profile 'default'
rootdir: /Users/vaibhavpandey/Projects/transformers
configfile: pyproject.toml
plugins: anyio-4.12.0, xdist-3.8.0, timeout-2.4.0, order-1.3.0, rerunfailures-15.1, asyncio-1.3.0, rich-0.2.0, hypothesis-6.148.8
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
collected 1 item                                                                                                        

tests/models/glm_image/test_modeling_glm_image.py::GlmImageModelTest::test_prompt_lookup_decoding_matches_greedy_search PASSED [100%]

=================================================== 1 passed in 0.94s ===================================================

Switched to branch 'siglip2_text_embedding'
Your branch is up to date with 'origin/siglip2_text_embedding'.
================================================== test session starts ==================================================
platform darwin -- Python 3.11.9, pytest-8.4.2, pluggy-1.6.0 -- /Users/vaibhavpandey/.pyenv/versions/transformer/bin/python
cachedir: .pytest_cache
hypothesis profile 'default'
rootdir: /Users/vaibhavpandey/Projects/transformers
configfile: pyproject.toml
plugins: anyio-4.12.0, xdist-3.8.0, timeout-2.4.0, order-1.3.0, rerunfailures-15.1, asyncio-1.3.0, rich-0.2.0, hypothesis-6.148.8
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
collected 1 item                                                                                                        

tests/models/glm_image/test_modeling_glm_image.py::GlmImageModelTest::test_prompt_lookup_decoding_matches_greedy_search PASSED [100%]

=================================================== 1 passed in 1.07s ==================================================
```

please take a look at the PR for the change it is intended to solve. I am not sure why the CI test is failing on glm_image. sorry to bother you guys again 
@ArthurZucker @zucchini-nlp

src/transformers/models/siglip2/modular_siglip2.py
```diff
@@ -34,10 +35,56 @@
...
+    if isinstance(current, normalizers.Sequence):
+        items = list(current.normalizers)
+        if not any(isinstance(n, normalizers.Lowercase) for n in items):
+            backend.normalizer = normalizers.Sequence([normalizers.Lowercase(), *items])
+        return
+
+    backend.normalizer = normalizers.Sequence([normalizers.Lowercase(), current])
+
+
+class Siglip2Tokenizer(TokenizersBackend):
```
zucchini-nlp: since siglip2 is basically gemma+lowercase, should we not inherit from `GemmaTokenizer` and let modular copy it?

zucchini-nlp:
The failing glm-image tests is supposed to be skipped. I see that there is a typo in test file and it's not being skipped, so feel free to ignore it

src/transformers/models/siglip2/modular_siglip2.py
vaibhav-research: Sure,  I started with TokenizersBackend to keep SigLIP2 decoupled from Gemma, but since the tokenizer behavior is Gemma-compatible, inheriting from GemmaTokenizer makes more sense. Updating the PR.

vaibhav-research: updated as suggested. please give it one more review when you get chance @zucchini-nlp 
thanks

vaibhav-research:
It looks like the PR is failing in the “CircleCI Failure Summary Comment” GitHub Action, not in CircleCI itself.

What I’m seeing:
	•	CircleCI jobs complete successfully for this commit.
	•	The GitHub Action fails while trying to find a collection_job and download failure_summary.json.
	•	In this run, the CircleCI workflow only contains an empty job, so the artifact lookup returns 404 and the script errors (jq: Cannot iterate over null).

I’ve rebased on latest, and the SigLIP2 changes don’t touch CI or CircleCI config, so this seems like an edge case in the failure-summary workflow. not sure, if I missed anything. @zucchini-nlp

vaibhav-research:
hi @zucchini-nlp 
just checking in to see if you have some bandwidth to take another look. I had pushed the fixes and addressed the earlier comments.
Let me know if there’s anything else you’d like me to change. 
Thanks

vaibhav-research:
looks like there is an open issue for t5 test failures, I will try and rebase again when this is fixed. all other tests are passing.
https://github.com/huggingface/transformers/pull/43374

src/transformers/models/siglip2/tokenization_siglip2.py
```diff
@@ -0,0 +1,123 @@
...
+            **kwargs,
+        )
+
+        # Persist for save/load + push_to_hub dynamic tokenizer test
+        if hasattr(self, "init_kwargs") and isinstance(self.init_kwargs, dict):
+            self.init_kwargs.setdefault("tokenizer_class", self.__class__.__name__)
+            self.init_kwargs["do_lower_case"] = do_lower_case
+
+        if do_lower_case:
+            _ensure_lowercase_normalizer(self)
```
zucchini-nlp: we can do easier with below, no if we know that the Gemma has only one normalizer as per L100?

```
normalizers.Sequence([normalizers.Lowercase(), current])
```

vaibhav-research: @zucchini-nlp Thanks for the suggestion.
I’ve removed the helper and simplified the logic to directly wrap the existing Gemma backend normalizer with normalizers.Sequence([normalizers.Lowercase(), current]), relying on the invariant that Gemma has a single normalizer.
I’ve updated both the modular and generated tokenizer files accordingly, and all tests are passing. Please let me know if anything else needs adjustment.

zucchini-nlp (approved):
LGTM, thanks a lot for iterating on it! The failing test is flaky and seems unrelated

zucchini-nlp:
run-slow: siglip2

vaibhav-research:
> LGTM, thanks a lot for iterating on it! The failing test is flaky and seems unrelated

@zucchini-nlp  Thanks a lot for the review and for pointing that out, really appreciate it!
happy to rebase or adjust anything if needed before merging.

zucchini-nlp:
They will be fixed soon, we're having some issues with the new release from torchcodec

src/transformers/models/siglip2/modular_siglip2.py
```diff
@@ -38,6 +40,47 @@
...
+            mask_token=mask_token,
+            **kwargs,
+        )
+
+        # Persist for save/load + push_to_hub dynamic tokenizer test
+        if hasattr(self, "init_kwargs") and isinstance(self.init_kwargs, dict):
+            self.init_kwargs.setdefault("tokenizer_class", self.__class__.__name__)
+            self.init_kwargs["do_lower_case"] = do_lower_case
+
+        if do_lower_case:
```
ArthurZucker: if its the default / always use let's just not have the flag!

src/transformers/models/siglip2/tokenization_siglip2.py
```diff
@@ -0,0 +1,103 @@
...
+            self.init_kwargs["do_lower_case"] = do_lower_case
+
+        if do_lower_case:
+            backend = getattr(self, "_tokenizer", None)
+            if backend is not None and backend.normalizer is not None:
+                backend.normalizer = normalizers.Sequence([normalizers.Lowercase(), backend.normalizer])
+
+    def _unk_id(self) -> int:
+        # Align with historical Siglip2 convention: pad, eos, bos, unk
+        return 3
```
ArthurZucker: Mmm i don't think this should be here at all.

tests/models/siglip2/test_tokenization_siglip2.py
ArthurZucker, line 1: the only test we want here is an integration test

src/transformers/models/siglip2/modular_siglip2.py
vaibhav-research: makes sense, I will remove the same.

src/transformers/models/siglip2/tokenization_siglip2.py
vaibhav-research: hmm, yeah this _unk_id() is not something I explicitly added. it’s currently being emitted by the modular converter when flattening GemmaTokenizer into the standalone tokenizer.
That said, I agree it doesn’t make sense to hard-code UNK behavior here and SigLIP2 shouldn’t diverge from standard TokenizersBackend handling. I’ll update the modular to avoid inheriting this and remove _unk_id() entirely from the generated tokenizer.

tests/models/siglip2/test_tokenization_siglip2.py
vaibhav-research: sure, I’ll simplify this to a single integration test that validates e2e behavior and drop the granular checks.

src/transformers/models/siglip2/tokenization_siglip2.py
vaibhav-research: hi @ArthurZucker 

need your input here please. as mentioned earlier _unk_id() isn’t explicitly added. it’s coming from [GemmaTokenizer,](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma/tokenization_gemma.py#L105) and the modular converter currently flattens base tokenizer methods into the generated file.

To address this, I see three options:
	1.	Keep the current inheritance from GemmaTokenizer and accept the inherited _unk_id().
	2.	Keep inheriting from GemmaTokenizer, but change the modular converter to avoid inlining _unk_id() into the generated tokenizer (larger change to the converter, but keeps SigLIP2 implementation minimal).
	3.	Stop inheriting from GemmaTokenizer and instead subclass TokenizersBackend directly in SigLIP2 (so no Gemma specific overrides like _unk_id() get pulled in), while still reusing the same backend construction/normalizer behavior.
	
awaiting your response.

vaibhav-research: also,  quick update @ArthurZucker  
after looking at modular_model_converter.py: there’s currently no per-class mechanism to tell the converter “don’t emit method X” (e.g. no _modular_exclude_methods support).
So option (2) would mean a converter change to introduce such an exclusion rule (bigger / global change). I believe we don't want that here.
Given that, the practical choices seem to be:
	1.	keep GemmaTokenizer inheritance and accept _unk_id being emitted (even though SigLIP2 doesn’t conceptually need it), or
	2.	switch Siglip2Tokenizer modular to subclass TokenizersBackend directly (copy Gemma backend init + add lowercase), which will avoid _unk_id entirely without touching the converter.
Which direction would you prefer?

ArthurZucker: there is you can doL
```python 
def _unk_id(self) -> int:
    raise ValueError("Not needed") 
```

vaibhav-research: Oh ok, sure i will add and push the changes 
Thanks

tests/models/siglip2/test_tokenization_siglip2.py
vaibhav-research: @ArthurZucker I have updated the test and included integration test only

src/transformers/models/siglip2/modular_siglip2.py
vaibhav-research: @ArthurZucker I have updated and removed the flag.

src/transformers/models/siglip2/tokenization_siglip2.py
vaibhav-research: @ArthurZucker Updated Siglip2Tokenizer to explicitly override _unk_id() and raise ValueError, so the generated tokenizer no longer inherits Gemma’s _unk_id behavior.

vaibhav-research:
@ArthurZucker thanks for your feedback, updated the PR based on your suggestions. Please take a look at it again when you get a chance.
Thanks

docs/source/en/model_doc/siglip2.md
```diff
@@ -145,7 +145,75 @@ logits_per_image = outputs.logits_per_image
...
+
+```
+### Text-only usage: Siglip2Tokenizer
+
+If you are encoding text without a processor (for example, via [`AutoTokenizer`]), use [`Siglip2Tokenizer`].
+`Siglip2Tokenizer` applies **lowercasing** at the tokenizer backend level (matching SigLIP2 training-time normalization),
+while keeping the same vocabulary/tokenization as the original tokenizer.
+
+- Lowercasing is handled via tokenizer normalizer
+- Padding/truncation defaults are handled by processor kwargs (`Siglip2ProcessorKwargs`)
```
zucchini-nlp: nit: when loading a tokenizer we don't have any `Siglip2ProcessorKwargs` anymore

zucchini-nlp (approved):
One tiny comment in docs, to be precise 😄

docs/source/en/model_doc/siglip2.md
zucchini-nlp: btw, can you also update the conversion script with the new tokenizer?

vaibhav-research: @zucchini-nlp sure,  I’ll adjust the docs to clarify  lowercasing comes from the tokenizer backend normalizer, and the padding/truncation defaults (max_length=64) come from Siglip2Processor defaults only when using the processor (not when loading the tokenizer alone).
Also yes,  I’ll update the SigLIP2 conversion script to save tokenizer artifacts/config with `tokenizer_class=Siglip2Tokenizer` so converted checkpoints resolve correctly.

vaibhav-research: @zucchini-nlp  I have updated the conversation script and doc.

vaibhav-research:
Thanks a ton @zucchini-nlp  and @ArthurZucker I really appreciate the patience and all the helpful nudges along the way 🙏
I’ve pushed the doc tweak and updated the conversion script to use the new Siglip2Tokenizer as suggested. 
Whenever you get a chance, would love another look. Thanks again for making this smooth (and for putting up with my many follow-ups 😄)

zucchini-nlp:
Thanks, we can merge imo. The CI is not feeling well since yesterday so let's wait a bit until it's fixed

vaibhav-research:
> Thanks, we can merge imo. The CI is not feeling well since yesterday so let's wait a bit until it's fixed

Thanks @zucchini-nlp 🙏 , will wait for CI to recover. I’ll keep an eye on the checks.

src/transformers/models/siglip2/modular_siglip2.py
```diff
@@ -39,6 +41,45 @@
...
+        # Persist for save/load + push_to_hub dynamic tokenizer test
+        if hasattr(self, "init_kwargs") and isinstance(self.init_kwargs, dict):
+            self.init_kwargs.setdefault("tokenizer_class", self.__class__.__name__)
+
+        backend = getattr(self, "_tokenizer", None)
+        if backend is not None and backend.normalizer is not None:
+            backend.normalizer = normalizers.Sequence([normalizers.Lowercase(), backend.normalizer])
+
+    def _unk_id(self) -> int:
+        raise ValueError("_unk_id is not needed for SigLIP2.")
```
ArthurZucker:
```suggestion
        raise AttributeError("_unk_id is not needed for SigLIP2.")
```

ArthurZucker (commented):
the attribute error should normally prevent the function to be written

src/transformers/models/siglip2/modular_siglip2.py
ArthurZucker: Ah shit it does not work one sec

vaibhav-research:
thanks @ArthurZucker for fixing the last issue and for merging the PR

ArthurZucker:
for sure, sorry for the delay

src/transformers/models/siglip2/tokenization_siglip2.py
```diff
@@ -0,0 +1,95 @@
...
+            **kwargs,
+        )
+
+        # Persist for save/load + push_to_hub dynamic tokenizer test
+        if hasattr(self, "init_kwargs") and isinstance(self.init_kwargs, dict):
+            self.init_kwargs.setdefault("tokenizer_class", self.__class__.__name__)
+
+        backend = getattr(self, "_tokenizer", None)
+        if backend is not None and backend.normalizer is not None:
+            backend.normalizer = normalizers.Sequence([normalizers.Lowercase(), backend.normalizer])
```
ArthurZucker, line 92: we should remove that tho

PR #43102, opened 2026-01-04
Add CPU vs GPU performance comparison example

kowshik-thatinati:
Added a section comparing inference speed on CPU vs GPU using Hugging Face pipeline and Accelerator.


kowshik-thatinati:
Hi 
I actually added a section comparing inference speed on CPU vs GPU using Hugging Face pipeline and Accelerator.

stevhliu (commented):
hey, thanks but i don't think its necessary to show multilingual text generation and ner as the goal is to keep this doc more general. likewise, not necessary to compare cpu/gpu performance

PR #43104, opened 2026-01-05
docs: clarify tokenizer decoder behavior in v5 (#43066)

### LINKED ISSUE #43066: Wrong tokenizer decoder type in Transformers v5
### System Info

Wrong decoder type with `5.0.0rc1`.

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

Run this:
```python

tokenizer = AutoTokenizer.from_pretrained("deepseek-ai/DeepSeek-R1-Distill-Llama-8B")
print(tokenizer.decoder)
```

You get:
```
Sequence(decoders=[Replace(pattern=String("▁"), content=" "), ByteFallback(), Fuse(), Strip(content=" ", start=1, stop=0)])
```

with Transformers v5.

And with Transformers 4.57.3 and earlier you get:

```
ByteLevel(add_prefix_space=True, trim_offsets=True, use_regex=True)
```

Is it expected that this changed?

### Expected behavior

The same decoder type as transformers 4.57.3

kowshik-thatinati:
Hey team 👋

Just adding a small clarification regarding the `decoder` property in AutoTokenizer for Transformers v5.

So basically, in v5, the decoder object might look different compared to v4. 
This is because the internal tokenizer backend has been redesigned. 
But don’t worry — the output of `tokenizer.decode(...)` stays the same, so functionality isn’t affected.

What this PR does:
1. Adds a comment in `tokenization_utils_base.py` near the `decoder` property for reference.
2. Adds a note in the docs (`tokenizer_notes.rst`) to explain this behavior for users.

Hope this helps avoid confusion for anyone upgrading to v5 🙂  

Closes #43066


kowshik-thatinati:
Hi 👋  
This PR only adds a clarification in docs and a code comment explaining the tokenizer decoder behavior in v5.  
No functional changes involved.

Please let me know if any changes are needed. Thanks!

ArthurZucker (commented):
hey! sorry there seems to be a bit too many changes for the fix!

kowshik-thatinati:
Hey, thanks for pointing that out 🙏
You’re right — the fix can be much smaller.

I’ll reduce this PR to only the minimal clarification needed (no extra docs / changes) and update it accordingly.
Thanks for the review!

PR #43109, opened 2026-01-05
Fix warnings poping up with `fixup` and `pytest`

Cyrilvallez:
# What does this PR do?

As per the title. I'm getting too annoyed to see them all the time, high-time to fix them

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -407,7 +407,6 @@ setup(
         "Intended Audience :: Developers",
         "Intended Audience :: Education",
         "Intended Audience :: Science/Research",
-        "License :: OSI Approved :: Apache Software License",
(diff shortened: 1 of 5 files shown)


ArthurZucker (approved):

setup.py
```diff
@@ -407,7 +407,6 @@ def run(self):
         "Intended Audience :: Developers",
         "Intended Audience :: Education",
         "Intended Audience :: Science/Research",
-        "License :: OSI Approved :: Apache Software License",
```
Cyrilvallez, line 410: Deprecated, https://peps.python.org/pep-0639/#deprecate-license-classifiers

PR #43111, opened 2026-01-05
Fix test_all_tensors_are_parameter_or_buffer

remi-or:
# What does this PR do?

This PR fixes the test `test_all_tensors_are_parameter_or_buffer` across 5 models, effectively making the test pass for all models. 

## Failure cases and fixes

There are 2 reasons this test used to fail:
- the model has some triton code which crash when we try to do a forward pass on the CPU:
To address this, we add a keyword arg `non_meta_device` which is the device the model is transferred to. It defaults to `cpu`, so the default test behavior does not change. If set to `cuda`, the model is transferred to the GPU and the triton code can run.
- the model has inputs with nested tensors (list, dict, or objects with tensor attibutes):
 To address this, we add the `transfer_nested_tensors` flag, turned off by default so the default test behavior does not change. If the flag is passed, we go through inputs and try to transfer nested tensors whether they are part of a list, tuple, dict or plain attributes. 

## Models affected

No modeling code was changed, but the test was overloaded with kwargs in the following models:
- layoutlmv2 (the `images` input has a nested tensor attribute)
- modernbert (cannot run on cpu because of triton)
- pe_audio (cannot run on cpu because of triton)
- pe_video (cannot run on cpu because of triton)
- perceiver (the model has nested tensors in its inputs)

### DIFF
--- a/tests/test_modeling_common.py
+++ b/tests/test_modeling_common.py
@@ -1286,16 +1286,11 @@ class ModelTesterMixin:
                 f"them correctly if the model is on meta device):\n{unique_bad_module_traceback}",
             )
 
-    def test_all_tensors_are_parameter_or_buffer(
-        self,
-        non_meta_device: str = "cpu",
-        transfer_nested_tensors: bool = False,
-    ) -> None:
+    def test_all_tensors_are_parameter_or_buffer(self) -> None:
         """Check that all tensors are registered as Parameter or Buffer, i.e. we don't have simple assignments such
         as `self.x = torch.tensor(...)` in a Module (as we cannot correctly recover from meta device if it's not
         registered as parameter/buffer). To test this, we initialize the model on a meta device and then move it onto
-        the non_meta_device (cpu by default) and perform a forward pass. Since some models like modernbert cannot run on
-        CPU, we can specify a different device to test on."""
+        the torch_device and perform a forward pass."""
         config, inputs_dict = self.model_tester.prepare_config_and_inputs_for_common()
 
         for model_class in self.all_model_classes:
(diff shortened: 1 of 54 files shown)


Cyrilvallez:
Nice, thanks for catching those! Instead of adding the keyword args, we can simply do `model.to(device=torch_device)` I believe, which is the default auto-detected device on which the tests are run. That would avoid overiding for the 5 models.
The logic of the test is not impacted by the device being cpu or cuda or whatever else, we only want to make sure it won't be meta!
Should for sure make the trick instead of `non_meta_device`. For `transfer_nested_tensors` I'm unsure, but I assume those models should prepare their nested inputs on the correct device (`torch_device`) by default as well, making it redundant as well?

remi-or:
Using `torch_device` is a much better solution tbh! Still leaving 3 overloads for the triton-related failures because they wont work on CPU, so they need the decorator `require_torch_gpu`

tests/test_modeling_common.py
```diff
@@ -1286,29 +1286,30 @@ def test_init_weights_can_init_buffers(self):
...
+            model.to_empty(device=torch_device)
             # Now, run all the inits
             model.init_weights()
 
             # Prepare inputs
             inputs = self._prepare_for_class(inputs_dict, model_class)
-            # Inputs may be on cuda -> move to cpu, we don't care about accelerator for this test
-            inputs = {k: v.to("cpu") if isinstance(v, torch.Tensor) else v for k, v in inputs.items()}
+            # Inputs may be on cuda -> move to torch_device, we usually don't care about accelerator for this test
+            inputs = {k: v.to(torch_device) if isinstance(v, torch.Tensor) else v for k, v in inputs.items()}
```
Cyrilvallez: This should already be on torch_device, let's remove! (it's the reason we had to move it explicitly to cpu haha)

Cyrilvallez (approved):

tests/test_modeling_common.py
remi-or: good catch!

PR #42151, opened 2025-11-11
SINQ quantization strategy integration

ChiaraBoretti:
# What does this PR do?
**Summary**
This PR introduces an integration of the [SINQ](https://huggingface.co/papers/2509.22944) quantization method into the Hugging Face Transformers library. It follows the pattern of existing quantization integrations such as HQQ, AWQ, and others.
The goal is to enable users to configure and apply SINQ quantization directly through model configuration parameters, which can then be passed via the ```from_pretrained()``` function within the Transformers framework.

**Example Usage**
The pipeline then will be the following:
```bash
cfg = SinqConfig(
    nbits=4,
    group_size=64,
    tiling_mode="1D",
    method="sinq",     
    modules_to_not_convert=["lm_head"],
    device="cuda:1"
)

model_name = "Qwen/Qwen3-1.7B"
tok = AutoTokenizer.from_pretrained(model_name)

qmodel = AutoModelForCausalLM.from_pretrained(
    model_name,
    cache_dir=cache_dir, 
    torch_dtype=torch.bfloat16,
    quantization_config=cfg,
)
```

**Model saving and loading**
Once the model has been quantized, it can be saved locally or pushed to the Hugging Face Hub.
Moreover with this integration it is also possible to download a sinq quantized model directly from the hub and use it in inference.
```bash
hf_path = "HF_hub/path"
tok  = AutoTokenizer.from_pretrained(hf_path)
qmodel = AutoModelForCausalLM.from_pretrained(hf_path, device_map="cuda:0")
```
**Installation**
All the required dependencies are installed by installing the official SINQ python package.
```bash
pip install sinq
```

Fixes #42116

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [x] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [x] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
[...]

PR #43112, opened 2026-01-05
SINQ quantization strategy integration (adapted for Transformers V5)

ChiaraBoretti:
# What does this PR do?
**Summary**
This PR introduces an integration of the [SINQ](https://huggingface.co/papers/2509.22944) quantization method into the Hugging Face Transformers library. It follows the pattern of existing quantization integrations such as HQQ, AWQ, and others.
The goal is to enable users to configure and apply SINQ quantization directly through model configuration parameters, which can then be passed via the ```from_pretrained()``` function within the Transformers framework.

**Example Usage**
The pipeline then will be the following:
```bash
cfg = SinqConfig(
    nbits=4,
    group_size=64,
    tiling_mode="1D",
    method="sinq",     
    modules_to_not_convert=["lm_head"],
    device="cuda:1"
)

model_name = "Qwen/Qwen3-1.7B"
tok = AutoTokenizer.from_pretrained(model_name)

qmodel = AutoModelForCausalLM.from_pretrained(
    model_name,
    cache_dir=cache_dir, 
    torch_dtype=torch.bfloat16,
    quantization_config=cfg,
)
```

**Model saving and loading**
Once the model has been quantized, it can be saved locally or pushed to the Hugging Face Hub.
```bash
from sinq.hf_io import patch_hf_pretrained_io
patch_hf_pretrained_io()

hf_path = "HF_hub/path"
qmodel.save_pretrained(local_path)
qmodel.push_to_hub(hf_path, safe_serialization=True)

tok  = AutoTokenizer.from_pretrained(hf_path)
qmodel = AutoModelForCausalLM.from_pretrained(hf_path, device_map="cuda:0")
```
**Installation**
All the required dependencies are installed by installing SINQ from the official github repository.

This pull request is a follow-up to #42151. I open a new PR to improve clarity and to keep this new integration adaptation separate. Please refer to this pull request and does not consider the previous one.
Fixes #42116

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @MekkCyber


docs/source/en/quantization/sinq.md
```diff
@@ -0,0 +1,182 @@
...
+Quantizing any 🤗 Hugging Face model with SINQ is simple and takes only a few lines of code. 
+First, create a [`SinqConfig`] and specify the following parameters:
+
+| Flag | Description | Type | Options | Default |
+|------|-------------|---------|---------|----------|
+| `--nbits` | Bit-width for weight quantization | int | 2, 3, 4, 5, 6, 8 | 4 |
+| `--tiling_mode` | Weight matrix tiling strategy | str | 1D, 2D | 1D |
+| `--group_size` | Weights per quantization group | int | 64, 128 | 64 |
+| `--method` | Quantization method | str | sinq, asinq | sinq |
+| `--dtype` | Data type of the original model | str | auto, float16, float32 | auto (bfloat16) |
```
SunMarc: It should be fine to remove this as we should already have this information when calling `from_pretrained`

docs/source/en/quantization/sinq.md
```diff
@@ -0,0 +1,182 @@
...
+| `--modules_to_not_convert` | List of the layers that are NOT quantize | List of str | [lm_head, ...] | [lm_head] |
+| `--device` | Device on which the model is loaded | str | cpu, cuda:0, cuda:1, etc | cuda:0 |
```
SunMarc: same for this one, the user can just pass `device_map` in `from_pretrained`.

docs/source/en/quantization/sinq.md
```diff
@@ -0,0 +1,182 @@
...
+    method="sinq",
+    modules_to_not_convert=["lm_head"],
+    device=device
+)
+
+tok = AutoTokenizer.from_pretrained(model_name)
+qmodel = AutoModelForCausalLM.from_pretrained(
+    model_name,
+    quantization_config=cfg,
+    torch_dtype=torch.bfloat16
```
SunMarc: `torch_dtype` is deprecated in favor of `dtype` 
```suggestion
    dtype=torch.bfloat16
```

docs/source/en/quantization/sinq.md
```diff
@@ -0,0 +1,182 @@
...
+### Save & reload
+
+If you want to reuse a quantized model later, save it to disk or push it on the HuggingFace Hub and reload it without needing base FP weights.
+If you installed SINQ from source you should call *patch_hf_pretrained_io* function:
+```python
+from sinq.hf_io import patch_hf_pretrained_io
+patch_hf_pretrained_io()
+# Save sinq quantized model
+model.save_pretrained("/path/to/save/qwen3-1.7B-sinq-4bit")
+model.push_to_hub("HF_Hub_username/qwen3-1.7B-sinq-4bit", safe_serialization=True, private=True)
```
SunMarc:
```suggestion
model.push_to_hub("HF_Hub_username/qwen3-1.7B-sinq-4bit")
```

docs/source/en/quantization/sinq.md
```diff
@@ -0,0 +1,182 @@
...
+Otherwise, if you installed SINQ through pip, you can simply use HF built-in functions:
+
+```python
+# --- Save to a folder (sharded safetensors) ---
+
+# 'model' must already be SINQ-quantized
+# Locally save
+qmodel.save_pretrained("/path/to/save/qwen3-1.7B-sinq-4bit")
+# Push to the Hub
+qmodel.push_to_hub("HF_Hub_username/qwen3-1.7B-sinq-4bit", safe_serialization=True)
```
SunMarc:
```suggestion
# Push to the Hub
qmodel.push_to_hub("HF_Hub_username/qwen3-1.7B-sinq-4bit")
```

docs/source/en/quantization/sinq.md
```diff
@@ -0,0 +1,182 @@
...
+
+---
+
+### Save & reload
+
+If you want to reuse a quantized model later, save it to disk or push it on the HuggingFace Hub and reload it without needing base FP weights.
+If you installed SINQ from source you should call *patch_hf_pretrained_io* function:
+```python
+from sinq.hf_io import patch_hf_pretrained_io
+patch_hf_pretrained_io()
```
SunMarc: instead of asking users to use this, we can move this code into hf_quantizer

docs/source/en/quantization/sinq.md
```diff
@@ -0,0 +1,182 @@
...
+      eprint={2509.22944},
+      archivePrefix={arXiv},
+      primaryClass={cs.LG},
+      url={http://arxiv.org/abs/2509.22944}
+}
+```
+
+## 3. Current Limitations
+
+At the moment the SINQ quantization strategy and SINQ quantized models do not support Multi-GPU option, so if your system counts multiple GPUs please specify which one should be used.
```
SunMarc, line 182: it doesn't work with `device_map` ?

doctest_list.txt
```diff
@@ -0,0 +1 @@
+docs/source/en/chat_response_parsing.md docs/source/en/continuous_batching.md docs/source/en/model_doc/colqwen2.md docs/source/en/model_doc/fuyu.md docs/source/en/quantization/sinq.md docs/source/en/serving.md docs/source/en/trainer.md src/transformers/models/auto/feature_extraction_auto.py src/transformers/models/auto/image_processing_auto.py src/transformers/models/fuyu/image_processing_fuyu_fast.py src/transformers/models/glpn/image_processing_glpn_fast.py src/transformers/models/lightglue/image_processing_lightglue_fast.py src/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py src/transformers/models/qwen2_5_vl/modular_qwen2_5_vl.py src/transformers/models/qwen2_vl/modeling_qwen2_vl.py src/transformers/models/qwen3_vl/modeling_qwen3_vl.py src/transformers/models/qwen3_vl/modular_qwen3_vl.py src/transformers/pipelines/deprecated/text2text_generation.py src/transformers/quantizers/quantizer_sinq.py src/transformers/utils/chat_parsing_utils.py
```
SunMarc: to remove

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,227 @@
+# src/transformers/integrations/sinq.py
```
SunMarc: add license 
```suggestion
```

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,227 @@
...
+            logger.warning_once(
+                "SinqQuantize.convert called without `model` or `full_layer_name`; "
+                "skipping SINQ quantization for this parameter."
+            )
+            key, values = next(iter(input_dict.items()))
+            val = values[0] if isinstance(values, list) else values
+            return {full_layer_name or key: val}
+
+        _, values = next(iter(input_dict.items()))
+        weight_tensor = values[0] if isinstance(values, list) else values
```
SunMarc: is it really needed ? In which cases it happens ?

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,227 @@
...
+
+        module, tensor_name = get_module_from_name(model, full_layer_name)
+        if tensor_name != "weight":
+            logger.warning_once(
+                f"SinqQuantize.convert called for non-weight parameter: {full_layer_name}; "
+                "treating as normal param."
+            )
+            return {full_layer_name: weight_tensor}
```
SunMarc: same for this one, I don't think we need this. It means that `param_needs_quantization` was not correctly coded.

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,227 @@
...
+        module_path, _, _ = full_layer_name.rpartition(".") 
+        parent_path, _, child_name = module_path.rpartition(".")
+        parent = model.get_submodule(parent_path) if parent_path else model
+
+        from sinq.sinqlinear import SINQLinear
+        from sinq.sinqlinear import sinq_base_quant_config as sinq_base_quant_config_fn
+
+        device_str = self._get_runtime_device_str()
+        device = torch.device(device_str)
+        compute_dtype = self.hf_quantizer.update_torch_dtype(None)
```
SunMarc: you can just use the device of the tensor as it should be loaded on the correct one, same for the dtype.

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,227 @@
...
+        sinq_layer = SINQLinear(
+            linear_layer=dense,
+            quant_config=sinq_quant_dict,
+            del_orig=True,
+            compute_dtype=compute_dtype,
+            device=device_str,
+            use_unpack_kernel=True,
+            layer_activations=None,
+        )
+
```
SunMarc: Is there really no other way to make this work ? I see that you took a lot of inspiration from hqq code but it's a nightmare to actually maintain it. Right now, even our hqq integration is broken due to v5 and I don't know if we will ever fix it. Part of the reason is that the api is too complicated / hard to maintain. Is it possible to think of a simpler api like we have for fp8/mxfp4 where we just replace the linear beforehand and then we just correctly set the params/buffer here ?

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,227 @@
...
+        if model is None or full_layer_name is None:
+            logger.warning_once(
+                "SinqDeserialize.convert called without `model` or `full_layer_name`; "
+                "returning raw tensors (no SINQ module reconstruction)."
+            )
+            out: Dict[str, "torch.Tensor"] = {}
+            for k, v in input_dict.items():
+                out[k] = v[0] if isinstance(v, list) else v
+            return out
```
SunMarc: probably not needed same as above

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,227 @@
...
+
+        if W_q is None or meta is None:
+            logger.warning_once(
+                "SinqDeserialize.convert did not find '.W_q' and '.meta' entries; "
+                "treating as normal (non-SINQ) parameter."
+            )
+            v = next(iter(input_dict.values()))
+            if isinstance(v, list):
+                v = v[0]
+            return {full_layer_name: v}
```
SunMarc, line 159: If there is indeed only one value, we should just return that as it means that we didn't collected sinq parameters as expected. We should also remove the warning. This is what we do for finegrained_fp8

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,227 @@
...
+
+        device_str = self._get_runtime_device_str()
+        device = torch.device(device_str)
+        compute_dtype = self.hf_quantizer.update_torch_dtype(None)
```
SunMarc: same as above, we shouldn't need this

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,227 @@
...
+            "W_q": W_q.to(device),
+            "meta": meta,
+        }
+        if bias is not None:
+            state["bias"] = bias.to(device)
+
+        sinq_layer.load_state_dict(state)
+
+        setattr(parent, child_name, sinq_layer)
+        sinq_layer._is_hf_initialized = True
```
SunMarc: Same as above, let's see if there is a way to simplify this.

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,227 @@
...
+        device = torch.device(device_str)
+        compute_dtype = self.hf_quantizer.update_torch_dtype(None)
+
+        module, tensor_name = get_module_from_name(model, full_layer_name)
+        if tensor_name != "weight":
+            logger.warning_once(
+                f"SinqDeserialize called for non-weight parameter: {full_layer_name}. "
+                "Skipping SINQ reconstruction."
+            )
+            return {full_layer_name: W_q}
```
SunMarc: no needed potentially !

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+          * weight-only SINQ
+          * param-level ConversionOps (`SinqQuantize`) during load for pure language models
+            (each Linear.weight is turned into a SINQLinear module)
+          * module-level quantization after load for multimodal models
+      - method="asinq":
+          * A-SINQ (activation-aware) SINQ quantization
+    """
+
+    requires_calibration: bool = False  # A-SINQ does its own internal calibration
+    required_packages: List[str] = ["sinq"]
```
SunMarc: we don't use this anymore 
```suggestion
```

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+    requires_parameters_quantization: bool = True
+
+    def __init__(self, quantization_config: SinqConfig, **kwargs):
+        super().__init__(quantization_config, **kwargs)
+        self._normalized_device_str: str | None = None
+        self.modules_to_not_convert: list[str] = []
```
SunMarc: not needed 
```suggestion
```

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+            - safe_serialization=False -> ok
+        """
+        if safe_serialization:
+            logger.warning(
+                "SINQ quantized models do not currently support `safe_serialization=True` "
+                "(safetensors) because SINQLinear.meta is a nested dict with non-tensor "
+                "entries. Please save with `safe_serialization=False` for now."
+            )
+            return False
+        return True
```
SunMarc: We don't do unsafe serialization anymore. Until sinq supports it, let's not enable serialization.

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+    def is_trainable(self) -> bool:
+        return True
+
+    def validate_environment(self, *args, **kwargs) -> None:
+        _import_sinq()
+
+        cfg: SinqConfig = self.quantization_config
+
+        if getattr(cfg, "dtype", "auto") not in ("auto", "bfloat16", "float16", "float32"):
+            raise ValueError(f"Unsupported dtype in SinqConfig: {cfg.dtype}")
```
SunMarc: let's do this check in `update_dtype` method instead

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+        if not torch.cuda.is_available():
+            raise RuntimeError("SINQ currently expects a CUDA device (for GemLite backend).")
+
+        device_map = kwargs.get("device_map", None)
+        device_str = _normalize_cuda_device(getattr(cfg, "device", "auto"))
+        _validate_cuda_device_str(device_str)
+        self._normalized_device_str = device_str
+
+        if device_str.startswith("cuda"):
+            torch.cuda.set_device(torch.device(device_str))
```
SunMarc: I prefer not to do that if possible. As I said somewhere, you can directly use the device of each loaded tensors

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+                    "SinqHfQuantizer: multi-GPU device_map detected, but SINQ currently supports only a single CUDA "
+                    f"device. Got {sorted(devs)}. Please use device_map=None and set SinqConfig(device='{device_str}')."
+                )
+            only = next(iter(devs))
+            if only != device_str and not (only == "cuda" and device_str == "cuda:0"):
+                raise RuntimeError(
+                    "SinqHfQuantizer: device_map device and SinqConfig.device disagree. "
+                    f"device_map={only!r}, SinqConfig.device={device_str!r}. "
+                    "Please pick one device and keep them consistent."
+                )
```
SunMarc: let's simplify this also, this is way too complex to just check values inside the `device_map`.

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+
+    def update_torch_dtype(self, torch_dtype: Optional[torch.dtype]) -> torch.dtype:
+        """
+        Decide compute dtype for quantized layers.
+        """
+        cfg: SinqConfig = self.quantization_config
+        if cfg.dtype == "auto":
+            return torch.bfloat16 if is_torch_bf16_gpu_available() else torch.float16
+        return getattr(torch, cfg.dtype)
```
SunMarc: let's not do that and just check if the torch_dtype passed is correct or not.

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+
+    # keep old name for BC
+    def update_dtype(self, dtype: Optional[torch.dtype]) -> torch.dtype:
+        return self.update_torch_dtype(dtype)
```
SunMarc: we don't to care about BC for `update_torch_dtype`, let's just use `update_dtype`

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+
+    def update_device_map(self, device_map):
+        # We already enforced single CUDA device in validate_environment;
+        # nothing else to do, let HF / accelerate handle placement.
+        return device_map
```
SunMarc: remove

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+
+    from sinq.sinqlinear import SINQLinear, sinq_base_quant_config as sinq_base_quant_config_fn
+    from sinq.awq import (
+        get_simple_calibration_data as awq_get_simple_calibration_data,
+        get_calib_dataset as awq_get_calib_dataset,
+        collect_activations as awq_collect_activations,
+        collect_activations_blockwise as awq_collect_activations_blockwise
+    )
+
+    _SINQ_IMPORTED = True
```
SunMarc: let's import the relevant lib in the corresponding method and not buckle everything here

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+    the quantization config dict.
+    """
+    global _SINQ_IMPORTED, SINQLinear, sinq_base_quant_config_fn
+    global awq_get_simple_calibration_data, awq_get_calib_dataset, awq_collect_activations, awq_collect_activations_blockwise
+
+    if _SINQ_IMPORTED:
+        return
+
+    if importlib.util.find_spec("sinq") is None:
+        raise ImportError("The 'sinq' package is not installed. Please `pip install sinq` to use SinqConfig.")
```
SunMarc: this should happen in validate_environment. Can you create a `is_sinq_available` method like the other quant methods ?

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+    def _should_skip(self, name: str) -> bool:
+        """
+        `name` can be a module path ("model.layers.0.self_attn.q_proj")
+        or a param path ("model.layers.0.self_attn.q_proj.weight").
+        We always compare against the *module* part.
+        """
+        if name.endswith(".weight") or name.endswith(".bias"):
+            name = name.rsplit(".", 1)[0]
+
+        return any(name == m or name.startswith(m + ".") for m in self.modules_to_not_convert)
```
SunMarc: We already have a utility function called `should_convert_module` . You can just reuse this

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+    def param_needs_quantization(self, model: "PreTrainedModel", param_name: str, **kwargs) -> bool:
+        """
+        Called per-parameter to decide whether to run `SinqQuantize` on it.
+
+        - If `self.pre_quantized`, we do *not* quantize again (handled by SinqDeserialize instead).
+        - For method="asinq": return False (ASINQ is module-level post-load).
+        - For method="sinq": True only for nn.Linear.weight not in modules_to_not_convert.
+        """
+        cfg: SinqConfig = self.quantization_config
+        method = str(getattr(cfg, "method", "sinq")).lower()
```
SunMarc: we don't need a getattr + lower() , `cfg` should always have a `method` and the value is either `sinq` or `asinq`. Also just use `self.quantization_config`

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+
+        if self.pre_quantized:
+            return False
+
+        if method == "asinq":
+            return False
```
SunMarc: why ?

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+            from ..integrations.sinq import SinqDeserialize
+
+            return [
+                WeightConverter(
+                    source_patterns=[
+                        ".W_q",
+                        ".meta",
+                        ".bias",
+                    ],
+                    target_patterns=[".weight", ".weight", ".weight"],
```
SunMarc: we only need to pass target_patterns=".weight"

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+        """
+        cfg: SinqConfig = self.quantization_config
+        self.modules_to_not_convert = self.get_modules_to_not_convert(
+            model, cfg.modules_to_not_convert, keep_in_fp32_modules
+        )
+
+        method = str(getattr(cfg, "method", "sinq")).lower()
+
+        # Allow user override via extra kwargs if you want (optional)
+        force = getattr(cfg, "force_param_level", None)  # if you add it
```
SunMarc: simplify

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+        Called after *all* weights have been loaded.
+
+        - For method="sinq": nothing to do, param-level quantization already
+          replaced Linear modules with SINQLinear during load.
+
+        - For method="asinq": run internal AWQ-like calibration and then replace
+          Linear modules with SINQLinear, using layer activations.
+        """
+        cfg: SinqConfig = self.quantization_config
+        method = str(getattr(cfg, "method", "sinq")).lower()
```
SunMarc: simplify

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+        if method == "asinq" and not self.pre_quantized:
+            self._run_asinq_calibration_and_replace(model, **kwargs)
+
+        if method == "sinq" and not self.pre_quantized and not self._do_param_level_sinq:
+            self._replace_linear_with_sinqlinear(model)
+        
+        try:
+            setattr(model, "_is_sinq_quantized", True)
+        except Exception:
+            pass
```
SunMarc: don't think this is needed

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+
+        - For method="asinq": run internal AWQ-like calibration and then replace
+          Linear modules with SINQLinear, using layer activations.
+        """
+        cfg: SinqConfig = self.quantization_config
+        method = str(getattr(cfg, "method", "sinq")).lower()
+
+        if method == "asinq" and not self.pre_quantized:
+            self._run_asinq_calibration_and_replace(model, **kwargs)
+
```
SunMarc: I think it will be better to leave methods that requires calibration to sinq lib. We prefer not to support these methods and usually ask users to use directly the original lib e.g. AutoAWQ or or AutoRound

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+        if method == "sinq" and not self.pre_quantized and not self._do_param_level_sinq:
+            self._replace_linear_with_sinqlinear(model)
+
```
SunMarc: I don't understand why some some arch, it is preferable to do all at once. It should be the same from what I see in the code. If we have to do all at once, it's might better to ask users to use sinq lib to quantize.

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+            if not isinstance(module, nn.Linear):
+                continue
+            if self._should_skip(full_name):
+                continue
+
+            parent_path, _, child_name = full_name.rpartition(".")
+            parent = model.get_submodule(parent_path) if parent_path else model
+
+            module = module.to(device=device, dtype=compute_dtype)
+
```
SunMarc: we can simplify a bunch of things

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,579 @@
...
+                del_orig=True,
+                compute_dtype=compute_dtype,
+                device=device_str,
+                use_unpack_kernel=True,
+                layer_activations=None,
+            )
+            setattr(parent, child_name, sinq_layer)
+            sinq_layer._is_hf_initialized = True
+
+    def _run_asinq_calibration_and_replace(self, model: "PreTrainedModel", **kwargs):
```
SunMarc: as said before, i prefer to redirect users to sinq lib if they want to calibrate models

src/transformers/utils/quantization_config.py
```diff
@@ -2092,3 +2093,142 @@ def to_dict(self) -> dict[str, Any]:
...
+        nbits: int = 4,
+        group_size: int = 64,
+        tiling_mode: str = "1D",
+        method: str = "sinq",  # "sinq" | "asinq"
+        per_channel: bool = True,
+        symmetric: bool = True,
+        use_nf4: bool = False,
+        modules_to_not_convert: Optional[List[str]] = None,
+        dtype: Optional[str] = "auto",  # "auto" | "bfloat16" | "float16" | "float32"
+        device: Optional[Union[str, int]] = "cpu",
```
SunMarc: let's remove this

src/transformers/utils/quantization_config.py
```diff
@@ -2092,3 +2093,142 @@ def to_dict(self) -> dict[str, Any]:
...
+        self.quant_method = QuantizationMethod.SINQ
+
+        self.nbits = int(nbits)
+        self.group_size = int(group_size)
+        self.tiling_mode = str(tiling_mode)
+        self.method = str(method).lower()
+
+        self.per_channel = bool(per_channel)
+        self.symmetric = bool(symmetric)
+        self.use_nf4 = bool(use_nf4)
```
SunMarc: let's not upcast here directly and check in the post init

src/transformers/utils/quantization_config.py
```diff
@@ -2092,3 +2093,142 @@ def to_dict(self) -> dict[str, Any]:
...
+
+        self.modules_to_not_convert = list(modules_to_not_convert) if modules_to_not_convert is not None else []
+
```
SunMarc: let's leave it as `None`

src/transformers/utils/quantization_config.py
```diff
@@ -2092,3 +2093,142 @@ def to_dict(self) -> dict[str, Any]:
...
+                "use_nf4": self.use_nf4,
+                "modules_to_not_convert": list(self.modules_to_not_convert or []),
+                "dtype": self.dtype,
+                "device": self.device,
+            }
+        )
+
+        base.update(self._extra_kwargs)
+
+        return base
```
SunMarc: `to_dict` should work fine no ?

src/transformers/utils/quantization_config.py
```diff
@@ -2092,3 +2093,142 @@ def to_dict(self) -> dict[str, Any]:
...
+                "this may be rejected by the backend."
+            )
+
+    @property
+    def is_trainable(self) -> bool:
+        return True
+
+    @property
+    def is_serializable(self) -> bool:
+        return True
```
SunMarc: not neeedd

src/transformers/utils/quantization_config.py
```diff
@@ -2092,3 +2093,142 @@ def to_dict(self) -> dict[str, Any]:
...
+        cfg = dict(config_dict)
+
+        cfg.pop("quant_method", None)
+
+        cfg.update(kwargs)
+
+        if return_unused_kwargs:
+            return cls(**cfg), {}
+
+        return cls(**cfg)
```
SunMarc: not sure what we are doing here

test_preparation/pipelines_torch_test_list.txt
```diff
@@ -0,0 +1,375 @@
+tests/models/aimv2/test_modeling_aimv2.py
```
SunMarc: remove

SunMarc (commented):
Thanks for the work, really appreciate to see that you adapted your work to fit the v5 !  Left a bunch of comments. I see that sinq api is quite similar to hqq and it would be great if we can improve a bit the api so that the integration is simpler and easier to maintain in general ! We had a lot of issues in our hqq integration and I prefer that we don't go through that again ;) The best would be to take inspiration from other quants methods like bnb, fp8, mxfp4 or torchao. Also, I would be great to reduce the complexity of this PR by simplifying as much as possible. You will see that the other integrations are much more easy to go through which makes the maintenance and usability in the future much better. Feel free to ask any questions, I'm really eager to merge this PR !

ChiaraBoretti:
Hi @SunMarc,

Thank you very much for taking the time to review my code and for all your helpful comments. I’ve just pushed an updated version that incorporates several of the changes you suggested.
That said, I still have a few questions, and I ran into some difficulties while trying to implement a couple of your recommendations. I was hoping you might be able to help me with those points.
I’ll reply directly to your specific comments with my questions. I’m keen to further improve the code and, with your guidance, get this pull request ready to be merged.

Thanks again for your support, and I really appreciate your help!

docs/source/en/quantization/sinq.md
ChiaraBoretti: I agree that your approach would be better overall. I also tried to implement your suggestion, but I ran into a few issues along the way.
In particular, I moved the relevant lines in quantizer_sinq.py to the beginning of the file so that my custom io_patch is applied as soon as the quantizer module is imported. While this works correctly for saving models, I’m encountering problems when loading already Sinq-quantized models.

At the moment, I’m working on a workaround where, during installation of the sinq package via pip, an autopatch is applied that automatically calls patch_hf_pretrained_io(). However, before going too far in that direction, I wanted to ask whether you might have any hints or suggestions on how to better implement your original approach.

src/transformers/integrations/sinq.py
ChiaraBoretti: Honestly, I tried to adapt my code to follow the FP8-style approach, but I wasn’t able to make it work successfully. My goal is to avoid modifying the original SINQ codebase, and the current implementation is the only one I’ve found that works reliably across all architectures (both pure LLMs and multimodal models) without touching the official SINQ code.
This is the part where I feel your help is really needed. If you have any suggestions or alternative ideas in mind, I’d be more than happy to try them out.

Thanks again for your time and support!

src/transformers/quantizers/quantizer_sinq.py
ChiaraBoretti: I’m not sure I fully understand your suggestion. Are you proposing to remove the asinq option and instead ask users to rely on the official SINQ repository if they want to use the calibrated variant of SINQ?
I just want to make sure I’m interpreting your point correctly before moving forward.

src/transformers/quantizers/quantizer_sinq.py
ChiaraBoretti: This is mainly because the current SINQ implementation works very well for pure LLM architectures, where parameter-level conversion is stable and behaves as expected. However, SINQ has not yet been fully tested on multimodal models.
I ran several experiments across different multimodal model families. While quantization of the linear layers generally works, I encountered issues with parameter-level conversion for certain architectures. For this reason, I adopted a hybrid approach: for pure LLMs, I rely on parameter-level conversion, which is well tested and robust; for more “risky” architectures (particularly multimodal models) I instead fall back to an all-at-once conversion strategy, which has proven to be more reliable in practice.

I arrived at this solution after testing models from the Gemma family. As I understand it, the root cause is that during ```from_pretrained()```, Transformers still calls the model’s ```_init_weights``` method on modules it considers missing or uninitialized. In models like Gemma3, specifically within its SigLIP component, ```_init_weights``` performs operations such as:

```init.xavier_uniform_(module.q_proj.weight)```

At that point, however, ```q_proj``` has already been replaced with SINQLinear. Since SINQLinear does not expose a ```.weight``` attribute, the initialization step fails.

Comment (SunMarc) [reply src/transformers/quantizers/quantizer_sinq.py]: > I arrived at this solution after testing models from the Gemma family. As I understand it, the root cause is that during from_pretrained(), Transformers still calls the model’s _init_weights method on modules it considers missing or uninitialized. In models like Gemma3, specifically within its SigLIP component, _init_weights performs operations such as:
> 
> init.xavier_uniform_(module.q_proj.weight)
> 
> At that point, however, q_proj has already been replaced with SINQLinear. Since SINQLinear does not expose a .weight attribute, the initialization step fails.
> 

You can skip the initialization for a specific parameter or even modules by setting `_is_hf_initialized` attribute ! We usually set it when we load the weights (e.g check the other quant methods) but we can also do it directly when we replace the modules if this is simpler/better.

File: src/transformers/quantizers/quantizer_sinq.py (hunk 37 above)
Comment (SunMarc) [reply src/transformers/quantizers/quantizer_sinq.py]: Yeah that's right. It will be better ! Maybe in the future it could make sense to add it, but for now let's remove this !

File: src/transformers/integrations/sinq.py (hunk 13 above)
Comment (SunMarc) [reply src/transformers/integrations/sinq.py]: A few things that could be nice to simplify if possible: 
- Can't we replace the linear with the `SINQLinear` before instead of here ? 
- Do we really have to pass the linear layer in `SINQLinear` ? We can maybe recreate a `SINQLinear` impl that is a bit more simpler ? 
- When does the quantization happens ? When the SINQLinear is created ? Maybe we can try to have a bit more flexibility here and create a quantize() method.

**Comment by SunMarc** [comment]
Thanks for taking the time to go through the comments, answered to a couple of them !

File: src/transformers/quantizers/quantizer_sinq.py (hunk 38 above)
Comment (ChiaraBoretti) [reply src/transformers/quantizers/quantizer_sinq.py]: Thank you for the suggestion about using `_is_hf_initialized`! After further investigation, I believe that the hybrid approach is not always required for multimodal models. In my understanding, whether it is needed depends largely on how `_init_weights` is implemented for a given architecture.
In the case of Gemma3, I ran into issues due to differences between how the language model and the vision encoder (SigLIP) handle weight initialization.
The language model in Gemma3 defines `_init_weights` in a module-level way, checking individual submodules:
```
if isinstance(module, nn.Linear):
    ...
```
and with param-level conversion:
- Each nn.Linear is replaced by SINQLinear
- Each SINQLinear has `_is_hf_initialized = True` set during conversion (line 117 in sinq.py)
- When `PreTrainedModel._initialize_weights` runs, it inspects each SINQLinear individually, sees the flag, and correctly skips initialization. So the param-level workflow works as intended for the language model.

On the other hand, the vision encoder (SigLIP) uses a different initialization pattern. Its `_init_weights` method checks parent modules and then directly initializes their children:
```
elif isinstance(module, SiglipAttention):
    init.xavier_uniform_(module.q_proj.weight)
    init.zeros_(module.q_proj.bias)
    # same for k_proj, v_proj, out_proj
```
This leads to the following failure mode:
- During param-level conversion, q_proj, k_proj, etc. are replaced with SINQLinear
- Each SINQLinear correctly has `_is_hf_initialized = True`, but the parent SiglipAttention module does not have this flag
- When `_initialize_weights` reaches SiglipAttention, it does not see `_is_hf_initialized`, so it calls `_init_weights`, which tries to access `module.q_proj.weight`, but `q_proj` is now a SINQLinear, which leads to a crash
- If SiglipAttention itself is marked with `_is_hf_initialized = True` to prevent the crash, another issue arises: other parameters in the vision encoder are never initialized, which results in incorrect model outputs.

I see two possible approaches to solve this problem:

1. Exclude the vision encoder from quantization. Add the vision modules to `modules_to_not_convert`. This is the simplest solution and, in my view, the most appropriate. The issue appears to be model-specific (related to how SigLIP implements `_init_weights`), rather than a general multimodal architecture problem.

2. Maintain the hybrid approach. Use standard nn.Linear layers during weight loading and initialization for the vision modules. Only after initialization is complete those layers are quantized and replaced with SINQLinear. This avoids the crash and preserves correct initialization, but at the cost of added complexity.

What do you think is the most appropriate solution to adopt?

Thanks!

File: docs/source/en/quantization/sinq.md (hunk 6 above)
Comment (SunMarc) [reply docs/source/en/quantization/sinq.md]: Can you add it in is_serializable method, so that it is only called when we want to actually save the model ?

File: src/transformers/quantizers/quantizer_sinq.py (hunk 38 above)
Comment (SunMarc) [reply src/transformers/quantizers/quantizer_sinq.py]: Oh okay I see the issue. Thanks for digging into this, a lot of things changed in the v5 and initialization was one big issue that still needs some minor fixes to work with quantization. But basically, if i understood well, for module.q_proj.weight is not available for `SINQLinear` unlike a classic `Linear`. Usually, `_is_hf_initialized` is initialized at the level of the parameters and we skip the initialization if the param has this param. But since here, we can't even access it.

```python
def zeros_(tensor: torch.Tensor) -> torch.Tensor:
    if not getattr(tensor, "_is_hf_initialized", False):
        return TORCH_INIT_FUNCTIONS["zeros_"](tensor)
    return tensor
```

A simple solution that we did for PretrainedModel class is just check whether weights and bias attribute exists or not. 

```python
 if getattr(module, "weight", None) is not None:
                init.ones_(module.weight)
```

Maybe you can open a separate PR to fix this issue for the models where this happens ? To me, it feels like the cleanest solution.

ChiaraBoretti:
Hi @SunMarc,

I’ve just pushed an updated version of the code with the following changes:

- Removal of asinq. I removed the asinq implementation and added an explicit error message when a user tries to set `method='asinq'`. The message directs users to the official SINQ repository for running the calibrated version of SINQ.

- Architecture-specific handling. I answered your comment regarding handling particular architectures with an all-in-one approach. For now, I’ve moved away from the all-in-one strategy. As mentioned in my answer, I believe the issue is model-specific rather than a general multimodal architecture problem. Based on this, I opted to explicitly specify the vision components in `modules_to_not_convert` for models that raise errors due to their `init_weights` implementation.

- Updated quantization flow. I reworked the quantization flow by modifying how SINQLinear is implemented, following your suggestion (I hope to understand them correctly). The new flow is: Create an empty SINQLinear -> Load the weights -> Explicitly call the weights quantization via a dedicated `quantize()` function instead of performing quantization when the SINQLinear is created (in the `__init__` of the class).
As a result, also the SinqQuantize and SinqDeserialize classes are now much simpler.

Please let me know if these changes are moving the PR in the right direction for approval, and if there’s anything else I should address.

Thanks!

SunMarc:
Thanks a lot for this ! 

> 
Architecture-specific handling. I answered your comment regarding handling particular architectures with an all-in-one approach. For now, I’ve moved away from the all-in-one strategy. As mentioned in my answer, I believe the issue is model-specific rather than a general multimodal architecture problem. Based on this, I opted to explicitly specify the vision components in modules_to_not_convert for models that raise errors due to their init_weights implementation.

Answered in the thread ! But this is a separate issue from this PR and it should be quite simple to fix. 

> Updated quantization flow. I reworked the quantization flow by modifying how SINQLinear is implemented, following your suggestion (I hope to understand them correctly). The new flow is: Create an empty SINQLinear -> Load the weights -> Explicitly call the weights quantization via a dedicated quantize() function instead of performing quantization when the SINQLinear is created (in the __init__ of the class).
> As a result, also the SinqQuantize and SinqDeserialize classes are now much simpler.

I had a quick look at it looks much better. The flow also makes more sense now.

Let me know if you want a new review but the simpler the code looks like, the better it is !

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,344 @@
...
+                continue
+
+            parent_path, _, child_name = full_name.rpartition(".")
+            parent = model.get_submodule(parent_path) if parent_path else model
+
+            # Create empty SINQLinear (no weights yet)
+            sinq_layer = SINQLinear(
+                in_features=module.in_features if not self.pre_quantized else None,
+                out_features=module.out_features if not self.pre_quantized else None,
+                bias=(module.bias is not None) if not self.pre_quantized else False,
```
SunMarc: put that in a replace function like the other methods

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,344 @@
...
+        - For method="asinq": please use the official SINQ repository.
+        """
+
+        if self.quantization_config.method == "asinq" and not self.pre_quantized:
+                raise ValueError(
+                "You are using `method='asinq'` in the quantization config. Right now the calibrated version of SINQ"
+                " is not supported in Hugging Face, please refer and use the official SINQ repository "
+                "`to quantized a model with this method. "
+            )
+
```
SunMarc: put that in validate_environment

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,344 @@
...
+        try:
+            if hasattr(model, "config") and hasattr(model.config, "_name_or_path"):
+                model_id = model.config._name_or_path
+            if model_id is None:
+                model_id = kwargs.get("pretrained_model_name_or_path", None)
+            if model_id is None and "config" in kwargs and hasattr(kwargs["config"], "_name_or_path"):
+                model_id = getattr(kwargs["config"], "_name_or_path", None)
+
+            logger.info(f"[SinqHfQuantizer] Detected model_id = {model_id}")
+
```
SunMarc: remove

test_preparation/pipelines_torch_test_list.txt
```diff
@@ -0,0 +1,374 @@
+tests/models/albert/test_modeling_albert.py
+tests/models/align/test_modeling_align.py
+tests/models/altclip/test_modeling_altclip.py
+tests/models/apertus/test_modeling_apertus.py
+tests/models/arcee/test_modeling_arcee.py
+tests/models/aria/test_modeling_aria.py
+tests/models/audio_spectrogram_transformer/test_modeling_audio_spectrogram_transformer.py
```
SunMarc: lot of txt files to detele

src/transformers/quantizers/quantizer_sinq.py
ChiaraBoretti: Did it! I added a ```replace_with_sinq_linear()``` function in sinq.py

src/transformers/quantizers/quantizer_sinq.py
ChiaraBoretti: moved to validate_environment()

test_preparation/pipelines_torch_test_list.txt
ChiaraBoretti: I deleted all these txt files

src/transformers/quantizers/quantizer_sinq.py
ChiaraBoretti: Thanks! I removed this function since it is no more used in the code.

ChiaraBoretti:
Thanks, @SunMarc, for your comments!

I’ve submitted a new version of the code, could you please take a look and let me know if there’s anything else I should change? I think that your suggestions helped me in making it simpler and cleaner.

Thanks again!

philippebich:
Hi @SunMarc ! Any news?

docs/source/en/quantization/sinq.md
SunMarc: can't be not put the device here ?

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,170 @@
...
+                input_dict[k] = v[0]
+
+        W_q = input_dict.get(".W_q", None)
+        meta = input_dict.get(".meta", None)
+        bias = input_dict.get(".bias", None)
+
+        if W_q is None or meta is None:
+            v = next(iter(input_dict.values()))
+            if isinstance(v, list):
+                v = v[0]
```
SunMarc, line 158: add a comment above to explain this path

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,299 @@
...
+    def __init__(self, quantization_config: SinqConfig, **kwargs):
+        super().__init__(quantization_config, **kwargs)
+
+        self._normalized_device_str: str | None = None
+        self._do_param_level_sinq: bool = False
+
+    def is_serializable(self, safe_serialization: Optional[bool] = None) -> bool:
+        if safe_serialization:
+            return True
+        return False
```
SunMarc: the safe_serialization=False option was removed. So just return True

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,299 @@
...
+            if passed_dtype not in (torch.float16, torch.bfloat16, torch.float32):
+                logger.warning(
+                    f"SINQ quantization with dtype={passed_dtype} may not be supported. "
+                    f"Recommended dtypes: torch.float16, torch.bfloat16, torch.float32"
+                )
+            
+            self.dtype = passed_dtype
+        else:
+            # Set default dtype if not provided
+            self.dtype = torch.bfloat16
```
SunMarc: don't do that here, there is a method for that called update_torch_dtype I think. You won't need to check for str as it will be a torch.dtype

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,299 @@
...
+
+        device_map = kwargs.get("device_map", None)
+
+        devs = _flatten_device_map(device_map)
+        if devs:
+            if len(devs) > 1:
+                raise RuntimeError(
+                    "SinqHfQuantizer: multi-GPU device_map detected, but SINQ currently supports only a single CUDA "
+                    f"device. Got {sorted(devs)}. Please use device_map=None."
```
SunMarc: we don't need `_flatten_device_map`. Just simplify to use the same check as the other methods.

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,299 @@
...
+        device_str = _normalize_cuda_device(getattr(self.quantization_config, "device", "auto"))
+
+        model = replace_with_sinq_linear(
+                model,
+                modules_to_not_convert=self.modules_to_not_convert,
+                quant_config=sinq_quant_dict,
+                compute_dtype=self.dtype,
+                device=device_str,
+                pre_quantized=self.pre_quantized,
+            )
```
SunMarc: You can use the value from device_map actually instead of using the one passed in quantization_config. So it should be easy to remove. Note that `_process_model_before_weight_loading` is actually performed under the with torch.device("meta") context manager, so all the tensors are on meta device actually

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        Called after *all* weights have been loaded.
+
+        We patch the HF save/load methods here so that subsequent save_pretrained
+        calls will correctly serialize SINQ modules. The patch must happen before
+        any save attempt, not during save (which would be too late for the first save).
+        """
+        from sinq.hf_io import patch_hf_pretrained_io
+        patch_hf_pretrained_io()
+        return model
```
SunMarc, line 261: Can't we put that in `is_serializable` instead of here ? Also in the docs, you are still asking the user to do it

src/transformers/utils/quantization_config.py
```diff
@@ -1925,3 +2093,106 @@ def to_dict(self) -> dict[str, Any]:
...
+        self,
+        nbits: int = 4,
+        group_size: int = 64,
+        tiling_mode: str = "1D",
+        method: str = "sinq",  # "sinq" | "asinq"
+        per_channel: bool = True,
+        symmetric: bool = True,
+        use_nf4: bool = False,
+        modules_to_not_convert: Optional[List[str]] = None,
+        device: Optional[Union[str, int]] = "cpu",
```
SunMarc: to remove

src/transformers/utils/quantization_config.py
```diff
@@ -1925,3 +2093,106 @@ def to_dict(self) -> dict[str, Any]:
...
+        self.group_size = int(self.group_size)
+        self.tiling_mode = str(self.tiling_mode)
+        self.method = str(self.method).lower()
+        
+        self.per_channel = bool(self.per_channel)
+        self.symmetric = bool(self.symmetric)
+        self.use_nf4 = bool(self.use_nf4)
+        
+        if self.modules_to_not_convert is not None:
+            self.modules_to_not_convert = list(self.modules_to_not_convert)
```
SunMarc: don't convert it here. Otherwise, we won't be able to add good default like the lm_head that shouldn't be quantized

src/transformers/utils/quantization_config.py
```diff
@@ -1925,3 +2093,106 @@ def to_dict(self) -> dict[str, Any]:
...
+                f"SINQ: group_size={self.group_size} is not a multiple of 8; "
+                "this may be rejected by the backend."
+            )
+
+    def to_dict(self) -> dict[str, Any]:
+        """
+        Convert configuration to a plain dict that will be stored in `config.json`
+        under `quantization_config`.
+        """
+        base = super().to_dict()
```
SunMarc: not needed

tests/quantization/sinq/test_sinq.py
```diff
@@ -0,0 +1,740 @@
...
+    def test_quantizer_init(self):
+        """Test basic quantizer initialization."""
+        config = SinqConfig()
+        quantizer = SinqHfQuantizer(quantization_config=config)
+
+        assert quantizer.quantization_config == config
+        assert quantizer.requires_calibration is False
+        assert quantizer.requires_parameters_quantization is True
+
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
```
SunMarc: we have specific decorator for this. use those instead of this one

tests/quantization/sinq/test_sinq.py
```diff
@@ -0,0 +1,740 @@
...
+
+All tests run with real SINQ library.
+"""
+
+import pytest
+import torch
+import torch.nn as nn
+from unittest.mock import Mock, patch
+import tempfile
+import os
```
SunMarc: those tests are good but it would be easier for us to try to respect the same structure as the other quantization tests. It will help us better maintain it

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,299 @@
...
+        Called on meta-initialized model, before loading any weights.
+
+        For SINQ, we replace nn.Linear modules with empty SINQLinear modules here.
+        The actual quantization happens later in SinqQuantize.convert() when weights are loaded.
+        """
+        from ..integrations.sinq import replace_with_sinq_linear
+
+        self.modules_to_not_convert = self.get_modules_to_not_convert(
+            model, (self.quantization_config.modules_to_not_convert or []), keep_in_fp32_modules
+        )
```
SunMarc, line 216: this is where the good default are added

SunMarc (commented):
Thanks, just a few more fixes and we should be good to merge. Thanks for spending lots of time on this PR ! Sorry I took a break hence I didn't review the PR since then.

src/transformers/quantizers/quantizer_sinq.py
ChiaraBoretti: I’m still a bit confused about this point. I tried removing the device specification from SinqConfig and instead relying directly on the device_map argument provided to from_pretrained(). However, this only works correctly when device_map=None.
If I set device_map to a CUDA device, the full-precision model gets loaded onto the GPU first, which is exactly what I want to avoid. My goal is for only the final quantized model to be loaded onto the user-specified GPU.

This is the change I introduced, (notice that with this I cannot give the user the possibility to specify where to locate the final quantized model):
```python
        # Get device from device_map
        if isinstance(device_map, dict):
            device_str = next(iter(device_map.values()), "cuda:0")
        else:
            device_str = "cuda:0"

        model = replace_with_sinq_linear(
                model,
                modules_to_not_convert=self.modules_to_not_convert,
                quant_config=sinq_quant_dict,
                compute_dtype=self.dtype,
                device=device_str,
                pre_quantized=self.pre_quantized,
            )
```
It effectively works only if I set device_map=None or if I don't specify any device_map in the from_pretrained() function. 
From my understanding I suppose that the issue is that device_map controls where accelerate loads each weight, and by the time SinqQuantize.convert() is executed, the full-precision weights have already been placed on the GPU. This motivates the approach of introducing a device parameter in SinqConfig: it allows the user to decide where the quantized model should reside, assuming no device_map is provided to from_pretrained().

I hope to be clear and I would really appreciate your help in this because probably there is something I'm missing to make it work.
Thanks!

src/transformers/quantizers/quantizer_sinq.py
ChiaraBoretti: I tried putting it in ```is_serializable```, but it doesn't work on the first save. The issue is that ```is_serializable()``` is called from within ```save_pretrained()```, so by that point we're already executing the unpatched method. The patch only takes effect for subsequent saves. Moving it to ```_process_model_after_weight_loading``` ensures the patch is applied immediately after model loading, before any save attempt.

SunMarc: sounds good then !

src/transformers/quantizers/quantizer_sinq.py
SunMarc: > If I set device_map to a CUDA device, the full-precision model gets loaded onto the GPU first, which is exactly what I want to avoid. My goal is for only the final quantized model to be loaded onto the user-specified GPU.

Indeed, the weights are loaded to the GPU but as soon as they are loaded, they should be converted thanks to your `SinqQuantize` ops, thus decreasing the memory. Since we are doing this async and with threads, it should not load all the weights then quantize but rather, load some weights -> quantize and so on. We tested with different methods like bnb and there was no issues. Maybe check on your side that `module.quantize(weight_tensor)` decreases the memory when `weight_tensor` is on a CUDA. Let me know if this is clear !

ChiaraBoretti: Yes, this makes more sense now, thanks!
However, I also tried the bitsandbytes (bnb) quantization strategy using the following simple script:
```python
qmodel_name = "Qwen/Qwen3-4B"
from transformers import BitsAndBytesConfig

cfg = BitsAndBytesConfig(load_in_4bit=True)
qtok = AutoTokenizer.from_pretrained(qmodel_name)
qmodel = AutoModelForCausalLM.from_pretrained(
    qmodel_name,
    dtype=torch.bfloat16,
    quantization_config=cfg,
    device_map="cuda:0"
)
```
In this case as well, if I set ```device_map="cuda:0"```, GPU memory usage is around 8 GB. If I don’t set ```device_map``` at all, the usage is still ~8 GB. 
I did also the same test for the finegrained_fp8 quantization and with ```device_map="cuda:0"``` or ```device_map="auto"``` I have a GPU occupation of ~8 GB.

For the SINQ case, with the code I’m currently testing, I can remove the ```device``` parameter from the config, but then ```device_map``` must be None (or not set) in the ```from_pretrained()```. In that situation, the quantized model is always loaded on the first available CUDA device. This also seems to be what happens with bnb, based on my observations.

Am I doing something wrong here, or is this the expected behavior when using bitsandbytes quantization?

Thanks!

SunMarc: For bnb, this is normal, we coded it this way and kept it this way due to prior limitation. For SINQ, I don't think this should happen.

SunMarc: see what we have in bnb 

```python
 def update_device_map(self, device_map):
        if device_map is None:
            if torch.cuda.is_available():
                device_map = {"": torch.cuda.current_device()}
            elif is_torch_npu_available():
                device_map = {"": f"npu:{torch.npu.current_device()}"}
            elif is_torch_hpu_available():
                device_map = {"": f"hpu:{torch.hpu.current_device()}"}
            elif is_torch_xpu_available():
                device_map = {"": torch.xpu.current_device()}
            else:
                device_map = {"": "cpu"}
            logger.info(
                "The device_map was not initialized. "
                f"Setting device_map to {device_map}. "
                "If you want to use the model for inference, please set device_map ='auto' "
            )
        return device_map
```

ChiaraBoretti: Ah, okay, thanks that clears things up.

Is this behavior also expected when using fine-grained FP8 quantization? I noticed that in the code there are some checks on the values of ```device_map```, but nothing that explicitly sets it when it is None.

With this type of quantization as well, when I explicitly specify ```device_map```, it looks like the full-precision model is first loaded onto the GPU (~8 GB), very similar to what happens with SINQ.
Is this the intended behavior, or am I misunderstanding how fine-grained FP8 loading is supposed to work?

Thanks again!

SunMarc: Oh you need to remove the cache allocator most probably as it makes you think that you are using more memory. You need to check for cuda allocated memory and not reserved memory. This is what I get with 

```
  ┌────────────────────────┬─────────┐
  │         Metric         │  Value  │
  ├────────────────────────┼─────────┤
  │ Model memory footprint │ 2.42 GB │
  ├────────────────────────┼─────────┤
  │ CUDA allocated         │ 2.66 GB │
  ├────────────────────────┼─────────┤
  │ CUDA reserved          │ 7.07 GB │
  └────────────────────────┴─────────┘
```

```python
  import torch
  from transformers import AutoTokenizer, AutoModelForCausalLM, BitsAndBytesConfig

  def get_gpu_memory():
      if torch.cuda.is_available():
          allocated = torch.cuda.memory_allocated() / 1024**3
          reserved = torch.cuda.memory_reserved() / 1024**3
          return allocated, reserved
      return 0, 0

  print("Initial GPU memory:")
  alloc, res = get_gpu_memory()
  print(f"  Allocated: {alloc:.2f} GB, Reserved: {res:.2f} GB")

  qmodel_name = "Qwen/Qwen3-4B"

  cfg = BitsAndBytesConfig(load_in_4bit=True)

  print(f"\nLoading tokenizer from {qmodel_name}...")
  qtok = AutoTokenizer.from_pretrained(qmodel_name)

  print(f"Loading model with 4-bit quantization...")
  qmodel = AutoModelForCausalLM.from_pretrained(
      qmodel_name,
      torch_dtype=torch.bfloat16,
      quantization_config=cfg,
      device_map="cuda:0"
  )

  print("\nAfter loading model:")
  alloc, res = get_gpu_memory()
  print(f"  Allocated: {alloc:.2f} GB, Reserved: {res:.2f} GB")

  # Also show nvidia-smi style info
  print("\nModel memory footprint:")
  print(f"  {qmodel.get_memory_footprint() / 1024**3:.2f} GB")
```

ChiaraBoretti: Ah ok now I get it! thanks. I checked the allocated memory after applying the Sinq quantization, and it’s around 2.7 GB, consistent with what you observed using bnb.

I’ve pushed a new version of the code where I removed the device specification from the SinqConfig and now rely on device_map instead. I also applied the other minor adjustments you suggested.
Could you please take a look and let me know if everything seems good to go, or if I may have missed anything?

Thanks a lot for your support!
Chiara

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,257 @@
...
+
+    requires_parameters_quantization: bool = True
+
+    def __init__(self, quantization_config: SinqConfig, **kwargs):
+        super().__init__(quantization_config, **kwargs)
+
+        self._normalized_device_str: str | None = None
+        self._do_param_level_sinq: bool = False
+
+    def is_serializable(self, safe_serialization: Optional[bool] = None) -> bool:
```
SunMarc: We don't have the safe_serialization flag anymore 
```suggestion
    def is_serializable(self) -> bool:
```

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,257 @@
...
+        if not self._do_param_level_sinq:
+            return False
+
+        module, tensor_name = get_module_from_name(model, param_name)
+
+        if tensor_name != "weight":
+            return False
+        module_name = param_name.rsplit(".", 1)[0] if param_name.endswith((".weight", ".bias")) else param_name
+        if not should_convert_module(module_name, self.modules_to_not_convert):
+            return False
```
SunMarc: not needed, just the `isinstance(module, SINQLinear)` is enough 
```suggestion
```

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,257 @@
...
+            model, (self.quantization_config.modules_to_not_convert or []), keep_in_fp32_modules
+        )
+
+        # Enable param-level quantization for SINQ method
+        self._do_param_level_sinq = (
+            self.quantization_config.method == "sinq" and not self.pre_quantized
+        )
+
+        # if not self.pre_quantized and self.quantization_config.method == "asinq":
+        #     raise ValueError("A-SINQ is not supported in HuggingFace integration")
```
SunMarc:
```suggestion
```

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,257 @@
...
+            device_str = "cuda:0" if torch.cuda.is_available() else "cpu"
+        elif isinstance(device_map, dict):
+            first_device = next(iter(device_map.values()), 0)
+            if isinstance(first_device, int):
+                device_str = f"cuda:{first_device}"
+            else:
+                device_str = str(first_device)
+        else:
+            # device_map is a string like "auto", "balanced", etc.
+            device_str = "cuda:0" if torch.cuda.is_available() else "cpu"
```
SunMarc: If you do that here, you should just update the `device_map` as we do for bnb

```python
def update_device_map(self, device_map):
        if device_map is None:
            if torch.cuda.is_available():
                device_map = {"": torch.cuda.current_device()}
```

tests/quantization/sinq/test_sinq.py
```diff
@@ -0,0 +1,573 @@
...
+        )
+
+        # Should return the original tensor
+        self.assertIn("layer1.weight", result)
+
+
+@require_torch_gpu
+@slow
+class SinqIntegrationTest(unittest.TestCase):
+    """End-to-end integration tests for SINQ quantization."""
```
SunMarc: Please only keep the integration tests that looks like the other quantization tests. I will do a refactor later on so that all the tests looks the same but for now, i prefer to keep the test short / same as the others

SunMarc (approved):
Thanks a lot ! This is ready to be merged. Just a few things to remove/improve + fix the CI. You need to run make style + check-repo or fix-repo

tests/quantization/sinq/test_sinq.py
ChiaraBoretti: Thanks for the feedback. Just to make sure I understand correctly, would you like me to remove the internal quantizer tests and the module-level tests, and keep only the Config tests and the end-to-end integration tests? Or do you prefer that I merge the quantizer validation and module conversion tests in the SinqIntegration one, like in ```finegrained_fp8```?

SunMarc: What would be nice is to just have something like `FineGrainedFP8ConfigTest` and `FP8QuantizerTest`. So just the config tests + end-to-end integrations for now

ChiaraBoretti: Ok, thanks! I modified the tests script. Hope now it is correct.

ChiaraBoretti:
Hi @SunMarc 
I pushed the latest version of the code with the modification you suggested. Then I run make style, make quality and make repo-consistency. Does the code now seems ready to be merged? Do I have to run other checks?
Thanks, 
Chiara

SunMarc:
The CI is still red, can you try to check why ? Also there are some conflicts

ChiaraBoretti:
Hi @SunMarc,

I’ve resolved the merge conflicts and rerun the test suite. At the moment, the two remaining failing tests appear to be related, with the root issue coming from test_torch. From what I can tell, the failure is tied to the Bark model tests rather than my integration code:

```bash
FAILED tests/models/bark/test_modeling_bark.py::BarkFineModelTest::test_inputs_embeds
NotImplementedError: Module [ModuleList] is missing the required "forward" function
```

Would you like me to investigate and fix this as well, or should it be handled by the maintainers of the Bark model implementation?

Thanks!

SunMarc:
No, this is fine ! Thanks for all your hardwork and keeping up with me. I'll ask @MekkCyber for a final review to see if I missed anything.

docs/source/en/quantization/selecting.md
```diff
@@ -56,6 +57,16 @@ See the [bitsandbytes documentation](./bitsandbytes) for more details.
 
 See the [HQQ documentation](./hqq) for more details.
 
+#### SINQ
+
+| Pros                                                                 | Cons                                                                       |
+|----------------------------------------------------------------------|----------------------------------------------------------------------------|
+| Super-fast but high-quality quantization process, no calibration data needed.              | Accuracy can degrade significantly at bit depths <=2-bit.                     |
```
MekkCyber:
```suggestion
| Super-fast but high-quality quantization process, no calibration data needed.              | Accuracy can degrade significantly at bit widths <=2-bit.                     |
```

docs/source/en/quantization/selecting.md
```diff
@@ -56,6 +57,16 @@ See the [bitsandbytes documentation](./bitsandbytes) for more details.
...
+| GemLite backend for faster inference.                                  | Slower inference for 3-bit models (no gemlite kernel)
+| Supports wide range of bit depths (8, 4, 3, 2 bit).              |                                                                            |
```
MekkCyber:
```suggestion
| Supports wide range of bit widths (8, 4, 3, 2 bits).              |                                                                            |
```

src/transformers/integrations/sinq.py
```diff
@@ -0,0 +1,173 @@
+# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
+#
```
MekkCyber:
```suggestion
# Copyright 2026 The HuggingFace Inc. team. All rights reserved.
#
```

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,259 @@
+# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
```
MekkCyber:
```suggestion
# Copyright 2026 The HuggingFace Inc. team. All rights reserved.
```

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def validate_environment(self, *args, **kwargs) -> None:
+        from ..utils import is_sinq_available
+
+        if not is_sinq_available():
+            raise ImportError("The 'sinq' package is not installed. Please install it with: pip install sinq")
+
+        if not torch.cuda.is_available():
+            raise RuntimeError("SINQ currently expects a CUDA device (for GemLite backend).")
+
```
MekkCyber, line 91: do we need cuda for `sinq` in general or just the `gemlite` backend ?

src/transformers/quantizers/quantizer_sinq.py
```diff
@@ -0,0 +1,259 @@
...
+                    "SinqHfQuantizer: multi-GPU device_map detected, but SINQ currently supports only a single CUDA "
+                    f"device. Got {sorted(device_map_values)}. Please use device_map=None."
+                )
+
+        if self.quantization_config.method == "asinq" and not self.pre_quantized:
+            raise ValueError(
+                "You are using `method='asinq'` in the quantization config. Right now the calibrated version of SINQ"
+                " is not supported in Hugging Face, please refer and use the official SINQ repository "
+                "`to quantized a model with this method. "
+            )
```
MekkCyber, line 107:
```suggestion
            raise ValueError(
                "You are using `method='asinq'` in the quantization config. Right now the calibrated version of SINQ"
                " is not supported in Hugging Face, please refer and use the official SINQ repository "
                "`to quantize a model with this method. "
            )
```

tests/quantization/sinq/test_sinq.py
```diff
@@ -0,0 +1,187 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
```
MekkCyber:
```suggestion
# Copyright 2026 The HuggingFace Team. All rights reserved.
```

tests/quantization/sinq/test_sinq.py
```diff
@@ -0,0 +1,187 @@
...
+@slow
+@require_torch_gpu
+class SinqTest(unittest.TestCase):
+    """Integration tests for SINQ quantization."""
+
+    model_name = "Qwen/Qwen2-0.5B"
+    input_text = "What is the capital of France?"
+    max_new_tokens = 10
+    device_map = torch_device
+
```
MekkCyber, line 100: maybe we can add an expected output slow test with a small model to make sure we don't generate gibberish

MekkCyber (commented):
Thank you for the contribution @ChiaraBoretti ! it looks great 🔥 ! Just have some few nits, mostly typos and tests and we are good to go

src/transformers/quantizers/quantizer_sinq.py
ChiaraBoretti: CUDA is only strictly required when using the GemLite backend. The core sinq algorithm can also run on the CPU, although this results in significantly slower inference and longer quantization times compared to GPU execution.

tests/quantization/sinq/test_sinq.py
ChiaraBoretti: Yes, that makes sense. Would you recommend adding a new test class with a small model definition and just a ```test_quantized_model``` function (like the one for the Qwen model test) to verify that the quantized model produces the expected output?

MekkCyber: nw we can just add it to the same `SinqTest` class

src/transformers/quantizers/quantizer_sinq.py
MekkCyber: but in the validation we are erroring out if no gpu is available, is that intended ?

ChiaraBoretti: You are right, raising an error may be too conservative here. I can replace it with a warning if you agree that this would be the better approach.

MekkCyber: yes that sounds better! let's do that, and update the tests, and maybe keep the error if gemlite is available and the dispatching to gemlite is done automatically in the library

ChiaraBoretti: Thank you for your suggestions. I have fixed the typos, added the requested test, and replaced the error with a warning. Please let me know if any further changes are needed.

Thanks!

tests/quantization/sinq/test_sinq.py
```diff
@@ -0,0 +1,196 @@
...
+            decoded = self.tokenizer.decode(output[0], skip_special_tokens=True)
+
+            self.assertIsNotNone(decoded)
+            self.assertGreater(len(decoded), len(self.input_text))
+
+            del loaded_model
+
+
+if __name__ == "__main__":
+    unittest.main()
```
MekkCyber:
```suggestion
```

MekkCyber (approved):
Very cool! thanks a lot for iterating on this!

tests/quantization/sinq/test_sinq.py
MekkCyber: removing since we don't need it in our test suite

MekkCyber:
@bot /style

PR #43113, opened 2026-01-05
Fix TypeError in SummarizationPipeline.check_inputs when max_new_tokens is None

majiayu000:
# What does this PR do?

Fixes #43093

This PR handles the case where `max_new_tokens` or `min_length` could be `None` in the `check_inputs` method of `SummarizationPipeline`. Previously, comparing `None` with integers would raise a `TypeError`.

The fix adds an early return when either value is `None`, avoiding the comparison that caused the error.

## Before submitting
      Pull Request section?
      to it if that's the case. Issue: #43093

## Who can review?

@Rocketknight1 (mentioned in the issue)


Rocketknight1:
Yes, fix looks good! While we're here, can you update the error message below to say "less than" rather than "inferior than"?

Rocketknight1:
Hi @majiayu000, I'm sorry about this! Since this PR was opened, we decided to finalize deprecation of the `SummarizationPipeline` (PR #43256). It's going to be removed entirely, so we're closing PRs related to it. This fix was good, though, so I hope this doesn't discourage you from contributing to Transformers in future!

PR #43120, opened 2026-01-05
[`GPT OSS`] Fix false flag

vasqu:
#42736 removed flex attn support but it turns out that #41083 added support for it in flex attention. There were typos that indicated wrong support for attn flavors, sorry about that should've noticed it / checked it properly.

So gpt oss does not support sdpa but everything else (limited FA support to the specific kernel).


Cyrilvallez (approved):
Yes thanks!

PR #43121, opened 2026-01-05
[`FA`] Generalize fa config checks and fix flags

vasqu:
As per title, oftentimes we have somethin along `config._attn_implementation == "flash_attention_2"`. This is unideal and very faulty as other FA flavors will be ignored, e.g. kernels and FA3. This simply changes this to the generalized version everywhere. 
Additionally, fixed a few wrong flags that still used `_supports_flash_attn_2` (including within tests), relevant commit a231a2be7fa81445085ecb50371606aeb7a113a7 

Note that sam2 video, sam3 tracker video, edgetam video do not have any FA tests but they only use a mask with target guided attention which is not used per default, so we add a fallback similar to sam3 to use FA for other attention blocks.

Note that I changed pixtral slightly more due to it doing unnecessary movements at each attention layer. Not sure why the mask was properly done but not the pos ids.


vasqu:
run-slow: pixtral

Cyrilvallez (commented):
Very needed indeed, now that we have many flavors!
Even if it's probably alright to keep a small check like this, wdyt about introducing a small helper `is_flash_implementation` (probably in `import_utils.py` along `is_tracing` etc?) or something? Would probably make it clearer/more maintainable if we keep adding flavors!

Cyrilvallez:
Sounds to me like it's quite important to have a unique entry-point for these cases, as otherwise we have the risk to have very surprising bugs/issues in just a few models because they simply miss some checks

vasqu:
Yes let me change the check to a separate fn, will probably go over more models then (which already do the proper `"flash" in ...` check)

vasqu:
Moved it to `utils.generic` (import utils doesn't fit IMO), but added this fn to all models now. Should make this way more maintainable

src/transformers/models/edgetam/modeling_edgetam.py
```diff
@@ -171,6 +174,15 @@ def forward(
...
 
+        if is_flash_attention_requested(self.config) and attention_similarity is not None:
+            # Target guided masks are represented as float masks and are incompatible with Flash Attention
+            # Fallback to SDPA for this call only so the rest of the model can still benefit from FA
+            attention_interface = ALL_ATTENTION_FUNCTIONS["sdpa"]
+            logger.warning_once(
+                "Falling back to SDPA for target-guided attention because "
+                "Flash Attention does not support additive bias masks."
+            )
+
```
Cyrilvallez, line 185: Was not there before, but trusting you on this one!

src/transformers/utils/generic.py
```diff
@@ -210,6 +210,27 @@ def is_mlx_array(x):
     return False if not _is_mlx_available else _is_mlx(x)
 
 
+def is_flash_attention_requested(config=None, requested_attention_implementation: str | None = None):
+    """
+    Checks whether some flavor of flash attention is requested or not.
+    Priority order first goes for any explicitly passed value `requested_attention_implementation` and
+    then checks the config's saved value `config._attn_implementation`.
```
Cyrilvallez: Maybe let's just raise if both are provided no? As it does not make sense to give both

src/transformers/utils/generic.py
```diff
@@ -210,6 +210,27 @@ def is_mlx_array(x):
...
+
+    The different versions of flash attention are usually
+    - Implementations based on the original flash attention repo: https://github.com/Dao-AILab/flash-attention
+    - Kernels implementations such as: https://huggingface.co/kernels-community/vllm-flash-attn3
+    """
+    checked_attention_implementation = requested_attention_implementation
+    if checked_attention_implementation is None:
+        if config and hasattr(config, "_attn_implementation"):
```
Cyrilvallez: nit: `_attn_implementation` is guaranteed to exist so we don't need to have this check

Cyrilvallez (approved):
Very nice, thanks a lot! It will truly make it easier!

src/transformers/utils/generic.py
vasqu: Yea makes sense, changed the logic and description to force to pass just one not both

src/transformers/models/edgetam/modeling_edgetam.py
vasqu: Discussed this with @yonigozlan internally. This feature seems to be rarely used (it should be None in most cases) + these models often have several different Attention mechanisms; we want to fallback here to support FA partially at least (similarly done in SAM3)

PR #43124, opened 2026-01-06
Use timm-side buffers initialization

Cyrilvallez:
# What does this PR do?

As per the title. Clean-up timm buffers initialization that was temporarily added in https://github.com/huggingface/transformers/pull/42941. It has now been upstreamed directly in `timm`, so let's simplify on the Transformers side!


rwightman:
@Cyrilvallez shouldn't this use `init_non_persistent_buffers()` instead of `_init_buffers()`?

A related FYI, there was a missed import error (in a deprecated path) due to a moved module location, still have to support those as lots of code uses it so will be doing a 1.0.24 release very shortly to fix it.

ArthurZucker (approved):

PR #43127, opened 2026-01-06
[Quantization] torchao serialization

MekkCyber:
# What does this PR do?

Fixes torchao serialization, some tests are still failling for Int8/Int4 quantization with: 

```
ValueError: Unsupported tensor type: <class 'torchao.dtypes.affine_quantized_tensor.AffineQuantizedTensor'>
``` 

which seems to be a problem on torchao's end


SunMarc (approved):
Thanks !

PR #43130, opened 2026-01-06
[mimi] mimi conv cache edge case

eustlb:
# What does this PR do?

two things;
1. factorize the cache init so that it can be overwritten
2. handle the specific edge case with short hidden state: if lenght is < from padding, then the cache init should be taking into account in the padding states. This edge case is never matched with mimi but happens with other models that would use this cache approach (#40546 )

### DIFF
--- a/src/transformers/models/kyutai_speech_to_text/modeling_kyutai_speech_to_text.py
+++ b/src/transformers/models/kyutai_speech_to_text/modeling_kyutai_speech_to_text.py
@@ -150,9 +150,38 @@ class KyutaiSpeechToTextConv1dPaddingCache:
         self.per_layer_padding = per_layer_padding
         self.per_layer_padding_mode = per_layer_padding_mode
         self.per_layer_in_channels = per_layer_in_channels
-        self.per_layer_is_init = [True] * num_layers
 
         self.padding_cache = [None] * num_layers
+
+    def _cache_init(self, hidden_states: torch.Tensor, layer_idx: int):
+        """
+        Initialize the cache for a specific layer.
+
+        Parameters:
+            hidden_states (`torch.Tensor`):
+                The hidden states to initialize the cache with.
+            layer_idx (`int`):
+                The index of the layer to initialize the cache for.
+        Returns:
+            `torch.Tensor`, the initialized cache.
+        """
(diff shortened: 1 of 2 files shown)


src/transformers/models/kyutai_speech_to_text/modeling_kyutai_speech_to_text.py
```diff
@@ -150,10 +150,39 @@ def __init__(
         self.per_layer_padding = per_layer_padding
         self.per_layer_padding_mode = per_layer_padding_mode
         self.per_layer_in_channels = per_layer_in_channels
-        self.per_layer_is_init = [True] * num_layers
```
eustlb, line 153: this is not used anywhere, let's remove

src/transformers/models/kyutai_speech_to_text/modeling_kyutai_speech_to_text.py
```diff
@@ -150,10 +150,39 @@ def __init__(
...
+        )
+
+        if padding_mode == "constant":
+            current_cache = torch.zeros(batch_size, in_channels, padding, device=device, dtype=dtype)
+        elif padding_mode == "replicate":
+            current_cache = (
+                torch.ones(batch_size, in_channels, padding, device=device, dtype=dtype) * hidden_states[..., :1]
+            )
+        else:
+            raise NotImplementedError(f"Padding mode {padding_mode} not supported")
```
ebezzam, line 178: Good idea to add `_cache_init` so that different models can initialize differently if needed

How about removing the check [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/mimi/modeling_mimi.py#L100-L103) inside init? Since you have a similar one in this new method?

eustlb: Agree we could remove, but also it can be convenient to have raise an error at init to avoid having to run the full forward

eustlb: hum actually you're right, overwise it will be inconvenient and we'll have to also overwrite the init when working with other modes! thanks, removing

ebezzam (approved):

PR #43134, opened 2026-01-06
[`Ernie 4.5 VL Moe`] Fix non contiguous params

vasqu:
As per title, this is caused by the `transpose` conversion op which can cause non-contiguous layouts (kind of expected tbh). Simply forcing contiguous solves this.

Repro:
```python
from transformers import AutoModelForImageTextToText

model = AutoModelForImageTextToText.from_pretrained(
    "baidu/ERNIE-4.5-VL-28B-A3B-PT",
    device_map="auto",
    dtype="auto",
    revision="refs/pr/10",
)

assert all(param.is_contiguous() for param in list(model.parameters())), "Some parameters are not contiguous"
```


Cyrilvallez (approved):
LGTM, thanks!!

PR #43135, opened 2026-01-06
Fix multi-label prediction thresholding in run_classification

sumukhacharya03:
Fixes incorrect multi-label prediction thresholding that caused empty
predictions in predict_results.txt for multi-label classification.


Rocketknight1:
`expit` is the sigmoid function and `sigmoid(0) = 0.5`, so `expit(p) > 0.5` and `p > 0` are the same boolean tensor. I'm not sure what this changes!

PR #43138, opened 2026-01-06
Train sampler unification

jiosephlee:
# What does this PR do?

Adds ability to control the sampling strategy of data during training, unifying `group_by_length` with the default random sampling, and additionally adds sequential sampling. Following up on previous PR #42265 

@SunMarc


jiosephlee:
This is my first PR, so I'm not too familiar with this process, but there's a natural consequence that this would outdate a lot of documentation. Should I go ahead and update all previous mentions of `group_by_length`?

SunMarc (approved):
Thanks, this looks good ! Please go ahead and update the docs

romitjain:
@jiosephlee @SunMarc 
Should we instead make the argument to accept a sampler function in addition to a strategy? That will enable users integrate their own samplers. WDYT?

jiosephlee:
@romitjain I agree that this will make the Trainer more flexible. There are, however, general hesitations to adding arguments and bloating Trainer even further.

Thoughts on adding an additional argument called `train_sampler_fn` that overrides the strategy when provided? @SunMarc? Also, all the docs that referenced group_by_length have been updated to use the new arg

ArthurZucker (approved):

ArthurZucker:
@SunMarc should I merge?

romitjain:
@ArthurZucker I would appreciate your opinion on this
https://github.com/huggingface/transformers/pull/43138#issuecomment-3776215277

SunMarc:
Pls merge @ArthurZucker 


> @romitjain I agree that this will make the Trainer more flexible. There are, however, general hesitations to adding arguments and bloating Trainer even further.
> 
> Thoughts on adding an additional argument called train_sampler_fn that overrides the strategy when provided? @SunMarc? Also, all the docs that referenced group_by_length have been updated to use the new arg

Maybe in the future when users request this ! They can also overwrite the `_get_train_sampler` directly !

SunMarc:
plz merge @ArthurZucker

jiosephlee:
@ArthurZucker should be ready to merge again

winglian:
@SunMarc Did we not want a deprecation period/notice on removing this from TraningArgs?

SunMarc:
Do you think this is a widely used args ? Happy to add it back otherwise. This PR was supposed to be merged before v5 but it got forgotten.

PR #37162, opened 2025-04-01
Fix load of rng state for resuming training from checkpoint

winglian:
# What does this PR do?
Fixes a regression from https://github.com/huggingface/transformers/pull/36991 whereby resuming training from a checkpoint would fail when attempting to load the rng state. The regression only affects torch versions < 2.6.0.

see https://github.com/axolotl-ai-cloud/axolotl/actions/runs/14185726834/job/39740935566 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

@zach-huggingface and @SunMarc

PR #43140, opened 2026-01-07
Fix unsafe torch.load() in _load_rng_state allowing arbitrary code execution

ColeMurray:
Add weights_only=True to torch.load() call in Trainer._load_rng_state() to prevent arbitrary code execution when loading malicious checkpoint files.

The existing safe_globals() context manager provides no protection for PyTorch < 2.6 as it returns contextlib.nullcontext(). This makes the torch.load() call at line 3059 vulnerable to pickle deserialization attacks, unlike all other torch.load() calls in the same file which correctly use weights_only=True.

Impact: Users loading untrusted checkpoints on PyTorch 2.2-2.5 are vulnerable to arbitrary code execution via malicious rng_state.pth files.


src/transformers/trainer.py
```diff
@@ -3056,7 +3056,7 @@ def _load_rng_state(self, checkpoint):
                 return
 
         with safe_globals():
-            checkpoint_rng_state = torch.load(rng_file)
+            checkpoint_rng_state = torch.load(rng_file, weights_only=True)
```
vasqu, line 3060:
```suggestion
            check_torch_load_is_safe()
            checkpoint_rng_state = torch.load(rng_file, weights_only=True)
```
We disallow this for `torch<2.6` (as it's only properly fixed then), must have slipped through

vasqu (approved):
LGTM, let's just add the quick check fn before

vasqu:
Sorry about the CI @ColeMurray, it is unrelated to your PR working on fixing it

vasqu:
Thx for sticking through, merging now!

vasqu:
Just wanted to check the history, seems like #37162 reintroduced removing that kwarg but we shouldn't allow it - i.e. `check_torch_load_is_safe()` was missing to throw an error in those cases

Can we safely do this instead, we really shouldn't allow torch<2.6 here cc @SunMarc @winglian

SunMarc (approved):
Thanks ! With the hard check, the issue that @winglian had should be also fixed as users know that they need to upgrade their version of torch. Even if it is a bit breaking, it should be fine for v5.

vasqu:
Perfect, merging then thx for checking in

PR #43143, opened 2026-01-07
Inherit from PreTrainedTokenizerBase

juliendenize:
# What does this PR do?

Hi !

This PR is another patch on the MistralCommonBackend. It does the following:
- `MistralCommonBackend` inherits from `PretrainedTokenizerBase`. This is to make the backend more Transformers native. Ultimately it would be nice to support this Backend in libraries such as TRL but it'd need to take care of loading the correct `ValidationMode` when training.
- Refactor necessary attributes and methods to ensure inheritance is correctly done. Irrelevant methods for `MistralCommonBackend` raise `NotImplementedError`.
- Add methods similar to `PythonBackend` to handle token types, special tokens to add, etc.
- Add necessary tests.
- Remove usage of deprecated attributes in `mistral-common` and some private attribute calls. There are few left, ultimately the goal is to expand `mistral-common` API to remove all calls to private methods are attributes.
- Use mistral-common util function to find local path instead of keeping duplicated code in Transformers.
- Upgrade mistral-common required version to 1.8.8 
- Fix kwargs arguments passed in `PretrainedTokenizerBase.encode`.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@patrickvonplaten for viz

@ArthurZucker and @itazap


src/transformers/tokenization_mistral_common.py
```diff
@@ -149,8 +155,35 @@ class MistralTokenizerType(str, Enum):
...
+    "pad_token": SpecialTokens.pad.value,
+    "unk_token": SpecialTokens.unk.value,
+}
+
+_VALID_INIT_KWARGS = {"_from_auto", "backend", "files_loaded"}
+
+
 @requires(backends=("mistral-common",))
-class MistralCommonBackend(PushToHubMixin):
+class MistralCommonBackend(PreTrainedTokenizerBase):
```
ArthurZucker, line 186: thanks! its a big win!

src/transformers/tokenization_mistral_common.py
```diff
@@ -1987,6 +1597,50 @@ def _get_validation_mode(mode: Union[str, ValidationMode]) -> ValidationMode:
...
+        raise NotImplementedError("`MistralCommonBackend` does not implement `add_special_tokens`.")
+
+    def add_tokens(  # type: ignore[override]
+        self,
+        special_tokens_dict: dict[str, Union[str, AddedToken, Sequence[Union[str, AddedToken]]]],
+        replace_extra_special_tokens: bool = True,
+    ):
+        r"""`MistralCommonBackend` does not implement `add_special_tokens`."""
+
+        raise NotImplementedError("`MistralCommonBackend` does not implement `add_tokens`.")
```
ArthurZucker, line 1606: Ok! Perfect this justifies not inheriting from python backend, and makes sense, adding tokens introduce brittle behaviours!

ArthurZucker (approved):
LGTM! Simplifies nicely thanks for the effort! 
I would add maybe "by design" to the non supported features!

itazap:
Looking forward to this addition, thanks!
I’d also be interested in any feedback on what it was like to add a new tokenizer backend in v5, especially how extensible the current design felt - if you have any comments!

src/transformers/tokenization_mistral_common.py
juliendenize: Exactly, I first considered inheriting from it as I had to copy paste some behavior but it was "forcing" me to define more functions and it created a slow down of about 5% that I didn't manage to locate when encoding. May be due to my own code though and not related to how things are done in `PythonBackend`, didn't push the investigations that far as in the end it made more sense to inherit from the base class :)

juliendenize:
@ArthurZucker

> I would add maybe "by design" to the non supported features!

Good point ! I added in the doc by design and some explications to the ones we won't support and suggest to open issues for the one that we don't see the point yet but if there are enough demands then we might do (Adding tokens or special tokens).

@itazap 
> Looking forward to this addition, thanks! I’d also be interested in any feedback on what it was like to add a new tokenizer backend in v5, especially how extensible the current design felt - if you have any comments!

Felt definitely nice to be able to inherit directly from the base class and delete some duplicated code. As we don't support pair of sequences nor adding tokens for Mistral models most of the code is to never allow this behavior. We prefer having a class raising errors instead of allowing behavior that is undefined in our case.
I think having private methods to implement such as `_decode` was a good idea for other backends. To me the difficulty still lies to have to find arguments that are hidden in `kwargs` of the  `__init__` or `from_pretrained` method for example. They do a lot of stuff with half of it hidden by not explicitly passed arguments.
As we can see in the PR, the number of lines decreased which is a good ! Inhereting from `PythonBackend` would be a nice future step but for now I believe it is not easily customizable and define too many methods and attributes not necessary to us at the time.

src/transformers/tokenization_mistral_common.py
ArthurZucker: most probably due to having to regex for all special tokens on any string -> to split the string on them!

ArthurZucker:
Failing (torch) is unrelated will merge once the rest is all unrelated don't worry

ArthurZucker:
Ty @juliendenize

PR #43144, opened 2026-01-07
[style] Rework ruff rules and update all files

Cyrilvallez:
# What does this PR do?

As per the title. Time to update type annotations mostly!

Removed the `C` group that only matches C901 that was removed as well, added C4 for all unnecessary comprehensions.

Removed most exceptions we had, that should really not be excluded, especially:
- F402 -> Checks for import bindings that are shadowed by loop variables.
- UP004 -> Checks for classes that inherit from object.
- UP028 -> Checks for for loops that can be replaced with yield from expressions.
- UP045 -> Check for typing.Optional annotations that can be rewritten based on PEP 604 syntax.
- UP007 -> Check for type annotations that can be rewritten based on PEP 604 syntax.
- F823 -> Checks for undefined local variables.
- SIM300 -> Checks for conditions that position a constant on the left-hand side of the comparison operator, rather than the right-hand side.
- SIM212 ->  Checks for if expressions that check against a negated condition.
- UP009 -> Checks for unnecessary UTF-8 encoding declarations.


setup.py
```diff
@@ -122,7 +122,7 @@
     "rhoknp>=1.1.0,<1.3.1",
     "rjieba",
     "rouge-score!=0.0.7,!=0.0.8,!=0.1,!=0.1.1",
-    "ruff==0.13.1",
+    "ruff==0.14.1",
```
Sai-Suraj-27: We can bump to the latest [0.14.10](https://github.com/astral-sh/ruff/releases) version? 🚀

Cyrilvallez: Oh indeed, was on ruff website and read `0.14.1` lol
Will switch!

PR #43001, opened 2025-12-22
fix failure of pytest tests/models/paddleocr_vl/test_modeling_paddleo…

sywangyi:
…cr_vl.py::PaddleOCRVLModelTest::test_flash_attn_2_fp32_ln

quantization_config is not set in text and visual part. so datatype conversion before flash attn will not be called.

- vision models: @yonigozlan @molbap
- multimodal models: @zucchini-nlp

```
pytest tests/models/paddleocr_vl/test_modeling_paddleocr_vl.py::PaddleOCRVLModelTest::test_flash_attn_2_fp32_ln
```
failure like
```
../bk/hub/models--kernels-community--flash-attn2/snapshots/172e23272e585d3c0d97124bc690593af81a0b95/build/torch29-cxx11-cu128-x86_64-linux/flash_attn2/flash_attn_interface.py:1199: in flash_attn_func
    return FlashAttnFunc.apply(
/mnt/disk0/wangyi/miniforge3/envs/transformers/lib/python3.11/site-packages/torch/autograd/function.py:581: in apply
    return super().apply(*args, **kwargs)  # type: ignore[misc]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../bk/hub/models--kernels-community--flash-attn2/snapshots/172e23272e585d3c0d97124bc690593af81a0b95/build/torch29-cxx11-cu128-x86_64-linux/flash_attn2/flash_attn_interface.py:837: in forward
    out_padded, softmax_lse, S_dmask, rng_state = _wrapped_flash_attn_forward(
/mnt/disk0/wangyi/miniforge3/envs/transformers/lib/python3.11/site-packages/torch/_ops.py:1255: in __call__
    return self._op(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
/mnt/disk0/wangyi/miniforge3/envs/transformers/lib/python3.11/site-packages/torch/_library/autograd.py:111: in autograd_impl
    result = forward_no_grad(*args, Metadata(keyset, keyword_only_args))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/mnt/disk0/wangyi/miniforge3/envs/transformers/lib/python3.11/site-packages/torch/_library/autograd.py:40: in forward_no_grad
    result = op.redispatch(keyset & _C._after_autograd_keyset, *args, **kwargs)
[...]

PR #43145, opened 2026-01-07
Fix flashattn wrt quantized models

SunMarc:
# What does this PR do?

This PR fixes an issue with `config` and `quantization_config`. Since we don't propagate the `quantization_config` into subconfig, we don't cast the tensors to the correct dtype when using flash attn. However, I prefer not to propagate this information for now. Instead, we just set _is_quantized flag. In the past, we were using `_pre_quantization_dtype`. 

Fixes https://github.com/huggingface/transformers/pull/43001

Regression caused by https://github.com/huggingface/transformers/pull/42882/files


MekkCyber (approved):
Thanks for fixing

PR #43146, opened 2026-01-07
run all test files on CircleCI

ydshieh:
# What does this PR do?

Maybe we need to verify that there is at least one python files being modified.

(i.e. if only doc MD files are modified, don't run)


Cyrilvallez (approved):
Thanks a lot! I think it will help quite a bit!

Cyrilvallez (approved):
Very nice, thanks!

Cyrilvallez (approved):
Let's go!! Thanks a lot!

PR #43147, opened 2026-01-07
[CB] Minor fix in kwargs

remi-or:
This PR fixes an error in the `fork` method of the continuous batching API. The arguments name were lacking a `_`


vasqu (commented):
The tests run now but they dont pass locally (on our dgx a100, torch 2.9.0; weirdly 2.6.0 passes). On which GPU have you checked?

The other failing CI tests seem unrelated to me, seems to be a general issues on main rn

remi-or:
> The tests run now but at least locally they seem at least flaky to me (on our dgx a100). On which GPU have you checked?

Checked on H100, what tests were flaky?

vasqu:
> what tests were flaky

Sorry switched torch versions, torch 2.6 passed case 0 but not case 1 and torch 2.9 they both failed locally (`test_num_return_sequences`). I suspect a GPU diff here so our CI (A10) still might fail here then

vasqu (approved):
Sorry still approving as it affects general usage overall

Can we add a todo to the test at least, commenting about the GPU diff here

PR #43150, opened 2026-01-07
[CB] Soft-reset offloading

remi-or:
# Summary
This PR adds a basic form of offloading to the continuous batching API. Currently, when we cannot schedule any request because the KV cache is full, we throw an error, which is less than optimal. This PR fixes this by adding a mechanism where instead of throwing an error, we de-schedule the oldest request by adding it back to the list of waiting request, but keeping the generated tokens by adding them to the initial prompt. Then, no new request can be scheduled until one is completed, which helps avoid infinite loops on scheduled-offloaded requests.

## Performance 

| Arguments                                          | Throughput on main | Throughput with PR | Delta   |
|----------------------------------------------------|--------------------|--------------------|---------|
| --samples 10                                       | 532.98             | 530.45             | -2.53   |
| --samples 20 --num-blocks 16                       | ERROR              | 100.14             | N/A     |
| --samples 50                                       | 1478.16            | 1477.42            | -0.74   |
| --samples 100                                      | 2254.38            | 2266.81            | +12.43  |
| --samples 100 --attn flash_attention_2             | 2062.56            | 2076.81            | +14.25  |
| --samples 100 --attn sdpa                          | 866.01             | 840.72             | -25.29  |
| --samples 500                                      | 5299.24            | 5281.98            | -17.26  |
| --samples 500 --add-prefix --compile               | 6774.53            | 6743.21            | -31.32  |
| --samples 50 --num-return-sequences 8 --do-sample  | 5319.79            | 5362.52            | +42.73  |
| --samples 100 --num-return-sequences 4 --do-sample | 4680.87            | 4674.91            | -5.96   |

Throughput are in tok/s. The script run is `examples/pytorch/continuous_batching.py`

## Tests
Tests pass, including the tests added for this PR.

## Sanity check
Looks good.


src/transformers/generation/continuous_batching/cache.py
```diff
@@ -237,16 +244,27 @@ def __init__(
         self.blocks_to_complete: dict[str, int] = {}
         self._total_prefix_length: int = 0  # a counter to measure the impact of prefix sharing, also used in tests
 
+    def will_allocation_be_successful(self, num_requested_blocks: int, allocated_blocks: int) -> bool:
+        needed_blocks = num_requested_blocks * self.num_full_attention_groups
+        if self.num_sliding_attention_groups:
+            blocks_left = max(self.max_sliding_window_blocks_per_request - allocated_blocks, 0)
+            needed_blocks += min(blocks_left, num_requested_blocks) * self.num_sliding_attention_groups
+        return needed_blocks <= self.get_num_free_blocks()
```
ArthurZucker, line 263: i really like this! very explicit what is happening, how many blocks you need. 

Just that there is requested in terms of token and needed in terms of attention type, I would probably document!

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -206,6 +206,20 @@ def free_blocks(self, blocks: list[int], shareable: bool) -> None:
         else:
             self._uninit_block_ids.extend(blocks)
 
+    def uninitialize_redundant_block(self, block_id: int) -> None:
```
```suggestion
    def reset_redundant_block(self, block_id: int) -> None:
```
if this is what it is doing

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -427,6 +427,29 @@ def _handle_request_error(self, error: Exception, state: RequestState) -> None:
         self.metrics.record_request_completion(state.created_time, state.request_id)
         self.output_queue.put(state.to_generation_output())
 
+    def soft_reset_one_request(self) -> None:
+        """Soft resets one active request by removing it from active requests and re-adding it to the waiting queue.
+
+        The generated tokens are kept as part of the new request's initial prompt. When `block_new_requests` is False,
```
ArthurZucker, line 435:
```suggestion
        The generated tokens are kept as part of the new request's initial prompt. When `offload_new_requests` is False,
```

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -427,6 +427,29 @@ def _handle_request_error(self, error: Exception, state: RequestState) -> None:
...
+        """
+        # The offloaded request is the newest (resp. oldest) if block_new_requests is True (resp. False)
+        if self.scheduler.block_new_requests:
+            request_id, state = self.scheduler.active_requests.popitem()
+        else:
+            request_id, state = next(iter(self.scheduler.active_requests.items()))
+        logger.info(
+            f"Soft resetting request {request_id} with {len(state.initial_tokens)} initial tokens and "
+            f"{len(state.generated_tokens)} generated tokens"
+        )
```
ArthurZucker, line 447: Is there a reason why we don't have an offloading criteria (appart from oldest / newest)? 

I don't know what is best, for server cases, you want to offload new, but for eval maybe not? Or is it always better?

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -427,6 +427,29 @@ def _handle_request_error(self, error: Exception, state: RequestState) -> None:
...
+        # Create a copy of the offloaded request keeping the generated tokens as addition to the initial prompt
+        new_state = state.create_equivalent_initial_request()
+        # Actual offloading of the request
+        self.scheduler.finish_request(request_id, evict_from_cache=True)
+        self.scheduler.add_waiting_request(new_state)
+        self.scheduler.block_new_requests = True
```
ArthurZucker, line 455: we want to block current scheduling, but we want next prepare next batch to still try to schedule, checking if this is done

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -591,6 +623,7 @@ def update_batch(self) -> None:
                 if is_finished:
                     self.metrics.record_request_completion(state.created_time, state.request_id)
                     self.scheduler.finish_request(state.request_id, evict_from_cache=(not self.manual_eviction))
+                    self.scheduler.block_new_requests = False
```
ArthurZucker, line 630: ok if any one request is finished we can accept new request SG

src/transformers/generation/continuous_batching/continuous_api.py
ArthurZucker: like should we offload the biggest ?

src/transformers/generation/continuous_batching/requests.py
```diff
@@ -220,6 +221,9 @@ def __repr__(self):
 
     def to_generation_output(self):
         """Convert the request state to a GenerationOutput object."""
+        if self._true_initial_tokens:
+            self.generated_tokens = self.initial_tokens[self._true_initial_tokens :] + self.generated_tokens
+            self.initial_tokens = self.initial_tokens[: self._true_initial_tokens]
```
ArthurZucker, line 227: maybe initi_tokens stayls as is and we just have `prefill_tokens` or something

src/transformers/generation/continuous_batching/scheduler.py
```diff
@@ -247,26 +285,32 @@ def schedule_batch(self, token_budget: int) -> list[RequestState]:
...
 
+        # If no requests were scheduled and the cache is full, we signal it by returning None
+        if not scheduled_requests and one_allocation_failed:
+            return None
+
         return scheduled_requests
 
 
 # FIXME: prioritize adding from waiting reqs before scheduling `RequestStatus.DECODING` when cache space allows it
+# TODO: further consolidate the code by making more of it common. The reference Scheduler is FIFO, not this one.
```
ArthurZucker, line 305: indeed

ArthurZucker (approved):
Missing a doc example of how to enable / why (or do we enable it by default)

src/transformers/generation/continuous_batching/cache.py
remi-or: Done! Good point, left a long docstring.

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: Not really, as the method checked the block was redundant. Move some things around and renamed to a more appropriate name.

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: The name is actually appropriate: the `block_new_requests` flag literally blocks the scheduler from picking up new requests, to avoid loops!

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: I am adding a TODO, but I guess we should have a few different offload policy as we do with the scheduling policy. This version is quite bare-bones for offloading, it just ensures nothing is going to crash, so that's the scope Im going for. I think I will test different policy once we have a model / dataset in mind!

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: At this point, there is no current scheduling, because all scheduling failed. What this does is to block any future scheduling until at least one request finishes. Adding a comment to make this more clear.

src/transformers/generation/continuous_batching/requests.py
remi-or: Good names, adding a TODO to not clog-up this PR and will do this in a downstream PR

remi-or:
We enable this per default because the alternative is crashing. When we add proper `OffloadPolicy` and the parameters around that, I will add a `NoOffload` policy which crashes, for the folks who like that.

PR #43152, opened 2026-01-07
[`Moe`] Pin more models to eager

vasqu:
As per title, some models still fluctuate keeping an eye out for a week to see if more models need to be added - e.g. qwen 3 omni moe was not caught in the initial nightly ci but the 2nd day

Keeping this as a draft to wait for potentially more models, DO NOT MERGE yet


vasqu:
run-slow: qwen3_omni_moe

vasqu:
Seems like that test was flaky either way so closing it

PR #43154, opened 2026-01-07
[`Fp8`] Fix experts

vasqu:
As per title, the current fp8 experts implementation is wrong - likely related to #42456 (cc @3outeille)

Before:
- The number of experts was treated as the number of `top k` experts that are hit, i.e. `num_experts = top_k_weights.shape[1]`
  - This is simply wrong as now our indices which can cover all experts is limited to `top_k`
  - This can trigger device assert errors as one hot might not have the correct indices for the number of classes indicated by num_experts
- The hit expert is now used which is correct if we passed the selection before
  - Only due to the indices being correctly limited 

After:
- We now follow the current eager implementation, e.g. Mixtral: https://github.com/huggingface/transformers/blob/88a5623361e1b3d844daef3c6c95535d12e70056/src/transformers/models/mixtral/modeling_mixtral.py#L82-L86
- The portion where we iterate through experts should only have changed the linear calls of https://github.com/huggingface/transformers/blob/88a5623361e1b3d844daef3c6c95535d12e70056/src/transformers/models/mixtral/modeling_mixtral.py#L88-L98
  - The key is to redirect the correctly hit expert to the top k weight via https://github.com/huggingface/transformers/blob/88a5623361e1b3d844daef3c6c95535d12e70056/src/transformers/models/mixtral/modeling_mixtral.py#L92

I'm not sure how long it has been this way, but it definitely is not correct atm. I'd be curious if I'm overseeing something here but without this fix #42028 will not work for the original model - my suspicion is on mixtral having a bias towards low indices which might have saved this oversight.


ArthurZucker (approved):
Yes absolutely! I did not check but if a test is missing let's add one!

ArthurZucker:
@3outeille note that this breaks TP but same as it is broken for now as well

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -535,18 +535,17 @@ def forward(
...
             expert_idx = expert_idx[0]
-            if expert_idx == num_experts:
+            if expert_idx == self.num_experts:
                 continue
-            _, token_idx = torch.where(expert_mask[expert_idx])
-            current_state = hidden_states.index_select(0, token_idx)
+            top_k_pos, token_idx = torch.where(expert_mask[expert_idx])
+            current_state = hidden_states[token_idx]
             gate, up = self.linear(
                 current_state, self.gate_up_proj[expert_idx], self.gate_up_proj_scale_inv[expert_idx]
```
SunMarc, line 552: thanks ! maybe we can add a small comment to say that it was mostly copied from deepspeed_v3 modeling, so that we should propagate the changes here also in the future

SunMarc (approved):
LGTM

src/transformers/integrations/finegrained_fp8.py
vasqu: Added a comment, referencing mixtral

tests/quantization/finegrained_fp8/test_fp8.py
```diff
@@ -368,6 +376,38 @@ def test_compute_module_sizes(self):
         # we should at least have 1.5 times memory reduction in total
         assert model_size[""] > quantized_model_size[""] * 1.5
 
+    @unittest.skip(reason="Dependent on #42028, will be removed alongside that PR")
+    def test_quantized_moe_forward(self):
```
vasqu, line 380: This acts as a sanity integration check but it depends on the minimax m2 PR (#42028) so I will remove this skip when merging that PR

I think this is the easiest way as these weights force the issue

PR #43158, opened 2026-01-07
[docs] embed dataset viewer

stevhliu:
replaces large markdown tables with the cleaner dataset viewer

Screenshot 2026-01-07 at 2 24 23 PM

### DIFF
--- a/docs/source/en/experts_interface.md
+++ b/docs/source/en/experts_interface.md
@@ -110,15 +110,15 @@ model.forward = torch.compile(model.forward, mode="max-autotune-no-cudagraphs")
 This [benchmark](https://github.com/user-attachments/files/24125816/bench.py) compares different input sizes and experts implementations with and without `torch.compile`.
 
 <iframe
-  src="https://huggingface.co/datasets/docs-benchmarks/experts-backends/embed/viewer/bs1-seq16/train"
-  frameborder="0"
-  width="100%"
-  height="560px"
-></iframe>
-
-<iframe
 	src="https://huggingface-moe-experts-benchmarks.static.hf.space"
 	frameborder="0"
 	width="100%"
 	height="800"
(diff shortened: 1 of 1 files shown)


IlyasMoutawwakil:
Love the idea @stevhliu !
Can you please update it with these results (include prefill, decode and generate times)
json: [results.json](https://github.com/user-attachments/files/24490807/results.json)
md: [results.md](https://github.com/user-attachments/files/24490813/results.md)
For reproduction we can link to this script: https://gist.github.com/IlyasMoutawwakil/8b70f5604c60efb56fcf8565fbdb05a5
We can also add these plots to summarize the numbers (plotting script in https://gist.github.com/IlyasMoutawwakil/6040b1e6803bf959138933e44fd93af2)
generate
decode
prefill

stevhliu:
updated the dataset with new results and thinking about embedding this chart in there as well (trying to figure out how to make it work on Spaces atm)!


https://github.com/user-attachments/assets/a3a38891-5c83-4866-8a41-e3408fba99b0

stevhliu:
@IlyasMoutawwakil, updated with the [plots](https://moon-ci-docs.huggingface.co/docs/transformers/pr_43158/en/experts_interface#benchmarks) as well now. let me know what you think!

IlyasMoutawwakil:
thanks @stevhliu ! isn't it a bit hard to read ? the generation / prefill / decode latencies are different metrics so it's counterintuitive to see them next to each other in the same bar plot, there is a way to put them in the same plot if that's the goal ; by stacking prefill and decode = generate:
prefill+decode
but we also lose the info about the speedup which is imo very eye-catching and straight-to-the-point

IlyasMoutawwakil (approved):
LGTM ! awesome docs !

PR #43166, opened 2026-01-08
Add Youtu-LLM model

LuJunru:
# What does this PR do?

This PR adds the implementation for the released Youtu-LLM model. The model has the following features:
- Type: Autoregressive Causal Language Models with Dense MLA
- Release versions: [Base](https://huggingface.co/tencent/Youtu-LLM-2B-Base) and [Instruct](https://huggingface.co/tencent/Youtu-LLM-2B)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker @Cyrilvallez


LuJunru:
> View the CircleCI Test Summary for this PR:
> 
> https://huggingface.co/spaces/transformers-community/circle-ci-viz?pr=43166&sha=5dab39

Hi @ArthurZucker @Cyrilvallez 

May I ask if it is possible to concentrate the test only on Youtu-LLM (the new model)? The summary here seems report errors raised by other models.
截屏2026-01-08 19 01 30

junru

LuJunru:
Hi @ArthurZucker @Cyrilvallez 

It seems Youtu-LLM-related codes have passed the auto review. The remaining check fails on other models.
截屏2026-01-09 09 49 13

src/transformers/models/youtu/modular_youtu.py
molbap, line 1: is the official name `YoutuLLM` or `Youtu` as in the prefixes here?

tests/models/youtu/test_modeling_youtu.py
```diff
@@ -0,0 +1,372 @@
...
+    @require_torch_large_accelerator
+    @slow
+    def test_eager_matches_sdpa_generate(self):
+        max_new_tokens = 30
+
+        tokenizer = AutoTokenizer.from_pretrained("tencent/Youtu-LLM-2B-Base")
+
+        model_sdpa = YoutuForCausalLM.from_pretrained(
+            "tencent/Youtu-LLM-2B-Base",
+            dtype=torch.float16,
```
molbap: let's make sdpa explicit here

molbap:
run-slow: youtu_llm

src/transformers/models/youtu_llm/modular_youtu.py
```diff
@@ -0,0 +1,83 @@
...
+            if module.bias is not None:
+                init.zeros_(module.bias)
+        elif isinstance(module, nn.Embedding):
+            init.normal_(module.weight, mean=0.0, std=embed_std)
+            if module.padding_idx is not None:
+                init.zeros_(module.weight.data[module.padding_idx])
+
+
+class YoutuModel(LlamaModel):
+    _keys_to_ignore_on_load_unexpected = [""]
```
molbap: is this to remove the Llama attribute? if so, ok

tests/models/youtu_llm/test_modeling_youtu.py
```diff
@@ -0,0 +1,372 @@
...
+class YoutuIntegrationTest(unittest.TestCase):
+    def tearDown(self):
+        # See LlamaIntegrationTest.tearDown(). Can be removed once LlamaIntegrationTest.tearDown() is removed.
+        cleanup(torch_device, gc_collect=False)
+
+    @slow
+    @require_torch_accelerator
+    @pytest.mark.torch_compile_test
+    @require_read_token
+    def test_compile_static_cache(self):
```
molbap: Thanks for adding an integration test! however naming-wise, seems to measure dynamic and static Cache no? By the way, could we have a simple no-compile integration test that works in the simplest setting, just to avoid regressions?

tests/models/youtu/test_modeling_youtu.py
```diff
@@ -0,0 +1,372 @@
...
+    def test_greedy_generate_dict_outputs_use_cache(self):
+        pass
+
+    @unittest.skip(reason="SDPA can't dispatch on flash due to unsupported head dims")
+    def test_sdpa_can_dispatch_on_flash(self):
+        pass
+
+    @unittest.skip(reason="Youtu-LLM is not suitable for testing with extreme small vocabulary")
+    def test_resize_tokens_embeddings(self):
+        pass
```
molbap: are all these tests indeed not working?

molbap (commented):
Seems clean, good modular file with simply Llama + MLA, beautiful. Asked a few questions, let me know and I'll re-review!

src/transformers/models/youtu/modular_youtu.py
LuJunru: We chose to use ```Youtu``` as the prefix of modules, as it is more suitable for extension (e.g., we plan to introduce ```YoutuVL``` in near future). ```Youtu-LLM``` is rather a brand name.

src/transformers/models/youtu_llm/modular_youtu.py
LuJunru: For the current version of the model (Youtu-LLM-2B family), this line of code could be removed.

tests/models/youtu_llm/test_modeling_youtu.py
LuJunru: We have provided inference tests below based on no-compile dynamic cache and no-compile static cache. Basically, I implemented this test function by referencing [test function of DeepSeek V3](https://github.com/huggingface/transformers/blob/main/tests/models/deepseek_v3/test_modeling_deepseek_v3.py#L393).

tests/models/youtu/test_modeling_youtu.py
LuJunru: Yes, exactly.

tests/models/youtu_llm/test_modeling_youtu.py
molbap: sure, can we update the name though to make it more clear and separate in two tests? that way if it breaks at some point it's easier to debug

LuJunru: Sure, is there any official examples that I can follow up?

LuJunru:
Hi @molbap 

I've updated a new version of code according to the discussion aforementioned. Can you help start a new solo test of Youtu-LLM (run-slow: youtu_llm)?

molbap:
run-slow: youtu_llm

tests/models/youtu/test_modeling_youtu.py
```diff
@@ -0,0 +1,423 @@
...
+            with self.subTest(f"{padding_side}"):
+                torch.testing.assert_close(
+                    res_eager,
+                    res_sdpa,
+                    msg=f"\n{tokenizer.batch_decode(res_eager)} \nvs\n{tokenizer.batch_decode(res_sdpa)}",
+                )
+
+
+@require_torch_accelerator
+class YoutuIntegrationTest(unittest.TestCase):
```
molbap, line 93:
```suggestion
@slow
class YoutuIntegrationTest(unittest.TestCase):
```

src/transformers/models/youtu/__init__.py
molbap, line 1: if the name is `youtu`, let's rename the directory thus as well. Any name is fine as long as we're keeping it consistently!

src/transformers/models/youtu/modular_youtu.py
molbap: then everything that has to see with model name (youtu) should be named as such, like the model directory

molbap (approved):
Sounds good to me, approving to move to core maintainer review - I might iterate one more time depending on feedback. Clean work!

src/transformers/models/youtu/__init__.py
LuJunru: ok, updated

tests/models/youtu/test_modeling_youtu.py
LuJunru: done

src/transformers/models/youtu/configuration_youtu.py
```diff
@@ -0,0 +1,200 @@
...
+            The dropout ratio for the attention probabilities.
+    ```python
+    >>> from transformers import YoutuModel, YoutuConfig
+    >>> # Initializing a Youtu-LLM-2B style configuration
+    >>> configuration = YoutuConfig()
+    >>> # Accessing the model configuration
+    >>> configuration = model.config
+    ```"""
+
+    model_type = "youtu_llm"
```
xenova: This needs to also be changed to "youtu", right? @molbap

molbap: yes! indeed

LuJunru: @xenova @molbap 

Understand, but since we already used ```youtu_llm``` in the [open-sourced models](https://huggingface.co/tencent/Youtu-LLM-2B/blob/main/config.json#L22), can we keep this unchanged?

xenova: If you'd like to use the same repo as before, then indeed you would need to update `model_type` to `youtu` for it to match this PR. In that update, you would also remove the auto_map/custom code section:
```
"auto_map": {
  "AutoConfig": "configuration_youtu.YoutuConfig",
  "AutoModel": "modeling_youtu.YoutuModel",
  "AutoModelForCausalLM": "modeling_youtu.YoutuForCausalLM"
},
```

so any user of the new models will see these changes.

Alternatively, if you'd prefer not to update the existing model repos (in case this causes issues with users that have not fixed the revision used), you may want to upload separate checkpoints `tencent/Youtu-LLM-2B-hf` for example. Perhaps @molbap can provide some more guidance here :)

molbap: that's a good use-case actually, we want to be able to load from the hub despite model keys not being matched perfectly, that's the role of `conversion_mapping.py`. For the `model_type` indeed as Joshua says there's the solution to change the key in your model repos.

Based on that I think the most convenient naming would be simply `youtu_llm` for everything, but if you want `youtu` without changing hub repos... then I would suggest doing something like touching the `CONFIG_MAPPING_NAMES` to have non-unique mapping keys to `YoutuConfig`, so that we know `youtu_llm` is a valid key. It's not standard, but loading directly from the hub wasn't standard until recently, hence "good use case".

cc @hmellor as it is similar to model/config mappings with vLLM x transformers modelling backend, and cc core reviewer @vasqu too (for review and opinion on this)

LuJunru: Yes, I guess many new models, once open-sourced on the hub, will have a pre-defined ```model_type```. Anyway, I've updated a version that explicitly sets the ```model_type``` to ```youtu``` and uses a different testing repo from the hub. Once the PR is merged, I can update the official repository (tencent/Youtu-LLM-2B, etc.) and add additional notes about the ```model_type``` modification.

docs/source/en/model_doc/youtu.md
```diff
@@ -0,0 +1,138 @@
+<!--Copyright 2025 The HuggingFace Team. All rights reserved.
```
vasqu:
```suggestion
<!--Copyright 2026 The HuggingFace Team. All rights reserved.
```
happy new year :D

docs/source/en/model_doc/youtu.md
```diff
@@ -0,0 +1,138 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+*This model was released on 2025-12-31 and added to Hugging Face Transformers on 2026-01-21.*
```
vasqu: To keep in mind, CI will complaing about the addition time so needs to be updated at last time - just for viz

docs/source/en/model_doc/youtu.md
```diff
@@ -0,0 +1,138 @@
...
+
+# 1. Configure Model
+model_id = "tencent/Youtu-LLM-2B"
+
+# 2. Initialize Tokenizer and Model
+tokenizer = AutoTokenizer.from_pretrained(model_id, trust_remote_code=True)
+model = AutoModelForCausalLM.from_pretrained(
+    model_id,
+    device_map="auto",
+    trust_remote_code=True
```
vasqu: Let's remove `trust_remote_code`, would be nice if we could either rely on a pr on the hub (revision="...")

docs/source/en/model_doc/youtu.md
vasqu: seeing this elsewhere but don't want to cluster the review

src/transformers/models/youtu/modular_youtu.py
```diff
@@ -0,0 +1,83 @@
...
+
+class YoutuRMSNorm(LlamaRMSNorm):
+    pass
+
+
+class YoutuRotaryEmbedding(LlamaRotaryEmbedding):
+    pass
+
+
+class YoutuMLP(LlamaMLP):
```
vasqu: Can be inherited from qwen3 or gemma

src/transformers/models/youtu/modular_youtu.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self, config: YoutuConfig, layer_idx: int):
+        nn.Module.__init__(self)
+        self.hidden_size = config.hidden_size
+
+        self.self_attn = YoutuMLAttention(config=config, layer_idx=layer_idx)
+
+        self.mlp = YoutuMLP(config)
+
+        self.input_layernorm = YoutuRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        self.post_attention_layernorm = YoutuRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
```
vasqu: I guess it clashes because of `YoutuMLAttention`, I think we can simply name it to `YoutuAttention` -> then we have the same prefix everywhere.

Otherwise, even if we have a difference, we can still use `super().__init__(config, layer_idx)`. You can remove attributes via `del attr` or overwrite those directly `self.attr = SpecialClass(...)`.

src/transformers/models/youtu/modular_youtu.py
```diff
@@ -0,0 +1,83 @@
...
+class YoutuPreTrainedModel(LlamaPreTrainedModel):
+    @torch.no_grad()
+    def _init_weights(self, module):
+        super()._init_weights(module)
+        std = getattr(self.config, "initializer_range", 0.02)
+        embed_std = getattr(self.config, "embedding_initializer_range", 2 * std)
+        if isinstance(module, nn.Linear):
+            init.normal_(module.weight, mean=0.0, std=std)
+            if module.bias is not None:
+                init.zeros_(module.bias)
```
vasqu: Should be handled by super, no? https://github.com/huggingface/transformers/blob/4a1ad8da8d6b36a60269ef1deadc474e9243aec1/src/transformers/modeling_utils.py#L2252-L2256

src/transformers/models/youtu/modular_youtu.py
```diff
@@ -0,0 +1,83 @@
...
+        elif isinstance(module, nn.Embedding):
+            init.normal_(module.weight, mean=0.0, std=embed_std)
+            if module.padding_idx is not None:
+                init.zeros_(module.weight.data[module.padding_idx])
```
vasqu: This look like the only case because we have a different std here

tests/models/youtu/test_modeling_youtu.py
```diff
@@ -0,0 +1,423 @@
...
+    torch.set_float32_matmul_precision("high")
+
+    from transformers import (
+        Cache,
+        YoutuForCausalLM,
+        YoutuModel,
+    )
+
+
+class YoutuModelTester:
```
vasqu: Let's use the `CausalLMTester`, it should reduce the code and simplify a lot. See https://github.com/huggingface/transformers/blob/4a1ad8da8d6b36a60269ef1deadc474e9243aec1/tests/models/solar_open/test_modeling_solar_open.py#L36

You can add special values for MLA yourself there if needed

tests/models/youtu/test_modeling_youtu.py
```diff
@@ -0,0 +1,423 @@
...
+            sequence_labels,
+            token_labels,
+            choice_labels,
+        ) = config_and_inputs
+        inputs_dict = {"input_ids": input_ids, "attention_mask": input_mask}
+        return config, inputs_dict
+
+
+@require_torch
+class YoutuModelTest(ModelTesterMixin, GenerationTesterMixin, PipelineTesterMixin, unittest.TestCase):
```
vasqu: This will also move to our wrapper then, e.g. https://github.com/huggingface/transformers/blob/4a1ad8da8d6b36a60269ef1deadc474e9243aec1/tests/models/solar_open/test_modeling_solar_open.py#L65

tests/models/youtu/test_modeling_youtu.py
vasqu: Let's check if we can fix the majority by moving the tests under the CausalLM wrapper classes

tests/models/youtu/test_modeling_youtu.py
```diff
@@ -0,0 +1,423 @@
...
+                    res_sdpa,
+                    msg=f"\n{tokenizer.batch_decode(res_eager)} \nvs\n{tokenizer.batch_decode(res_sdpa)}",
+                )
+
+
+@slow
+class YoutuIntegrationTest(unittest.TestCase):
+    def tearDown(self):
+        # See LlamaIntegrationTest.tearDown(). Can be removed once LlamaIntegrationTest.tearDown() is removed.
+        cleanup(torch_device, gc_collect=False)
```
vasqu, line 95: Let's also do setup https://github.com/huggingface/transformers/blob/4a1ad8da8d6b36a60269ef1deadc474e9243aec1/tests/models/llama/test_modeling_llama.py#L338-L342

No need for comments here

tests/models/youtu/test_modeling_youtu.py
```diff
@@ -0,0 +1,423 @@
...
+            "Simply put, the theory of relativity states that ",
+            "My favorite all time favorite condiment is ketchup.",
+        ]
+        tokenizer = AutoTokenizer.from_pretrained("Junrulu/Youtu-LLM-2B-Base-hf")
+        model = YoutuForCausalLM.from_pretrained(
+            "Junrulu/Youtu-LLM-2B-Base-hf", device_map=torch_device, dtype=torch.float16
+        )
+        if model.config.tie_word_embeddings:
+            # Youtu-LLM-2B-Base contains extra repeated weights for the tied embeddings, we can tie weights here according to its config
+            model.tie_weights()
```
vasqu, line 120: Should happen automatically, let's remove(?) Don't really see the reason this is needed as it should be done during load time. Same for the other tests

utils/check_docstrings.py
```diff
@@ -508,6 +508,7 @@ class DecoratedItem:
     "YolosImageProcessor",
     "YolosModel",
     "YosoConfig",
+    "YoutuConfig",
```
vasqu: Let's not do this, we should aim to properly support this. Checking to fix

src/transformers/models/youtu/modular_youtu.py
```diff
@@ -0,0 +1,83 @@
...
+        self.mlp = YoutuMLP(config)
+
+        self.input_layernorm = YoutuRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        self.post_attention_layernorm = YoutuRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+
+
+class YoutuPreTrainedModel(LlamaPreTrainedModel):
+    @torch.no_grad()
+    def _init_weights(self, module):
+        super()._init_weights(module)
```
vasqu:
```suggestion
class YoutuPreTrainedModel(LlamaPreTrainedModel, PreTrainedModel):
    @torch.no_grad()
    def _init_weights(self, module):
        PreTrainedModel._init_weights(self, module)
```
super doesn't find this

utils/check_docstrings.py
vasqu: Found the culprits it's the docstrings in the config
```text
initializer_range (`float`, *optional*):
            The standard deviation of the truncated_normal_initializer for initializing all weight matrices, except embedding matrices.
        embedding_initializer_range (`float`, *optional*):
            The standard deviation of the truncated_normal_initializer for initializing all embedding matrices.
```
Will take another look there either ways

src/transformers/models/youtu/configuration_youtu.py
```diff
@@ -0,0 +1,200 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Youtu-LLM model configuration"""
+
+from ...configuration_utils import PreTrainedConfig
+from ...modeling_rope_utils import RopeParameters
+
+
+Youtu_PRETRAINED_CONFIG_ARCHIVE_MAP = {}
```
vasqu:
```suggestion
```
isnt used anymore

src/transformers/models/youtu/configuration_youtu.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+class YoutuConfig(PreTrainedConfig):
```
vasqu: Can we inherit from deepseekv3 or similar in the modular file? I had comment about how to delete attributes / overwrite attributes - this would massively simplify the config

src/transformers/models/youtu/configuration_youtu.py
```diff
@@ -0,0 +1,200 @@
...
+        num_key_value_heads: int | None = 16,
+        kv_lora_rank: int | None = 512,
+        q_lora_rank: int | None = 1536,
+        qk_rope_head_dim: int | None = 64,
+        v_head_dim: int | None = 128,
+        qk_nope_head_dim: int | None = 128,
+        hidden_act: str | None = "silu",
+        max_position_embeddings: int | None = 131072,
+        initializer_range: float | None = None,
+        embedding_initializer_range: float | None = None,
```
vasqu, line 133: Let's default to values directly, I don't see a reason why not

src/transformers/models/youtu/configuration_youtu.py
```diff
@@ -0,0 +1,200 @@
...
+        self.head_dim = qk_rope_head_dim
+        self.rope_interleave = rope_interleave
+
+        # for backward compatibility
+        if num_key_value_heads is None:
+            num_key_value_heads = num_attention_heads
+
+        self.mlp_bias = False
+        self.num_key_value_heads = num_key_value_heads
+        self.hidden_act = hidden_act
```
vasqu, line 166: A lot of this for example would be handled by modular, e.g. https://github.com/huggingface/transformers/blob/4a1ad8da8d6b36a60269ef1deadc474e9243aec1/src/transformers/models/solar_open/modular_solar_open.py#L34

src/transformers/models/youtu/configuration_youtu.py
```diff
@@ -0,0 +1,200 @@
...
+        if self.hidden_size != 0:
+            self.initializer_range = (
+                (2.0 / (5.0 * self.hidden_size)) ** 0.5 if initializer_range is None else initializer_range
+            )
+        else:
+            self.initializer_range = 0.02
+        # if embedding_initializer_range is None, set it to 2.0 * self.initializer_range
+        self.embedding_initializer_range = (
+            self.initializer_range * 2.0 if embedding_initializer_range is None else embedding_initializer_range
+        )
```
vasqu: `embedding_initializer_range` should just directly be defaulted in the kwargs

same for `initializer_range` but I can see at least a bit more special so it does make more sense - although, we don't need the if else: it would be weird if the hidden size were to be 0

src/transformers/models/youtu/configuration_youtu.py
```diff
@@ -0,0 +1,200 @@
...
+        # Standardize and validate the correctness of rotary position embeddings parameters
+        self.rope_parameters.setdefault("rope_theta", kwargs.pop("rope_theta", self.default_theta))
+        self.standardize_rope_params()
+        self.validate_rope(ignore_keys=ignore_keys_at_rope_validation)
+
+        # Convert to float because RoPE fn expect a float. Models on the hub were saved as int
+        for key in ["beta_fast", "beta_slow", "factor"]:
+            if key in self.rope_parameters:
+                self.rope_parameters[key] = float(self.rope_parameters[key])
+        return kwargs
```
vasqu: Shouldn't be needed and directly fixed on the hub instead! This overwrite was mostly meant for a few special models that need it for BC but we have the option to properly provide the correct `rope_parameters` from the get go

vasqu (commented):
Some initial comments from my side. I think we can simplify a few more things especially on the testing side with our causal lm tester. Looks overall solid tho :hugs:

src/transformers/models/youtu/configuration_youtu.py
vasqu: Ideally we should match the hub `model_type` with what we have integrated here, i.e. `youtu`. So this sounds like the best solution re updating later on - it is slightly breaking for users but unsure how many users really rely on the `model_type`

We could add onto `CONFIG_MAPPING_NAMES` but this feels like a dirty solution, wouldn't recommend tbh.

xenova:
I noticed that the [Youtu-LLM-2B](https://huggingface.co/tencent/Youtu-LLM-2B) model on the HF hub include duplicated embed_tokens + lm_head weights, even though according to the [config](https://huggingface.co/tencent/Youtu-LLM-2B/blob/main/config.json#L33), these should be tied.

```py
>>> from transformers import AutoModelForCausalLM
>>> model = AutoModelForCausalLM.from_pretrained("tencent/Youtu-LLM-2B")
Loading weights: 100%|███████████████| 387/387 [00:01<00:00, 283.68it/s, Materializing param=model.norm.weight]
The tied weights mapping and config for this model specifies to tie model.embed_tokens.weight to lm_head.weight, but both are present in the checkpoints, so we will NOT tie them. You should update the config with `tie_word_embeddings=False` to silence this warning
>>> model.num_parameters()
2224228352
>>> model.tie_weights()
>>> model.num_parameters()
1961560064
```

https://huggingface.co/tencent/Youtu-LLM-2B?show_file_info=model.safetensors

image

I've opened a PR for the config and weight updates in https://huggingface.co/tencent/Youtu-LLM-2B/discussions/17.

Saves ~500MB in the model weights. 💪 
image

src/transformers/models/youtu/modular_youtu.py
LuJunru: Yes, we design to use different std for embedding layers

src/transformers/models/youtu/modular_youtu.py
LuJunru: I guess not, since we hope to use different std for certain layers.

tests/models/youtu/test_modeling_youtu.py
LuJunru: Could be moved after pr merged? (https://huggingface.co/tencent/Youtu-LLM-2B/discussions/17)

src/transformers/models/youtu/configuration_youtu.py
LuJunru: Not quite similar, as we remove moe-related parameters

src/transformers/models/youtu/configuration_youtu.py
LuJunru: these stds would be decided according to the hidden size

```
self.initializer_range = (
    (2.0 / (5.0 * self.hidden_size)) ** 0.5 if initializer_range is None else initializer_range
)
self.embedding_initializer_range = (
    self.initializer_range * 2.0 if embedding_initializer_range is None else embedding_initializer_range
)
```

src/transformers/models/youtu/configuration_youtu.py
LuJunru: Indeed. Add this due to the suggestions from ```make style``` or ```make check-repo```.

tests/models/youtu/test_modeling_youtu.py
LuJunru: I found some attributes are incompatible:
```
FAILED tests/models/youtu/test_modeling_youtu.py::YoutuModelTest::test_model_rope_scaling_from_config_0_linear - AttributeError: 'CausalLMOutputWithPast' object has no attribute 'last_hidden_state'
FAILED tests/models/youtu/test_modeling_youtu.py::YoutuModelTest::test_model_rope_scaling_from_config_1_dynamic - AttributeError: 'CausalLMOutputWithPast' object has no attribute 'last_hidden_state'
FAILED tests/models/youtu/test_modeling_youtu.py::YoutuModelTest::test_model_rope_scaling_from_config_2_yarn - AttributeError: 'CausalLMOutputWithPast' object has no attribute 'last_hidden_state'
FAILED tests/models/youtu/test_modeling_youtu.py::YoutuModelTest::test_training_overfit - AttributeError: 'BaseModelOutputWithPast' object has no attribute 'loss'
```

docs/source/en/model_doc/youtu.md
LuJunru: fixed.

docs/source/en/model_doc/youtu.md
LuJunru: ok

src/transformers/models/youtu/modular_youtu.py
LuJunru: turn to qwen3

src/transformers/models/youtu/modular_youtu.py
LuJunru: Fixed to ```YoutuAttention```

tests/models/youtu/test_modeling_youtu.py
LuJunru: redundant codes removed.

utils/check_docstrings.py
LuJunru: Currently not able to be removed, as suggested by ```make check-repo```

src/transformers/models/youtu/modular_youtu.py
LuJunru: fixed

src/transformers/models/youtu/configuration_youtu.py
LuJunru: fixed

src/transformers/models/youtu/configuration_youtu.py
LuJunru: removed

src/transformers/models/youtu/configuration_youtu.py
vasqu: You could remove unrelated attributes after the super with `self.attr`

src/transformers/models/youtu/configuration_youtu.py
vasqu: We could at least avoid the else branch, but yes fair I overlooked that embedding was also dependent there

src/transformers/models/youtu/configuration_youtu.py
```diff
@@ -0,0 +1,181 @@
...
+        self.qk_rope_head_dim = qk_rope_head_dim
+        self.v_head_dim = v_head_dim
+        self.qk_nope_head_dim = qk_nope_head_dim
+        self.qk_head_dim = qk_nope_head_dim + qk_rope_head_dim
+        self.head_dim = qk_rope_head_dim
+        self.rope_interleave = rope_interleave
+
+        # for backward compatibility
+        if num_key_value_heads is None:
+            num_key_value_heads = num_attention_heads
```
vasqu, line 163: Shouldnt be needed, we should have this directly as arg if we already have the power to change this

src/transformers/models/youtu/modular_youtu.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(self, config: YoutuConfig, layer_idx: int):
+        nn.Module.__init__(self)
+        self.hidden_size = config.hidden_size
+
+        self.self_attn = YoutuAttention(config=config, layer_idx=layer_idx)
+
+        self.mlp = YoutuMLP(config)
+
+        self.input_layernorm = YoutuRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        self.post_attention_layernorm = YoutuRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
```
vasqu:
```suggestion
class YoutuDecoderLayer(LlamaDecoderLayer):
    pass
```
Should now be the same as Llama, no? The init should overlap now too

src/transformers/models/youtu/modular_youtu.py
vasqu: But you use `initializer_range` to init linear layers so we handle this already, see https://github.com/huggingface/transformers/blob/0189e042b8bd73573b95d7be7910f0ceab9c23f1/src/transformers/modeling_utils.py#L2286-L2287 get initializer range
https://github.com/huggingface/transformers/blob/0189e042b8bd73573b95d7be7910f0ceab9c23f1/src/transformers/modeling_utils.py#L2296-L2300 update linear layer

Hence, you can leave that part out and only update the embedding part

src/transformers/utils/auto_docstring.py
```diff
@@ -72,6 +72,7 @@
     "parakeet": "ParakeetCTCConfig",
     "lasr": "LasrCTCConfig",
     "wav2vec2-with-lm": "Wav2Vec2Config",
+    "youtu": "YoutuConfig",
```
vasqu: Why is this needed, would like to leave this out if possible

utils/check_docstrings.py
vasqu: Please run `make fix-repo` (without having this added to any exceptions) --> you will see it will update a few strings like I showed above, namely `initializer_range` and `embedding_initializer_range` - you might need to left tab them to have the same spacing again as before.

tests/models/youtu/test_modeling_youtu.py
vasqu: Yes gotcha

vasqu (commented):
Pushing some quick fixes so some comments are outdated but it will still need another look into the tests especially in regards to the causal lm tester (I have done something quickly but shapes are not working as expected for some tests so either needs to fix the init properly or overwrite some tests that check shapes)

I hope that lifts some confusions

src/transformers/models/youtu/modular_youtu.py
LuJunru: ok, got it

src/transformers/models/youtu/configuration_youtu.py
LuJunru: ok

src/transformers/models/youtu/configuration_youtu.py
LuJunru: somewhat simplified right now

src/transformers/models/youtu/configuration_youtu.py
LuJunru: Now the ```YoutuConfig``` is inherited from ```DeepseekV3Config```, so most codes removed.

src/transformers/models/youtu/modular_youtu.py
LuJunru: sure

src/transformers/utils/auto_docstring.py
LuJunru: removed

LuJunru:
Hi @vasqu 

I have fixed most of the issues mentioned above, please check again. 

There is one specific issue related to `check_docstring.py`. After using ```make fix-repo```, the original docstring will be changed from:
```
    Args:
            vocab_size (`int`, *optional*, defaults to 128256):
                Vocabulary size of the Deep model. Defines the number of different tokens that can be represented by the
                `inputs_ids` passed when calling [`YoutuModel`]
            hidden_size (`int`, *optional*, defaults to 2048):
                Dimension of the hidden representations.
            intermediate_size (`int`, *optional*, defaults to 6144):
                Dimension of the MLP representations.
            num_hidden_layers (`int`, *optional*, defaults to 32):
                Number of hidden layers in the Transformer decoder.
            num_attention_heads (`int`, *optional*, defaults to 16):
                Number of attention heads for each attention layer in the Transformer decoder.
            num_key_value_heads (`int`, *optional*, defaults to 16):
                In MLA, num_key_value_heads=num_attention_heads.
            kv_lora_rank (`int`, *optional*, defaults to 512):
                Rank of the LoRA matrices for key and value projections.
            q_lora_rank (`int`, *optional*, defaults to 1536):
                Rank of the LoRA matrices for query projections.
            qk_rope_head_dim (`int`, *optional*, defaults to 64):
                Dimension of the query/key heads that use rotary position embeddings.
            v_head_dim (`int`, *optional*, defaults to 128):
                Dimension of the value heads.
            qk_nope_head_dim (`int`, *optional*, defaults to 128):
                Dimension of the query/key heads that don't use rotary position embeddings.
            hidden_act (`str` or `function`, *optional*, defaults to `"silu"`):
                The non-linear activation function (function or string) in the decoder.
            max_position_embeddings (`int`, *optional*, defaults to 131072):
                The maximum sequence length that this model might ever be used with.
            initializer_range (`float`, *optional*):
                The standard deviation of the truncated_normal_initializer for initializing all weight matrices, except embedding matrices.
            embedding_initializer_range (`float`, *optional*):
                The standard deviation of the truncated_normal_initializer for initializing all embedding matrices.
            rms_norm_eps (`float`, *optional*, defaults to 1e-06):
                The epsilon used by the rms normalization layers.
            use_cache (`bool`, *optional*, defaults to `True`):
                Whether or not the model should return the last key/values attentions (not used by all models). Only
                relevant if `config.is_decoder=True`.
            pad_token_id (`int`, *optional*):
                Padding token id.
            bos_token_id (`int`, *optional*, defaults to 128000):
                Beginning of stream token id.
            eos_token_id (`int`, *optional*, defaults to 128001):
                End of stream token id.
            tie_word_embeddings (`bool`, *optional*, defaults to `True`):
                Whether to tie weight embeddings
            rope_parameters (`RopeParameters`, *optional*):
                Dictionary containing the configuration parameters for the RoPE embeddings. The dictionary should contain
                a value for `rope_theta` and optionally parameters used for scaling in case you want to use RoPE
                with longer `max_position_embeddings`.
            rope_interleave (`bool`, *optional*, defaults to `True`):
                Whether to interleave the rotary position embeddings.
            attention_bias (`bool`, defaults to `False`, *optional*, defaults to `False`):
                Whether to use a bias in the query, key, value and output projection layers during self-attention.
            attention_dropout (`float`, *optional*, defaults to 0.0):
                The dropout ratio for the attention probabilities.
```
to
```
    Args:
            vocab_size (`int | None`, *optional*, defaults to 128256): <fill_docstring>
            hidden_size (`int | None`, *optional*, defaults to 2048): <fill_docstring>
            intermediate_size (`int | None`, *optional*, defaults to 6144): <fill_docstring>
            num_hidden_layers (`int | None`, *optional*, defaults to 32): <fill_docstring>
            num_attention_heads (`int | None`, *optional*, defaults to 16): <fill_docstring>
            num_key_value_heads (`int | None`, *optional*, defaults to 16): <fill_docstring>
            kv_lora_rank (`int | None`, *optional*, defaults to 512): <fill_docstring>
            q_lora_rank (`int | None`, *optional*, defaults to 1536): <fill_docstring>
            qk_rope_head_dim (`int | None`, *optional*, defaults to 64): <fill_docstring>
            v_head_dim (`int | None`, *optional*, defaults to 128): <fill_docstring>
            qk_nope_head_dim (`int | None`, *optional*, defaults to 128): <fill_docstring>
            hidden_act (`str | None`, *optional*, defaults to `"silu"`): <fill_docstring>
            max_position_embeddings (`int | None`, *optional*, defaults to 131072): <fill_docstring>
            initializer_range (`float | None`, *optional*): <fill_docstring>
            embedding_initializer_range (`float | None`, *optional*): <fill_docstring>
            rms_norm_eps (`int | None`, *optional*, defaults to 1e-06): <fill_docstring>
            use_cache (`bool | None`, *optional*, defaults to `True`): <fill_docstring>
            pad_token_id (`int | None`, *optional*): <fill_docstring>
            bos_token_id (`int | None`, *optional*, defaults to 128000): <fill_docstring>
            eos_token_id (`int | None`, *optional*, defaults to 128001): <fill_docstring>
            tie_word_embeddings (`bool | None`, *optional*, defaults to `True`): <fill_docstring>
            rope_parameters (`transformers.modeling_rope_utils.RopeParameters | dict[str, transformers.modeling_rope_utils.RopeParameters]`, *optional*): <fill_docstring>
            rope_interleave (`bool | None`, *optional*, defaults to `True`): <fill_docstring>
            attention_bias (`bool | None`, *optional*, defaults to `False`): <fill_docstring>
            attention_dropout (`float | None`, *optional*, defaults to 0.0): <fill_docstring>
```

I noticed many ```<fill_docstring>``` placeholder here, is this correct? Meanwhile, even though I use the updated docstring by ```make fix-repo```, it still raise ```check_repository_consistency``` errors.

src/transformers/models/youtu/modular_youtu.py
```diff
@@ -0,0 +1,251 @@
...
+    LlamaRMSNorm,
+    LlamaRotaryEmbedding,
+)
+from ..qwen3.modeling_qwen3 import Qwen3MLP
+
+
+logger = logging.get_logger(__name__)
+
+
+class YoutuConfig(DeepseekV3Config):
```
vasqu, line 43: Sorry, that was confusing on my side --> I meant to say to add this to modular then. You can see that it will unfold the inherited attributes in the config file (which also solves the consistency issues) but better double check I haven't missed something

src/transformers/models/youtu/modular_youtu.py
```diff
@@ -0,0 +1,251 @@
...
+            self.initializer_range = 2.0 / (5.0 * self.hidden_size) ** 0.5
+
+        # if embedding_initializer_range is None, set it to 2.0 * self.initializer_range
+        if embedding_initializer_range is None:
+            self.embedding_initializer_range = 2.0 * self.initializer_range
+        else:
+            self.embedding_initializer_range = embedding_initializer_range
+
+    def convert_rope_params_to_dict(self, ignore_keys_at_rope_validation: set | None = None, **kwargs):
+        raise AttributeError("Not overwritten for the Youtu model!")
```
vasqu, line 206: Fyi, this way you can disable inheriting function from others

tests/models/youtu/test_modeling_youtu.py
```diff
@@ -0,0 +1,192 @@
...
+            batch_size,
+            getattr(config, "num_key_value_heads", config.num_attention_heads),
+            seq_length,
+        )
+        expected_key_shape = expected_common_shape + (config.qk_nope_head_dim + config.qk_rope_head_dim,)
+        expected_value_shape = expected_common_shape + (config.v_head_dim,)
+
+        for layer in past_key_values.layers:
+            self.assertEqual(layer.keys.shape, expected_key_shape)
+            self.assertEqual(layer.values.shape, expected_value_shape)
```
vasqu, line 85: Refactored this to use our causal lm class - it makes it easier for use to refactor tests in the future

tests/models/youtu/test_modeling_youtu.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+@slow
+class YoutuIntegrationTest(unittest.TestCase):
+    def tearDown(self):
+        cleanup(torch_device, gc_collect=False)
```
vasqu, line 95: Let's also clean on setup e.g. https://github.com/huggingface/transformers/blob/a1f63d55b9e3fc38e70a5d10df01b7a94f3cbd91/tests/models/llama/test_modeling_llama.py#L65-L73
(no need to copy the comment)

vasqu (approved):
I fixed a few things in regards to the config and tests, that resolves your issue as well it seems. Lmk if not!

Just a few last nits but approving since it's nothing major, quick checking with our slow CI in a second (might need to adjust values because of GPU differences)

vasqu:
It does seem we need to fix at least: `FAILED tests/models/youtu/test_modeling_youtu.py::YoutuModelTest::test_config - ZeroDivisionError: float division by zero`

vasqu:
run-slow: youtu

src/transformers/models/youtu/modular_youtu.py
LuJunru: ok, got it

LuJunru:
> It does seem we need to fix at least: `FAILED tests/models/youtu/test_modeling_youtu.py::YoutuModelTest::test_config - ZeroDivisionError: float division by zero`

alright, i guess this was the error that led me to add conditions of ```hidden_size = 0``` in the ```configuration_youtu.py``` previously.

vasqu:
`test_sdpa_can_dispatch_on_flash` seems to fail because of some hidden dim incompatibility - you can either skip with the proper reason or try to adjust the values in the tester

vasqu:
run-slow: youtu

vasqu:
@LuJunru I just updated some last few nits, everything passes (the other tests are unrelated) so I merged 

Thanks for iterating and gz on the model addition :hugs:

LuJunru:
> @LuJunru I just updated some last few nits, everything passes (the other tests are unrelated) so I merged
> 
> Thanks for iterating and gz on the model addition 🤗

@molbap @vasqu @xenova 🤗 Thank you for the professional suggestions! I'm going to update the usage in our official repos.

tests/models/youtu/test_modeling_youtu.py
```diff
@@ -0,0 +1,196 @@
...
+    torch_device,
+)
+
+from ...causal_lm_tester import CausalLMModelTest, CausalLMModelTester
+
+
+if is_torch_available():
+    import torch
+
+    torch.set_float32_matmul_precision("high")
```
ydshieh, line 36: Hi @LuJunru Is any particular reason to set it to "high" here? It causes the issue described in 

https://github.com/huggingface/transformers/pull/45248#issuecomment-4188677910

LuJunru: To avoid test failures caused by minor numerical fluctuations across different testing environments when using different GPUs.

ydshieh: Thanks. 

I will change it to `highest` which should work too. If you are curious why we need to change it, see #45252

PR #43168, opened 2026-01-08
[cache] Remove all deprecated classes

Cyrilvallez:
# What does this PR do?


ArthurZucker (approved):
let's keep an alias or no?

stevhliu:
i think we'll also need to update the docs [here](https://huggingface.co/docs/transformers/main/en/optimization_overview#caching) and [here](https://huggingface.co/docs/transformers/main/en/kv_cache) which mention the `OffloadedCache`?

Cyrilvallez:
I think we really want to move away from the many many cache classes we had before, so let's not keep any alias!

Cyrilvallez:
@stevhliu, we only remove the `OffloadedCache` class, but `cache_implementation="offloaded"` or `cache_implementation="offloaded_static"` are still valid! It's just the classes that are removed, but the logic for those stay the same!

PR #43171, opened 2026-01-08
[makefile] Cleanup and improve the rules

Cyrilvallez:
# What does this PR do?

As per the title. Along the years, the rules accumulated and are now weirdly named/hard to remember.
This PR cleans everything, and adds the following 3 important rules:

- `style` -> runs only linting/formatting automatic checks
- `check-repo` -> performs all repo health checks, without fixing even if a fix is available
- `fix-repo` -> performs all repo automatic fixes, e.g. modular conversions, Copied from, dummy etc


Makefile
```diff
@@ -1,46 +1,28 @@
...
 
 # Check that the repo is in a good state
-
-repo-consistency:
+check-repo:
+	ruff check $(check_dirs) setup.py conftest.py
+	ruff format --check $(check_dirs) setup.py conftest.py
+	@python -c "from transformers import *" || (echo '🚨 import failed, this means you introduced unprotected imports! 🚨'; exit 1)
+	python utils/sort_auto_mappings.py --check_only
+	python utils/check_doc_toc.py
```
ArthurZucker:
```suggestion
	python utils/check_doc_toc.py ||
```

ArthurZucker (approved):
i would run even if previous fails to have loads of failures as they can guide the user

Makefile
```diff
@@ -1,46 +1,28 @@
...
-	else \
-		modified_py_files=$$(python utils/get_modified_files.py $(check_dirs)); \
-		if [ -n "$$modified_py_files" ]; then \
-			echo "Checking/fixing files: $${modified_py_files}"; \
-			ruff check $${modified_py_files} --fix --exclude $(exclude_folders); \
-			ruff format $${modified_py_files} --exclude $(exclude_folders); \
-		else \
-			echo "No library .py files were modified"; \
-		fi; \
-	fi
```
Cyrilvallez, line 24: This is really not needed, ruff runs on ALL the repo in a really negligible time (less than 0.1s)

Makefile
Cyrilvallez: Nice idea!

PR #43172, opened 2026-01-08
Check kernel FA when creating mask

zucchini-nlp:
# What does this PR do?

If we request kernels FA2 when the model is loaded, the attention mask is created as 4D which causes issues downstream (e.g in Siglip). We can either manually force `config.attn_implementation` to be only `flash_attention_2/3` by mapping it back from kernels, or we can check both until new masking API is used everywhere

IMO we need to keep `config.attn_implementation` pointing to correct impl, so I went to adjust the name in all other files

Btw, some models don't even use the general attention interface yet 🫠 I didn't check model-by-model and made changes in batch automatically.

### DIFF
--- a/src/transformers/masking_utils.py
+++ b/src/transformers/masking_utils.py
@@ -632,6 +632,8 @@ class AttentionMaskInterface(GeneralInterface):
         "eager": eager_mask,
         "flash_attention_2": flash_attention_mask,
         "flash_attention_3": flash_attention_mask,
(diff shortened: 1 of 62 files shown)


vasqu:
Does #43121 solve these issues, had the same thought as you and it annoyed me so I went into it :joy:

zucchini-nlp:
oh cool, did not see that 😄

PR #43175, opened 2026-01-08
Fix generation config validation

zucchini-nlp:
# What does this PR do?

Fixes an issue reported by @BenjaminBossan . The default of caching was `True` so we assume that new default of `None` is also `True`. So we need to check only if `False` is set explicitly by the user

```python

model_id = 'hf-internal-testing/tiny-random-Gemma3ForCausalLM'
tmp_dirname = "/tmp/save-gemma3"
model.save_pretrained(tmp_dirname)

```

### DIFF
--- a/src/transformers/generation/configuration_utils.py
+++ b/src/transformers/generation/configuration_utils.py
@@ -687,7 +687,7 @@ class GenerationConfig(PushToHubMixin):
                 )
 
         # 2.5. check cache-related arguments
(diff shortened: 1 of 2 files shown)


vasqu (approved):
Let's add a quick test for this otherwise lgtm

zucchini-nlp:
Oke, extending the prev test on "generation config default values"

BenjaminBossan:
Thanks.

PR #43177, opened 2026-01-08
make devices consistent in cache reordering

pandey-ps:
# What does this PR do?
This PR fixes cache reordering causing runtime error when keys and values are on different devices

While working with beam search, with the cache operations, i got this error attached

Screenshot 2026-01-08 210423

this was due to beam_idx was being moved to different devices without considering that keys and values can be on different devices when during memory offloading.

changes i made are just to get the device info from keys and values separately solving the mismatch, hope this helps!

      Pull Request section?
      to it if that's the case.


pandey-ps:
@vasqu kindly review when you get a moment, thanks!

src/transformers/cache_utils.py
```diff
@@ -77,8 +77,12 @@ def reset(self) -> None:
     def reorder_cache(self, beam_idx: torch.LongTensor) -> None:
         """Reorders this layer's cache for beam search."""
         if self.get_seq_length() > 0:
-            self.keys = self.keys.index_select(0, beam_idx.to(self.keys.device))
-            self.values = self.values.index_select(0, beam_idx.to(self.values.device))
+            # Get device for each tensor and then move beam_idx to that device
+            keys_device = self.keys.device
+            self.keys = self.keys.index_select(0, beam_idx.to(keys_device))
```
vasqu, line 82: Isn't that exactly the same as what's already happening
```python
self.keys = self.keys.index_select(0, beam_idx.to(self.keys.device))
```
vs
```python
keys_device = self.keys.device
self.keys = self.keys.index_select(0, beam_idx.to(keys_device))
```

Do you have a reproducer to check what's going on?

PR #43181, opened 2026-01-08
[`Generate`] Allow custom config values in generate config

vasqu:
As per title, followup to a series of PRs that address the new config logic
- #43000
- #42958
- #42702

tl;dr: custom values in an uploaded generation config are ignored atm


src/transformers/generation/utils.py
```diff
@@ -1797,6 +1797,11 @@ def _prepare_generation_config(
         generation_config.update(**self.generation_config.to_dict(), defaults_only=True)
         generation_config.update(**global_defaults, defaults_only=True)
 
+        # Add custom keys not in global defaults, e.g. as in `VibeVoice` for `noise_scheduler_class`
+        for key, value in self.generation_config.to_dict().items():
+            if not hasattr(generation_config, key):
+                setattr(generation_config, key, value)
+
```
ebezzam: thanks @vasqu 👍 

This is indeed what works for me with VibeVoice (see [here](https://github.com/pengzhiliang/transformers/blob/1204d12f7a04370b17212103d2713d4e266e33c7/src/transformers/generation/utils.py#L1800) from VibeVoice PR)

vasqu: Updated it to be within the update fn directly instead

src/transformers/generation/utils.py
```diff
@@ -1791,10 +1791,13 @@ def _prepare_generation_config(
...
-        # Do not update any values that aren't `None`, i.e. if set by users explicitly and passed
-        # to `generate()`. Thus the `defaults_only=True` is used
+        #
+        # Only update values that are `None`, i.e. these values were not explicitly set by by users to `generate()`,
+        # or values that are not present in the current config, i.e. custom entries that were set via `**kwargs`.
+        # Thus we use the specific kwargs `defaults_only=True` (`None` values only) and `allow_custom_entries=True`
+        # (custom entries are carried over).
         global_defaults = self.generation_config._get_default_generation_params()
-        generation_config.update(**self.generation_config.to_dict(), defaults_only=True)
+        generation_config.update(**self.generation_config.to_dict(), defaults_only=True, allow_custom_entries=True)
```
zucchini-nlp, line 1800: Ah, that is what you meant earlier. I assumed that `generation_config` at this point will already be `self.generation_config`

Btw, if a few lines above we create not  `generation_config = GenerationConfig()` but `generation_config = self.generation_config` it will be much easier. I can't think of why it is a bad idea. WDYT?

zucchini-nlp (commented):
Also, let's allow loading custom fields when running `generation_config.from_pretrained`. After the recent changes, there is no need for a `_from_model_config` private attr and we are guaranteed to get only generation attr at init time. So we could just save all kwargs at `__init__` as attributes, criiw

src/transformers/generation/utils.py
vasqu: Nope, not the case, e.g.
```python
from transformers import AutoModelForCausalLM, AutoTokenizer, GenerationConfig

    "meta-llama/Llama-3.2-3B-Instruct",
    device_map="auto",
    torch_dtype="auto",
    attn_implementation="flash_attention_2",
).eval()
tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-3.2-3B-Instruct")
tokenizer.pad_token = tokenizer.eos_token
inputs = tokenizer(
    ["Hello, how are you?", "is this life?"],
    padding=True,
    padding_side="left",
    return_tensors="pt",
).to(model.device)

outputs = model.generate(**inputs, max_new_tokens=50, generation_config=GenerationConfig(do_sample=False))
print(tokenizer.batch_decode(outputs, skip_special_tokens=True))
```
We will have different generation configs so it's not guaranteed.

zucchini-nlp: Indeed 🥲  
I guess there is no cleaner option then

src/transformers/generation/utils.py
```diff
@@ -1791,10 +1791,13 @@ def _prepare_generation_config(
         generation_config = copy.deepcopy(generation_config)
 
         # First set values from the loaded `self.generation_config`, then set default values (BC)
-        # Do not update any values that aren't `None`, i.e. if set by users explicitly and passed
-        # to `generate()`. Thus the `defaults_only=True` is used
+        #
+        # Only update values that are `None`, i.e. these values were not explicitly set by by users to `generate()`,
```
ebezzam: nit typo: 2x "by"

ebezzam (commented):
thanks @vasqu and @zucchini-nlp for iterating! I tried these changes in my VibeVoice branch, and they work 👍

src/transformers/generation/utils.py
vasqu: Oops, thanks for spotting :grin:

zucchini-nlp (approved):
Seems like the CI is flaky again 🥲

vasqu:
Finally got it

PR #43185, opened 2026-01-09
[docs] llama.cpp

stevhliu:
- adds llama.cpp guide explaining how it uses transformers to the new community integrations section
- create separate section for deployment (more to follow soon here!)


ngxson (approved):
nice, thanks!!

PR #43187, opened 2026-01-09
🚨Fix floating-point precision in JanusImageProcessor resize

raimbekovm:
Fixes #43186

## What does this PR do?

Fixes floating-point precision issue in `JanusImageProcessor.resize()` where `int()` truncation caused incorrect output dimensions.

**Problem:**
```python
height = 2522, size = 384
delta = 384 / 2522 = 0.15225218081681999...
height * delta = 383.9999999999...
int(height * delta) = 383  # Wrong!
round(height * delta) = 384  # Correct!
```

**Solution:**
Replace `int()` with `round()` for proper rounding of scaled dimensions.

## Why edit `modular_janus.py`?

The file `image_processing_deepseek_vl.py` (mentioned in the issue) is auto-generated from `modular_janus.py`. The header states:
```
🚨 This file was automatically generated from src/transformers/models/janus/modular_janus.py.
   Do NOT edit this file manually...
```

DeepseekVL imports `JanusImageProcessor`, so fixing the source fixes all dependent models:
- Janus
- DeepseekVL
- DeepseekVL Hybrid


Rocketknight1:
Bug seems legit but I'd like someone with more image processor experience to check that we're still matching the behaviour that these models expect, because sometimes models are trained with slightly buggy preprocessing code and it becomes important to match that behaviour later. cc @molbap @yonigozlan @zucchini-nlp

zucchini-nlp (approved):
Makes sense to me, we need the sides to be exactly 'size' length. Can you run `make fix-repo` to convert modular files?

raimbekovm:
Done. Ran the modular converter and pushed the regenerated file.

Note: The same int() truncation issue exists in two other places:
- `image_processing_deepseek_vl.py` (generated from modular_deepseek_vl.py which inherits from JanusImageProcessor)
- `image_processing_janus_fast.py` (not a modular file)

Should I include these in this PR or handle them separately?

zucchini-nlp:
> Should I include these in this PR or handle them separately?

Running `make fix-repo` should have auto-fixed those, no? I see the CI is still complaining about these models

raimbekovm:
Fixed. Added the missing changes for deepseek_vl and janus_fast.

zucchini-nlp:
@bot /repo

yonigozlan:
run-slow: deepseek_vl, janus, deepseek_vl_hybrid

zucchini-nlp:
run-slow: deepseek_vl, janus, deepseek_vl_hybrid

zucchini-nlp:
slow tests cannot be triggered 😢

molbap:
run-slow: deepseek_vl, deepseek_vl_hybrid, janus

yonigozlan:
I got similar results locally with slow tests now failing due to this changes. However it seems to me that the processing now works as it's supposed to , so ok to change the expected values (and add a 🚨 as it breaks BC a bit)? @molbap @zucchini-nlp

zucchini-nlp:
Yeah, sounds good to me. Let's just update test values with the new ones and merge as breaking

yonigozlan:
run-slow: deepseek_vl, deepseek_vl_hybrid, janus

PR #43194, opened 2026-01-09
Fix MusicGen generation config

zucchini-nlp:
# What does this PR do?

Some audio models have their own `generate()` which doesn't call `prepare_generation_config`. We need to call `prepare_generation_config` to update values to prev defaults

Also, I added a small deprecation warning preparing us and users to passing either generation config of `kwargs`, only one. We will anyway have to support it for a long time, it is a common pattern among users


src/transformers/models/musicgen_melody/modeling_musicgen_melody.py
```diff
@@ -1102,12 +1101,8 @@ def generate(
                     - [`~generation.GenerateBeamEncoderDecoderOutput`]
         """
         # 1. Handle `generation_config` and kwargs that might update it, and validate the resulting objects
-        if generation_config is None:
-            generation_config = self.generation_config
-
-        generation_config = copy.deepcopy(generation_config)
-        model_kwargs = generation_config.update(**kwargs)  # All unused kwargs must be model kwargs
-        generation_config.validate()
+        generation_config = self.generation_config if generation_config is None else generation_config
```
zucchini-nlp: this line is used in all overriden `generate()` to use custom gen-params. Need to clean up after your PR on custom generation parameters

zucchini-nlp:
run-slow: musicgen_melody

src/transformers/generation/utils.py
```diff
@@ -1807,6 +1807,16 @@ def _prepare_generation_config(
...
 
+        # It doesn't make sense to allow kwargs and `generation_config`, that should be mutually exclusive
+        if generation_config_provided is not None and set(kwargs.keys()) - set(model_kwargs.keys()):
+            generation_kwargs = set(kwargs.keys()) - set(model_kwargs.keys())
+            logger.warning_once(
+                f"Passing `generation_config` together with generation-related "
+                f"arguments=({generation_kwargs}) is deprecated and will be removed in future versions. "
+                "Please pass either a `generation_config` object OR all generation "
+                "parameters explicitly, but not both.",
+            )
```
vasqu, line 1821: I would love to have this raise an error but it's probably too breaking?

src/transformers/models/musicgen_melody/modeling_musicgen_melody.py
vasqu: Do we want to prio #43181 first then, I changed the logic to be within update directly (via a new kwarg), lmk what you think otherwise fine to fix this first

vasqu (approved):
Fine to fix this first before we forget, `test_generate_text_audio_prompt` still fails not sure if it failed before as well

vasqu:
CI is going crazy with the flakiness :cry:

src/transformers/generation/utils.py
zucchini-nlp: yeah, we don't want to break it yet. This is to nudge users for better practices and give them time to get used to it. Maybe we can raise errors after 10-15 releases

src/transformers/models/musicgen_melody/modeling_musicgen_melody.py
zucchini-nlp: yeah, I don't mind waiting for your PR to be merged first

zucchini-nlp:
CI is sooo flaky

PR #43195, opened 2026-01-09
Remove more deprecated objects/args

Cyrilvallez:
# What does this PR do?

As per the title!
Most important change is the removal of the 2 deprecated auto classes


ArthurZucker (approved):

djstrong:
@Cyrilvallez @ArthurZucker Why e.g. `return_all_scores` support in text classification was removed?

Cyrilvallez:
Hey @djstrong! See the following snippet that was removed:
```python
elif return_all_scores is not None:
    warnings.warn(
        "`return_all_scores` is now deprecated,  if want a similar functionality use `top_k=None` instead of"
        " `return_all_scores=True` or `top_k=1` instead of `return_all_scores=False`.",
        UserWarning,
    )
    if return_all_scores:
        postprocess_params["top_k"] = None
    else:
        postprocess_params["top_k"] = 1
```

PR #43201, opened 2026-01-09
[bugfix] fix qwen3vlmoe loading

JJJYmmm:
# What does this PR do?

Fix the loading of qwen3vlmoe experts.

test script:

```
from transformers import AutoProcessor, AutoModelForImageTextToText

model = AutoModelForImageTextToText.from_pretrained("Qwen/Qwen3-VL-30B-A3B-Instruct", torch_dtype="auto", device_map="auto")
```

before:
```
model.language_model.layers.{0...47}.mlp.experts.down_proj    | MISMATCH | Reinit due to size mismatch ckpt: torch.Size([128, 768, 2048]) vs model:torch.Size([128, 2048, 768])  
model.language_model.layers.{0...47}.mlp.experts.gate_up_proj | MISMATCH | Reinit due to size mismatch ckpt: torch.Size([128, 2048, 1536]) vs model:torch.Size([128, 1536, 2048])
```

The reason is that the official checkpoint of Qwen3VLMoe follows the shape `[num_experts, out_features, in_features]`, but in the latest code, these weights are transposed in the last two dimensions. This pr transposes them back during conversion.

The model now loads successfully after the fix. 🫡

### DIFF
--- a/src/transformers/conversion_mapping.py
+++ b/src/transformers/conversion_mapping.py
@@ -234,6 +234,18 @@ def _build_checkpoint_conversion_mapping():
     mapping["qwen3_omni_moe"] = mapping["qwen2_moe"].copy()
     mapping["qwen3_next"] = mapping["qwen2_moe"].copy()
     mapping["qwen3_vl_moe"] = mapping["qwen2_moe"].copy()
+    mapping["qwen3_vl_moe"] += [
+        WeightConverter(
+            source_patterns="mlp.experts.gate_up_proj",
+            target_patterns="mlp.experts.gate_up_proj",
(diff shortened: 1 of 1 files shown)


vasqu:
Similar issue as in #43227 where reverse mapping fails cc @Cyrilvallez

IlyasMoutawwakil:
@vasqu a quik fix that i found for now is to rename the tensor first and then transpose it.

JJJYmmm:
I'll wait for #43227 to be merged and then apply the final fix for the transposed experts. 🫡

Cyrilvallez:
Closing as superseded by #43307!

PR #28538, opened 2024-01-16
[`gradient_checkpointing`] default to use it for torch 2.3

ArthurZucker:
# What does this PR do?

Fixes #28536 in preparation for next torch release

Issue #28536, opened 2024-01-16
Gradient checkpointing throws use_reentrant warning on PyTorch 2.1


rosario-purple:
### System Info

- `transformers` version: 4.36.2
- Platform: Linux-5.15.0-91-generic-x86_64-with-glibc2.35
- Python version: 3.10.13
- Huggingface_hub version: 0.19.4
- Safetensors version: 0.4.0
- Accelerate version: 0.25.0
- Accelerate config:    - compute_environment: LOCAL_MACHINE
        - distributed_type: DEEPSPEED
        - mixed_precision: bf16
        - use_cpu: False
        - debug: False
        - num_processes: 8
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {'gradient_accumulation_steps': 1, 'offload_optimizer_device': 'none', 'offload_param_device': 'none', 'zero3_init_flag': True, 'zero3_save_16bit_model': False, 'zero_stage': 3}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
- PyTorch version (GPU?): 2.1.1+cu121 (True)
- Tensorflow version (GPU?): not installed (NA)
- Flax version (CPU?/GPU?/TPU?): 0.7.5 (cpu)
- Jax version: 0.4.21
- JaxLib version: 0.4.21
- Using GPU in script?: Yes
- Using distributed or parallel set-up in script?: Yes

### Who can help?

@ArthurZucker @younesbelkada 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

Training any text model with gradient checkpointing enabled on PyTorch 2.1 and higher produces this warning:

```
/scratch/miniconda3/envs/brr/lib/python3.10/site-packages/torch/utils/checkpoint.py:429: Warning: torch.utils.checkpoint: please pass in use_reentrant=True or use_reentrant=False explicitly. The default value of use_reentrant will be updated to be False in the future. To maintain current behavior, pass use_reentrant=True. It is recommended that you use use_reentrant=False. Refer to docs for more details on the differences between the two variants.
```

This can be resolved by manually monkey-patching the model code with `use_reentrant=True`, eg. like so:

```
[...]

PR #43203, opened 2026-01-09
Switch gradient checkpointing default to `use_reentrant=False` (PyTorch recommended)

qgallouedec:
### Summary

This PR changes our gradient checkpointing default from `use_reentrant=True` to `use_reentrant=False`.

Two years ago we explicitly set `use_reentrant=True` in #28538 because PyTorch started warning that the default would change in the future, and recommending users choose a value explicitly:

> ```
> /scratch/miniconda3/envs/brr/lib/python3.10/site-packages/torch/utils/checkpoint.py:429: Warning: 
> torch.utils.checkpoint: please pass in use_reentrant=True or use_reentrant=False explicitly. The default 
> value of use_reentrant will be updated to be False in the future. To maintain current behavior, pass 
> use_reentrant=True. It is recommended that you use use_reentrant=False. Refer to docs for more details 
> on the differences between the two variants.
> ```
> PyTorch warning shown with torch 2.3, see #28536

At the time, defaulting to `True` was the safest choice to preserve the behavior of earlier releases.

PyTorch now **recommends** the non-reentrant variant (`use_reentrant=False`) see, https://docs.pytorch.org/docs/stable/checkpoint.html, and is moving toward making it the default. Aligning with this upstream recommendation gives us several benefits:

Note: training and checkpointing behavior remains functionally equivalent in typical use cases, with the main difference being **how activations are recomputed during backward** (non-reentrant uses a safer mechanism).

### Don't skip tests that pass!

Some gradient checkpointing tests were skipped, but they actually pass. This PR removes these unwanted skipping.

```diff
- @unittest.skip
- def test_training_gradient_checkpointing(self):
-     pass
```

67 new passing tests!

### Test renaming

Now that `True` isn't the default anymore, we rename

```diff
- def test_training_gradient_checkpointing_use_reentrant(self):
+ def test_training_gradient_checkpointing_use_reentrant_false(self):
```

### Fix the `reason` of skipping

Some tests are failing, but the reported reason was not correct, this PR fixes all the occurrences:

```diff
- @unittest.skip(reason="This architecture seem to not compute gradients properly when using GC, check: https://github.com/huggingface/transformers/pull/27124")
+ @unittest.skip(reason="This module does not support standalone training")
  def test_training_gradient_checkpointing_use_reentrant_false(self):
      pass
```

It's the case for example for
- `tests/models/align/test_modeling_align.py`
- `tests/models/align/test_modeling_altclip/.py`
- etc

### `xfail` instead of `skip` for model that don't properly compute the gradient

`xfail` suit better in these cases

```diff
- @unittest.skip(reason="This architecture seem to not compute gradients properly when using GC, check https://github.com/huggingface/transformers/pull/27124")
+ @pytest.mark.xfail(reason="This architecture seems to not compute gradients for some layer.")
  def test_training_gradient_checkpointing_use_reentrant_false(self):
-     pass
+     super().test_training_gradient_checkpointing_use_reentrant_false()
```

it's the case for example for 

- `tests/models/autoformer/test_modeling_autoformer.py`
- `tests/models/aya_vision/test_modeling_aya_vision.py`
- `tests/models/big_bird/test_modeling_big_bird.py`
- etc


tests/models/align/test_modeling_align.py
```diff
@@ -209,22 +209,6 @@ def check_hidden_states_output(inputs_dict, config, model_class):
...
-    )
-    def test_training_gradient_checkpointing_use_reentrant(self):
-        pass
-
-    @unittest.skip(
-        reason="This architecture seem to not compute gradients properly when using GC, check: https://github.com/huggingface/transformers/pull/27124"
-    )
-    def test_training_gradient_checkpointing_use_reentrant_false(self):
-        pass
-
```
qgallouedec, line 227: It seems a big number of this ignored test actually pass. I check them all

src/transformers/models/recurrent_gemma/modeling_recurrent_gemma.py
```diff
@@ -489,6 +489,8 @@ def forward(
                 x_branch = x_branch.unsqueeze(-1)
                 self.conv1d_state = conv_state[:, :, 1:]
         else:
+            self.conv1d_state = None
+            self.rg_lru.recurrent_states = None
```
qgallouedec, line 491: Clearing recurrent cache state when `use_cache=False` so the recurrent block doesn’t reuse graph-attached state between forwards, which fixes the double-backward error (see below) in the gradient checkpointing tests.

Result before the change:

```
$ pytest tests/models/recurrent_gemma/test_modeling_recurrent_gemma.py::RecurrentGemmaModelTest::test_training_gradient_checkpointing
================================================ test session starts ================================================
platform linux -- Python 3.12.12, pytest-8.4.2, pluggy-1.6.0
rootdir: /fsx/qgallouedec/transformers
configfile: pyproject.toml
plugins: timeout-2.4.0, asyncio-1.3.0, rich-0.2.0, anyio-4.12.0, cov-7.0.0, xdist-3.8.0, hypothesis-6.150.0, order-1.3.0, rerunfailures-15.1
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
collected 1 item                                                                                                    

tests/models/recurrent_gemma/test_modeling_recurrent_gemma.py::RecurrentGemmaModelTest::test_training_gradient_checkpointing FAILED [100%]

===================================================== FAILURES ======================================================
___________________________ RecurrentGemmaModelTest.test_training_gradient_checkpointing ____________________________

self = <tests.models.recurrent_gemma.test_modeling_recurrent_gemma.RecurrentGemmaModelTest testMethod=test_training_gradient_checkpointing>

    def test_training_gradient_checkpointing(self):
        # Scenario - 1 default behaviour
>       self.check_training_gradient_checkpointing()

tests/test_modeling_common.py:1620: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_modeling_common.py:1572: in check_training_gradient_checkpointing
    loss.backward()
../miniconda3/envs/trl/lib/python3.12/site-packages/torch/_tensor.py:625: in backward
    torch.autograd.backward(
../miniconda3/envs/trl/lib/python3.12/site-packages/torch/autograd/__init__.py:354: in backward
    _engine_run_backward(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

t_outputs = (tensor(4.6717, device='cuda:0', grad_fn=<NllLossBackward0>),)
args = ((tensor(1., device='cuda:0'),), False, False, ())
kwargs = {'accumulate_grad': True, 'allow_unreachable': True}, attach_logging_hooks = False

    def _engine_run_backward(
        t_outputs: Sequence[Union[torch.Tensor, GradientEdge]],
        *args: Any,
        **kwargs: Any,
    ) -> tuple[torch.Tensor, ...]:
        attach_logging_hooks = log.getEffectiveLevel() <= logging.DEBUG
        if attach_logging_hooks:
            unregister_hooks = _register_logging_hooks_on_whole_graph(t_outputs)
        try:
>           return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
                t_outputs, *args, **kwargs
            )  # Calls into the C++ engine to run the backward pass
E           RuntimeError: Trying to backward through the graph a second time (or directly access saved tensors after they have already been freed). Saved intermediate values of the graph are freed when you call .backward() or autograd.grad(). Specify retain_graph=True if you need to backward through the graph a second time or if you need to access saved tensors after calling backward.

../miniconda3/envs/trl/lib/python3.12/site-packages/torch/autograd/graph.py:841: RuntimeError
============================================== short test summary info ==============================================
FAILED tests/models/recurrent_gemma/test_modeling_recurrent_gemma.py::RecurrentGemmaModelTest::test_training_gradient_checkpointing - RuntimeError: Trying to backward through the graph a second time (or directly access saved tensors after they ha...
================================================ 1 failed in 10.78s =================================================
```

src/transformers/modeling_utils.py
```diff
@@ -3079,7 +3079,7 @@ def gradient_checkpointing_enable(self, gradient_checkpointing_kwargs=None):
             raise ValueError(f"{self.__class__.__name__} does not support gradient checkpointing.")
 
         if gradient_checkpointing_kwargs is None:
-            gradient_checkpointing_kwargs = {"use_reentrant": True}
+            gradient_checkpointing_kwargs = {"use_reentrant": False}
```
qgallouedec, line 3066: This is the main change

SunMarc (approved):
Thanks for changing the default and going through the existing tests ! cc @molbap for viz as you worked on gradient checkpointing !

PR #43207, opened 2026-01-10
Fix Mamba2ForCausalLM weight tying

Anri-Lombard:
## What does this PR do?

Fixes #43206

Adds the `_tied_weights_keys` mapping to `Mamba2ForCausalLM` to enable proper weight tying when `tie_word_embeddings=True`.

## The Bug

When `tie_word_embeddings=True`, the embedding weights should be shared with the `lm_head`. However, `Mamba2ForCausalLM` had:
```python
_tied_weights_keys = {}  # Empty - weight tying never happens
```

This caused:
- Weights not actually tied (different tensors)
- `resize_token_embeddings()` not resizing `lm_head` properly

## The Fix

```python
_tied_weights_keys = {"lm_head.weight": "backbone.embeddings.weight"}
```

This is the standard pattern used by `MambaForCausalLM` (v1), GPT2, LLaMA, and other models.

## Verification

```python
from transformers import Mamba2ForCausalLM, Mamba2Config

config = Mamba2Config(vocab_size=1000, tie_word_embeddings=True)
model = Mamba2ForCausalLM(config)

# Weights are now properly tied
assert model.lm_head.weight.data_ptr() == model.backbone.embeddings.weight.data_ptr()

# Resize works correctly
model.resize_token_embeddings(1100)
assert model.lm_head.weight.shape[0] == 1100
```

### DIFF
--- a/tests/models/mamba2/test_modeling_mamba2.py
+++ b/tests/models/mamba2/test_modeling_mamba2.py
@@ -90,7 +90,7 @@ class Mamba2ModelTester:
         num_labels=3,
         num_choices=4,
         scope=None,
-        tie_word_embeddings=True,
+        tie_word_embeddings=False,
     ):
         self.parent = parent
         self.num_heads = num_heads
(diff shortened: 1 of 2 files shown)


vasqu (approved):
Can you add a fast test as regression test

Anri-Lombard:
Thanks! Enabled `tie_word_embeddings=True` in the ModelTester to run the standard weight tying regression test. Are you happy with that? 🙏

vasqu:
Can we instead create a small test for this? It seems that the original model did not have tied weights so this is more of an addition IMO and we check this as an explicit test (would be nice to link / mention your issue as well)

Anri-Lombard:
@vasqu Added an explicit regression test that checks both tie_word_embeddings=True and tie_word_embeddings=False 🙏

tests/models/mamba2/test_modeling_mamba2.py
```diff
@@ -341,6 +341,26 @@ def recursive_check(tuple_object, dict_object):
             dict_inputs = self._prepare_for_class(inputs_dict, model_class, return_labels=True)
             check_equivalence(model, tuple_inputs, dict_inputs, {"output_hidden_states": True})
 
+    def test_tied_weight_embeddings(self):
+        """Regression test for https://github.com/huggingface/transformers/issues/43206."""
```
vasqu, line 345: Awesome, thanks for linking :pray:

vasqu (approved):
Thank you, just checking with run-slow once just to be paranoid safe then merging :D

vasqu:
run-slow: mamba2

PR #39769, opened 2025-07-29
Fix Evolla and xLSTM tests

Cyrilvallez:
# What does this PR do?

As per the title

PR #43209, opened 2026-01-10
[xLSTM] Fix bugs preventing small model training

Anri-Lombard:
# What does this PR do?

Fixes three bugs in the xLSTM implementation that prevent training models smaller than 7B parameters.

Fixes #43208

## Changes

1. **Line 235**: Fixed typo where tensor was called as function instead of using `.reshape()`
2. **Lines 184-185**: Fixed shape mismatch by using `dqk // nc` instead of `dqk` for proper head dimension (verified against [NX-AI/xlstm original](https://github.com/NX-AI/xlstm/blob/main/xlstm/blocks/mlstm/backends.py))
3. **Line 783**: Fixed `return_last_states` default from `bool = False` to `bool | None = None` to match docstring and enable config-based control

## Testing

Tested with 125M parameter model (hidden_size=768, num_heads=4, layers=4):

**Before fix:**
```
RuntimeError: shape '[2, 4, 2, 64, 192]' is invalid for input of size 98304
```

**After fix:**
```
Model created successfully! Parameters: 129.01M
Forward pass successful!
Output shape: torch.Size([2, 128, 65536])
```

All existing xLSTM tests pass (59 passed, 162 skipped for hardware-specific tests).

## Before submitting
- [x] This PR fixes a typo or improves the docs (if applicable)
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md)?
- [x] Was this discussed/approved via a Github issue? #43208


src/transformers/models/xlstm/modeling_xlstm.py
```diff
@@ -180,8 +180,8 @@ def mlstm_chunkwise_parallel_fw_H(
         vecN_k_states = vecN_states.view(batch_size, nh, nc, dqk // nc)
         scaMinter_k_states = scaMinter_states
 
-        matQ = matQ.view(batch_size, nh, nc, chunk_size, dqk)
-        matK = matK.view(batch_size, nh, nc, chunk_size, dqk)
+        matQ = matQ.view(batch_size, nh, nc, chunk_size, dqk // nc)
+        matK = matK.view(batch_size, nh, nc, chunk_size, dqk // nc)
```
vasqu: I need more verification here, I don't think it's correct and maybe your custom config just has incompatible values

I looked into the original code and ours and it already splits correctly, e.g. going through q
- `bsz, seq_len, hidden_size`
- `bsz, seq_len, num_heads, head_dim`
- `bsz, num_heads, seq_len, head_dim` (this is what this function gets passed)
- `bsz, num_heads, num_chunks, chunk_size, head_dim` (this explicit reshape)

It does not make sense to make the hidden dimension dependent on the chunk size (just to fit arbitrary calculations). My guess is that the chunk size is not properly splitable with the sequence len and something goes wrong then, but this is not the correct solution

vasqu:
The other changes do make sense tho on a quick glance

src/transformers/models/xlstm/modeling_xlstm.py
Anri-Lombard: Thanks for taking a look! So I think it's correct - let me explain why.

At line 178, `dqk` is extracted from `matC_states` shape:
```python
batch_size, nh, dqk, dhv = matC_states.shape
```

However, `matC_states` has shape `(batch_size, nh, nc*dhqk, dhv)` where chunk states are concatenated, so `dqk` actually equals `nc × dhqk`, not the head dimension itself. This variable naming causes the confusion.

The original matQ has shape `(batch_size, nh, nc × chunk_size, dhqk)`. The buggy code tried:
```python
matQ.view(batch_size, nh, nc, chunk_size, dqk)  # where dqk = nc × dhqk
# Total elements: batch × nh × nc × chunk_size × (nc × dhqk) - mismatch!
```

The fix uses `dqk // nc` which equals `dhqk`:
```python
matQ.view(batch_size, nh, nc, chunk_size, dqk // nc)  # where dqk // nc = dhqk
# Total elements: batch × nh × nc × chunk_size × dhqk - correct!
```

This matches the original NX-AI implementation ([source](https://github.com/NX-AI/xlstm/blob/main/xlstm/blocks/mlstm/backends.py#L169-L171)) where queries and keys are reshaped using `DH` (the head dimension):
```python
q = queries.view(B, NH, NS, CS, DH) / math.sqrt(DH)
k = keys.view(B, NH, NS, CS, DH)
```

The `DH` in the original is exactly what `dqk // nc` gives us, which is why the 125M parameter model now works correctly.

vasqu: Gotcha, thanks for clarifying. I missed that we take the shape from `matC` and that that matrix has that shape. If I see it correctly, then these are the intermediate states after each chunk (and the initial state).

Imo, it would make more sense to predefine the shape `dhqk` then instead, making it clear that we have the head dim with this. Also, let's add a small test that forces the correct shape issue (with smaller values).

It seems that the integration tests have been skipped for quite a while now, checking if this fixes it :eyes:

vasqu: Seeing #39769, seems still to be broken (even with this fix)

Where these tests written with the kernels inference or with the pure torch version? cc @kpoeppel if you have some time

Anri-Lombard: > Gotcha, thanks for clarifying. I missed that we take the shape from `matC` and that that matrix has that shape. If I see it correctly, then these are the intermediate states after each chunk (and the initial state).
> 
> Imo, it would make more sense to predefine the shape `dhqk` then instead, making it clear that we have the head dim with this. Also, let's add a small test that forces the correct shape issue (with smaller values).
> 
> It seems that the integration tests have been skipped for quite a while now, checking if this fixes it 👀

Updated 👍

ArthurZucker:
@vasqu should we go ahead and merge?

vasqu (approved):
Sorry, yes forgot about this one 😓 the fixes are correct anyways, integration tests are sadly broken (for a while now)

vasqu:
Looks like the added test fails, can you check @Anri-Lombard?

Anri-Lombard:
Fixed @vasqu 🙏

vasqu:
run-slow: xlstm

vasqu:
Just for safety checking with run-slow but should be good, thanks for iterating :)

vasqu:
Thanks again @Anri-Lombard and sorry for the delays

Anri-Lombard:
No problem @vasqu, thanks for the reviews! 🙏

PR #43210, opened 2026-01-10
Fix Qwen2VLVideoProcessor ignoring runtime min_pixels/max_pixels kwargs

### LINKED ISSUE #43183: Regression in Qwen2_5_VLProcessor video kwargs routing (4.56.2 -> 4.57.1+): videos_kwargs{min_pixels,max_pixels} ignored
### System Info

Transformers versions: 4.56.2 (works), 4.57.1 (regression), 4.57.3 (still regressed)
numpy==2.2.6
pillow==12.1.0
torch==2.8.0+cu128
torchvision==0.23.0

CUDA 12.8
Python 3.10.12

CPU: Intel(R) Xeon(R) Platinum 8480C
GPU: NVIDIA H100 80GB HBM3

Model / processor: nvidia/cosmos-reason1-7b via AutoProcessor -> Qwen2_5_VLProcessor (video input)
Input: 1 video as np.uint8 array with shape T=10, H=1080, W=1920, C=3
Constraints used in repro: min_pixels=100352, max_pixels=151200

### Who can help?

@zucchini-nlp 

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

1) Enter python environment with specific transformers version
2) Run the repro script:
```bash
python test_qwen2_5_vl_kwargs_routing.py --model nvidia/cosmos-reason1-7b
```
2) Observe the printed `video_grid_thw` for these cases:
  A) `videos_kwargs={"min_pixels": min_pixels, "max_pixels": max_pixels}`
  B) `videos_kwargs={"size": {"shortest_edge": min_pixels, "longest_edge": max_pixels}}`
  C) `top-level size={"shortest_edge": min_pixels, "longest_edge": max_pixels}`

[test_qwen2_5_vl_kwargs_routing.py](https://github.com/user-attachments/files/24510556/test_qwen2_5_vl_kwargs_routing.py)

### Expected behavior

`videos_kwargs={"min_pixels":..., "max_pixels":...}` should continue to constrain video resizing (or raise a clear error / emit a deprecation warning if the API changed), because it works in transformers==4.56.2.

Concretely, for the same input and pixel budget, case A should produce a downsized grid similar to `video_grid_thw=[[5, 20, 36]]` (small patch/token count), not the large grid.

### What actually happens (summary of the issue)
In 4.56.2, case A downsizes as expected, but cases B/C do not.

In 4.57.1+, this flips: case A no longer downsizes (appears ignored), while cases B/C downsize.
This silent routing change causes a large increase in `video_grid_thw` (e.g., `[[5,20,36]]` -> `[[5,78,138]]`), which is ~15x more patches/visual tokens and leads to major performance regressions in downstream serving (e.g., vLLM throughput drop).

### API naming/semantics confusion
For Qwen2‑VL video resizing, the code path uses `size["shortest_edge"]` and `size["longest_edge"]` as pixel-budget bounds passed into `smart_resize(min_pixels=..., max_pixels=...)` (i.e., total pixel count constraints), not literal edge lengths in pixels. This makes it non-obvious that `size={"shortest_edge": min_pixels, "longest_edge": max_pixels}` is effectively the same as `{min_pixels, max_pixels}`, and it’s easy for users to interpret these keys as geometric edge lengths instead of area bounds. Is this intended?

Anri-Lombard:
## Summary
- Add `_further_process_kwargs` method to `Qwen2VLVideoProcessor` to handle runtime `min_pixels`/`max_pixels` kwargs
- Add test to verify runtime kwargs work correctly

## What does this PR do?

Fixes #43183

The video processor was missing the `_further_process_kwargs` method that the image processor has (see `image_processing_qwen2_vl_fast.py:87-108`). This caused `videos_kwargs{min_pixels, max_pixels}` to be silently ignored when passed at call time, resulting in ~15x more visual tokens than expected.

## Test plan
- [x] Added `test_video_processor_runtime_min_max_pixels` test
- [x] All existing video processor tests pass

### DIFF
--- a/src/transformers/models/qwen2_vl/video_processing_qwen2_vl.py
+++ b/src/transformers/models/qwen2_vl/video_processing_qwen2_vl.py
@@ -107,6 +107,29 @@ class Qwen2VLVideoProcessor(BaseVideoProcessor):
 
         super().__init__(size=size, min_pixels=min_pixels, max_pixels=max_pixels, **kwargs)
 
+    def _further_process_kwargs(
+        self,
+        size: SizeDict | None = None,
+        min_pixels: int | None = None,
+        max_pixels: int | None = None,
+        **kwargs,
+    ) -> dict:
+        """
+        Update kwargs that need further processing before being validated
+        Can be overridden by subclasses to customize the processing of kwargs.
+        """
+        if min_pixels is not None and max_pixels is not None:
+            size = {"shortest_edge": min_pixels, "longest_edge": max_pixels}
+        elif size is not None:
+            if "shortest_edge" not in size or "longest_edge" not in size:
+                raise ValueError("size must contain 'shortest_edge' and 'longest_edge' keys.")
+            min_pixels = size["shortest_edge"]
+            max_pixels = size["longest_edge"]
+        else:
+            size = {**self.size}
+
+        return super()._further_process_kwargs(size=size, min_pixels=min_pixels, max_pixels=max_pixels, **kwargs)
+
     def sample_frames(
         self,
         metadata: VideoMetadata,
--- a/src/transformers/models/video_llama_3/video_processing_video_llama_3.py
+++ b/src/transformers/models/video_llama_3/video_processing_video_llama_3.py
@@ -109,6 +109,29 @@ class VideoLlama3VideoProcessor(BaseVideoProcessor):
 
         super().__init__(size=size, min_pixels=min_pixels, max_pixels=max_pixels, **kwargs)
 
+    def _further_process_kwargs(
+        self,
+        size: SizeDict | None = None,
+        min_pixels: int | None = None,
+        max_pixels: int | None = None,
+        **kwargs,
+    ) -> dict:
+        """
+        Update kwargs that need further processing before being validated
+        Can be overridden by subclasses to customize the processing of kwargs.
+        """
+        if min_pixels is not None and max_pixels is not None:
+            size = {"shortest_edge": min_pixels, "longest_edge": max_pixels}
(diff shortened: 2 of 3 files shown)


brandonpelfrey:
Hi @Anri-Lombard thanks for this PR! Can you confirm that running your test without the fix failed in the same way described in the original issue? I am sure the test passes with your fix, I just want to make sure we know we're replicating the original problem.

Anri-Lombard:
Hi @brandonpelfrey, thanks for pointing that out 🙏

Yes, I confirmed the test replicates the original problem.

**Without the fix** (with `_further_process_kwargs` removed):
- The test fails with: `AssertionError: 256 not less than 256`
- Both default and constrained calls produce identical `video_grid_thw` values
- The runtime `min_pixels`/`max_pixels` kwargs are silently ignored

**With the fix:**
- The test passes
- Default processing with `max_pixels=224*224` produces 256 patches
- Constrained processing with `max_pixels=56*56` produces fewer patches as expected

I've also strengthened the test to verify that the grid size actually changes when custom pixel constraints are passed, rather than just checking for key existence.

brandonpelfrey:
Thanks @Anri-Lombard . Unfortunately, I'm not a maintainer 😃 but this looks like it will resolve the originally reported problem.

Rocketknight1:
I think this is a duplicate of #43228!

Anri-Lombard:
I guess since this one was first that would be the duplicate 😅

Rocketknight1:
lol, fair point, your one was first! That one is more advanced in the review process now though

PR #43211, opened 2026-01-10
Fix flaky SAM-HQ integration tests

### LINKED ISSUE #42890: tests/models/sam_hq/test_modeling_sam_hq.py::SamHQModelIntegrationTest may fail since a lot of cases are lack of set_seed()

engmohamedsalah:
## Summary

Fixes #42890 

SAM-HQ integration tests were failing due to non-deterministic positional embeddings. The `positional_embedding` parameter in `SamHQPositionalEmbedding` is:
1. Randomly initialized using `torch.randn()` (line 418 in `modeling_sam_hq.py`)
2. Excluded from checkpoint loading via `_keys_to_ignore_on_load_missing` (line 1236)

This meant every test run generated different random values, causing flaky test failures.

## Solution

Added `set_seed(0)` in the `setUp()` method of `SamHQModelIntegrationTest` class to ensure deterministic random initialization across all integration tests.

This approach:
- ✅ **KISS**: Single line in setUp() method
- ✅ **DRY**: One fix applies to all 14 integration test methods
- ✅ **Clean**: Follows existing transformers testing patterns
- ✅ **Minimal**: Only 8 lines added (including comments)

## Changes

- Added `from transformers.trainer_utils import set_seed` import
- Added `setUp()` method with `set_seed(0)` to `SamHQModelIntegrationTest` class
- Added clear comments explaining why seed is needed

## Test Plan

- [x] Python syntax check passed
- [ ] Integration tests will be run by CI


Rocketknight1:
As mentioned [here](https://github.com/huggingface/transformers/issues/42890#issuecomment-3739434107) I'm not certain setting the seed is the right solution, and the test failure might indicate an actual issue with weight loading somewhere. Holding off on merging this until it's investigated!

engmohamedsalah:
Thank you @Rocketknight1 for catching this! You were absolutely right - this is a **weight loading bug**, not just a randomness issue.

## Root Cause Analysis

Commit `0b369802cf` ("fix sam family!") removed `_tied_weights_keys` because positional embeddings are buffers (not parameters) and can't be tied. However, the commit left `_keys_to_ignore_on_load_missing` in place, which causes `prompt_encoder.shared_embedding.positional_embedding` to be **randomly initialized on every model load** instead of being loaded from the checkpoint.

The flaky tests were actually **exposing this bug** - outputs varied because one embedding was randomly initialized each time!

## Proper Fix

I've created a new PR with the correct fix: removing `_keys_to_ignore_on_load_missing` entirely so both positional embeddings are properly loaded from checkpoint.

The new PR: #TBD (will update once created)

Closing this PR in favor of the proper fix. Thank you again for the thorough review!

engmohamedsalah:
Update: New PR with proper fix is #43463

Issue #33698, opened 2024-09-25
Refactor `output_hidden_states` to allow index selection


hlky:
### Feature request

Models such as Llava use CLIPVision with `output_hidden_states=True` then select the index of `hidden_states` that it needs e.g. `image_outputs.hidden_states[vision_feature_layer]`.

`output_hidden_states` could be refactored to `output_hidden_states: Optional[Union[bool, int]] = None` and `hidden_states` output changed to `hidden_states: Optional[Union[torch.FloatTensor, Tuple[torch.FloatTensor, ...]]] = None`.

The layer index would need to be normalized to account for negative index:
```python
if type(output_hidden_states) is int and output_hidden_states < 0:
    output_hidden_states = min(len(self.layers), len(self.layers) + output_hidden_states + 1) # or config.num_hidden_layers
```

`CLIPEncoder` could be changed from
```python
for idx, encoder_layer in enumerate(self.layers):
    if output_hidden_states:
        encoder_states = encoder_states + (hidden_states,)
```
to
```python
for idx, encoder_layer in enumerate(self.layers):
    if type(output_hidden_states) is int and output_hidden_states == idx:
        encoder_states = hidden_states
    elif output_hidden_states:
        encoder_states = encoder_states + (hidden_states,)
```
and after the loop changed from
```python
if output_hidden_states:
    encoder_states = encoder_states + (hidden_states,)
```
to
```python
if type(output_hidden_states) is int and output_hidden_states == len(self.layers):
    encoder_states = hidden_states
elif output_hidden_states:
    encoder_states = encoder_states + (hidden_states,)
```

Models such as Llava would then be able to do:
```python
image_outputs = self.vision_tower(pixel_values, output_hidden_states=vision_feature_layer)
selected_image_feature = image_outputs.hidden_states
```

### Motivation

Memory efficiency, as per the [comment](https://github.com/huggingface/transformers/blob/68049b17a6bb4c9b0d499e9e77121effa2f5a6c0/src/transformers/models/llava/modeling_llava.py#L454) in Llava `"this is not memory efficient at all (output_hidden_states=True) will save all the hidden stated."`

This would also improve memory efficiency in CLIP Text, specifically in Diffusers pipelines such as Stable Diffusion XL where the penultimate layer is used or when `clip skip` is used.

### Your contribution

I can submit a PR.

amyeroberts:
Hi @hlky, thanks for this proposal! 

[...]

PR #43213, opened 2026-01-10
feat: allow output_hidden_states and output_attensions to record outputs of specific layers

cloudhan:
# What does this PR do?

This PR enable model forward to record optional outputs at specified layers. This will be particularly useful for large model with long context when explorering the aesthetics of the attention maps to design sparse attention. Without it, with moderate size model (say 7B), it can easily OOM with only 1k level context.

```py
outputs = model.forward(input_ids, output_hidden_states=10, output_attentions=[10])
```
now it only keeps `outputs.attentions[10]`, outputs of other layers are set to `None` to save memory.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


Rocketknight1:
Hmmn - this adds some code complexity that may complicate maintenance for a relatively niche feature. Probably the easiest way to extract these without huge memory is just to make a function that extracts the layers you want and then compile it? Torch should be able to correctly drop the unwanted tensors and avoid OOM in that scenario.

zucchini-nlp:
Might be useful for certain use-cases as in https://github.com/huggingface/transformers/issues/33698 ig. Depends on how much memory is being wasted on VLMs/diffusers, it's a common pattern for them to output hidden states

Rocketknight1:
I guess, but I think we should try to avoid adding features that need updates across all modeling code unless they're really important. Plus I think the `compile()` solution should work fine!

zucchini-nlp:
Yeah, we defi don't want to update all models. Anyway, just wanted to link related issues to have a full picture :)

cloudhan:
@Rocketknight1 this is actually a common pattern for this feature.
https://github.com/huggingface/transformers/blob/b40b7a767d569b067a115ce115fd270ac323cd36/src/transformers/models/qwen3/modeling_qwen3.py#L358-L361
and then it will be handled by the `generic.py`

Unless you want to support all models, you don't need to update all modeling. Only some legacy models are using per modeling pattern
https://github.com/huggingface/transformers/blob/b40b7a767d569b067a115ce115fd270ac323cd36/src/transformers/models/gpt_neo/modeling_gpt_neo.py#L449-L452

Rocketknight1:
Good point, I forgot we'd factored that out into `_can_record_outputs` for a lot of models. If you want to attempt a PR to update the generic code there to allow specific layer selection, I guess we could consider it! No guarantees, though, I'd need to talk with the rest of the team about whether the extra complexity is worth it

PR #43219, opened 2026-01-11
Fix tokenizer auto_map being ignored for custom models

Anri-Lombard:
Fixes #43202

PR #42894 introduced an early-exit to `TokenizersBackend` when `tokenizer_class` doesn't match the registered tokenizer for a `model_type`. However, this check was placed before the `auto_map` extraction, causing custom tokenizers (with `trust_remote_code=True`) to be ignored.

Reproduction:
```python
tokenizer = AutoTokenizer.from_pretrained("IQuestLab/IQuest-Coder-V1-40B-Loop-Instruct", trust_remote_code=True)
print(tokenizer.decode(tokenizer.encode("This is a test")))
# Expected: "This is a test"
# Actual (bug): "Th is <unk> is <unk> a <unk> te st"
```

For models with unregistered `model_type` (like `iquestloopcoder`), the condition `TOKENIZER_MAPPING_NAMES.get(model_type, "") != tokenizer_config_class` is always True, causing early-exit to `TokenizersBackend` without checking if `auto_map` exists.

This fix moves the `auto_map` extraction before the early-exit check and adds `tokenizer_auto_map is None` to the condition.

Added regression test `test_custom_tokenizer_with_mismatched_tokenizer_class`.


vasqu:
cc @itazap @ArthurZucker

ArthurZucker (approved):
This looks good to me! 
Thankd for also adding a test!

awni:
@Anri-Lombard @ArthurZucker thanks for the fix here. We are depending on it in mlx-lm and looking forward to the next RC.

ArthurZucker:
Can you fix the quality checks please?! 🤗

vasqu:
@bot /style

vasqu:
I was so free to make the style check myself to merge, thanks for contributing :heart:

Issue #43215, opened 2026-01-10
Add output_attentions flag to DinoV3


GaspTO:
### Feature request

Hi,

I was trying to output the attention maps of **DINOv3** as it is possible for other models, such as DETR, by passing the flag `output_attentions=True` during the main call:
```
model = DetrForObjectDetection.from_pretrained("facebook/detr-resnet-50")
url = "http://images.cocodataset.org/val2017/000000039769.jpg"
image = Image.open(requests.get(url, stream=True).raw)
inputs = processor(images=image, return_tensors="pt")
with torch.no_grad():
    outputs = model(**inputs, output_attentions=True)
```
But instead of DETR, I want to add for DINOV3.

Is this a good idea for a PR? I already made the changes locally, it's just about whether I open a PR.

### Motivation

Seeing the attention maps of a vision transformer, especially one trained with self-supervised learning, is a very important thing for studying and debugging it. In fact, this was shown in the paper of the first DINO model, called _Emerging Properties in Self-Supervised Vision Transformers_ , where the natural proclivity of these models to learn segmentation masks was seen in its attention maps.

I was trying to see these masks in DINOv3, but could not do it.

### Your contribution

I can do this change. In fact, it would be my first change, which would be a great introduction to, hopefully, more.

qgallouedec:
Hey @GaspTO, sorry for the slow reply, we're catching up on the backlog. DINOv3 already supports this, attentions are captured from every attention layer:

https://github.com/huggingface/transformers/blob/b3c3f5c15b3bfc30ea6d613e24841f73922235a5/src/transformers/models/dinov3_vit/modeling_dinov3_vit.py#L472-L475

What you're running into is that the model loads with SDPA by default, which never materializes the attention weights, so `attentions` comes back empty (with a warning). Load it with `attn_implementation="eager"` in `from_pretrained` and your `output_attentions=True` call returns one map per layer.

Closing since nothing needs to change on our side.

PR #43220, opened 2026-01-11
Add output_attentions to dinov3

GaspTO:
Add the changes proposed in the issue #43215 

(The issue has not been approved yet, but I already had these changes, so it made sense to just open a PR to not lose this).


zucchini-nlp (commented):
The model already can output attentions with `_can_record_outputs` and `check_model_inputs`. Is there any repro for when it didn't work?

GaspTO:
It seems you're correct. I'll be closing this.

PR #43157, opened 2026-01-07
Remove dtensor dependency in Tensor Parallel

3outeille:
# Motivation

- API stays the same but implementation details are different. It will be easier for us to debug whihc distributed calls are called we differentiated them by ourselves (example of `VocabParallelEmbedding`) and thus can put a breakpoint.
- Benchmark shows speedup with `torch.compile` on/off 

- An example with `ColwiseParallel`
```python
class _AllReduceBackward(torch.autograd.Function):
    """Identity forward, all-reduce backward. Used before colwise layers (f in Megatron)."""

    @staticmethod
    def forward(ctx, x, device_mesh):
        ctx.device_mesh = device_mesh
        return x

    @staticmethod
    def backward(ctx, grad_output):
        device_mesh = ctx.device_mesh
        if device_mesh.size() == 1:
            return grad_output, None
        dist.all_reduce(grad_output, op=dist.ReduceOp.SUM, group=device_mesh.get_group())
        return grad_output, None


def all_reduce_backward(x, device_mesh):
    """Identity forward, all-reduce backward. Use before colwise layers."""
    return _AllReduceBackward.apply(x, device_mesh)
    
class ColwiseParallel(TensorParallelLayer):
    """
    Column-wise parallel: weight is sharded on dim -2 (output features).
    Forward: input replicated -> output sharded on last dim.
    If gather_output=True, output is all-gathered to produce full tensor.
    """

    def __init__(self, gather_output: bool = False, **kwargs):
        super().__init__(**kwargs)
        self.gather_output = gather_output

    def _prepare_input_fn(self, mod, inputs, device_mesh):
        input_tensor = inputs[0] if inputs else inputs
        return all_reduce_backward(input_tensor, device_mesh)

    def _prepare_output_fn(self, mod, outputs, device_mesh):
        if self.gather_output:
            return all_gather(outputs, device_mesh)
        return outputs

    def shard_tensor(
        self,
        param,
        param_type=None,
        param_casting_dtype=None,
        to_contiguous=None,
        rank=None,
        device_mesh=None,
        tensor_idx=None,
    ):
        device_mesh = self.device_mesh
        empty_param = self.empty_param
        rank = self.rank
        if param_type == "bias":
            parameter = get_tensor_shard(param, empty_param, device_mesh, rank, -1, tensor_idx)
        else:
            parameter = get_tensor_shard(param, empty_param, device_mesh, rank, -2, tensor_idx)
        parameter = parameter.to(param_casting_dtype)
        return parameter, None

[...]

PR #43226, opened 2026-01-12
fix crash in when running FSDP2+TP

sywangyi:
- distributed: @3outeille @ArthurZucker


sywangyi:
running https://github.com/huggingface/accelerate/blob/main/examples/torch_native_parallelism/nd_parallel_trainer.py as accelerate launch --config-file configs/tp_hsdp.yaml nd_parallel_trainer.py. crash like
[rank1]: Traceback (most recent call last):
[rank1]:   File "/mnt/disk3/wangyi/accelerate/examples/torch_native_parallelism/nd_parallel_trainer.py", line 82, in <module>
[rank1]:     main()
[rank1]:   File "/mnt/disk3/wangyi/accelerate/examples/torch_native_parallelism/nd_parallel_trainer.py", line 77, in main
[rank1]:     trainer.train()
[rank1]:   File "/mnt/disk3/wangyi/transformers/src/transformers/trainer.py", line 2174, in train
[rank1]:     return inner_training_loop(
[rank1]:            ^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/mnt/disk3/wangyi/transformers/src/transformers/trainer.py", line 2339, in _inner_training_loop
[rank1]:     self.optimizer = self.accelerator.prepare(self.optimizer)
[rank1]:                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/mnt/disk3/wangyi/accelerate/src/accelerate/accelerator.py", line 1515, in prepare
[rank1]:     raise ValueError(
[rank1]: ValueError: When using FSDP2, a model and optimizer must be passed together to `Accelerator.prepare()` as the optimizer needs to have its parameters modified after the model is converted.

src/transformers/trainer.py
```diff
@@ -2335,7 +2335,7 @@ def _inner_training_loop(
             self.model.train()
             if hasattr(self.lr_scheduler, "step"):
                 # We should avoid accelerate preparing the model in TP case since we dont need it as it is handled by transformers from_pretrained and also it goes into DDP based preparation.
-                if self.is_tp_enabled:
+                if self.is_tp_enabled and not is_fsdp2:
                     self.optimizer = self.accelerator.prepare(self.optimizer)
                 else:
```
SunMarc: I think it should be safe to do that no and let accelerate handle the rest, it should make things less complicated.
```suggestion
```

src/transformers/trainer.py
```diff
@@ -2566,7 +2566,7 @@ def _inner_training_loop(
                                 _grad_norm = self.optimizer.clip_master_grads(args.max_grad_norm)
                             else:
                                 grad_norm_context = contextlib.nullcontext
-                                if self.is_tp_enabled:
+                                if self.is_tp_enabled and not self.is_fsdp_enabled:
```
SunMarc: why this won't work in case of fsdpv2 ? Also it seems that we are getting rid of dtensor for tp here so this will probably change https://github.com/huggingface/transformers/pull/43157#issuecomment-3738339409

SunMarc (commented):
Thanks ! Just a nit and we can merge

src/transformers/trainer.py
sywangyi: not could delete this, for trainer + TP case, using accelerator.prepare(self.model) will go to DDP part and through error.

3outeille:
making FSDP2 + TP integration smoother is in the roadmap. Almost done with TP, going to tackle FSDP2 next

src/transformers/trainer.py
SunMarc: it shouldn't no ? In prepare_model, it should go there: 

```python
elif self.parallelism_config and self.parallelism_config.tp_enabled:
                if not hasattr(model, "tp_size"):
                    raise NotImplementedError(
                        "Model should undergo tensor parallel before passing it to accelerate."
                        "You can use .from_pretrained(..., tp_plan='auto') if the model supports"
                    )
                if model.tp_size != self.parallelism_config.tp_size:
                    raise ValueError(
                        f"tp_size in the plugin {self.parallelism_config.tp_size} should be same as model's tp size {model.tp_size}"
                    )
```

instead of here:
``` python 
            if self.multi_device and not (self.parallelism_config and self.parallelism_config.tp_enabled):
                if model_has_dtensor(model):
                    raise ValueError(
                        "Your model contains `DTensor` parameters, which is incompatible with DDP. Maybe you loaded your model with `device_map='auto'`? Specify `device_map='cuda'` or 'cpu' instead."
                    )
                if any(p.requires_grad for p in model.parameters()):
                    kwargs = self.ddp_handler.to_kwargs() if self.ddp_handler is not None else {}
                    # TODO: Look at enabling native TP training directly with a proper config
                    if os.environ.get("ACCELERATE_BYPASS_DEVICE_MAP", "false") != "true":
                        if self.device.type == "hpu":
                            device_ids, output_device = [self.device.index], self.device.index
                        else:
                            device_ids, output_device = [self.local_process_index], self.local_process_index
                    else:
                        device_ids, output_device = None, None
                    model = torch.nn.parallel.DistributedDataParallel(
                        model, device_ids=device_ids, output_device=output_device, **kwargs
                    )
                    if self.ddp_handler is not None:
                        self.ddp_handler.register_comm_hook(model)
```

sywangyi: it's crash in tp only case, I fix the issue in https://github.com/huggingface/accelerate/pull/3908

SunMarc (approved):
Thanks a lot !

src/transformers/trainer.py
SunMarc: Nice, thanks !

ArthurZucker (approved):
Thanks!

PR #43227, opened 2026-01-12
GptOss experts implementation

### LINKED ISSUE #43193: `torch._grouped_mm` support for `GptOssExperts`
I've seen that https://github.com/huggingface/transformers/pull/42697 introduced the possibility of using grouped matrix multiplication for MoE layers. Any reason why `gpt-oss` was left out (I know that its mlp layer is decorated with `MegaBlocksMoeMLP`)?

IlyasMoutawwakil:
# What does this PR do?

[wip] Fixes #43193 @vasqu @mattteochen
generate

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


IlyasMoutawwakil:
run-slow: gpt_oss

src/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -64,86 +64,54 @@ def extra_repr(self):
...
-        self.expert_dim = self.intermediate_size
-        self.gate_up_proj = nn.Parameter(torch.zeros(self.num_experts, self.hidden_size, 2 * self.expert_dim))
-        self.gate_up_proj_bias = nn.Parameter(torch.zeros(self.num_experts, 2 * self.expert_dim))
-        self.down_proj = nn.Parameter(torch.empty((self.num_experts, self.expert_dim, self.hidden_size)))
-        self.down_proj_bias = nn.Parameter(torch.zeros(self.num_experts, self.hidden_size))
+        self.experts_are_transposed = True
+        self.gate_up_proj = nn.Parameter(torch.empty(self.num_experts, self.hidden_size, 2 * self.intermediate_size))
+        self.gate_up_proj_bias = nn.Parameter(torch.empty(self.num_experts, 2 * self.intermediate_size))
+        self.down_proj = nn.Parameter(torch.empty((self.num_experts, self.intermediate_size, self.hidden_size)))
+        self.down_proj_bias = nn.Parameter(torch.empty(self.num_experts, self.hidden_size))
```
IlyasMoutawwakil: i avoided transposing the weights and made minimal changes that don't impact how computation is done in eager mode because the slow tests are already failing on main from very long so it's hard to debug.

IlyasMoutawwakil: @ArthurZucker keeping the weights like this also avoids problems with megablocks expectations

src/transformers/integrations/moe.py
```diff
@@ -230,6 +254,15 @@ def __init__(self, config, *args, **kwargs):
...
         experts_forward = original_forward
 
+        if self.config._experts_implementation == "grouped_mm":
+            if self.gate_up_proj.data_ptr() % 16 != 0 or self.down_proj.data_ptr() % 16 != 0:
+                logger.warning(
+                    "'grouped_mm' experts implementation requires 16-byte aligned expert weights. "
+                    "We will fall back to 'eager' implementation to avoid potential crashes. "
+                    "Please re-initialize the expert weights with 16-byte alignment."
+                )
+                self.config._experts_implementation = "eager"
```
IlyasMoutawwakil: this only happens in tests that load, save and then reload the model on cpu, 
the problem is that `safetensors` adds metadata that's only 8 bytes aligned, causing the model weights to start from an offset of 8 bytes.

IlyasMoutawwakil: the expert weights not being 16 bytes aligned shouldn't be an issue on cpu though because that constraint only applies to cuda kernels using TMA https://github.com/pytorch/pytorch/issues/172440

src/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -64,86 +64,54 @@ def extra_repr(self):
...
         super().__init__()
         self.intermediate_size = config.intermediate_size
         self.num_experts = config.num_local_experts
         self.hidden_size = config.hidden_size
-        self.expert_dim = self.intermediate_size
-        self.gate_up_proj = nn.Parameter(torch.zeros(self.num_experts, self.hidden_size, 2 * self.expert_dim))
-        self.gate_up_proj_bias = nn.Parameter(torch.zeros(self.num_experts, 2 * self.expert_dim))
-        self.down_proj = nn.Parameter(torch.empty((self.num_experts, self.expert_dim, self.hidden_size)))
-        self.down_proj_bias = nn.Parameter(torch.zeros(self.num_experts, self.hidden_size))
+        self.experts_are_transposed = True
```
ArthurZucker: this is the only thing I don't like! 
We could just put it like above ⬆️

src/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -64,86 +64,54 @@ def extra_repr(self):
         return f"{tuple(self.weight.shape)}, eps={self.variance_epsilon}"
 
 
+@use_experts_implementation
```
ArthurZucker:
```suggestion
@use_experts_implementation(transposed_experts=True)
```

src/transformers/models/gpt_oss/modeling_gpt_oss.py
```diff
@@ -64,86 +64,54 @@ def extra_repr(self):
...
-        """
-        When training it is more efficient to just loop over the experts and compute the output for each expert
-        as otherwise the memory would explode.
+    def _apply_gate(self, gate_up: torch.Tensor) -> torch.Tensor:
+        gate, up = gate_up[..., ::2], gate_up[..., 1::2]
+        gate = gate.clamp(min=None, max=self.limit)
+        up = up.clamp(min=-self.limit, max=self.limit)
+        glu = gate * torch.sigmoid(gate * self.alpha)
+        gated_output = (up + 1) * glu
+        return gated_output
```
ArthurZucker, line 87: This is fine, makes it more readable!

ArthurZucker (commented):
Looks good! Let's agree on transposed vs not transposed (shape ccould help?)

src/transformers/integrations/moe.py
vasqu: Maybe adding the link to the issue? Would be nice for users I think if they encounter this (which shouldn't really happen but you never know)

src/transformers/models/gpt_oss/modeling_gpt_oss.py
vasqu: Do we need to init with `torch.empty`? Just thinking about the case where we would not use `from_pretrained`, these weights would be random (I think)

src/transformers/integrations/moe.py
```diff
@@ -176,19 +189,30 @@ def grouped_mm_experts_forward(
...
+        # Default gating mechanism
+        gate, up = gate_up_out.chunk(2, dim=-1)  # (S, intermediate_dim)
+        hidden_after_activation = self.act_fn(gate) * up  # (S, intermediate_dim)
 
     # --- Down projection per expert (grouped_mm) ---
-    out_per_sample_g = torch._grouped_mm(hidden_after_activation, self.down_proj.transpose(-2, -1), offs=offsets)
+    if getattr(self, "experts_are_transposed", False):
+        out_per_sample_g = torch._grouped_mm(hidden_after_activation, self.down_proj, offs=offsets)
+    else:
+        out_per_sample_g = torch._grouped_mm(hidden_after_activation, self.down_proj.transpose(-2, -1), offs=offsets)
```
vasqu: A lot of these functions essentially only either change the order or add a transpose. Could we somehow solve this more elegantly with `functools.partial` or similarly? Just not a fan of all these if-else's

vasqu (commented):
Hmm, so we no longer transpose during weight conversion but do it manually? Not against it but we will need to solve the loading/reverse issue either way, not necessarily here 

Left a few comments but they are mostly aesthetic / nits

IlyasMoutawwakil:
> Not against it but we will need to solve the loading/reverse issue either way, not necessarily here

yes and the transposition can be achieved and works fine (all tests passà with the temporary weight renaming trick, however the biggest problem here is that megablocks expect the transposed format of expert weights so it might make more sense to transpose the other MoEs to be compatible with megablocks (if it's worth the speedup)

src/transformers/models/gpt_oss/modeling_gpt_oss.py
IlyasMoutawwakil: not really needed, but I thought that using empty initializer will uncover any uninitialized weights since the real initialization should be done in `_init_weights` iiuc, maybe @Cyrilvallez

vasqu:
> the biggest problem here is that megablocks expect the transposed format of expert weights

Argh, yea this will be hard to solve otherwise (would need to rewrite that kernel). Maybe we transpose the other way around for the other models at some point - depends on how much improvement that kernel provides 

On another note, is the transpose expensive on runtime? I would expect slight perf regression having to call transpose here and there (in `grouped_mm` implementation, I see a few).

IlyasMoutawwakil:
> On another note, is the transpose expensive on runtime?

no, the grouped_mm work on strided tensors so the transpose is just a view (no op).

src/transformers/integrations/moe.py
IlyasMoutawwakil: @vasqu good news i was able to force 16 bytes alignment using the weight loader.

src/transformers/conversion_mapping.py
```diff
@@ -332,6 +345,6 @@ def get_model_conversion_mapping(
 
     # Add the ones from the quantizer as well if provided
     if hf_quantizer is not None:
-        weight_conversions.extend(hf_quantizer.get_weight_conversions())
+        weight_conversions = hf_quantizer.get_weight_conversions() + weight_conversions
```
IlyasMoutawwakil: it's seems that the order of operations matters when both quantization conversions and weight conversions are defined. e.g. with gpt_oss if I want to apply the Force16ByteAlignment conversion, it has to be placed after the dequantization otherwise the loader is confused (see output below). @vasqu @ArthurZucker 

```python
device = torch.device("cuda:1")
model = GptOssForCausalLM.from_pretrained(
    model_id,
    device_map=device,
    dtype=torch.bfloat16,
    quantization_config=Mxfp4Config(dequantize=True),
).eval()
```
results in:
```
Loading weights: 100%|█████████████████████████████████████████████| 363/363 [00:00<00:00, 1003.90it/s, Materializing param=model.norm.weight]
GptOssForCausalLM LOAD REPORT from: openai/gpt-oss-20b
Key                                                   | Status     | 
------------------------------------------------------+------------+-
model.layers.{0...23}.mlp.experts.down_proj_scales    | UNEXPECTED | 
model.layers.{0...23}.mlp.experts.down_proj_blocks    | UNEXPECTED | 
model.layers.{0...23}.mlp.experts.gate_up_proj_scales | UNEXPECTED | 
model.layers.{0...23}.mlp.experts.gate_up_proj_blocks | UNEXPECTED | 
model.layers.{0...23}.mlp.experts.gate_up_proj        | MISSING    | 
model.layers.{0...23}.mlp.experts.down_proj           | MISSING    | 

Notes:
- MISSING       :those params were newly initialized because missing from the checkpoint. Consider training on your downstream task.
```

src/transformers/models/gpt_oss/modeling_gpt_oss.py
Cyrilvallez: Will not really help to uncover, but yes it's preferred to have `empty` rather than `zeros` etc (though now that we really init on meta device, it does not really change anything, more for the convention)

src/transformers/conversion_mapping.py
IlyasMoutawwakil: also @Cyrilvallez

Cyrilvallez: Yes! Since `get_weight_conversions` is only to DEQUANTIZE it makes sense that it's first. However, as discussed offline, for now it's not possible to match 1 param with 2 converter (i.e. 1 dequant converter, and 1 from the hardcoded mapping). So it means that any model with a mapping registered cannot be dequantized 😭 So in theory you're right it should come first, but as it does not work anyway currently it does not make a difference 😭 Let's still keep this change adding a comment on all that please!

IlyasMoutawwakil: thanks @Cyrilvallez added a comment mostly rewording your explanation !

tests/models/gpt_oss/test_modeling_gpt_oss.py
```diff
@@ -120,6 +120,9 @@ def test_flex_attention_with_grads(self):
     def test_generate_compile_model_forward_fullgraph(self):
         return super().test_generate_compile_model_forward_fullgraph()
 
+    def test_reverse_loading_mapping(self, check_keys_were_modified=False):
+        super().test_reverse_loading_mapping(check_keys_were_modified)
```
vasqu, line 124: I think it's still important to have the test running just to be sure, so I added a kwarg for cases where we do not modify keys

tests/test_modeling_common.py
```diff
@@ -4499,8 +4505,9 @@ def test_reverse_loading_mapping(self):
                     # Get all the serialized keys that we just saved according to the reverse mapping
                     serialized_keys = list(state_dict.keys())
 
-                # They should be different, otherwise we did not perform any mapping
-                self.assertNotEqual(sorted(serialized_keys), sorted(model_keys), "No key mapping was performed!")
+                if check_keys_were_modified:
+                    # They should be different, otherwise we did not perform any mapping
+                    self.assertNotEqual(sorted(serialized_keys), sorted(model_keys), "No key mapping was performed!")
```
vasqu, line 4515: This is what I meant

src/transformers/models/gpt_oss/modular_gpt_oss.py
```diff
@@ -61,86 +61,52 @@ def forward(self, hidden_states):
         return (self.weight * hidden_states).to(input_dtype)  # main diff with Llama
 
 
+@use_experts_implementation(is_transposed=True, has_bias=True)
```
vasqu, line 64: Nice, the additional bias is a good addition!

src/transformers/conversion_mapping.py
```diff
@@ -347,6 +360,10 @@ def get_model_conversion_mapping(
 
     # Add the ones from the quantizer as well if provided
     if hf_quantizer is not None:
-        weight_conversions.extend(hf_quantizer.get_weight_conversions())
+        # NOTE: Since get_weight_conversions() only serves to dequantize, we need to put them first in the list.
+        # However, for now it's not possible to match 1 param with 2 converters (i.e. 1 dequantization converter
+        # and 1 model-specific converter. Which means that if a model that has model-specific conversions and is being
+        # dequantized, the model-specific conversion that has patterns matching the dequantization patterns will be ignored.
```
vasqu:
```suggestion
        # NOTE: Since get_weight_conversions() only serves to dequantize, we need to put them first in the list.
        # However, for now it's not possible to match 1 param with 2 converters (i.e. 1 dequantization converter
        # and 1 model-specific converter). Which means that if a model that has model-specific conversions and is being
        # dequantized, the model-specific conversion that has patterns matching the dequantization patterns will be ignored.
```

src/transformers/core_model_loading.py
```diff
@@ -440,6 +440,40 @@ def reverse_op(self) -> ConversionOps:
         return ErnieFuseAndSplitTextVisionExperts(stack_dim=self.stack_dim, concat_dim=self.concat_dim)
 
 
+class Force16BytesAlignment(ConversionOps):
+    """
+    Ensures that the given tensor is 16-bytes aligned in memory.
```
vasqu: Wdyt about giving an explicit use case? I.e. `torch._grouped_mm` - might come off weird otherwise what's behind this converter

src/transformers/core_model_loading.py
```diff
@@ -440,6 +440,40 @@ def reverse_op(self) -> ConversionOps:
...
+    """
+
+    @torch.no_grad()
+    def convert(
+        self, input_dict: dict[str, torch.Tensor], source_patterns: list[str], target_patterns: list[str], **kwargs
+    ) -> dict[str, torch.Tensor]:
+        target_pattern = self.get_target_pattern(input_dict, source_patterns, target_patterns)
+        tensors = next(iter(input_dict.values()))
+        tensor = tensors[0] if isinstance(tensors, list) else tensors
+        return {target_pattern: tensor.clone()}
```
vasqu: I copied the fix on transpose; what I'm unsure about is whether we always clone or not (in the original version I saw it being cloned either way, might be wrong)

src/transformers/integrations/moe.py
```diff
@@ -75,54 +108,93 @@ def batched_mm_experts_forward(
     final_hidden_states = torch.zeros_like(hidden_states)
 
     # Flatten top_k_index to get expert_ids per selected sample
-    expert_ids = top_k_index.reshape(-1)
+    expert_ids = top_k_index.view(-1)
```
vasqu: Trusting you on this, I always feel a bit unsafe with `view` xD

src/transformers/integrations/moe.py
```diff
@@ -75,54 +108,93 @@ def batched_mm_experts_forward(
...
     # - (num_tokens, num_top_k) Qwen2MoE style
     # - (num_tokens, num_experts) DeepseekV2 style
     if top_k_weights.shape == (num_tokens, num_top_k):
-        sample_weights = top_k_weights.reshape(-1)  # (S,)
+        sample_weights = top_k_weights.view(-1, 1)  # (S, 1)
     elif top_k_weights.shape == (num_tokens, num_experts):
-        sample_weights = top_k_weights[token_idx, expert_ids]  # (S,)
+        # TODO: routers that output full expert distribution
+        # should probably be corrected to output only top_k weights
+        sample_weights = top_k_weights[token_idx, expert_ids].view(-1, 1)  # (S, 1)
```
vasqu: This is from e.g. deepseek v2? Or at least before it was fixed

src/transformers/integrations/moe.py
```diff
@@ -141,60 +213,64 @@ def grouped_mm_experts_forward(
...
     expert_ids_g = expert_ids[perm]
     sample_weights_g = sample_weights[perm]
-    current_states_g = current_hidden_states[perm]
+    selected_hidden_states_g = selected_hidden_states[perm]
+
+    # Select expert weights and biases for selected samples
+    # NOTE: We keep all experts here and rely on offsets to target the active ones.
+    # I have already implemented a version that only passes the active experts, but
+    # to do so I had to use torch.unique which breaks the graph capture (data-dependent).
+    # Also there were no speedup gains from it in my experiments, even in eager mode.
```
vasqu, line 251: Did it change memory requirements? But honestly I rather have a torch compile friendly version either way

src/transformers/integrations/moe.py
```diff
@@ -141,60 +213,64 @@ def grouped_mm_experts_forward(
...
+    selected_down = self.down_proj
+    selected_gate_up_bias = self.gate_up_proj_bias[expert_ids_g] if self.has_bias else None
+    selected_down_bias = self.down_proj_bias[expert_ids_g] if self.has_bias else None
 
     # Compute offsets for grouped_mm
     # using histc instead of bincount to avoid cuda graph issues
-    # (grouped_mm_experts_forward still fails with cuda graphs but because of _grouped_mm internals)
     num_tokens_per_expert = torch.histc(expert_ids_g.float(), bins=num_experts, min=0, max=num_experts - 1)
-    offsets = torch.cumsum(num_tokens_per_expert, dim=0, dtype=torch.int32)
+    offs = torch.cumsum(num_tokens_per_expert, dim=0, dtype=torch.int32)
```
vasqu: Any reason to shorten this, let's keep it as fullname if possible

src/transformers/integrations/moe.py
```diff
@@ -217,24 +293,56 @@ class ExpertsInterface(GeneralInterface):
...
+
+        if not hasattr(experts_class, "_apply_gate"):
 
-    @wraps(original_init)
-    def __init__(self, config, *args, **kwargs):
-        original_init(self, config, *args, **kwargs)
-        self.config = config
+            def _apply_gate(self, gate_up_out: torch.Tensor) -> torch.Tensor:
+                gate, up = gate_up_out.chunk(2, dim=-1)  # (S, intermediate_dim)
+                return self.act_fn(gate) * up  # (S, intermediate_dim)
```
vasqu: This makes it a bit hidden, can we move this a bit outside?

vasqu (commented):
I added some changes to the conversion + test, so please take another look there 

Most comments are more nits about style / fmi

src/transformers/core_model_loading.py
IlyasMoutawwakil: ah yes it's a typo, only clone if non-aligned

src/transformers/integrations/moe.py
IlyasMoutawwakil: yep exactly

src/transformers/integrations/moe.py
IlyasMoutawwakil: no it's the same memory footprint, it's also sometimes slower because the torch.unique overhead.
passing all experts seems to be the intended way, the kernel takes care of loading active experts to memory.

src/transformers/integrations/moe.py
IlyasMoutawwakil: because grouped_mm names it offs, no opinion about it, reverting to full name

src/transformers/integrations/moe.py
IlyasMoutawwakil: makes sense ! moving

src/transformers/models/gpt_oss/modular_gpt_oss.py
IlyasMoutawwakil: yeah better be explicit than rely on getattrs

tests/models/gpt_oss/test_modeling_gpt_oss.py
IlyasMoutawwakil: makes sense ! thanks

src/transformers/core_model_loading.py
IlyasMoutawwakil: added

vasqu (approved):
LGTM, I guess the outputs slightly change now with the interface's `grouped_mm/batched_mm` or even the eager rewrite?

IlyasMoutawwakil:
> LGTM, I guess the outputs slightly change now with the interface's grouped_mm/batched_mm or even the eager rewrite?

the eager for training is the same, I kept everything as is when i realized that the slow tests are brokens (see https://github.com/huggingface/transformers/pull/43246), I removed the batched implementation from there as it is redundant, so users will have to set experts impl to batched to get the same eval output as before.

vasqu:
Probably merging next week, currently don't have the permission to merge with red CI :smile:

src/transformers/integrations/moe.py
IlyasMoutawwakil: > When possible, the returned tensor will be a view of input

you are right ! they should be equivalent if tensor mem is contiguous (when possible) so reverting to reshapes !

src/transformers/conversion_mapping.py
```diff
@@ -347,6 +360,10 @@ def get_model_conversion_mapping(
 
     # Add the ones from the quantizer as well if provided
     if hf_quantizer is not None:
-        weight_conversions.extend(hf_quantizer.get_weight_conversions())
+        # NOTE: Since get_weight_conversions() only serves to dequantize, we need to put them first in the list.
+        # However, for now it's not possible to match 1 param with 2 converters (i.e. 1 dequantization converter
+        # and 1 model-specific converter). Which means that if a model that has model-specific conversions and is being
+        # dequantized, the model-specific conversion that has patterns matching the dequantization patterns will be ignored.
+        weight_conversions = hf_quantizer.get_weight_conversions() + weight_conversions
```
ArthurZucker: I am not sure it makes a difference at all no? Because the operation are ordered by length of collected tensors I thing.

src/transformers/core_model_loading.py
```diff
@@ -440,6 +440,42 @@ def reverse_op(self) -> ConversionOps:
         return ErnieFuseAndSplitTextVisionExperts(stack_dim=self.stack_dim, concat_dim=self.concat_dim)
 
 
+class Force16BytesAlignment(ConversionOps):
+    """
+    Ensures that the given tensor is 16-bytes aligned in memory and clones it if not.
+    This garantees 16-bytes alignmenet for kernels / implementations that use TMA or SIMD instructions like torch._grouped_mm.
+    """
```
ArthurZucker, line 447: very nice 🫡

ArthurZucker (approved):
Great work! LGTM, lets make sure quantized tests pass but let's go!

src/transformers/conversion_mapping.py
IlyasMoutawwakil: Yeah now that I'm digging deeper into the weight loader I think the reason I got the error above is because it's not possible to cascade converters (i.e., applying model-specific conversions on top of tensors created by the dequantization conversions). Not because you can't match one tensor with two converters (that's a valid limitation, but not the one happening here in gpt oss).

I added $ to the end of Force16BytesAlignement source pattern to fix my error without changing this order. Basically making sure that the sources of the mxfp dequant converter and Force16BytesAlignement are exclusive.
I will revert the line change and make this comment instead:
```py
        # NOTE: Since get_weight_conversions() only serve to dequantize, we normally want to apply them first.
        # However, for now it's not possible to cascade converters (i.e., applying model-specific conversions on top
        # of tensors created by the dequantization conversions)
        # This means that if a model has model-specific conversions and is being dequantized, the model-specific conversion
        # that relies on tensors created by dequantization conversions will not be applied.
        # GptOss example: with Mxfp4Config(dequantize=True), Force16BytesAlignment converters are ignored because the tensors
        # "mlp.experts.gate_up_proj$" and "mlp.experts.down_proj$" are only created after dequantization conversions are applied.
        weight_conversions.extend(hf_quantizer.get_weight_conversions())
```

IlyasMoutawwakil:
Should be good to merge

vasqu:
#43353 leads me to believe that a simple weight converter won't be enough. Somehow under the concurrent workers we hit the same 16byte alignment issue on our CI with (qwen 3 omni moe).

I'll take a proper look tomorrow but it might make sense to add the fallback regardless sadly. Just wanted to put it out before we hastly merge

IlyasMoutawwakil:
@vasqu I think what you are dealing with might be the cpu+safetensors memory misalignement issue i have seen with gptoss ; there are two kinds of misalignment issues i have encountered so far:
- on cuda: this only happens if the strided tensor has shapes that are not divisible by 8 (for fp16/bf16) or 4 (for fp32), otherwise the cuda allocator will allocate an aligned tensor. That's the issue I was solving in the first experts PR, because I was testing on the dgx so tests were automatically running on cuda.
- on cpu: Safetensors adds an 8 bytes-aligned metadata on top of the file ([see this issue from which the 8 bytes were decided](https://github.com/huggingface/safetensors/issues/231)),  so when reloaded on cpu the metadata can "misalign" the tensors in the memmaped file. This is what the weight converter defined here solves.

I would suggest you try the weight converter defined here because the ci runner uses cpu (I had to set TRANSFORMERS_TEST_DEVICE=cpu to reproduce the gpt_oss issue locally)

IlyasMoutawwakil:
To sum up the conversation I had with @vasqu internally:

> All moes that have a converter that creates new tensors, for example all the ones that merge their experts using cat/stack, will create 16-bytes aligned tensors on cpu. This is because they will use the allocator to create *new* merged tensors, and [the allocator will return a 16-bytes aligned memory](https://stackoverflow.com/questions/70692795/why-is-malloc-16-byte-aligned). 
If any MoE fails because of 16 bytes misalignment, it's gonna be because it's not allocating its tensors ; for example using safetensors memmaped tensors (which are only 8 bytes aligned). Those are the ones that will need the Force16BytesAlignement converter.

src/transformers/conversion_mapping.py
ArthurZucker: cool, as we talked offline let's add more comments about how to handle the weight converter

ArthurZucker (approved):

PR #43228, opened 2026-01-12
Qwen-VL video processor accepts min/max pixels

### LINKED ISSUE #43183: Regression in Qwen2_5_VLProcessor video kwargs routing (4.56.2 -> 4.57.1+): videos_kwargs{min_pixels,max_pixels} ignored

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/43183 and makes sure we can pass `min/max pixels` to video processor's call.

### DIFF
--- a/src/transformers/models/qwen2_vl/video_processing_qwen2_vl.py
+++ b/src/transformers/models/qwen2_vl/video_processing_qwen2_vl.py
@@ -122,7 +122,7 @@ class Qwen2VLVideoProcessor(BaseVideoProcessor):
             size = {"shortest_edge": min_pixels, "longest_edge": max_pixels}
         elif size is not None:
             if "shortest_edge" not in size or "longest_edge" not in size:
-                raise ValueError("size must contain 'shortest_edge' and 'longest_edge' keys.")
+                raise ValueError("dictionary `size` must contain 'shortest_edge' and 'longest_edge' keys.")
             min_pixels = size["shortest_edge"]
             max_pixels = size["longest_edge"]
         else:
--- a/src/transformers/models/qwen2_vl/video_processing_qwen2_vl.py
+++ b/src/transformers/models/qwen2_vl/video_processing_qwen2_vl.py
@@ -107,6 +107,29 @@ class Qwen2VLVideoProcessor(BaseVideoProcessor):
 
         super().__init__(size=size, min_pixels=min_pixels, max_pixels=max_pixels, **kwargs)
 
+    def _further_process_kwargs(
+        self,
+        size: SizeDict | None = None,
+        min_pixels: int | None = None,
+        max_pixels: int | None = None,
+        **kwargs,
+    ) -> dict:
+        """
+        Update kwargs that need further processing before being validated
+        Can be overridden by subclasses to customize the processing of kwargs.
+        """
+        if min_pixels is not None and max_pixels is not None:
+            size = {"shortest_edge": min_pixels, "longest_edge": max_pixels}
+        elif size is not None:
+            if "shortest_edge" not in size or "longest_edge" not in size:
+                raise ValueError("size must contain 'shortest_edge' and 'longest_edge' keys.")
+            min_pixels = size["shortest_edge"]
+            max_pixels = size["longest_edge"]
+        else:
+            size = {**self.size}
+
+        return super()._further_process_kwargs(size=size, min_pixels=min_pixels, max_pixels=max_pixels, **kwargs)
+
     def sample_frames(
         self,
         metadata: VideoMetadata,
--- a/src/transformers/models/video_llama_3/modular_video_llama_3.py
+++ b/src/transformers/models/video_llama_3/modular_video_llama_3.py
@@ -1474,8 +1474,8 @@ class VideoLlama3VideoProcessor(Qwen2VLVideoProcessor):
     model_input_names = ["pixel_values_videos", "video_grid_thw", "video_merge_sizes", "video_compression_mask"]
 
     def _further_process_kwargs(self):
-        raise AttributeError("VideoLlama3 never supported min/max pixels")
-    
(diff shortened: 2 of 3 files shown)


zucchini-nlp:
@bot /style

src/transformers/models/video_llama_3/modular_video_llama_3.py
```diff
@@ -1473,6 +1473,9 @@ class VideoLlama3VideoProcessor(Qwen2VLVideoProcessor):
     valid_kwargs = VideoLlama3VideoProcessorInitKwargs
     model_input_names = ["pixel_values_videos", "video_grid_thw", "video_merge_sizes", "video_compression_mask"]
 
+    def _further_process_kwargs(self):
+        raise AttributeError("VideoLlama3 never supported min/max pixels")
```
molbap: maybe a bit confusing to super() a very generic method with a very specific error message

src/transformers/models/qwen2_vl/video_processing_qwen2_vl.py
```diff
@@ -107,6 +107,29 @@ def __init__(self, **kwargs: Unpack[Qwen2VLVideoProcessorInitKwargs]):
...
+    ) -> dict:
+        """
+        Update kwargs that need further processing before being validated
+        Can be overridden by subclasses to customize the processing of kwargs.
+        """
+        if min_pixels is not None and max_pixels is not None:
+            size = {"shortest_edge": min_pixels, "longest_edge": max_pixels}
+        elif size is not None:
+            if "shortest_edge" not in size or "longest_edge" not in size:
+                raise ValueError("size must contain 'shortest_edge' and 'longest_edge' keys.")
```
molbap: formatting suggestion
```suggestion
                raise ValueError("dictionary `size` must contain 'shortest_edge' and 'longest_edge' keys.")
```

molbap (approved):
Thanks for the fix!

src/transformers/models/video_llama_3/modular_video_llama_3.py
zucchini-nlp: it's to bypass modular, modular doesn't allow to delete methods from parent class unless we do `AttributeError`

molbap: yes I understand, I mean the `raise` message is a bit confusing - the method itself is fine

brandonpelfrey:
@zucchini-nlp Thanks for helping here. I asked this in the other PR that was trying to fix this: Can you confirm that the new unit test failed without the fix and after the fix the unit test passes?

brandonpelfrey:
@zucchini-nlp Looks like the only blocker left here is some whitespace introduced on line 1478. I am not sure but I think the other CI failure appears like a transient glitch.

Rocketknight1:
cc @zucchini-nlp `test_processing_qwen2_5_omni.py::Qwen2_5OmniProcessorTest::test_apply_chat_template_video_frame_sampling` is failing in the CI after this PR! I'm not sure if we just need to update the test values or if this uncovered some deeper bug in the handling for `size` and `min/max_pixels`

PR #43230, opened 2026-01-12
simplify extra tokens logic in base

itazap:
simplify extra tokens logic in base

some light refactoring, hopefully a bit more readable


itazap:
[run-sl](run-slow: florence2, nllb, tapas)

src/transformers/tokenization_utils_base.py
```diff
@@ -1251,81 +1246,50 @@ def pad_token_type_id(self) -> int:
...
     def __setattr__(self, key, value):
-        key_without_id = key
-        key_is_special_id = key.endswith("_id") or key.endswith("_ids")
-        if key_is_special_id:
-            key_without_id = key[:-3] if not key.endswith("_ids") else key[:-4]
+        # Handle _id/_ids suffix (eg. bos_token_id -> bos_token)
+        key_without_id = key.removesuffix("_ids").removesuffix("_id") if key.endswith(("_id", "_ids")) else key
+
+        # Named special tokens (bos_token, eos_token, etc.)
+        if key_without_id in self.SPECIAL_TOKENS_ATTRIBUTES:
```
ArthurZucker, line 1253: at this point special tokens attributes contains the `extra_special_tokens`?

ArthurZucker: Because that's the most important point: you want the extra special token to define the names that are allowed in the set attribute and get attributes

ArthurZucker (approved):
One of the main thing that is missing is an update to the documentation for the V5 transition. Basically just explicitly that this argument is just to use when initializing.

Otherwise it looks good to me thanks for the simplification. Given the extensive tests that they pass I suppose it's all good!

src/transformers/tokenization_utils_base.py
itazap: no "extra_special_tokens" would not be a part of any other attribute ! it would always be separate

ArthurZucker:
Force merging as the errors are overall the same as on main

Rocketknight1:
Seeing some processor errors after this PR including `Qwen3VLProcessorTest::test_apply_chat_template_video_frame_sampling` and `Glm46VProcessorTest::test_apply_chat_template_video_frame_sampling`, plus possibly others. I think the CI would have caught them but it was hard to tell in the cloud of other errors!

Rocketknight1:
Pushed fixes for Glm46 and Glm4 in #43537

PR #43233, opened 2026-01-12
Use `HF_TOKEN` directly and remove `require_read_token` 

ydshieh:
# What does this PR do?

This makes the process more smooth: we only need to make sure the bot has access to the gated repositories.


ydshieh:
run-slow: whisper, llama

ydshieh:
run-slow: whisper, llama

PR #43244, opened 2026-01-13
Add Solar-Open Model

oesni:
# What does this PR do?

Implements Solar-Open model. 
Solar Open is the open-weights MoE Solar LLM created by Upstage.
- model repo: https://huggingface.co/upstage/Solar-Open-100B
- technical report: https://huggingface.co/papers/2601.07022

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


vasqu:
@oesni you can ping me when you think it's ready for review (assuming it's not yet because it's a draft)

oesni:
It's ready for review! @vasqu 
But after rebase, it seems some tests fails. 
I'll work on it.

oesni:
wonder if it's okay to add `SolarOpenConfig` class to `OBJECTS_TO_IGNORE` in `utils/check_docstrings.py` since auto-generated config class fails docstring check.
There is comment for `OBJECTS_TO_IGNORE`: `Do not add anything here ...`

docs/source/en/model_doc/solar_open.md
```diff
@@ -0,0 +1,66 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+*This model was released on 2025-12-31 and added to Hugging Face Transformers on 2026-01-13.*
```
vasqu: Just as reminder to keep track of this when we merge

docs/source/en/model_doc/solar_open.md
```diff
@@ -0,0 +1,66 @@
...
+
+We introduce Solar Open, a 102B-parameter bilingual Mixture-of-Experts language model for underserved languages.
+Solar Open demonstrates a systematic methodology for building competitive LLMs by addressing three interconnected challenges.
+First, to train effectively despite data scarcity for underserved languages, we synthesize 4.5T tokens of high-quality, domain-specific, and RL-oriented
+data.
+Second, we coordinate this data through a progressive curriculum jointly optimizing composition, quality thresholds, and domain coverage across 20 trillion tokens.
+Third, to enable reasoning capabilities through scalable RL, we apply our proposed framework SnapPO for efficient optimization.
+Across benchmarks in English and Korean, Solar Open achieves competitive performance, demonstrating the effectiveness of this methodology for underserved language AI development.
+
+Tips:
```
vasqu:
```suggestion
## Usage Tips
```
Might be worth to highlight a bit more

docs/source/en/model_doc/solar_open.md
```diff
@@ -0,0 +1,66 @@
...
+
+Recommended inference parameters for optimal performance:
+
+```
+temperature=0.8
+top_p=0.95
+top_k=50
+```
```
vasqu, line 50: Let's add a small inference script here instead! Something along https://github.com/huggingface/transformers/blob/0a5d574dd0d0dadfbbc981a572bd585284c2ed59/docs/source/en/model_doc/minimax_m2.md?plain=1#L32-L56

We can also add the recommended parameters there or similarly

src/transformers/conversion_mapping.py
```diff
@@ -231,6 +231,7 @@ def _build_checkpoint_conversion_mapping():
     mapping["glm4_moe_lite"] = mapping["qwen2_moe"].copy()
     mapping["glm4v_moe"] = mapping["qwen2_moe"].copy()
     mapping["longcat_flash"] = mapping["qwen2_moe"].copy()
+    mapping["solar_open"] = mapping["qwen2_moe"].copy()
```
vasqu, line 249: Nice!

src/transformers/models/auto/configuration_auto.py
vasqu, line 1: I assume we use `TokenizersBackend` so we don't need anything in `tokenization_auto` but asking just in case

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,238 @@
+# Copyright 2025 Upstage and HuggingFace Inc. team.
```
vasqu:
```suggestion
# Copyright 2026 Upstage and HuggingFace Inc. team.
```
Happy new year :D (probably elsewhere as well)

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,238 @@
...
+        tie_word_embeddings (`bool`, *optional*, defaults to `False`):
+            Whether the model's input and output word embeddings should be tied.
+        rope_parameters (`RopeParameters`, *optional*):
+            Dictionary containing the configuration parameters for the RoPE embeddings. The dictionary should contain
+            a value for `rope_theta` and optionally parameters used for scaling in case you want to use RoPE
+            with longer `max_position_embeddings`.
+        attention_bias (`bool`, *optional*, defaults to `False`):
+            Whether to use a bias in the projection layers.
+        attention_dropout (`float`, *optional*, defaults to 0.0):
+            The dropout ratio for the attention probabilities.
```
vasqu, line 83: Are there models that use it differently? We can IMO remove these if not

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,238 @@
...
+        routed_scaling_factor (`float`, *optional*, defaults to 1.0):
+            Scaling factor for routed experts.
+        n_group (`int`, *optional*, defaults to 1):
+            Number of groups for routed experts.
+        topk_group (`int`, *optional*, defaults to 1):
+            Number of selected groups for each token.
+        norm_topk_prob (`bool`, *optional*, defaults to `True`):
+            Whether to normalize the topk probabilities.
+        use_qk_norm (`bool`, *optional*, defaults to `False`):
+            Whether to use query-key normalization in the attention.
```
vasqu: Same here

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,238 @@
...
+    Glm4MoePreTrainedModel,
+    Glm4MoeRMSNorm,
+)
+from ..llama.modeling_llama import LlamaDecoderLayer
+
+
+logger = logging.get_logger(__name__)
+
+
+class SolarOpenConfig(PretrainedConfig):
```
vasqu: It seems fairly similar to deepseek v3, any chance we can inherit here and simplify a bit

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,238 @@
...
+        rms_norm_eps: float = 1e-5,
+        use_cache: bool = True,
+        tie_word_embeddings: bool = False,
+        rope_parameters: RopeParameters | None = {
+            "rope_theta": 1000000.0,
+            "rope_type": "yarn",
+            "partial_rotary_factor": 1.0,
+            "factor": 2.0,
+            "original_max_position_embeddings": 65536,
+        },
```
vasqu: Mutable kwargs are painful, let's default to None and create the default values if it is None indeed

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,238 @@
...
+        n_routed_experts: int = 128,
+        head_dim: int = 128,
+        hidden_act: str = "silu",
+        max_position_embeddings: int = 131072,
+        initializer_range: float = 0.02,
+        rms_norm_eps: float = 1e-5,
+        use_cache: bool = True,
+        tie_word_embeddings: bool = False,
+        rope_parameters: RopeParameters | None = {
+            "rope_theta": 1000000.0,
```
vasqu: We have `default_theta` to simplify

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,238 @@
...
+            "rope_type": "yarn",
+            "partial_rotary_factor": 1.0,
```
vasqu: Can be left out? I.e. this is not used with `1.0` (and we default to 1.0)

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,238 @@
...
+        if "rope_scaling" in kwargs:
+            self.rope_parameters = kwargs.pop("rope_scaling")
+        if self.rope_parameters is not None:
+            if "rope_type" not in self.rope_parameters and "type" in self.rope_parameters:
+                self.rope_parameters["rope_type"] = self.rope_parameters["type"]
+            if "rope_theta" not in self.rope_parameters and "rope_theta" in kwargs:
+                self.rope_parameters["rope_theta"] = kwargs.pop("rope_theta")
+            if "partial_rotary_factor" not in self.rope_parameters and "partial_rotary_factor" in kwargs:
+                self.rope_parameters["partial_rotary_factor"] = kwargs.pop("partial_rotary_factor")
+        self.validate_rope()
```
vasqu: We should be able to leave this out, it's handled internally ourselves. If you encounter issues, i.e. because values don't default like you want to, keys should be ignored but are warned, you can do something along this https://github.com/huggingface/transformers/blob/0a5d574dd0d0dadfbbc981a572bd585284c2ed59/src/transformers/models/minimax_m2/configuration_minimax_m2.py#L189
overwriting how we read the rope parameters

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+class SolarOpenDecoderLayer(LlamaDecoderLayer):
+    def __init__(self, config: SolarOpenConfig, layer_idx: int):
+        nn.Module.__init__(self)
+        self.hidden_size = config.hidden_size
+        self.self_attn = SolarOpenAttention(config=config, layer_idx=layer_idx)
+        self.mlp = SolarOpenMoE(config)
+        self.input_layernorm = SolarOpenRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        self.post_attention_layernorm = SolarOpenRMSNorm(config.hidden_size, eps=config.rms_norm_eps)
```
vasqu:
```suggestion
    pass
```
Am I missing something, it looks identical to me (except the few extra empty lines)

vasqu: Ah I think I'm seeing the issue, it's the moe? You can simply do
```python
super().__init__()
self.mlp = SolarOpenMoE(config)
```
That will inherit everything that is similar and redefining some attribute will overwrite it (if you use `del self.mlp` you could also delete unnecessary parts)

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,238 @@
...
+class SolarOpenRMSNorm(Glm4MoeRMSNorm):
+    pass
+
+
+class SolarOpenPreTrainedModel(Glm4MoePreTrainedModel):
+    pass
+
+
+class SolarOpenModel(Glm4MoeModel):
+    _keys_to_ignore_on_load_unexpected = []
```
vasqu, line 234: Would be nicer in `SolarOpenPreTrainedModel` but more of a nit

tests/models/solar_open/test_modeling_solar_open.py
```diff
@@ -0,0 +1,94 @@
...
+        self.routed_scaling_factor = routed_scaling_factor
+        self.norm_topk_prob = norm_topk_prob
+        self.use_qk_norm = use_qk_norm
+
+
+@require_torch
+class SolarOpenModelTest(CausalLMModelTest, unittest.TestCase):
+    model_tester_class = SolarOpenModelTester
+    # used in `test_torch_compile_for_training`. Skip as "Dynamic control flow in MoE"
+    _torch_compile_train_cls = None
```
vasqu: Can be maybe removed, we allow compile as of now, see #42697 (let me know if not)

tests/models/solar_open/test_modeling_solar_open.py
```diff
@@ -0,0 +1,94 @@
...
+        model = SolarOpenForCausalLM.from_pretrained(model_id, device_map="auto", torch_dtype=torch.bfloat16)
+        inputs = tokenizer(prompts, return_tensors="pt", padding=True).to(model.device)
+
+        generated_ids = model.generate(**inputs, max_new_tokens=20, do_sample=False)
+        generated_texts = tokenizer.batch_decode(generated_ids, skip_special_tokens=False)
+
+        for text in generated_texts:
+            # Check that the generated continuation starts with <|think|>
+            generated_part = text.split("<|begin|>assistant")[-1]
+            self.assertTrue(generated_part.startswith("<|think|>"))
```
vasqu: The 100B model is too big for our CIs, so it would be nice if you could create a dummy small model that we can run (24GB Vram as we have A10 GPUs). 

I assume it was sanity checked to work as intended on your side?

vasqu (commented):
Looks already super good, my main points are mostly related to making the config more aligned with the current way we handle rope + tests to add a small dummy model for us - 100B is sadly too heavy for our CI :cry:

vasqu:
> wonder if it's okay to add SolarOpenConfig class to OBJECTS_TO_IGNORE in utils/check_docstrings.py since auto-generated config class fails docstring check.
There is comment for OBJECTS_TO_IGNORE: Do not add anything here ...

Just checked why it failed, we should not add it there. You can run `make fix-repo` and you will see that it complains because the config has a wrong default for `rope_parameters` but since we will change it (we should default to None --> non-mutable args) either way let's wait. We can take a look afterwards or I just quickly fix that no worries

docs/source/en/model_doc/solar_open.md
oesni: f80d1d5df31aaf9ce4681b220a9ed160355a36d6
Added example code (which is from our model card example)

src/transformers/models/auto/configuration_auto.py
oesni: 038b1c1f691df3c5b4ac38d8ee5be6c07dcbb67f

src/transformers/models/solar_open/modular_solar_open.py
oesni: Thanks! Happy new year XD
5a8f8621d7876cc2c5b91616f512b1063e645082

src/transformers/models/solar_open/modular_solar_open.py
oesni: It is more like glm4_moe, so I have inherited `Glm4MoeConfig`.
However, it's kinda tricky since `Glm4MoeConfig` has a line with `kwargs.setdefault` for `partial_rotary_factor`
(which seems impossible to remove or override).
https://github.com/huggingface/transformers/blob/a85d6b2a1ecae042cba82663080d4c9b42fceb72/src/transformers/models/glm4_moe/modular_glm4_moe.py#L193

src/transformers/models/solar_open/modular_solar_open.py
oesni: e0487c356b7441badf19f1908afe8fcc03177760

src/transformers/models/solar_open/modular_solar_open.py
oesni: e0487c356b7441badf19f1908afe8fcc03177760

src/transformers/models/solar_open/modular_solar_open.py
oesni: 25d84b9e56324c67979b98f48b8b7ccb6bc7a07d

src/transformers/models/solar_open/modular_solar_open.py
oesni: I intended to override `_keys_to_ignore_on_load_unexpected` from `Glm4MoeModel`, so I'll leave it here :)

tests/models/solar_open/test_modeling_solar_open.py
oesni: f09cac58859f524fa52e76296d98e4514ee488ff
checked test passes without `_torch_compile_train_cls = None`

tests/models/solar_open/test_modeling_solar_open.py
oesni: Yes, I have tested model implementations with the cases model always starts with `<|think|>` token. (on 8*H100)
I'll try to write test with small dummy model.

oesni: b194b851b39f5ac180d172572131b05e618e1146

docs/source/en/model_doc/solar_open.md
```diff
@@ -0,0 +1,108 @@
...
+MODEL_ID = "upstage/Solar-Open-100B"
+
+# Load model and tokenizer
+tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
+
+model = AutoModelForCausalLM.from_pretrained(
+    pretrained_model_name_or_path=MODEL_ID,
+    torch_dtype=torch.bfloat16,
+    device_map="auto",
+    trust_remote_code=True,
```
vasqu:
```suggestion
```
When we merge we hopefully don't need the remote code anymore

src/transformers/models/auto/configuration_auto.py
vasqu: Does it also work if we do not add it to the auto-mapping? We should auto fallback to `TokenizersBackend` but if not let's keep it

src/transformers/models/solar_open/configuration_solar_open.py
```diff
@@ -0,0 +1,197 @@
...
+        attention_dropout: float = 0.0,
+        num_experts_per_tok: int = 8,
+        routed_scaling_factor: float = 1.0,
+        n_group: int = 1,
+        topk_group: int = 1,
+        norm_topk_prob: bool = True,
+        use_qk_norm: bool = False,
+        **kwargs,
+    ):
+        kwargs.setdefault("partial_rotary_factor", 1.0)  # override default from Glm4MoeConfig
```
vasqu: I don't think it worked as expected :cry: see below it's set again

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,235 @@
...
+        attention_dropout: float = 0.0,
+        num_experts_per_tok: int = 8,
+        routed_scaling_factor: float = 1.0,
+        n_group: int = 1,
+        topk_group: int = 1,
+        norm_topk_prob: bool = True,
+        use_qk_norm: bool = False,
+        **kwargs,
+    ):
+        kwargs.setdefault("partial_rotary_factor", 1.0)  # override default from Glm4MoeConfig
```
vasqu: Can we set it after the super? Otherwise, we still can make adjustments in `convert_rope_params_to_dict` instead, wdyt? Just needs to comments properly what happens

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,235 @@
...
+class SolarOpenMoE(Glm4MoeMoE):
+    pass
+
+
+class SolarOpenAttention(Glm4MoeAttention):
+    def __init__(self, config: SolarOpenConfig, layer_idx: int | None = None):
+        super().__init__(config, layer_idx)
+        self.head_dim = config.head_dim
+
+    pass
```
vasqu:
```suggestion
```

src/transformers/models/solar_open/modular_solar_open.py
vasqu: Gotcha, no worries

tests/models/solar_open/test_configuration_solar_open.py
vasqu, line 1: We shouldn't need this file, we automatically test configs alongside in the `test_modeling_xx.py`

tests/models/solar_open/test_modeling_solar_open.py
```diff
@@ -0,0 +1,102 @@
...
+@slow
+class SolarOpenIntegrationTest(unittest.TestCase):
+    def tearDown(self):
+        cleanup(torch_device, gc_collect=False)
+
+    def test_batch_generation_dummy_bf16(self):
+        torch.manual_seed(42)
+        torch.cuda.manual_seed_all(42)
+        torch.backends.cudnn.deterministic = True
+        torch.use_deterministic_algorithms(True, warn_only=True)
```
vasqu: I don't think we need to seed this much if we don't sample or have you found it flaky - more of a nit

tests/models/solar_open/test_modeling_solar_open.py
```diff
@@ -0,0 +1,102 @@
...
+            "Lorem ipsum dolor sit amet=√=√ 치수=√ 치수 치수 치수 치수 치수 Shelley Shelley Shelley Shelley Shelley Shelley Shelley Shelley Shelley площа площа",
+        ]
+
+        tokenizer = AutoTokenizer.from_pretrained(model_id)
+        model = SolarOpenForCausalLM.from_pretrained(model_id, device_map=torch_device, torch_dtype=torch.bfloat16)
+        inputs = tokenizer(prompts, return_tensors="pt", padding=True).to(model.device)
+
+        generated_ids = model.generate(**inputs, max_new_tokens=20, do_sample=False)
+        generated_texts = tokenizer.batch_decode(generated_ids, skip_special_tokens=False)
+        print(generated_texts)
```
vasqu:
```suggestion
```

tests/models/solar_open/test_modeling_solar_open.py
```diff
@@ -0,0 +1,102 @@
...
+
+@require_torch
+class SolarOpenModelTest(CausalLMModelTest, unittest.TestCase):
+    model_tester_class = SolarOpenModelTester
+    model_split_percents = [0.5, 0.85, 0.9]  # it tries to offload everything with the default value
+
+
+@require_torch_accelerator
+@slow
+class SolarOpenIntegrationTest(unittest.TestCase):
```
vasqu, line 88: I will update the values on our CI (A10), thanks for the model :pray:

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -0,0 +1,235 @@
...
+logger = logging.get_logger(__name__)
+
+
+class SolarOpenConfig(Glm4MoeConfig):
+    r"""
+    This is the configuration class to store the configuration of a [`SolarOpenModel`]. It is used to instantiate a
+    SolarOpen model according to the specified arguments, defining the model architecture.
+
+    Configuration objects inherit from [`Glm4MoeConfig`] and can be used to control the model outputs. Read the
+    documentation from [`Glm4MoeConfig`] for more information.
```
vasqu: Should be `PreTrainedConfig` iirc, can you check with llama config for example.

vasqu (commented):
Last few nits, updating the expected output on our CI 

Should be good after that :saluting_face:

tests/models/solar_open/test_modeling_solar_open.py
vasqu: [0fda12a](https://github.com/huggingface/transformers/pull/43244/commits/0fda12a0625ac61c83035189d4bdb846f1f7616e)

tests/models/solar_open/test_modeling_solar_open.py
vasqu: Fyi, moe implementations by default now use `torch._grouped_mm` and `torch.bmm` for more speed (however they are more prone to fluctuate; except you seed like you did here) - I switched to our eager implementation instead as it's way more stable but less efficient (which matters less in the context of slow tests)

docs/source/en/model_doc/solar_open.md
vasqu: It's now enforced on our CI, will need `make fix-repo` but that happens automatically then

vasqu:
run-slow: solar_open

src/transformers/models/auto/configuration_auto.py
oesni: ahh It worked without adding to auto mapping. Would it be better to remove it then?

src/transformers/models/solar_open/configuration_solar_open.py
oesni: afaik, `setdefault` does not override value if it's set already.
So, It actually works as I intended (code is a bit messy though) and tested in [test_configuration_solar_open.py](https://github.com/huggingface/transformers/pull/43244/files/b194b851b39f5ac180d172572131b05e618e1146#diff-52e71302f79b08f704ddf9e998a80ba091f58cbb771e4529effb01b00f6cd7b0)

src/transformers/models/solar_open/modular_solar_open.py
oesni: I tried setting it after super, but it doesn't fix the problem. In the converted `SolarOpenConfig` class, `setdefault` is called after `super().__init__()`, which invokes `convert_rope_params_to_dict`. At that point, `convert_rope_params_to_dict` sets `rope_parameters["partial_rotary_factor"]` to `0.5` (since that's the default value). 
Also, it seems difficult to handle this in `convert_rope_params_to_dict` since the function can't tell whether `partial_rotary_factor` was set explicitly or is just the default (when the value is `0.5`). 😢
(Just in case it's unclear: I want `partial_rotary_factor` to default to `1.0` but remain configurable.)

---
Do you have any ideas on a better way to handle this?

tests/models/solar_open/test_configuration_solar_open.py
oesni: Does the auto-test check default values as well? I wrote test because I found cases where default values weren't set as intended (related to the rope params issues discussed above).

src/transformers/models/solar_open/modular_solar_open.py
oesni: 107899945e497a7339083563925c7efbe2f16845

src/transformers/models/auto/configuration_auto.py
vasqu: Yes please, it would be indeed better that way (i.e. to remove)

src/transformers/models/solar_open/configuration_solar_open.py
vasqu: Gotcha, then let's specify this with comments a bit more, i.e. default values if already set won't be overriden

src/transformers/models/solar_open/modular_solar_open.py
vasqu: I think your initial solution is fine, it's a bit awkward but alright - I'd just prefer to define/elaborate the comment a bit more then :hugs: pointing to it not being overridden, maybe even mark the reason due to modular inheritance issue

tests/models/solar_open/test_configuration_solar_open.py
vasqu: I'd prefer to create a singular extra test then instead of a whole new file (I'm not even sure if we would fetch this test in our CI). The core is to check the default value

vasqu (approved):
Looks overall good, just a few last changes needed, i.e. the config test to a singular extra test (instead of a file) + defining the comment a bit more + sync with main and run `make fix-repo` x `make style` for our CI

src/transformers/models/solar_open/modular_solar_open.py
oesni: [33771e5](https://github.com/huggingface/transformers/pull/43244/commits/33771e54fb76c1b14f9468688da7ffbeec6c304c)

tests/models/solar_open/test_configuration_solar_open.py
oesni: [2563541](https://github.com/huggingface/transformers/pull/43244/commits/25635415f5966da5b25214c3ccf6685daaa1701b)

src/transformers/models/auto/configuration_auto.py
oesni: c118e871abcb5b3efc42f43ffcbbdfdf38bbc8bd

src/transformers/models/solar_open/modular_solar_open.py
oesni: No models currently use these configs, but can we keep them? 
I prefer models to be more configurable for potential future use.

oesni:
All comments have been addressed, and I confirmed that the failing tests are related to a different models. 🤗 
(`Idefics2ProcessorTest` and `GlmImageModelTest`)
@vasqu

src/transformers/models/solar_open/modular_solar_open.py
vasqu: We usually only support extra branches / features when they are actually used within a model

tests/models/solar_open/test_modeling_solar_open.py
```diff
@@ -0,0 +1,193 @@
...
+        inputs = tokenizer(prompts, return_tensors="pt", padding=True).to(model.device)
+
+        generated_ids = model.generate(**inputs, max_new_tokens=20, do_sample=False)
+        generated_texts = tokenizer.batch_decode(generated_ids, skip_special_tokens=False)
+
+        for text, expected_text in zip(generated_texts, EXPECTED_DECODED_TEXT):
+            self.assertEqual(text, expected_text)
+
+
+class SolarOpenConfigTest(unittest.TestCase):
```
vasqu: Ah I meant to move this under the causal lm tester, e.g. gpt2 https://github.com/huggingface/transformers/blob/314f10929a2215b74c2ad6ecf7b2f380c9b7468a/tests/models/gpt2/test_modeling_gpt2.py#L199

tests/models/solar_open/test_modeling_solar_open.py
```diff
@@ -0,0 +1,193 @@
...
+        self.assertEqual(config.max_position_embeddings, 131072)
+        self.assertEqual(config.partial_rotary_factor, 1.0)
+        self.assertEqual(config.rope_parameters["rope_theta"], 1_000_000)
+        self.assertEqual(config.rope_parameters["rope_type"], "yarn")
+        self.assertEqual(config.rope_parameters["partial_rotary_factor"], 1.0)
+        self.assertEqual(config.rope_parameters["factor"], 2.0)
+        self.assertEqual(config.rope_parameters["original_max_position_embeddings"], 65_536)
+        self.assertEqual(config.tie_word_embeddings, False)
+
+    def test_rope_parameters_partially_initialized(self):
```
vasqu: Imo, this is the only important one as it matches to check the partial rotary factor which is a tad different than elsewhere

vasqu (approved):
Super small nits, let's move the test(s) under causal lm tester, mostly the one test to check if partial rotary factor has the correct default

vasqu:
Yup, dont worry about the CI - it's been a bit flaky these past few days/weeks

tests/models/solar_open/test_modeling_solar_open.py
oesni: Gotcha!
[583a8bf](https://github.com/huggingface/transformers/pull/43244/commits/583a8bf84ffeeef606d3da73d8315119c9e090d5)

src/transformers/models/solar_open/modular_solar_open.py
oesni: Gotcha. I'll follow the policy 🫡
9023688db77be07f11f99ea7578bca8c6e7258e7

vasqu: Be careful, looks like due to inheritance we now have a few things to potentially adjust, e.g. the attention is probably now completely aligned with llama

oesni: Confirmed that `Glm4MoeAttention` without `use_qk_norm` is basically just `LlamaAttention`.
Can we inherit `LlamaAttention` and  keep `attention_bias` and `attention_dropout` since both parameters seem quite common and removing those parameters requires modifying parent implementation.

vasqu: Yea should be fine!

oesni: ad2d31040f97316ea39b7dc1af21032976f6ec17
(found little difference in `o_proj` 😃 )

vasqu:
run-slow: solar_open

vasqu:
run-slow: solar_open

vasqu:
Merging now :hugs: thanks for the contribution

oesni:
Thanks for the review! 🤗  @vasqu

PR #43254, opened 2026-01-13
Add supported kwargs to fixed_cross_entropy

### LINKED ISSUE #43240: kwargs are not passed to loss calculation function.

Rocketknight1:
This PR allows some specific kwargs to be passed through to `fixed_cross_entropy`. Fixes #43240

cc @arthurzucker because I think you wrote the original?

### DIFF
--- a/src/transformers/loss/loss_utils.py
+++ b/src/transformers/loss/loss_utils.py
@@ -35,7 +35,9 @@ def fixed_cross_entropy(
     **kwargs,
 ) -> torch.Tensor:
     reduction = "sum" if num_items_in_batch is not None else "mean"
-    loss = nn.functional.cross_entropy(source, target, ignore_index=ignore_index, weight=weight, reduction=reduction, label_smoothing=label_smoothing)
+    loss = nn.functional.cross_entropy(
+        source, target, ignore_index=ignore_index, weight=weight, reduction=reduction, label_smoothing=label_smoothing
+    )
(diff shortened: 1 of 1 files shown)


stas00:
After this change I'd then drop `**kwargs` arg, since it silently ignores them. This is not good for the user - it either should do something with kwargs and assert if the use is unexpected or it shouldn't absorb args that shouldn't be there.

There is another version of this PR here https://github.com/huggingface/transformers/pull/43251

Rocketknight1:
Hey @stas00, good to hear from you again! I suspect the `**kwargs` there is because the function is often called with `**kwargs` in other contexts, so doing it like that lets it catch any named kwargs that match while ignoring ones that don't. That makes me a little nervous about updating it, because I think those kwargs are flowing through from a lot of different possible sources.

stas00:
I understand your reasoning, Matt

Perhaps with v5 transformers could fix the silent `kwargs` black holes in its code - `kwargs` is designed for passing forward arguments to the next function in the stack. In this situation the function isn't passing `kwargs` forward therefore it shouldn't have `kwargs` in its API.

PR #43256, opened 2026-01-13
Delete Text2TextPipeline, SummarizationPipeline and TranslationPipeline

Rocketknight1:
`Text2TextPipeline` has been deprecated for a while, as well as its sub-pipelines `TranslationPipeline` and `SummarizationPipeline`. It was a big maintenance headache, and users should probably use modern chat models for tasks like translation and summarization. This PR finally deletes it all for `v5`!


Rocketknight1:
Tests are passing (one unrelated one failing, I'll make a PR for that). Even though they're no longer SOTA, there is probably still some usage of Summarization/Translation pipelines though, especially the Marian models. Those models will still be usable, but there won't be a clean pipeline for it! We might want to write something about the change in the v5 release notes - I can do that if you're okay with the change cc @lysandrejik @arthurzucker ?

Cyrilvallez (approved):
Very nice, thanks a lot! Indeed, you should definitely add a note in `MIGRATION_GUIDE_V5.md` to notify users, and provide the new alternative for those pipelines!

PR #43273, opened 2026-01-14
fix(processing): use feature_extractor.sampling_rate as default in apply_chat_template

### LINKED ISSUE #43262: Audio processors: `apply_chat_template()` defaults to 16kHz sampling rate, even if the processor config sets a different value
Firstly, thanks for a fantastic library!

### System Info

- `transformers` version: 4.57.5
- Platform: macOS-26.2-arm64-arm-64bit-Mach-O
- Python version: 3.14.2
- Safetensors version: 0.7.0
- Accelerate version: not installed
- PyTorch version (accelerator?): 2.9.1 (NA)
- Jax version: not installed

### Who can help?

Assuming processors fall under tokenisers: @ArthurZucker and @itazap

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

When using a subclass of `transformers.processing_utils.ProcessorMixin` to load an audio file with `apply_chat_template()`, if no sampling rate is explicitly specified, a default of 16kHz is used, even if the processor config calls for a different sampling rate.

Using `CsmProcessor` as an example:
```python
# pip install transformers torch scipy librosa

import torch
import scipy.io.wavfile

# Create example audio
audio = torch.randn(24000)  # 1s of audio at 24kHz
scipy.io.wavfile.write(filename="./example.wav", rate=24000, data=audio.numpy())

# Setup processor
chat_messages = [
    dict(
        role="0",
        content=[
            {"type": "text", "text": "Example"},
            {"type": "audio", "path": "./example.wav"},
        ],
    )
]
processor = AutoProcessor.from_pretrained("sesame/csm-1b")
print(f"{processor.feature_extractor.sampling_rate=}")  # 24000
print()

# Load audio
batch = processor.apply_chat_template(
    chat_messages,
    return_dict=True,
    return_tensors="pt",
    tokenize=True,
)
print("batch['input_values'] should contain 24000 elements")
print(f"{batch['input_values'].shape=}")  # 16000 - bug
print()

# Load audio again, specifying correct rate
batch = processor.apply_chat_template(
    chat_messages,
    return_dict=True,
    return_tensors="pt",
    tokenize=True,
    sampling_rate=processor.feature_extractor.sampling_rate,
)
print("batch['input_values'] should contain 24000 elements")
print(f"{batch['input_values'].shape=}")  # 24000 - correct
```

The default is specified in `transformers.processing_utils.ChatTemplateLoadKwargs`, and is used whenever the `sampling_rate` parameter is not provided in the call to `apply_chat_template()`.

### Expected behavior

Ideally, the default sampling rate should be that of the processor subclass' config.

I'm not sure how would be neatest to implement this - possibly by overriding `apply_chat_template()` in the processor subclasses to inject a custom default value?

theonlypal:
## Summary

- Use processor's `feature_extractor.sampling_rate` as default when loading audio in `apply_chat_template()`
- Fall back to 16kHz only when no feature_extractor is available

## Problem

`apply_chat_template()` hardcoded 16kHz as the default sampling rate, ignoring the processor's configured value. This caused audio to be incorrectly resampled (e.g., 24kHz audio resampled to 16kHz for CsmProcessor).

## Fix

1. Changed `ChatTemplateLoadKwargs.sampling_rate` default from `16_000` to `None`
2. Added runtime resolution using `self.feature_extractor.sampling_rate` when available

Fixes #43262

### DIFF
--- a/src/transformers/processing_utils.py
+++ b/src/transformers/processing_utils.py
@@ -514,12 +514,15 @@ class ChatTemplateLoadKwargs(TypedDict, total=False):
 
     num_frames (`int`, *optional*):
         Number of frames to sample uniformly. If not passed, the whole video is loaded.
+    sampling_rate (`int`, *optional*):
+        The sampling rate at which the audio should be loaded. If not provided, defaults to the
+        processor's feature_extractor.sampling_rate, or 16000 if not available.
     load_audio_from_video (`bool`, *optional*):
             Whether to use the audio track of input video. If `True` the audio track will be loaded and passed to the
             processor. This flag has no effect if the model doesn't support audio modality.
     """
 
-    sampling_rate: int | None = 16_000
+    sampling_rate: int | None = None
     load_audio_from_video: bool | None = False
 
 
@@ -1737,6 +1740,13 @@ class ProcessorMixin(PushToHubMixin):
         tokenize = processed_kwargs["template_kwargs"].pop("tokenize", False)
         return_dict = processed_kwargs["template_kwargs"].pop("return_dict", True)
         mm_load_kwargs = processed_kwargs["mm_load_kwargs"]
+
+        # Use the processor's feature_extractor sampling rate as default if not explicitly provided
+        if mm_load_kwargs.get("sampling_rate") is None:
+            if hasattr(self, "feature_extractor") and hasattr(self.feature_extractor, "sampling_rate"):
+                mm_load_kwargs["sampling_rate"] = self.feature_extractor.sampling_rate
+            else:
+                mm_load_kwargs["sampling_rate"] = 16_000
 
         if tokenize:
             batch_images, batch_videos = [], []


src/transformers/processing_utils.py
```diff
@@ -514,12 +514,15 @@ class ChatTemplateLoadKwargs(TypedDict, total=False):
...
+    sampling_rate (`int`, *optional*):
+        The sampling rate at which the audio should be loaded. If not provided, defaults to the
+        processor's feature_extractor.sampling_rate, or 16000 if not available.
     load_audio_from_video (`bool`, *optional*):
             Whether to use the audio track of input video. If `True` the audio track will be loaded and passed to the
             processor. This flag has no effect if the model doesn't support audio modality.
     """
 
-    sampling_rate: int | None = 16_000
+    sampling_rate: int | None = None
```
zucchini-nlp, line 525: I see that the `AudioKwargs` already have a `sampling_rate`. In that case, this one is redundant and can be deleted imo

src/transformers/processing_utils.py
```diff
@@ -1738,6 +1741,13 @@ def apply_chat_template(
         return_dict = processed_kwargs["template_kwargs"].pop("return_dict", True)
         mm_load_kwargs = processed_kwargs["mm_load_kwargs"]
 
+        # Use the processor's feature_extractor sampling rate as default if not explicitly provided
+        if mm_load_kwargs.get("sampling_rate") is None:
+            if hasattr(self, "feature_extractor") and hasattr(self.feature_extractor, "sampling_rate"):
+                mm_load_kwargs["sampling_rate"] = self.feature_extractor.sampling_rate
+            else:
+                mm_load_kwargs["sampling_rate"] = 16_000
```
zucchini-nlp, line 1749: hmm, this would work if we want to fallback to defaults. Though in some models we might have a default set in `ProcessorKwargs` as here. We can't reliably infer the sample rate if the processor default is different from feature extractor default

https://github.com/huggingface/transformers/blob/e8ae373133be1eff2254c5dd71fcd628445cb4a4/src/transformers/models/voxtral/processing_voxtral.py#L47-L54


cc @ebezzam @eustlb do you have any ideas? In videos we had two options: 1) load and then resample in processing or 2) load inside video processing. Ig you guys already have some plans for standardization, so I'll let you decide

src/transformers/processing_utils.py
ebezzam: I think you're right @zucchini-nlp, it can be removed

src/transformers/processing_utils.py
ebezzam: I think this is a nice idea for a fix until @eustlb completes the standardization for the audio processor.

One idea to make @theonlypal's approach slightly simpler (if we also remove `sampling_rate` from `ChatTemplateLoadKwargs` as mentioned above) would be:

```python
if hasattr(self, "feature_extractor") and hasattr(self.feature_extractor, "sampling_rate"):
    sampling_rate = self.feature_extractor.sampling_rate
else:
    sampling_rate = 16_000

...

      # Audio models do not accept nested list of audios (yet!) so we construct a flat input audio list
      if not mm_load_kwargs["load_audio_from_video"]:
          for fname in audio_fnames:
              batch_audios.append(load_audio(fname, sampling_rate=sampling_rate))
      else:
          for fname in video_fnames:
              batch_audios.append(load_audio(fname, sampling_rate=sampling_rate))
```
namely simply creating a `sampling_rate` variable within the `apply_chat_template` method.

--

More broadly, tbh I find "sampling_rate" to also be redundant in `audio_kwargs`, as it can be obtained from `self.feature_extractor.sampling_rate` (in all cases I can think of, but may break for older models so probably good to have a default). 

Moreover, for most processors, the sampling rate is only used for saving audio (e.g. [CSM](https://github.com/huggingface/transformers/blob/0a5d574dd0d0dadfbbc981a572bd585284c2ed59/src/transformers/models/csm/processing_csm.py#L163)), which I think @eustlb is planning to move such methods out of individual processors for having a standardized method? And for such a method, I think it would be fine to default to the feature extractor's sampling rate. And even let the user specify the sampling rate in case they want the audio at a different sampling rate.

PR #43279, opened 2026-01-14
[Qwen-VL] min/max pixels gone forever

zucchini-nlp:
# What does this PR do?

As per title, deletes min/max pixels from class attributes, but still allows passing them with kwargs at `__init__` and during `__call__`. 

And fixes https://github.com/huggingface/transformers/pull/43228#issuecomment-3745649304. Caused by us setting `self.min_pixels` to config value and giving it highest priority than `size` in `_further_process_kwargs`, even when `size` is a user-defined argument.

### DIFF
--- a/tests/models/qwen2_vl/test_video_processing_qwen2_vl.py
+++ b/tests/models/qwen2_vl/test_video_processing_qwen2_vl.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import tempfile
 import unittest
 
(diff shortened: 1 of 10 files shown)


tests/models/qwen2_vl/test_video_processing_qwen2_vl.py
```diff
@@ -58,7 +59,7 @@ def __init__(
         max_pixels=100 * 100,
         merge_size=2,
     ):
-        size = size if size is not None else {"shortest_edge": 20}
+        size = size if size is not None else {"shortest_edge": 400, "longest_edge": 10000}
```
molbap, line 62: 10000?

molbap (approved):
Thanks for the cleanup :) just default values seeming weird in a test

tests/models/qwen2_vl/test_video_processing_qwen2_vl.py
zucchini-nlp: yeah, that is not really a size but `max_pixels`. So each image can be resized as max of 100x100 and min of 20x20

Issue #40827, opened 2025-09-11
[`Cache`] Proper support for Linear Attention (related) caches


vasqu:
Linear Attention covers many flavors that only came up in recent years. The most famous one is Mamba(2) - which was initially often seen/described as (selective) SSM. However, Qwen3 Next includes `Gated Delta Net`, another different flavor.

I expect more and more to follow suite and our current way of handling this, is ignoring everything and rewrite the Cache... Imo less than ideal especially since these caches are static (no dynamic growth or anything), they should be easier to handle than kv caches. 

My suggestion proposal/suggestion would be to introduce a layer type based cache for linear attention similar to swa / dynamic / static. Going for full integration.

cc @ArthurZucker @Cyrilvallez @LysandreJik @gante

vasqu:
Semi related: Also going for a standardization as done for attention implementations 👀

Cyrilvallez:
Indeed, should be extremely straightforward and will simplify quite a bit

gante:
If we can fit the existing `Cache` abstractions (or if we can integrate them with minimal changes to the base abstraction), I think it makes complete sense! It would also be much better for hybrid models.

Same interface for all cache/states = much easier time managing them on model-agostic parts of the codebase. For instance, we have much poorer `generate` support of mamba-like models due to cache interfaces. We would still need some internal attribute to distinguish the two kinds -- e.g., we can't roll back states, some generation methods can be made more efficient with rollbacks.

The trickiest part is perhaps `update`, to find a design where we don't need to add a new abstraction / manual cache manipulation every time a new type of state comes out (`conv_states`, `recurrent_states`, `ssm_states`, ...). Perhaps we could make the logic in `update` dependent on the received kwargs? E.g. in `Qwen3NextDynamicCache`, `cache.update(key_states=..., value_states=...)` would run the logic to update KV, `cache.update(conv_states=...)` would update `conv_states`, ...

Cyrilvallez:
Still on my roadmap, hopefully will find some time soon!

jagmarques:
The Gemma 4 issue filed in #45242 makes this more timely. Hybrid models with mixed attention types (SWA + full attention + linear/SSM layers) are arriving faster than the Cache abstraction can handle.

[...]

PR #43280, opened 2026-01-14
Prepare inputs for generation is called from `super()`

zucchini-nlp:
# What does this PR do?

As per title, less custom code and more standardization. There are still a few models with completely custom `prepare_inputs_for_generation`, I can't make those models happy with `super().prepare_inputs_for_generation` (musicgen, xlm, reformer)


src/transformers/models/bert/modeling_bert.py
```diff
@@ -1013,30 +1014,6 @@ def forward(
...
-
-        attention_mask = torch.cat([attention_mask, attention_mask.new_zeros((attention_mask.shape[0], 1))], dim=-1)
-        dummy_token = torch.full(
-            (effective_batch_size, 1), self.config.pad_token_id, dtype=torch.long, device=input_ids.device
-        )
-        input_ids = torch.cat([input_ids, dummy_token], dim=1)
-
-        return {"input_ids": input_ids, "attention_mask": attention_mask}
-
-    @classmethod
```
zucchini-nlp, line 1032: not even used across the library because the model isn't generative. Should have been deleted in the past

src/transformers/models/bamba/modular_bamba.py
```diff
@@ -1152,55 +1154,24 @@ def prepare_inputs_for_generation(
...
-                or cache_position[-1] >= input_ids.shape[1]  # Exception 3
-            ):
-                input_ids = input_ids[:, -cache_position.shape[0] :]
-            elif input_ids.shape[1] != cache_position.shape[0]:  # Default case (the "else", a no op, is Exception 2)
-                input_ids = input_ids[:, cache_position]
-        else:
+        if past_key_values is None:
             past_key_values = HybridMambaAttentionDynamicCache(
                 self.config, input_ids.shape[0], self.dtype, device=self.device
             )
```
zucchini-nlp, line 1160: special cache classes can't yet be prepared in `generate`, let's keep it as custom code for now. If needed, we can allow these models to have a `self.prepare_model_cache` method in the future and call it from within `generate`

src/transformers/models/rwkv/modeling_rwkv.py
```diff
@@ -683,29 +683,6 @@ def get_output_embeddings(self):
...
-        if state is not None:
-            input_ids = input_ids[:, -1].unsqueeze(-1)
-
-        # if `inputs_embeds` are passed, we only want to use them in the 1st generation step
-        if inputs_embeds is not None and state is None:
-            model_inputs = {"inputs_embeds": inputs_embeds}
-        else:
-            model_inputs = {"input_ids": input_ids}
-
-        model_inputs["state"] = state
```
zucchini-nlp, line 699: any arg which is non-common will be simply fast-forwarded in `model_kwargs` as is, no need to overwrite anything

src/transformers/models/bert/modeling_bert.py
vasqu: It even has `can_generate` returning False :cry: 

Are `encoder_decoder` [over here](https://github.com/huggingface/transformers/tree/main/src/transformers/models/encoder_decoder) still working? Just want to make sure that there is no weird connection as they reuse bert for example as decoder

src/transformers/models/bamba/modular_bamba.py
```diff
@@ -1154,55 +1154,24 @@ def prepare_inputs_for_generation(
...
-        # Exception 3: with synced GPUs cache_position may go out of bounds, but we only want dummy token in that case.
-        #              (we can't check exception 3 while compiling)
-        if not empty_past_kv:
-            if (
-                inputs_embeds is not None  # Exception 1
-                or cache_position[-1] >= input_ids.shape[1]  # Exception 3
-            ):
-                input_ids = input_ids[:, -cache_position.shape[0] :]
-            elif input_ids.shape[1] != cache_position.shape[0]:  # Default case (the "else", a no op, is Exception 2)
-                input_ids = input_ids[:, cache_position]
```
vasqu, line 1170: Just trying to check if I understand things correctly, this will be handled by `_cache_dependant_input_preparation`

src/transformers/models/bamba/modular_bamba.py
vasqu: I think we want to make linear attention caches part of the normal caches (#40827), at that point we should be able to remove it as well

src/transformers/models/bamba/modular_bamba.py
```diff
@@ -1154,55 +1154,24 @@ def prepare_inputs_for_generation(
...
-        model_inputs.update(
-            {
-                "position_ids": position_ids,
-                "past_key_values": past_key_values,
-                "use_cache": use_cache,
-                "attention_mask": attention_mask,
-                "logits_to_keep": self.config.num_logits_to_keep,
-                "cache_position": cache_position,
-            }
+        kwargs["logits_to_keep"] = self.config.num_logits_to_keep
```
vasqu, line 1161: Nit: but why not pass it directly as kwarg?

src/transformers/models/ctrl/modeling_ctrl.py
```diff
@@ -482,28 +482,16 @@ def prepare_inputs_for_generation(
...
-            if input_ids.shape[1] > past_length:
-                remove_prefix_length = past_length
-            else:
-                # Default to old behavior: keep only final ID
-                remove_prefix_length = input_ids.shape[1] - 1
-
-            input_ids = input_ids[:, remove_prefix_length:]
-
-        model_inputs = {"input_ids": input_ids, "past_key_values": past_key_values, "use_cache": use_cache}
+        model_inputs = super().prepare_inputs_for_generation(
```
vasqu, line 485: `# Overwritten -- inputs_embeds not working properly` should be changed above ig? Seems like only the `token_type_ids` are unwanted

src/transformers/models/ernie/modeling_ernie.py
```diff
@@ -1064,30 +1064,6 @@ def forward(
             attentions=outputs.attentions,
         )
 
-    def prepare_inputs_for_generation(self, input_ids, attention_mask=None, **model_kwargs):
```
vasqu, line 1066: Same as bert :cry:

src/transformers/models/kosmos2_5/modeling_kosmos2_5.py
```diff
@@ -1642,20 +1631,21 @@ def prepare_inputs_for_generation(
...
+            past_key_values=past_key_values,
+            attention_mask=attention_mask,
+            use_cache=use_cache,
+            cache_position=cache_position,
+            position_ids=position_ids,
+            is_first_iteration=is_first_iteration,
+            **model_kwargs,
+        )
+        # Kosmos2.5 has offset for position ids, so we need to create them correctly in PositionEmbedding layer
+        model_inputs.pop("position_ids", None)
```
vasqu: Can we not create them here before as well? E.g. a few lines above pos_ids is created when we have a cache

src/transformers/models/mamba/modeling_mamba.py
```diff
@@ -753,40 +753,32 @@ def prepare_inputs_for_generation(
         **kwargs,
     ):
         # Overwritten -- uses `cache_params` as opposed to `past_key_values`
-        model_inputs = {"input_ids": input_ids.contiguous()}
```
vasqu, line 756: Let's update the reason here as well (cache_position / non-standard cache)

src/transformers/models/mamba2/modeling_mamba2.py
```diff
@@ -957,40 +957,33 @@ def prepare_inputs_for_generation(
         **kwargs,
     ):
         # Overwritten -- uses `cache_params` as opposed to `past_key_values`
-        model_inputs = {"input_ids": input_ids.contiguous()}
```
vasqu, line 960: Same here

src/transformers/models/prophetnet/modeling_prophetnet.py
```diff
@@ -1851,27 +1851,17 @@ def prepare_inputs_for_generation(
     ):
         # Overwritten -- our tests complain if we use GenerationMixin.prepare_inputs_for_generation
```
vasqu: Update reason?

src/transformers/models/rwkv/modeling_rwkv.py
vasqu: yup, seeing the `# 7. Forward ALL kwargs that are uninitialized (e.g. `use_cache`).`

tests/models/git/test_modeling_git.py
```diff
@@ -482,6 +482,12 @@ def test_greedy_generate_dict_outputs_use_cache(self):
     def test_forward_with_logits_to_keep(self):
         pass
 
+    @unittest.skip(
+        reason="GIT cannot continue from past-kv due to image tokens merged into cache. Not wroth fixing, rarely used model"
```
vasqu:
```suggestion
        reason="GIT cannot continue from past-kv due to image tokens merged into cache. Not worth fixing, rarely used model"
```

vasqu (commented):
Looks good overall, would just double check with `run-slow` for each kind of category
- Bert (encoder)
- Encoder_Decoder (encoder-decoder models that are created from other models e.g. bert, gpt2)
- Bamba (Hybrid mamba2-attn)
- Falcon Mamba (Hybrid mamba-attn)
- Mamba (pure model)
- Mamba2 (pure model)
- Roformer

Just those I found on a first glance that seemed slightly different from each other

src/transformers/models/bamba/modular_bamba.py
zucchini-nlp: yes correct, it's the same thing

src/transformers/models/bert/modeling_bert.py
zucchini-nlp: it is working. I tracked it down to when Bert was split into two classes, one for generation and one for masked LM. But the contributor didn't remove `prepare_inputs_for_generation`...

src/transformers/models/bamba/modular_bamba.py
zucchini-nlp: it might be already in `kwargs` because `generate` passes this arg. I don't know why these models are defaulting to a config attribute instead of letting `generate` handle it 🙃 

Would love to delete this, even though it's technically "breaking". WDYT?

src/transformers/models/ctrl/modeling_ctrl.py
zucchini-nlp: Oh right, some of the comments are a bit off imo. I wouldn't guess why `inputs_embeds` doesn't work looking at the code

src/transformers/models/ernie/modeling_ernie.py
zucchini-nlp: too many models, because of copying from bert

src/transformers/models/kosmos2_5/modeling_kosmos2_5.py
zucchini-nlp: I copied it from Kosmos2. IIRC using `create_position_ids_from_input_ids` caused another issue and letting the model handle it worked the best. I can re-check

zucchini-nlp:
run_slow: bert,  encoder_decoder, bamba, mamba, mamba2, falcon_h1, falcon_mamba, roformer, git

zucchini-nlp:
run-slow: kosmos2_5

src/transformers/models/kosmos2_5/modeling_kosmos2_5.py
zucchini-nlp: Oke, seems to work because this model specifically doesn't support `inputs_embeds`, bringing it back

src/transformers/models/bamba/modular_bamba.py
vasqu: Imo, it should've been a custom generate arg in the generation config, now we have this mess :cry: It's minor so I feel like it's not worth breaking but trusting on your feeling here, not opposed to deleting it (let's mark it as breaking in the title then tho :red_circle:)

src/transformers/models/falcon_h1/modeling_falcon_h1.py
```diff
@@ -43,7 +43,7 @@
 from ...modeling_rope_utils import ROPE_INIT_FUNCTIONS, dynamic_rope_update
 from ...modeling_utils import ALL_ATTENTION_FUNCTIONS, PreTrainedModel
 from ...processing_utils import Unpack
-from ...utils import TransformersKwargs, auto_docstring, can_return_tuple, is_torchdynamo_compiling, logging
```
vasqu, line 46: Good that this is removed, `is_torchdynamo_compiling` is old and should be deprecated - `is_tracing` should be used instead

Not really related to this PR, keeping notes myself

src/transformers/models/git/modeling_git.py
```diff
@@ -203,6 +203,7 @@ def forward(
         else:
             embeddings = inputs_embeds
 
+        print("position_ids", position_ids)
```
vasqu:
```suggestion
```
oops :D

src/transformers/models/git/modeling_git.py
```diff
@@ -1122,6 +1127,7 @@ def forward(
             attention_mask = torch.cat([extended_attention_mask, attention_mask], dim=-1)
 
         # Images attend each other bidirectionally while text remains causal
+        print(cache_position)
```
vasqu:
```suggestion
```

src/transformers/models/kosmos2_5/modeling_kosmos2_5.py
```diff
@@ -1610,52 +1611,48 @@ def prepare_inputs_for_generation(
...
-            "position_ids": position_ids,
-            "use_cache": use_cache,
-        }
-
-        # Forward ALL kwargs that are uninitialized (e.g. `use_cache`).
-        for key, value in model_kwargs.items():
-            if key not in model_inputs:
-                model_inputs[key] = value
+            # Kosmos2.5 has offset for position ids, so we need to create them correctly in PositionEmbedding layer
+            model_inputs.pop("position_ids", None)
```
vasqu, line 1659: Should this not happen either way if we do not have a cache? I.e. `if past_key_values is not None... else: pop...`

vasqu (approved):
Looks overall good, let's fix the last edge case (kosmos 2.5)

Most models seem to work as expected, the one config thing (num_keep_logits) - not opposed to removing but let's add a :red_circle: or similar to the title just to be sure.

vasqu:
CI failures are unrelated :eyes:

src/transformers/models/kosmos2_5/modeling_kosmos2_5.py
zucchini-nlp: nope, this would simply be prepared from attention mask in all cases by `cumsum`'ing, so we drop it off just in case

zucchini-nlp:
Huh, interesting, the kosmos tests are still fine in dgx. Let me try runners as well

Passes inside the runners, what!? 🫠

zucchini-nlp:
run-slow: kosmos2_5

vasqu:
Super weird, maybe something broke in main temporarily?

zucchini-nlp:
It might have grabbed incorrect commit prob, I pushed two commits at the same time 🙈

PR #43286, opened 2026-01-14
Crash Fix for GLM-Image: TopPLogitsWarper numerical stability on MPS

inferencers:
# What does this PR do?
Using -inf as a filter value in masked_fill causes NaN on MPS devices.

Changing default filter_value to a large negative float -1e9 instead is a commonly used safe alternative for softmax and numerically stable across all devices (CPU, CUDA, MPS).

Fixes # (issue)
Runtime crash when generating image with GLM-Image on MPS.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/transformers/generation/logits_process.py
+++ b/src/transformers/generation/logits_process.py
@@ -500,7 +500,7 @@ class TopPLogitsWarper(LogitsProcessor):
     ```
     """
 
-    def __init__(self, top_p: float, filter_value: float = -float("Inf"), min_tokens_to_keep: int = 1):
+    def __init__(self, top_p: float, filter_value: float = -1e9, min_tokens_to_keep: int = 1):
         top_p = float(top_p)
         if top_p < 0 or top_p > 1.0:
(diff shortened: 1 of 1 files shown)


src/transformers/generation/logits_process.py
```diff
@@ -500,7 +500,7 @@ class TopPLogitsWarper(LogitsProcessor):
     ```
     """
 
-    def __init__(self, top_p: float, filter_value: float = -float("Inf"), min_tokens_to_keep: int = 1):
+    def __init__(self, top_p: float, filter_value: float = -1e9, min_tokens_to_keep: int = 1):
```
vasqu, line 503: Sorry but we cannot do this as if you were combining multiple logits processors we would encounter weird behavior. Isolated this is fine but not grouped with others.

Unsure how to handle this tbh but it's certainly not good :cry:

inferencers: Agreed

PR #43288, opened 2026-01-14
Fix eager experts for DeepSeekV2

### LINKED ISSUE #43224: DeepSeek-V2-Lite-Chat repetitive generation when `_experts_implementation="eager"`
### System Info

- `transformers` version: 5.0.0.dev0
- Platform: Linux-5.4.0-169-generic-x86_64-with-glibc2.35
- Python version: 3.11.10
- Huggingface_hub version: 1.3.1
- Safetensors version: 0.4.5
- Accelerate version: 1.10.1
- PyTorch version (accelerator?): 2.9.0+cu128 (CUDA)

### Who can help?

text models: @ArthurZucker @Cyrilvallez

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

```python


import torch
import transformers

model_name = "deepseek-ai/DeepSeek-V2-Lite-Chat"

device = torch.device("cuda" if torch.cuda.is_available() else "cpu")

with torch.no_grad(), torch.device(device):
    trust_remote_code = False
    tokenizer = AutoTokenizer.from_pretrained(model_name,trust_remote_code=trust_remote_code)
        model_name,
        torch_dtype="auto",
        trust_remote_code=trust_remote_code,
        _experts_implementation="eager"
        )
    # generate some text
    inputs = tokenizer("Hello, my dog is", return_tensors="pt").to(device)
    outputs = model.generate(**inputs, max_new_tokens=20)
    print(tokenizer.decode(outputs[0], skip_special_tokens=True))
    
    
# Setting `pad_token_id` to `eos_token_id`:100001 for open-end generation.
# Hello, my dog isodododododododododododododododododododod

```

### Expected behavior

The output should be reasonable.

Rocketknight1:
The eager expert computation for DeepSeekV2 inherited from Qwen2MoE, but they pass expert weights differently. Qwen2MoE selects experts before `forward()`, it passes `top_k_weights` with shape  `(num_tokens, num_top_k)`. DeepSeekV2 does not, so `top_k_weights` has shape `(num_tokens, num_experts)`. The result is that DeepSeekV2 selected incorrect experts when `_experts_implementation == "eager"` and the output was garbage.

This PR overrides the `forward()` method for DeepSeek so we select the right experts!

cc @IlyasMoutawwakil, fixes #43224


IlyasMoutawwakil:
thanks @Rocketknight1 i was surprised as well when writing batched_mm and grouped_mm that top_k_weights can have two possible shapes while the eager path seemed very standardized across MoEs. so i added support for both shapes without touching the eager path (assuming that it works fine somehow 😅).

Rocketknight1:
Yeah, I'm not sure how this ever worked! I wonder how long this bug has existed for

IlyasMoutawwakil:
not for long, i see it was introduced in https://github.com/huggingface/transformers/pull/42456

Rocketknight1:
cc @ydshieh more flaky tests here, I think possibly Hub issues again?

Rocketknight1:
run-slow: deepseek_v2

Rocketknight1:
@IlyasMoutawwakil tests are green! Can you approve it so I can merge?

src/transformers/models/deepseek_v2/modular_deepseek_v2.py
```diff
@@ -233,6 +233,33 @@ def __init__(self, config):
         super().__init__(config)
         self.num_experts = config.n_routed_experts
 
+    def forward(
```
vasqu: Sorry I have to say that this is likely the wrong fix and will make things incompatible with things like fp8, we should change the gates instead. 

Essentially, this scatter changes the shapes https://github.com/huggingface/transformers/blob/f92c55356ea6227515a18750ea9e762ccfcacad4/src/transformers/models/deepseek_v2/modular_deepseek_v2.py#L299
We should simply remove that line and everything else should be resolved then (without these changes here)

IlyasMoutawwakil: good catch ! standardizing the router's outputs would definitely make things easier

Rocketknight1: Yes, agreed! I wasn't too familiar with the code and your fix seems like a better one

Rocketknight1: I don't understand how the change breaks FP8 compatibility, though? If you look at the modeling file, it only updates one slice

vasqu: Because the weights are exchanged as is and the forward needs to be overwritten --> it has to expect the same input as the base version in the modeling code. FP8 currently assumes the version where we do not use the scatter

Reference: https://github.com/huggingface/transformers/blob/38e5987e1b9b320d44cb0ece00c4a4e34256964a/src/transformers/integrations/finegrained_fp8.py#L530-L563

IlyasMoutawwakil: so with this deepseekv2 will have its `top_k_weights` with shape `(num_tokens, num_top_k)` ? like mixtral, qwen moe, etc

Rocketknight1:
run-slow: deepseek_v2

src/transformers/models/deepseek_v2/modular_deepseek_v2.py
vasqu: Yup, exactly

Rocketknight1:
@vasqu confirmed in local testing that your fix also recovers the correct output!

vasqu (approved):
Thank you :)

IlyasMoutawwakil (approved):
LGTM ! thanks !

PR #43290, opened 2026-01-14
[`Docs`] Minimax M2 add release date

vasqu:
New make automatically added the correct line but there was no original release date


stevhliu (approved):
thanks!

PR #43301, opened 2026-01-15
Simplify `_get_resolved_checkpoint_files` and do not try safetensors bot conversion with a subfolder

Cyrilvallez:
# What does this PR do?

As per the title. Also takes the opportunity to refactor `_get_resolved_checkpoint_files` a bit, as it's currently a big mess
Also, `use_safetensors` should be set to None in `from_pretrained`, so that we check `.bin` files automatically if there are no `safetensors` file. Otherwise, it requires passing an explicit `use_safetensors=False` from the user which is weird - also related to [this issue](https://github.com/huggingface/transformers/pull/43178#discussion_r2691060130)

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -3629,7 +3629,7 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         local_files_only: bool = False,
         token: str | bool | None = None,
         revision: str = "main",
-        use_safetensors: bool | None = True,
+        use_safetensors: bool | None = None,
         weights_only: bool = True,
         **kwargs,
     ) -> SpecificPreTrainedModelType:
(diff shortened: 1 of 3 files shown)


src/transformers/modeling_utils.py
```diff
@@ -562,25 +564,38 @@ def _get_resolved_checkpoint_files(
...
+                "_commit_hash": commit_hash,
+                **has_file_kwargs,
+            }
+            can_auto_convert = (
+                not is_offline_mode()  # for obvious reasons
+                # If we are in a CI environment or in a pytest run, we prevent the conversion
+                and not is_env_variable_true("DISABLE_SAFETENSORS_CONVERSION")
+                and not is_remote_code  # converter bot does not work on remote code
+                and subfolder == ""  # converter bot does not work on subfolders
+            )
```
LysandreJik, line 590: clean!

LysandreJik (approved):
Thanks @Cyrilvallez for the refactor

PR #43304, opened 2026-01-15
[Quantization] Add cutlass kernel for FP8

MekkCyber:
# What does this PR do?

adds the cutlass kernel for scaled matmul, the performance is much better than triton for the specific block size : (128, 128):
```
================================================================================
COMPARISON: CUTLASS vs Triton Speedup
================================================================================

     M x      N x      K |  Triton (ms) | CUTLASS (ms) |    Speedup
----------------------------------------------------------------------
     1 x   4096 x   4096 |       0.0571 |       0.0317 |       1.80x
     4 x   4096 x   4096 |       0.0571 |       0.0516 |       1.11x
    16 x   4096 x   4096 |       0.0563 |       0.0511 |       1.10x
    32 x   4096 x   4096 |       0.0573 |       0.0509 |       1.13x
    64 x   4096 x   4096 |       0.0581 |       0.0509 |       1.14x
   128 x   4096 x   4096 |       0.0807 |       0.0509 |       1.59x
   256 x   4096 x   4096 |       0.0798 |       0.0511 |       1.56x
   512 x   4096 x   4096 |       0.0826 |       0.0509 |       1.62x
  1024 x   4096 x   4096 |       0.1069 |       0.0510 |       2.10x
  2048 x   4096 x   4096 |       0.2187 |       0.0660 |       3.31x
  4096 x   4096 x   4096 |       0.4310 |       0.1256 |       3.43x
```

All FP8 tests passing !


src/transformers/integrations/finegrained_fp8.py
```diff
@@ -26,6 +26,61 @@
...
+def _get_quantization_kernel():
+    """Lazily load the CUTLASS quantization kernel from HuggingFace Hub."""
+    global _quantization_kernel
+    if _quantization_kernel is None:
+        try:
+            from .hub_kernels import get_kernel
+
+            _quantization_kernel = get_kernel("RedHatAI/quantization")
+        except Exception as e:
+            logger.warning_once(f"Failed to load CUTLASS quantization kernel: {e}. Falling back to Triton.")
```
sayakpaul, line 43: Do we want to also log that we're using the Redhat kernel in case it was successfully loaded?

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -26,6 +26,61 @@
...
+            _quantization_kernel = False  # Mark as unavailable
+    return _quantization_kernel if _quantization_kernel else None
+
+
+def _supports_cutlass_blockwise_fp8(block_size: list[int] | None) -> bool:
+    """
+    Check if CUTLASS blockwise FP8 matmul is supported for the given block size.
+
+    CUTLASS blockwise kernels require:
+    - SM90+ (Hopper or newer)
```
sayakpaul, line 53: This is fine IMO. When hardware is available, users should be able to max them out!

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -26,6 +26,61 @@
...
+        return False
+    if len(block_size) != 2 or block_size[0] != 128 or block_size[1] != 128:
+        return False
+
+    # Check GPU capability (SM90+)
+    try:
+        capability = torch.cuda.get_device_capability()
+        cuda_capability = capability[0] * 10 + capability[1]
+    except Exception:
+        return False
```
SunMarc: no need for try except

src/transformers/integrations/finegrained_fp8.py
SunMarc, line 43: if the kernel is not compatible, it will raise an error ?

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -338,6 +393,84 @@ def grid(META):
...
+    - Output dtype is bfloat16 or float16 (CUTLASS requirement)
+    - Tensor dimensions are compatible (divisible by 16)
+
+    Otherwise falls back to Triton.
+    """
+    # Check if we can use CUTLASS
+    # CUTLASS only supports bfloat16/float16 output
+    cutlass_compatible_dtype = output_dtype in (torch.bfloat16, torch.float16)
+
+    if cutlass_compatible_dtype and _supports_cutlass_blockwise_fp8(block_size):
```
SunMarc: let's put the dtype check in `_supports_cutlass_blockwise_fp8`

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -338,6 +393,84 @@ def grid(META):
...
+                # Bs should be [K//128, N//128] with K-major layout (stride(0)==1)
+
+                # As: reshape to [M, K//128], then make M-major via t().contiguous().t()
+                As_2d = As.view(M, -1).contiguous()
+                As_2d = As_2d.t().contiguous().t()  # [M, K//128] with stride(0)==1
+
+                # Bs: our input is [N//128, K//128], need [K//128, N//128] with stride(0)==1
+                # Transpose to get [K//128, N//128], then make K-major via t().contiguous().t()
+                Bs_km = Bs.contiguous().t()  # [K//128, N//128]
+                Bs_km = Bs_km.t().contiguous().t()  # Make K-major (stride(0)==1)
```
SunMarc, line 459: any idea of the overhead of this ?

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -338,6 +393,84 @@ def grid(META):
...
+    Otherwise falls back to Triton.
+    """
+    # Check if we can use CUTLASS
+    # CUTLASS only supports bfloat16/float16 output
+    cutlass_compatible_dtype = output_dtype in (torch.bfloat16, torch.float16)
+
+    if cutlass_compatible_dtype and _supports_cutlass_blockwise_fp8(block_size):
+        kernel = _get_quantization_kernel()
+        if kernel is not None:
+            try:
```
SunMarc, line 423: even after loading the kernels and checking that everything is supported, there is still a risk of failure ?

src/transformers/integrations/finegrained_fp8.py
```diff
@@ -26,6 +26,61 @@
...
+    try:
+        capability = torch.cuda.get_device_capability()
+        cuda_capability = capability[0] * 10 + capability[1]
+    except Exception:
+        return False
+
+    # Try to load the kernel and check if blockwise FP8 is supported
+    kernel = _get_quantization_kernel()
+    if kernel is None:
+        return False
```
SunMarc, line 79: I think we also need to check that kernels is installed no?

src/transformers/integrations/finegrained_fp8.py
SunMarc, line 43: instead of a single kernel, can we try to create a dict where we store multiple kernels ? We can leave this for a follow-up PR but the idea would be to also move the triton kernels in kernels.

SunMarc (commented):
Thanks ! Just a few nits

src/transformers/integrations/finegrained_fp8.py
MekkCyber: it will just log it and fallback to triton

MekkCyber: Yes we can do that I think

src/transformers/integrations/finegrained_fp8.py
MekkCyber: the overhead is already taken into consideration in the benchmark, the gains outweight the overhead

src/transformers/integrations/finegrained_fp8.py
MekkCyber: as discussed offline, yes if no compatible kernel is found for the torch version and cuda version used

SunMarc (approved):
Thanks ! Let's merge this

SunMarc:
@ArthurZucker if you can merge this !

ArthurZucker (approved):

ArthurZucker:
nice work and perfs!

PR #43307, opened 2026-01-15
[loading] Fix Transpose Operation, and qwen3_vl_moe mapping

Cyrilvallez:
# What does this PR do?

As per the title. Transpose needs to be more general in order to be used in reverse mode in a chain of several operations.
Supersedes https://github.com/huggingface/transformers/pull/43201 as it needed more work on the Transpose Operation!

Will also help https://github.com/huggingface/transformers/pull/43227

### DIFF
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -249,6 +249,7 @@ class Transpose(ConversionOps):
     """
     Transposes the given tensor along dim0 and dim1.
     """
+
     # BIG TODO: This Ops may clash with TP if the required transpose dim are the same as the TP shard dim
     # Would require to check if this Ops is used before sharding with TP
 
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -250,9 +250,6 @@ class Transpose(ConversionOps):
     Transposes the given tensor along dim0 and dim1.
     """
 
-    # BIG TODO: This Ops may clash with TP if the required transpose dim are the same as the TP shard dim
-    # Would require to check if this Ops is used before sharding with TP
-
     def __init__(self, dim0: int = 0, dim1: int = 1):
         self.dim0 = dim0
         self.dim1 = dim1
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -249,6 +249,8 @@ class Transpose(ConversionOps):
     """
     Transposes the given tensor along dim0 and dim1.
     """
+    # BIG TODO: This Ops may clash with TP if the required transpose dim are the same as the TP shard dim
+    # Would require to check if this Ops is used before sharding with TP
 
     def __init__(self, dim0: int = 0, dim1: int = 1):
         self.dim0 = dim0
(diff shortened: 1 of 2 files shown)


src/transformers/core_model_loading.py
```diff
@@ -245,6 +245,45 @@ def reverse_op(self) -> ConversionOps:
...
+        # Here it's the first operation of a chain, so return the source
+        if len(target_patterns) > 1:
+            # Here it's the first operation of a chain, so return the source
+            if len(source_patterns) == 1:
+                return source_patterns[0]
+            else:
+                raise ValueError("Undefined Operation encountered!")
+        # Here it's the only operation, or the last operation in a chain, so we return the target
+        else:
+            return target_patterns[0]
```
IlyasMoutawwakil, line 279: so it has to either be the first or the last in a chain of ops ? I can see the dequantization ops breaking this as they extend the chain 🥲 correct me if i'm wrong

Cyrilvallez: Nop, should be alright with quantization! They do not change the targets/sources!

src/transformers/core_model_loading.py
```diff
@@ -245,6 +245,44 @@ def reverse_op(self) -> ConversionOps:
...
 
 
+class Transpose(ConversionOps):
+    """
+    Transposes the given tensor along dim0 and dim1.
+    """
+
+    def __init__(self, dim0: int = 0, dim1: int = 1):
+        self.dim0 = dim0
+        self.dim1 = dim1
```
ArthurZucker, line 255: should we pass the full converter to change the shard dim at init?

ArthurZucker (commented):
Sounds good but concerned about TP + Transpose, we need a more abstract solution like updating the shard dim based on op at init of ops?

src/transformers/core_model_loading.py
vasqu: Hmm, but aren't you assuming that we need to change the shard dim based on transpose? At least the transpose that do use it make it so that they are aligned with other models, i.e. they will need to be sharded the same way as intended.

There might be a few models that shard differently tho, I would consider this not part of the conversion op tho - otherwise we will mix in our assumptions :thinking:

Cyrilvallez: It's due to ops order - sharding happens before Transpose, so then if you Transpose the dim that was sharded on, you've got an issue

vasqu: Ah I messed up the order, I thought it was transpose then shard

Cyrilvallez: Life would be too nice... 😆

Cyrilvallez:
Actually I checked and the params on which `Transpose` are used are not part of the TP plan anyway. So i'd rather merge this PR now as it fixes the model and unblocks other PRs, then we can see in the future about the TP shard dim if the issue comes up

vasqu:
Can we add a comment at least, TODO/FIXME or similar at least? This is quite important (although not used with TP atm)

Cyrilvallez:
Done. I think it's easier this way as it's not as straightforward as I thought it would be to always switch the shard dim if required

vasqu (approved):
Discussed internally, we will need to double-check many current ops that we already have

PR #43323, opened 2026-01-16
🚨 Delete duplicate code in backbone utils

zucchini-nlp:
# What does this PR do?

This PR cleans up backbone utilities. Specifically, we have currently 5 different config attr to decide which backbone to load, most of which can be merged into one and seem redundant
After this PR, we'll have only one `config.backbone_config` as a single source of truth. The models will load the backbone `from_config` and load pretrained weights only if the checkpoint has any weights saved. The overall idea is same as in other composite models

I removed these config attr:
- `backbone` - the backbone model id is now used to create a `backbone_config` by loading it from the hub or from timm
- `backbone_kwargs` - it is used to update `backbone_config` with user-provided kwargs (i.e. `backbone_config = CONFIG_MAPPING[model_type](**backbone_kwargs)`)
- `use_pretrained_backbone` - we don't load a pretrained backbone anymore unless the user is calling `from_pretrained`. The default is to initialize a model with random weights and let the users either tune it from scratch or load pretrained weights themselves
- `use_timm_backbone` - we can infer model type from config and the requested backbone type, so this arg was redundant

Along the way, I also updated the tests and docs. Recommended review path: `modeling_backbone_utils.py` -> `auto_factory.py` -> `timm backbone model files` -> couple other models of your choice


docs/source/en/backbones.md
```diff
@@ -69,12 +69,12 @@ When you know a model supports a backbone, you can load the backbone and neck di
...
 
-Set `backbone` to a pretrained model and  `use_pretrained_backbone=True` to use pretrained weights instead of randomly initialized weights.
+Note that initializing from config will create the model with random weights. If you want to load a pretrained model, use `from_pretrained` API.
 
 ```py
 from transformers import MaskFormerConfig, MaskFormerForInstanceSegmentation
 
-config = MaskFormerConfig(backbone="microsoft/resnet-50", use_pretrained_backbone=True)
+config = MaskFormerConfig(backbone="microsoft/resnet-50")
 model = MaskFormerForInstanceSegmentation(config)
```
zucchini-nlp, line 79: imo it doesn't serve much purpose, loading a random init model with pretrained backbone. User still has to tune the model so it can be used

Therefore I deleted this feature. Pretrained weights are loaded `from_pretrained` and random weights `from_config`, same way as any other model

docs/source/en/backbones.md
```diff
@@ -96,14 +96,12 @@ model = MaskFormerForInstanceSegmentation(config)
...
 
-[timm](https://hf.co/docs/timm/index) is a collection of vision models for training and inference. Transformers supports timm models as backbones with the [`TimmBackbone`] and [`TimmBackboneConfig`] classes.
-
-Set `use_timm_backbone=True` to load pretrained timm weights, and `use_pretrained_backbone` to use pretrained or randomly initialized weights.
+[timm](https://hf.co/docs/timm/index) is a collection of vision models for training and inference. Transformers supports timm models as backbones with the [`TimmBackbone`] and [`TimmBackboneConfig`] classes. Set the neccessary backbone checkpoint in `backbone` to create a model with Timm backbone with randomly initialized weights.
 
 ```py
 from transformers import MaskFormerConfig, MaskFormerForInstanceSegmentation
 
-config = MaskFormerConfig(backbone="resnet50", use_timm_backbone=True, use_pretrained_backbone=True)
```
zucchini-nlp, line 106: `use_timm_backbone` is not really needed imo. We can infer if the requested checkpoint is from timm or HF by checking if repo exists on the hub with a valid config

Deleted it as well as a redudant arg

src/transformers/models/beit/configuration_beit.py
```diff
@@ -186,9 +186,9 @@ def __init__(
 
         # backbone attributes
         self.stage_names = ["stem"] + [f"stage{idx}" for idx in range(1, self.num_hidden_layers + 1)]
-        self._out_features, self._out_indices = get_aligned_output_features_output_indices(
-            out_features=out_features, out_indices=out_indices, stage_names=self.stage_names
-        )
+        out_indices = list(out_indices) if out_indices is not None else None
+        self._out_features, self._out_indices = out_features, out_indices
+        self.align_output_features_output_indices()
```
zucchini-nlp: The feature-index aligning happens in `Mixin` when we call `align_output_features_output_indices`

src/transformers/modeling_backbone_utils.py
```diff
@@ -0,0 +1,343 @@
...
+    # Attribute to indicate if the backbone has attention and can return attention outputs.
+    # Should be set to `False` for conv-based models to be able to run `forward_with_filtered_kwargs`
+    has_attentions: bool = True
+
+    def _init_timm_backbone(self, config) -> None:
+        """
+        Initialize the backbone model from timm The backbone must already be loaded to self._backbone
+        """
+
+        # These will diagree with the defaults for the transformers models e.g. for resnet50
```
yonigozlan:
```suggestion
        # These will disagree with the defaults for the transformers models e.g. for resnet50
```

src/transformers/models/beit/configuration_beit.py
yonigozlan: Nit: Can we set `self._out_features` and `self. _out_indices` in `align_output_features_output_indices(out_features, out_indices)`? To only call `self.align_output_features_output_indices(out_features, out_indices)` in backbone configs instead of these three lines. It would also simplify the setters

src/transformers/modeling_backbone_utils.py
```diff
@@ -0,0 +1,343 @@
...
+                The names of the features for the backbone to output. Defaults to `config._out_features` if not provided.
+            out_indices (`list[int]` or `tuple[int]`, *optional*):
+                The indices of the features for the backbone to output. Defaults to `config._out_indices` if not provided.
+            stage_names (`list[str]`, *optional*):
+                The names of the stages of the backbone. Defaults to `config.stage_names` if not provided.
+        """
+        self.stage_names = stage_names if stage_names is not None else self.stage_names
+        self._out_features = out_features if out_features is not None else self._out_features
+        self._out_indices = out_indices if out_indices is not None else self._out_indices
+        self.align_output_features_output_indices()
```
yonigozlan: This doesn't seem to be used anywhere for now, is it really needed?

src/transformers/modeling_backbone_utils.py
```diff
@@ -0,0 +1,343 @@
...
+class BackboneMixin:
+    backbone_type: BackboneType | None = None
+
+    # Attribute to indicate if the backbone has attention and can return attention outputs.
+    # Should be set to `False` for conv-based models to be able to run `forward_with_filtered_kwargs`
+    has_attentions: bool = True
+
+    def _init_timm_backbone(self, config) -> None:
+        """
+        Initialize the backbone model from timm The backbone must already be loaded to self._backbone
```
yonigozlan:
```suggestion
        Initialize the backbone model from timm. The backbone must already be loaded to self._backbone
```

src/transformers/modeling_backbone_utils.py
```diff
@@ -0,0 +1,343 @@
...
+            self.backbone_type = BackboneType.TIMM
+        else:
+            self.backbone_type = BackboneType.TRANSFORMERS
+
+        if self.backbone_type == BackboneType.TIMM:
+            self._init_timm_backbone(config)
+        elif self.backbone_type == BackboneType.TRANSFORMERS:
+            self._init_transformers_backbone(config)
+        else:
+            raise ValueError(f"backbone_type {self.backbone_type} not supported.")
```
yonigozlan: It would be great if we could get this to be called when `super().__init__` is called, so that we don't have to call it manually everytime in the model init.
If we make sure that BackboneMixin is always the first in the inheritance list with an mro check, we could override init in BackboneMixin to call super.__init__ then call _init_backbone. wdyt?

src/transformers/modeling_backbone_utils.py
```diff
@@ -0,0 +1,343 @@
...
+    backbone = kwargs.pop("backbone") if kwargs.get("backbone") is not None else default_backbone
+    kwargs.pop("use_pretrained_backbone", None)
+
+    # Init timm backbone with hardcoded values for BC
+    if (
+        timm_default_kwargs is not None
+        and use_timm_backbone
+        and backbone is not None
+        and backbone_config is None
+        and backbone_kwargs is None
```
yonigozlan: Can this be None? we have "backbone_kwargs = kwargs.pop("backbone_kwargs", {})" above

yonigozlan (approved):
Thank you this is greatly needed! I left a few comments, mainly to see if we can simplify further

src/transformers/models/beit/configuration_beit.py
zucchini-nlp: sure, I actually added a `set_output_features_output_indices` as well so we can use that to "set+align" values

src/transformers/modeling_backbone_utils.py
zucchini-nlp: right, I wanted to use it instead of calling `align` and didn't yet have time. I'll just merge it into "align_features" i think

src/transformers/modeling_backbone_utils.py
zucchini-nlp: good point, lemme try

src/transformers/modeling_backbone_utils.py
zucchini-nlp: i remember this part caused some issues due to being the default in some old models. I think it was set to `None` explicitly in tests, but lemme recheck

The idea is to fast exit with this path whenever a config has "default_timm_kwargs" and nothing else is set to `True`

docs/source/en/tasks/training_vision_backbone.md
```diff
@@ -38,13 +38,18 @@ Initialize [`DetrConfig`] with the pre-trained DINOv3 ConvNext backbone. Use `nu
...
+
+config = DetrConfig(backbone_config=backbone_config,
                     num_labels=1, id2label={0: "license_plate"}, label2id={"license_plate": 0})
 model = DetrForObjectDetection(config)
 
-for param in model.model.backbone.parameters():
-    param.requires_grad = False
+# Assign pretrained backbone checkpoint and freeze the weights
+model.model.backbone = backbone
+model.model.freeze_backbone()
```
zucchini-nlp, line 51: backbones are usually set under different module names, so I'm thinking we can add a `self.get_backbone` and `self.set_backbone` for easier manipulation

Maybe next PR, this one is already big

tests/test_backbone_common.py
```diff
@@ -140,14 +140,6 @@ def test_create_from_modified_config(self):
             self.assertEqual(len(result.feature_maps), 1)
             self.assertEqual(len(model.channels), 1)
 
-            # Check backbone can be initialized with fresh weights
```
zucchini-nlp, line 143: backbone is always init with fresh weight unless it is `from_pretrained`, test is not needed anymore

tests/test_configuration_common.py
```diff
@@ -133,7 +133,8 @@ def create_and_test_config_from_and_save_pretrained_composite(self):
             for sub_config_key, sub_class in sub_configs.items():
                 if general_config_dict[sub_config_key] is not None:
                     if sub_class.__name__ == "AutoConfig":
-                        sub_class = sub_class.for_model(**general_config_dict[sub_config_key]).__class__
+                        sub_config_dict = copy.deepcopy(general_config_dict[sub_config_key])
+                        sub_class = sub_class.for_model(**sub_config_dict).__class__
```
zucchini-nlp, line 137: `for_model` was changing the config in-place so we deepcopy before

zucchini-nlp:
run-slow: auto, beit, bit, conditional_detr, convnext, convnextv2, d_fine, dab_detr, deformable_detr, depth_anything, detr, timm_backbone

Cyrilvallez:
Hey! Very nice initiative, indeed backbones have been quite annoying and have no reason to exist at all in general (we can simply do usual composite models)! Especially the `timm_backbone` and its pretrained weights loading which does not work correctly with our `from_pretrained`.
In general, we could completely remove `timm_backbone` in favor of `timm_wrapper` I believe if we remove the `use_pretrained_weights` anyway 🤗
And about
> use_pretrained_backbone - we don't load a pretrained backbone anymore unless the user is calling from_pretrained

I believe we need to remove it completely - otherwise the loading is being done in the `__init__`, and when `from_pretrained` is called, if those weights are not in the main hf repo because they are assumed to only be on timm repo, then the weights will be considered missing and will be reinitialized.... So we can fully remove it IMO and assume weights need to live in hf repo, which will solve a long due issue!

src/transformers/utils/backbone_utils.py
```diff
@@ -1,379 +1,19 @@
...
-    out_indices = list(out_indices) if out_indices is not None else None
-    # First verify that the out_features and out_indices are valid
-    verify_out_features_out_indices(out_features=out_features, out_indices=out_indices, stage_names=stage_names)
-    output_features, output_indices = _align_output_features_output_indices(
-        out_features=out_features, out_indices=out_indices, stage_names=stage_names
+class BackboneConfigMixin(BackboneConfigMixin):
+    warnings.warn(
+        "Importing `BackboneConfigMixin` from `utils/backbone_utils.py` is deprecated and will be removed in "
+        "Transformers v5.10. Import as `from transformers.modeling_backbone_utils import BackboneConfigMixin` instead.",
+        FutureWarning,
```
Cyrilvallez, line 10: I don't really mind the renaming in general, but not sure if it's really needed

src/transformers/models/bit/modeling_bit.py
```diff
@@ -759,12 +759,11 @@ def forward(
     BiT backbone, to be used with frameworks like DETR and MaskFormer.
     """
 )
-class BitBackbone(BitPreTrainedModel, BackboneMixin):
+class BitBackbone(BackboneMixin, BitPreTrainedModel):
```
Cyrilvallez, line 762: Why do we need to switch the order here? Cause of `__init__`?

Cyrilvallez:
In general we want to move away from the backbone API as much as possible in favor of standard composite models so this will help 🤗

src/transformers/utils/backbone_utils.py
zucchini-nlp: you mean moving to a different place? It was hitting a circular import  with `PreTrainedConfig`, atm I imported it lazily in the new file so it doesn't get imported at the top

src/transformers/models/bit/modeling_bit.py
zucchini-nlp: yeah, to init the backbone stuff first and then call `BitPreTrainedModel.__init__` from within it

zucchini-nlp:
> In general, we could completely remove timm_backbone in favor of timm_wrapper I believe if we remove the use_pretrained_weights anyway 🤗

100% , that is my goal for the subsequent PR. This is already hard to manage with gh conflicts and models, so we need to first clean-up extra kwargs and keep it as `timm_backbone`

> I believe we need to remove it completely. So we can fully remove it IMO and assume weights need to live in hf repo, which will solve a long due issue!

It is removed already completely. We pop the kwarg and never use it, it's hardcoded as `False` in timm backbone modeling file. Atm the slow tests are passing which I believe means that the weights are already in hf repo for official releases

src/transformers/utils/backbone_utils.py
Cyrilvallez: No I meant you renamed `backbone_utils` -> `modeling_backbone_utils` haha

Cyrilvallez (approved):
Alright very nice! Super happy to gradually move away from the backbone API! Feel free to merge once CI is back to green (after the conflict handling issues)

src/transformers/utils/backbone_utils.py
zucchini-nlp: oh that! No reason behind, just a matter of taste. Can rename it back for sure

Cyrilvallez: Yes I believe it's a little easier as we want to deprecate the backbone api, then it avoids maintaining yet another BC entry point!

zucchini-nlp:
run-slow: auto, beit, bit, conditional_detr, convnext, convnextv2, d_fine, dab_detr, deformable_detr, depth_anything

zucchini-nlp:
run-slow: beit, bit, conditional_detr, convnext, convnextv2, d_fine, dab_detr, deformable_detr, depth_anything, timm_backbone, detr, pp_doclayout_v3, resnet, vitpose

zucchini-nlp:
Nice, slow CI passes and failing ones are just flaky

PR #43326, opened 2026-01-16
Fix Mxfp4 dequantize

Cyrilvallez:
# What does this PR do?

As per the title. Fix https://github.com/huggingface/transformers/issues/43317.
They move the devices without being asked to do it, resulting in OOMs when using with a device_map="auto". Also, no need to upscale the dtype to int64 as they are only used to indexing and it takes up more memory in constrained device_map="auto" settings

### DIFF
--- a/src/transformers/integrations/mxfp4.py
+++ b/src/transformers/integrations/mxfp4.py
@@ -128,20 +128,19 @@ class Mxfp4Dequantize(ConversionOps):
         missing_keys=None,
         **kwargs,
     ) -> dict[str, torch.Tensor]:
-        param_data = {}
         if "_blocks" in input_dict.keys():
             if isinstance(input_dict["_blocks"], list):
-                param_data["_blocks"] = input_dict["_blocks"][0]
+                blocks = input_dict["_blocks"][0]
             else:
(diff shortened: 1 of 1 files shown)


src/transformers/integrations/mxfp4.py
```diff
@@ -465,12 +456,9 @@ def dequantize(module, param_name, param_value, target_device, dq_param_name, **
                 delattr(module, scales_attr)
 
 
-def dequantize_convertops(blocks, scales, target_device):
+def dequantize_convertops(blocks, scales):
     dequantized = convert_moe_packed_tensors(blocks, scales)
-    if target_device == "cpu" and torch.cuda.is_available():
-        torch.cuda.empty_cache()
```
Cyrilvallez, line 471: This is extremely bad! Very slow, and works against the memory allocator during loading! So it takes a lot of time to run, and we need more time later to re-allocate that memory!

SunMarc: Indeed !

SunMarc (approved):
Thanks for fixing ! Can you have a quick look @MekkCyber ?

MekkCyber (approved):
Yes we can do that, lgtm

PR #43327, opened 2026-01-16
Getting closer

BenjaminBossan:
It was necessary to flatten the LoRA weights for 3d MoE, as LoRA always expected 2d weights (being nn.Linear).

Therefore, I added new conversion ops: `PermuteDims` and `FlattenDims`. I also added a convenience function `_block_diag_3d`, it could probably be optimized.

Furthermore, I now have separate branches for `down_proj` and `gate_up_proj` as they are different enough that treating them in one branch is too confusing.

### DIFF
--- a/src/transformers/integrations/peft.py
+++ b/src/transformers/integrations/peft.py
@@ -21,6 +21,7 @@ from typing import TYPE_CHECKING, Any, Literal, Optional
 from ..core_model_loading import (
     Concatenate,
     ConversionOps,
+    MergeModulelist,
     WeightConverter,
     WeightRenaming,
 )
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Current error:

```
MixtralForCausalLM LOAD REPORT from: peft-internal-testing/mixtral-pre-v5-lora
Key                                                              | Status   | Details                                                                                  
-----------------------------------------------------------------+----------+------------------------------------------------------------------------------------------
model.layers.{0, 1}.mlp.experts.lora_A.default.weight            | MISMATCH | Reinit due to size mismatch ckpt: torch.Size([64, 3584]) vs model:torch.Size([64, 1024]) 
model.layers.{0, 1}.mlp.experts.lora_B.default.weight            | MISMATCH | Reinit due to size mismatch ckpt: torch.Size([64, 1024]) vs model:torch.Size([3584, 64]) 
model.layers.{0, 1}.mlp.experts.base_layer.lora_A.default.weight | MISMATCH | Reinit due to size mismatch ckpt: torch.Size([128, 1024]) vs model:torch.Size([64, 7168])
model.layers.{0, 1}.mlp.experts.base_layer.lora_B.default.weight | MISMATCH | Reinit due to size mismatch ckpt: torch.Size([128, 7168]) vs model:torch.Size([1024, 64])
```

So there isn't much missing to align the shapes.

PR #43330, opened 2026-01-17
fix: _get_num_multimodal_tokens video branch (#43329)

floor-licker:
# What does this PR do?

Fixes #43329

 ### Summary
  `_get_num_multimodal_tokens(video_sizes=...)` crashed in multiple multimodal processors because the
  video branch called `self.video_processor.get_number_of_video_patches(...)` (not implemented on several video
  processors), and used `merge_size` without initializing it (it was only set in the image branch).

  This meant CI never exercised the video route (tests only covered `image_sizes`).

  ### Changes

  - Initialize merge_size in the video_sizes branch (from videos_kwargs.get("merge_size") fallback to
    self.video_processor.merge_size).
  - Add get_number_of_video_patches() to the relevant *VideoProcessor implementations (either as an
    alias to the existing helper or as a small utility consistent with preprocessing).
  - Add unit coverage that calls _get_num_multimodal_tokens(video_sizes=[(8, 224, 224)]) so the video
    branch is exercised in CI.

  ### Affected processors

  - Ernie4_5_VL_MoeProcessor
  - Glm4vProcessor
  - Glm46VProcessor
  - Qwen2VLProcessor
  - Qwen2_5_VLProcessor
  - Qwen3VLProcessor
  - VideoLlama3Processor

  ### Relevant Tests

  python -m pytest -q \
    tests/models/qwen2_vl/test_processing_qwen2_vl.py::Qwen2VLProcessorTest::test_get_num_vision_tokens
  \
    tests/models/qwen2_5_vl/
  test_processing_qwen2_5_vl.py::Qwen2_5_VLProcessorTest::test_get_num_vision_tokens \
    tests/models/qwen3_vl/test_processing_qwen3_vl.py::Qwen3VLProcessorTest::test_get_num_vision_tokens
  \
    tests/models/video_llama_3/
  test_processing_video_llama_3.py::VideoLlama3ProcessorTest::test_get_num_vision_tokens \
    tests/models/ernie4_5_vl_moe/
  test_processing_ernie4_5_vl_moe.py::Ernie4_5_VL_MoeProcessorTest::test_get_num_vision_tokens \
    tests/models/glm4v/test_processor_glm4v.py::Glm4vProcessorTest::test_get_num_vision_tokens_video \
    tests/models/glm46v/test_processor_glm46v.py::Glm46VProcessorTest::test_get_num_vision_tokens_video

### DIFF
--- a/src/transformers/models/ernie4_5_vl_moe/video_processing_ernie4_5_vl_moe.py
+++ b/src/transformers/models/ernie4_5_vl_moe/video_processing_ernie4_5_vl_moe.py
@@ -535,6 +535,42 @@ class Ernie4_5_VL_MoeVideoProcessor(BaseVideoProcessor):
             tensor_type=return_tensors,
         )
 
+    def get_number_of_video_patches(self, num_frames: int, height: int, width: int, videos_kwargs=None):
+        """
+        A utility that returns number of video patches for a given video size.
+
+        Args:
+            num_frames (`int`):
+                Number of frames in the input video.
+            height (`int`):
+                Height of the input video.
+            width (`int`):
+                Width of the input video.
+            videos_kwargs (`dict`, *optional*):
+                Any kwargs to override defaults of the video processor.
+        Returns:
+            `int`: Number of video patches per video.
+        """
+        size = videos_kwargs.get("size", None) or self.size
+        patch_size = videos_kwargs.get("patch_size", None) or self.patch_size
+        merge_size = videos_kwargs.get("merge_size", None) or self.merge_size
+        do_resize = videos_kwargs.get("do_resize", None)
+        do_resize = self.do_resize if do_resize is None else do_resize
+
+        resized_height, resized_width = height, width
+        if do_resize:
+            resized_height, resized_width = smart_resize(
+                height,
+                width,
+                factor=patch_size * merge_size,
+                min_pixels=size["shortest_edge"],
+                max_pixels=size["longest_edge"],
+            )
+
+        grid_h, grid_w = resized_height // patch_size, resized_width // patch_size
+        grid_t = num_frames + int(num_frames % 2 != 0)
+        return grid_t * grid_h * grid_w
+
     @add_start_docstrings(
         BASE_VIDEO_PROCESSOR_DOCSTRING,
     )
--- a/src/transformers/models/qwen2_vl/processing_qwen2_vl.py
+++ b/src/transformers/models/qwen2_vl/processing_qwen2_vl.py
@@ -162,6 +162,7 @@ class Qwen2VLProcessor(ProcessorMixin):
         if video_sizes is not None:
             videos_kwargs = Qwen2VLProcessorKwargs._defaults.get("videos_kwargs", {})
             videos_kwargs.update(kwargs)
+            merge_size = videos_kwargs.get("merge_size", None) or self.video_processor.merge_size
             num_video_patches = [
                 self.video_processor.get_number_of_video_patches(*video_size, videos_kwargs)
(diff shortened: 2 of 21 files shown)


Rocketknight1:
cc @zucchini-nlp, feels like a code agent PR but feel free to merge it if it works

stefgina:
cc @Rocketknight1 @zucchini-nlp i created this issue initially asking to also do the PR, but this came in automatically before i got an answer. no worries, if it works it works.

suggestions: 
I would suggest we also include a batch to test the function _get_num_multimodal_tokens() -like we do in the image branch-, instead of a single item, and also improve a bit the depth of these tests (eg. `do_resize` ).

src/transformers/models/qwen2_vl/video_processing_qwen2_vl.py
```diff
@@ -323,5 +323,10 @@ def get_num_of_video_patches(self, num_frames: int, height: int, width: int, vid
         grid_t = num_frames // temporal_patch_size
         return grid_t * grid_h * grid_w
 
+    def get_number_of_video_patches(self, num_frames: int, height: int, width: int, videos_kwargs=None):
+        return self.get_num_of_video_patches(
+            num_frames=num_frames, height=height, width=width, videos_kwargs=videos_kwargs
+        )
+
```
zucchini-nlp, line 330: ?

src/transformers/models/qwen2_vl/processing_qwen2_vl.py
```diff
@@ -162,6 +162,7 @@ def _get_num_multimodal_tokens(self, image_sizes=None, video_sizes=None, **kwarg
         if video_sizes is not None:
             videos_kwargs = Qwen2VLProcessorKwargs._defaults.get("videos_kwargs", {})
             videos_kwargs.update(kwargs)
+            merge_size = videos_kwargs.get("merge_size", None) or self.video_processor.merge_size
```
zucchini-nlp, line 165: good catch here!

src/transformers/models/ernie4_5_vl_moe/video_processing_ernie4_5_vl_moe.py
```diff
@@ -535,6 +535,42 @@ def _preprocess(
             tensor_type=return_tensors,
         )
 
+    def get_number_of_video_patches(self, num_frames: int, height: int, width: int, videos_kwargs=None):
+        """
+        A utility that returns number of video patches for a given video size.
+
```
zucchini-nlp, line 541: quite sad that these are not copied with modular, even though there are overlapping parts with qwen2-vl. Ernie is different but other models are 90% identical to qwen

Let me take a note for myself and modularize it all. That way we don't have to add `get_num_video_patches` manually like this in all models

zucchini-nlp (commented):
The video path is not supported well currently and it is not a "missed bug". We were moving gradually and started with images. There is one community contrib working on video on vLLM side so I pushed by local changes in https://github.com/huggingface/transformers/pull/42919 which should fix `_get_num_multimodal_tokens` for videos

Though it is stale atm and from what I see is not super high in priority. I will take a look at the PR

That said, this PR is in general in good shape if we take advantage of modular. But since it is a code agent PR, I am not sure if it is worth asking for iterations. I will close it in favor of #42919

src/transformers/models/ernie4_5_vl_moe/video_processing_ernie4_5_vl_moe.py
jtranq: Fair enough, makes sense

PR #43341, opened 2026-01-19
skip some unsupported tests for glm_image model 

kaixuanliu:
@ydshieh pls help review, thx!

### DIFF
--- a/tests/models/glm_image/test_modeling_glm_image.py
+++ b/tests/models/glm_image/test_modeling_glm_image.py
@@ -283,6 +283,30 @@ class GlmImageModelTest(ModelTesterMixin, GenerationTesterMixin, unittest.TestCa
     def test_multi_gpu_data_parallel_forward(self):
         pass
 
+    @unittest.skip(
+        "GlmImage has a VQ module that uses `weight.data` directly in forward which prevent offloading on that module"
+    )
+    def test_disk_offload_safetensors(self):
+        pass
+
+    @unittest.skip(
+        "GlmImage has a VQ module that uses `weight.data` directly in forward which prevent offloading on that module"
+    )
+    def test_disk_offload_bin(self):
+        pass
+
+    @unittest.skip(
+        "GlmImage has a VQ module that uses `weight.data` directly in forward which prevent offloading on that module"
+    )
+    def test_cpu_offload(self):
+        pass
+
+    @unittest.skip(
+        "GlmImage has a VQ module that uses `weight.data` directly in forward which prevent offloading on that module"
+    )
+    def test_model_parallelism(self):
+        pass
+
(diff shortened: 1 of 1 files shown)


tests/models/glm_image/test_modeling_glm_image.py
```diff
@@ -345,6 +375,27 @@ def test_retain_grad_hidden_states_attentions(self):
...
+    )
+    def test_flash_attn_2_fp32_ln(self):
+        pass
+
+    @unittest.skip(
+        reason="GlmImage is a multimodal model that requires pixel_values and image_grid_thw. "
+        "This test only uses input_ids and attention_mask which causes NoneType iteration error."
+    )
+    def test_flash_attn_2_from_config(self):
+        pass
```
vasqu, line 398: They will be fixed by #43346, no?

vasqu: Ah nvm, sorry confused the model name, it's a different model :cry:

IlyasMoutawwakil (approved):
LGTM ! some PRs are red because of these

tests/models/glm_image/test_modeling_glm_image.py
```diff
@@ -337,6 +361,12 @@ def test_training_gradient_checkpointing_use_reentrant(self):
     def test_training_gradient_checkpointing_use_reentrant_false(self):
         pass
 
+    @unittest.skip(
+        reason="This architecture seem to not compute gradients properly when using GC, check: https://github.com/huggingface/transformers/pull/27124"
+    )
+    def test_training_gradient_checkpointing_use_reentrant_true(self):
+        pass
```
vasqu, line 369: As far as I can see this is the only test that is currently blocking us on CI

tests/models/glm_image/test_modeling_glm_image.py
```diff
@@ -283,6 +283,30 @@ def test_sdpa_can_dispatch_on_flash(self):
...
+        "GlmImage has a VQ module that uses `weight.data` directly in forward which prevent offloading on that module"
+    )
+    def test_cpu_offload(self):
+        pass
+
+    @unittest.skip(
+        "GlmImage has a VQ module that uses `weight.data` directly in forward which prevent offloading on that module"
+    )
+    def test_model_parallelism(self):
+        pass
```
vasqu: I'd like to avoid skipping (maybe xfail in this case) too much and it seems to be something we should support at some point cc @zucchini-nlp

vasqu (approved):
Let's change the offloading tests at least to xfail, we should support them and they are not those that are blocking our CI

vasqu:
Changed some to xfail myself since it's blocking CI

PR #43353, opened 2026-01-19
[`Qwen3 Omni Moe`] Fix 16 byte alignment in tests

vasqu:
As per title, this is required for `torch._grouped_mm`; related failures e.g. https://app.circleci.com/pipelines/github/huggingface/transformers/160384/workflows/1c47e8a5-8aea-4fd5-9ec6-b6aab32e4e55/jobs/2110308/parallel-runs/3

I cannot reproduce these issues locally, would be nice to have something that could Edit: needs to be run on CPU to trigger


vasqu:
run-slow: qwen3_omni_moe, qwen3_vl_moe

vasqu:
Discussed internally, we will move this to the conversions instead for all moes for now

PR #43354, opened 2026-01-19
Skip glm_image integration tests (wrong checkpoint)

raimbekovm:
# What does this PR do?

The integration tests for `GlmImageForConditionalGeneration` were failing because they load `zai-org/GLM-4.5V` which is a `glm4v_moe` model, not `glm_image`. This causes shape mismatches:

```
lm_head.weight: [151552, 4096] vs [16512, 4096]  
patch_embed.proj.weight: [1536, 3, 2, 14, 14] vs [1536, 3, 14, 14]
```

The correct checkpoint is `zai-org/GLM-Image` with `subfolder='vision_language_encoder'`.

There's also a deeper issue — tests use VLM prompts ("What kind of dog is this?") but GlmImage is for image generation, not visual Q&A. So the test logic itself needs rewriting.

This PR skips the tests until they can be properly fixed.

Fixes #43344

## Before submitting
      Pull Request section?
      to it if that's the case. #43344

## Who can review?

@zucchini-nlp @ydshieh

### DIFF
--- a/tests/models/glm_image/test_modeling_glm_image.py
+++ b/tests/models/glm_image/test_modeling_glm_image.py
@@ -347,6 +347,12 @@ class GlmImageModelTest(ModelTesterMixin, GenerationTesterMixin, unittest.TestCa
 
 
 @require_torch
+@unittest.skip(
+    reason="Integration tests use wrong checkpoint (zai-org/GLM-4.5V is glm4v_moe, not glm_image). "
+    "The correct checkpoint is zai-org/GLM-Image with subfolder='vision_language_encoder'. "
+    "Tests also need to be rewritten as GlmImage is for image generation, not VLM-style Q&A. "
+    "See https://github.com/huggingface/transformers/issues/43344"
+)
 @slow
 class GlmImageIntegrationTest(unittest.TestCase):
     @classmethod


raimbekovm:
For context, here's how the checkpoint should be loaded according to the docs:

```python
model = GlmImageForConditionalGeneration.from_pretrained(
    "zai-org/GLM-Image",
    subfolder="vision_language_encoder",
    ...
)
processor = AutoProcessor.from_pretrained(
    "zai-org/GLM-Image",
    subfolder="processor",
    ...
)
```

The model outputs vision tokens for a diffusion decoder, not text answers. So the test prompts and expected outputs would need to match that behavior.

Looks like the tests were added in #43100 and probably written before the checkpoint was fully available on the hub.

Rocketknight1:
cc @ydshieh as well

tests/models/glm_image/test_modeling_glm_image.py
```diff
@@ -347,6 +347,12 @@ def test_generate_compile_model_forward_fullgraph(self):
 
 
 @require_torch
+@unittest.skip(
+    reason="Integration tests use wrong checkpoint (zai-org/GLM-4.5V is glm4v_moe, not glm_image). "
+    "The correct checkpoint is zai-org/GLM-Image with subfolder='vision_language_encoder'. "
+    "Tests also need to be rewritten as GlmImage is for image generation, not VLM-style Q&A. "
+    "See https://github.com/huggingface/transformers/issues/43344"
+)
```
zucchini-nlp, line 355: the model was released day-0 so we indeed had no integration tests yet. As we discussed internally with @ydshieh, we plan to improve on slow tests for day-0 releases.

Now that the model is already released, would you mind updating these to use the correct ckpt? If not, I can do it myself. IMO we should not be skipping it, we'll never get back to update checkpoints otherwise 😢

zucchini-nlp: A tiny tests checking generation with `bs=1` and `do_sample=True` will be enough. Also make sure to set seed to repro
The whole model pipe is anyway tested in diffusers I believe

zucchini-nlp:
Actually, I think slow tests will be added as part of https://github.com/huggingface/transformers/pull/43342 by the authors :)

raimbekovm:
Glad I could help identify the issue — feel free to ping me if you need any help! рахмет:)

zucchini-nlp:
Let's close the PR here, thanks for your help!

PR #43362, opened 2026-01-20
Fix non-deterministic output in MoE experts with batched_mm/grouped_mm

### LINKED ISSUE #43360: [v5] do_sample false not working ('batched_mm'). transformers 5.0 main branch
### System Info

Image

main branch transformers 5.0.0.dev

```python
import torch

MODEL_PATH = "zai-org/GLM-4.7-Flash"
messages = [{"role": "user", "content": "hello"}]
tokenizer = AutoTokenizer.from_pretrained(MODEL_PATH)
inputs = tokenizer.apply_chat_template(
    messages,
    tokenize=True,
    add_generation_prompt=True,
    return_dict=True,
    return_tensors="pt",
)
    pretrained_model_name_or_path=MODEL_PATH,
    torch_dtype=torch.bfloat16,
    device_map="auto",
    experts_implementation='batched_mm',
)

inputs = inputs.to(model.device)
generated_ids = model.generate(**inputs, max_new_tokens=128, do_sample=False)
output_text = tokenizer.decode(generated_ids[0][inputs.input_ids.shape[1]:])
print(output_text)
```

### Who can help?

-

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

-

### Expected behavior

.

raimbekovm:
# What does this PR do?

Fixes non-deterministic output when using `do_sample=False` with MoE models and `experts_implementation='batched_mm'` or `'grouped_mm'`.

The issue is `index_add_` with duplicate indices on CUDA. When `top_k > 1`, `token_idx` contains duplicates (`[0,0,1,1,2,2,...]`) and CUDA's `atomicAdd` doesn't guarantee operation order. Since FP addition isn't associative, this causes different rounding errors on each run.

The fix replaces `index_add_` with `view().sum()` - this works because each token appears exactly `top_k` times in sequence, so we can reshape and sum deterministically.

Fixes #43360

## Before submitting
      Pull Request section?
      to it if that's the case. #43360

## Who can review?

@IlyasMoutawwakil

### DIFF
--- a/src/transformers/integrations/moe.py
+++ b/src/transformers/integrations/moe.py
@@ -71,8 +71,8 @@ def batched_mm_experts_forward(
     device = hidden_states.device
     num_top_k = top_k_index.size(-1)
     num_tokens = hidden_states.size(0)
+    hidden_dim = hidden_states.size(-1)
     num_experts = self.gate_up_proj.size(0)
-    final_hidden_states = torch.zeros_like(hidden_states)
 
     # Flatten top_k_index to get expert_ids per selected sample
     expert_ids = top_k_index.reshape(-1)
@@ -117,10 +117,11 @@ def batched_mm_experts_forward(
     # Apply routing weights
     out_per_sample = out_per_sample * sample_weights.unsqueeze(-1)  # (S, hidden_dim)
 
-    # Accumulate results back to the final_hidden_states using original token indices
-    final_hidden_states.index_add_(0, token_idx, out_per_sample.to(final_hidden_states.dtype))
+    # Accumulate results using deterministic reshape+sum instead of index_add_
+    # (index_add_ with duplicate indices is non-deterministic on CUDA due to atomicAdd)
+    final_hidden_states = out_per_sample.view(num_tokens, num_top_k, hidden_dim).sum(dim=1)
 
-    return final_hidden_states
+    return final_hidden_states.to(hidden_states.dtype)
 
 
 def grouped_mm_experts_forward(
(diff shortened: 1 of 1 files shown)


src/transformers/integrations/moe.py
```diff
@@ -117,10 +117,11 @@ def batched_mm_experts_forward(
     # Apply routing weights
     out_per_sample = out_per_sample * sample_weights.unsqueeze(-1)  # (S, hidden_dim)
 
-    # Accumulate results back to the final_hidden_states using original token indices
-    final_hidden_states.index_add_(0, token_idx, out_per_sample.to(final_hidden_states.dtype))
+    # Accumulate results using deterministic reshape+sum instead of index_add_
+    # (index_add_ with duplicate indices is non-deterministic on CUDA due to atomicAdd)
+    final_hidden_states = out_per_sample.view(num_tokens, num_top_k, hidden_dim).sum(dim=1)
```
IlyasMoutawwakil, line 122: tbh I prefer this approach ! I only left the `final_hidden_states.index_add_` to stay as close to the eager impl as possible and hopefully get numerically close outputs, does token_idx contain duplicates in the eager path as well or is it impossible ?

raimbekovm: In eager, duplicates are basically impossible - would only happen if a token picked the same expert twice in top_k, but topk() can't return duplicate indices.

In batched_mm it's different - token_idx is `[0,0,1,1,2,2,...]` by design since we process all (token, expert) pairs together.

So eager index_add_ is fine, but batched one hits the atomicAdd race.

About numerical results - batched already doesn't match eager exactly (different accumulation order), so view().sum() doesn't change that. It just makes things deterministic.

IlyasMoutawwakil: thanks ! makes sense

IlyasMoutawwakil (approved):
LGTM ! thanks for the clean fix !

raimbekovm:
thanks for the quick review!

vasqu (approved):
LGTM, merging later today fixing up some CI related issues first

PR #43365, opened 2026-01-20
Fix: resolve lazy-load ImportError for PreTrainedModel (#42548)

karthik-0306:
## Description
This PR fixes the `ImportError: cannot import name 'PreTrainedModel' from 'transformers'` reported in #42548. 

**The Issue:**
`PreTrainedModel` was missing from the `_import_structure` in `src/transformers/__init__.py`. This caused the lazy-loading mechanism to fail when core frameworks (like PyTorch) were not present or specifically disabled, even though the source file existed.

**Changes:**
- Added `modeling_utils` mapping to `_import_structure` including `PreTrainedModel`, `PushToHubMixin`, and `unwrap_model`.
- Updated the `TYPE_CHECKING` block to include these exports for IDE/static analysis support.
- Formatted with `ruff` to ensure style compliance.

**Verification:**
Running the following in a minimal environment now succeeds:
`python -c "from transformers import PreTrainedModel; print('Success')"`

Fixes #42548

### DIFF
--- a/src/transformers/__init__.py
+++ b/src/transformers/__init__.py
@@ -133,6 +133,7 @@ _import_structure = {
         "is_wandb_available",
     ],
     "loss": [],
+    "modeling_utils": ["PreTrainedModel", "PushToHubMixin", "unwrap_model"],
     "pipelines": [
         "AnyToAnyPipeline",
         "AudioClassificationPipeline",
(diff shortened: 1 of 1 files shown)


karthik-0306:
I noticed a failure in tests/models/glm_image/test_modeling_glm_image.py. It seems unrelated to this PR's changes in __init__.py, as it appears to be a gradient checkpointing issue in that specific model.

PR #43369, opened 2026-01-20
fix: add device cast to position_ids_expanded in LlamaRotaryEmbedding

sasankkurnella:
## What does this PR do?

Fixes #43350

The `position_ids_expanded` tensor was not being moved to the same device as the input tensor `x`, causing a `RuntimeError` when `position_ids` is on CPU while the model is on CUDA.

### The Problem

In `LlamaRotaryEmbedding.forward()`, `inv_freq_expanded` is correctly moved to `x.device`, but `position_ids_expanded` was left on whatever device `position_ids` came from. This causes the matrix multiplication to fail:

```
RuntimeError: Expected all tensors to be on the same device, but got mat2 is on cpu, different from other tensors on cuda:0
```

### The Fix

Added `.to(x.device)` to `position_ids_expanded`, mirroring the behavior of `inv_freq_expanded`:

```python
# Before
position_ids_expanded = position_ids[:, None, :].float()

# After
position_ids_expanded = position_ids[:, None, :].float().to(x.device)
```

This is a safe no-op when tensors are already on the same device.

### Impact

This fix cascades to all models inheriting from `LlamaRotaryEmbedding` (Mistral, Ministral3, Qwen2, Gemma2, DeepSeek, etc.).

## Before submitting

- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request)?
- [x] Was this discussed/approved via a Github issue? Yes, #43350

### DIFF
--- a/src/transformers/models/afmoe/modeling_afmoe.py
+++ b/src/transformers/models/afmoe/modeling_afmoe.py
@@ -93,7 +93,7 @@ class AfmoeRotaryEmbedding(nn.Module):
     @dynamic_rope_update  # power user: used with advanced RoPE types (e.g. dynamic rope)
     def forward(self, x, position_ids):
         inv_freq_expanded = self.inv_freq[None, :, None].float().expand(position_ids.shape[0], -1, 1).to(x.device)
-        position_ids_expanded = position_ids[:, None, :].float()
+        position_ids_expanded = position_ids[:, None, :].float().to(x.device)
 
         device_type = x.device.type if isinstance(x.device.type, str) and x.device.type != "mps" else "cpu"
         with maybe_autocast(device_type=device_type, enabled=False):  # Force float32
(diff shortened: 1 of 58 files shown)


sasankkurnella:
I've updated `modeling_llama.py` with the fix.

To ensure consistency, I also ran:
1. `python utils/modular_model_converter.py --all` (propagated changes to ~47 modular models)
2. `python utils/check_copies.py --fix_and_overwrite` (synced docstrings/copies for models like Falcon)
3. Manually updated `kyutai_speech_to_text` which was missed by the automated tools.

**Update:** `check_repository_consistency` is now **PASSING** ✅. The PR should be ready for review!

sasankkurnella:
**Re: `tests_torch` failure on `Qwen3OmniMoe`**

The failing tests (`test_eager_matches_sdpa_inference_*` in `Qwen3OmniMoeThinkerForConditionalGenerationModelTest`) are **unrelated to this PR**.

This PR only modifies `LlamaRotaryEmbedding` and its downstream auto-generated files. The `Qwen3OmniMoe` model:
- Uses its own `Qwen3OmniMoeRotaryEmbedding` class
- - Is not touched by this PR
- - Fails with `RuntimeError: expected data_ptr to be aligned to 16 bytes` - a memory alignment issue in SDPA, not a device mismatch
I believe this test may be flaky or failing on `main`. Could a maintainer please verify and/or re-run CI?

Rocketknight1:
Not necessary as discussed in #43350!

PR #43371, opened 2026-01-20
Remove GPU tracking from TrackioCallback and remove env var support

qgallouedec:
Trackio now natively supports GPU tracking, see https://github.com/gradio-app/trackio/releases/tag/trackio%400.14.0, so I suggest that we remove it from this callback.


MekkCyber (approved):
Makes sense! Thank you

qgallouedec:
I'm not sure why the CI is red, but I'm quite sure it's not related to the changes in this PR, so I'll go ahead and merge. If I'm wrong don't be mad and forgive me <3

MekkCyber:
Yes the CI is broken these days! But in any case if you are not a core maintainer you can't merge with red CI 😅 , auto-merge only works in case the CI is green

SunMarc (approved):
Nice, thanks for this PR !

PR #43374, opened 2026-01-20
Fix t5 failures

Abdennacer-Badaoui:
Fix all the remaining T5 failures. 

- The test `test_export_t5_summarization` was failing because `eos_token_id` was not being passed to GenerationConfig in `Seq2SeqLMExportableModule`, causing the generate loop to never stop at the end-of-sequence token and produce extra tokens. 
- The `test_fp16_fp32_conversion` test was failing because T5's `_keep_in_fp32_modules = ["wo"]` was unconditionally forcing the `wo` layer to FP32 for both FP16 and BF16 loading, so I created a centralized `_build_dtype_plan_for_loading()` method that encapsulates all the `dtype_plan` building logic in one place, taking the target dtype as input and properly applying `_keep_in_fp32_modules` (FP16 only) and `_keep_in_fp32_modules_strict` (FP16 and BF16) based on the actual loading `dtype`.
- Fix some expectations (tested for both AMD and Nvidia)

### DIFF
--- a/src/transformers/integrations/executorch.py
+++ b/src/transformers/integrations/executorch.py
@@ -879,6 +879,7 @@ class Seq2SeqLMExportableModule(torch.nn.Module):
                 "batch_size": batch_size,
                 "max_cache_len": max_cache_length,
             },
+            eos_token_id=model.config.eos_token_id,
         )
         self.exported_encoder = None
         self.exported_decoder = None
--- a/src/transformers/integrations/executorch.py
+++ b/src/transformers/integrations/executorch.py
@@ -879,7 +879,7 @@ class Seq2SeqLMExportableModule(torch.nn.Module):
                 "batch_size": batch_size,
                 "max_cache_len": max_cache_length,
             },
-            eos_token_id=model.config.eos_token_id,
+            eos_token_id=model.generation_config.eos_token_id,
         )
         self.exported_encoder = None
         self.exported_decoder = None
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -1338,6 +1338,21 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         self.init_weights()
         self._backward_compatibility_gradient_checkpointing()
 
+    def _build_dtype_plan_for_loading(self, dtype: torch.dtype) -> dict:
+        """
+        Builds dtype_plan based on target dtype and keep_in_fp32 module settings.
+        """
+
+        # Add _keep_in_fp32_modules only for FP16 loading
+        if isinstance(self._keep_in_fp32_modules, list) and dtype == torch.float16:
+            self.dtype_plan.update(dict.fromkeys(self._keep_in_fp32_modules, torch.float32))
+
+        # Add _keep_in_fp32_modules_strict for both FP16 and BF16
+        if isinstance(self._keep_in_fp32_modules_strict, list) and dtype in (torch.float16, torch.bfloat16):
+            self.dtype_plan.update(dict.fromkeys(self._keep_in_fp32_modules_strict, torch.float32))
+
+        return self.dtype_plan
+
     @property
     def tp_plan(self) -> dict[str, str]:
         """
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -1338,21 +1338,6 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         self.init_weights()
         self._backward_compatibility_gradient_checkpointing()
 
-    def _build_dtype_plan_for_loading(self, dtype: torch.dtype) -> dict:
-        """
-        Builds dtype_plan based on target dtype and keep_in_fp32 module settings.
-        """
-
-        # Add _keep_in_fp32_modules only for FP16 loading
-        if isinstance(self._keep_in_fp32_modules, list) and dtype == torch.float16:
-            self.dtype_plan.update(dict.fromkeys(self._keep_in_fp32_modules, torch.float32))
-
-        # Add _keep_in_fp32_modules_strict for both FP16 and BF16
-        if isinstance(self._keep_in_fp32_modules_strict, list) and dtype in (torch.float16, torch.bfloat16):
-            self.dtype_plan.update(dict.fromkeys(self._keep_in_fp32_modules_strict, torch.float32))
-
-        return self.dtype_plan
-
     @property
     def tp_plan(self) -> dict[str, str]:
         """
--- a/tests/models/t5/test_modeling_t5.py
+++ b/tests/models/t5/test_modeling_t5.py
@@ -1431,8 +1431,8 @@ class T5ModelIntegrationTests(unittest.TestCase):
     def test_compile_static_cache(self):
         NUM_TOKENS_TO_GENERATE = 40
         EXPECTED_TEXT_COMPLETION = [
-            "theory of relativity states that 1) the speed of light is constant in all inertial reference frames. the laws of physics are the same for all inertial reference frames.",
-            "ketchup is my favorite condiment.",
+            "theory of relativity states that 1) the speed of light is constant in all inertial reference frames . the laws of physics are the same for all inertial reference frames .",
+            "ketchup is my favorite condiment .",
         ]
 
         prompts = [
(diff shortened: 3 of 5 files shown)


vasqu:
This is weird because we explicitly have `_keep_in_fp32_modules` and `_keep_in_fp32_modules_strict` - the later being for bf16 as well, the former only for fp16 

Something must have changed for loading models

Abdennacer-Badaoui:
That's true. I think I found the problem. 
So the `dtype_plan` initialization in `PreTrainedModel.__init__` was unconditionally adding `_keep_in_fp32_modules` entries (intended for FP16 only) even when loading with BF16, so I will fix the core loading logic in `modeling_utils.py` to only add `_keep_in_fp32_modules` to `dtype_plan` when `dtype == torch.float16`, respecting the intended distinction between `_keep_in_fp32_modules` (FP16 only) and `_keep_in_fp32_modules_strict` (FP16 and BF16).

Abdennacer-Badaoui:
I created a centralized `_build_dtype_plan_for_loading()` method that encapsulates all the` dtype_plan` building logic in one place, taking the target `dtype` as input and        properly applying `_keep_in_fp32_modules` (FP16 only) and `_keep_in_fp32_modules_strict` (FP16 and BF16) based on the actual loading `dtype`, replacing my previous fragmented logic.

src/transformers/integrations/executorch.py
```diff
@@ -879,6 +879,7 @@ def __init__(
                 "batch_size": batch_size,
                 "max_cache_len": max_cache_length,
             },
+            eos_token_id=model.config.eos_token_id,
```
ydshieh: I am not sure why `Seq2SeqLMExportableModule` doesn't use something similar to other place like `generation_config = model.generation_config`.
(maybe @vasqu or @zucchini-nlp could say sth here)

For the change here, I am also not sure `model.config` is the best thing to use, or it should use `model.generation_config` instead.

And one more thing: if we change here, would other tests will be affected. I guess no as I only see only T5 test file use `Seq2SeqLMExportableModule`.

ydshieh:
> The test_fp16_fp32_conversion test was failing because T5's _keep_in_fp32_modules = ["wo"] was unconditionally forcing the wo layer to FP32 for both FP16 and BF16 loading (even though BF16 is numerically stable enough and doesn't need FP32), so we added a post_init() override in T5PreTrainedModel that checks torch.get_default_dtype() and removes wo from dtype_plan when loading with bfloat16, allowing wo to stay in FP32 only for FP16 loading while using BF16 for BF16 loading.

What kind of failures we have here? I mean could we just adjust the test logic (if that makes sense)? To change the logic in `modeling_utils.py`, it's best to make good and precise definitions of what those flags should do. So my question is: do the currently logic (on `main`) around what you changed in `modeling_utils.py` is wrong itself, or the change is mainly to pass the test (which has a flaw testing logic and to be adjusted)?

tests/models/t5/test_modeling_t5.py
```diff
@@ -1033,7 +1033,7 @@ def test_small_v1_1_integration_test(self):
         loss = model(input_ids.to(torch_device), labels=labels.to(torch_device)).loss
         mtf_score = -(labels.shape[-1] * loss.item())
 
-        EXPECTED_SCORE = -59.0293
+        EXPECTED_SCORE = -40.1645
```
ydshieh, line 1041: Let's try to use more `Expectation` class.

BTW, do you know why such difference?

ydshieh: For below line

> self.assertTrue

let's avoid such usage, but `torch.testing.assert_close`

ydshieh: it seems failing since Nov 14 2025. Is this failing since this test is added, or there is some PRs fails it as some point while it passed before ?

when we see or make such big change, it's better to check and give more context, so we are sure we are not doing something wrong.

Abdennacer-Badaoui: I think it’s difficult to track the origin of what led to this change. There are many factors that could cause the expected scores or text completions to change. However, I checked that the new value has been consistent for a good period of time.

src/transformers/integrations/executorch.py
Abdennacer-Badaoui: You're right. It's better to use `model.generation_config`, I'll do it. 
Yes, only T5 uses `Seq2SeqLMExportableModule`.

Abdennacer-Badaoui:
> > The test_fp16_fp32_conversion test was failing because T5's _keep_in_fp32_modules = ["wo"] was unconditionally forcing the wo layer to FP32 for both FP16 and BF16 loading (even though BF16 is numerically stable enough and doesn't need FP32), so we added a post_init() override in T5PreTrainedModel that checks torch.get_default_dtype() and removes wo from dtype_plan when loading with bfloat16, allowing wo to stay in FP32 only for FP16 loading while using BF16 for BF16 loading.
> 
> What kind of failures we have here? I mean could we just adjust the test logic (if that makes sense)? To change the logic in `modeling_utils.py`, it's best to make good and precise definitions of what those flags should do. So my question is: do the currently logic (on `main`) around what you changed in `modeling_utils.py` is wrong itself, or the change is mainly to pass the test (which has a flaw testing logic and to be adjusted)?



The current logic in `modeling_utils.py` on main seems to be wrong, it doesn't respect the documented distinction between `_keep_in_fp32_modules` and `_keep_in_fp32_modules_strict`.
   According to the documentation, `_keep_in_fp32_modules` should "avoid casting to anything other than float32, except bfloat16" (only prevent FP16 casting), while `_keep_in_fp32_modules_strict` is meant to prevent both FP16 and BF16 casting. However, the current implementation in `__init__` unconditionally adds both  
  to `dtype_plan` during initialization, before knowing the target dtype, which means `_keep_in_fp32_modules` incorrectly forces modules to FP32 even for BF16 loading. This causes T5's `wo` layer to stay in FP32 when loading with BF16 (when using `_keep_in_fp32_modules = ["wo"]`). The test ( `tests/models/t5/test_modeling_t5.py::T5ModelFp16Tests::test_fp16_fp32_conversion` ) is correctly catching this bug by expecting wo to be BF16 when loading with dtype=torch.bfloat16.

src/transformers/integrations/executorch.py
vasqu: Yes, it makes sense to set via `generation_config` - note that generate is always custom overwritten for executorch related models, e.g. the encoder-decoder overwrite is here https://github.com/huggingface/transformers/blob/de306e8e14672dd8392b4bd344054a6a18de8613/src/transformers/integrations/executorch.py#L967-L1000

It's always as simple as possible to mimic just simple `do_sample=False`, the other variants seem to rely on max length for example: https://github.com/huggingface/transformers/blob/de306e8e14672dd8392b4bd344054a6a18de8613/src/transformers/integrations/executorch.py#L621

vasqu: Note https://github.com/huggingface/transformers/blob/de306e8e14672dd8392b4bd344054a6a18de8613/src/transformers/integrations/executorch.py#L996-L998 which needs the proper eos token

vasqu:
Re: keeping modules in fp32, see my comment https://github.com/huggingface/transformers/pull/43374#issuecomment-3774271277

I'm pretty sure this behavior is wrong and was only recently changed - iirc @eustlb added the strict flag, maybe cc @Cyrilvallez for model loading?

ArthurZucker (commented):
I think the change is just that now `dtype="auto"` by default, which will follow the `config` and if no config IDK. We should try to just for fp32 in the test model lkoading imo

src/transformers/modeling_utils.py
```diff
@@ -1362,6 +1359,21 @@ def post_init(self):
...
+        # Add _keep_in_fp32_modules only for FP16 loading
+        if isinstance(self._keep_in_fp32_modules, list) and dtype == torch.float16:
+            self.dtype_plan.update(dict.fromkeys(self._keep_in_fp32_modules, torch.float32))
+
+        # Add _keep_in_fp32_modules_strict for both FP16 and BF16
+        if isinstance(self._keep_in_fp32_modules_strict, list) and dtype in (torch.float16, torch.bfloat16):
+            self.dtype_plan.update(dict.fromkeys(self._keep_in_fp32_modules_strict, torch.float32))
+
+        return self.dtype_plan
+
```
Abdennacer-Badaoui: I see @ArthurZucker 
I actually added a function that correctly builds the dtype_plan during loading. Let me know what you think. Otherwise, I can revert it and just update the test.

Cyrilvallez (commented):
Opened https://github.com/huggingface/transformers/pull/43683 (will merge asap when the tests finish running) for the dtype issue because I wanted to add very strict tests on this!
The rest LGTM, let's simply revert that part on pull main! Sorry for the inconvenience 😬

Cyrilvallez (approved):
Alright LGTM! Thanks a lot!

Cyrilvallez:
Force merging as failures are completely unrelated!

PR #43379, opened 2026-01-20
Fix tiles mixing for batched input, add tie_word_embeddings to LFM2VL config

ankke:
Fixes: For batched input with multiple images tiled input was mixed between batches.Now, using a util function.
Adds: `tie_word_embeddings` to Lfm2VlConfig. Tests for above bug + updated integration test for the newest lfm2vl model.

cc: @zucchini-nlp


src/transformers/models/lfm2_vl/configuration_lfm2_vl.py
```diff
@@ -87,6 +88,7 @@ def __init__(
 
         self.vision_config = vision_config
         self.text_config = text_config
+        self.tie_word_embeddings = getattr(text_config, "tie_embedding", tie_word_embeddings)
```
zucchini-nlp, line 93: oh right. Interesting that the integration tests didn't complain though

tests/models/lfm2_vl/test_image_processing_lfm2_vl.py
```diff
@@ -831,3 +831,132 @@ def test_batch_with_tiling(self):
...
+        # All patches should have zero red, zero green, high blue
+        for i in range(num_patches_per_image, 2 * num_patches_per_image):
+            first_patch = pixel_values[i][0].view(3, patch_size, patch_size)
+            red_mean = first_patch[0].float().mean().item()
+            blue_mean = first_patch[2].float().mean().item()
+            self.assertGreater(
+                blue_mean, red_mean, f"Blue image patch {i} has more red than blue - patches may be interleaved"
+            )
+
+    def test_batch_tiling_with_thumbnail(self):
```
zucchini-nlp: nit: this test is same as prev but uses thumbnail. We can decorate it with `parameterized` and allow `thumbnail=True/False`

tests/models/lfm2_vl/test_modeling_lfm2_vl.py
```diff
@@ -234,7 +234,7 @@ def test_training_gradient_checkpointing_use_reentrant_true(self):
 @slow
 class Lfm2VlForConditionalGenerationIntegrationTest(unittest.TestCase):
     def setUp(self):
-        self.processor = AutoProcessor.from_pretrained("LiquidAI/LFM2-VL-1.6B")
+        self.processor = AutoProcessor.from_pretrained("LiquidAI/LFM2.5-VL-1.6B")
```
zucchini-nlp, line 320: no clear why we use a new released ckpt? Is the Lfm-VL v2 removed from the hub and not used anymore?

src/transformers/models/lfm2_vl/configuration_lfm2_vl.py
ankke: They were failing for me without this field

zucchini-nlp: ah prob they were failing in `main` branch. We usually don't check why the test failed if it is broken already on `main`, which I think we'll work on improving

tests/models/lfm2_vl/test_modeling_lfm2_vl.py
ankke: added as separate test case

zucchini-nlp (approved):
Thanks, I will trigger slow tests and check

zucchini-nlp:
run-slow: lfm2_vl

zucchini-nlp:
I can't push to this branch apparently, can you update the docstring @ankke and add `tie_word_embeddings`?

zucchini-nlp (approved):
Thanks, merging!

zucchini-nlp:
run-slow: lfm2_vl

PR #43380, opened 2026-01-20
fix(models) : string mutation bug in from_proteinnet_string

aarushisingh04:
### what does this PR do?

fixes a `TypeError` in `openfold_utils/protein.py` where `from_proteinnet_string` attempts to mutate an immutable string in-place.

the original code tried to assign seq[i] = "X" to handle unknown residues, which crashes. this pr replaces that logic with a list comprehension to safely sanitize the sequence, resolving the existing `FIXME`.

### before submitting
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).

### who can review?

@ArthurZucker

### DIFF
--- a/src/transformers/models/esm/openfold_utils/protein.py
+++ b/src/transformers/models/esm/openfold_utils/protein.py
@@ -81,9 +81,8 @@ def from_proteinnet_string(proteinnet_str: str) -> Protein:
     for g in groups:
         if g[0] == "[PRIMARY]":
             seq = g[1][0].strip()
-            for i in range(len(seq)):
-                if seq[i] not in residue_constants.restypes:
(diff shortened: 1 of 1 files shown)


Rocketknight1 (approved):
Yup, makes sense! I doubt this code path was ever taken, but the fix is clean and obviously correct so we might as well include it.

PR #43389, opened 2026-01-21
Fix always issuing generation deprecation warning

githubnemo:
# What does this PR do?

**Note:** already discussed with @zucchini-nlp internally

Commit 38e5987 from PR #43194 introduced a minor bug that causes the issuing of a deprecation warning regardless of the user using the deprecated interface.

For example:
```
model.generate(**input_ids, do_sample=False)
```
would issue a deprecation warning not to pass `GenerationConfig` as well as generation kwargs even though we're not passing a `GenerationConfig`.

Along the fix this also introduces a unit test to check both the intended and the bug behavior.

### DIFF
--- a/tests/generation/test_utils.py
+++ b/tests/generation/test_utils.py
@@ -2691,6 +2691,46 @@ class GenerationIntegrationTests(unittest.TestCase):
             model.generation_config.use_cache = None
             model.save_pretrained(tmpdirname)
 
+    def test_generation_config_deprecation(self):
+        import logging as pylogging
+
+        model = AutoModelForCausalLM.from_pretrained("hf-internal-testing/tiny-random-gpt2").to(torch_device)
+        tokenizer = AutoTokenizer.from_pretrained("hf-internal-testing/tiny-random-gpt2")
+        input_ids = tokenizer("Hello", return_tensors="pt").input_ids.to(torch_device)
+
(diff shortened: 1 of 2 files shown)


tests/generation/test_utils.py
```diff
@@ -2691,6 +2691,46 @@ def test_generation_config_defaults(self):
...
+                if "Passing `generation_config` together with" in msg:
+                    self.warnings.append(msg)
+
+        warningHandler = OnWarningHandler()
+        logger.addHandler(warningHandler)
+
+        try:
+            # Providing generation_config and kwargs is deprecated, so we expect a warning here.
+            #
+            # We're using logging_once, make sure that we emit this warning in any case by clearing the
```
zucchini-nlp, line 2719: nit: what is the purpose of `try`, and can we let it fail explicitly to know that this test has issues instead of silent skip

zucchini-nlp (approved):
Thanks, one tiny question and then feel free to merge :)

tests/generation/test_utils.py
githubnemo: `try` without `except` doesn't catch the exceptions, it executes the code in `try` and then invokes `finally` to clean up - that's the only purpose of this construct is to remove the logging handler after the test is done in any case (exception thrown because of assertion failure or not)

PR #43392, opened 2026-01-21
pin `torch==2.9.1` for now

ydshieh:
# What does this PR do?

Apparently, the new release has some issues on our CI.

### DIFF
--- a/docker/consistency.dockerfile
+++ b/docker/consistency.dockerfile
@@ -5,7 +5,7 @@ ARG REF=main
 RUN apt-get update && apt-get install -y time git g++ pkg-config make git-lfs
 ENV UV_PYTHON=/usr/local/bin/python
 RUN pip install uv && uv pip install --no-cache-dir -U pip setuptools GitPython
-RUN uv pip install --no-cache-dir --upgrade 'torch' 'torchaudio' 'torchvision' --index-url https://download.pytorch.org/whl/cpu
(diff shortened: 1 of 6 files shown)


Rocketknight1 (approved):

.circleci/create_circleci_config.py
```diff
@@ -21,6 +21,7 @@
 
 import yaml
 
+a = 1
```
vasqu: Not to forget

docker/consistency.dockerfile
```diff
@@ -5,7 +5,7 @@ ARG REF=main
 RUN apt-get update && apt-get install -y time git g++ pkg-config make git-lfs
 ENV UV_PYTHON=/usr/local/bin/python
 RUN pip install uv && uv pip install --no-cache-dir -U pip setuptools GitPython
-RUN uv pip install --no-cache-dir --upgrade 'torch' 'torchaudio' 'torchvision' --index-url https://download.pytorch.org/whl/cpu
+RUN uv pip install --no-cache-dir --upgrade 'torch==2.9.1' 'torchaudio' 'torchvision' --index-url https://download.pytorch.org/whl/cpu
```
vasqu, line 8: Maybe we can do < 2.10 and raise from time to time instead?

vasqu (approved):
Thank you

docker/consistency.dockerfile
ydshieh: yeah, i haven't spend time on think a better way, like having some flag  in a single place that would make the change more easily. Add to TODO however. Just want have this done quickly :-) 🙏

PR #43396, opened 2026-01-21
chore(typing): initial `ty` integration

tarekziade:
# What does this PR do?

Initial `ty` integration. To avoid a gigantic, risky patch, let's start with a baby step where we add the tooling to `make repo-check` and activate it on a subset of the repo. 

That gives us a human-readable patch, and allows us to get confortable with the tool before we expand it.

This patch:

- adds `ty` to `make repo-check` and the CI
- runs only against `src/transformers/utils`
- has a large list of ignores in its configuration for now, that we will loosen later

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


src/transformers/utils/backbone_utils.py
```diff
@@ -137,8 +137,33 @@ def get_aligned_output_features_output_indices(
...
+
+
+class _TimmBackbone(Protocol):
+    feature_info: _TimmFeatureInfo
+
+
+class _SupportsToDict(Protocol):
+    """Protocol for classes that have a to_dict method returning a dictionary."""
+
+    def to_dict(self) -> dict[str, Any]: ...
```
Cyrilvallez: Yes I also think this is too much. Especially since we already started deprecating the backbone API with https://github.com/huggingface/transformers/pull/43323, so let's not add to it!

src/transformers/utils/chat_template_utils.py
```diff
@@ -393,6 +397,7 @@ def _compile_jinja_template(chat_template):
         raise ImportError(
             "apply_chat_template requires jinja2 to be installed. Please install it using `pip install jinja2`."
         )
+    assert jinja2 is not None
```
Cyrilvallez: Do we have to add those asserts? 🤔
What's their purpose?

src/transformers/utils/chat_template_utils.py
```diff
@@ -405,16 +410,17 @@ def __init__(self, environment: ImmutableSandboxedEnvironment):
...
             body = parser.parse_statements(["name:endgeneration"], drop_needle=True)
-            return jinja2.nodes.CallBlock(self.call_method("_generation_support"), [], [], body).set_lineno(lineno)
+            return jinja2.nodes.CallBlock(self.call_method("_generation_support"), [], [], body).set_lineno(lineno)  # type: ignore[attr-defined]
 
         @jinja2.pass_eval_context
         def _generation_support(self, context: jinja2.nodes.EvalContext, caller: jinja2.runtime.Macro) -> str:
             rv = caller()
             if self.is_active():
                 # Only track generation indices if the tracker is active
+                assert self._rendered_blocks is not None and self._generation_indices is not None
```
Cyrilvallez: Same here and another one below

src/transformers/utils/generic.py
```diff
@@ -897,7 +897,7 @@ def wrapper(self, *args, **kwargs):
                     # Arg-specific handling
                     if arg_name == "use_cache":
                         if getattr(self, "gradient_checkpointing", False) and self.training and arg_value:
-                            logger.warning_once(
```
Cyrilvallez, line 902: We have a lot of `warning_once` everywhere in the lib... If `ty` complains about them, is there anything we can do in the `toml` config directly to silence it? We will have A LOT of `# type: ignore[attr-defined]` otherwise...

src/transformers/utils/doc.py
```diff
@@ -1091,6 +1091,7 @@ def copy_func(f):
     """Returns a copy of a function f."""
     # Based on http://stackoverflow.com/a/6528148/190597 (Glenn Maynard)
     g = types.FunctionType(f.__code__, f.__globals__, name=f.__name__, argdefs=f.__defaults__, closure=f.__closure__)
-    g = functools.update_wrapper(g, f)
-    g.__kwdefaults__ = f.__kwdefaults__
-    return g
+    wrapped = functools.update_wrapper(g, f)
+    if hasattr(f, "__kwdefaults__"):
+        setattr(wrapped, "__kwdefaults__", f.__kwdefaults__)
```
Cyrilvallez: It requires an explicit `setattr` here, or simply the existence `hasattr` check? 🤔

src/transformers/utils/hub.py
```diff
@@ -168,7 +168,8 @@ def define_sagemaker_information():
 
     sagemaker_params = json.loads(os.getenv("SM_FRAMEWORK_PARAMS", "{}"))
     runs_distributed_training = "sagemaker_distributed_dataparallel_enabled" in sagemaker_params
-    account_id = os.getenv("TRAINING_JOB_ARN").split(":")[4] if "TRAINING_JOB_ARN" in os.environ else None
+    training_job_arn = os.getenv("TRAINING_JOB_ARN")
+    account_id = training_job_arn.split(":")[4] if training_job_arn is not None else None
```
Cyrilvallez, line 172: What's the reason of `ty` complaining on this one?

src/transformers/utils/import_utils.py
```diff
@@ -534,10 +540,12 @@ def enable_tf32(enable: bool) -> None:
     pytorch_version = version.parse(get_torch_version())
     if pytorch_version >= version.parse("2.9.0"):
         precision_mode = "tf32" if enable else "ieee"
-        torch.backends.fp32_precision = precision_mode
+        if hasattr(torch.backends, "fp32_precision"):
+            torch.backends.fp32_precision = precision_mode
```
Cyrilvallez, line 544: It seem to require a lot of existence checks. Do you know how they are chosen to be required or not? Does `ty` base itself e.g. on your currently installed version of `torch` in the env? Because that may become an issue

src/transformers/utils/notebook.py
```diff
@@ -307,6 +307,7 @@ def on_train_begin(self, args, state, control, **kwargs):
 
     def on_step_end(self, args, state, control, **kwargs):
         epoch = int(state.epoch) if int(state.epoch) == state.epoch else f"{state.epoch:.2f}"
+        assert self.training_tracker is not None, "on_train_begin must be called before on_step_end"
```
Cyrilvallez: Same here, we generally don't want asserts

src/transformers/utils/quantization_config.py
```diff
@@ -1254,7 +1257,8 @@ def is_quantized(self):
     def is_quantization_compressed(self):
         from compressed_tensors.quantization import QuantizationStatus
 
-        return self.is_quantized and self.quantization_config.quantization_status == QuantizationStatus.COMPRESSED
+        qc = self.quantization_config
+        return qc is not None and bool(qc.config_groups) and qc.quantization_status == QuantizationStatus.COMPRESSED
```
Cyrilvallez: Here it's not the same return as before

Cyrilvallez (commented):
ALright! Happy to start with `ty`! Asked quite a few questions in comments more for my understanding and get familiar with ty than anything else

src/transformers/utils/backbone_utils.py
tarekziade: Yes I was not sure about this, happy to follow that rule of not adding protocols

src/transformers/utils/generic.py
tarekziade: One option would be to implement the logger class for real, and not monkey patch it.

e.g.

```
class TransformersLogger(Logger):
    def warning_once(self, xxx):
        xxx
```

src/transformers/utils/chat_template_utils.py
tarekziade: weird left over

src/transformers/utils/chat_template_utils.py
tarekziade: technically it could be called outside the context manager. I suspect we could extract the `AssistantTracker` class from that function and clean it up a bit to remove that

tarekziade: triggered by `possibly-missing-attribute`

src/transformers/utils/doc.py
tarekziade: that is triggered by 

```
Unresolved attribute `__kwdefaults__` on type `_Wrapped[(...), Unknown, (...), Unknown]
```

because `update_wrapper` returns a  `_Wrapped` 

A better fix is:

```
diff --git a/src/transformers/utils/doc.py b/src/transformers/utils/doc.py
index a2113fcf29..0521d346d9 100644
--- a/src/transformers/utils/doc.py
+++ b/src/transformers/utils/doc.py
@@ -1091,7 +1091,6 @@ def copy_func(f):
     """Returns a copy of a function f."""
     # Based on http://stackoverflow.com/a/6528148/190597 (Glenn Maynard)
     g = types.FunctionType(f.__code__, f.__globals__, name=f.__name__, argdefs=f.__defaults__, closure=f.__closure__)
-    wrapped = functools.update_wrapper(g, f)
-    if hasattr(f, "__kwdefaults__"):
-        setattr(wrapped, "__kwdefaults__", f.__kwdefaults__)
-    return wrapped
+    g = cast(types.FunctionType, functools.update_wrapper(g, f))
+    g.__kwdefaults__ = f.__kwdefaults__
+    return g
```

src/transformers/utils/hub.py
tarekziade: `possibly-missing-attribute`because `os.getenv` can return None and `ty` can't know that the `in os.environ` check is similar. I think it's a pretty reasonable tradeoff, we remove that uncertainty by storing the getenv() value

src/transformers/utils/import_utils.py
tarekziade: In this very specific case `ty` can't infere the existence of this attributes:

  - At runtime, torch.backends module replaces itself with a GenericModule instance (line 122 in torch/backends/__init__.py)
  - GenericModule defines fp32_precision as a ContextProp descriptor (lines 116-119)
  - However, there's no .pyi stub file for torch.backends, so type checkers can't see that the attribute exists

>  Does ty base itself e.g. on your currently installed version of torch in the env?

yes it uses its stubs

src/transformers/utils/notebook.py
tarekziade: You are making the assumption that the `on_train_begin` is always called prior to `on_step_end` but there's no guarantee other than the caller behaving that way. so `ty` detects it which makes sense. What would you suggest?

src/transformers/utils/quantization_config.py
tarekziade: what should it return in case `self.quantization_config` is `None` ? 

or is this good enough?

```
return self.is_quantized and (qc is not None and qc.quantization_status == QuantizationStatus.COMPRESSED)
```

tarekziade:
> ALright! Happy to start with `ty`! Asked quite a few questions in comments more for my understanding and get familiar with ty than anything else

Thanks for the first pass!

one remark about the `assert is not None` checks:

it's all the same pattern:

- a class has an attribute it sets to `None` in its constructor
- a method `B` makes the assumption is has another value than `None`, set by another method `A`

So it makes sense for a type checker to detect that one possible code execution pass is to call `B` with no previous call to `A`. Adding assertions like this is the simplest non intrusive fix. With no assertions, we need to refactor the class so calls to `B` can't happen before calls to `A` by design.

tarekziade:
run-slow: dia, whisper

tarekziade:
supersed by https://github.com/huggingface/transformers/pull/44167
(which is a branch in the same repo for convenency)

Issue #43334, opened 2026-01-17
Qwen3-VL can't be loaded in transformers dev: AttributeError: 'Qwen3VLTextConfig' object has no attribute 'pad_token_id'


albertvillanova:
Qwen3-VL can't be loaded in transformers dev branch (5.0.0.dev0 at commit 24807bfcf4a21286fa2a7e728f381ddaaca7bbc7): 
> AttributeError: 'Qwen3VLTextConfig' object has no attribute 'pad_token_id'

After investigation, the issue appeared after the merge of this PR:
- https://github.com/huggingface/transformers/pull/41541

As @zucchini-nlp pointed out, there are no slow tests for `qwen3VL`, which could have prevented this.

See related issue in TRL:
- https://github.com/huggingface/trl/issues/4842

### Reproduction

```python
from transformers import AutoModel

AutoModel.from_pretrained('Qwen/Qwen3-VL-2B-Instruct')
```

Traceback:
```python
  File "<string>", line 1, in <module>
  File "transformers/src/transformers/models/auto/auto_factory.py", line 372, in from_pretrained
    return model_class.from_pretrained(
  File "transformers/src/transformers/modeling_utils.py", line 4020, in from_pretrained
    model = cls(config, *model_args, **model_kwargs)
  File "transformers/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py", line 938, in __init__
    self.language_model = Qwen3VLTextModel._from_config(config.text_config)
  File "transformers/src/transformers/modeling_utils.py", line 1464, in _from_config
    model = cls(config, **kwargs)
  File "transformers/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py", line 807, in __init__
    self.padding_idx = config.pad_token_id
  File "transformers/src/transformers/configuration_utils.py", line 164, in __getattribute__
    return super().__getattribute__(key)
AttributeError: 'Qwen3VLTextConfig' object has no attribute 'pad_token_id'. Did you mean: 'bos_token_id'?
```

### Who can help?

@zucchini-nlp, who created the PR

JesseChenJesse:
I tested Qwen3-VL-2B-Thinking in huggingface just now, and then occurred:
> AttributeError: 'Qwen3VLTextConfig' object has no attribute 'pad_token_id'. Did you mean: 'bos_token_id'? 

then i tried to see model file, and copy line `"pad_token_id": 151643,` in file "generation_config.json" to "config.json", i found it run.
[...]

PR #43398, opened 2026-01-22
Fix: adding pad_token_id in Qwen3VLTextConfig

### LINKED ISSUE #43393: Qwen3-VL checkpoints don't have pad_token
`AutoModelForImageTextToText.from_pretrained("Qwen/Qwen3-VL-2B-Instruct")` fails for me because of the line `self.padding_idx = config.pad_token_id`, but the checkpoints do not have `pad_token_id` set and the configs don't have a default value for it. Should we update the checkpoints or change that line?

cc @zucchini-nlp

vaibhav-research:
# What does this PR do?

This PR fixes an initialization error in Qwen3-VL text model construction when loading checkpoints whose nested text_config does not explicitly define pad_token_id.

there are  2 main issues
	•	Qwen3VLTextModel.__init__ unconditionally accesses config.pad_token_id, but Qwen3VLTextConfig did not define this attribute.
	•	When loading via from_pretrained, this resulted in an AttributeError during model initialization.

The fix adds pad_token_id to Qwen3VLTextConfig (defaulting to None) and forwards it through PreTrainedConfig, aligning the config schema with what the model expects. 
A regression test is added to ensure Qwen3VLTextModel can be instantiated when pad_token_id is not provided.

Fixes #43393 #43334 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@Rocketknight1 @zucchini-nlp 

### DIFF
--- a/src/transformers/models/qwen3_vl/modular_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/modular_qwen3_vl.py
@@ -213,10 +213,10 @@ class Qwen3VLTextConfig(PreTrainedConfig):
         self.attention_bias = attention_bias
         self.attention_dropout = attention_dropout
         self.rope_parameters = rope_parameters
+        self.pad_token_id = pad_token_id
 
         super().__init__(
             ignore_keys_at_rope_validation={"mrope_section", "mrope_interleaved"},
-            pad_token_id=pad_token_id,
             **kwargs,
         )
 
--- a/src/transformers/models/qwen3_vl/modular_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/modular_qwen3_vl.py
@@ -153,7 +153,7 @@ class Qwen3VLTextConfig(PreTrainedConfig):
             Whether to use a bias in the query, key, value and output projection layers during self-attention.
         attention_dropout (`float`, *optional*, defaults to 0.0):
             The dropout ratio for the attention probabilities.
-        pad_token_id (`int`, *optional*, defaults to `None`):
+        pad_token_id (`int`, *optional*):
             The id of the padding token. If unset, the config is treated as not having a dedicated padding token.
 
     ```python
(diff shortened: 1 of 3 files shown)


tests/models/qwen3_vl/test_modeling_qwen3_vl.py
```diff
@@ -190,6 +192,34 @@ def setUp(self):
     def test_config(self):
         self.config_tester.run_common_tests()
 
+    def test_nested_text_config_missing_pad_token_id(self):
+        """
+        Mirrors the from_pretrained failure when loading a Qwen3VLModel:
+        a text_config dict loaded from config.json may omit pad_token_id, but should still
+        results in a Qwen3VLTextConfig with pad_token_id=None and allow Qwen3VLModel init.
+        """
+        config = self.model_tester.get_config()  # Qwen3VLConfig
```
zucchini-nlp: no test needed, we didn't see the bug before merging the PR because of the absence of slow integration tests. Can you revert diff on this file?

src/transformers/models/qwen3_vl/modular_qwen3_vl.py
```diff
@@ -213,6 +216,7 @@ def __init__(
 
         super().__init__(
             ignore_keys_at_rope_validation={"mrope_section", "mrope_interleaved"},
+            pad_token_id=pad_token_id,
```
zucchini-nlp: nit: let's explicitly assign it as `self.pad_token_id = pad_token_id`. The `super` doesn't have defaults for `token_ids` anymore and we don't need to pass it here

tests/models/qwen3_vl/test_modeling_qwen3_vl.py
vaibhav-research: sure, I will revert.

src/transformers/models/qwen3_vl/modular_qwen3_vl.py
vaibhav-research: sure, I will more it out of `super` block

vaibhav-research:
@zucchini-nlp thanks for the review. I updated the PR based on your feedback. 
reverted all changes in tests/models/qwen3_vl/test_modeling_qwen3_vl.py per your suggestion. Also updated Qwen3VLTextConfig to explicitly set self.pad_token_id = pad_token_id and removed passing it to super().__init__.

zucchini-nlp (approved):
Great, thanks for iterating. Merging!

PR #43403, opened 2026-01-22
handle 1D position_ids for modeling_flash_attention_utils as well

kaixuanliu:
This PR fixes failed test case of 
`pytest -rA tests/models/lighton_ocr/test_modeling_lighton_ocr.py::LightOnOcrForConditionalGenerationModelTest::test_eager_matches_fa2_generate`, in this case `position_ids` is concated to 1D in [position_ids_in_meshgrid](https://github.com/huggingface/transformers/blob/main/src/transformers/models/pixtral/modeling_pixtral.py#L37) func, and hence cause the crash.


kaixuanliu:
Also fix 2 failed test cases for lighton_ocr:
```
tests/models/lighton_ocr/test_modeling_lighton_ocr.py::LightOnOcrForConditionalGenerationModelTest::test_flash_attn_2_inference_equivalence
tests/models/lighton_ocr/test_modeling_lighton_ocr.py::LightOnOcrForConditionalGenerationModelTest::test_flash_attn_2_inference_equivalence_right_padding
```
where in `get_image_features` phase, `image_sizes` is None, and cause the crash.

kaixuanliu:
@ArthurZucker @Rocketknight1 Pls help review, thx! The failed CI should have nothing to do with this PR.

Rocketknight1:
Hi @kaixuanliu, given that this only affects LightOnOCR, would it be better to modify the code there (for example, adding a dummy batch dim to `position_ids` before passing them to this function) instead of needing modifications to the base class?

vasqu:
Just seeing this now, it would be indeed better to just `unsqueeze(0)` for example. Likely somewhere in the code this isn't properly done. We always expect `bsz, seq_len` and allowing only `seq_len` would go against our conventions

kaixuanliu:
@Rocketknight1 @vasqu Thx for your advice. Have updated the code. Pls have a review again.

src/transformers/models/lighton_ocr/modular_lighton_ocr.py
```diff
@@ -365,6 +365,10 @@ def forward(
             )
             inputs_embeds = inputs_embeds.masked_scatter(special_image_mask, image_features)
 
+        # Ensure position_ids has batch dimension for flash attention compatibility
+        if position_ids is not None and position_ids.ndim == 1:
+            position_ids = position_ids.unsqueeze(0)
```
vasqu: Could it be that we need to properly take care of cache pos and pos ids? https://github.com/huggingface/transformers/blob/236a8e9d9dc72f97f227d7eab6675651d50d990f/src/transformers/models/llama/modeling_llama.py#L399-L406

tests/test_modeling_common.py
```diff
@@ -3132,6 +3132,9 @@ def flash_attn_inference_equivalence(
                     if "image_grid_thw" in inputs_dict:
                         continue
                     first_inputs["pixel_values"] = inputs_dict["pixel_values"][:1].to(torch.bfloat16)
+                    # Also include image_sizes if present (needed for vision models)
+                    if "image_sizes" in inputs_dict:
+                        first_inputs["image_sizes"] = inputs_dict["image_sizes"][:1]
```
vasqu: Hmm, I'm a bit unsure about the generality here. Is it specific to lighton or do more models depend on this

vasqu (commented):
Some quick comments

tests/test_modeling_common.py
kaixuanliu: Actually I am not sure if other models depend on this... I run into this error just w/ lighton tests. Or we can fix the bug in lighton specific tests like in https://github.com/huggingface/transformers/pull/43403/changes/dfbc63326ee807a9a2f6863e83cd9ae430334efd?

vasqu: I've checked locally with #43443 and it should fix the issue, it will only need this adjustment here. But yes, let's move this to the specific model test for now

kaixuanliu: Thx, will update the code once https://github.com/huggingface/transformers/pull/43443 is merged.

vasqu:
run-slow: lighton_ocr

vasqu (approved):
Lgtm, just sanity checking with run-slow that nothing else got broken in the process

PR #43407, opened 2026-01-22
Cleanup rope conversion

vasqu:
As per title, cleanup as this overwrite was not necessary

Can be verified by running (with and without this change)
```python


config = AutoConfig.from_pretrained("upstage/Solar-Open-100B")
print(config)
```

cc @oesni for viz


vasqu:
run-slow: solar_open

ArthurZucker (approved):
ty there are probably other models that don't require this, would be nice if you can do a pass on them as well!

vasqu:
run-slow: minimax_m2

vasqu:
I've checked through all the models that overwrite it and only found minimax m2 

The other models need it to be BC with the already released models.

PR #43410, opened 2026-01-22
[`RoPE`] Make explicit inheritance

vasqu:
As per title, not every config needs this mixin - it should be only done for those that have it


vasqu:
Moving too fast, need to check bit by bit for the last failures

ArthurZucker (approved):
LGTM ty! 🤗

PR #43411, opened 2026-01-22
[`HunYuan`] Fix RoPE init

vasqu:
As per title, currently init weights assumes everything is uniform but these dynamic inits are slightly different


vasqu:
cc @xenova if you could check if it works for you, just made a quick draft for this as we talked internally about this models series

xenova (commented):
I can confirm this PR fixes the bug I encountered! 🙌

vasqu:
run-slow: hunyuan_v1_dense, hunyuan_v1_moe

vasqu:
run-slow: hunyuan_v1_dense, hunyuan_v1_moe

xenova (approved):
works for me (on the test case that caused this discussion)!

vasqu:
run-slow: hunyuan_v1_moe

PR #43418, opened 2026-01-22
Fix in-place modification of inputs_embeds in Kosmos-2.5 forward

Neuro1729:
# What does this PR do?

Fixes a PyTorch autograd error when fine-tuning `Kosmos2_5ForConditionalGeneration` with PEFT (LoRA).

The model forward pass performed an in-place assignment on `inputs_embeds`, which can be a leaf tensor with `requires_grad=True` during training. This caused the following runtime error:

RuntimeError: a leaf Variable that requires grad is being used in an in-place operation

This PR avoids the in-place modification by cloning the tensor before assignment, allowing training to proceed correctly while preserving existing behavior.

I have verified the fix using the reproduction script I provided in the issue; the RuntimeError is resolved and training proceeds as expected.

Fixes #43416

cc @zucchini-nlp

### DIFF
--- a/src/transformers/models/kosmos2_5/modeling_kosmos2_5.py
+++ b/src/transformers/models/kosmos2_5/modeling_kosmos2_5.py
@@ -1091,7 +1091,7 @@ class Kosmos2_5TextTransformer(nn.Module):
 
         # Ignore copy
         if image_embeds is not None:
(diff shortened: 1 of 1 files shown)


zucchini-nlp (approved):
Thanks!

PR #43419, opened 2026-01-22
Fix per-sample nested label truncation in gather_for_metrics

karthik-0306:
This PR fixes an issue in `nested_truncate` that affects metric gathering
for models using per-sample nested label structures such as
`tuple[list[...], ...]` (e.g. Mask2Former).

Previously, truncation during metric gathering could occur at the wrong
nesting level, leading to tensor-dimension truncation or dropped label
components in last-batch edge cases (e.g. remainder == 1).

This change fixes the root cause by:
- Correctly truncating per-sample nested structures at the list (sample) level
- Preserving tuple structure and tensor contents
- Handling last-batch edge cases safely
- Adding a focused regression test that fails on main and passes with this fix

The fix avoids introducing task-specific logic into `Trainer` and does not
impact processor outputs.

Fixes #43388  
Related to #43395

### DIFF
--- a/src/transformers/trainer_pt_utils.py
+++ b/src/transformers/trainer_pt_utils.py
@@ -363,16 +363,16 @@ def expand_like(arrays, new_seq_length, padding_index=-100):
 def nested_truncate(tensors, limit):
     """Truncate `tensors` at `limit` (even if it's a nested list/tuple/dict of tensors)."""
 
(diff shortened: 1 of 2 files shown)


Rocketknight1:
Sorry, these code agent PRs are really annoying to review - they sound useful at first but the logic is always very unclear, especially when it clashes with an existing open PR. I'm just gonna close this and we'll see if any issues remain after #43395 is merged

PR #43424, opened 2026-01-22
Add test to ensure executorch exportability with dynamic shapes

justinchuby:
# What does this PR do?

Previously the models were not exportable with dynamic shapes due to slices removed in https://github.com/huggingface/transformers/pull/41900.

This PR adds a test to ensure executorch exportability with dynamic shapes.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/tests/test_executorch.py
+++ b/tests/test_executorch.py
@@ -127,3 +127,37 @@ class ExecutorchTest(unittest.TestCase):
             inputs_embeds=self.inputs_embeds, cache_position=self.cache_position
         )
         torch.testing.assert_close(eager_output_embeds, exported_output_embeds, atol=1e-4, rtol=1e-4)
+
+    def test_decoder_only_lm_export_with_dynamic_shapes(self):
+        """Test TorchExportableModuleForDecoderOnlyLM export with dynamic sequence lengths."""
+        module = TorchExportableModuleForDecoderOnlyLM(self.model)
+
+        # Test export with input_ids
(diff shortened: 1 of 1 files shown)


tests/test_executorch.py
```diff
@@ -127,3 +127,37 @@ def test_decoder_only_lm_export(self):
             inputs_embeds=self.inputs_embeds, cache_position=self.cache_position
         )
         torch.testing.assert_close(eager_output_embeds, exported_output_embeds, atol=1e-4, rtol=1e-4)
+
+    def test_decoder_only_lm_export_with_dynamic_shapes(self):
```
vasqu, line 131: I re-added the slice ops and it still passed (torch 2.9.0) so we should make it a bit more constrained

I assume this is due to a simple forward not really changing shapes, hence the slicing doesn't matter? Or given that there is not attention mask, it isn't really triggered?

justinchuby: I see. Yes an attention mask is probably needed

justinchuby:
@vasqu even though this test does not include masks, it is still useful to guard general exportability with dynamic shapes. Would it be possible to merge this change first? I can create a follow up with more test variations.

vasqu:
@justinchuby Not opposed to it in general but I don't really see the merit in having a test that doesn't cover its original purpose? Would it be possible to extend these tests here instead?

PR #43427, opened 2026-01-23
Adjust distributed backend in tensor parallel's test file

kaixuanliu:
This PR fixes bug when `param` is a torch.Tensor, it does not have `get_shape()` method. You can reproduce this bug using:
`pytest -rA tests/tensor_parallel/test_tensor_parallel.py::TestTensorParallel2Proc::test_model_dense_forward_compile_eval`


kaixuanliu:
@CyrilVallez, can you help review? thx! The failed CI tests shoule have nothing to do with this PR.

kaixuanliu:
For cuda device, nccl backend is more widely used, and for some devices like xpu, "gloo" backend is not supported. @Cyrilvallez , can you help review? Thx!

Cyrilvallez:
Hey @kaixuanliu! All TP tests are passing, and `gloo` is valid for cuda as well. We don't run such tests on `xpu`, so I don't think there is any need to complicate the test code depending on hardware rn!
Feel free to reopen if you think otherwise!

PR #43428, opened 2026-01-23
benchmark-v2: minor fix for benchmark-v2, kernel is not correctly called

sywangyi:
kernels is not correctly called in benchmark_v2, should register before kernelize, see https://github.com/huggingface/transformers/blob/10e97cd508218546ef681a2c9b4c519ac0d927c3/src/transformers/modeling_utils.py#L3607, use 'use_kernels' in AutoModelForCausalLM.from_pretrained could handle register kernels and kernelize


sywangyi:
@remi-or , pls help review, thx very much.

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -205,15 +205,16 @@ def setup_benchmark(self, model_id: str, config: BenchmarkConfig) -> None:
         # Load model
         self.logger.debug(f"Loading model {model_id} on device {config.device}...")
         dtype = getattr(torch, config.dtype.removeprefix("torch."))
+        use_kernels = config.kernelize and kernelize is not None and Mode is not None
```
remi-or, line 208: I think this change makes little sense: we cannot control MODE and doing it this way is the same as calling `kernelize`. Unless you found a difference in practice?

benchmark_v2/framework/benchmark_config.py
```diff
@@ -40,13 +40,14 @@ def is_fa2_or_kernel_available() -> bool:
...
 
-        get_kernel("kernels-community/flash-attn")
+        get_kernel("kernels-community/flash-attn2")
     except Exception as _:
         logger.warning(
             "flash_attention_2 is not available. kernels is installed, but the flash_attn kernel is not available."
             "Benchmarking flash_attention_2 will not be possible."
         )
         return False
+    return True
```
remi-or, line 51: Good catch, thanks.

benchmark_v2/framework/benchmark_config.py
```diff
@@ -40,13 +40,14 @@ def is_fa2_or_kernel_available() -> bool:
     try:
         from kernels import get_kernel
 
-        get_kernel("kernels-community/flash-attn")
+        get_kernel("kernels-community/flash-attn2")
```
remi-or, line 44: Could you add `version=1` please? This will make things more steady.
Not sure about the exact syntax, if you can test it out it would be great. Thanks!

remi-or (commented):
Thanks for the contribution! A few nits to address but otherwise we should be good.

benchmark_v2/framework/benchmark_runner.py
sywangyi: yes, I run the benchmark test with kernelize = true. there's warning like
No kernel mapping found for layer `rotary_pos_emb`. Check if the layer name matches one of the kernels in the mapping or add the kernel you want to use to the mapping. Defaulting to original forward implementation.
  warnings.warn(
/opt/venv/lib/python3.12/site-packages/kernels/layer/layer.py:317: UserWarning:
No kernel mapping found for layer `SiLU`. Check if the layer name matches one of the kernels in the mapping or add the kernel you want to use to the mapping. Defaulting to original forward implementation.
  warnings.warn(
/opt/venv/lib/python3.12/site-packages/kernels/layer/layer.py:317: UserWarning:
No kernel mapping found for layer `RMSNorm`. Check if the layer name matches one of the kernels in the mapping or add the kernel you want to use to the mapping. Defaulting to original forward implementation.
  warnings.warn(

I did an investigation, and seems the kernel is not registered so it could not be used. reference code like https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L3651.
see https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L4430, in kernelize(), the model is set due to the model.training states, and all the above logic is contained in from_pretrained use_kernel logic.

benchmark_v2/framework/benchmark_config.py
sywangyi: there's no version_1. see https://github.com/huggingface/kernels-community, only flash-attn2 and flash-attn3

remi-or: No, I mean the `version` kwargs! You are right for the repos URL.

sywangyi: version=1 is introduced in kernels 0.12.0. see https://github.com/huggingface/kernels/releases/tag/v0.12.0
I check transformers dependence of kernels, it is "kernels>=0.10.2,<0.11"(https://github.com/huggingface/transformers/blob/main/setup.py#L93), so, could we use it after kernels upgrade?

benchmark_v2/framework/benchmark_runner.py
remi-or: So is it the difference in mode that is making a change? If so, it's still better to target inference IMO

benchmark_v2/framework/benchmark_config.py
remi-or: Ha yes you seem to be right... just add a #TODO then please!

benchmark_v2/framework/benchmark_runner.py
sywangyi: no, setting "use_kernels = True" in AutoModelForCausalLM.from_pretrained bring two steps.
1. register the kernels, so it could be used later
2. kernelize(self, device=Device(type=self.device.type), mode=mode), default mode is Mode.INFERENCE

if we miss 1, available kernels are not registered, a lot warn like "No kernel mapping found for layer SiLU" occurs..

the previous logic wo the PR is actually losing step 1. make available kernels not be used.

remi-or: Ok, it seems you are right. Thank you for your investigation.

remi-or (approved):
LGTM, thanks for the contribution and the insights!

yao-matrix:
@SunMarc , could you help merge it? Thx very much.

PR #43429, opened 2026-01-23
Fix lm_head weight tying in Mistral3ForConditionalGeneration

aswin00000:
## Problem
Ministral-3-3B-Instruct-2512 generates gibberish because `lm_head.weight` is not tied to `embed_tokens.weight`.

## Root Cause  
The Ministral-3-3B config has `text_config.tie_word_embeddings=True`, and the checkpoint only stores `model.language_model.embed_tokens.weight` (no `lm_head.weight`).  
However, for Mistral3 the top-level `config.tie_word_embeddings` was never set, so the base tying logic that relies on `_tied_weights_keys` was effectively disabled and `lm_head.weight` was randomly initialized.

## Solution
Instead of overriding `tie_weights()`, set `tie_word_embeddings=True` by default in `Mistral3Config` so that the existing weight-tying mechanism (`_tied_weights_keys = {"lm_head.weight": "model.language_model.embed_tokens.weight"}`) is applied correctly to Mistral3-based VLMs.

## Changes
- Updated `src/transformers/models/mistral3/configuration_mistral3.py`:
  - Added a `tie_word_embeddings: bool = True` argument to `Mistral3Config.__init__`
  - Stored it as `self.tie_word_embeddings`

### DIFF
--- a/src/transformers/models/mistral3/modeling_mistral3.py
+++ b/src/transformers/models/mistral3/modeling_mistral3.py
@@ -529,7 +529,8 @@ class Mistral3ForConditionalGeneration(Mistral3PreTrainedModel, GenerationMixin)
             should_tie = getattr(self.config.text_config, "tie_word_embeddings", should_tie)
 
         if should_tie and hasattr(self.model, 'language_model') and hasattr(self.model.language_model, 'embed_tokens'):
-            self._tie_or_clone_weights(self.lm_head, self.model.language_model.embed_tokens)
+            # Tie lm_head weight to embed_tokens weight
+            self.lm_head.weight = self.model.language_model.embed_tokens.weight
 
 
 __all__ = ["Mistral3Model", "Mistral3PreTrainedModel", "Mistral3ForConditionalGeneration"]
--- a/src/transformers/models/mistral3/modeling_mistral3.py
+++ b/src/transformers/models/mistral3/modeling_mistral3.py
@@ -528,7 +528,11 @@ class Mistral3ForConditionalGeneration(Mistral3PreTrainedModel, GenerationMixin)
         if hasattr(self.config, "text_config"):
             should_tie = getattr(self.config.text_config, "tie_word_embeddings", should_tie)
 
-        if should_tie and hasattr(self.model, 'language_model') and hasattr(self.model.language_model, 'embed_tokens'):
+        if (
+            should_tie
+            and hasattr(self.model, "language_model")
+            and hasattr(self.model.language_model, "embed_tokens")
+        ):
             # Tie lm_head weight to embed_tokens weight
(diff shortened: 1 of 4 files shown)


aswin00000:
Requesting review from @zucchini-nlp as discussed in the original bug report.
   ;;  https://github.com/huggingface/transformers/issues/43404
This PR fixes the weight tying issue that was causing gibberish output in Ministral-3-3B-Instruct-2512.
## Summary
- **Problem**: Mistral3 models generate gibberish due to improper weight tying
- **Root Cause**: Base class only checked `config.tie_word_embeddings` but multimodal models use `config.text_config.tie_word_embeddings`
- **Solution**: Added custom `tie_weights()` method that checks both locations

src/transformers/models/mistral3/modeling_mistral3.py
```diff
@@ -517,5 +517,24 @@ def prepare_inputs_for_generation(
...
         return model_inputs
 
+    def tie_weights(self, missing_keys: set[str] | None = None, recompute_mapping: bool = True):
+        """
+        Tie the weights between the input embeddings and the output embeddings.
+        """
+        super().tie_weights(missing_keys=missing_keys, recompute_mapping=recompute_mapping)
+
+        # Check for tie_word_embeddings in text_config (for VLMs) or top-level
+        should_tie = getattr(self.config, "tie_word_embeddings", False)
```
zucchini-nlp: no, no, we need simply to add `tie_word_embeddings` with default value of `True` in `Mistral3Config`. That will solve it for the 1B model and the other 2 ckpt already have `tie_word_embeddings=False` saved in their config as you mentioned

aswin00000: Thanks,  I updated the PR
- Added `tie_word_embeddings=True` default to `Mistral3Config`

CI is re-running now.

zucchini-nlp: nice, one last thing is to make CI happy and run `make fix-repo` 😄  Then we can merge

aswin00000:
@zucchini-nlp

I’m a junior engineer and this is one of my first OSS contributions

If this change looks good, If it’s not too much trouble, even a short written confirmation (here or by email) that this fix was useful would mean a lot to me as I’m building my career.

Thank you again for the reviews!  

aswinmraju2002@gmail.com

zucchini-nlp:
Thanks for the PR @aswin00000 ! LGTM at this point and I will merge it when CI turns green, the failing tests are just flake and not related :)

zucchini-nlp (approved):
forgot to ✔️

ArthurZucker:
Ty both!

PR #43430, opened 2026-01-23
Make DataCollatorMixin handle BatchFeature

jp1924:
# What does this PR do?

Modify so that if the batch returned from DataCollatorMixin is a dict, it is converted to BatchFeature and output.
In most learning and projects, instead of returning the value returned from processing_class performed in pt_call or np_call as is, there are many cases where an additional dict is made and returned, and there are many times when the functionality provided by BatchFeature cannot be used.
So, I tweaked the collator so that if the batch is a simple dict, it can be converted to BatchFeature.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@SunMarc

### DIFF
--- a/src/transformers/data/data_collator.py
+++ b/src/transformers/data/data_collator.py
@@ -46,7 +46,7 @@ class DataCollatorMixin:
         else:
             raise ValueError(f"Framework '{return_tensors}' not recognized!")
 
-        if not isinstance(batch, BatchFeature):
+        if isinstance(batch, dict):
             batch = BatchFeature(batch, return_tensors)
 
         return batch
--- a/src/transformers/data/data_collator.py
+++ b/src/transformers/data/data_collator.py
@@ -21,6 +21,7 @@ from typing import Any
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


jp1924:
@SunMarc 

Looking at this failing in the test, originally it should have gone through np's shape and returned as a tuple, 
but since it changed to pt's tensor, it's returning as a torch.Size object and causing an assert error?

Rocketknight1:
Hmn - interesting idea, but the specific object being returned can vary. It's sometimes `BatchEncoding` rather than `BatchFeature`, so I think this PR will be quite breaking. The cases when an extra key is added and a dict is returned may also break the batch assumptions for BatchFeature/BatchEncoding, so this definitely needs some extra testing!

jp1924:
@Rocketknight1 Hmm... Yeah, since the Collator is used in various areas, 
if I add it carelessly, it could cause quite a bit of confusion. 
It seems right not to add it separately. 
It would be better to make the user return it as the necessary BatchFeature.

PR #43442, opened 2026-01-23
add trackio to training notebooks

merveenoyan:
@stevhliu I also need to remove IDEFICS example (too old) and fix image captioning one (dataset was taken down, although it's hard to find an unbroken image captioning dataset), will do on a follow-up PR


stevhliu (approved):
thanks!

PR #43443, opened 2026-01-23
Fix Pixtral with flash attention

zucchini-nlp:
# What does this PR do?
Fixes https://github.com/huggingface/transformers/issues/43441

Pixtral always has packed inputs because inputs images are of different lengths. However position ids was a 1D tensor causing an issue in FA2. Unsqueezing a bs of 1 helps to solve the issue

Also, tiny bug in `torch._check`. Having it as `def msg(): return msg` will not resolve back to a string due to naming, so I renamed it to `msg_callable`

### DIFF
--- a/src/transformers/models/pixtral/modeling_pixtral.py
+++ b/src/transformers/models/pixtral/modeling_pixtral.py
@@ -507,7 +507,7 @@ class PixtralVisionModel(PixtralPreTrainedModel):
         position_ids = position_ids_in_meshgrid(
             patch_embeds_list, max_width=self.config.image_size // self.config.patch_size
         )
-        kwargs["position_ids"] = position_ids.to(patch_embeds.device, non_blocking=True)
+        kwargs["position_ids"] = position_ids.unsqueeze(0).to(patch_embeds.device, non_blocking=True)
 
         position_embeddings = self.patch_positional_embedding(patch_embeds, position_ids)
 
--- a/src/transformers/utils/import_utils.py
+++ b/src/transformers/utils/import_utils.py
@@ -1421,16 +1421,18 @@ def torch_compilable_check(cond: Any, msg: str | Callable[[], str], error_type: 
 
     if not callable(msg):
         # torch._check requires msg to be a callable but we want to keep the API simple for users
-        def msg():
+        def msg_callable():
             return msg
+    else:
+        msg_callable = msg
 
     if callable(cond):
         cond = cond()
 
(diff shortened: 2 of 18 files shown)


src/transformers/models/pixtral/modeling_pixtral.py
```diff
@@ -507,7 +507,7 @@ def forward(
         position_ids = position_ids_in_meshgrid(
             patch_embeds_list, max_width=self.config.image_size // self.config.patch_size
         )
-        kwargs["position_ids"] = position_ids.to(patch_embeds.device, non_blocking=True)
+        kwargs["position_ids"] = position_ids.unsqueeze(0).to(patch_embeds.device, non_blocking=True)
```
vasqu, line 510: Let's move this to `position_ids_in_meshgrid`, no?

src/transformers/utils/import_utils.py
```diff
@@ -1421,16 +1421,18 @@ def torch_compilable_check(cond: Any, msg: str | Callable[[], str], error_type:
 
     if not callable(msg):
         # torch._check requires msg to be a callable but we want to keep the API simple for users
-        def msg():
+        def msg_callable():
             return msg
+    else:
+        msg_callable = msg
```
vasqu, line 1427: Good catch, could we add some small test? What was the error message :eyes:

vasqu (approved):
LGTM, it probably also goes to fix #43403 (?) cc @kaixuanliu

src/transformers/utils/import_utils.py
zucchini-nlp: It complained that we can't parse `fn.local.callable` as str when trying to raise `ValueError(msg)`

zucchini-nlp: I think there should have been several tests on VLMs triggering the same issue, will check why we didn't catch it

Edit: we weren't checking error message 🤦🏻  fixed!

src/transformers/models/pixtral/modeling_pixtral.py
zucchini-nlp: no particular preference, oke

zucchini-nlp: oh actually no, them we'd have to squeeze it back in position embeddings. The extra dim is needed only for attn module

zucchini-nlp:
@bot /repo

PR #42564, opened 2025-12-02
[`BC`] Update `get_(text|image|audio|video)_features` methods to return `BaseModelOutputWithPooling`

tomaarsen:
# What does this PR do?
* Add return_dict to get_text_features, get_image_features, get_audio_features, get_video_features methods to return 'BaseModelOutputWithPooling' by default

Fixes #42401

The architectures supporting `get_image_features` are all extremely different, with wildly different outputs for the get_image_features methods: 
* 2d outputs,
* 3d outputs,
* lists of 2d outputs (due to non-matching shapes),
* existing 'return_attentions' resulting in returning 2-tuple,
* existing 'return_dict' resulting in returning 3-tuples (???),
* high quality image embeddings,
* low quality image embeddings,
* deepstack image embeddings,
* etc. etc. etc.

This PR aims to normalize these, so that users and third party libraries can perform inference on individual modalities (e.g. via `get_image_features`) despite the full model requiring multiple modalities.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

@zucchini-nlp @ArthurZucker @Cyrilvallez

- Tom Aarsen

PR #43446, opened 2026-01-23
[`typings`] Automatically type decorator return types as `tuple | X`

tomaarsen:
# What does this PR do?

This PR updates the `can_return_tuple` and ~~`check_model_inputs`~~ `capture_outputs` (see https://github.com/huggingface/transformers/pull/43446#issuecomment-3932796710) typings such that:
```python
from transformers.utils import can_return_tuple
from transformers.utils.generic import check_model_inputs
try:
    from typing import reveal_type
except ImportError:
    from typing_extensions import reveal_type


@can_return_tuple
def my_func(foo: int, bar: int, **kwargs) -> dict[str, int]:
    return {
        "sum": foo + bar,
        "product": foo * bar,
    }

result = my_func(1, 2)
reveal_type(result)  # tuple[Unknown, ...] | dict[str, int]


@check_model_inputs
def my_second_func(foo: int, bar: int, **kwargs) -> dict[str, int | float]:
    return {
        "minus": float(foo - bar),
        "div": foo / bar,
    }

result2 = my_second_func(1, 2)
reveal_type(result2)  # tuple[Unknown, ...] | dict[str, int | float]
```

In short: functions and method that use `can_return_tuple` or `check_model_inputs` must currently be typed like `tuple | X`, but can now be typed like `X` as well.

I also had Copilot help me write a `check_decorator_return_types.py` file that ensures that functions decorated with `can_return_tuple` or `check_model_inputs`:

1. Have an explicit, non-`None` return annotation.
2. Are not annotated with a union that already includes `tuple`.

The latter can be auto-fixed with `python utils/check_decorator_return_types.py --fix_and_overwrite`, and the former must be manually fixed. This checking script is included in `make fix-repo` and the CI. The script only runs in about ~5 seconds for me.

## Before submitting
      Pull Request section?
      to it if that's the case: https://github.com/huggingface/transformers/pull/42564#discussion_r2716072890

## Who can review?

cc @zucchini-nlp @vasqu @tarekziade

- Tom Aarsen


src/transformers/utils/generic.py
```diff
@@ -25,14 +25,19 @@
...
 
 import numpy as np
+from typing_extensions import ParamSpec
 
 from ..utils import logging
 from .import_utils import is_mlx_available, is_torch_available, is_torch_fx_proxy, requires
 
 
+P = ParamSpec("P")
+T = TypeVar("T")
```
tomaarsen, line 40: Reviewer note: the ParamSpec and TypeVar as used here match Python's own `typing` documentation exactly: https://docs.python.org/3.10/library/typing.html#typing.ParamSpec

They also use an example of a decorator, so the implementation here is quite standard/common.

src/transformers/models/blt/modular_blt.py
```diff
@@ -673,7 +673,6 @@ def __init__(self, config: BltLocalDecoderConfig):
 
         self.post_init()
 
-    @check_model_inputs
```
tomaarsen: Reviewer note: There are only 2 actual code changes under `src/transformers/models`, and this is one of them. This class does not return a `BaseModelOutput`, and the other related private classes (`BltLocalEncoder`, `BltGlobalTransformer`) don't use this `@check_model_inputs` either.

src/transformers/models/altclip/modeling_altclip.py
```diff
@@ -964,6 +964,7 @@ def get_input_embeddings(self):
     def set_input_embeddings(self, value):
         self.embeddings.word_embeddings = value
 
+    @can_return_tuple
```
tomaarsen, line 952: Reviewer note: There are only 2 actual code changes under `src/transformers/models`, and this is one of them. This class copies from CLAP, which uses `@can_return_tuple`, but this class did not. I've added it here.

I also updated this class and the CLAP variant to remove the `return_dict = return_dict if return_dict is not None else self.config.use_return_dict` line, as that was just dead code.

zucchini-nlp: nice! Ideally would be great to start using `check_model_inputs` for pretrained models. Though it might require more manual work than current state of PR

src/transformers/models/blt/modeling_blt.py
```diff
@@ -742,7 +742,6 @@ def __init__(self, config: BltLocalDecoderConfig):
 
         self.post_init()
 
-    @check_model_inputs
```
zucchini-nlp: accidentally deleted?

src/transformers/models/blt/modular_blt.py
zucchini-nlp: ah I see, so it is an issue with how the model was implemented. So many tiny inconsistencies 😅

vasqu: Actually, they shouldn't inherit from PreTrained model at all it seems like :sweat_smile: it's not a high priority model so fine for now

src/transformers/utils/generic.py
vasqu: Doesn't hurt to add the doc link as comment to the wrappers we use it at below - it's like 1:1 matching the example :D

utils/check_decorator_return_types.py
```diff
@@ -0,0 +1,351 @@
...
+        target = target.func
+
+    if isinstance(target, ast.Name):
+        name = target.id
+    elif isinstance(target, ast.Attribute):
+        name = target.attr
+    else:
+        return None
+
+    if name in TARGET_DECORATORS:
```
tarekziade: dead code (previous return none in all cases)

utils/check_decorator_return_types.py
```diff
@@ -0,0 +1,351 @@
...
+def _is_none_annotation(returns: ast.expr | None) -> bool:
+    if returns is None:
+        return True
+
+    # -> None
+    if isinstance(returns, ast.Constant) and returns.value is None:
+        return True
+
+    # -> None (as a name)
+    if isinstance(returns, ast.Name) and returns.id == "None":
```
tarekziade: nit

```
return isinstance(returns, ast.Name) and returns.id == "None"
```

utils/check_decorator_return_types.py
```diff
@@ -0,0 +1,351 @@
...
+The intention is that the decorators themselves are responsible for
+adding the `tuple` part of the return type, so the underlying function
+should be annotated with just the base return type.
+
+Usage (from the root of the repo):
+
+```bash
+python utils/check_decorator_return_types.py
+```
+"""
```
zucchini-nlp, line 31: 🤩

utils/check_decorator_return_types.py
```diff
@@ -0,0 +1,351 @@
...
+from __future__ import annotations
+
+import argparse
+import ast
+import os
+from collections.abc import Iterable
+from dataclasses import dataclass
+
+
+PATH_TO_TRANSFORMERS = "src/transformers"
```
zucchini-nlp: should we do "src/transformers/models"?

utils/check_decorator_return_types.py
```diff
@@ -0,0 +1,351 @@
...
+    return any(_is_tuple_type(member) for member in members)
+
+
+def _is_delegating_to_super(func_node: ast.AST) -> bool:
+    """Return True if the function body starts with a super(...) delegation.
+
+    We ignore functions whose first non-docstring statement is either:
+    - `return super(...` (possibly via an attribute like `super().foo(...)`), or
+    - `super(...` as a bare expression.
+    """
```
zucchini-nlp, line 146: usually this happens in modular or due to bad practice (like in SigLip). I am not sure if modular copies type hints though, so we might want to check fn even if they call `super()`?

utils/check_decorator_return_types.py
```diff
@@ -0,0 +1,351 @@
...
+    functions: list[tuple[ast.AST, str]] = []
+    for node in ast.walk(tree):
+        if not isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
+            continue
+        if not node.decorator_list:
+            continue
+        for deco in node.decorator_list:
+            name = _decorator_name(deco)
+            if name is not None:
+                functions.append((node, name))
```
zucchini-nlp: if name in `can_return_tuple/check_model_inputs` or did I miss the part where decorators are filetred? We have several decorators in the library

zucchini-nlp (commented):
Great work! I have just a few questions about the auto-checker

utils/check_decorator_return_types.py
vasqu: +1, it is specific to models - not expecting much perf improvement but better than nothing

utils/check_decorator_return_types.py
vasqu: Modular copies everything from its parents, typing and decorators included but let's cc @Cyrilvallez just in case

utils/check_decorator_return_types.py
tomaarsen: Refactored in aa97fc675a29859b43d8e856b39ffdc5715580a7

utils/check_decorator_return_types.py
vasqu: The function name is a bit confusing, see https://github.com/huggingface/transformers/blob/6ad9269daa6470fb9bc24513961ccc833f1ea517/utils/check_decorator_return_types.py#L84-L93

I suspect the else was never caught and then we check if it's a target name. We should rename this fn tho to something along `_valid_target_decorator_name` (hmm sounds awful but you get what I mean)

tomaarsen: That was done in `_decorator_name`, but that's super confusing, so I refactored it in aa97fc675a29859b43d8e856b39ffdc5715580a7

Now it checks if the name matches one of our targets (currently just `can_return_tuple` and `check_model_inputs`.

vasqu (approved):
Lgtm

Maybe we should start splitting into more than `fix-repo` again @Cyrilvallez :eyes:

utils/check_decorator_return_types.py
tomaarsen: It only happens in modular indeed. Modular does copy the type hint, and if there's issues, then they'll appear in the `modeling` already. I think it's fine to not require types in `modular`: the more we can copy the better.

In short, I think it's fine to skip the `super`'s. If I include them, these failures are reported:

```
src/transformers\models\csm\modular_csm.py:404: function 'forward' decorated with '@check_model_inputs' must have a non-None return annotation
src/transformers\models\ernie4_5\modular_ernie4_5.py:108: function 'forward' decorated with '@can_return_tuple' must have a non-None return annotation
src/transformers\models\ernie4_5_moe\modular_ernie4_5_moe.py:305: function 'forward' decorated with '@can_return_tuple' must have a non-None return annotation
src/transformers\models\jais2\modular_jais2.py:168: function 'forward' decorated with '@can_return_tuple' must have a non-None return annotation
src/transformers\models\metaclip_2\modular_metaclip_2.py:359: function 'forward' decorated with '@check_model_inputs' must have a non-None return annotation
src/transformers\models\metaclip_2\modular_metaclip_2.py:422: function 'forward' decorated with '@check_model_inputs' must have a non-None return annotation
src/transformers\models\metaclip_2\modular_metaclip_2.py:513: function 'forward' decorated with '@can_return_tuple' must have a non-None return annotation
src/transformers\models\metaclip_2\modular_metaclip_2.py:660: function 'forward' decorated with '@check_model_inputs' must have a non-None return annotation
src/transformers\models\metaclip_2\modular_metaclip_2.py:731: function 'forward' decorated with '@check_model_inputs' must have a non-None return annotation
```
All just cases where the modular returns `None` because they inherit the return type.

utils/check_decorator_return_types.py
vasqu: Concurrent reviewing and changes :joy: yea much better

utils/check_decorator_return_types.py
tomaarsen: Updated in 864a6a7e0559fff9429c0c1be861f05fe95064f1

utils/check_decorator_return_types.py
zucchini-nlp: sure, wanted to make sure modular does indeed copy everything

tomaarsen: Explained it a bit more in code as well: 3b0808e1498889d9e3f7ad8ae1624a13fa6bb992

utils/check_decorator_return_types.py
tomaarsen: Refactored this function in 456ee32a96a590486d16d8a7fd842cabd9ac5a5c, it really didn't need to be that large.

src/transformers/utils/generic.py
tomaarsen: Very true! Done in dbd43d60a1ddd6ba900b6c63ce99f33ad6f7d2dc & 3fd9571c96449256980f52d29d7883b1cd1eebef

tomaarsen:
I updated this PR with @Cyrilvallez' latest changes regarding `model_check_inputs` -> `capture_outputs`, so that methods wrapped with `capture_outputs` are also nicely typed:

```python
from transformers.utils import can_return_tuple
from transformers.utils.output_capturing import capture_outputs

try:
    from typing import reveal_type
except ImportError:
    from typing_extensions import reveal_type


@can_return_tuple
def my_func(foo: int, bar: int, **kwargs) -> dict[str, int]:
    return {
        "sum": foo + bar,
        "product": foo * bar,
    }


result = my_func(1, 2)
reveal_type(result)  # tuple[Unknown, ...] | dict[str, int]


@capture_outputs
def my_second_func(foo: int, bar: int, **kwargs) -> dict[str, int | float]:
    return {
        "minus": float(foo - bar),
        "div": foo / bar,
    }


result2 = my_second_func(1, 2)
reveal_type(result2)  # tuple[Unknown, ...] | dict[str, int | float]
```
```
ty check demo.py
```
```
info[revealed-type]: Revealed type
  --> demo_can_return_tuple_hint.py:19:13
   |
18 | result = my_func(1, 2)
19 | reveal_type(result)  # tuple[Unknown, ...] | dict[str, int]
   |             ^^^^^^ `tuple[Unknown, ...] | dict[str, int]`
   |

info[revealed-type]: Revealed type
  --> demo_can_return_tuple_hint.py:31:13
   |
30 | result2 = my_second_func(1, 2)
31 | reveal_type(result2)  # tuple[Unknown, ...] | dict[str, int | float]
   |             ^^^^^^^ `tuple[Unknown, ...] | dict[str, int | float]`
   |

Found 2 diagnostics
```
This should help cut down some bloat from our modeling/modular files without considerably complicating any of the decorators. Ideally, we can get this merged somewhat quickly, as the merge conflicts start quickly too.

also cc @molbap as you're also working on decorator-related cleanup in #43590 right now

- Tom Aarsen

PR #43447, opened 2026-01-23
[`chore`] Add missing r-prefix to Gemma3nAudioEncoderModelOutput

tomaarsen:
# What does this PR do?

This PR adds a required `r`-prefix to a docstring. It'll be added automatically via `utils/check_docstrings.py` if you've made changes in gemma3n, but we should just get ahead of that and fix it now.

It wasn't spotted previously as I believe the `check_docstrings.py` might only run on files that have local changes?

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@vasqu

- Tom Aarsen


vasqu (approved):
Thx!

zucchini-nlp:
> It wasn't spotted previously as I believe the check_docstrings.py might only run on files that have local changes?

It doesn't run on modular files, I've been trying to make it work in the past and forgot about it with v5 😢

tomaarsen:
> It doesn't run on modular files, I've been trying to make it work in the past and forgot about it with v5 😢

Ahh, that's it! Thanks for the intel.

- Tom Aarsen

PR #43449, opened 2026-01-23
pin `torchcodec==0.9.1`

ydshieh:
# What does this PR do?

0.10 is out but we still have the pin torch==2.9.1

Let's pin torchcodec first and come back to check the new versions later


vasqu (approved):
thx

PR #43459, opened 2026-01-24
Sanitize invalid UTF‑16 surrogate codepoints before calling the Rust encode_batch, replacing them with U+FFFD

josephrocca:
Without this fix, this crashes:
```py
from transformers import T5TokenizerFast
tok = T5TokenizerFast.from_pretrained("t5-small")
tok("ok" + "\ud800" + "bad")
```

The error message implies a wrong input type: `TypeError: TextEncodeInput must be Union[TextInputSequence, Tuple[InputSequence, InputSequence]]` - but the input is a valid Python `str`.

Last known working version: v4.44.2 (tokenizers v0.19.1)

## Who can review?

@ArthurZucker and @itazap

### DIFF
--- a/src/transformers/tokenization_utils_tokenizers.py
+++ b/src/transformers/tokenization_utils_tokenizers.py
@@ -772,6 +772,21 @@ class TokenizersBackend(PreTrainedTokenizerBase):
         **kwargs,
     ) -> BatchEncoding:
         # Input validation (from _call_one)
+        def _replace_invalid_surrogates(s: str) -> str:
+            # Replace unpaired UTF-16 surrogate code points with U+FFFD.
+            if any(0xD800 <= ord(ch) <= 0xDFFF for ch in s):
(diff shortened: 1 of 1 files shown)


ArthurZucker (commented):
Hey! sorry I think this should rather be done in `tokenizers` no?

josephrocca:
@ArthurZucker Ah, sorry about that. I'll close this an open a PR over there. I noticed that for *non-fast* tokenizers:

* BertTokenizer: does not crash, tokenizes to ['ok', '##bad'] and encodes to [101, 7929, 9024, 102]
* T5Tokenizer (SentencePiece): raises TypeError: not a string

Is there a "correct" way to handle this (at the Python-bindings level) from your perspective RE the tokenizers PR? Or should fast tokenizers match the slow ones on a per-class level?

ArthurZucker:
There is no longer any `Fast` and slow tokenizers in `v5`! thus is there is such an issue it would need a fix in `tokenizers`!

PR #43463, opened 2026-01-24
Fix SAM-HQ weight loading bug causing flaky tests

### LINKED ISSUE #42890: tests/models/sam_hq/test_modeling_sam_hq.py::SamHQModelIntegrationTest may fail since a lot of cases are lack of set_seed()

engmohamedsalah:
## What does this PR do?

Fixes #42890 by addressing the **root cause**: a weight loading bug introduced in commit `0b369802cf`.

## Root Cause

Commit `0b369802cf` ("fix sam family!") removed `_tied_weights_keys` because positional embeddings are buffers (not parameters) and cannot be tied. However, it left `_keys_to_ignore_on_load_missing` in place, causing:

```python
# BUGGY CODE (before this PR):
_keys_to_ignore_on_load_missing = ["prompt_encoder.shared_embedding.positional_embedding"]
```

This caused `prompt_encoder.shared_embedding.positional_embedding` to be **randomly initialized on every model load** instead of being loaded from the checkpoint.

The flaky integration tests were actually **exposing this weight loading bug** - outputs varied because one embedding was randomly initialized each time with `torch.randn()`.

## The Fix

Remove `_keys_to_ignore_on_load_missing` entirely so both positional embeddings are properly loaded from checkpoint:

```python
# FIXED CODE (this PR):
class SamHQModel(SamModel):
    # (removed _keys_to_ignore_on_load_missing)
    def __init__(self, config):
        ...
```

## Changes

- **src/transformers/models/sam_hq/modular_sam_hq.py**: Remove `_keys_to_ignore_on_load_missing`  
- **src/transformers/models/sam_hq/modeling_sam_hq.py**: Regenerated from modular file
- **tests/models/sam_hq/test_modeling_sam_hq.py**: Remove temporary `set_seed()` workaround from previous PR

## Testing

The integration tests should now pass consistently without requiring `set_seed()` because the model loads the correct weights from the checkpoint.

## Credit

Special thanks to @Rocketknight1 for identifying that this was a weight loading bug rather than just a randomness issue!

---

**Note**: This PR supersedes #43211 which attempted to fix the symptom (randomness) rather than the root cause (weight loading bug).

### DIFF
--- a/src/transformers/models/sam_hq/modular_sam_hq.py
+++ b/src/transformers/models/sam_hq/modular_sam_hq.py
@@ -440,6 +440,10 @@ class SamHQVisionModel(SamVisionModel):
     """
 )
 class SamHQModel(SamModel):
+    _tied_weights_keys = {
+        "prompt_encoder.shared_embedding.positional_embedding": "shared_image_embedding.positional_embedding"
+    }
+
     def __init__(self, config):
         super().__init__(config)
         self.vision_encoder = SamHQVisionEncoder(config.vision_config)


Rocketknight1:
cc @nielsrogge @yonigozlan @molbap I think this makes sense and I think you're the right maintainers for SAM, can you review?

yonigozlan:
Thanks for raising this! However this is already fixed on main by https://github.com/huggingface/transformers/pull/43420, so closing this PR.

PR #43467, opened 2026-01-24
Fix GLM-Image integration tests to use correct model class

engmohamedsalah:
## What does this PR do?

Fixes #43344 

The integration tests for GLM-Image were failing because they were attempting to load the `zai-org/GLM-4.5V` checkpoint (a GLM-4V MoE model) using `GlmImageForConditionalGeneration` classes, causing architecture mismatches.

## Issue Details

The tests were failing with the following mismatches:
1. **lm_head vocab size mismatch**: checkpoint has 151,552 tokens (text vocab) but GlmImage expects 16,512 (vision vocab for image generation)
2. **patch_embed shape mismatch**: checkpoint uses Conv3d `[1536, 3, 2, 14, 14]` but GlmImage uses Conv2d `[1536, 3, 14, 14]`

## Root Cause

GLM-4V and GLM-Image are fundamentally different model architectures:
- **GLM-4V**: Vision-to-text model (similar to LLaVA) using Conv3d for video support
- **GLM-Image**: Text-to-image generation model using Conv2d for images only

## Solution

Updated the integration tests to use `Glm4vMoeForConditionalGeneration` instead of `GlmImageForConditionalGeneration`, which correctly matches the `zai-org/GLM-4.5V` checkpoint architecture.

## Changes Made

- Added import for `Glm4vMoeForConditionalGeneration`
- Updated `get_model()` method to use `Glm4vMoeForConditionalGeneration`
- Updated `test_small_model_integration_test_batch_flashatt2()` to use `Glm4vMoeForConditionalGeneration`

## Testing

The tests were previously failing with size mismatch errors. With this fix, the model loads correctly with the matching architecture.

### DIFF
--- a/tests/models/glm_image/test_modeling_glm_image.py
+++ b/tests/models/glm_image/test_modeling_glm_image.py
@@ -20,6 +20,7 @@ from parameterized import parameterized
 
 from transformers import (
     AutoProcessor,
+    Glm4vMoeForConditionalGeneration,
     GlmImageConfig,
     GlmImageForConditionalGeneration,
     GlmImageModel,
@@ -420,7 +421,7 @@ class GlmImageIntegrationTest(unittest.TestCase):
     @classmethod
     def get_model(cls):
         if cls.model is None:
-            cls.model = GlmImageForConditionalGeneration.from_pretrained(
+            cls.model = Glm4vMoeForConditionalGeneration.from_pretrained(
                 "zai-org/GLM-4.5V", dtype="auto", device_map="auto"
             )
         return cls.model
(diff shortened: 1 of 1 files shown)


Rocketknight1:
cc @zucchini-nlp do we want these tests in this file, then, or should they move to the MoE class instead?

tests/models/glm_image/test_modeling_glm_image.py
```diff
@@ -420,7 +421,7 @@ def setUpClass(cls):
     @classmethod
     def get_model(cls):
         if cls.model is None:
-            cls.model = GlmImageForConditionalGeneration.from_pretrained(
+            cls.model = Glm4vMoeForConditionalGeneration.from_pretrained(
```
zucchini-nlp, line 424: this doesn't make sense, we don't want to run MoE tests in GlmImage model. Also, the tests are already being added by glm team in a separate PR, I will close this PR and the issue (I think code agents are opening PRs because the issue is open)

PR #43469, opened 2026-01-24
argparser: Allow optional bool flags without values

antznette1:
## Summary
Improve HfArgumentParser boolean UX so `Optional[bool]` arguments can be passed as flags without an explicit value.

## What changed
- Updated HfArgumentParser to treat `typing.Optional[bool]` `(Union[bool, None])` the same as `bool | None`:
   - keeps default `None`
   - allows `--flag` (no value) to parse as `True` via `nargs="?"` + `const=True`
- Updated unit tests in tests/utils/test_hf_argparser.py to reflect the new argparse configuration and verify `--opt` works without an explicit value.

### DIFF
--- a/src/transformers/hf_argparser.py
+++ b/src/transformers/hf_argparser.py
@@ -189,6 +189,12 @@ class HfArgumentParser(ArgumentParser):
         # A variable to store kwargs for a boolean field, if needed
         # so that we can init a `no_*` complement argument (see below)
         bool_kwargs = {}
+        is_optional_bool_type = (
+            origin_type is Union
+            and hasattr(field.type, "__args__")
+            and bool in field.type.__args__
+            and type(None) in field.type.__args__
+        )
         if origin_type is Literal or (isinstance(field.type, type) and issubclass(field.type, Enum)):
             if origin_type is Literal:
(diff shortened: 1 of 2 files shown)


Rocketknight1:
Is there an issue or reproducer to show the bug that this is fixing..?

antznette1:
> Is there an issue or reproducer to show the bug that this is fixing..?

This is fixing a CLI UX bug in HfArgumentParser for `Optional[bool]`.
Before this PR, an `Optional[bool]` flag behaves like a value-taking option, so passing the flag without a value errors (argparse expects an argument), even though most CLIs treat `--flag` as “enable”.

Reproducer:
>> from dataclasses import dataclass
from typing import Optional
from transformers.hf_argparser import HfArgumentParser
@dataclass
class Args:
    opt: Optional[bool] = None
parser = HfArgumentParser(Args)
parser.parse_args_into_dataclasses(["--opt"])  # previously errors, now sets opt=True

This PR makes `Optional[bool]` behave like a standard boolean flag: `--opt` sets `True`, while still allowing explicit values (`--opt false`) and preserving the default `None`.

PR #43477, opened 2026-01-25
Remove stale TODO comments in UDOP tied weights

raimbekovm:
## Summary
Remove outdated TODO comments claiming patch embedding weight tying is "not working".

## Details
Testing confirms the tying mechanism works correctly:
- `patch_embed.proj.weight` and `encoder.embed_patches.proj.weight` share the same tensor
- Only source weights are saved to state_dict
- Weight tying is correctly re-established when loading

The TODO was added during #41580 refactor but the mechanism works identically to token embedding tying.


Rocketknight1 (approved):
Yes, LGTM! (And I'm generally a fan of getting rid of outdated TODOs and open issues because they attract bad code agents)

raimbekovm:
lol yeah that happens

PR #43478, opened 2026-01-25
refactor: use BatchLinear in Qwen2Moe to enable PEFT/LoRA support

karthikthota-03:
**Related Issue** Fixes #43472

**Overview**  This PR is to refactor the expert implementation in Qwen2Moe by introducing a standardized BatchLinear utility.

The current implementation of Qwen2MoeExperts relies on a custom expert registry and manual looping. While this works for standard inference, it creates a major bottleneck for Parameter-Efficient Fine-Tuning (PEFT). Specifically, because the expert weights aren't stored as standard nn.Parameter tensors in a way that maps to linear logic, tools like LoRA cannot easily target the experts.

**Changes**

src/transformers/pytorch_utils.py: Added a BatchLinear class. This module stores expert weights in a single 3D tensor [num_experts, out_features, in_features] and utilizes torch.bmm for the forward pass. This makes the weights discoverable as standard parameters for PEFT and quantization.

src/transformers/models/qwen2_moe/modeling_qwen2_moe.py: Refactored Qwen2MoeExperts to use the new BatchLinear layers for gate_up_proj and down_proj. I've also removed the @use_experts_implementation dependency for this model to simplify the logic.

src/transformers/__init__.py: Added BatchLinear to the lazy-loading structure.

**Scope and Future Work** I have started with Qwen2-Moe to demonstrate the effectiveness of this architecture. By moving to a 3D tensor format, we unlock the ability for LoRA to target these layers directly.

If the maintainers prefer, I am happy to include similar refactors for all other MoE models (such as Mixtral, DeepSeek, etc.) directly within this PR. This would ensure a unified implementation across the library and provide library-wide support for MoE-LoRA immediately.

**Testing** I've verified the implementation locally:

Confirmed the forward pass produces the correct output shapes.

Verified that weight parameters are now correctly registered and accessible via model.named_parameters().

Looking forward to your feedback on whether I should proceed with the other models here!

### DIFF
--- a/src/transformers/__init__.py
+++ b/src/transformers/__init__.py
@@ -457,7 +457,7 @@ else:
         "get_wsd_schedule",
         "get_reduce_on_plateau_schedule",
     ]
-    _import_structure["pytorch_utils"] = ["Conv1D", "apply_chunking_to_forward"]
+    _import_structure["pytorch_utils"] = ["Conv1D", "BatchLinear", "apply_chunking_to_forward"]
     _import_structure["time_series_utils"] = []
     _import_structure["trainer"] = ["Trainer"]
     _import_structure["trainer_pt_utils"] = ["torch_distributed_zero_first"]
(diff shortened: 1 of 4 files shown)


Rocketknight1:
Sorry, we generally don't review "drive-by" code agent PRs from new contributors anymore! They usually have a lot of issues and it's very annoying to take the time to review/comment on them.

PR #43481, opened 2026-01-26
Fix Llama4 vision attention scaling for TP compatibility

raimbekovm:
# What does this PR do?

Fixes a bug where `vision_eager_attention_forward` ignored the passed `scaling` parameter and used `module.head_dim**-0.5` instead. This causes incorrect attention scores under Tensor Parallelism (TP) where head dimensions may differ from the original config.

**Changes:**
- Use passed `scaling` parameter in `vision_eager_attention_forward` instead of computing it internally
- Pass `self.scaling` at call site instead of `None`

**Test showing the bug:**
```python
# With different head_dim (simulating TP split)
original_head_dim = 64  # correct scaling = 0.125
tp_head_dim = 32        # buggy scaling = 0.177

# Output difference up to 0.55 between correct and buggy versions
```

## Before submitting
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case.

## Who can review?

@zucchini-nlp @ArthurZucker @molbap

### DIFF
--- a/src/transformers/models/llama4/modeling_llama4.py
+++ b/src/transformers/models/llama4/modeling_llama4.py
@@ -301,7 +301,7 @@ def vision_eager_attention_forward(
     key_states = repeat_kv(key, module.num_key_value_groups)
     value_states = repeat_kv(value, module.num_key_value_groups)
 
-    attn_weights = torch.matmul(query, key_states.transpose(2, 3)) * module.head_dim**-0.5
+    attn_weights = torch.matmul(query, key_states.transpose(2, 3)) * scaling
     if attention_mask is not None:
         causal_mask = attention_mask[:, :, :, : key_states.shape[-2]]
         attn_weights = attn_weights + causal_mask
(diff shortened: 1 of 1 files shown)


zucchini-nlp (approved):
Thanks, sounds reasonable to me. Not sure why it was left out when adding the model. Btw, I also see that ModernBert computes scale inside `eager_attention`

vasqu:
TP should not split on head dim (but on the number of available heads). Hence, it doesn't really make sense imo

cc @molbap (since you added this iirc)

raimbekovm:
Yeah you're right, I checked - colwise shards on the output dim so it's num_heads that gets split, not head_dim. The TODO comment was misleading.

Still think the fix makes sense though - the function accepts `scaling` but ignores it, which is kinda weird. Better to use what's passed in imo.

molbap:
Hey! I agree with the theory (that it does not make sense) but from what I remember of this (complicated) release, it was computed like this in the original code perhaps? I added the `None` (If I remember) to bypass this. Let's please take a look at the slow tests in `eager` and `sdpa` both before merging in case there's something off.

vasqu:
Not feeling strongly here and since it seems to work I'd rather go "don't fix what's working" - if there is an instance where you can see something not working then I'm all up for it!

raimbekovm:
fair enough, I was mass-fixing TODOs and got carried away lol

vasqu:
No worries, appreciate it! :hugs: :heart:

vasqu:
Closing this for now as discussed but feel free to reopen if you notice something (that could be added with a test)

PR #43483, opened 2026-01-26
Add pytest-random-order for reproducible test randomization

tarekziade:
# What does this PR do?

Uses bucket=module to randomize tests within files, compatible with --dist=loadfile. CI uses CIRCLE_BUILD_NUM as seed for reproducibility across parallel containers. Local runs use random seed to catch order dependencies over time. Enables debugging by reproducing exact test order from failed CI builds.

### DIFF
--- a/.circleci/create_circleci_config.py
+++ b/.circleci/create_circleci_config.py
@@ -151,6 +151,13 @@ class CircleCIJob:
         pytest_flags.append(
             f"--make-reports={self.name}" if "examples" in self.name else f"--make-reports=tests_{self.name}"
         )
+        # Use CIRCLE_BUILD_NUM as seed for pytest-random-order to ensure:
+        # - Different randomization for each build/PR
+        # - Same seed across all parallel containers in a build (reproducible)
+        # - Deterministic test order for debugging failures
+        # - bucket=module randomizes within each file, works well with --dist=loadfile
+        pytest_flags.append("--random-order-bucket=module")
+        pytest_flags.append("--random-order-seed=${CIRCLE_BUILD_NUM:-0}")
                 # Examples special case: we need to download NLTK files in advance to avoid cuncurrency issues
         timeout_cmd = f"timeout {self.command_timeout} " if self.command_timeout else ""
         marker_cmd = f"-m '{self.marker}'" if self.marker is not None else ""
--- a/.circleci/create_circleci_config.py
+++ b/.circleci/create_circleci_config.py
@@ -32,7 +32,7 @@ COMMON_ENV_VARIABLES = {
     "DISABLE_SAFETENSORS_CONVERSION": True,
 }
 # Disable the use of {"s": None} as the output is way too long, causing the navigation on CircleCI impractical
-COMMON_PYTEST_OPTIONS = {"max-worker-restart": 0, "vvv": None, "rsfE":None}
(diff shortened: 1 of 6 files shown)


tarekziade:
run-slow: vivit

tarekziade:
run-slow: vivit

.circleci/create_circleci_config.py
```diff
@@ -151,6 +153,13 @@ def to_dict(self):
         pytest_flags.append(
             f"--make-reports={self.name}" if "examples" in self.name else f"--make-reports=tests_{self.name}"
         )
+        # Use CIRCLE_BUILD_NUM as seed for pytest-random-order to ensure:
+        # - Different randomization for each build/PR
+        # - Same seed across all parallel containers in a build (reproducible)
+        # - Deterministic test order for debugging failures
+        # - bucket=module randomizes within each file, works well with --dist=loadfile
+        pytest_flags.append("--random-order-bucket=module")
+        pytest_flags.append("--random-order-seed=${CIRCLE_BUILD_NUM:-0}")
```
ydshieh: maybe consider using 

COMMON_PYTEST_OPTIONS 

for this part?

But ok if we keep this as it is.

Makefile
```diff
@@ -61,11 +61,11 @@ fix-repo: style
 
 # Run tests for the library
 test:
-	python -m pytest -n auto --dist=loadfile -s -v ./tests/
+	python -m pytest -n auto --dist=loadfile -s -v --random-order-bucket=module ./tests/
```
ydshieh: maybe add a comment to remind to install the dependency

pyproject.toml
```diff
@@ -75,6 +75,10 @@ markers = [
     "generate: marks tests that use the GenerationTesterMixin",
     "is_training_test: marks tests that use the TrainingTesterMixin (deselect with '-m \"not is_training_test\"')",
 ]
+# pytest-random-order randomizes test order to catch test interdependencies
+# Uses bucket=module to randomize tests within each file (works well with --dist=loadfile)
+# In CI: seed is set to CIRCLE_BUILD_NUM for reproducibility across parallel containers
+# Locally: uses random seed by default, or specify with --random-order-seed=<number>
```
ydshieh: is this a good place to have this comment?

I mean we only have new comment here but no actual code change?

ydshieh:
Nice to have this  🚀 ! LGTM just a few nit comments. Let you to make the final call if to update.

.circleci/create_circleci_config.py
ydshieh: question:

what 

`CIRCLE_BUILD_NUM:-0` means?

Can't just use 

`CIRCLE_BUILD_NUM`??

are they different?

ydshieh (approved):

.circleci/create_circleci_config.py
tarekziade: `{CIRCLE_BUILD_NUM:-0}` means if that variable is not found it fallbacks to `0`

tarekziade: re COMMON_PYTEST_OPTIONS great idea thanks!

Makefile
tarekziade: good point, we can also use pytest `-p` option for this: so it fails with an explicit error

PR #43490, opened 2026-01-26
Fix extras on all supported Python versions

tarekziade:
# What does this PR do?

It makes sure `extras` can be installed on all supported Python versions.

- cleaned up extras (removed natten, tweaked mistral-common etc,)
- adds a supported Python version range (10->14)
- dynamically update the metadata
- run a smoke test in the CI every night to verify `pip install` works on all extras


setup.py
```diff
@@ -224,7 +228,10 @@ def run(self):
 
 extras = {}
 
-extras["ja"] = deps_list("fugashi", "ipadic", "unidic_lite", "unidic", "sudachipy", "sudachidict_core", "rhoknp")
+extras["ja"] = deps_list("fugashi", "ipadic", "unidic_lite", "unidic", "rhoknp")
+if PYTHON_MINOR_VERSION < 14:
+    extras["ja"] += deps_list("sudachipy", "sudachidict_core")
```
Cyrilvallez: Does it make sense to have half the dependencies here for python >= 3.14? Probably we would the extra itself to be conditional no?

Cyrilvallez (commented):
Very nice idea in general!
I assume it's because those lib versions are not compatible with the python version above the conditional? In some cases, would probably make more sense to have the full extra in the conditional, instead of some of the deps!

tarekziade:
> Very nice idea in general! I assume it's because those lib versions are not compatible with the python version above the conditional? In some cases, would probably make more sense to have the full extra in the conditional, instead of some of the deps!

yes correct, recent Python versions are not well supported yet for all wheels/platform, in some projects

setup.py
tarekziade: I think it would still be useful. Without sudachipy/sudachidict_core the `ja` extra would have:

- MeCab tokenization via fugashi + dictionaries (ipadic, unidic_lite, unidic)
- Juman++ tokenization via rhoknp

tarekziade:
Supersed by @Cyrilvallez work on https://github.com/huggingface/transformers/commit/e46d21afe2aaf942fd9515b2fba1d6f3affd61a6

src/transformers/dependency_versions_table.py
```diff
@@ -91,5 +86,9 @@
     "opentelemetry-api": "opentelemetry-api",
     "opentelemetry-exporter-otlp": "opentelemetry-exporter-otlp",
     "opentelemetry-sdk": "opentelemetry-sdk",
+    "sudachipy": "sudachipy>=0.6.6",
```
ydshieh: just curious why changing the order

utils/extract_metadata.py
```diff
@@ -0,0 +1,69 @@
...
+    max_ver: int
+    min_ver, max_ver = setup.SUPPORTED_PYTHON_VERSIONS
+    versions: list[str] = [f"3.{v}" for v in range(min_ver, max_ver + 1)]
+    print(json.dumps(versions))
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        print("Usage: python utils/extract_metadata.py {extras|python-versions}", file=sys.stderr)
+        sys.exit(1)
```
ydshieh, line 58: why not just raise?

.github/workflows/extras-smoke-test.yml
ydshieh, line 1: i will let you play with this first.

we can talk about some aspects in more details in person

src/transformers/dependency_versions_table.py
tarekziade: this file was updated by `make fix-repo` - not sure why it changed the order, I will revert, run `make fix-repo` and see if it keeps it unchanged

utils/extract_metadata.py
tarekziade: it's mostly similar.

`sys.exit(1)` is like raising `SystemExit` and when you raise another error, you also get an exit code of `1`

here I chose to display a clean message and exit 1, rather than a full traceback.

.github/workflows/extras-smoke-test.yml
```diff
@@ -0,0 +1,204 @@
+name: Extras Smoke Test
+
+on:
+  schedule:
+    # Run every night at 3 AM UTC
+    - cron: "0 3 * * *"
+  workflow_dispatch:
```
ydshieh: not sure if we do need this event though

tarekziade: Sure I will remove . I guess it's a convenience to manually trigger

.github/workflows/extras-smoke-test.yml
```diff
@@ -0,0 +1,204 @@
...
+    steps:
+      - id: chk
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_CIFEEDBACK_BOT_TOKEN }}
+        run: |
+          if [ -n "$SLACK_BOT_TOKEN" ]; then
+            echo "has_token=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "has_token=false" >> "$GITHUB_OUTPUT"
+          fi
```
ydshieh, line 124: don't really need this but just something simple below

like

> if:  ... &&  secrets.SLACK_CIFEEDBACK_BOT_TOKEN != ''

ydshieh (approved):
Nice, just a few remaining nit comments

.github/workflows/extras-smoke-test.yml
tarekziade: this is what I initially did, but you can't access `secrets` from a forked repo https://github.com/huggingface/transformers/actions/runs/21511453228

utils/aggregate_failure_reports.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python
+# Copyright 2025 The HuggingFace Inc. team.
```
Cyrilvallez: Very very nit: it's now 2026! On all new files haha

setup.py
```diff
@@ -156,10 +157,24 @@
...
+    ]
+
+if PYTHON_MINOR_VERSION < 13:
+    _deps += ["kenlm"]
+
+# this is a lookup table with items like:
+#
+# tokenizers: "tokenizers==0.9.4"
+# packaging: "packaging"
+#
```
Cyrilvallez: Let's keep this comment instead, small conflict handling error
```suggestion
# This is a lookup table with items like: {"tokenizers": "tokenizers==0.9.4", "packaging": "packaging"}, i.e.
```

setup.py
```diff
@@ -172,29 +187,35 @@ def deps_list(*pkgs):
...
+    extras["audio"] += deps_list("kenlm")
 extras["video"] = deps_list("av")
 extras["timm"] = deps_list("timm")
 extras["quality"] = deps_list("datasets", "ruff", "GitPython", "urllib3", "libcst", "rich")
 extras["kernels"] = deps_list("kernels")
 extras["sentencepiece"] = deps_list("sentencepiece", "protobuf")
 extras["tiktoken"] = deps_list("tiktoken", "blobfile")
-extras["mistral-common"] = deps_list("mistral-common[image]")
+if PYTHON_MINOR_VERSION < 14:
+    extras["mistral-common"] = deps_list("mistral-common[image]")
```
Cyrilvallez, line 190: Damn, it's not working on 3.14? 😳

setup.py
```diff
@@ -293,38 +318,46 @@ def run(self):
...
+            "Intended Audience :: Developers",
+            "Intended Audience :: Education",
+            "Intended Audience :: Science/Research",
+            "Operating System :: OS Independent",
+        ]
+        + python_classifiers
+        + [
+            "Topic :: Scientific/Engineering :: Artificial Intelligence",
+        ],
+        cmdclass={"deps_table_update": DepsTableUpdateCommand},
```
Cyrilvallez, line 356: I believe we need to fix a Python version to run that, otgherwise `make fix-repo` will constantly update deps based on local python version of contributors!

Cyrilvallez (commented):
Very very nice! Thanks a lot!
I think we simply need to fix the python version on which we want to run the update of dependency table!

src/transformers/dependency_versions_table.py
tarekziade: I confirm that `make fix-repo` changes the order. We could tweak it so it outputs alpha sorted

setup.py
tarekziade: great catch, thanks. `3.10` I guess?

setup.py
tarekziade: it is working I verified separately, this is a bug. good catch thx, will tweak

setup.py
Cyrilvallez: Yes, let's always keep the oldest indeed! `SUPPORTED_PYTHON_VERSIONS[0]`

setup.py
tarekziade: no it is indeed failing with that version, 

```
m5 ➜testmistral bin/pip install "mistral-common[image]>=1.8.8"
ERROR: Ignored the following versions that require a different python version: 1.6.0 Requires-Python >=3.9.0,<3.14; 1.6.1 Requires-Python >=3.9.0,<3.14; 1.6.2 Requires-Python >=3.9.0,<3.14; 1.6.3 Requires-Python >=3.9.0,<3.14; 1.6.4 Requires-Python >=3.9.0,<3.14; 1.7.0 Requires-Python >=3.9.0,<3.14; 1.8.0 Requires-Python >=3.9.0,<3.14; 1.8.1 Requires-Python >=3.9.0,<3.14; 1.8.2 Requires-Python >=3.9.0,<3.14; 1.8.3 Requires-Python >=3.9.0,<3.14; 1.8.4 Requires-Python >=3.9.0,<3.14; 1.8.5 Requires-Python >=3.9.0,<3.14; 1.8.6 Requires-Python >=3.10.0,<3.14; 1.8.7 Requires-Python >=3.10.0,<3.14; 1.8.8 Requires-Python >=3.10.0,<3.14; 1.9.0 Requires-Python >=3.10.0,<3.14
ERROR: Could not find a version that satisfies the requirement mistral-common>=1.8.8 (from versions: 1.0.0, 1.0.1, 1.0.2, 1.1.0, 1.2.0, 1.2.1, 1.3.0, 1.3.1, 1.3.2, 1.3.3, 1.3.4, 1.4.0, 1.4.1, 1.4.2, 1.4.3, 1.4.4, 1.5.0, 1.5.1, 1.5.2, 1.5.3, 1.5.4, 1.5.5, 1.5.6)
ERROR: No matching distribution found for mistral-common>=1.8.8
```

they started to specifically require <3.13

Cyrilvallez (approved):
Perfect thanks for making the changes!

.github/workflows/extras-smoke-test.yml
```diff
@@ -0,0 +1,198 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          pattern: failure-report-*
+          path: failure_reports/
+          merge-multiple: true
+        continue-on-error: true
+
+      - name: Aggregate failures
+        run: |
+          python utils/aggregate_failure_reports.py \
```
ydshieh, line 150: ok. just a remark : I tend to (since one year) not create new script files unless it will be used in several workflow files. But it's fine, I will explain why on Monday.

PR #43495, opened 2026-01-26
fix: add compatible_model_types to suppress model type mismatch warnings

leoneperdigao:
## Related Issue
Fixes #43408

**Issue:** Warning: You are using a model of type sam3_video to instantiate a model of type sam3_tracker
**URL:** https://github.com/huggingface/transformers/issues/43408

### System Info

When using the sample code to create a SAM3 Tracker Model like:
```python
from transformers import Sam3TrackerProcessor, Sam3TrackerModel

model = Sam3TrackerModel.from_pretrained("facebook/sam3")
processor = Sam3TrackerProcessor.from_pretrained("facebook/sam3")
```

The following warning will be displayed:
```ba  
You are using a model of type sam3_video to instantiate a model of type sam3_tracker. This is not supported for all configurations of models and can yield errors.
```

This happens with version `5.0.0.dev0`

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

```python
from transformers import Sam3TrackerProcessor, Sam3TrackerModel

model = Sam3TrackerModel.from_pretrained("facebook/sam3")
processor = Sam3TrackerProcessor.from_pretrained("facebook/sam3")
```

### Expected behavior

Idealy the model should be instantiated without a Warning.

## Checklist
- [ ] Code follows project style guidelines
- [ ] Tests added/updated and passing
- [ ] Documentation updated (if applicable)
- [ ] Self-review completed

### DIFF
--- a/src/transformers/models/sam3/configuration_sam3.py
+++ b/src/transformers/models/sam3/configuration_sam3.py
@@ -402,6 +402,14 @@ class Sam3Config(PreTrainedConfig):
 
     This is the main configuration class that combines all sub-configurations for the SAM3 model.
 
+    <Tip>
+
+    SAM3 checkpoints with `model_type="sam3_video"` are compatible with `Sam3Model` since the video variant weights
+    are a superset of the image-only model weights. You may see a warning about model type mismatch when loading
+    such checkpoints, which can be safely ignored in this case.
+
+    </Tip>
+
     Args:
         vision_config (`dict` or `Sam3VisionConfig`, *optional*):
             Configuration for the vision encoder.
--- a/src/transformers/configuration_utils.py
+++ b/src/transformers/configuration_utils.py
@@ -542,8 +542,11 @@ class PreTrainedConfig(PushToHubMixin, RotaryEmbeddingConfigMixin):
             # raise warning only if we still can't see a match in `model_type`
             if config_dict["model_type"] != cls.model_type:
                 logger.warning(
-                    f"You are using a model of type {config_dict['model_type']} to instantiate a model of type "
-                    f"{cls.model_type}. This is not supported for all configurations of models and can yield errors."
+                    f"You are using a model of type `{config_dict['model_type']}` to instantiate a model of type "
+                    f"`{cls.model_type}`. This may be expected if you are loading a checkpoint that shares a subset "
+                    f"of the architecture (e.g., loading a `sam2_video` checkpoint into `Sam2Model`), but is otherwise "
+                    f"not supported and can yield errors. Please verify that the checkpoint is compatible with the "
+                    f"model you are instantiating."
                 )
 
         return cls.from_dict(config_dict, **kwargs)
--- a/src/transformers/models/edgetam/configuration_edgetam.py
+++ b/src/transformers/models/edgetam/configuration_edgetam.py
@@ -250,6 +250,14 @@ class EdgeTamConfig(PreTrainedConfig):
     Configuration objects inherit from [`PreTrainedConfig`] and can be used to control the model outputs. Read the
     documentation from [`PreTrainedConfig`] for more information.
 
+    <Tip>
+
+    EdgeTAM checkpoints with `model_type="edgetam_video"` are compatible with `EdgeTamModel` since the video variant
+    weights are a superset of the image-only model weights. You may see a warning about model type mismatch when
+    loading such checkpoints, which can be safely ignored in this case.
+
+    </Tip>
+
     Args:
         vision_config (Union[`dict`, `EdgeTamVisionConfig`], *optional*):
             Dictionary of configuration options used to initialize [`EdgeTamVisionConfig`].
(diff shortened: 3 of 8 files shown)


leoneperdigao:
Hi! I analyzed the CI failures and they appear to be unrelated to the changes in this PR:

- `test_tokenization_class_priority` - Marked as flaky, HerbertTokenizer assertion issue
- `test_apply_chat_template_with_audio` - Audio file loading error (`bcn_weather.mp3`)
- `test_apply_chat_template_with_image` / `test_batch_apply_chat_template` - Image loading errors (`picsum_237_200x300.jpg`)
- `FuyuProcessingTest` tests - Image loading errors (`bus.png`)

The test for the feature introduced in this PR (`test_compatible_model_types_no_warning`) passes locally.

That said, I might be missing something. Please let me know if there is anything I should look into further or if you have any suggestions. Happy to make adjustments!

Rocketknight1:
cc @nielsrogge @molbap @yonigozlan for SAM

src/transformers/models/sam3/configuration_sam3.py
```diff
@@ -434,6 +434,9 @@ class Sam3Config(PreTrainedConfig):
     """
 
     model_type = "sam3"
+    # sam3_video checkpoints can be loaded with Sam3Model since they share
+    # the same architecture and weights - only the model_type in config differs
```
yonigozlan: Just for precision, they don't share the same architecture and weights, sam3's weights are a subset of sam3_video's weights.
Same for the other models

tests/utils/test_configuration_utils.py
```diff
@@ -357,3 +357,51 @@ def __init__(
         self.assertIsInstance(new_config_instance.inf_positive, float)
         self.assertIsInstance(new_config_instance.inf_negative, float)
         self.assertIsInstance(new_config_instance.nan, float)
+
+    def test_compatible_model_types_suppresses_warning(self):
+        """Test that compatible_model_types suppresses the model type mismatch warning."""
+        from transformers import logging
+        from transformers.testing_utils import CaptureLogger, LoggingLevel
```
yonigozlan: Let's put these at the top of the file

yonigozlan (approved):
Hey @leoneperdigao ! Thank you for opening this PR! Happy to merge this, let's wait on core maintainers approval @vasqu @Cyrilvallez @ArthurZucker.
Also, we have the same issue with sam2/sam2_video, would you mind adding the fix for sam2 as well? Thanks!

tests/utils/test_configuration_utils.py
leoneperdigao: Done

src/transformers/models/sam3/configuration_sam3.py
leoneperdigao: Ok, check. Thanks! Adjusted the comment.

leoneperdigao:
> Hey @leoneperdigao ! Thank you for opening this PR! Happy to merge this, let's wait on core maintainers approval @vasqu @Cyrilvallez @ArthurZucker. Also, we have the same issue with sam2/sam2_video, would you mind adding the fix for sam2 as well? Thanks!

Thanks @yonigozlan! I fixed it for SAM2 as well

yonigozlan:
Bump @vasqu @Cyrilvallez @ArthurZucker ;)

Cyrilvallez:
Humm, not super convinced by the added flag. Shouldn't we simply update the MODEL_MAPPINGS here?

leoneperdigao:
> Humm, not super convinced by the added flag. Shouldn't we simply update the MODEL_MAPPINGS here?

@Cyrilvallez I considered MODEL_MAPPINGS, but I believe `compatible_model_types` may be more appropriate because:

1. MODEL_MAPPINGS are for `AutoModel` automatic dispatch, whereas this warning appears during explicit config loading (e.g., `Sam2Config.from_pretrained("sam2_video_checkpoint")`)

2. `compatible_model_types` explicitly documents the compatibility relationship in the config itself

3. Modifying MODEL_MAPPINGS could have unintended side effects on AutoModel behavior

That said, I'm open to discussion! Could you clarify which specific MODEL_MAPPING you had in mind and how it would address the warning for explicit config loading? I want to make sure I understand your suggestion correctly.

Cyrilvallez:
Oh indeed you're not using AutoModel in your example, sorry. Then I believe the warning is very legitimate. You're not using the correct model class, so I think it's quite important that the user be aware that the weights may not load correctly. 
IDK what you think @vasqu @ArthurZucker but I tend to lend against this change, as I think it would open a pandora box of potential issues if we start relying on it, and I believe the warning is legitinate/not harmful in general here

leoneperdigao:
> Oh indeed you're not using AutoModel in your example, sorry. Then I believe the warning is very legitimate. You're not using the correct model class, so I think it's quite important that the user be aware that the weights may not load correctly. IDK what you think @vasqu @ArthurZucker but I tend to lend against this change, as I think it would open a pandora box of potential issues if we start relying on it, and I believe the warning is legitinate/not harmful in general here


That's a fair point - I understand the concern.

On the other hand the original issue (#43408) raises a valid UX concern: the warning can be misleading for architecturally compatible models (like sam2/sam3 and their video variants), potentially causing "warning fatigue" or unnecessary confusion.

If we feel it is not good to incorporate this change as is, I'd then suggest the following:

1. Adjust the warning text to indicate this *might* be intentional (e.g., "This may be expected if the models share compatible architectures")

2. Document the relationship in the sam2/sam3 config docstrings without suppressing the warning

vasqu:
Agreeing with @Cyrilvallez here, it should be a rare occurrence and it will open a whole other set of issues in the future for us if we get lenient on config standards. I think that adjusting the warning message + docstring is the best solution for now as you pointed out

leoneperdigao:
@Cyrilvallez @vasqu  I adjusted as agreed. Can you verify ?

vasqu (approved):
Thanks, lgtm

PR #43496, opened 2026-01-26
Revert tokenization regression, add test

tomaarsen:
# What does this PR do?

This PR fixes the following regression:
```python

tokenizer = AutoTokenizer.from_pretrained("sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2")
sentences = ["C'est une belle journée.", "Das ist ein schöner Tag.", "It's a beautiful day."]

tokenized = tokenizer(sentences, padding=True, truncation=True, return_tensors="pt")
decoded = tokenizer.batch_decode(tokenized["input_ids"])
print(decoded)
# `main` before #42563 or with this PR:
["<s> C'est une belle journée.</s>", '<s> Das ist ein schöner Tag.</s><pad>', "<s> It's a beautiful day.</s>"]
# `main` since #42563:
["<s> c ' est une belle <unk>. </s>", '<s> das ist ein <unk> tag. </s> <pad>', "<s> it ' s a <unk> day. </s>"]
```

The issue was introduced in #42563. In essence, alongside removing a lot of sentencepiece conversion code, this PR also removed this code (https://github.com/huggingface/transformers/pull/42563/files#r2727583533) that loads the `tokenizer_file` if it exists:
```python
        # If tokenizer_file exists and tokenizer has a TokenizersBackend, replace the blank tokenizer with tokenizer.json if tokenizer_file is not None and hasattr(tokenizer, "_tokenizer"):
            from tokenizers import Tokenizer as TokenizerFast

            tokenizer._tokenizer = TokenizerFast.from_file(tokenizer_file)
            # Re-run post-initialization if the tokenizer has it
            if hasattr(tokenizer, "_post_init"):
                tokenizer._post_init()
```

Because this code was removed, the tokenizer from https://huggingface.co/sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2 loads with the default WordPiece model, instead of the Unigram that the model was trained for. I've added a test that fails on `main` that's resolved by reintroducing this code.

This is currently heavily affecting the performance of the https://huggingface.co/sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2 embedding model, causing test failures on my side.

### Note
I also simplified some
```python
        for tokenizer_class in [BertTokenizer, BertTokenizer]:
            with self.subTest(f"{tokenizer_class}"):
```
test cases. This is likely a leftover from replacing BertTokenizerFast.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@ArthurZucker @itazap

- Tom Aarsen

### DIFF
--- a/tests/tokenization/test_tokenization_utils.py
+++ b/tests/tokenization/test_tokenization_utils.py
@@ -127,30 +127,28 @@ class TokenizerUtilsTest(unittest.TestCase):
 
     @require_tokenizers
     def test_decoding_single_token(self):
-        for tokenizer_class in [BertTokenizer, BertTokenizer]:
-            with self.subTest(f"{tokenizer_class}"):
-                tokenizer = tokenizer_class.from_pretrained("google-bert/bert-base-cased")
+        tokenizer = BertTokenizer.from_pretrained("google-bert/bert-base-cased")
 
-                token_id = 2300
-                decoded_flat = tokenizer.decode(token_id)
-                decoded_list = tokenizer.decode([token_id])
+        token_id = 2300
+        decoded_flat = tokenizer.decode(token_id)
+        decoded_list = tokenizer.decode([token_id])
 
-                self.assertEqual(decoded_flat, "Force")
-                self.assertEqual(decoded_list, "Force")
+        self.assertEqual(decoded_flat, "Force")
+        self.assertEqual(decoded_list, "Force")
 
-                token_id = 0
-                decoded_flat = tokenizer.decode(token_id)
-                decoded_list = tokenizer.decode([token_id])
+        token_id = 0
+        decoded_flat = tokenizer.decode(token_id)
+        decoded_list = tokenizer.decode([token_id])
 
-                self.assertEqual(decoded_flat, "[PAD]")
-                self.assertEqual(decoded_list, "[PAD]")
+        self.assertEqual(decoded_flat, "[PAD]")
+        self.assertEqual(decoded_list, "[PAD]")
 
-                last_item_id = tokenizer.vocab_size - 1
-                decoded_flat = tokenizer.decode(last_item_id)
-                decoded_list = tokenizer.decode([last_item_id])
+        last_item_id = tokenizer.vocab_size - 1
+        decoded_flat = tokenizer.decode(last_item_id)
+        decoded_list = tokenizer.decode([last_item_id])
 
-                self.assertEqual(decoded_flat, "##：")
-                self.assertEqual(decoded_list, "##：")
+        self.assertEqual(decoded_flat, "##：")
+        self.assertEqual(decoded_list, "##：")
 
     def test_extra_special_tokens_multimodal(self):
         attribute_special_tokens_list = [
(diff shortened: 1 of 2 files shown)


tests/tokenization/test_tokenization_utils.py
```diff
@@ -260,19 +256,37 @@ def test_instantiation_from_tokenizers_json_file(self):
...
+            self.assertEqual(tokenizer._tokenizer.model.unk_token, "[UNK]")
+
+    @require_tokenizers
+    def test_instantiation_from_non_default_tokenizers_json_file_from_pretrained(self):
+        bert_tokenizer = Tokenizer(Unigram([("test", 0.2), ("[UNK]", 0.01), ("token", 0.3)], unk_id=1))
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            bert_tokenizer.save(os.path.join(tmpdirname, "tokenizer.json"))
+            tokenizer = BertTokenizer.from_pretrained(tmpdirname)
+            self.assertIsInstance(tokenizer._tokenizer.model, Unigram)
+            self.assertEqual(tokenizer._tokenizer.model.token_to_id("token"), 2)
```
tomaarsen, line 279: This is the core new test, it fails on `main` with:
```
AssertionError: WordPiece(unk_token="[UNK]", continuing_subword_prefix="##", max_input_chars_per_word=100, vocab={"test":0, "[UNK]":1, "token":2}) is not an instance of <class 'tokenizers.models.Unigram'>
```
as the BertTokenizer ignores the tokenizer file and just relies on its defaults instead.

tomaarsen:
This PR is causing a regression of its own with 
```
FAILED tests/models/auto/test_tokenization_auto.py::AutoTokenizerTest::test_do_lower_case - AssertionError: '[UNK]' != 'hello'
```
Perhaps this PR is a bit too naive: it likely ignores explicit kwargs when loading the `AutoTokenizer.from_pretrained` in favor of the tokenizer file. There's likely a cleaner fix. cc @ArthurZucker as you refactored this initially.

- Tom Aarsen

tomaarsen:
I've just spotted this section in the migration guide: https://github.com/huggingface/transformers/blob/main/MIGRATION_GUIDE_V5.md#tokenization
Which seems to imply that the change was intended, and that "custom" tokenizer options will no longer be supported. 

How should existing models tackle this? Introduce a custom tokenizer with `auto_map` and spring `trust_remote_code` on their users?

- Tom Aarsen

ArthurZucker (commented):
hey @tomaarsen ! I am not sure we can do anything for such old models! 
We want to support custom models, but we cannot be backward compatible on this one with the proposed fix. 

The best is to push a proper tokenizer.json imo!

tomaarsen:
I ended up patching this by setting `"tokenizer_class": "PreTrainedTokenizerFast"` in the `tokenizer_config.json`. This prevents the `BertTokenizer` code from overriding the `tokenizer.json` Unigram configs, and works on both Transformers v4 and v5. Much better.

Relevant commit: https://huggingface.co/sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2/commit/e8f8c211226b894fcb81acc59f3b34ba3efd5f42

- Tom Aarsen

PR #43497, opened 2026-01-26
[Bugfix] Extract pad_token_id from text config for Llama-4

eldarkurtic:
# What does this PR do?

Loading Llama-4 model with `Llama4ForConditionalGeneration` fails because `self.config.pad_token_id` doesn't exist. For Llama-4 models, `pad_token_id` is inside `text_config` not the general config.


Rocketknight1:
cc @zucchini-nlp I think this is the same bug as Qwen3

tomaarsen:
Was this now resolved in #43539, @Rocketknight1? If so, we can close.

- Tom Aarsen

zucchini-nlp:
Yes, I think we can close it

PR #43505, opened 2026-01-26
fix(models): Migrate legacy segmentation_indices to out_indices in BeitConfig

harshaljanjani:
### What does this PR do?

→ Add backward compat for legacy `segmentation_indices` field in `BeitConfig`.
→ Fix the conversion script to set `out_indices` for base models; should the model checkpoints be re-exported after this PR.
→ Improve validation error messages to catch missing `out_indices` and display clearer errors.

Fixes #43504.

### Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/src/transformers/models/beit/modeling_beit.py
+++ b/src/transformers/models/beit/modeling_beit.py
@@ -1194,6 +1194,12 @@ class BeitForSemanticSegmentation(BeitPreTrainedModel):
         self.beit = BeitModel(config, add_pooling_layer=False)
 
         # FPNs
+        if self.config.out_indices is None:
+            raise ValueError(
+                f"BeitForSemanticSegmentation requires config.out_indices to be set. "
+                f"For a base model ({config.num_hidden_layers} layers), use: config.out_indices = [3, 5, 7, 11]. "
+                f"For a large model (24 layers), use: config.out_indices = [7, 11, 15, 23]."
+            )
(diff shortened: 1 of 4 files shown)


harshaljanjani:
Following up on this PR, happy to make changes or add context if helpful!

src/transformers/models/beit/modeling_beit.py
```diff
@@ -1194,6 +1194,12 @@ def __init__(self, config: BeitConfig) -> None:
         self.beit = BeitModel(config, add_pooling_layer=False)
 
         # FPNs
+        if self.config.out_indices is None:
+            raise ValueError(
+                f"BeitForSemanticSegmentation requires config.out_indices to be set. "
+                f"For a base model ({config.num_hidden_layers} layers), use: config.out_indices = [3, 5, 7, 11]. "
+                f"For a large model (24 layers), use: config.out_indices = [7, 11, 15, 23]."
```
Rocketknight1: We can probably just remove these errors! I don't think they're correct - in particular, printing `config.num_hidden_layers` in the error is only correct if the model is a base model. Let's just keep the configuration change and revert the rest.

harshaljanjani: Thanks for the review; comments addressed!

Rocketknight1 (approved):
LGTM now!

PR #43509, opened 2026-01-26
Guard torch.is_autocast_enabled(device_type) for torch<2.4

oliverholworthy:
This PR restores compatibility with PyTorch versions < 2.4 while preserving the current behavior on newer versions.

`torch.is_autocast_enabled(device_type)` was introduced in PyTorch 2.4, but Transformers currently supports torch>=2.2.  
On torch 2.2 and 2.3, calling this function with an argument raises a `TypeError`.

This change gates the call based on the installed torch version:

```python
if is_torch_greater_or_equal_than_2_4:
    is_autocast_active = torch.is_autocast_enabled(device_type)
else:
    is_autocast_active = torch.is_autocast_enabled()
````

This ensures:

* Newer PyTorch versions use the device-specific autocast API.
* Older supported PyTorch versions continue to work without runtime errors.

Motivation: avoid breaking users on torch 2.2/2.3 while keeping the existing minimum torch requirement unchanged.

Fixes #43508

## Who can review?

Tagging for awareness based on prior related changes and pipeline context:

@Rocketknight1

### DIFF
--- a/src/transformers/utils/generic.py
+++ b/src/transformers/utils/generic.py
@@ -45,9 +45,9 @@ if is_torch_available():
     from torch.types import _dtype
 
     from ..model_debugging_utils import model_addition_debugger_context
+    from ..pytorch_utils import is_torch_greater_or_equal_than_2_4
 
     _is_torch_available = True
-
 
 # required for @can_return_tuple decorator to work with torchdynamo
 _is_mlx_available = False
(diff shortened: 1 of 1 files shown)


Rocketknight1:
I think we're just going to modify https://github.com/huggingface/transformers/pull/41307 to bump the minimum version to 2.4 instead! It's about time to do so anyway

PR #43510, opened 2026-01-26
Fix and re-enable extra_state tests

pstjohn:
This PR fixes and re-enables the skipped `extra_state` tests, since we use this functionality when embedding NVIDIA TransformerEngine layers in PreTrainedModels.

Thanks!

@ArthurZucker for review


pstjohn:
@ArthurZucker, gentle ping

pstjohn:
Pinging @Cyrilvallez as an alternate reviewer

pstjohn:
@Rocketknight1 maybe? 🥺

ArthurZucker:
Sorry missed the ping

ArthurZucker (commented):
Is this super useful (i mean in terms of usage?)

pstjohn:
> Is this super useful (i mean in terms of usage?)

TransformerEngine uses it to serialize their FP8 metadata information; it's been part of the pytorch API for ages. We're currently stripping that information when saving models with `save/load pretrained`, but it does add some extra legwork on our end.

In terms of other folks using these hooks, it looks like they're out there:
https://github.com/search?q=%22def+set_extra_state(self,+state%22&type=code

🤷

ArthurZucker (approved):
Okay, ty for fixing then!

src/transformers/core_model_loading.py
```diff
@@ -897,6 +897,10 @@ def set_param_for_module(
     module_path, _, param_name = target_name.rpartition(".")
     module_obj = model.get_submodule(module_path) if module_path else model
 
+    if param_name == torch.nn.modules.module._EXTRA_STATE_KEY_SUFFIX:
+        module_obj.set_extra_state(param_value)
+        return
```
ArthurZucker, line 904:
```suggestion
        module_obj.set_extra_state(target_name)
        loading_info.missing_keys.discard(target_name)
        return
```
should you not use the target name (tho probably no one uses conversion on extra-states

ArthurZucker: also IDK how extra states behave but we nee dot set 
````
param_value._is_hf_initialized = True
```
to avoid overwriting it if its in the init scheme

pstjohn: 🤔 i'm not sure we need (or can easily support) `_is_hf_initialized` here, since we don't know _what_ `set_extra_state` is going to do with that information. That might be the user's responsibility inside `set_extra_state` if we indeed see issues with resetting.

Typically though this isn't being used for parameters, since those are already in the state dict, but for misc. extra data that we want to get stored in the state_dict.

pstjohn:
@ArthurZucker this is RTM on my end, I think the tests_generate failure is unrelated

PR #43513, opened 2026-01-27
Fix cardinality error for DETR models without explicit background class

### LINKED ISSUE #42199: Cardinality error is incorrect for models derived from DETR that do not have an explicit background class
##  Issue

For DETR variants, the cardinality errors that are reported during training are incorrect. This was reported in the DeformableDETR repository, and was acknowledged but not resolved:

https://github.com/fundamentalvision/Deformable-DETR/issues/24

Since all the derived models no longer include an explicit background class, the returned cardinality error is wrong. This affects at least Deformable DETR. I'm not sure about Conditional as it doesn't seem to be reproducible in the fine-tuning example.

While `loss_cardinality` is not strictly a loss, it is a bug to include it. It's confusing to users because it's returned as part of the loss dict. It also breaks a common pattern in logging frameworks of calling `logger.log('total_loss', sum(loss_dict.items())`, because `loss_cardinality` is both huge compared to the real losses, and does not change during training in a meaningful way.

## Context

In the original DETR implementation (broadly copied here), the class embedding outputs +1 logits to explicitly model the background (which is tacked on as index `-1`). Hence the recommendation to instantiate models with `max(class_ids)`. In later works, the loss function uses a sigmoid/focal loss and the output logits have len(num_classes). It's mentioned in passing in the DeformableDETR paper, but not clear to me if it was strictly necessary for the decoder design or faster convergence. I'm not aware of an earlier experiment, but every other paper seems to now use that convention.

The ImageLoss function has this method: https://github.com/huggingface/transformers/blob/8cb5963cc22174954e7dca2c0a3320b7dc2f4edc/src/transformers/loss/loss_for_object_detection.py#L143-L157

Note that the assumption is “no object” is explicitly class index -1. In other version, like Deformable DETR, ImageLoss is sub-classed. But since this index actually refers to a genuine class, what this actually computes is the number of predictions that are classes[-1] versus ground truth. As an obvious example, in the case of a 1-class model, the output is always `num_queries - num_gt`.

## Proposed solution

Remove the entry in the `loss_dict` entirely and don't calculate cardinality error. I'm happy to submit a PR for this.

Alternatively we could compute it based on the score threshold, but that's a post processing step and not in the forward pass. There may be a more useful metric to return here, but since these models always predict `num_queries` objects, the concept of cardinality only makes sense once you filter them.

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

Run the forward pass with any of these models, including targets. This is a 1-class model:

Image

### Expected behavior

Either return a true cardinality error or nothing. However as above - it doesn't make sense because the model returns all queries as foreground predictions.

heathdutton:
For DETR-derived models that use sigmoid/focal loss (Deformable DETR, Grounding DINO, LW-DETR, RT-DETR, D-FINE), the cardinality error calculation was incorrect. These models don't have an explicit background class, so checking `argmax(-1) != shape[-1] - 1` is meaningless.

This PR fixes the cardinality calculation to use `sigmoid().max(-1).values > 0.5` to determine if a query confidently predicts an object.

The original DETR's `ImageLoss.loss_cardinality()` remains unchanged since it correctly uses softmax with an explicit background class.

Fixes #42199


Rocketknight1:
cc @nielsrogge @yonigozlan

yonigozlan (approved):
Thanks for fixing @heathdutton ! Indeed the cardinality loss was wrong for these models. LGTM!

PR #43517, opened 2026-01-27
Make it possible to evaluate when using sequence parallel in HF Trainer

jp1924:
# What does this PR do?

When using that sp(sequence parallel) in evaluate, the sequence length must be divisible by `world_size`, and the eval batch size must be the same as the train batch size to be able to use it, and this is a PR to modify that.
When using deepspeed sp, due to monkey patching the attention function, the following problems occur.

1. Eval batch size must be the same as train batch size.
2. During eval, the sequence must always be divisible by world_size.

In this situation, I propose to add the following two things in this PR.

1. A callback that restores the monkey-patched attention function to the original function during evaluate, and then reapplies it when training starts.
2. Check if it's training or not with self.model.training in the compute_loss section.

I think it should be made so that it can work without sequence parallel during evaluate.

However, some discussion is needed regarding these changes.
I'm not sure if when you first designed sequence parallel, you only considered it for training, or if you designed it to work together in evaluate as well.
Looking at the current code, it seems like it was made assuming it would be used in training, so I modified the code so that it can work normally in evaluate.

---

- `transformers` version: 5.0.0.dev0
- Platform: Linux-5.15.0-153-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- Huggingface_hub version: 1.3.3
- Safetensors version: 0.7.0
- Accelerate version: 1.12.0
- DeepSpeed version: 0.18.4
- PyTorch version (accelerator?): 2.9.1+cu128 (CUDA)
- GPU type: NVIDIA H100 NVL

I'm using the dev version via `pip install git+https://github.com/huggingface/transformers.git@dfe30827b8ebdd974eb7ce69c7d5d8cf8e6cf852`

## Before submitting

      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc


src/transformers/integrations/integration_utils.py
```diff
@@ -2416,3 +2418,17 @@ def get_reporting_integration_callbacks(report_to):
...
+
+    def on_step_begin(self, *args, **kwargs) -> None:
+        if self._local_mapping:
+            ALL_ATTENTION_FUNCTIONS._local_mapping = self._local_mapping
+            self._local_mapping = None
+
+    def on_step_end(self, args, state, control, **kwargs) -> None:
+        if control.should_evaluate:
+            self._local_mapping = deepcopy(ALL_ATTENTION_FUNCTIONS._local_mapping)
+            ALL_ATTENTION_FUNCTIONS._local_mapping = ALL_ATTENTION_FUNCTIONS._global_mapping
```
SunMarc: can you show me where the monkey patching due to sp happens ?

SunMarc (commented):
Thanks for the PR, left a comment

src/transformers/integrations/integration_utils.py
kashif: that is in the deepspeed i beleive: https://github.com/deepspeedai/DeepSpeed/blob/master/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L465-L466

kashif:
good question @jp1924 lets me ask around if we want default, i.e. no SP for eval mode and then decide

src/transformers/integrations/integration_utils.py
jp1924: @SunMarc 

The attention function changes occur through deepspeed's Accelerator's [accelerator.py > Accelerator > _prepare_deepspeed](https://github.com/huggingface/accelerate/blob/00e4d264f66d9694ba6116a0dbe7ecec83655b43/src/accelerate/accelerator.py#L2369-L2387) and then in deepspeed's [ulysses_sp.py > UlyssesSPAttentionHF > register_with_transformers](https://github.com/deepspeedai/DeepSpeed/blob/c89e0db8e2b061e9a1f872731b4d7bf92f628231/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L345-L468).

When checking the actual values in the vscode debug console, the results appear as shown in the image.

image


Looking at the image, you can see that it's divided into `_local_mapping` and `_global_mapping`, and you can confirm that `_local_mapping` has been changed to `UlyssesSPAttentionHF`.

jp1924: I have some points I'd like to discuss:

1. Is it appropriate to add this to transformers?
   - I'm thinking it might be more appropriate to add this kind of functionality to accelerate instead

2. Is it right to handle this with callbacks?
   - Couldn't we make it automatically flow to global or local in `ALL_ATTENTION_FUNCTIONS` based on whether 
   model.training is True or False?
   - If there's a singleton-like approach (similar to Accelerator's PartialState) that's globally accessible without overhead and can globally check whether model.training is True or False, it seems like this could be easily applied here

What do you think?

kashif:
@jp1924, another option might be to make the changes in deepspeed? we can make a context variable for eval that then when toggled skips the SP sharding in the deepspeed's uattn_wrapper ?

jp1924:
Oh? You're right! Now that I think about it, modifying transformers or accelerate is just a temporary workaround, not a fundamental solution. @kashif your approach seems to be a more fundamental way to solve the problem.

jp1924:
@kashif @SunMarc 
Great — the changes have been applied to DeepSpeed and Accelerate. 
Now it looks like we just need to apply the changes to transformers.

If writing tests is necessary, we might consider adding the test from this comment to transformers: https://github.com/deepspeedai/DeepSpeed/pull/7821#issuecomment-3822303457

SunMarc:
Yes, would you like to try writing a simple test to check that everything works ? You should try to come up with a test that doesn't take too much time to run !

kashif:
@SunMarc I added a change in the SP test to do eval on an odd sized input, but we will need to wait for a new version of deepspeed and accelerate before we can test this right?

SunMarc:
Yeah indeed ! But we can merge this even if it will fail for now.

SunMarc (approved):
Thanks a lot !

jp1924:
@kashif @SunMarc 
First, I'm sorry for being late
I've been busy and couldn't check properly. But when I came back, 
the tests were already created? 

Thanks for helping! 
I'm sorry it seems like I left the work I should have done to you... ugh

SunMarc:
> Thanks for helping!
> I'm sorry it seems like I left the work I should have done to you... ugh

It's @kashif who did it ! Thanks for your work and making the deepseed pr happen

jp1924:
thank you!

PR #43518, opened 2026-01-27
[Kernels] kernel migration updates for activation kernels

ariG23498:
CC: @MekkCyber


MekkCyber (commented):
Thank you! Just one small question, does this work with previous `kernels` versions too or only the latest ?

ariG23498:
>  does this work with previous kernels versions too or only the latest ?

I am using the latest one.

Do you think we should add a test for previous kernel version?

MekkCyber:
Does this pr fail with previous kernels versions ?

ariG23498:
> Does this pr fail with previous kernels versions ?

I could not give it a try, do you want me to try and revert back?

MekkCyber:
Yes I think it would be cool to test with other versions before merging this 🤗

ariG23498:
@MekkCyber running with `v0.12.1` and `v0.11.7` on the following
```

    model_id,
    dtype="auto",
    device_map="auto",
    use_kernels=True,
)
```

resulted in proper downloading of kernels. (I had also removed the hf cache before running them sequentially)

MekkCyber (approved):
Thanks for testing !

ariG23498:
@MekkCyber I think the failing CI is due to other issues (unrelated to the PR)

Let me know if there is something else I need to work on! 🤗

MekkCyber:
cc @ArthurZucker ! if you can merge please 🤗 thanks !

PR #43520, opened 2026-01-27
fix: initialize BatchNorm2d buffers only on meta

tarekziade:
# What does this PR do?

Commit 8dd9c999a6262d6ceb48f4a2da7acaccfa80e3bc introduced a regression by unconditionally reinitializing BatchNorm2d buffers (running_mean, running_var, num_batches_tracked) in the _init_weights() method.

The problem: When loading pretrained timm models, the flow is:
1. timm.create_model(pretrained=True) correctly loads pretrained BatchNorm statistics
2. post_init() calls _init_weights() on all modules
3. _init_weights() overwrites the pretrained BatchNorm buffers with zeros/ones
4. Model produces incorrect results because normalization uses wrong statistics

The error shows a mismatch at index (0, 178, 0, 0) with a difference of 80224.0 on CircleCI on AVX 512 CPUs

The fix: added a check to only initialize BatchNorm2d buffers when they're on the meta device (uninitialized): This ensures pretrained buffers are preserved while still handling the meta device initialization case.

### DIFF
--- a/src/transformers/models/timm_backbone/modeling_timm_backbone.py
+++ b/src/transformers/models/timm_backbone/modeling_timm_backbone.py
@@ -118,9 +118,27 @@ class TimmBackbone(PreTrainedModel, BackboneMixin):
         if hasattr(module, "init_non_persistent_buffers"):
             module.init_non_persistent_buffers()
         elif isinstance(module, nn.BatchNorm2d):
+            # PyTorch's BatchNorm2d always registers all three buffers together, so we only check one
             running_mean = getattr(module, "running_mean", None)
-            if running_mean is not None and running_mean.device.type == "meta":
-                # Only initialize if on meta device (uninitialized), not if already loaded from pretrained
+
+            # Determine if buffers need initialization based on two cases:
+            # 1. Meta device case: Model created with `with torch.device("meta"):`
+            #    - Buffers are on meta device and uninitialized
+            # 2. to_empty() case: Model moved from meta to real device via `model.to_empty(device="cpu")`
+            #    - Buffers are moved to real device but contain invalid values (zeros)
+            #    - running_var with zeros is invalid (should be ones for proper normalization)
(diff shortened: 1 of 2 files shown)


src/transformers/models/timm_backbone/modeling_timm_backbone.py
```diff
@@ -117,10 +117,26 @@ def _init_weights(self, module):
...
         if hasattr(module, "init_non_persistent_buffers"):
             module.init_non_persistent_buffers()
-        elif isinstance(module, nn.BatchNorm2d) and getattr(module, "running_mean", None) is not None:
-            init.zeros_(module.running_mean)
-            init.ones_(module.running_var)
-            init.zeros_(module.num_batches_tracked)
+        elif isinstance(module, nn.BatchNorm2d):
+            running_mean = getattr(module, "running_mean", None)
+            running_var = getattr(module, "running_var", None)
+            num_batches_tracked = getattr(module, "num_batches_tracked", None)
```
Cyrilvallez: I one exists, all are guaranteed to exist, so we can only check one (see https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/batchnorm.py#L61-L79)

src/transformers/models/timm_backbone/modeling_timm_backbone.py
```diff
@@ -117,10 +117,26 @@ def _init_weights(self, module):
...
+            if (
+                running_mean is not None
+                and running_var is not None
+                and (
+                    running_mean.device.type == "meta"
```
Cyrilvallez: We can simply check meta device here I believe! Should not need to check for limit values

Cyrilvallez (approved):
Ha indeed, very nice thanks! We can simplify a bit the different checks though!

tarekziade:
run-slow: timm_backbone, timm_wrapper

PR #43521, opened 2026-01-27
[CB] Minor perf improvements and ty compatibility

remi-or:
# Summary
This PR introduces minor changes to the continuous batching feature:

#### Performance
- optimize the `get_seqlens_k` function to avoid looping over all cache managers
- logits indexing is now done inside the cuda graph, as there was no apparent reason it should be done outside
- fuse the `get_read/write_indices` functions, which should also make the code more readable
#### Typing
- add a `ProtoPretrainedModel` class for typing purposes
- make the code `ty` compliant, using as little `# type:ignore` as possible
- add the `set_continuous_batching_context` to the `LogitsProcessorsList` class to stay coherent with typing 
#### Usability
- rename a variable in `copy_cache`
- we now add a temporary token to the request that get scheduled, which is replaced when a new real token is produced. This is not very usefull as of now but will be when we add CPU / GPU overlapping scheduling, which is soon™

## Performances
| Arguments                                          | Throughput on main | Throughput with PR | Delta    |
|----------------------------------------------------|--------------------|--------------------|----------|
| --samples 10                                       | 530.45             | 531.02             | +0.57    |
| --samples 20 --num-blocks 16/20                    | 100.14             | 126.91             | N/A*     |
| --samples 50                                       | 1477.42            | 1478.66            | +1.24    |
| --samples 100                                      | 2266.81            | 2243.32            | -23.49   |
| --samples 100 --attn flash_attention_2             | 2076.81            | 2050.43            | -26.38   |
| --samples 100 --attn sdpa                          | 840.72             | 831.78             | -8.94    |
| --samples 500                                      | 5281.98            | 5287.99            | +6.01    |
| --samples 500 --add-prefix --compile               | 6743.21            | 6732.60            | -10.61   |
| --samples 50 --num-return-sequences 8 --do-sample  | 5362.52            | 5359.59            | -2.93    |
| --samples 100 --num-return-sequences 4 --do-sample | 4674.91            | 4697.05            | +22.14   |

## Tests
Tests pass, including the tests added for this PR.

## Sanity check
Looks good.


src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -253,12 +286,14 @@ def setup_static_tensors(self, num_groups: int) -> None:
         self.tensor_metadata = {"dtype": torch.int32, "device": self.model_device}
```
ArthurZucker: you no longer use these no? you could type them as type dict if you still wanted to use them

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -1328,6 +1375,8 @@ def generate_batch(
                     else:
                         if not manager.is_running():
                             logger.error("Generation thread terminated unexpectedly.")
+                            # This helps get some information in stdout
```
ArthurZucker, line 1377: not sure its really necessary

ArthurZucker (approved):
thanks

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: Right, removing it.

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: Actually it is kind of helpful if you call CB through `subcommand.run` because trying to capture `stderr` means you also capture the TQDM bars, which is a pain. Open to any other alternatives!

PR #43523, opened 2026-01-27
Tie the weights even if initializing from a config on meta device

Cyrilvallez:
# What does this PR do?

Fix https://github.com/huggingface/transformers/issues/43522.
TLDR we want to skip tying when inside `from_pretrained` (so we add the context manager), but always tie when initializing from config (even with meta device).

Consider the following:

```python
import torch
from transformers import AutoConfig,  PaliGemmaForConditionalGeneration


repo_id = "google/paligemma-3b-mix-224"
config = AutoConfig.from_pretrained(repo_id)
with torch.device("meta"):
    model = PaliGemmaForConditionalGeneration._from_config(config)

lm_head_in_params = "lm_head.weight" in [n for n, _ in model.named_parameters()]
print(f"Weights were tied correctly: {not lm_head_in_params}")
```

this was False before, True now

### DIFF
--- a/src/transformers/initialization.py
+++ b/src/transformers/initialization.py
@@ -243,3 +243,25 @@ def no_init_weights():
                 setattr(module, func_name, func)
         # Set back `init_weights`
         PreTrainedModel.init_weights = original_init_weights
+
+
+@contextmanager
+def no_tie_weights():
+    """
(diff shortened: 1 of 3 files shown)


ArthurZucker (approved):
LGTM let' s enforce this via a test

PR #43527, opened 2026-01-27
fix: return labels instead of label in reduce_label method in BeitImageProcessorFast

sbucaille:
# What does this PR do?
Fixes #43526 

## Who can review?

@yonigozlan @molbap


yonigozlan (approved):
Thanks for fixing and adding tests!

PR #43534, opened 2026-01-27
[Benchmark] Disable GPU monitoring by default on AMD GPUs

Abdennacer-Badaoui:
GPU monitoring causes ~2-3x slowdown on AMD GPUs during benchmark runs. This PR automatically disables GPU monitoring when an AMD GPU is detected, with a warning message explaining the reason.
Users can still force-enable GPU monitoring with `--force-gpu-monitoring` if needed.


Abdennacer-Badaoui:
This will be closed if we merge this https://github.com/huggingface/transformers/pull/43552

PR #43536, opened 2026-01-27
Rework dependencies and extras + Remove outdated `templates` folder

Cyrilvallez:
# What does this PR do?

The dependencies and extras have almost never been touched and are quite old.
We are suffering quite a bit from the following:
- duplicated dependencies (e.g. being still in extra etc even if they are main dependencies, such as tokenizers or tqdm, or an extra needing 2 extras that are redundant)
- dependencies upon non-used library (e.g. `ftfy`)
- dependencies that are already main dependencies of REQUIRED libs (i.e. `hf_xet` il only ever used as a dep of `huggingface_hub`, not directly in Transformers)


templates/adding_a_new_model/open_model_proposals/ADD_BIG_BIRD.md
ArthurZucker, line 1: haha this is a bit bold, we do need to update it tho so down to remove! (fyi @itazap )

ArthurZucker (approved):
lgtm

ArthurZucker:
Can you run the docker file generation btw to make sure we test it

templates/adding_a_new_model/open_model_proposals/ADD_BIG_BIRD.md
Cyrilvallez: Yes, quite outdated and redundant with `add_new_model.md` in the doc , but can still probably be used as a ref if we want to change the doc later on as well, so maybe cc @stevhliu as well just in case - let's keep in mind we had that once if we want to comeback to it

Cyrilvallez:
All right, dockers have been re-built and run on this PR, and all failed tests have been seen elsewhere and are due to the recent CI difficulties.
Will rebase to remove the change to dockerfile, then merge it

PR #43537, opened 2026-01-27
Operation Green CI II

Rocketknight1:
We get flaky generation tests in the CI a lot. A very common cause is the `has_similar_generate_outputs` helper function - this function is called by many tests, but it has very narrow tolerances by default in the CI (`atol=1e-5, rtol=1e-5`). Example here: https://app.circleci.com/jobs/github/huggingface/transformers/2127052

Especially for `bfloat16` models, this seems very likely to cause unnecessary errors, so I expanded the tolerances to `1e-3`.

Example failing run [here](https://app.circleci.com/pipelines/github/huggingface/transformers/161587/workflows/f87424b2-d037-4309-a864-08a8d62ae201/jobs/2127052)

### DIFF
--- a/tests/generation/test_utils.py
+++ b/tests/generation/test_utils.py
@@ -4820,7 +4820,10 @@ class GenerationIntegrationTests(unittest.TestCase):
         incremental_outputs = outputs_2b
 
         # The two sets of generated text and past kv should be equal to each other
-        atol, rtol = (1e-3, 1e-3) if getattr(model.config, "_experts_implementation", None) else (1e-5, 1e-5)
+        if is_moe_model(model.config):
+            atol = rtol = 1e-3
+        else:
+            atol = rtol = 1e-5
         self.assertTrue(has_similar_generate_outputs(traditional_outputs, incremental_outputs, atol=atol, rtol=rtol))
         cache1, cache2 = traditional_outputs.past_key_values, incremental_outputs.past_key_values
         for idx in range(len(cache1)):
--- a/tests/generation/test_utils.py
+++ b/tests/generation/test_utils.py
@@ -1179,7 +1179,9 @@ class GenerationTesterMixin:
             else:
                 atol = rtol = 1e-5
             if not has_complex_embeds_computation:
-                self.assertTrue(has_similar_generate_outputs(outputs_from_ids, outputs_from_embeds, atol=atol, rtol=rtol))
+                self.assertTrue(
+                    has_similar_generate_outputs(outputs_from_ids, outputs_from_embeds, atol=atol, rtol=rtol)
+                )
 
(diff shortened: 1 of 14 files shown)


tests/generation/test_utils.py
```diff
@@ -5074,7 +5074,7 @@ def test_update_candidate_strategy_with_matches_1(self, sklearn_available):
             self.assert_no_sklearn()
 
 
-def has_similar_generate_outputs(output_1, output_2, atol=1e-5, rtol=1e-5) -> bool:
+def has_similar_generate_outputs(output_1, output_2, atol=1e-3, rtol=1e-3) -> bool:
```
vasqu: Not feeling super confident about this, it's likely due to moe model's using `grouped_mm` which is inherently a bit flaky

It would be nicer to isolate this to moes only

Rocketknight1: Done! I check `_experts_implementation` to decide if a model is an MoE or not

tests/generation/test_utils.py
```diff
@@ -717,12 +721,12 @@ def test_assisted_decoding_matches_greedy_search(self, assistant_type):
...
             output_assisted = model.generate(**generation_kwargs, **inputs_dict, **logits_processor_kwargs)
 
-            # default values of `has_similar_generate_outputs`
-            atol, rtol = 1e-5, 1e-5
             # `gpt_oss` seems to have larger differences on CPU every other generated tokens, sth. like
             # 1e-9, 1e-5, 1e-9, 1e-5. While on GPU, they are all very small 1e-9.
-            if model.config.model_type == "gpt_oss" and torch_device == "cpu":
-                atol, rtol = 1e-4, 1e-4
+            if is_moe_model(config):
+                atol = rtol = 1e-3
```
vasqu, line 727: We are raising this arbitrarily, imo it wouldn't hurt to check for `1e-4` first

Nit: would it make sense to make this to a small helper instead of always having this if else everywhere?

tests/models/cpmant/test_modeling_cpmant.py
```diff
@@ -205,7 +205,7 @@ def test_inference_causal(self):
 
     @tooslow
     def test_simple_generation(self):
-        model_path = "openbmb/cpm-ant-10b"
+        model_path = "Rocketknight1/cpm-ant-10b-testing"
```
ArthurZucker: can we push to hf-internal-testing?

ArthurZucker (approved):

tests/models/cpmant/test_modeling_cpmant.py
Rocketknight1: Done!

PR #43542, opened 2026-01-28
fix: output router capture wrong router logits in qwen moe models

ITcarrot:
# What does this PR do?

This PR fixes a bug in the router implementation of several MoE models (Qwen Moe like models, `Olmoe`, `FlexOlmo`). 

Previously, the raw `router_logits` were being overwritten by the result of the softmax operation:
```python
router_logits = F.linear(hidden_states, self.weight)
router_logits = torch.nn.functional.softmax(router_logits, dtype=torch.float, dim=-1)
```
This behavior destroys the raw logits information, which is often required for:
1. Casting returned `router_top_value` back to input type in `router_top_value = router_top_value.to(router_logits.dtype)`
2. Calculating auxiliary losses (e.g., load balancing loss) which typically operate on raw logits.
3. Returning correct router_logits in the model output if requested.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@Cyrilvallez @IlyasMoutawwakil

### DIFF
--- a/src/transformers/models/flex_olmo/modeling_flex_olmo.py
+++ b/src/transformers/models/flex_olmo/modeling_flex_olmo.py
@@ -343,8 +343,8 @@ class FlexOlmoTopKRouter(nn.Module):
     def forward(self, hidden_states):
         hidden_states = hidden_states.reshape(-1, self.hidden_dim)
         router_logits = F.linear(hidden_states, self.weight)  # (seq_len, num_experts)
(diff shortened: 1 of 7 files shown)


vasqu (approved):
Good catch yes, the logits should not go through softmax :sweat_smile: 

Could we have a small test added to qwen2 moe (since everyone else inherits from this) + are there any more such cases?

PR #43547, opened 2026-01-28
fix the error of tests/quantization/fbgemm_fp8/test_fbgemm_fp8.py::Fb…

sywangyi:
…gemmFp8Test::test_change_loading_attributes

- quantization: @SunMarc @MekkCyber
input_scale_ub is not correctly updated, since replace_with_fbgemm_fp8_linear is called under meta device.


SunMarc (approved):
Thanks for spotting this !

sywangyi:
The failed CI tests should have nothing to do with this PR.

PR #43548, opened 2026-01-28
[Qwen2VL] Fix missing min_pixels/max_pixels attributes in fast image processor

tomaszcichy98:
## What does this PR do?

Fixes a bug where `Qwen2VLImageProcessorFast` doesn't set `min_pixels` and `max_pixels` instance attributes, breaking compatibility with code that expects these attributes.

## The Problem

The slow processor (`Qwen2VLImageProcessor`) sets `self.min_pixels` and `self.max_pixels` as instance attributes. The fast processor converts these to `size["shortest_edge"]` and `size["longest_edge"]` but never sets the original attributes.

This causes vLLM (and potentially other libraries) to fail when loading Qwen2-VL/Qwen2.5-Omni models:

```
File "transformers/models/qwen2_vl/image_processing_qwen2_vl.py", line 92, in smart_resize
    if h_bar * w_bar > max_pixels:
TypeError: '>' not supported between instances of 'int' and 'NoneType'
```

### Why config values don't help

Even if `preprocessor_config.json` contains `min_pixels` and `max_pixels` values, this doesn't fix the issue because:

1. `use_fast: false` in the config is **ignored** - transformers 5.0 always auto-upgrades to the fast processor
2. The fast processor receives these values as kwargs but only stores them in `self.size` dict, not as instance attributes
3. Code accessing `processor.min_pixels` gets `None` instead of the config value

## The Fix

Set `min_pixels` and `max_pixels` attributes after `super().__init__()` to maintain backward compatibility:

```python
super().__init__(size=size, **kwargs)
self.min_pixels = self.size.get("shortest_edge")
self.max_pixels = self.size.get("longest_edge")
```

## Verification

```python

# Before fix
processor = AutoProcessor.from_pretrained("Qwen/Qwen2.5-VL-7B-Instruct")
print(processor.image_processor.min_pixels)  # None
print(processor.image_processor.max_pixels)  # None

# After fix
print(processor.image_processor.min_pixels)  # 3136
print(processor.image_processor.max_pixels)  # 12845056
```

### DIFF
--- a/src/transformers/models/glm_image/image_processing_glm_image_fast.py
+++ b/src/transformers/models/glm_image/image_processing_glm_image_fast.py
@@ -115,6 +115,9 @@ class GlmImageImageProcessorFast(BaseImageProcessorFast):
             raise ValueError("size must contain 'shortest_edge' and 'longest_edge' keys.")
 
         super().__init__(size=size, **kwargs)
+        # Maintain backward compatibility by setting min_pixels/max_pixels attributes
+        self.min_pixels = self.size.get("shortest_edge")
+        self.max_pixels = self.size.get("longest_edge")
 
     def _further_process_kwargs(
         self,
--- a/src/transformers/models/qwen2_vl/image_processing_qwen2_vl_fast.py
+++ b/src/transformers/models/qwen2_vl/image_processing_qwen2_vl_fast.py
@@ -81,6 +81,9 @@ class Qwen2VLImageProcessorFast(BaseImageProcessorFast):
             raise ValueError("size must contain 'shortest_edge' and 'longest_edge' keys.")
 
         super().__init__(size=size, **kwargs)
+        # Maintain backward compatibility by setting min_pixels/max_pixels attributes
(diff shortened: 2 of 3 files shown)


Rocketknight1:
cc @zucchini-nlp maybe?

zucchini-nlp (commented):
It was removed as part of v5 release in favor of `self.size`. I believe the error is raised from within vLLM code? 

We can ask vLLM team to update the code on their side, not sure if they already bumped to v5

zucchini-nlp:
fyi @hmellor ,can be added in https://github.com/vllm-project/vllm/pull/30566 if it's indeed not working

tomaszcichy98:
> It was removed as part of v5 release in favor of `self.size`. I believe the error is raised from within vLLM code?
> 
> We can ask vLLM team to update the code on their side, not sure if they already bumped to v5

Yeah it is coming from vLLM @zucchini-nlp

hmellor:
This is already fixed in vLLM https://github.com/vllm-project/vllm/pull/33208

PR #43549, opened 2026-01-28
[kernels] exception handling for fa kernels

MekkCyber:
# What does this PR do?

Before we were just silently skipping parameters that are passed by the user like `s_aux` in case they are not supported by the attention backend specified, it would be better to raise an exception instead.

cc @danieldk

### DIFF
--- a/src/transformers/modeling_flash_attention_utils.py
+++ b/src/transformers/modeling_flash_attention_utils.py
@@ -533,11 +533,24 @@ def _process_flash_attention_kwargs(
         flash_kwargs (`dict`):
             A dict of kwargs that are requested and supported.
     """
+
+    user_kwargs = {
+        "dropout_p": dropout,
+        "window_size": sliding_window,
+        "deterministic": deterministic,
+        "softcap": softcap,
+        "s_aux": s_aux,
+    }
+    # Note 'window_size' in supports_mapping maps to our 'sliding_window' param
+    for k, v in user_kwargs.items():
+        if not supports_mapping[k] and v is not None:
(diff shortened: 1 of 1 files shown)


src/transformers/modeling_flash_attention_utils.py
```diff
@@ -533,6 +533,21 @@ def _process_flash_attention_kwargs(
         flash_kwargs (`dict`):
             A dict of kwargs that are requested and supported.
     """
+
+    user_kwargs = {
+        "dropout_p": dropout,
+        "window_size": sliding_window,
+        "deterministic": deterministic,
```
danieldk, line 540: I am not 100% sure about this one. If someone requests deterministic output, it might still be ok to emit non-deterministic output with a warning?

danieldk (approved):
Looks good to me, but I am not very familiar with this code, so someone else should review it too.

src/transformers/modeling_flash_attention_utils.py
MekkCyber: Yes it depends I think! I will wait for @ArthurZucker and @vasqu inputs since they maintain the fa implementation

vasqu: Don't feel super confident about this
- dropout is fair
- sliding window might still unintentionally work if `key_length <= sliding_window`
- deterministic has the env variable as well
- s_aux is only supported in gpt oss and we check that people use it https://github.com/huggingface/transformers/blob/1be63d94f63c98e9a94de40b2a2fa4d12d9767f0/src/transformers/models/gpt_oss/configuration_gpt_oss.py#L131

Additionally, I'm also for being more lenient at least on deterministic (warning)

TL;DR: The conditions below should be counted in when we raise the error + a small test would be nice

PR #43556, opened 2026-01-28
[`Modular`] Allow to add new bases that are not present in the inherited class

vasqu:
As per title, it is an edge case for lasr only at the moment but #41212 will also need it. 

I think this is the easiest solution (and fastest) because there are several edge cases with additional bases.

### DIFF
--- a/src/transformers/models/emu3/modular_emu3.py
+++ b/src/transformers/models/emu3/modular_emu3.py
@@ -866,7 +866,7 @@ class Emu3ImageVocabularyMapping:
         return img_tokens.to(device)
 
 
-class Emu3PreTrainedModel(ChameleonPreTrainedModel, Emu3VQVAE):
+class Emu3PreTrainedModel(ChameleonPreTrainedModel):
     _no_split_modules = [
         "Emu3DecoderLayer",
     ]
--- a/src/transformers/models/glm_ocr/modular_glm_ocr.py
+++ b/src/transformers/models/glm_ocr/modular_glm_ocr.py
@@ -184,7 +184,7 @@ class GlmOcrTextConfig(Glm4vTextConfig):
         super().__init__(**super_kwargs)
 
 
-class GlmOcrConfig(Glm4vConfig, nn.Module):
+class GlmOcrConfig(Glm4vConfig):
     r"""
     This is the configuration class to store the configuration of a [`GlmOcrModel`]. It is used to instantiate a
     GLM-OCR model according to the specified arguments, defining the model architecture. Instantiating a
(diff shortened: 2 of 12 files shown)


src/transformers/models/lasr/modular_lasr.py
vasqu, line 1: Lasr was the only model with this issue but given that the new eomt model also has it, I'd rather avoid many exceptions and integrate it as full feature, mainly for mixins

src/transformers/models/emu3/modular_emu3.py
```diff
@@ -866,7 +866,7 @@ def convert_bpe2img(self, img_batch: torch.Tensor) -> torch.Tensor:
         return img_tokens.to(device)
 
 
-class Emu3PreTrainedModel(ChameleonPreTrainedModel, Emu3VQVAE):
+class Emu3PreTrainedModel(ChameleonPreTrainedModel):
```
Cyrilvallez, line 869: Really not sure why it was there, was never supposed to work to inherit from 2 like that

src/transformers/models/glm_ocr/modular_glm_ocr.py
```diff
@@ -184,7 +184,7 @@ def __init__(
         super().__init__(**super_kwargs)
 
 
-class GlmOcrConfig(Glm4vConfig, nn.Module):
+class GlmOcrConfig(Glm4vConfig):
```
Cyrilvallez, line 187: ????? It's a config....

src/transformers/models/qwen2_moe/modular_qwen2_moe.py
```diff
@@ -129,9 +129,9 @@ def forward(self, hidden_states: torch.Tensor) -> tuple[torch.Tensor, torch.Tens
         return expert_output
 
 
-class Qwen2MoeDecoderLayer(LlamaDecoderLayer, nn.Module):
+class Qwen2MoeDecoderLayer(LlamaDecoderLayer):
     def __init__(self, config: Qwen2MoeConfig, layer_idx: int):
-        nn.Module.__init__()
+        nn.Module.__init__(self)
```
Cyrilvallez, line 134: No need for additional base, it's part of MRO

src/transformers/models/sam2/modeling_sam2.py
```diff
@@ -914,7 +914,7 @@ def forward(
         return attn_output, attn_weights
 
 
-class Sam2TwoWayAttentionBlock(nn.Module):
+class Sam2TwoWayAttentionBlock(GradientCheckpointingLayer):
```
Cyrilvallez, line 917: All of those were inherited with `GradientCheckpointingLayer` in modular, I suppose to add it. It was not working until now

Cyrilvallez (approved):
That's a bit of a self-approval but we discussed with @vasqu offline and he cannot approve his own PR haha 🤝

vasqu:
Merging since it unblocks another model PR and CI is green :exploding_head:

PR #43557, opened 2026-01-28
fix(converter): speed up `MistralConverter.extract_vocab_merges_from_model`

tarekziade:
# What does this PR do?

This test was taking 3 to 4mn on the CI  `test_voxtral_tokenizer_converts_from_tekken`

https://github.com/huggingface/transformers/blob/main/tests/models/auto/test_tokenization_auto.py#L213

Turns out we have a super slow function on large vocabularies (100k+ tokens)

This patch improves `MistralConverter.extract_vocab_merges_from_model`

Performance improvement:
- Before: O(n) linear search for each lookup, causing O(n²) behavior overall - 3/4mn
- After: O(1) dictionary lookup, reducing to O(n log n) for sorting - 1 second


src/transformers/convert_slow_tokenizer.py
```diff
@@ -1943,6 +1942,7 @@ def token_bytes_to_string(b):
             vocab[token.content] = idx
         bpe_ranks = [base64.b64decode(k["token_bytes"]) for k in bpe_ranks]
         rank_set = set(bpe_ranks)
+        token_to_rank = {token: rank for rank, token in enumerate(bpe_ranks)}
```
Cyrilvallez, line 1946: Very slightly concerned about `rank_set = set(bpe_ranks)` which might suggest that some items are duplicated? In which case the index in `token_to_rank` will be the last occurence and `bpe_rank.index` will be the first - but probably out of scope?

Cyrilvallez (approved):
Very nice catch!
Small detail, not sure if we can ever have duplicates in `bpe_ranks` or not! Maybe cc @ArthurZucker here? Feel free to merge if it's not the case

ArthurZucker (approved):
yep ty for the quick fix

PR #43566, opened 2026-01-28
Fix Wav2vec and a few others

Cyrilvallez:
# What does this PR do?

https://github.com/huggingface/transformers/pull/43523 broke Wav2Vec and a few others as they have their own `tie_weights`, which is actually not tying the weights... Who would have thought that changing the purpose of a method in a few models could create a lot of problems later on?? 🥲🥲😅😅


Rocketknight1 (approved):

Cyrilvallez:
CI still not triggering... Merging this one!

PR #43567, opened 2026-01-28
[`Sam`] Fixup training flags

vasqu:
Kind of a follow up to #43556 which fixed a bunch of wrong inheritance with gradient ckpting. Just dummy exchanged the proper flags and I didn't have to skip a lot of these tests


yonigozlan (approved):
Great to less tests skipped :)

PR #43569, opened 2026-01-28
[CB] Refactor logic for inputs and outputs outside of the main API

remi-or:
This PR refactors all of the logic linked to the inputs and outputs of continuous batching into a new class `ContinuousBatchingIOs` itself in a new file. This will be quite useful when we introduce async scheduling, because there will be two sets of host-side IOs and one set of device-side IOs. 
To avoid having this major refactoring AND the async logic in the same PR, I think it would be good if we can first merge this part, which does not change much, and then add the async logic.
This PR also makes the test file `ty` compatible and adds a bit more documentation.


ArthurZucker (approved):
this makes the code a bit more stateful, but it might be better. I am mostly thinking of multi-gpu runs, this should make stuff simpler.

PR #43583, opened 2026-01-29
[MoE] Use int input for histc on CUDA to support deterministic algorithms

YangKai0616:
# What does this PR do?

`torch.histc` with deterministic algorithms enabled behaves differently across devices:

**CPU**: only supports float input
**CUDA**: only supports int input

This PR updates `grouped_mm_experts_forward` to use the appropriate dtype based on device type, enabling compatibility with `torch.use_deterministic_algorithms(True)` on CUDA.

Simple reproduction script:
```
import torch

def test_histc_deterministic(device: str):
    print(f"\n=== Testing on {device.upper()} ===")
    
    num_experts = 8
    expert_ids = torch.randint(0, num_experts, (100,), device=device)
    
    torch.use_deterministic_algorithms(True)
    
    # Test float input
    try:
        result = torch.histc(expert_ids.float(), bins=num_experts, min=0, max=num_experts - 1)
        print(f"✅ float input works with deterministic mode")
    except RuntimeError as e:
        print(f"❌ float input fails: {e}")
    
    # Test int input
    try:
        result = torch.histc(expert_ids, bins=num_experts, min=0, max=num_experts - 1)
        print(f"✅ int input works with deterministic mode")
    except (RuntimeError, NotImplementedError) as e:
        print(f"❌ int input fails: {e}")
    
    torch.use_deterministic_algorithms(False)

if __name__ == "__main__":
    print(f"PyTorch version: {torch.__version__}")
    
    test_histc_deterministic("cpu")
    
    if torch.cuda.is_available():
        test_histc_deterministic("cuda")
```

Hi @vasqu , please help review this PR, thanks!


src/transformers/integrations/moe.py
```diff
@@ -226,7 +226,9 @@ def grouped_mm_experts_forward(
 
     # Compute offsets for grouped_mm
     # using histc instead of bincount to avoid cuda graph issues
-    num_tokens_per_expert = torch.histc(expert_ids_g.float(), bins=self.num_experts, min=0, max=self.num_experts - 1)
+    # With deterministic algorithms, CPU only supports float input, CUDA only supports int input.
+    histc_input = expert_ids_g.float() if device.type == "cpu" else expert_ids_g
```
IlyasMoutawwakil: we can also apply .int() in the case of cuda to be sure
```suggestion
    histc_input = expert_ids_g.float() if device.type == "cpu" else expert_ids_g.int()
```

IlyasMoutawwakil (commented):
LGTM ! thanks for the fix !

IlyasMoutawwakil (approved):

PR #43584, opened 2026-01-29
Fix mistral checkpoint loading in `utils/fetch_hub_objects_for_ci.py`: avoid too many requests and/or timeout

ydshieh:
# What does this PR do?

Just reuse the local files as much as possible


tarekziade (approved):
LGTM, thx

PR #43261, opened 2026-01-13
Fix PEFT x MoEs

ArthurZucker:
# What does this PR do?
Fixes #42491

This should serve as an example of how the weight loader can be re-used in other project. 
The content is probably gonna be upstreamed to peft!
Current status:
image

What to expect:
1. You already loaded a transformers model
2. You want to load only the Peft adapters:
    a. we check the weight conversion mapping
    b. if there are ops, we replace them with mapped peft ops
    c. we just collect lora_A and lora_B together, process them like so:
```python
class PeftConcatenate(Concatenate):
    """Convert per-expert LoRA weights to merged MoE weights using SVD."""
    @torch.no_grad
    def convert(
        self, input_dict: dict[str, list[torch.Tensor]], source_patterns: list[str], target_patterns: list[str], **kwargs
    ) -> dict[str, list[torch.Tensor]]:
        lora_a_out = []
        lora_b_out = []
        for k,v in input_dict.items():
            if "lora_A" in k:
                lora_a_out.append(v)
            elif "lora_B" in k:
                lora_b_out.append(v)
        lora_a_out = torch.cat(lora_a_out, dim=0)
        for i in range(len(lora_b_out)):
            lora_b_out.append(torch.block_diag(lora_b_out[0][i], lora_b_out[1][i]))
        lora_b_out = torch.stack(lora_b_out[2:], dim=0)
        return {
            target_patterns[0]+".lora_A.weight": [lora_a_out],
            target_patterns[0]+".lora_B.weight": [lora_b_out],
        }
```

        d. The output fused gate_up.lora_A/B are loaded in the model

CF: (Credits to @BenjaminBossan for the pic)
image

PR #43585, opened 2026-01-29
Regex post processing in loading

Cyrilvallez:
# What does this PR do?

The change introduced in https://github.com/huggingface/transformers/pull/43261 with `__setattr__` is quite dangerous, as `source_patterns` and `target_patterns` live together and cannot be unentangled from one another if there is a capturing group. 
Thus, changing one with `__setattr__` without changing the other, or simply changing the wrong one first can lead to very unexpected behavior and wrong pattern matching/replacing.
For this reason, it's better to reinstantiate an object directly in `peft` instead of switching the targets/sources.

### DIFF
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -694,7 +694,7 @@ class WeightConverter(WeightTransform):
     operations: list[ConversionOps] = field(default_factory=list, repr=False)
 
     def __post_init__(self):
-        super().__post_init__()
+        WeightTransform.__post_init__(self)
         if bool(len(self.source_patterns) - 1) + bool(len(self.target_patterns) - 1) >= 2:
             # We allow many-to-many only if we use an internal operation that can handle it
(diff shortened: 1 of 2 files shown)


src/transformers/core_model_loading.py
```diff
@@ -520,46 +520,13 @@ class WeightTransform:
...
-                if r"(.+)" in pattern:
-                    pattern = pattern.replace(r"(.+)", r"\1")
-                normalized.append(pattern)
-            object.__setattr__(self, name, normalized)
-            return
+    def __setattr__(self, name, value):
+        # We do not allow to re-set the patterns, as they are linked between each other and changing one
+        # without the other can mess-up with the capturing groups/compiled sources
+        if name in ("source_patterns", "target_patterns") and hasattr(self, name):
+            raise ValueError(f"Cannot assign to field {name}")
```
ArthurZucker:
```suggestion
            raise ValueError(f"Cannot assign to field {name}, you should create a new instance")
```

ArthurZucker (approved):
Fair enough! ty

PR #43586, opened 2026-01-29
Increase default ReadTimeout in tests

Wauplin:
# What does this PR do?

This PR sets `HF_HUB_DOWNLOAD_TIMEOUT=60` as environment variable when running tests. I've also added [`pytest-env`](https://github.com/pytest-dev/pytest-env) as a dev dependency to do that.

I've also reverted https://github.com/huggingface/transformers/pull/43560 which is not needed anymore.

**Note:** there might still be some `ReadTimeout`s in CI when running scripts that are not run with `pytest`. I preferred not to touch that part as I don't have a broad understanding on what the CI does but any failing workflow (due to timeouts) should be fixable with a `HF_HUB_DOWNLOAD_TIMEOUT=60` env variable.

**Note 2:** this PR requires [huggingface_hub v1.3.5](https://pypi.org/project/huggingface-hub/1.3.5/) to work. I don't think we should pin this version as minimal dependency though, since it's just to fix some tests. In theory the CI will always pull the latest version so it should be fine.


Rocketknight1 (approved):
LGTM!

PR #43589, opened 2026-01-29
Simplify loading structure

Cyrilvallez:
# What does this PR do?

As per the title. This simplifies the use of `LoadStateDictInfo` a bit everywhere, and makes it clear that the entries within the struct are mutated in-place all the time.

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -4079,7 +4079,7 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
             download_kwargs=download_kwargs,
         )
         loading_info, disk_offload_index = cls._load_pretrained_model(model, state_dict, checkpoint_files, load_config)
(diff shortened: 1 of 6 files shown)


src/transformers/modeling_utils.py
```diff
@@ -4227,73 +4209,49 @@ def _load_pretrained_model(
...
-            unexpected_keys=unexpected_keys,
-            mismatched_keys=mismatched_keys,
-            disk_offload_index=disk_offload_index,
-            error_msgs=error_msgs,
-            conversion_errors=conversion_errors,
-        )
-
-    @staticmethod
-    def _finalize_load_state_dict(
-        model,
```
ArthurZucker, line 4243: run peft tests pretty sure we need the split between the tow, it allows for better granularity

ArthurZucker (commented):
LGTM otherwise

Cyrilvallez:
Alright, resplit the 2 methods indeed!
Since nothing really changed after last review and the test failing is unrelated, merging!

PR #43591, opened 2026-01-29
custom tok init fix

itazap:
custom tokenizers fail on super._init_() call that tries to update vocab size before all vocab attrs are defined

### DIFF
--- a/tests/models/auto/test_tokenization_auto.py
+++ b/tests/models/auto/test_tokenization_auto.py
@@ -461,7 +461,7 @@ class AutoTokenizerTest(unittest.TestCase):
 
     @slow
     def test_custom_tokenizer_init(self):
-        tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen-VL", trust_remote_code=True)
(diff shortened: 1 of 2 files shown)


tests/models/auto/test_tokenization_auto.py
```diff
@@ -473,6 +474,12 @@ def test_from_pretrained_dynamic_tokenizer(self):
         self.assertIsNot(tokenizer.__class__, reloaded_tokenizer.__class__)
         self.assertTrue(reloaded_tokenizer.special_attribute_present)
 
+    @slow
+    def test_custom_tokenizer_init(self):
+        tokenizer = AutoTokenizer.from_pretrained("Qwen/Qwen-VL", trust_remote_code=True)
```
ArthurZucker: works, but let's make sure we pin latest commit (security)

ArthurZucker (commented):
`test_tokenizer_defaults_preserved_by_kwargs` I'm not sure its related no?

ArthurZucker (approved):

itazap:
run-slow: auto

PR #43593, opened 2026-01-29
fix: AttributeError for Qwen3_omni_moe

Vallabh-1504:
# What does this PR do?

This PR fixes a crash when initializing `Qwen3OmniMoeTalkerCodePredictorConfig` due to a missing attribute reference.

Specifically, it:
1.  **Removes** the reference to the non-existent `use_sliding_window` attribute, which was causing an `AttributeError`.
2.  **Adds** the missing `max_window_layers` initialization (defaulting to `28`). The existing `layer_types` logic relies on this attribute, and without it, `sliding_window` causes a secondary crash.
3.  **Adds** a new test case (`test_code_predictor_config_init`) to ensure the configuration initializes correctly. 

Fixes #43531

## Before submitting
      Pull Request section?
      to it if that's the case. (Issue #43531)

### DIFF
--- a/src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
+++ b/src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
@@ -582,8 +582,6 @@ class Qwen3OmniMoeTalkerCodePredictorConfig(PreTrainedConfig, RotaryEmbeddingCon
         eos_token_id: int | None = None,
         **kwargs,
     ):
-        self.sliding_window = sliding_window
-        self.max_window_layers = max_window_layers
         self.num_code_groups = num_code_groups
         self.vocab_size = vocab_size
         self.max_position_embeddings = max_position_embeddings
--- a/src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
+++ b/src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
@@ -573,6 +573,7 @@ class Qwen3OmniMoeTalkerCodePredictorConfig(PreTrainedConfig, RotaryEmbeddingCon
         rope_parameters: int | None = None,
         attention_bias: bool | None = False,
         sliding_window: int | None = None,
+        max_window_layers: int | None = 28,
         layer_types: list[str] | None = None,
         attention_dropout: int | None = 0,
         num_code_groups: int | None = 32,
@@ -582,6 +583,7 @@ class Qwen3OmniMoeTalkerCodePredictorConfig(PreTrainedConfig, RotaryEmbeddingCon
         **kwargs,
     ):
         self.sliding_window = sliding_window
+        self.max_window_layers = max_window_layers
         self.num_code_groups = num_code_groups
         self.vocab_size = vocab_size
         self.max_position_embeddings = max_position_embeddings
@@ -589,7 +591,8 @@ class Qwen3OmniMoeTalkerCodePredictorConfig(PreTrainedConfig, RotaryEmbeddingCon
         self.intermediate_size = intermediate_size
         self.num_hidden_layers = num_hidden_layers
         self.num_attention_heads = num_attention_heads
-        self.sliding_window = sliding_window if self.use_sliding_window else None
+        self.sliding_window = sliding_window
+        self.max_window_layers = max_window_layers
 
         # for backward compatibility
         if num_key_value_heads is None:
--- a/src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py
+++ b/src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py
@@ -474,8 +474,6 @@ class Qwen3OmniMoeTalkerCodePredictorConfig(Qwen3Config):
         eos_token_id: int | None = None,
         **kwargs,
     ):
-        self.sliding_window = sliding_window
-        self.max_window_layers = max_window_layers
         self.num_code_groups = num_code_groups
         super().__init__(
             vocab_size,
@@ -505,6 +503,7 @@ class Qwen3OmniMoeTalkerCodePredictorConfig(Qwen3Config):
         )
         del self.use_sliding_window
         self.sliding_window = sliding_window
+        self.max_window_layers = max_window_layers
 
 
 class Qwen3OmniMoeTalkerTextConfig(Qwen3MoeConfig):
--- a/src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py
+++ b/src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py
@@ -465,6 +465,7 @@ class Qwen3OmniMoeTalkerCodePredictorConfig(Qwen3Config):
         rope_parameters: int | None = None,
         attention_bias: bool | None = False,
         sliding_window: int | None = None,
+        max_window_layers: int | None = 28,
         layer_types: list[str] | None = None,
         attention_dropout: int | None = 0,
         num_code_groups: int | None = 32,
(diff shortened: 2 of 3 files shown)


Rocketknight1:
cc @vasqu since you commented on the issue!

src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
```diff
@@ -573,6 +573,7 @@ def __init__(
         rope_parameters: int | None = None,
         attention_bias: bool | None = False,
         sliding_window: int | None = None,
+        max_window_layers: int = 28,
```
vasqu: Seems irrelevant?

tests/models/qwen3_omni_moe/test_configuration_qwen3_omni_moe.py
vasqu, line 1: The test should be still under `test_modeling_xxx`

vasqu (commented):
`max_window_layers` seems like an unnecessary attribute added + let's move the test, please no new files

Vallabh-1504:
hi, @vasqu, thanks for the review.
when I was running pytest I was getting an error regarding `max_window_layers`.
This attribute is used in the `layer_types` list comprehension (around line 612).

```
    def __getattribute__(self, key):
        if key != "attribute_map" and key in super().__getattribute__("attribute_map"):
            key = super().__getattribute__("attribute_map")[key]
>       return super().__getattribute__(key)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AttributeError: 'Qwen3OmniMoeTalkerCodePredictorConfig' object has no attribute 'max_window_layers'

src\transformers\configuration_utils.py:163: AttributeError
```

src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
```diff
@@ -582,14 +583,14 @@ def __init__(
...
         self.sliding_window = sliding_window
+        self.max_window_layers = max_window_layers
         self.num_code_groups = num_code_groups
         self.vocab_size = vocab_size
         self.max_position_embeddings = max_position_embeddings
         self.hidden_size = hidden_size
         self.intermediate_size = intermediate_size
         self.num_hidden_layers = num_hidden_layers
         self.num_attention_heads = num_attention_heads
-        self.sliding_window = sliding_window if self.use_sliding_window else None
```
vasqu, line 591: Ok I checked what happened and it seems it was unintentionally added in #41541, but since I'm not super familiar with this model I'd rather wait for @zucchini-nlp to answer here

Imo we should just do `self.sliding_window = sliding_window` (`use_sliding_window` was never used at all and should be removed from the docstring) - `max_window_layers` should be removed alongside it (not reintroduced)

vasqu: Also the changes should be done in modular and then reapplied via `python utils/modular_model_converter.py qwen3_omni_moe`

src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
```diff
@@ -582,14 +583,14 @@ def __init__(
         **kwargs,
     ):
         self.sliding_window = sliding_window
+        self.max_window_layers = max_window_layers
```
zucchini-nlp: it is a recurring pattern in many qwen-multimodal models and I think authors incorrectly copied it when adding Qwen3-Omni-MoE. The official checkpoints have `max_window_layers` saved in config therefore we don't see errore at inference

IMO we need to keep it to match the model's default behavior and to match with docstring

src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
zucchini-nlp: Indeed a bad copy, no need for `use_sliding_window`. Model always uses sliding layers together with full attention

zucchini-nlp:
@Vallabh-1504 hey, any updates on the PR? Would be great to merge it before the next planned release, which I believe should be around this week

Vallabh-1504:
hi @zucchini-nlp, I was working on it but hitting a bit of a wall.

I was applying changes through `modular_qwen3_omni_moe.py`, as previously i did manual changes which was wrong. but for some reason, the `modular_model_converter.py` keeps regenerating the `use_sliding_window` logic in the final config, even after i removed it from `modular_qwen3_omni_moe.py`.

kinda stuck in a loop where the generator keeps the code i'm trying to remove.

I also don't know what to do with the `max_window_layers`, as this is also not initialized anywhere.

Can you guide me through this. If i'm missing a step or if there's some cache with the converter?

zucchini-nlp:
Smth like this should work, we need to delete unused attributed explicitly and re-assign the "differing" attr. For `max_window_layers`, let's keep it without deleting

https://github.com/Vallabh-1504/transformers/blob/9041720191cdef9348f09f7c1695db606223d353/src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py#L571-L572

Vallabh-1504:
@zucchini-nlp, I have applied the changes as requested!

src/transformers/models/qwen3_omni_moe/configuration_qwen3_omni_moe.py
```diff
@@ -581,14 +583,16 @@ def __init__(
...
         self.num_code_groups = num_code_groups
         self.vocab_size = vocab_size
         self.max_position_embeddings = max_position_embeddings
         self.hidden_size = hidden_size
         self.intermediate_size = intermediate_size
         self.num_hidden_layers = num_hidden_layers
         self.num_attention_heads = num_attention_heads
-        self.sliding_window = sliding_window if self.use_sliding_window else None
+        self.sliding_window = sliding_window
+        self.max_window_layers = max_window_layers
```
zucchini-nlp, line 593: nit: duplicate

src/transformers/models/qwen3_omni_moe/modular_qwen3_omni_moe.py
```diff
@@ -473,6 +475,7 @@ def __init__(
         **kwargs,
     ):
         self.sliding_window = sliding_window
+        self.max_window_layers = max_window_layers
```
zucchini-nlp: not needed, super assigns it already which is why we got duplicates in auto-generated code  :)

tests/models/qwen3_omni_moe/test_modeling_qwen3_omni_moe.py
```diff
@@ -911,3 +912,30 @@ def test_small_model_integration_test_batch_flashatt2(self):
...
+
+
+class TestQwen3OmniMoeCodePredictorConfig(unittest.TestCase):
+    def test_code_predictor_config_init(self):
+        """
+        Test that Qwen3OmniMoeTalkerCodePredictorConfig initializes correctly
+        and accepts max_window_layers while removing use_sliding_window.
+        """
+
+        config = Qwen3OmniMoeTalkerCodePredictorConfig(
```
zucchini-nlp: ideally we need a complete model test for the 'TalkerModel' model. But I realize that the model doesn't follow transformers standards and we'll skip anyway many tests from `ModelTesterMixin`, or override a lot of them

I'm fine with deleting the current test in that case, @vasqu wdyt?

zucchini-nlp (approved):
One tiny comment and let's wait for @vasqu's review

tests/models/qwen3_omni_moe/test_modeling_qwen3_omni_moe.py
vasqu: Imo, we can add a small regression test under https://github.com/huggingface/transformers/blob/d75ecdb9f2bac143185591994f921b370d8ebf21/tests/models/qwen3_omni_moe/test_modeling_qwen3_omni_moe.py#L61 at least, we don't need a separate class for this

Just a tad weird because the naming is weird but if we don't test it we ought to repeat it in a refactor :grimacing:

vasqu (approved):
Don't have much to add onto @zucchini-nlp's comments, I'd just prefer to move the test under the general tester class

PR #43594, opened 2026-01-29
[CB] [Serve] Fix broken serve tests

remi-or:
This PR fixes two issues related to CB and serve:
- when the `max_length` argument as set to `None` when creating a request, CB had an error because it always expected max_length to be an int. This is now an expected behavior, and we have a hidden variable to avoid checking if the variable is None
- the CB streamer did not yield the last generated token when the request finished. This was not an issue before the last PR, because the result of each request was yielded twice (with the same amount of tokens). Now that this is fixed, it revealed the bug related to the streamer, which is now fixed.


ArthurZucker (approved):
thanks!

PR #43603, opened 2026-01-29
Fix unintended Hub metadata calls from _patch_mistral_regex

vaibhav-research:
# What does this PR do?

`TokenizersBackend._patch_mistral_regex()` is a Mistral-specific tokenizer patch, but the current implementation may call `huggingface_hub.model_info()` during detection. That triggers an HTTP request to `/api/models/<repo_id>` and can occur even for non-Mistral repos in environments where outbound network calls are blocked.

This PR adds minimal guardrails:
- Return early when `local_files_only=True` or in offline mode.
- Return early for non-Mistral repo ids before calling `model_info()`.

This keeps the Mistral behavior unchanged while preventing unnecessary metadata network requests for non-Mistral models.

Fixes #43502 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@zucchini-nlp @ArthurZucker 

### DIFF
--- a/src/transformers/tokenization_utils_tokenizers.py
+++ b/src/transformers/tokenization_utils_tokenizers.py
@@ -1285,7 +1285,7 @@ class TokenizersBackend(PreTrainedTokenizerBase):
 
         if local_files_only or is_offline_mode():
             return tokenizer
-        
+
         if not is_local and isinstance(pretrained_model_name_or_path, str):
             _id = pretrained_model_name_or_path.lower()
             # Only continue for likely Mistral repos/ids; otherwise no-op.
--- a/src/transformers/tokenization_utils_tokenizers.py
+++ b/src/transformers/tokenization_utils_tokenizers.py
@@ -1283,21 +1283,15 @@ class TokenizersBackend(PreTrainedTokenizerBase):
 
         from transformers.utils.hub import cached_file
 
-        if local_files_only or is_offline_mode():
-            return tokenizer
-
-        if not is_local and isinstance(pretrained_model_name_or_path, str):
-            _id = pretrained_model_name_or_path.lower()
-            # Only continue for likely Mistral repos/ids; otherwise no-op.
-            if not (_id.startswith("mistralai/") or "mistral" in _id):
-                return tokenizer
-
         def is_base_mistral(model_id: str) -> bool:
             model = model_info(model_id)
             if model.tags is not None:
                 if re.search("base_model:.*mistralai", "".join(model.tags)):
                     return True
             return False
+
+        if local_files_only or is_offline_mode():
+            is_local = True
 
         if pretrained_model_name_or_path is not None and (
             is_local or (not is_local and is_base_mistral(pretrained_model_name_or_path))
--- a/src/transformers/tokenization_utils_tokenizers.py
+++ b/src/transformers/tokenization_utils_tokenizers.py
@@ -1283,15 +1283,21 @@ class TokenizersBackend(PreTrainedTokenizerBase):
 
         from transformers.utils.hub import cached_file
 
+        if local_files_only or is_offline_mode():
+            return tokenizer
+        
+        if not is_local and isinstance(pretrained_model_name_or_path, str):
+            _id = pretrained_model_name_or_path.lower()
+            # Only continue for likely Mistral repos/ids; otherwise no-op.
+            if not (_id.startswith("mistralai/") or "mistral" in _id):
+                return tokenizer
+
         def is_base_mistral(model_id: str) -> bool:
             model = model_info(model_id)
             if model.tags is not None:
                 if re.search("base_model:.*mistralai", "".join(model.tags)):
                     return True
             return False
-
-        if is_offline_mode():
-            is_local = True
 
         if pretrained_model_name_or_path is not None and (
             is_local or (not is_local and is_base_mistral(pretrained_model_name_or_path))
--- a/src/transformers/tokenization_utils_tokenizers.py
+++ b/src/transformers/tokenization_utils_tokenizers.py
@@ -1277,14 +1277,20 @@ class TokenizersBackend(PreTrainedTokenizerBase):
                 >> Tags including `base_model:.*mistralai`
         """
         import re
+        from functools import lru_cache
 
         from huggingface_hub import model_info
         from packaging import version
 
         from transformers.utils.hub import cached_file
 
+        @lru_cache(maxsize=128)
         def is_base_mistral(model_id: str) -> bool:
-            model = model_info(model_id)
+            try:
+                model = model_info(model_id)
+            except Exception:
+                # Never block tokenizer init on a Hub error — assume non-Mistral.
+                return False
             if model.tags is not None:
                 if re.search("base_model:.*mistralai", "".join(model.tags)):
                     return True
--- a/src/transformers/tokenization_utils_base.py
+++ b/src/transformers/tokenization_utils_base.py
@@ -1828,6 +1828,7 @@ class PreTrainedTokenizerBase(PushToHubMixin):
 
         init_kwargs["name_or_path"] = pretrained_model_name_or_path
         init_kwargs["is_local"] = _is_local
+        init_kwargs["local_files_only"] = local_files_only
 
         #### Handle tokenizer serialization of added and special tokens
         added_tokens_decoder: dict[int, AddedToken] = {}
--- a/tests/tokenization/test_tokenization_fast.py
+++ b/tests/tokenization/test_tokenization_fast.py
@@ -287,3 +287,47 @@ class ReduceMutableBorrowTests(unittest.TestCase):
 
     def fetch(self, tokenizer, text):
         return tokenizer.encode(text, truncation="longest_first", padding="longest")
+
+
+@require_tokenizers
+class PatchMistralRegexHubCallTests(unittest.TestCase):
+    """
+    Regression tests for https://github.com/huggingface/transformers/issues/44749 /
+    https://github.com/huggingface/transformers/issues/43502
+
+    `_patch_mistral_regex` used to unconditionally call `huggingface_hub.model_info`
+    for any tokenizer with vocab > 100k (e.g. Qwen3), causing a large slowdown
+    and breaking offline / local-path loading.
+    """
+
+    def _dummy_backend_tokenizer(self):
+        tok = Tokenizer(WordLevel({"[UNK]": 0, "a": 1}, unk_token="[UNK]"))
+        tok.pre_tokenizer = pre_tokenizers.Whitespace()
+        return tok
+
+    def test_local_files_only_skips_model_info(self):
+        from unittest.mock import patch
+
+        tok = self._dummy_backend_tokenizer()
+        with patch(
+            "huggingface_hub.model_info",
+            side_effect=AssertionError("model_info must not be called when local_files_only=True"),
+        ):
+            result = PreTrainedTokenizerFast._patch_mistral_regex(
+                tok,
+                "some/non-mistral-repo-id",
+                local_files_only=True,
+            )
+        self.assertIsNotNone(result)
+
+    def test_hub_error_does_not_break_init(self):
+        from unittest.mock import patch
+
+        tok = self._dummy_backend_tokenizer()
+        with patch("huggingface_hub.model_info", side_effect=RuntimeError("hub down")):
+            # Must not raise — a Hub failure should be swallowed for non-Mistral models.
+            result = PreTrainedTokenizerFast._patch_mistral_regex(
+                tok,
+                "not-a-real/hub-id",
+            )
+        self.assertIsNotNone(result)


src/transformers/tokenization_utils_tokenizers.py
```diff
@@ -1196,16 +1196,22 @@ def _patch_mistral_regex(
...
+
         def is_base_mistral(model_id: str) -> bool:
             model = model_info(model_id)
             if model.tags is not None:
                 if re.search("base_model:.*mistralai", "".join(model.tags)):
                     return True
             return False
 
-        if is_offline_mode():
-            is_local = True
```
vasqu, line 1294: Wouldn't it make more sense to just adjust these lines here to something along `local_files_only or is_offline_mode()` - to me it looks like the core issue is that `is_local` can be false even if we have `local_files_only=True`

Also let's add a small test

vaibhav-research: @vasqu Thanks, agreed on the direction.

I think we’re largely aligned. my understanding from reproducing this is that the core issue isn’t just how `is_local` is set, but that we can hit a Hub metadata call before the offline / local-only intent is fully respected.

In particular, this helper:

```
def is_base_mistral(model_id: str) -> bool:
            model = model_info(model_id)
            if model.tags is not None:
                if re.search("base_model:.*mistralai", "".join(model.tags)):
                    return True
            return False
```

unconditionally calls `model_info(model_id)`, which triggers a `/api/models/<repo>` request. In my repro, that happens even when `local_files_only=True`, because the call occurs before we can short-circuit based on offline intent.

I initially tried forcing `is_local = True` when `local_files_only or is_offline_mode()`, but since the `model_info()` call is reached regardless, it didn’t fully prevent the network access in practice. That’s why I opted for an early return before we ever reach `is_base_mistral()` for `non-Mistral / offline` cases.

also, will add a test for this.

vasqu: Sorry, I pushed to your branch directly 990bc92e9f62180ee30a8e6a49e794abfaead8e0 - I think this is easier that way than to collect the lines on git :sweat_smile: 

The problem was that `local_files_only` was never passed making it always default to false. The issue is that the API call only happens when `is_local=False` and hence `local_files_only` had no effect, it made the call regardless

src/transformers/tokenization_utils_tokenizers.py
```diff
@@ -377,7 +377,6 @@ def __init__(self, *args, **kwargs):
                 self._tokenizer,
                 self.init_kwargs.get("name_or_path", None),
                 init_kwargs=self.init_kwargs,
-                fix_mistral_regex=kwargs.get("fix_mistral_regex"),
                 **kwargs,
```
vasqu, line 482: That didn't really make sense, we pass kwargs either way :eyes:

src/transformers/tokenization_utils_tokenizers.py
vaibhav-research: Makes sense, its way more cleaner approach. thanks a lot for pushing that change 😃    @vasqu 

One thing, even after this, in online mode `local_files_only=False` we still call `model_info()` for any model that hits the patch path, e.g. `Qwen2Tokenizer -> _patch_mistral_regex -> is_base_mistral() -> model_info()`.

My repro PHASE 2 blocks `/api/models/<repo>` for non-Mistral and it still triggers for `Qwen`, because `is_base_mistral` currently always calls `model_info()`.

If we want to avoid unnecessary Hub metadata calls for non-Mistral models, we likely need a cheap guard (e.g. only run is_base_mistral if repo_id looks `Mistral-ish: mistralai/*` or `contains “mistral”`), either as an early return or inside `is_base_mistral()`.

I am going to paste the test I am running to prove my point. Thanks again for your time.

vasqu: I get your point but the problem is with custom repos and custom models, we do not have much freedom here and have to check just in case. We cannot assume that only mistral ai will use mistral tokenizers. 

And in this case, it's fairly inexpensive call for making sure we catch as many edge cases as possible

vaibhav-research:
the test I am running is following

Step 1: preparing cache


```
from huggingface_hub import snapshot_download

MODELS = [
    "Qwen/Qwen3-30B-A3B",
    "Qwen/Qwen2.5-7B-Instruct",
    "Qwen/Qwen2-7B-Instruct",
    "mistralai/Mistral-7B-v0.1",
    "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
    "microsoft/phi-3-mini-4k-instruct",
    "gpt2",
]

ALLOW = [
    "config.json",
    "tokenizer.json",
    "tokenizer.model",
    "tokenizer_config.json",
    "special_tokens_map.json",
    "added_tokens.json",
    "merges.txt",
    "vocab.json",
    "*.tiktoken",
]

for model_id in MODELS:
    snapshot_download(repo_id=model_id, allow_patterns=ALLOW)
    print(f"cached minimal: {model_id}")
```

```
python prepare_cache.py 
Fetching 5 files: 100%|█████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 10160.62it/s]
Download complete: : 0.00B [00:00, ?B/s]              cached minimal: Qwen/Qwen3-30B-A3B           | 0/5 [00:00<?, ?it/s]
Fetching 5 files: 100%|█████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 12953.38it/s]
Downloading (incomplete total...): 0.00B [00:00, ?B/s]cached minimal: Qwen/Qwen2.5-7B-Instruct
Download complete: : 0.00B [00:00, ?B/s]                                                           | 0/5 [00:00<?, ?it/s]
Download complete: : 0.00B [00:00, ?B/s]
Fetching 5 files: 100%|█████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 46707.17it/s]
                                                      cached minimal: Qwen/Qwen2-7B-Instruct
Fetching 5 files: 100%|█████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 12694.62it/s]
Download complete: : 0.00B [00:00, ?B/s]              cached minimal: mistralai/Mistral-7B-v0.1    | 0/5 [00:00<?, ?it/s]
Fetching 5 files: 100%|█████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 60963.72it/s]
Downloading (incomplete total...): 0.00B [00:00, ?B/s]cached minimal: TinyLlama/TinyLlama-1.1B-Chat-v1.0
Fetching 6 files: 100%|██████████████████████████████████████████████████████████████████| 6/6 [00:00<00:00, 9597.95it/s]
Fetching 5 files:   0%|                               cached minimal: microsoft/phi-3-mini-4k-instruct/5 [00:00<?, ?it/s]
Download complete: : 0.00B [00:00, ?B/s]
Download complete: : 0.00B [00:00, ?B/s] [00:00, ?B/s]
Download complete: : 0.00B [00:00, ?B/s]                                                           | 0/6 [00:00<?, ?it/s]
Fetching 5 files: 100%|█████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 15196.75it/s]
Downloading (incomplete total...): 0.00B [00:00, ?B/s]cached minimal: gpt2
Download complete: : 0.00B [00:00, ?B/s]
Download complete: : 0.00B [00:00, ?B/s]
```


Step2:  run online and offline test using following script


```
import httpx
import traceback

MODELS = [
    "Qwen/Qwen3-30B-A3B",
    "Qwen/Qwen2.5-7B-Instruct",
    "Qwen/Qwen2-7B-Instruct",
    "mistralai/Mistral-7B-v0.1",
    "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
    "microsoft/phi-3-mini-4k-instruct",
    "gpt2",
]

_real_send = httpx.Client.send


def _is_hf_model_info_root(req: httpx.Request) -> bool:

    try:
        host = (req.url.host or "").lower()
        path = req.url.path or ""
    except Exception:
        return False

    if "huggingface.co" not in host:
        return False
    if not path.startswith("/api/models/"):
        return False
    return "/tree/" not in path


def _blocked_send_all(self, request, *args, **kwargs):
    raise RuntimeError(f"blocked HTTP call: {request.method} {request.url}")


def _blocked_send_model_info_for_non_mistral(self, request, *args, **kwargs):
    if _is_hf_model_info_root(request):
        repo_id = request.url.path[len("/api/models/") :].split("?", 1)[0]
        repo_id_l = repo_id.lower()
        mistralish = repo_id_l.startswith("mistralai/") or ("mistral" in repo_id_l)
        if not mistralish:
            raise RuntimeError(f"blocked model_info for non-mistral: {request.method} {request.url}")
    return _real_send(self, request, *args, **kwargs)


def run_phase(name: str, *, local_files_only: bool, block_mode: str, show_stack: bool = False):
    if block_mode == "all":
        httpx.Client.send = _blocked_send_all
    elif block_mode == "model_info_non_mistral":
        httpx.Client.send = _blocked_send_model_info_for_non_mistral
    elif block_mode == "none":
        httpx.Client.send = _real_send
    else:
        raise ValueError(f"unknown block_mode={block_mode}")

    failures = []

    for model_id in MODELS:
        try:
            AutoTokenizer.from_pretrained(
                model_id,
                local_files_only=local_files_only,
                trust_remote_code=False,
            )
        except Exception as e:
            failures.append((model_id, e))
            if show_stack:
                print("\n--- stack (trimmed) ---")
                traceback.print_exc(limit=12)

    print("\n" + "=" * 88)
    print(name)
    print("=" * 88)

    if not failures:
        print("result: OK (no failures)")
        return True

    print(f"result: FAIL ({len(failures)} failures)")
    for model_id, e in failures:
        print(f"- {model_id}: {repr(e)}")

    return False


def main():
    ok1 = run_phase(
        "phase 1: local_files_only=True, block all HTTP (should be fully offline)",
        local_files_only=True,
        block_mode="all",
        show_stack=False,
    )

    ok2 = run_phase(
        "phase 2: local_files_only=False, block model_info(/api/models/<repo>) for non-mistral",
        local_files_only=False,
        block_mode="model_info_non_mistral",
        show_stack=False,
    )

    httpx.Client.send = _real_send

    print("\nsummary:")
    print(f"- phase 1: {'ok' if ok1 else 'fail'}")
    print(f"- phase 2: {'ok' if ok2 else 'fail'}")


if __name__ == "__main__":
    main()
    
```

test results:

```
python repro.py                                

========================================================================================
phase 1: local_files_only=True, block all HTTP (should be fully offline)
========================================================================================
result: OK (no failures)

========================================================================================
phase 2: local_files_only=False, block model_info(/api/models/<repo>) for non-mistral
========================================================================================
result: FAIL (3 failures)
- Qwen/Qwen3-30B-A3B: RuntimeError('blocked model_info for non-mistral: GET https://huggingface.co/api/models/Qwen/Qwen3-30B-A3B')
- Qwen/Qwen2.5-7B-Instruct: RuntimeError('blocked model_info for non-mistral: GET https://huggingface.co/api/models/Qwen/Qwen2.5-7B-Instruct')
- Qwen/Qwen2-7B-Instruct: RuntimeError('blocked model_info for non-mistral: GET https://huggingface.co/api/models/Qwen/Qwen2-7B-Instruct')

summary:
- phase 1: ok
- phase 2: fail
```
@vasqu I pulled the changes you pushed and re ran the test that I ran when the issue was reported in #43502. please let me know if my test is good or if I am missing anything. this is only happening with `qwen` to be specific.

vasqu:
Re tests, I think it makes sense to extend (the following and other similar tests for other tokenizer types) https://github.com/huggingface/transformers/blob/071e178be163917777dac272e8e26525bc20db08/tests/test_tokenization_common.py#L2704 :eyes:

src/transformers/tokenization_utils_tokenizers.py
vaibhav-research: makes sense, I will leave it as is then and work on adding a test for this.

vaibhav-research:
> Re tests, I think it makes sense to extend (the following and other similar tests for other tokenizer types)
> 
> https://github.com/huggingface/transformers/blob/071e178be163917777dac272e8e26525bc20db08/tests/test_tokenization_common.py#L2704
> 
> 👀

sure, will extend this.

src/transformers/tokenization_utils_tokenizers.py
```diff
@@ -1203,7 +1202,7 @@ def is_base_mistral(model_id: str) -> bool:
                     return True
             return False
```
ArthurZucker, line 1297:
```suggestion
        from functools import lru_cache

        @lru_cache(maxsize=128)
        def is_base_mistral(model_id: str) -> bool:
            try:
                model = model_info(model_id)
            except Exception:
                # Never block tokenizer init on a Hub error — assume non-Mistral.
                return False
            if model.tags is not None:
                if re.search("base_model:.*mistralai", "".join(model.tags)):
                    return True
            return False
```

ArthurZucker (commented):
Nice, small patch — thanks! A few follow-ups worth tacking on while we're here:

1. Cache `is_base_mistral` with `@lru_cache` so repeated loads of the same Hub id (notebooks, rollout loops, DDP workers) don't each hit `/api/models/...`.
2. Wrap `model_info()` in `try/except` and return `False` on any error — a Hub hiccup / 5xx / ratelimit shouldn't break tokenizer init for non-Mistral models.
3. Worth pairing this with #43212 (offline-load regression test) or adding a minimal test here that monkeypatches `huggingface_hub.model_info` to assert it isn't called for non-Mistral local paths.

Inline suggestions below.

PR #43609, opened 2026-01-29
[`Rope`] Re-Enable remote code

vasqu:
As per title, draft to show what would be roughly needed


src/transformers/configuration_utils.py
```diff
@@ -215,8 +216,8 @@ def __init__(
         # BC for rotary embeddings. We will pop out legacy keys from kwargs and rename to new format
         if hasattr(self, "rope_parameters"):
             ignore_keys_at_rope_validation = kwargs.pop("ignore_keys_at_rope_validation", None)
-            kwargs = self.convert_rope_params_to_dict(
-                ignore_keys_at_rope_validation=ignore_keys_at_rope_validation, **kwargs
+            kwargs = RotaryEmbeddingConfigMixin.convert_rope_params_to_dict(
```
hmellor: Some configs in Transformers do define this method, so we can't always call the method from the mixin

PR #43614, opened 2026-01-30
add contiguous operation for diffllama model for xpu to enable compile mode.

kaixuanliu:
We need to add contiguous operation in sdpa stage for xpu as well for this model so as we can use compile mode. pls help review, thx! @IlyasMoutawwakil


kaixuanliu:
Related test case:
`tests/models/diffllama/test_modeling_diffllama.py::DiffLlamaModelTest::test_sdpa_can_compile_dynamic`

IlyasMoutawwakil (approved):
LGTM

PR #43592, opened 2026-01-29
Fixes configuration default values

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/43525
Fixes https://github.com/huggingface/transformers/issues/43572

Adds missing `pad_token_id` and `tie_word_embeddings` to config classes with their defaults

PR #43617, opened 2026-01-30
Fix tie_word_embedding issue for llava_onevision model

kaixuanliu:
@zucchini-nlp pls help review, thx! We have to add back the changes in https://github.com/huggingface/transformers/pull/42523. As for llava_onevision model, in its checkpoint config file, the model's `tie_word_embeddings` is Flase, and model's text_config's `tie_word_embeddings` is True: [L171](https://huggingface.co/llava-hf/llava-onevision-qwen2-0.5b-ov-hf/blob/main/config.json#L171),  which causes when loading the model, the `lm_head.weight` of the model will get missing, hence cause totally wrong output. You can run this unit test to reproduce:
`RUN_SLOW=1 pytest -rA tests/models/llava_onevision/test_modeling_llava_onevision.py::LlavaOnevisionForConditionalGenerationIntegrationTest::test_small_model_integration_test_multi_image_nested`

### DIFF
--- a/src/transformers/models/llava_next_video/configuration_llava_next_video.py
+++ b/src/transformers/models/llava_next_video/configuration_llava_next_video.py
@@ -162,5 +162,11 @@ class LlavaNextVideoConfig(PreTrainedConfig):
 
         super().__init__(**kwargs)
 
+        # Due to a mismatch at model addition-time, the `tie_word_embeddings` was saved in the text config, even
+        # though it concerns the main model, while it was set to False by default in the main model... So we hardcode a fix here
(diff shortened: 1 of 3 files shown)


zucchini-nlp (commented):
Hey @kaixuanliu , thanks a lot for catching this. I am fixing all config default value issues in https://github.com/huggingface/transformers/pull/43592 instead of fixing per model in separate PRs

I'll add llava models in the PR to keep it in one place

kaixuanliu:
Great! Will close this PR once https://github.com/huggingface/transformers/pull/43592 is merged.

zucchini-nlp:
PR merged and will be in the next patch release, prob on Monday

PR #43619, opened 2026-01-30
Don't modify `tied_weight_keys` in-place

zucchini-nlp:
# What does this PR do?

As per title, some models add or delete entries in tied weights depending on configuration. If we load two models consecutively with different configs, it fails to tie weights correctly

I am copying it in `__init__` same way as `keep_in_fp32`. We could also simply `copy` the cls attribute in these two models, where in-place modification happens. WDYT?

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -1321,6 +1321,7 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         # when a different component (e.g. language_model) is used.
         self._keep_in_fp32_modules = copy.copy(self.__class__._keep_in_fp32_modules)
         self._keep_in_fp32_modules_strict = copy.copy(self.__class__._keep_in_fp32_modules_strict)
+        self._tied_weight_keys = copy.copy(self.__class__._tied_weight_keys)
         self.dtype_plan = {}
 
         if isinstance(self._keep_in_fp32_modules, list):
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -1318,7 +1318,8 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         self.warnings_issued = {}
         # Overwrite the class attribute to make it an instance attribute, so models like
         # `InstructBlipForConditionalGeneration` can dynamically update it without modifying the class attribute
-        # when a different component (e.g. language_model) is used.
+        # when a different component (e.g. language_model) is used. Same for `_tied_weights_keys` which pops/adds
+        # new keys dynamically depending on config values
         self._keep_in_fp32_modules = copy.copy(self.__class__._keep_in_fp32_modules)
         self._keep_in_fp32_modules_strict = copy.copy(self.__class__._keep_in_fp32_modules_strict)
         self._tied_weights_keys = copy.copy(self.__class__._tied_weights_keys)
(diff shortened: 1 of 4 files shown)


tests/models/deformable_detr/test_modeling_deformable_detr.py
```diff
@@ -243,6 +243,25 @@ def test_deformable_detr_object_detection_head_model(self):
...
+        self.assertTrue("model.decoder.bbox_embed" in model._tied_weights_keys)
+        self.assertTrue("model.decoder.class_embed" in model._tied_weights_keys)
+
+        # if we update config attr, model's tied weights keys also change
+        config.with_box_refine = False
+        config.two_stage = False
+
+        model = DeformableDetrForObjectDetection(config)
+        self.assertFalse("model.decoder.bbox_embed" in model._tied_weights_keys)
+        self.assertFalse("model.decoder.class_embed" in model._tied_weights_keys)
```
vasqu, line 263: Aren't we checking the exact same keys, not the modified ones? E.g. once we should check for `bbox_embed.0` and `bbox_embed` instead, no?

Same for the dino test

src/transformers/modeling_utils.py
```diff
@@ -1321,6 +1321,7 @@ def __init__(self, config: PreTrainedConfig, *inputs, **kwargs):
         # when a different component (e.g. language_model) is used.
         self._keep_in_fp32_modules = copy.copy(self.__class__._keep_in_fp32_modules)
         self._keep_in_fp32_modules_strict = copy.copy(self.__class__._keep_in_fp32_modules_strict)
+        self._tied_weights_keys = copy.copy(self.__class__._tied_weights_keys)
```
vasqu, line 1325: I think this is fine, it also allows for a bit more freedom on custom models ig

vasqu (commented):
Just a bit confused about the test since we check for the same keys? Does it work as intended

tests/models/deformable_detr/test_modeling_deformable_detr.py
zucchini-nlp: no, we are checking the modified ones. These are manually added/popped from the dict, so we verify that they are present when weights should be tied. Otherwise it is absent

This test fails on main and passes under the PR

vasqu: My bad, I didn't see it was `assertFalse` in the second case - makes total sense, it's adding a whole new key pair

vasqu (approved):
Thanks for clarifiying, my bad didn't properly see the second case

src/transformers/modeling_utils.py
vasqu, line 1325: Nit: Can we maybe add a small comment here as well and move it before the keep in xxx? So
```
# Comment
self._tied_weights_keys = copy.copy(self.__class__._tied_weights_keys)
# Existing comment
self._keep_in_fp32_modules = copy.copy(self.__class__._keep_in_fp32_modules)
self._keep_in_fp32_modules_strict = copy.copy(self.__class__._keep_in_fp32_modules_strict)
....
```
makes it more ordered to what's supposed to be together

zucchini-nlp: I think `Overwrite the class attribute to make it an instance attribute, blah blah blah` is descriptive enough as it applied to class attributes in general

I added one small sentence at the end of that comment to make it explicit :)

tests/models/mm_grounding_dino/test_modeling_mm_grounding_dino.py
```diff
@@ -327,6 +327,11 @@ def test_feed_forward_chunking(self):
     def test_load_save_without_tied_weights(self):
         pass
 
+    # Ignore copy
+    def test_tie_weights_is_not_modified(self):
+        # this model doesn't need a test
+        pass
```
molbap, line 333: Just passing through, why doesn't this need a test? (wrt comment)

zucchini-nlp: the model definition is slightly different from `GroundingDino` but the tests are `Copied from`. I don't see `tied_weights_keys` being modified for `MMGroundingDino`

zucchini-nlp:
CI is still super flaky :(

vasqu:
As discussed CI is still flaky, merging as the test that fails is unrelated - can you write something internally in slack

PR #43620, opened 2026-01-30
[`Rope`] Revert #43410 and make inheritance implicit again

vasqu:
As per title, this is mainly due to the fact that we have little control over remote code which could lead to solutions like
- Closes #43609 --> makes the whole mixin behave like a static holder for methods... 
- Modify methods/inherited classes on the fly --> risky and unnatural behavior might be introduced

Imo, this is the best for now without incurring too many risks


zucchini-nlp (approved):
Nice!

ArthurZucker (approved):
😢 I'm sorry I was the one who suggested this change! 
I agree that forcing remote code config to inherit is brittle and it does not really hurt so yeah let's keep that!

vasqu:
No worries! Glad it's at least after the big release so no patch required, merging since the failing test is unrelated

PR #43621, opened 2026-01-30
[vllm compat] Separate renaming from conversion ops

Cyrilvallez:
# What does this PR do?

As per the title. Following specs described by @hmellor for vllm compatibility, no logic change should be required. Only the following small changes to mappings should be enough


ArthurZucker (approved):
I don't mind, but its not always super convenient when you have a lot of renaming

hmellor:
> but its not always super convenient when you have a lot of renaming

Some day I plan to support `WeightConverter` as well as `WeightRenaming` in vLLM so this won't be a problem. But for now I only have a WIP for `WeightRenaming`

PR #43622, opened 2026-01-30
Allow Attention and Experts to be used as standalone modules

Cyrilvallez:
# What does this PR do?

As per the title. Also fix a vllm regresion on experts. Allow Experts and Attention module to be used on their own, such as

```py
import torch
from transformers.models.mixtral.modeling_mixtral import MixtralExperts
config = AutoConfig.from_pretrained("mistralai/mixtral-8x7b-v0.1")
experts = MixtralExperts(config)
experts.forward(torch.randn((1, 64, experts.config.hidden_size)))
```

```py
 File "/home/harry/transformers/src/transformers/integrations/moe.py", line 317, in forward
    experts_forward = ALL_EXPERTS_FUNCTIONS[self.config._experts_implementation]
                      ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/harry/transformers/src/transformers/utils/generic.py", line 1070, in __getitem__
    return self._global_mapping[key]
           ~~~~~~~~~~~~~~~~~~~~^^^^^
KeyError: None
```


vasqu (approved):
Could we add a small test so that we don't have a regression?

src/transformers/integrations/moe.py
```diff
@@ -265,6 +269,20 @@ class ExpertsInterface(GeneralInterface):
         "grouped_mm": grouped_mm_experts_forward,
     }
 
+    def get_interface(self, experts_implementation: str, default: Callable) -> Callable:
+        """Return the requested `experts_implementation`. Also strictly check its validity, and raise if invalid."""
+        if experts_implementation is None:
+            logger.warning_once(
+                "You tried to access the `ExpertsInterface` with a `config.experts_implementation` set to `None`. This "
+                "is expected if you use an Expert Module as a standalone Module. If this is not the case, something went "
+                "wrong with the dispatch of `config.experts_implementation`"
```
vasqu:
```suggestion
                "wrong with the dispatch of `config._experts_implementation`"
```
technically, no?

tests/utils/test_modeling_utils.py
```diff
@@ -3084,6 +3084,43 @@ def test_not_available_kernels(self):
...
+
+        # Try the Experts (check it works + raises the warning)
+        hidden_states = hidden_states.reshape(-1, hidden_size)
+        dummy_scores = torch.randn(seq_len, config.num_experts_per_tok)
+        dummy_indices = torch.randint(0, config.num_local_experts, (seq_len, config.num_experts_per_tok))
+        with CaptureLogger(logging.get_logger("transformers.integrations.moe")) as cl:
+            _ = experts_module(hidden_states, dummy_indices, dummy_scores)
+        self.assertIn(
+            "You tried to access the `ExpertsInterface` with a `config.experts_implementation` set to `None`.", cl.out
+        )
```
vasqu, line 3126: Nit: let's also add the case where we have something invalid (unregistered implementation)

vasqu (approved):
LGTM, just 2 nits

tests/utils/test_modeling_utils.py
Cyrilvallez: Indeed good idea

PR #43623, opened 2026-01-30
Tie zamba weights correctly

zucchini-nlp:
# What does this PR do?

as per title

### DIFF
--- a/src/transformers/models/zamba/modeling_zamba.py
+++ b/src/transformers/models/zamba/modeling_zamba.py
@@ -844,10 +844,11 @@ class ZambaModel(ZambaPreTrainedModel):
                 layers.append(ZambaHybridLayer(ZambaAttentionDecoderLayer(config), linear, mamba))
                 if self._tied_weights_keys is None:
                     self._tied_weights_keys = {
-                        rf"layers.(?![{layer_id}])\d+.shared_transf": f"layers.{layer_id}.shared_transf"
+                        rf"layers.(?!{layer_id}\.)\d+.shared_transf": f"layers.{layer_id}.shared_transf"
                     }
             else:
                 layers.append(mamba)
+        print()
         self.layers = nn.ModuleList(layers)
 
         self._attn_implementation = config._attn_implementation
--- a/src/transformers/models/zamba/modeling_zamba.py
+++ b/src/transformers/models/zamba/modeling_zamba.py
@@ -848,7 +848,6 @@ class ZambaModel(ZambaPreTrainedModel):
                     }
             else:
                 layers.append(mamba)
-        print()
         self.layers = nn.ModuleList(layers)
 
         self._attn_implementation = config._attn_implementation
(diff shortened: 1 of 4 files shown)


zucchini-nlp:
run-slow: zamba, zamba2

src/transformers/models/zamba/modeling_zamba.py
```diff
@@ -844,10 +844,11 @@ def __init__(self, config: ZambaConfig):
                 layers.append(ZambaHybridLayer(ZambaAttentionDecoderLayer(config), linear, mamba))
                 if self._tied_weights_keys is None:
                     self._tied_weights_keys = {
-                        rf"layers.(?![{layer_id}])\d+.shared_transf": f"layers.{layer_id}.shared_transf"
+                        rf"layers.(?!{layer_id}\.)\d+.shared_transf": f"layers.{layer_id}.shared_transf"
```
zucchini-nlp, line 847: it was not matching `layer.25.shared_transform` when we need to tie layer 25 to layer 2. Adding a "dot" amkes sure that the pattern matches all layers except for "2"

vasqu: Yea honestly how was it supposed to work without the dot :sweat_smile:

src/transformers/models/zamba2/modular_zamba2.py
```diff
@@ -967,12 +967,26 @@ def get_layers(self, blocks, linear_layers, mamba_layers):
...
             if layer_type == "hybrid":
                 block = next(blocks)
-                if self.config.num_mem_blocks * len(self.config.hybrid_layer_ids) > 1:
-                    prefix_pattern = f"layers.{layer_id}.shared_transformer"
-                    self._tied_weights_keys.update({prefix_pattern: "layers.0.shared_transformer"})
+                prefix_pattern = f"layers.{layer_id}.shared_transformer"
+
+                # Zamba ties Hybrid module weights by repeating blocks after every
+                # `num_mem_blocks`. So if `num_mem_blocks=2`, the blocks looks like
+                # [1, 2, 1, 2, 1, 2] where all "ones" share the same set of weights.
```
vasqu, line 967: Wouldn't it be all same digits share the same set of weights?

vasqu (approved):
Sorry for the delays, had to debug myself to check what was going on :D 

Do we have a test that catches this (except for vllm catching it for us) - some small dummy config ought to replicate it

src/transformers/models/zamba2/modular_zamba2.py
zucchini-nlp: i think with official weights, it's yes because `num_mem_blocks = 1`. Just wanted to make it technically correct wrt to v4

zucchini-nlp:
Yeah, I'd expect the slow tests to trigger tbh but prob they've been failing in `main` for long time. Will think of a fast test to add

PR #43400, opened 2026-01-22
Add VibeVoice Acoustic Tokenizer

ebezzam:
# What does this PR do?

Splitting off acoustic tokenizer from https://github.com/huggingface/transformers/pull/40546
Such that [VibeVoice ASR](https://huggingface.co/microsoft/VibeVoice-ASR) can be done in a separate / independent PR

Model card: https://huggingface.co/bezzam/VibeVoice-AcousticTokenizer

cc @eustlb

PR #43625, opened 2026-01-30
Add VibeVoice ASR

ebezzam:
# What does this PR do?

Adds VibeVoice ASR, branches off of (and hence depends on) https://github.com/huggingface/transformers/pull/43400

Original checkpoint: https://huggingface.co/microsoft/VibeVoice-ASR
Original modeling code: https://github.com/microsoft/VibeVoice/blob/main/vibevoice/modular/modeling_vibevoice_asr.py

Transformers compatible checkpoint: https://huggingface.co/bezzam/VibeVoice-ASR-7B
Example usage (after installing with `pip install git+https://github.com/ebezzam/transformers.git@vibevoice_asr`): https://github.com/ebezzam/transformers/blob/vibevoice_asr/docs/source/en/model_doc/vibevoice_asr.md#usage

---

TODO
 - [x] checkpoint conversion: https://huggingface.co/bezzam/VibeVoice-ASR-7B
 - [x] passing reproducer: https://gist.github.com/ebezzam/e1200bcecdc29e87dadd9d8423ae7ecb#file-reproducer_vibevoice_asr-py
 - [x] cleanup
 - [x] finish tests
 - [x] propagate some changes back to https://github.com/huggingface/transformers/pull/43400
 - [x] convenience function for transcribing (like Audio Flamingo)

cc: @eustlb


src/transformers/processing_utils.py
ebezzam, line 1: Some fixes here are needed for chat template to use audio that aren't 16 kHz

related to https://github.com/huggingface/transformers/pull/43273

src/transformers/models/glmasr/modular_glmasr.py
```diff
@@ -116,11 +116,11 @@ def apply_transcription_request(
                 Custom prompt(s) to include in the user turn. A list must be the same length as the batch. When `None`,
                 each sample uses `"Transcribe the input speech."`.
             **kwargs:
-                Additional keyword arguments forwarded to [`~AudioFlamingo3Processor.apply_chat_template`] (for example
```
eustlb: nice catch! thanks

docs/source/en/model_doc/vibevoice_asr.md
```diff
@@ -0,0 +1,360 @@
...
+
+
+model_id = "bezzam/VibeVoice-ASR-7B"
+
+# Load processor and model
+processor = AutoProcessor.from_pretrained(model_id)
+model = VibeVoiceAsrForConditionalGeneration.from_pretrained(model_id, device_map="auto")
+print(f"Model loaded on {model.device} with dtype {model.dtype}")
+
+# Prepare inputs using `apply_transcription_request`
```
ebezzam, line 64: chat template usage comes later on. I think it's more intuitive for users to see this API first?

docs/source/en/model_doc/vibevoice_asr.md
```diff
@@ -0,0 +1,360 @@
...
+# Apply model with batch inputs
+inputs = processor.apply_transcription_request(audio, prompt=prompts).to(model.device, model.dtype)
+output_ids = model.generate(**inputs)
+generated_ids = output_ids[:, inputs["input_ids"].shape[1] :]
+transcription = processor.batch_decode(generated_ids, extract_transcription=True)
+
+print(transcription)
+```
+
+### Adjusting tokenizer chunk (e.g. if out-of-memory)
```
ebezzam, line 182: A community member reached out to me about VibeVoice ASR, and mentioned that they had OOM issues (with the original checkpoint/codebase) when using the default chunk size of 60s (see [here](https://github.com/microsoft/VibeVoice/blob/0aa8cb4c64fc87f281f89ea5e733b7005eff36a8/vibevoice/modular/modeling_vibevoice_asr.py#L213)).

So I think it's good to expose this parameter, so people can tweak the chunk size according to their hardware.

Curious on your thoughts of the best way to expose/handle. See processor.

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,493 @@
...
+        audio_token_id (`int`, *optional*, defaults to 151648):
+            The audio token index to encode the audio prompt.
+        audio_bos_token_id (`int`, *optional*, defaults to 151646):
+            The audio begin-of-sequence token index.
+        audio_eos_token_id (`int`, *optional*, defaults to 151647):
+            The audio end-of-sequence token index.
+        acoustic_vae_std (`float`, *optional*, defaults to 0.625):
+            Standard deviation used during acoustic VAE sampling.
+        tokenizer_chunk_size (`int`, *optional*, defaults to 1440000):
+            The chunk size (in number of samples) to use when tokenizer audio inputs. Default corresponds to 60 seconds at 24kHz.
```
ebezzam: I'm torn between keeping the parameter `tokenizer_chunk_size` in the model config or in its generation config, as there may be cases where the user wants to set it when calling `generate` to avoid OOM. What do you think?

From my understanding of chats with @vasqu and @zucchini-nlp it's better to avoid custom parameters in the generation config? As they weren't intended to be covered with https://github.com/huggingface/transformers/pull/42702 but they were covered in the end. I'm for whatever is better practice within Transformers!

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,493 @@
...
+    def forward(self, acoustic_latents, semantic_latents):
+        acoustic_features = self.acoustic_linear_1(acoustic_latents)
+        acoustic_features = self.acoustic_norm(acoustic_features)
+        acoustic_features = self.acoustic_linear_2(acoustic_features)
+
+        semantic_features = self.semantic_linear_1(semantic_latents)
+        semantic_features = self.semantic_norm(semantic_features)
+        semantic_features = self.semantic_linear_2(semantic_features)
+
+        return acoustic_features + semantic_features
```
ebezzam, line 175: New type of multi-modal projector with acoustic and semantic paths

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,493 @@
...
+@auto_docstring(
+    custom_intro="""
+    The VibeVoice ASR model with pre-trained acoustic tokenizers and a language model.
+    """
+)
+class VibeVoiceAsrForConditionalGeneration(AudioFlamingo3ForConditionalGeneration):
+    def __init__(self, config: VibeVoiceAsrConfig):
+        super().__init__(config)
+        self.acoustic_tokenizer = AutoModel.from_config(config.acoustic_tokenizer_config)
+        self.semantic_tokenizer = AutoModel.from_config(config.semantic_tokenizer_config)
```
ebezzam: FYI both are `VibeVoiceAsrEncoderModel`

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,493 @@
...
+
+        if tokenizer_chunk_size is None:
+            tokenizer_chunk_size = self.config.tokenizer_chunk_size
+        else:
+            hop_length = np.prod(self.acoustic_tokenizer.config.downsampling_ratios)
+            if tokenizer_chunk_size % hop_length != 0:
+                tokenizer_chunk_size = int((tokenizer_chunk_size // hop_length) * hop_length)
+                logger.warning(
+                    f"`tokenizer_chunk_size` has been adjusted to {tokenizer_chunk_size} to be a multiple of hop length ({hop_length})."
+                )
```
ebezzam: What do you think about this handling to ensure that the user-specified `tokenizer_chunk_size` is a multiple of the hop-length?

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,493 @@
...
+                semantic_encoder_cache = semantic_encoder_output.padding_cache
+
+            acoustic_latents = torch.cat(acoustic_latents, dim=1)
+            semantic_latents = torch.cat(semantic_latents, dim=1)
+
+            # Sample acoustic tokens
+            noise_std = self.config.acoustic_vae_std * torch.randn(
+                acoustic_latents.shape[0], device=acoustic_latents.device, dtype=acoustic_latents.dtype
+            )
+            acoustic_latents = acoustic_latents + noise_std[:, None, None] * torch.randn_like(acoustic_latents)
```
ebezzam, line 262: Pulled sampling out of acoustic tokenizer so that:
1. Same `VibeVoiceAsrEncoderModel` can be used for both acoustic and semantic paths
2. Replicate [their approach](https://github.com/microsoft/VibeVoice/blob/0aa8cb4c64fc87f281f89ea5e733b7005eff36a8/vibevoice/modular/modeling_vibevoice_asr.py#L324-L326) of sampling on the concatenated acoustic latents

src/transformers/models/vibevoice_asr/processing_vibevoice_asr.py
```diff
@@ -0,0 +1,319 @@
...
+        return self.apply_chat_template(
+            conversations,
+            tokenize=True,
+            add_generation_prompt=True,
+            return_dict=True,
+            **kwargs,
+        )
+
+    def batch_decode(
+        self, *args, return_as_dicts=False, extract_transcription=False, skip_special_tokens=True, **kwargs
```
ebezzam: What do you think of these different flags for user convenience? You can see the different outputs [here](https://github.com/huggingface/transformers/pull/43625/changes#r2764366984)

For reference, their post-processing method (which uses try/except): https://github.com/microsoft/VibeVoice/blob/0aa8cb4c64fc87f281f89ea5e733b7005eff36a8/vibevoice/processor/vibevoice_asr_processor.py#L490

docs/source/en/model_doc/vibevoice_asr.md
```diff
@@ -0,0 +1,332 @@
...
+print("=" * 60)
+for speaker_transcription in transcription:
+    print(speaker_transcription)
+
+# Remove speaker labels, only get raw transcription
+transcription = processor.batch_decode(generated_ids, extract_transcription=True)[0]
+print("\n" + "=" * 60)
+print("TRANSCRIPTION ONLY")
+print("=" * 60)
+print(transcription)
```
ebezzam, line 90: Thoughts on the different flags for user convenience? (outputs are just below)

We can discuss at the code [here](https://github.com/huggingface/transformers/pull/43625/changes#r2764362701) (in processor)

tests/models/vibevoice_asr/test_modeling_vibevoice_asr.py
ebezzam, line 1: Only integration tests are complete

tests/models/vibevoice_asr/test_processing_vibevoice_asr.py
ebezzam, line 1: TODO

ebezzam (commented):
@eustlb self-review of VibeVoice ASR if you can take a look 🤗

Integration tests are done. The reproducers from the original model can be found here (`reproducer_vibevoice_asr*.py`): https://gist.github.com/ebezzam/e1200bcecdc29e87dadd9d8423ae7ecb
(_the other "boilerplate" tests still need to be finished_)

docs/source/en/model_doc/vibevoice_asr.md
```diff
@@ -0,0 +1,332 @@
...
+The model supports various automatic speech recognition functionalities.
+
+### Speaker-timestamped transcription
+
+A notable feature of VibeVoice ASR is its ability to transcribe multi-speaker content, denoting who spoke and when.
+
+```python
+from transformers import AutoProcessor, VibeVoiceAsrForConditionalGeneration
+
+model_id = "bezzam/VibeVoice-ASR-7B"
```
ebezzam: update here and elsewhere with final

tests/models/vibevoice_asr/test_modeling_vibevoice_asr.py
```diff
@@ -0,0 +1,394 @@
...
+    def test_get_audio_features_attentions(self):
+        pass
+
+    @unittest.skip(reason="VibeVoiceAsr has unique audio processing with acoustic and semantic tokenizers.")
+    def test_get_audio_features_hidden_states(self):
+        pass
+
+    @unittest.skip(reason="VibeVoiceAsr has slight randomness due to VAE sampling.")
+    def test_determinism(self):
+        pass
```
ebezzam, line 180: Note that a few tests (this and below) need to be skipped because of the acoustic latent sampling causing two runs to be slightly different

tests/test_modeling_common.py
```diff
@@ -5235,30 +5235,36 @@ def test_get_audio_features_output(self, return_dict: bool | None):
                 outputs = model.get_audio_features(**inputs_dict)
 
             if return_dict in (True, None):
-                self.assertTrue(isinstance(outputs, ModelOutput), "get_audio_features() must return a BaseModelOutput")
+                self.assertTrue(
+                    isinstance(outputs, ModelOutput), "get_audio_features() must return a BaseModelOutputWithPooling"
+                )
```
ebezzam, line 5226: In reality it should be `BaseModelOutputWithPooling` as there is a check below that the `pooler_output` attribute exists

tests/test_modeling_common.py
```diff
@@ -5235,30 +5235,36 @@ def test_get_audio_features_output(self, return_dict: bool | None):
...
                 if getattr(self, "skip_test_audio_features_output_shape", False):
                     return
 
                 last_hidden_state_shape = outputs.last_hidden_state.shape
-                batch_size = inputs_dict["input_features"].shape[0]
+
+                if "input_features" in inputs_dict:
+                    batch_size = inputs_dict["input_features"].shape[0]
+                else:
+                    batch_size = inputs_dict["input_values"].shape[0]
```
ebezzam, line 5253: Audio models like VibeVoice ASR use `input_values` instead of `input_features`

tests/test_processing_common.py
```diff
@@ -104,7 +104,8 @@ class ProcessorTesterMixin:
     text_input_name = "input_ids"
     images_input_name = "pixel_values"
     videos_input_name = "pixel_values_videos"
-    audio_input_name = "input_features"
+    audio_input_name_values = "input_values"  # raw/normalized audio
+    audio_input_name = "input_features"  # computed features, e.g. Mel spectrogram, STFT
```
ebezzam, line 108: Add relevant testing logic for audio models like VibeVoice ASR that use `input_values` instead of `input_features`

tests/test_processing_common.py
```diff
@@ -742,7 +743,13 @@ def test_processor_with_multiple_inputs(self):
...
         # Verify output contains all expected keys
         for key in expected_output_keys:
-            self.assertIn(key, inputs)
+            if key == self.audio_input_name:
+                self.assertTrue(
+                    self.audio_input_name_values in inputs or self.audio_input_name in inputs,
+                    f"Expected either '{self.audio_input_name_values}' or '{self.audio_input_name}' in inputs",
+                )
+            else:
+                self.assertIn(key, inputs)
```
ebezzam, line 772: Add relevant testing logic for audio models like VibeVoice ASR that use `input_values` instead of `input_features`

ebezzam (commented):
@eustlb self-review of the tests, which take inspiration from Audio Flamingo 3

Note that I suggest some changes to core testing utilities, to handle audio models like VibeVoice ASR that take as input `input_values` instead of `input_features`

src/transformers/models/vibevoice_acoustic_tokenizer/feature_extraction_vibevoice_acoustic_tokenizer.py
```diff
@@ -68,7 +68,9 @@ def __call__(
         sampling_rate: int | None = None,
         padding: bool | str | PaddingStrategy | None = True,
         pad_to_multiple_of: int | None = None,
+        max_length: int | None = None,
         return_attention_mask: bool | None = True,
+        return_tensors: str | None = "pt",
```
ebezzam, line 75: Note: these changes were needed to make the common processing tests "happy"

ebezzam:
run-slow: auto, vibevoice_asr

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,521 @@
...
+
+    >>> # Initializing a model from the vibevoice_asr style configuration
+    >>> model = VibeVoiceAsrForConditionalGeneration(configuration)
+
+    >>> # Accessing the model configuration
+    >>> configuration = model.config
+    ```"""
+
+    model_type = "vibevoice_asr"
+    is_composition = True
```
eustlb: is_composition is not used in the lib (must have been removed or I don't know cause it is not used anywhere)

```suggestion
    model_type = "vibevoice_asr"
```

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,521 @@
...
+    VibeVoiceAcousticTokenizerModel,
+    VibeVoiceAcousticTokenizerPreTrainedModel,
+)
+
+
+logger = logging.get_logger(__name__)
+
+
+class VibeVoiceAsrEncoderConfig(VibeVoiceAcousticTokenizerConfig):
+    r"""
```
eustlb: the issue here is that it is misleading for the users, in our AudioLM paradigm there is an encoder that encode before projecting to the backbone language model. Here we'll have twice this config, and two encoders that are actually sub-parts of the encoding logic

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
eustlb: yep, this is not clear IMO

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
eustlb: ok but then we'll rather use self.config.acoustic_tokenizer_config.vae_std

eustlb (commented):
Nice work, looks already clean 👏
A first review to iterate on:

**encoder config**
This does not make sense IMO. I get it's been introduced to circumvent the fact that we cannot do `AutoModel.from_config` to initiate the acoustic/semantic tokenizer encoders, so we define it via modular. Shouldn't this mean we should rather open them to the API, meaning that we should be able to do `from transformers import VibevoiceAcousticTokenizerEncoderModel` and in `VibeVoiceAsrForConditionalGeneration` directly initiate the semantic/ acoustic encoder with those configs 
→  let's update vibevoice acoustic tokenizer

src/transformers/models/vibevoice_acoustic_tokenizer/configuration_vibevoice_acoustic_tokenizer.py
```diff
@@ -83,30 +238,55 @@ def __init__(
...
+            )
+        elif isinstance(encoder_config, dict):
+            encoder_config = VibeVoiceAcousticTokenizerEncoderConfig(**encoder_config)
+        self.encoder_config = encoder_config
 
-    @property
-    def upsampling_ratios(self):
-        return self.downsampling_ratios[::-1]
+        if decoder_config is None:
+            # NOTE (ebezzam/eustlb) add condition to deduce decoder config if encoder_config provided?
```
ebezzam: As discussed, maybe we do another fallback in case the user just wants to specify `encoder_config` in order to have a symmetric `decoder_config`?

src/transformers/models/vibevoice_acoustic_tokenizer/modular_vibevoice_acoustic_tokenizer.py
```diff
@@ -294,20 +309,42 @@ def __init__(self, config):
...
+        latents = hidden_states.permute(0, 2, 1)
+        return VibeVoiceAcousticTokenizerEncoderOutput(latents=latents, padding_cache=padding_cache)
 
 
 class VibeVoiceAcousticTokenizerDecoderStem(nn.Module):
     def __init__(self, config):
         super().__init__()
 
-        intermediate_channels = int(config.num_filters * 2 ** (len(config.decoder_depths) - 1))
+        intermediate_channels = int(config.num_filters * 2 ** (len(config.depths) - 1))
```
ebezzam, line 326: Since the encoder and decoder configs are split, we can just call it `depths` in the decoder

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
ebezzam: Now they are `VibeVoiceAcousticTokenizerEncoderModel`!

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
ebezzam: Now that we have a separate encoder config ([ref](https://github.com/ebezzam/transformers/blob/vibevoice_asr/src/transformers/models/vibevoice_acoustic_tokenizer/configuration_vibevoice_acoustic_tokenizer.py#L226-L228)), `self.config.acoustic_tokenizer_config` doesn't have an attribute `vae_std`. And I don't think it should since the semantic tokenizer doesn't need it?

So is using `self.config.acoustic_vae_std` fine? we could also shorten it to `self.config.vae_std` like in the tokenizer model ([ref](https://github.com/ebezzam/transformers/blob/522cf5314543131e7437867b9d6d92276a3b128d/src/transformers/models/vibevoice_acoustic_tokenizer/configuration_vibevoice_acoustic_tokenizer.py#L280))

ebezzam (commented):
@eustlb a self-review replying to your comments

eustlb (commented):
Nice for opening encoder/ decoder to the API to enable AutoModel!!
For config thing,  what would you think of doing:
```python
class VibeVoiceAcousticTokenizerConfig(PretrainedConfig):
    model_type = "vibevoice_acoustic_tokenizer"

    ...

    @property
    def encoder_config(self):
        return VibeVoiceAcousticTokenizerEncoderConfig(**self.to_dict())

    @property
    def decoder_config(self):
        return VibeVoiceAcousticTokenizerDecoderConfig(**self.to_dict())


class VibeVoiceAcousticTokenizerEncoderConfig(VibeVoiceAcousticTokenizerConfig):
    model_type = "vibevoice_acoustic_tokenizer_encoder"
    base_config_key = "encoder_config"


class VibeVoiceAcousticTokenizerDecoderConfig(VibeVoiceAcousticTokenizerConfig):
    model_type = "vibevoice_acoustic_tokenizer_decoder"
    base_config_key = "decoder_config"
```

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
eustlb: It's been a common practice for audio models to add custom fields (done before me, and I reproduced it), nevertheless, I think audio models are the only ones doing it so I would follow thier recommandation. Could we have it as a forward kwargs and use default if not specified? something like [here](https://github.com/huggingface/transformers/blob/9a7d817885aab0728cd4894798790e47607920c7/src/transformers/models/voxtral_realtime/modular_voxtral_realtime.py#L718-L724), but no need to log a warning I thing just explain in the forward docstring that it defaults to config value

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,360 @@
...
+        if isinstance(semantic_tokenizer_config, dict):
+            semantic_tokenizer_config["model_type"] = semantic_tokenizer_config.get(
+                "model_type", "vibevoice_acoustic_tokenizer_encoder"
+            )
+            semantic_tokenizer_config = CONFIG_MAPPING[semantic_tokenizer_config["model_type"]](
+                **semantic_tokenizer_config
+            )
+        elif semantic_tokenizer_config is None:
+            semantic_tokenizer_config = CONFIG_MAPPING["vibevoice_acoustic_tokenizer_encoder"](hidden_size=128)
+        self.semantic_tokenizer_config = semantic_tokenizer_config
```
eustlb: when chaning default values I would usually do it [like that](https://github.com/huggingface/transformers/blob/9a7d817885aab0728cd4894798790e47607920c7/src/transformers/models/voxtral_realtime/configuration_voxtral_realtime.py#L183-L197) **but we can leave it here as it is minimal**

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,360 @@
...
+    >>> model = VibeVoiceAsrForConditionalGeneration(configuration)
+
+    >>> # Accessing the model configuration
+    >>> configuration = model.config
+    ```"""
+
+    model_type = "vibevoice_asr"
+    sub_configs = {
+        "acoustic_tokenizer_config": AutoConfig,
+        "semantic_tokenizer_config": AutoConfig,
```
eustlb: maybe acoustic_tokenizer_encoder_config (and same for semantic) ?

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,360 @@
...
+            tokenizer_chunk_size = self.config.tokenizer_chunk_size
+        else:
+            if tokenizer_chunk_size % self.config.acoustic_tokenizer_config.hop_length != 0:
+                tokenizer_chunk_size = int(
+                    (tokenizer_chunk_size // self.config.acoustic_tokenizer_config.hop_length)
+                    * self.config.acoustic_tokenizer_config.hop_length
+                )
+                logger.warning(
+                    f"`tokenizer_chunk_size` has been adjusted to {tokenizer_chunk_size} to be a multiple of hop length ({self.config.acoustic_tokenizer_config.hop_length})."
+                )
```
eustlb, line 232: here I would rather raise a ValueError just saying it must be a multiple and let the user adapt, WDYT?

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,360 @@
...
+
+        with torch.no_grad():
+            acoustic_encoder_cache = None
+            semantic_encoder_cache = None
+            acoustic_latents = []
+            semantic_latents = []
```
eustlb: cannot help stopping to try to save lines lol
```suggestion
            acoustic_encoder_cache, semantic_encoder_cache = None, None
            acoustic_latents, semantic_latents = [], []
```

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
```diff
@@ -0,0 +1,360 @@
...
+
+        if cache_position is not None and cache_position[0] == 0:
+            if input_values is not None:
+                model_inputs["input_values"] = input_values
+            if padding_mask is not None:
+                model_inputs["padding_mask"] = padding_mask
+            if tokenizer_chunk_size is not None:
+                model_inputs["tokenizer_chunk_size"] = tokenizer_chunk_size
+
+        return model_inputs
```
eustlb: `cache_position` is about to be deprecated, can you rather use `is_first_iteration` that you can catch diretly:
```python
def prepare_inputs_for_generation(self, *arg, is_first_iteration=False, **kwargs):
```

src/transformers/models/vibevoice_asr/processing_vibevoice_asr.py
```diff
@@ -0,0 +1,317 @@
...
```
eustlb: `batch_decode` is deprecated, use decode directly

src/transformers/models/vibevoice_asr/processing_vibevoice_asr.py
eustlb: convenient! just the API is skip_special_tokens=False by default, so maybe we should align with that

src/transformers/models/vibevoice_asr/convert_vibevoice_asr_to_hf.py
```diff
@@ -0,0 +1,389 @@
...
+    )
+    parser.add_argument(
+        "--float32", action="store_true", help="Whether to use float32 precision. Default is bfloat16."
+    )
+    parser.add_argument(
+        "--max_shard_size",
+        type=str,
+        default="2.5GB",
+        help="Maximum shard size for safetensors files in GB.",
+    )
```
eustlb, line 372: no sure why we'd want to use this, can't we just use the default value from save_pretrained/ push_to_hub?

eustlb (commented):
Looking good! Let me just ping @ydshieh here before approving.
Indeed, he mentionned to me he's not a fan of the `expected_result....json` (I don't use them myself) and I'd like to get his opinion on that.
**Edit:** let me also ping you @ArthurZucker on that question

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
ebezzam: haha I did have the same reflex! just wasn't sure if that's done within transformers

src/transformers/models/vibevoice_asr/convert_vibevoice_asr_to_hf.py
ebezzam: I think the default was changed for `push_to_hub` recently? Because the [original checkpoint](https://huggingface.co/microsoft/VibeVoice-ASR/tree/main) seems to have this as a max, and a user reached out to me that they could load the original checkpoint but got OOM with my checkpoint (before adding this).

So to not disrupt people's workflows when the official checkpoint change, I think it would be best to set the max size? (Otherwise, for other cases I agree on using the default).

EDIT: https://github.com/huggingface/transformers/pull/42734 changed the default

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
ebezzam: yeah I think it's better in case they don't pay attention to warnings and are "surprised" by the `tokenizer_chunk_size` that is set. The error can suggest a value just under what they set (so they don't need to calculate themselves or trial/error)

src/transformers/models/vibevoice_asr/convert_vibevoice_asr_to_hf.py
eustlb: hum interesting, it is unclear to me how this could cause OOM issues since to my understanding the shard boundaries have no effect on device mapping

src/transformers/models/vibevoice_asr/modular_vibevoice_asr.py
ebezzam: yep agree on avoiding using custom generation fields to align with other modalities 👍 

I guess you wrote this comment before seeing below? `tokenizer_chunk_size` is exposed in the forward kwargs

src/transformers/models/vibevoice_asr/convert_vibevoice_asr_to_hf.py
ebezzam: I'm not too sure either, because I didn't see the exact error and they just told me they got out of memory. Maybe it's CPU RAM issue (only when loading) when the shard is too large?

eustlb (approved):
Nice work! thanks a lot @ebezzam 👏
for the unadressed:
- shard size: let's use the default behavior, it should work, let's have an issue raised if necessary so that we can upstream it
- test outputs in json: this can be latter changed an not important RN

→  LGTM

cc @ArthurZucker for a last review

eustlb (commented):
I actually have one comment concerning processor's decode:
what would you think of rather having a return_format=str, defaulting to "raw" and with possible values "raw", "parsed", "transcription_only"? + adding in the tests checking they work

ebezzam:
@eustlb done!
- single `return_format` option: https://github.com/ebezzam/transformers/blob/cd3e35bcad75b4ecf397f3d6a8163f087028c621/src/transformers/models/vibevoice_asr/processing_vibevoice_asr.py#L251
- processor test: https://github.com/ebezzam/transformers/blob/cd3e35bcad75b4ecf397f3d6a8163f087028c621/tests/models/vibevoice_asr/test_processing_vibevoice_asr.py#L119

ebezzam:
run-slow: vibevoice_acoustic_tokenizer, vibevoice_asr

src/transformers/models/vibevoice_asr/processing_vibevoice_asr.py
```diff
@@ -0,0 +1,339 @@
...
+            skip_special_tokens = kwargs.get("skip_special_tokens", False)
+
+        decoded = self.tokenizer.decode(*args, skip_special_tokens=skip_special_tokens, **kwargs)
+
+        if return_format != "raw":
+            decoded = [self._parse_dict_output(text) for text in decoded]
+        if return_format == "transcription_only":
+            return [self._extract_content_from_dict(dict_output) for dict_output in decoded]
+        else:
+            return decoded
```
eustlb: let's use clearer code paths
```suggestion
        if return_format not in ["raw", "parsed", "transcription_only"]:
            raise ValueError(f"return_format must be one of {return_types}.")
        elif return_format != "raw":
            kwargs["skip_special_tokens"] = True  # for other formats this does not make sense, we can silently ignore
        
        decoded = self.tokenizer.decode(*args, **kwargs)
        
        if return_format == "parsed":
            return [self._parse_dict_output(el) for el in decoded]
        elif return_format == "transcription_only":
            return [self._extract_content_from_dict(el) for el in decoded]
        return decoded
```

ebezzam: I'll take everything but the parsing logic at the end is slightly different
```python
        if return_format in ["parsed", "transcription_only"]:
            decoded = [self._parse_dict_output(el) for el in decoded]
        if return_format == "transcription_only":
            return [self._extract_content_from_dict(el) for el in decoded]
        else:
            return decoded
```

ebezzam:
run-slow: vibevoice_asr

src/transformers/models/vibevoice_asr/processing_vibevoice_asr.py
eustlb: ha indeed! my bad

ebezzam:
run-slow: vibevoice_asr

docs/source/en/model_doc/vibevoice_asr.md
```diff
@@ -0,0 +1,455 @@
...
+print("\n" + "=" * 60)
+print("DICT OUTPUT")
+print("=" * 60)
+dict_output = pipe.processor.extract_speaker_dict(outputs[0]["generated_text"])
+print(dict_output)
+
+print("\n" + "=" * 60)
+print("TRANSCRIPT OUTPUT")
+print("=" * 60)
+transcription = pipe.processor.extract_transcription(outputs[0]["generated_text"])
```
ebezzam, line 431: For convenient pipeline usage, these methods are needed because the `any-to-any` pipeline doesn't call `self.processor.decode` but rather `self.tokenizer.decode` ([here](https://github.com/huggingface/transformers/blob/91d7b6456c5ef62d72ffd9faac5d21260b91df5b/src/transformers/pipelines/any_to_any.py#L448) to [here](https://github.com/huggingface/transformers/blob/main/src/transformers/processing_utils.py#L1870-L1915)).

This means that passing `return_format` to `pipe` isn't taken into account to reach these methods for parsing the JSON-like output of VibeVoice ASR.

docs/source/en/model_doc/vibevoice_asr.md
```diff
@@ -0,0 +1,455 @@
...
+  Unlike conventional ASR models that slice audio into short chunks (often losing global context), VibeVoice ASR accepts up to **60 minutes** of continuous audio input within 64K token length. This ensures consistent speaker tracking and semantic coherence across the entire hour.
+
+- **👤 Customized Hotwords**:
+  Users can provide customized hotwords (e.g., specific names, technical terms, or background info) to guide the recognition process, significantly improving accuracy on domain-specific content.
+
+- **📝 Rich Transcription (Who, When, What)**:
+  The model jointly performs ASR, diarization, and timestamping, producing a structured output that indicates *who* said *what* and *when*.
+  
+- **🌍 Multilingual & Code-Switching Support**:
+  It supports over 50 languages, requires no explicit language setting, and natively handles code-switching within and across utterances. Language distribution can be found [here](#language-distribution).
```
ArthurZucker, line 44: Huge! 🚀

docs/source/en/model_doc/vibevoice_asr.md
```diff
@@ -0,0 +1,455 @@
...
+
+"""
+WITHOUT CONTEXT: Revevoices is a novel framework designed for generating expressive, long-form, multi-speaker conversational audio.
+WITH CONTEXT   : VibeVoice is this novel framework designed for generating expressive, long-form, multi-speaker, conversational audio.
+"""
+```
+
+### Batch inference
+
+Batch inference is possible by passing a list of audio and (if provided) a list of prompts of equal length.
```
ArthurZucker:
```suggestion
Batch inference is possible by passing a list of audio and (if provided) a list of prompts padded to equal length.
```
Or padding is not possible?

docs/source/en/model_doc/vibevoice_asr.md
```diff
@@ -0,0 +1,455 @@
...
+transcription = processor.decode(generated_ids, return_format="transcription_only")
+
+print(transcription)
+```
+
+### Adjusting tokenizer chunk (e.g. if out-of-memory)
+
+A key feature of VibeVoice ASR is that it can transcribe up to 60 minutes of continuous audio. This is done by chunking audio into 60-second segments (1440000 samples at 24kHz) and caching the convolution states between each segment.
+
+However, if chunks of 60 seconds are too large for your device, the `tokenizer_chunk_size` argument passed to `generate` can be adjusted. *Note it should be a multiple of the hop length (3200 for the original acoustic tokenizer).*
```
ArthurZucker: Naming is a bit weird coming from LLMs but its fine and its true that it IS a `tokenizer`, just and audio one! SG

docs/source/en/model_doc/vibevoice_asr.md
ArthurZucker: sg, we might want to update the pipeline in the futur!

docs/source/en/model_doc/vibevoice_asr.md
ArthurZucker, line 1: very nice and detailed md! kudos

src/transformers/models/vibevoice_acoustic_tokenizer/configuration_vibevoice_acoustic_tokenizer.py
```diff
@@ -101,12 +102,54 @@ def __init__(
         self.vae_std = vae_std
 
     @property
-    def upsampling_ratios(self):
-        return self.downsampling_ratios[::-1]
+    def hop_length(self):
+        return int(np.prod(self.downsampling_ratios))
```
ArthurZucker: not sure depending on np just for this is worth it :)

src/transformers/models/vibevoice_acoustic_tokenizer/modeling_vibevoice_acoustic_tokenizer.py
```diff
@@ -406,20 +421,42 @@ def __init__(self, config):
...
+                per_layer_in_channels.extend([block.mixer.conv.in_channels for block in layer.stage])
+            per_layer_padding.append(self.head.causal_padding)
+            per_layer_in_channels.append(self.head.conv.in_channels)
+
+            padding_cache = VibeVoiceAcousticTokenizerConv1dPaddingCache(
+                num_layers=len(per_layer_padding),
+                per_layer_padding=per_layer_padding,
+                per_layer_padding_mode=["constant"] * len(per_layer_padding),
+                per_layer_in_channels=per_layer_in_channels,
+            )
```
ArthurZucker: I think @eustlb implemented this in a "cleaner" way lazily in voxtral, wuold be nice to use it https://github.com/huggingface/transformers/blob/91d7b6456c5ef62d72ffd9faac5d21260b91df5b/src/transformers/models/voxtral_realtime/modular_voxtral_realtime.py#L53-L71

src/transformers/models/vibevoice_acoustic_tokenizer/modeling_vibevoice_acoustic_tokenizer.py
```diff
@@ -490,29 +527,36 @@ def __init__(self, config):
...
         )
+        self.post_init()
+
+    def forward(self, hidden_states, padding_cache=None, use_cache=False, **kwargs):
+        if use_cache and padding_cache is None:
+            per_layer_padding = [self.stem.conv.causal_padding]
+            per_layer_in_channels = [self.stem.conv.conv.in_channels]
+            per_layer_padding.extend([block.mixer.causal_padding for block in self.stem.stage])
+            per_layer_in_channels.extend([block.mixer.conv.in_channels for block in self.stem.stage])
+            for layer in self.conv_layers:
```
ArthurZucker: same comment!

src/transformers/models/vibevoice_asr/processing_vibevoice_asr.py
```diff
@@ -0,0 +1,378 @@
...
+            self.audio_token = "<|box_start|>"
+            self.audio_token_id = tokenizer.convert_tokens_to_ids(self.audio_token)
+        else:
+            self.audio_token = tokenizer.audio_token
+            self.audio_token_id = tokenizer.audio_token_id
+
+        if not hasattr(tokenizer, "audio_duration_token"):
+            self.audio_duration_token = "<|AUDIO_DURATION|>"
+        else:
+            self.audio_duration_token = tokenizer.audio_duration_token
```
ArthurZucker: there should be better ways to do this no? like default `tokenizer_kwargs`?

ArthurZucker: `self.audio_token_id = tokenizer.audio_token_id` should always work if properly added

src/transformers/models/vibevoice_asr/processing_vibevoice_asr.py
```diff
@@ -0,0 +1,378 @@
...
+        audio = make_list_of_audio(audio)
+        if len(text) != len(audio):
+            raise ValueError(f"Got {len(text)} text but {len(audio)} audios; they must match 1:1.")
+        data = self.feature_extractor(audio, **audio_kwargs)
+
+        # Replace audio duration placeholders in text
+        audio_lengths = data["padding_mask"].sum(dim=-1).cpu().numpy()
+        audio_durations = audio_lengths / self.feature_extractor.sampling_rate
+        for i in range(len(text)):
+            text[i] = text[i].replace(self.audio_duration_token, f"{audio_durations[i]:.2f}")
```
ArthurZucker: idk if it makes a big diff, i would compile the regex replace once to apply it but its a nit

src/transformers/models/vibevoice_asr/processing_vibevoice_asr.py
```diff
@@ -0,0 +1,378 @@
...
+
+        # Expand audio tokens in text
+        num_audio_tokens = np.ceil(audio_lengths / audio_kwargs["pad_to_multiple_of"]).astype(int).tolist()
+        for i, num_tokens in enumerate(num_audio_tokens):
+            text[i] = re.sub(re.escape(self.audio_token), self.audio_token * num_tokens, text[i])
```
ArthurZucker: smae here compile the regex outside then apply it

ArthurZucker (approved):
Great work!! 
Make sure the conv1D cache API is synch / standard and let's shipppp

docs/source/en/model_doc/vibevoice_asr.md
ebezzam: I should word it better, I meant that the user should provide the same number of audio and prompts to the processor. For example:
```python
audio = [
    "https://huggingface.co/datasets/bezzam/vibevoice_samples/resolve/main/realtime_model/vibevoice_tts_german.wav",
    "https://huggingface.co/datasets/bezzam/vibevoice_samples/resolve/main/example_output/VibeVoice-1.5B_output.wav"
]
prompts = ["About VibeVoice", None]

processor = AutoProcessor.from_pretrained("bezzam/VibeVoice-ASR-7B")
inputs = processor.apply_transcription_request(audio, prompt=prompts).to(model.device, model.dtype)
```

docs/source/en/model_doc/vibevoice_asr.md
ebezzam: ah I can see the confusion! a bit wordy but what about `acoustic_tokenizer_chunk_size`?

docs/source/en/model_doc/vibevoice_asr.md
ArthurZucker: ah okay!

src/transformers/models/vibevoice_acoustic_tokenizer/modular_vibevoice_acoustic_tokenizer.py
eustlb, line 102: here we can inherit directly from VoxtralRealtimeCausalConv1d

src/transformers/models/vibevoice_acoustic_tokenizer/modular_vibevoice_acoustic_tokenizer.py
```diff
@@ -411,8 +416,8 @@ def _init_weights(self, module):
 class VibeVoiceAcousticTokenizerModel(VibeVoiceAcousticTokenizerPreTrainedModel):
     def __init__(self, config):
         super().__init__(config)
-        self.encoder = VibeVoiceAcousticTokenizerEncoder(config)
-        self.decoder = VibeVoiceAcousticTokenizerDecoder(config)
+        self.encoder = AutoModel.from_config(config.encoder_config)
+        self.decoder = AutoModel.from_config(config.decoder_config)
```
eustlb, line 421: yes!

src/transformers/models/vibevoice_acoustic_tokenizer/configuration_vibevoice_acoustic_tokenizer.py
```diff
@@ -101,12 +102,54 @@ def __init__(
...
+
+    @property
+    def decoder_config(self):
+        config_dict = self.to_dict()
+        config_dict["depths"] = list(reversed(config_dict["depths"]))
+        return VibeVoiceAcousticTokenizerDecoderConfig(**config_dict)
+
+
+class VibeVoiceAcousticTokenizerEncoderConfig(VibeVoiceAcousticTokenizerConfig):
+    r"""
```
eustlb, line 120: yes!

eustlb (approved):
One last thing but kudos @ebezzam, great (and clean) work!

src/transformers/models/vibevoice_acoustic_tokenizer/modular_vibevoice_acoustic_tokenizer.py
ebezzam: (discussed offline) for a couple reasons (exposing `groups` parameter in [VoxtralRealtimeCausalConv1d](https://github.com/huggingface/transformers/blob/dd5f1a1ec8d7133329548f6be4dcbb0693750c99/src/transformers/models/voxtral_realtime/modeling_voxtral_realtime.py#L201) and state dict changing which would affect [current checkpoint](https://huggingface.co/microsoft/VibeVoice-AcousticTokenizer)), we leave this as a TODO

PR #43626, opened 2026-01-30
[CB] Keep order of incoming requests

remi-or:
The test `tests/generation/test_paged_attention.py::TestBatchGeneration::test_generate_batch_consistency` is failing because CB shuffles incoming request to maximize prefix caching. This is not a problem unless the user expects outputs to have the same order as the inputs when using `generate_batch`, which is reasonable. This PR ensures this and fixes the test on GPU.

On CPU, the test is still half broken, probably because of numerical difference between the implementations. I added a TODO and a FIXME to later check this out. In the meantime, this PR 1. fixes the CI and 2. adds order to the API which was needed anyway


Cyrilvallez (approved):
Trusting you on this!

PR #43559, opened 2026-01-28
[`feat`] Allow loading T5Gemma2Encoder with AutoModel

tomaarsen:
# What does this PR do?

* Allow the encoder of T5Gemma2 to be loaded standalone

### Details
This is valuable for Sentence Transformers, which may want to load the encoder only (see https://github.com/huggingface/sentence-transformers/pull/3604). Here, we grab and train the encoder only, resulting in e.g.: https://huggingface.co/tomaarsen/t5gemma2-270m-gooaq-cmnrl

#### Usage:
```python
from transformers import T5Gemma2Encoder, AutoTokenizer
import torch

model_name = "tomaarsen/t5gemma2-270m-gooaq-cmnrl"
model = T5Gemma2Encoder.from_pretrained(model_name)
processor = AutoTokenizer.from_pretrained(model_name)

queries = [
    "Which planet is known as the Red Planet?",
]
documents = [
    "Venus is often called Earth's twin because of its similar size and proximity.",
    "Mars, known for its reddish appearance, is often referred to as the Red Planet.",
    "Jupiter, the largest planet in our solar system, has a prominent red spot.",
    "Saturn, famous for its rings, is sometimes mistaken for the Red Planet.",
]

query_inputs = processor(text=queries, truncation=True, padding=True, return_tensors="pt")
document_inputs = processor(text=documents, truncation=True, padding=True, return_tensors="pt")

with torch.no_grad():
    query_embeddings = model(**query_inputs).last_hidden_state[:, 0, :]
    document_embeddings = model(**document_inputs).last_hidden_state[:, 0, :]

def mean_pooling(model_output, attention_mask):
    token_embeddings = model_output.last_hidden_state
    input_mask_expanded = attention_mask.unsqueeze(-1).expand(token_embeddings.size()).float()
    return torch.sum(token_embeddings * input_mask_expanded, 1) / torch.clamp(input_mask_expanded.sum(1), min=1e-9)

query_embeddings = mean_pooling(model(**query_inputs), query_inputs['attention_mask'])
document_embeddings = mean_pooling(model(**document_inputs), document_inputs['attention_mask'])

similarities = torch.nn.functional.cosine_similarity(
    query_embeddings.unsqueeze(1), document_embeddings.unsqueeze(0), dim=-1
)
print(similarities.tolist())
# [[0.37183186411857605, 0.8092442750930786, 0.6081508994102478, 0.7218592762947083]]
# As expected: The second document is most similar to the query.
```

[...]

PR #43633, opened 2026-01-30
🚨 T5Gemma2 model structure

zucchini-nlp:
# What does this PR do?

Makes sure that the attn implementation is set to all sub-configs. The `config.encoder.text_config` was not getting its attn set because we aren't passing it to `PreTrainedModel.__init__`. We can't change the model structure without breaking so I manually re-added a call to `self.adjust_attn_implemetation` in modeling code

Also deleted `__setattr__`, not sure what is the reason behind having them. Composite configs usually don't need to force-set the same attr in all subconfigs magically


src/transformers/models/t5gemma2/modeling_t5gemma2.py
```diff
@@ -804,6 +804,11 @@ def __init__(
 
         self.text_config = text_config
 
+        # Set attn implementation manually because `text-config` is never passed to `super()`
+        self.text_config._attn_implementation_internal = self._check_and_adjust_attn_implementation(
+            self.text_config._attn_implementation, is_init_check=True
+        )
+
```
zucchini-nlp: @tomaarsen here is why the attn implementation was not being set

tomaarsen: Just making sure: this is not also necessary for the `vision_config`, right?
Will review in more details next week.

zucchini-nlp: nope, the vision config is used a few lines above to init a `PreTrainedModel` and thus it is passed to `PreTrainedModel.__init__`

tomaarsen (approved):
Just ran some more extensive tests, this seems to work for me now. The attn_implementation is set automatically and my ST training works as expected. 
This PR should supersede parts of #43559 now, ~~which can be closed once this is merged.~~

zucchini-nlp:
Will finish up in a while and ask for review

tomaarsen:
Somewhat related: If I train a T5GemmaEncoder, I end up with a repository like: https://huggingface.co/tomaarsen/t5gemma2-270m-gooaq-cmnrl/tree/main
Note that the `model_type` is `t5gemma2_encoder`, which I can't load with `AutoConfig` like other architectures. Perhaps I should still expand on #43559 to turn `t5gemma2_encoder` into an architecture with AutoConfig and AutoModel support? Otherwise I still can't conveniently load https://huggingface.co/tomaarsen/t5gemma2-270m-gooaq-cmnrl.

- Tom Aarsen

zucchini-nlp:
@tomaarsen yes, it's expected that `AutoConfig` will not work on the encoder-part. If it's needed for ST, we need the other PR you had as well. I was assuming ST will load directly with `T5GemmaEncoderConfig` in a similar fashion to old T5 family

tomaarsen:
> If it's needed for ST, we need the other PR you had as well.

Agreed, will re-add it. 

> I was assuming ST will load directly with `T5GemmaEncoderConfig` in a similar fashion to old T5 family

I will, but the old T5 family is also loaded with:
load AutoConfig -> Check config type -> Check for edge cases (T5, MT5, T5Gemma, T5Gemma2), otherwise AutoModel

And then I need to be able to use
```python

config = AutoConfig.from_pretrained("tomaarsen/t5gemma2-270m-gooaq-cmnrl")
```

- Tom Aarsen

zucchini-nlp:
Failing test is flaky, ready for review

src/transformers/models/t5gemma2/modeling_t5gemma2.py
vasqu: Hmm, could we refactor with the conversion mapping / checkpoint conversion mapping? Iiuc, if we refactor the text specific things into its own module then we won't have this issue

I think it needs conversion mapping because you also want to use the encoder as standalone model

vasqu (commented):
The config changes are good! 

I just think we could maybe convert the model instead via conversion_mapping or similar? This is a quick and dirty workaround so would be pro making this properly converted into its own encoder text module if possible

src/transformers/models/t5gemma2/modeling_t5gemma2.py
zucchini-nlp: will be a huge change though, let me see

zucchini-nlp:
run-slow: t5gemma2

zucchini-nlp:
run-slow: t5gemma2

zucchini-nlp:
run-slow: t5gemma2

zucchini-nlp:
run-slow: t5gemma2

src/transformers/models/t5gemma2/modular_t5gemma2.py
```diff
@@ -815,50 +793,130 @@ def prepare_decoder_input_ids_from_labels(self, input_ids):
...
+        self,
+        input_ids: torch.LongTensor | None = None,
+        attention_mask: torch.Tensor | None = None,
+        position_ids: torch.LongTensor | None = None,
+        inputs_embeds: torch.FloatTensor | None = None,
+        # Unused for processor compatibility kept in signature.
+        token_type_ids: torch.Tensor | None = None,
+        **kwargs: Unpack[TransformersKwargs],
+    ) -> BaseModelOutput:
+        del token_type_ids
```
vasqu:
```suggestion
```
I don't think we need this (I know you didn't add this but can be removed nonetheless)

src/transformers/models/t5gemma2/modular_t5gemma2.py
```diff
@@ -937,56 +978,26 @@ def forward(
...
-        position_embeddings = {}
-        for layer_type in self.text_config.layer_types:
-            position_embeddings[layer_type] = self.rotary_emb(hidden_states, position_ids, layer_type)
+            image_mask = self.get_image_placeholder_mask(
+                input_ids, inputs_embeds=inputs_embeds, image_features=image_features
+            )
 
-        # dropout
-        hidden_states = self.dropout(hidden_states)
+            inputs_embeds = inputs_embeds.masked_scatter(image_mask, image_features)
```
vasqu, line 990: Much nicer thanks (also more aligned to what other VLMs do)

tests/models/t5gemma2/test_modeling_t5gemma2.py
```diff
@@ -1013,3 +1023,54 @@ def test_retain_grad_hidden_states_attentions(self):
     )
     def test_sdpa_can_dispatch_on_flash(self):
         pass
+
+
+@require_torch_accelerator
+@slow
+class T5Gemma2IntegrationTest(unittest.TestCase):
+    def tearDown(self):
+        cleanup(torch_device, gc_collect=True)
```
vasqu, line 1037: Should have the same on setup?

vasqu (approved):
LGTM, thanks a lot. Much better this way! Just one nit on the test to maybe add cleanup on setup as well

tests/models/t5gemma2/test_modeling_t5gemma2.py
zucchini-nlp: I see it's not always used for setup. Can add, no harm for sure. Seems like it was needed with compile tests or heavy models, this one is 270m

src/transformers/models/t5gemma2/modular_t5gemma2.py
zucchini-nlp: oh right, it's prob "google" code standards, they delete unused vars explicitly 😅

tests/models/t5gemma2/test_modeling_t5gemma2.py
vasqu: Rather keep it consistent and it doesn't hurt

zucchini-nlp: i'll wait for Tom's PR and resolve conflicts tmrw

tomaarsen: @zucchini-nlp you mean that you want #43559 to be merged first? That works for me. I can help give a final review of this (optionally in the context of ST) once you're ready.

zucchini-nlp: yes, so we don't break the main I'd love to merge that one first and test the current branch again

tomaarsen: Agreed, I've merged my PR.

zucchini-nlp:
Checked with rebase, everything is still working and tests are passing. Will merge after CI turns green

PR #43636, opened 2026-01-30
Add _metrics dict to Trainer for custom metric logging

abigailllr:
Adds a private _metrics dict to the Trainer class that allows custom trainers to log metrics without overriding log.   

Custom trainers can now simply do:                                                                                                                                                                                                          
`self._metrics[mode]["my_metric"].append(value)`                                                                          
                                                                                                                                 
And the metrics will be automatically averaged, merged into logs and cleared when log() is called.                                                                                                                                                          

Fixes [#43599 ](https://github.com/huggingface/transformers/issues/43599)

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -30,6 +30,7 @@ import sys
 import tempfile
 import time
 import warnings
+from collections import defaultdict
 from collections.abc import Callable, Iterator, Mapping
 from functools import partial
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


Rocketknight1:
cc @qgallouedec @SunMarc

src/transformers/trainer.py
```diff
@@ -435,6 +436,8 @@ def __init__(
         self._memory_tracker = TrainerMemoryTracker(self.args.skip_memory_metrics)
         self._memory_tracker.start()
 
+        self._metrics = {"train": defaultdict(list), "eval": defaultdict(list)}
```
qgallouedec, line 439: I'm wondering if we can leverage this `self._metrics` to log the loss

abigailllr: though about it, but loss uses tensor accumulation to avoid TPU sync overhead. metrics is more for auxiliary custom metrics that don't need that optimization

PR #43639, opened 2026-01-31
Fix Quickstart Colab dependency mismatch

Justynita:
# What does this PR do?

The Quickstart "Open in Colab" example can fail in a fresh Google Colab environment due to an incompatible preinstalled version of `huggingface_hub`.

This PR updates the installation command to explicitly constrain `huggingface_hub` to the version range required by `transformers`, ensuring the example runs as intended.


Rocketknight1:
We definitely do not want to constrain `huggingface_hub<1.0` for modern versions of Transformers! I'm not sure how you encountered this bug.

PR #43651, opened 2026-01-31
Add _loss_is_scaled_for_ga to allow custom trainers to control gradient accumulation loss scaling

abigailllr:
Added a _loss_is_scaled_for_ga property that custom trainers can override to explicitly control gradient accumulation loss scaling. The default implementation preserves backward compatibility. Custom trainers can now simply override this property to return False instead of manipulating model_accepts_loss_kwargs.

Fixes https://github.com/huggingface/transformers/issues/43595?issue=huggingface%7Ctransformers%7C43604

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -1941,7 +1941,7 @@ class Trainer:
         Override to return False if you implement custom compute_loss that needs
         the Trainer to handle gradient accumulation scaling.
         """
-        return self.model_accepts_loss_kwargs and self.compute_loss_func is None
+        return self.model_accepts_loss_kwargs or self.compute_loss_func is not None
 
     def compute_loss(
         self,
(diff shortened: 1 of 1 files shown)


Rocketknight1:
cc @qgallouedec

qgallouedec:
If I understand correctly `model_accepts_loss_kwargs` checks if the model forward has kwargs, but is actually used to decide if the loss should be scaled or not, am I right? The underlying assumption if that if the model accepts kwargs, then it takes `num_items_in_batch` as argument, which means that it scales the loss by itself.
TBH, I'm wondering if we should instead aim for a more ambitious refactor here.

abigailllr:
> If I understand correctly `model_accepts_loss_kwargs` checks if the model forward has kwargs, but is actually used to decide if the loss should be scaled or not, am I right? The underlying assumption if that if the model accepts kwargs, then it takes `num_items_in_batch` as argument, which means that it scales the loss by itself. TBH, I'm wondering if we should instead aim for a more ambitious refactor here.

yees, thats right. I'd be open to a more ambitious refactor, do you have a specific direction in mind?

PR #43654, opened 2026-01-31
fix(tokenizer): Avert special token property overwrites in batch add_tokens calls

harshaljanjani:
### What does this PR do?

→ Fixes `test_modeling_big_bird.py::BigBirdModelIntegrationTest::test_fill_mask`.

For more details on reproducing the bug, please visit the linked issue!

Fixes #43653.

### Before submitting
      Pull Request section?
      to it if that's the case.

**CI Failure:**

Image

**Current Output:**

Image

**Output After the Fix:**

Image


src/transformers/models/big_bird/tokenization_big_bird.py
```diff
@@ -135,6 +136,11 @@ def __init__(
             **kwargs,
         )
 
+        if isinstance(mask_token_obj, AddedToken):
+            mask_id = self._tokenizer.token_to_id(str(mask_token_obj))
+            if mask_id is not None:
+                self._tokenizer.add_special_tokens([mask_token_obj])
+
```
ArthurZucker: hey this is wrong 😓 the call to `super()` should already be adding all of the `special` tokens. 
The reason they are skipped when decoding is probably because `skip_special_tokens=True` by default

harshaljanjani: Thanks for the review @ArthurZucker; as it turns out, I underestimated the bug, but I think this is a much more informed reasoning chain since it required a few layers deeper analysis.

→ `BigBirdTokenizer` defines mask_token with `lstrip=True`, but [this batch add_tokens call](https://github.com/harshaljanjani/transformers/blob/1c35dbb62097c2d89ab52b424a3df805e1e18bd6/src/transformers/tokenization_utils_tokenizers.py#L366) processes multiple [MASK] copies with conflicting `lstrip` values, one with `lstrip=True` from `_special_tokens_map`, one with `lstrip=False` from saved configs (written code to demonstrate the same; output attached with [TRACE] notes showing the conflicting values).
→  In [Rust's add_tokens method](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/added_vocabulary.rs#L258-L323). `AddedToken` has a [Hash implementation](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/added_vocabulary.rs#L90-L94) using only content "[MASK]", but [Eq/PartialEq](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/added_vocabulary.rs#L15) checks all fields. Two tokens with same content but different `lstrip` are equal for hashing but different for equality.
→ Within a batch, [existing.contains()](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/added_vocabulary.rs#L291) uses full equality, so `[MASK](lstrip=False)` doesn't match `[MASK](lstrip=True)` and [overwrites it](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/added_vocabulary.rs#L309). Whichever appears later is seen as the final state, in our buggy case, `lstrip=False`.
So, the fix would be to make [add_tokens calls](https://github.com/harshaljanjani/transformers/blob/1c35dbb62097c2d89ab52b424a3df805e1e18bd6/src/transformers/tokenization_utils_tokenizers.py#L368-L372) for each special token, so the correct `lstrip=True` version is seen as the final state.

The current output:

before_bigbird

The fixed output would look something like the following:

after_bigbird

src/transformers/tokenization_utils_tokenizers.py
```diff
@@ -365,6 +365,12 @@ def __init__(self, *args, **kwargs):
                 # These tokens are from the special tokens map
                 self.add_tokens(tokens)
 
+        for special_token_value in self._special_tokens_map.values():
+            if special_token_value is not None and isinstance(special_token_value, AddedToken):
+                if not special_token_value.special:
+                    special_token_value.special = True
+                self._tokenizer.add_tokens([special_token_value])
+
```
ArthurZucker: above you have:
```python
        # if some of the special tokens are not already in the tokenizer, add them
        # V5: Check both named special tokens and extra special tokens
        # Iterate over _special_tokens_map to preserve AddedToken properties (lstrip, rstrip, etc.)
        for special_token_value in self._special_tokens_map.values():
            if special_token_value is None:
                continue
            if str(special_token_value) not in encoder and special_token_value not in tokens_to_add:
                tokens_to_add.append(special_token_value)
```
so we are doing it twice which I don't think makes sense. 

Now I am all in for a bug fix, so the bug you are fixing -> add an equivalent small test for bigbird maybe? 

Your deep dive is interesting, I did not fully check it, but whatever was serialized (in the tokenizer_config.json or tokenizer.json) takes precedence

ArthurZucker (commented):
I need to check your answer a bit more

ArthurZucker:
> → Within a batch, [existing.contains()](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/added_vocabulary.rs#L291) uses full equality, so [MASK](lstrip=False) doesn't match [MASK](lstrip=True) and [overwrites it](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/tokenizer/added_vocabulary.rs#L309). Whichever appears later is seen as the final state, in our buggy case, lstrip=False.

This is the expected behavior yes.

ArthurZucker:
> one with lstrip=False from saved configs 

Config always takes precedence

src/transformers/tokenization_utils_tokenizers.py
harshaljanjani: Thanks for being so patient with the review! Hopefully this addresses the concerns :)

> Config takes precedence.

The bug tmk, is that `[MASK]` appears with conflicting properties within the preset itself, in **both** `_special_tokens_map` (with `lstrip=True` from [google/bigbird-roberta-base/tokenizer_config.json](https://huggingface.co/google/bigbird-roberta-base/resolve/main/tokenizer_config.json)) and `_extra_special_tokens` (with default properties `lstrip=False, normalized=False` when loaded from [google/bigbird-roberta-base/spiece.model](https://huggingface.co/google/bigbird-roberta-base/resolve/main/spiece.model)), and thus the fix is needed since it conflicts within the preset itself.

image<br>

> Your deep dive is interesting, I did not fully check it, but whatever was serialized (in the tokenizer_config.json or tokenizer.json) takes precedence

In this case it doesn't which is the MO of the fix. However, I did find a more efficient way to do it thanks to your review: I could de-duplicate, since the whole reason it's happening is because of the duplication of the mask. We could prevent the duplicate `[MASK]` from entering the batch in the first place instead of checking after. Turns out that works a charm as well. I initially tried to solve this by making individual `add_tokens` calls for each `_special_tokens_map` entry **after** the batch to ensure the correct properties reach the backend, but I think this is a better solution. I've also added a test case per your suggestion.

Hope this better aligns with what the norm is by design here, thanks for taking the time!!

harshaljanjani:
Thanks for taking the time to review! I've left a reply [here](https://github.com/huggingface/transformers/pull/43654#discussion_r2759109412) :)
Also, [ci/circleci: tests_exotic_models](https://circleci.com/gh/huggingface/transformers/2143035) seems to be flaky, the failing test appears unrelated to this change and is passing locally with the latest version of the repo (commit: https://github.com/huggingface/transformers/commit/affcf45994a8b356f96c2a58e88ed8d679e46539).

image

src/transformers/tokenization_utils_tokenizers.py
```diff
@@ -346,7 +346,7 @@ def __init__(self, *args, **kwargs):
 
         # Also check extra special tokens
         for token in self._extra_special_tokens:
-            if str(token) not in encoder and token not in tokens_to_add:
+            if str(token) not in encoder and str(token) not in {str(t) for t in tokens_to_add}:
```
ArthurZucker: Can you fix the complexity? :) 
Create `{str(t) for t in tokens_to_add}` once to then check existence in O(1) instead of O(len(tokens_to_add) :)

harshaljanjani:
Resolved :)

tests/tokenization/test_tokenization_utils.py
```diff
@@ -352,3 +352,29 @@ def test_special_tokens_overwrite(self):
...
+        tokenizer = BigBirdTokenizer.from_pretrained("google/bigbird-roberta-base")
+
+        # Check that mask_token in _special_tokens_map has lstrip=True
+        mask_in_special = tokenizer._special_tokens_map.get("mask_token")
+        self.assertIsNotNone(mask_in_special)
+        self.assertTrue(mask_in_special.lstrip, "mask_token in _special_tokens_map should have lstrip=True")
+        mask_id = tokenizer.convert_tokens_to_ids("[MASK]")
+
+        # Check that the backend also has lstrip=True
+        backend_mask = tokenizer._tokenizer.get_added_tokens_decoder()[mask_id]
```
ArthurZucker: this is not the test we should enforce it was already working

tests/tokenization/test_tokenization_utils.py
```diff
@@ -352,3 +352,29 @@ def test_special_tokens_overwrite(self):
...
+    @slow
+    def test_mask_token_lstrip_preserved(self):
+        from transformers import BigBirdTokenizer
+
+        tokenizer = BigBirdTokenizer.from_pretrained("google/bigbird-roberta-base")
+
+        # Check that mask_token in _special_tokens_map has lstrip=True
+        mask_in_special = tokenizer._special_tokens_map.get("mask_token")
+        self.assertIsNotNone(mask_in_special)
+        self.assertTrue(mask_in_special.lstrip, "mask_token in _special_tokens_map should have lstrip=True")
```
ArthurZucker: this already passes for me locally

ArthurZucker (commented):
Can you work a bit more on the test please? let's explicit was is not passing on main today
```python
# on main:
In [3]: tokenizer._special_tokens_map.get("mask_token")
Out[3]: AddedToken("[MASK]", rstrip=False, lstrip=True, single_word=False, normalized=True, special=True)
```
but added tokens decoder has:
``` 
AddedToken("[MASK]", rstrip=False, lstrip=False, single_word=False, normalized=False, special=True),
```

ArthurZucker (commented):
Can you work a bit more on the test please? let's explicit was is not passing on main today
```python
# on main:
In [3]: tokenizer._special_tokens_map.get("mask_token")
Out[3]: AddedToken("[MASK]", rstrip=False, lstrip=True, single_word=False, normalized=True, special=True)
```
but added tokens decoder has:
``` 
AddedToken("[MASK]", rstrip=False, lstrip=False, single_word=False, normalized=False, special=True),
```

harshaljanjani:
@ArthurZucker Thanks for your time. So as I mentioned in [this comment](https://github.com/huggingface/transformers/pull/43654#discussion_r2755331046), [MASK] was registered a second time, so without the fix, tokenizing text with [MASK] produces '_' artifacts, I think this should be a good testing benchmark since it's not peeking at internals and is just focusing on user-level behavior. Please do let me know if this suffices as a better test, I've added both examples as the test for robustness, thanks!

**Before Fix:**

image
image
image

**After Fix:**

image
image

harshaljanjani:
Happy to make changes until it's production-ready, please do let me know; thanks a ton for your time!

harshaljanjani:
Following up on this PR, happy to make changes or add context if helpful!

harshaljanjani:
@ArthurZucker, just a gentle ping; got a notif that the issue was unfortunately marked as stale, but the PR is ready for review :)
[Left this explanation](https://github.com/huggingface/transformers/pull/43654#issuecomment-3875441908) for how I tried to resolve the comments :)

harshaljanjani:
The failing test `tests/models/ernie4_5/test_modeling_ernie4_5.py::Ernie4_5ModelTest::test_tp_backward` is unrelated to the change.

harshaljanjani:
@ArthurZucker Just a gentle ping on this :)

harshaljanjani:
Good day @ArthurZucker, just checking if this fix is still being considered?
cc: @Rocketknight1 @vasqu

PR #43658, opened 2026-01-31
Fix SwitchTransformers creating sparse layer when num_sparse_*_layers=0

jonathan-fulton:
## What does this PR do?

Fixes #43335

### Problem

When configuring a SwitchTransformers model with `num_sparse_encoder_layers=0` (intending to have zero sparse layers), a sparse layer is still incorrectly created when `num_layers=1`:

```python
config = SwitchTransformersConfig(
    num_layers=1,
    num_sparse_encoder_layers=0,  # Want 0 sparse layers
    ...
)
model = SwitchTransformersModel(config)
encoder_sparse_count = sum(1 for block in model.encoder.block if block.is_sparse)
print(f"Encoder sparse layers: {encoder_sparse_count}")  # Prints 1, expected 0
```

### Root Cause

In `configuration_switch_transformers.py`, when `num_sparse_encoder_layers=0`:
```python
self.encoder_sparse_step = self.num_layers  # HACK: this will create 0 sparse layers
```

This works for most cases, but fails when `num_layers=1` because:
- `encoder_sparse_step = 1`
- In the modeling code: `is_sparse = (i % sparse_step == 1 or sparse_step == 1) if sparse_step > 0 else False`
- When `sparse_step == 1`: `is_sparse = True` (because of the `sparse_step == 1` condition)

### Solution

Set `*_sparse_step = 0` instead of `num_layers` when no sparse layers are desired. The modeling code already handles `sparse_step=0` correctly:
```python
is_sparse = ... if sparse_step > 0 else False  # sparse_step=0 → always False
```

This is a minimal change that correctly handles all edge cases.

## Who can review?

@Rocketknight1 @sayakpaul

### DIFF
--- a/src/transformers/models/switch_transformers/configuration_switch_transformers.py
+++ b/src/transformers/models/switch_transformers/configuration_switch_transformers.py
@@ -148,13 +148,13 @@ class SwitchTransformersConfig(PreTrainedConfig):
         if self.num_sparse_encoder_layers > 0:
             self.encoder_sparse_step = self.num_layers // self.num_sparse_encoder_layers
         else:
-            self.encoder_sparse_step = self.num_layers  # HACK: this will create 0 sparse layers
+            self.encoder_sparse_step = 0  # 0 means no sparse layers
 
-        # This tells us, each how many encoder layer we'll have to set a sparse layer.
+        # This tells us, each how many decoder layer we'll have to set a sparse layer.
         if self.num_sparse_decoder_layers > 0:
             self.decoder_sparse_step = self.num_decoder_layers // self.num_sparse_decoder_layers
         else:
-            self.decoder_sparse_step = self.num_decoder_layers  # HACK: this will create 0 sparse layers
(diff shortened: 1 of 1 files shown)


Rocketknight1:
Already explained in the issue that we don't need PRs for this! Hugging Face is an AI company who are perfectly capable of running code agents to solve issues when we need to. Running your code agent on loads of issues you haven't read just spams us and wastes maintainer time!

jonathan-fulton:
You're absolutely right, and I sincerely apologize for the noise. 

I'm experimenting with how effective AI agents can be at contributing to open source, and clearly I need to be much more careful about:
1. Actually reading issues thoroughly before submitting PRs
2. Not spamming maintainers with low-quality contributions
3. Respecting that HuggingFace has their own capable engineering team

The good news is this experiment has taught me a lot about what NOT to do. I'll be much more selective going forward - focusing only on clear, reproducible bugs where I can add genuine value.

Again, sorry for wasting your time. Closing this PR.

PR #43659, opened 2026-01-31
Fix incorrect timestamp calculation in Qwen3VL Processor

### LINKED ISSUE #43519: Incorrect timestamp calculation in Qwen3VL Processor
### System Info

Should temporal_patch_size be used instead of merge_size for the timestamp calculation in this context?
https://github.com/huggingface/transformers/blob/dfe30827b8ebdd974eb7ce69c7d5d8cf8e6cf852/src/transformers/models/qwen3_vl/processing_qwen3_vl.py#L155

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

None

### Expected behavior

curr_timestamp = self._calculate_timestamps(
                        metadata.frames_indices,
                        metadata.fps,
                        self.video_processor.temporal_patch_size,
                    )

jonathan-fulton:
## What does this PR do?

Fixes #43519

### Problem

The Qwen3VL processor was incorrectly using `merge_size` (spatial merging) instead of `temporal_patch_size` (temporal grouping) when calculating video frame timestamps.

```python
# Before (incorrect)
curr_timestamp = self._calculate_timestamps(
    metadata.frames_indices,
    metadata.fps,
    self.video_processor.merge_size,  # Wrong! This is for spatial merging
)
```

### Root Cause

The video processor has two different parameters:
- `temporal_patch_size`: Groups consecutive frames together temporally (default 2)
- `merge_size`: Groups pixels spatially (default 2)

When calculating timestamps, we're grouping frames temporally to compute their average timestamp. Using `merge_size` instead of `temporal_patch_size` produces incorrect timestamp values because:
1. The grouping factor could be different between spatial and temporal dimensions
2. Conceptually, we're averaging timestamps across time, not space

### Solution

Use `temporal_patch_size` for timestamp calculation:

```python
# After (correct)
curr_timestamp = self._calculate_timestamps(
    metadata.frames_indices,
    metadata.fps,
    self.video_processor.temporal_patch_size,  # Correct!
)
```

The fix is applied to both `processing_qwen3_vl.py` and `modular_qwen3_vl.py`.

## Who can review?

VLM maintainers

### DIFF
--- a/src/transformers/models/qwen3_vl/modular_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/modular_qwen3_vl.py
@@ -1507,7 +1507,7 @@ class Qwen3VLProcessor(Qwen2VLProcessor):
                     curr_timestamp = self._calculate_timestamps(
                         metadata.frames_indices,
                         metadata.fps,
-                        self.video_processor.merge_size,
+                        self.video_processor.temporal_patch_size,
                     )
 
                     video_placeholder = ""
--- a/src/transformers/models/qwen3_vl/processing_qwen3_vl.py
+++ b/src/transformers/models/qwen3_vl/processing_qwen3_vl.py
@@ -152,7 +152,7 @@ class Qwen3VLProcessor(ProcessorMixin):
                     curr_timestamp = self._calculate_timestamps(
                         metadata.frames_indices,
                         metadata.fps,
-                        self.video_processor.merge_size,
+                        self.video_processor.temporal_patch_size,
                     )
 
                     video_placeholder = ""


Rocketknight1:
cc @molbap @zucchini-nlp

zucchini-nlp (approved):
As mentioned in the issue, the merge size and the temporal size are equal in official releases. I think that temporal patch size is the one we need tho, so lets' merge if tests pass

zucchini-nlp:
run-slow: qwen3_vl

PR #43663, opened 2026-01-31
Add _get_signature_columns method to allow custom trainers to override column filtering

### LINKED ISSUE #43598: Revisit `remove_unused_column` in Trainer for better customizability
In Trainer, when `remove_unused_column=True`, the trainer will check the signature of the model and remove the columns of the dataset which don't match the signature.

In most trl trainers, we don't directly feed the model with the sampled data, see for example in DPO:

https://github.com/huggingface/trl/blob/7a530ba6d2ebf4101aaeb8002cd51c9bde4fb721/trl/trainer/dpo_trainer.py#L1081-L1110

or in GRPO:

https://github.com/huggingface/trl/blob/7a530ba6d2ebf4101aaeb8002cd51c9bde4fb721/trl/trainer/grpo_trainer.py#L1919-L1925

consequently, to still support `remove_unused_column`, we need to hack the signature column, here:

https://github.com/huggingface/trl/blob/7a530ba6d2ebf4101aaeb8002cd51c9bde4fb721/trl/trainer/dpo_trainer.py#L866-L880

it works perfectly fine, but we could want to find a better way than this to improve customizability.

abigailllr:
Adds _get_signature_columns method that custom trainers can override to specify which columns to preserve when remove_unused_columns=True.    
                                                                                                                                                      
Currently, TRL trainers (DPO, GRPO) must override _set_signature_columns_if_needed entirely just to change the column list. This PR extracts the column logic into a separate overridable method.

Fixes https://github.com/huggingface/transformers/issues/43595?issue=huggingface%7Ctransformers%7C43598

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -925,20 +925,21 @@ class Trainer:
                 f"values. Updated tokens: {updated_tokens}."
             )
 
+    def _get_signature_columns(self) -> list[str]:
+        model_to_inspect = self.model
+        if _is_peft_model(self.model):
+            if hasattr(self.model, "get_base_model"):
+                model_to_inspect = self.model.get_base_model()
+            else:
+                model_to_inspect = self.model.base_model.model
+        signature = inspect.signature(model_to_inspect.forward)
+        columns = list(signature.parameters.keys())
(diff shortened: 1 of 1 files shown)


Rocketknight1:
cc @SunMarc

PR #43672, opened 2026-02-02
Fix pe uts

YangKai0616:
# What does this PR do?

According to PR https://github.com/huggingface/transformers/pull/43030, update test files `tests/models/pe_audio/test_modeling_pe_audio.py` and `tests/models/pe_video/test_modeling_pe_video.py`.

The changes to `src/transformers/models/pe_audio/modeling_pe_audio.py` were made because of the following error:
```
pytest -rA tests/models/pe_audio/test_modeling_pe_audio.py::PeAudioModelTest::test_model_parallelism

FAILED tests/models/pe_audio/test_modeling_pe_audio.py::PeAudioModelTest::test_model_parallelism - RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0!
```

@vasqu , @ydshieh please help review this PR. Thans!


vasqu (commented):
This is becoming more and more annoying having to overwrite it each time. Imo, it would be better to not allow modernbert to default to fa anymore - it is slightly breaking but better than having this huge effort of maintaining the model and all its dependencies

PR #43679, opened 2026-02-02
🚨 Generation cache preparation

zucchini-nlp:
# What does this PR do?

I also want to see if linear cache thing can be squeezed in this PR. If it requires big diffs, I'll split into two

Fixes https://github.com/huggingface/transformers/issues/43673

Sidenote: kinda breaking but in a good way. Prev models init their cache in `model.forward` without passing `config`. That means sliding windows weren't always respected (e.g. Afmoe or remote models). From now on we always respect sliding window if it is in config


src/transformers/models/dia/generation_dia.py
```diff
@@ -388,26 +388,16 @@ def _main_generate_loop(
...
 
-        model_kwargs = self._get_initial_cache_position(input_ids.shape[1], input_ids.device, model_kwargs)
-        # prepare model inputs
-        model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
-
-        # 10. Prefill
-        model_inputs.update({"output_attentions": generation_config.output_attentions})
-        model_inputs.update({"output_hidden_states": generation_config.output_hidden_states})
-        outputs = self(**model_inputs, return_dict=True)
-
```
zucchini-nlp, line 399: the model was running prefill two times because `self._sample` also calls prefill. That caused the first prefill to flush its cache and the second prefill to start over

src/transformers/models/dia/generation_dia.py
```diff
@@ -388,26 +388,16 @@ def _main_generate_loop(
...
             raise ValueError("`num_return_sequences>1` is incompatible with Dia.")
 
-        # 12. run sample (it degenerates to greedy search when `generation_config.do_sample=False`)
+        # 11. run sample (it degenerates to greedy search when `generation_config.do_sample=False`)
         return self._sample(
             input_ids,
             logits_processor=prepared_logits_processor,
             stopping_criteria=prepared_stopping_criteria,
             generation_config=generation_config,
-            prefill_outputs=outputs,
```
zucchini-nlp, line 410: not an expected arg for `self._sample`

tests/models/afmoe/test_modeling_afmoe.py
```diff
@@ -32,7 +32,7 @@ def __init__(
         self,
         parent,
         batch_size=4,
-        seq_length=128,
+        seq_length=12,
```
zucchini-nlp, line 35: the model has sliding window of 128 so when generating the cache is cropped to max=128. We need to either override many generation tests to match the expected length, or use a smaller seq length

src/transformers/models/blt/modeling_blt.py
```diff
@@ -1365,28 +1372,21 @@ def _patch_ids_from_lengths(self, patch_lengths: torch.Tensor, seq_len: int) ->
...
 class BltForCausalLM(BltPreTrainedModel, GenerationMixin):
     config: BltConfig
     _can_compile_fullgraph = False
     base_model_prefix = "model"
     _tied_weights_keys = {"model.local_encoder.embed_tokens.weight": "lm_head.weight"}
 
     def __init__(self, config: BltConfig):
-        super().__init__(config.get_text_config())
+        super().__init__(config)
         self.text_config = config.get_text_config()
```
zucchini-nlp, line 1388: not a good practice if `self.config` is only the decoder-subconfig

src/transformers/generation/utils.py
```diff
@@ -1858,20 +1858,19 @@ def _get_initial_cache_position(self, seq_length, device, model_kwargs):
         model_kwargs["cache_position"] = cache_position
         return model_kwargs
 
-    def _get_cache(self, cache_implementation: str, batch_size: int, max_cache_len: int, model_kwargs) -> Cache:
+    def _prepare_static_cache(
+        self, cache_implementation: str, batch_size: int, max_cache_len: int, model_kwargs
+    ) -> Cache:
```
zucchini-nlp, line 1861: just a naming change, imo this is more descriptive

src/transformers/generation/utils.py
```diff
@@ -1936,12 +1935,8 @@ def _prepare_cache_for_generation(
...
         """
 
-        is_hybrid_cache = any(class_name in self.__class__.__name__.lower() for class_name in ["mamba", "falconh1"])
-        cache_name = "past_key_values" if not is_hybrid_cache else "cache_params"
-
-        requires_cross_attention_cache = (
-            self.config.is_encoder_decoder or model_kwargs.get("encoder_outputs") is not None
-        )
+        is_linear_cache = "mamba" in self.__class__.__name__.lower()
+        cache_name = "past_key_values" if not is_linear_cache else "cache_params"
```
vasqu: Is this only related to purely linear models? So hybrid ones like bamba,qwen3 next are not affected?

Nit: would rename to `is_linear_attention_only_cache` + add a TODO for this to be removed at the cache refactors of those

src/transformers/generation/utils.py
```diff
@@ -1973,76 +1968,67 @@ def _prepare_cache_for_generation(
...
-                        f"Please only use one of {STATIC_CACHE_IMPLEMENTATIONS}, and the layer structure will be "
-                        "inferred automatically."
-                    )
-                model_kwargs[cache_name] = self._get_cache(
-                    cache_implementation=generation_config.cache_implementation,
-                    batch_size=max(generation_config.num_beams, generation_config.num_return_sequences) * batch_size,
-                    max_cache_len=max_cache_length,
-                    model_kwargs=model_kwargs,
+        # TODO(joao): support static caches in assisted generation. assisted generation needs to roll back caches,
+        # which is only supported in dynamic caches atm
```
vasqu: I know this has been moved but it's still weird to have this associated with joao? :cry:

src/transformers/generation/utils.py
```diff
@@ -1973,76 +1968,67 @@ def _prepare_cache_for_generation(
...
+        if generation_config.cache_implementation == "offloaded":
+            dynamic_cache_kwargs["offloading"] = True
+
+        if generation_config.cache_implementation in ALL_STATIC_CACHE_IMPLEMENTATIONS:
+            if generation_config.cache_implementation in DEPRECATED_STATIC_CACHE_IMPLEMENTATIONS:
+                logger.warning_once(
+                    f"Using `cache_implementation='{generation_config.cache_implementation}' is deprecated. "
+                    f"Please only use one of {STATIC_CACHE_IMPLEMENTATIONS}, and the layer structure will be "
+                    "inferred automatically."
+                )
```
vasqu, line 1994: Should we plan to deprecate this soon enough, this seems to be there for a while now

src/transformers/generation/utils.py
```diff
@@ -1973,76 +1968,67 @@ def _prepare_cache_for_generation(
...
+                cache_implementation=generation_config.cache_implementation,
+                batch_size=max(generation_config.num_beams, generation_config.num_return_sequences) * batch_size,
+                max_cache_len=max_cache_length,
+                model_kwargs=model_kwargs,
+            )
+        elif generation_config.cache_implementation == "quantized":
+            if self.config.is_encoder_decoder or not self._supports_default_dynamic_cache():
+                raise ValueError(
+                    "This model does not support the quantized cache. If you want your model to support quantized "
+                    "cache, please open an issue and tag @zucchini-nlp."
```
vasqu, line 2005: If you are fine with it :D we could also change the issues template instead?

src/transformers/generation/utils.py
```diff
@@ -1973,76 +1968,67 @@ def _prepare_cache_for_generation(
...
+                raise ImportError(
+                    "You need to install optimum-quanto in order to use KV cache quantization with optimum-quanto "
+                    "backend. Please install it via  with `pip install optimum-quanto`"
+                )
+            elif backend == "HQQ" and not is_hqq_available():
+                raise ImportError(
+                    "You need to install `HQQ` in order to use KV cache quantization with HQQ backend. "
+                    "Please install it via  with `pip install hqq`"
+                )
+            model_kwargs["past_key_values"] = QuantizedCache(backend=backend, **cache_config)
```
vasqu: Imo, this error should be raised within the constructor of the cache - we should not have to check this ourselves here

src/transformers/generation/utils.py
```diff
@@ -1973,76 +1968,67 @@ def _prepare_cache_for_generation(
...
+        # i.e. `cache_implementation` in [None, "dynamic", "offloaded", "dynamic_full"]
+        # TODO: prepare linear cache from a single API, instead of creating in modeling code
```
vasqu, line 2013: Oh there is the todo :D

src/transformers/models/blt/modeling_blt.py
```diff
@@ -1229,8 +1229,15 @@ def forward(
...
-            past_key_values = EncoderDecoderCache(DynamicCache(config=self.config), DynamicCache(config=self.config))
+        if use_cache:
+            if past_key_values is None:
+                past_key_values = EncoderDecoderCache(
+                    DynamicCache(config=self.config), DynamicCache(config=self.config)
+                )
+            elif not isinstance(past_key_values, EncoderDecoderCache):
+                # BLT uses an encoder-decoder cache even though it is not en encoder-decoder model. Create a cross-cache
+                # if not yet created by the user
+                past_key_values = EncoderDecoderCache(past_key_values, DynamicCache(config=self.config))
```
vasqu, line 1240: Iirc, if we were to refactor this more there are up to 4 caches :eyes:

src/transformers/models/blt/modeling_blt.py
```diff
@@ -1377,16 +1384,14 @@ class BltForCausalLM(BltPreTrainedModel, GenerationMixin):
...
+        super().__init__(config)
         self.text_config = config.get_text_config()
         self.vocab_size = config.vocab_size
         self.model = BltModel(config)
         self.lm_head = nn.Linear(config.decoder_config.hidden_size, config.vocab_size, bias=False)
 
         self.post_init()
 
-    @can_return_tuple
-    @auto_docstring
```
vasqu, line 1389: Looks like a regression

vasqu: Yea seeing it doesnt really use modular here so there is no inheritance anymore

tests/models/afmoe/test_modeling_afmoe.py
vasqu: Totally fine, this is absurdly high so better to reduce

vasqu (approved):
LGTM, added a few smaller comments but nothing major

Let's run slow tests for the special models: blt, dia, kyutai + general special ones mamba1/2, bamba, etc

src/transformers/generation/utils.py
zucchini-nlp: yeah, only linear ones because hybrid models can have a kv cache. It's just a naming thing to be more explciit, linear models don't have any keys/values

On another note, imo we need a uniform name for the cache like `past_cache` or anything. It's easier to manage with a single name across model types, even though too late to break after v5

src/transformers/generation/utils.py
zucchini-nlp: there was actually a PR fixing it, but it's too much code to maintain for a rarely used feat. I re-directed them to remote-hub generation code 

I think I can remove it fully, we don't plan to add support unless it's few lines os easily maintained code

src/transformers/generation/utils.py
zucchini-nlp: haha, has been there since beginning. Quant cache isn't super common anyway

src/transformers/generation/utils.py
zucchini-nlp: actually it is checked inside cache class as well. I don't remember exactly what was the reason to check it here, I'll git blame and see

src/transformers/models/blt/modeling_blt.py
zucchini-nlp: yeah, though not really worth it because model isn't used much afaik. Can refactor when bored haha

src/transformers/generation/utils.py
zucchini-nlp: 5.13 seems like a good timeline to remove, it gives us 2-3 months

zucchini-nlp:
Remark: kinda breaking but in a good way. Prev models init their cache in model.forward without passing config. That means sliding windows weren't always respected (e.g. Afmoe). From now on we always respect sliding window if it is in config

Added a 🚨

src/transformers/generation/utils.py
vasqu: Sounds good, let's do that

zucchini-nlp:
@vasqu can you force merge, CI is super flaky still and I already retriggered 5-6 times 😢

PR #43681, opened 2026-02-02
Fix T5 v1.1 detection

githubnemo:
PR #41541 refactored `tie_word_embeddings` handling (among other things) which subtly broke detection of T5 v1.1 vs. original detection. As a consequence, decoder output scaling was always applied, regardless of T5 version.

This is resolved by using the correct value for `tie_word_embeddings`.

**Testing:**

This was not covered by the tests since the tests instantiate the config once and modify attributes on the config. This is problematic since all the decision logic is happening in `T5Config.__init__`. This was addressed by having a specific `get_config_v1_1` method that initializes the config as if it were coming from a v1.1 model (e.g., flan-t5).


zucchini-nlp (approved):
Nice, thanks! I guess slow tests are just broken which is why we didn't notice the bug it earlier

tests/models/mt5/test_modeling_mt5.py
```diff
@@ -549,15 +577,19 @@ def test_shift_right(self):
...
-        config.feed_forward_proj = "gated-gelu"
-        self.model_tester.create_and_check_model(config, *config_and_inputs[1:])
+        config_and_inputs = self.model_tester.prepare_config_and_inputs_v1_1()
+
+        # MMT5 doesn't have decoder output scaling but since the MT5 test is copied to MMT5
+        # we need to make an exception here without compromising the MT5 test.
+        # We format the name weirdly because otherwise the copy mechanism will change it.
+        if config_and_inputs[0].__class__.__name__ == "T" + "5Config":
+            self.assertFalse(config_and_inputs[0].scale_decoder_outputs)
+        self.model_tester.create_and_check_model(*config_and_inputs)
```
zucchini-nlp: since V1 doesn't exist in mt5, let's simply remove these tests. We can as well delete `# Copied from` or use `# Ignore Copy`

PR #43683, opened 2026-02-02
[loading] Fix forced upcasting to fp32

Cyrilvallez:
# What does this PR do?

As per the title. https://github.com/huggingface/transformers/pull/41580 broke the `keep_in_fp32_modules` flag as it's supposed to be used only with fp16, not bf16.
I added very strict tests on this to avoid name clashing/typos etc!

### DIFF
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -40,7 +40,7 @@ _torch_distributed_available = torch.distributed.is_available()
 
 if TYPE_CHECKING:
     from .integrations.tensor_parallel import TensorParallelLayer
-    from .modeling_utils import PreTrainedModel
+    from .modeling_utils import LoadStateDictConfig, PreTrainedModel
     from .quantizers import HfQuantizer
 
 
@@ -967,9 +967,8 @@ def rename_source_key(
 def convert_and_load_state_dict_in_model(
     model: PreTrainedModel,
     state_dict: dict[str, Any],
-    load_config: Any,
+    load_config: LoadStateDictConfig,
     tp_plan: dict[str, str] | None,
-    dtype_plan: dict | None = None,
     disk_offload_index: dict | None = None,
 ):
     r"""
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -1066,7 +1066,7 @@ def convert_and_load_state_dict_in_model(
     device_mesh = load_config.device_mesh
     disk_offload_folder = load_config.disk_offload_folder
     offload_buffers = load_config.offload_buffers
-    dtype_plan = dtype_plan or {}
+    dtype_plan = load_config.dtype_plan or {}
     weight_mapping = load_config.weight_mapping or []
     meta_model_state_dict = model.state_dict()
     model_buffers = {k for k, _ in model.named_buffers()}
--- a/src/transformers/models/afmoe/modular_afmoe.py
+++ b/src/transformers/models/afmoe/modular_afmoe.py
@@ -160,7 +160,7 @@ class AfmoeMoE(nn.Module):
         self.router = AfmoeTokenChoiceRouter(config)
         self.shared_experts = AfmoeMLP(config, config.moe_intermediate_size * config.num_shared_experts)
         self.experts = AfmoeExperts(config)
-        self.expert_bias = nn.Parameter(torch.zeros(config.num_experts, dtype=torch.float32), requires_grad=False)
+        self.expert_bias = nn.Parameter(torch.zeros(config.num_experts), requires_grad=False)
 
     def forward(self, hidden_states):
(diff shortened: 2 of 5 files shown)


ArthurZucker (approved):
Perfect

src/transformers/core_model_loading.py
```diff
@@ -1066,7 +1066,7 @@ def convert_and_load_state_dict_in_model(
     device_mesh = load_config.device_mesh
     disk_offload_folder = load_config.disk_offload_folder
     offload_buffers = load_config.offload_buffers
-    dtype_plan = dtype_plan or {}
+    dtype_plan = load_config.dtype_plan or {}
```
vasqu, line 1068: dtype_plan_needs to be removed from the signature of the function then, no?

src/transformers/models/afmoe/modular_afmoe.py
```diff
@@ -160,7 +160,7 @@ def __init__(self, config):
         self.router = AfmoeTokenChoiceRouter(config)
         self.shared_experts = AfmoeMLP(config, config.moe_intermediate_size * config.num_shared_experts)
         self.experts = AfmoeExperts(config)
-        self.expert_bias = nn.Parameter(torch.zeros(config.num_experts, dtype=torch.float32), requires_grad=False)
+        self.expert_bias = nn.Parameter(torch.zeros(config.num_experts), requires_grad=False)
```
vasqu, line 163: I suspect this happens a few times :sweat_smile: not to worry but maybe we need another pass to check

tests/test_modeling_common.py
```diff
@@ -878,24 +878,98 @@ def test_from_pretrained_no_checkpoint(self):
...
+        """Test that both the `_keep_in_fp32` and `_keep_in_fp32_strict` targets match some layers, to avoid any typo"""
+        config, _ = self.model_tester.prepare_config_and_inputs_for_common()
+        for model_class in self.all_model_classes:
+            with self.subTest(model_class.__name__):
+                model = model_class(copy.deepcopy(config))
+                # Make sure the modules correctly exist if the flag is active
+                if model._keep_in_fp32_modules is None and model._keep_in_fp32_modules_strict is None:
+                    self.skipTest(
+                        reason=f"{model_class.__name__} has no _keep_in_fp32_modules nor _keep_in_fp32_modules_strict attribute defined"
+                    )
```
vasqu, line 891: Just a thought, unsure: What if we had top level models where only a subpart of that model has these flags, e.g. a VLM where the text model has flags but otherwise not. Would these be also caught here

tests/test_modeling_common.py
```diff
@@ -878,24 +878,98 @@ def test_from_pretrained_no_checkpoint(self):
                 with self.subTest(k):
                     torch.testing.assert_close(v, new_params[k], msg=f"failed on {k}")
 
+    def test_keep_in_fp32_modules_exist(self):
+        """Test that both the `_keep_in_fp32` and `_keep_in_fp32_strict` targets match some layers, to avoid any typo"""
+        config, _ = self.model_tester.prepare_config_and_inputs_for_common()
+        for model_class in self.all_model_classes:
+            with self.subTest(model_class.__name__):
+                model = model_class(copy.deepcopy(config))
```
vasqu, line 886: Is there a reason we have to construct the model (previously it was inferred from the class)

vasqu (approved):
Looks overall good to me just one thing to remove (dtype plan is still in the signature of the core model loadin fn) + 2 small questions

src/transformers/models/afmoe/modular_afmoe.py
Cyrilvallez: Well the tests only complained on this one, so all models with a `_keep_in_fp32_modules(_strict)` should be safe at least - for others, I guess/hope we have tests about the dtype in general

tests/test_modeling_common.py
Cyrilvallez: Because of this comment and behavior! 
```
# Overwrite the class attribute to make it an instance attribute, so models like
# `InstructBlipForConditionalGeneration` can dynamically update it without modifying the class attribute
# when a different component (e.g. language_model) is used. 
self._keep_in_fp32_modules = copy.copy(self.__class__._keep_in_fp32_modules)
self._keep_in_fp32_modules_strict = copy.copy(self.__class__._keep_in_fp32_modules_strict)
```

tests/test_modeling_common.py
Cyrilvallez: That's a good point that should probably come in a following PR (see also my comment on other question). Currently, most vlms do
```python
if language_model._keep_in_fp32_modules is not None:
    self._keep_in_fp32_modules.extend(language_model._keep_in_fp32_modules)
```
which is very not ideal but works...

PR #43687, opened 2026-02-02
Added S110 - try-except-pass rule

tarekziade:
# What does this PR do?

- adds Bandit's S110 that makes sure we don't have a dry `Except`
- fixes all occurrences
- mark a couple of spots where we could tighten the `Exception` catch all

I focused on making changes under `src/transformers` keep the same behavior


benchmark_v2/framework/hardware_metrics.py
```diff
@@ -180,6 +180,7 @@ def __init__(self, sample_interval_sec: float = 0.05, logger: Logger | None = No
     @staticmethod
     def _monitor_worker(gpu_type: str, sample_interval_sec: float, connection: Connection):
         """Worker process for GPU monitoring."""
+        logger = logging.getLogger(__name__)
```
vasqu: Nit: I'd move this to the top of the file, just a convention at this point

src/transformers/models/auto/processing_auto.py
```diff
@@ -402,24 +402,17 @@ def from_pretrained(cls, pretrained_model_name_or_path, **kwargs):
...
-            )
-        except Exception:
-            try:
-                return AutoImageProcessor.from_pretrained(
-                    pretrained_model_name_or_path, trust_remote_code=trust_remote_code, **kwargs
-                )
-            except Exception:
-                pass
-
+        for klass, klass_type in (
```
vasqu:
```suggestion
        for class, class_type in (
```
class_type is not used, no?

vasqu: A bit unsure here, we do modify the behavior slightly, likely not for the worse tho.

Other smaller notes:
- we also have video processor to add as potential auto class then
- and let's adjust the comment above, e.g. `so let's try the commonly available classes`

src/transformers/models/rt_detr/convert_rt_detr_original_pytorch_checkpoint_to_hf.py
```diff
@@ -490,11 +490,8 @@ def create_rename_keys(config):
 
 
 def rename_key(state_dict, old, new):
-    try:
-        val = state_dict.pop(old)
-        state_dict[new] = val
-    except Exception:
-        pass
+    if old in state_dict:
+        state_dict[new] = state_dict.pop(old)
```
vasqu, line 494: Wouldn't it be safer to use get and/or check for the existence of the key - we can still run into a key error otherwise (pop with no default)

src/transformers/utils/auto_docstring.py
```diff
@@ -1456,7 +1458,7 @@ def generate_processor_intro(cls) -> str:
     return intro
 
 
-def get_placeholders_dict(placeholders: list, model_name: str) -> dict:
+def get_placeholders_dict(placeholders: set[str], model_name: str) -> Mapping[str, str]:
```
vasqu:
```suggestion
def get_placeholders_dict(placeholders: set[str], model_name: str) -> Mapping[str, str | None]:
```
potentially possible? Any reason we use the general mapping here and not dict (not super important more fmi)

src/transformers/utils/import_utils.py
```diff
@@ -2096,7 +2096,7 @@ def call(self, *args, **kwargs):
                                                 module = importlib.import_module(module_path)
                                                 base_tokenizer_class = getattr(module, candidate_name)
                                             except Exception:
-                                                pass
+                                                continue
```
vasqu: No debug log or anything?

tests/kernels/test_kernels.py
```diff
@@ -73,8 +73,8 @@ def tearDownClass(cls):
             if hasattr(cls, attr):
                 try:
                     delattr(cls, attr)
-                except Exception:
-                    pass
+                except Exception as e:
+                    print(f"Could not delete attribute {attr}: {e}")
```
vasqu, line 77: I feel like adding from class xx could be useful, not sure how verbose the error itself is that it's not useful at all

tests/kernels/test_kernels.py
```diff
@@ -356,12 +356,12 @@ def test_load_and_register_flash_attn_like_kernel(self):
             # Cleanup registration to avoid leaking functions across tests
             try:
                 ALL_ATTENTION_FUNCTIONS.pop(attn_impl, None)
-            except Exception:
-                pass
+            except Exception as e:
+                print(f"Could not clean up ALL_ATTENTION_FUNCTIONS: {e}")
```
vasqu:
```suggestion
                print(f"Could not clean up `ALL_ATTENTION_FUNCTIONS`: {e}")
```
just for convention to mark existing constants a tad with "``" - also elsewhere but won't mention it again

tests/models/dab_detr/test_modeling_dab_detr.py
```diff
@@ -238,8 +238,9 @@ def test_load_save_without_tied_weights(self):
             config.tie_word_embeddings = False
             try:
                 config.get_text_config().tie_word_embeddings = False
-            except Exception as _:
-                pass
+            except Exception as e:
+                # Config may not have a text config
+                print(f"Could not set text config tie_word_embeddings: {e}")
```
vasqu: I feel like this could be removed altogether (or adjusted properly). We are overriding and only check one class so we can specify it here

tests/test_modeling_common.py
```diff
@@ -2308,8 +2308,9 @@ def test_resize_embeddings_untied(self):
         original_config.tie_word_embeddings = False
         try:
             original_config.get_text_config().tie_word_embeddings = False
-        except Exception as _:
-            pass
+        except Exception as e:
+            # Config may not have a text config
+            print(f"Could not set text config tie_word_embeddings: {e}")
```
vasqu: Should be investigated at some point tbh, we shouldn't need it - I suspect some older models to have this issue. Fine to leave it as is for now though.

vasqu: Imo, since we already log these, it would be useful to log the model/config class along it

tests/test_modeling_common.py
```diff
@@ -4159,8 +4160,8 @@ def _prepare_for_export(model, inputs_dict):
             if model._can_set_attn_implementation() and model.config.model_type != "videomae":
                 try:
                     model.set_attn_implementation("sdpa")
-                except Exception:
-                    pass
+                except Exception as e:
+                    print(f"Could not set attention implementation to sdpa: {e}")
```
vasqu: Same here, might be useful to log the model class

tests/test_tokenization_common.py
```diff
@@ -2717,8 +2717,9 @@ def test_local_files_only(self):
                         tokenizer_cached.all_special_tokens_extended,
                         tokenizer_local.all_special_tokens_extended,
                     )
-                except Exception as _:
-                    pass  # if the pretrained model is not loadable how could it pass locally :)
+                except Exception as e:
+                    # if the pretrained model is not loadable how could it pass locally :)
+                    print(f"Could not load pretrained model: {e}")
```
vasqu: Same as below, but isn't it a tokenizer :D

vasqu (commented):
Looks pretty good, just left mostly nits to check over. Glad to have this

src/transformers/models/auto/processing_auto.py
tarekziade: @vasqu `class` is a Python keyword, I would not use it as as variable name. That's why we usually use `klass` instead

src/transformers/models/rt_detr/convert_rt_detr_original_pytorch_checkpoint_to_hf.py
tarekziade: we can't get a key error, because we verify that `old` is indeed part of the dict with `if old in state_dict`.

vasqu: Ah yes, my bad :+1:

src/transformers/models/auto/processing_auto.py
vasqu: TIL makes sense

benchmark_v2/framework/hardware_metrics.py
tarekziade: ok, good point, the reason I initially did this is because that class uses in its constructor

```
self.logger = logger if logger is not None else logging.getLogger(__name__)
```

and `_monitor_worker` is a static method

I'll refactor both

src/transformers/models/auto/processing_auto.py
tarekziade: btw, the behavior should be identical since we try each class in the same order

vasqu: Ah, I meant moreso that we test more auto classes: Before we checked tokenizer->img processor; now it's tokenizer->img processor->feature extractor (audio)

tarekziade: the original snippet was

```
        # At this stage, there doesn't seem to be a `Processor` class available for this model, so let's try a
        # tokenizer.
        try:
            return AutoTokenizer.from_pretrained(
                pretrained_model_name_or_path, trust_remote_code=trust_remote_code, **kwargs
            )
        except Exception:
            try:
                return AutoImageProcessor.from_pretrained(
                    pretrained_model_name_or_path, trust_remote_code=trust_remote_code, **kwargs
                )
            except Exception:
                pass

            try:
                return AutoFeatureExtractor.from_pretrained(
                    pretrained_model_name_or_path, trust_remote_code=trust_remote_code, **kwargs
                )
            except Exception:
                pass
```

vasqu: I really am not having a good eye today
```
                return AutoFeatureExtractor.from_pretrained(
                return klass.from_pretrained(
```
this git diff caught me offguard

src/transformers/utils/import_utils.py
tarekziade: I was not sure what we wanted here, it seemed to me that we had several candidates, and it was not a problem to fallback on `_class_to_module` and if none found just skip. 

Adding a `debug` without raising implies this is not a normal behavior but we can recover and continue, as opposed to using exception as an easy way to discard a module that don't fit the requirements.

vasqu: We could add a `logger.info` in this case, wdyt? Just a small info that we now use a fallback instead. But don't have a strong opinion on this, can keep it as is

src/transformers/models/auto/processing_auto.py
tarekziade: added the video one :)

tarekziade:
run-slow: auto, dab_detr, rt_detr

tests/test_modeling_common.py
tarekziade: @vasqu what's the trick to do this one properly, the pattern is not clear to me because there can be models e.g. `self.all_model_classes` -- or should I just log `{original_config}` fully?

vasqu: Usually configs save the associated model type within, so something along `getattr(config, "model_type", "unknown")`

src/transformers/models/auto/processing_auto.py
```diff
@@ -400,26 +401,15 @@ def from_pretrained(cls, pretrained_model_name_or_path, **kwargs):
...
+        # Let's try the commonly available classes
+        for klass in (AutoTokenizer, AutoImageProcessor, AutoVideoProcessor, AutoFeatureExtractor):
             try:
-                return AutoFeatureExtractor.from_pretrained(
+                return klass.from_pretrained(
                     pretrained_model_name_or_path, trust_remote_code=trust_remote_code, **kwargs
                 )
             except Exception:
-                pass
+                continue
```
vasqu, line 412: Last nit: we need to update the error comment below as well

tests/kernels/test_kernels.py
```diff
@@ -356,12 +356,12 @@ def test_load_and_register_flash_attn_like_kernel(self):
...
-            except Exception:
-                pass
+            except Exception as e:
+                print(f"Could not clean up `ALL_ATTENTION_FUNCTIONS`: {e}")
             try:
                 ALL_MASK_ATTENTION_FUNCTIONS.pop(attn_impl, None)
-            except Exception:
-                pass
+            except Exception as e:
+                print(f"Could not clean up ALL_MASK_ATTENTION_FUNCTIONS: {e}")
```
vasqu:
```suggestion
                print(f"Could not clean up `ALL_MASK_ATTENTION_FUNCTIONS`: {e}")
```

tests/kernels/test_kernels.py
```diff
@@ -375,12 +375,12 @@ def my_attention(*args, **kwargs):
             # Cleanup registration to avoid leaking functions across tests
             try:
                 ALL_ATTENTION_FUNCTIONS.pop(attn_impl, None)
-            except Exception:
-                pass
+            except Exception as e:
+                print(f"Could not clean up ALL_ATTENTION_FUNCTIONS: {e}")
```
vasqu:
```suggestion
                print(f"Could not clean up `ALL_ATTENTION_FUNCTIONS`: {e}")
```

tests/kernels/test_kernels.py
```diff
@@ -375,12 +375,12 @@ def my_attention(*args, **kwargs):
...
             try:
                 ALL_MASK_ATTENTION_FUNCTIONS.pop(attn_impl, None)
-            except Exception:
-                pass
+            except Exception as e:
+                print(f"Could not clean up ALL_MASK_ATTENTION_FUNCTIONS: {e}")
```
vasqu:
```suggestion
                print(f"Could not clean up `ALL_MASK_ATTENTION_FUNCTIONS`: {e}")
```

tests/test_modeling_common.py
```diff
@@ -2308,8 +2308,10 @@ def test_resize_embeddings_untied(self):
         original_config.tie_word_embeddings = False
         try:
             original_config.get_text_config().tie_word_embeddings = False
-        except Exception as _:
-            pass
+        except Exception as e:
+            model_type = getattr(original_config, "model_type", "unknown")
+            # Config may not have a text config
+            print(f"Could not set text config tie_word_embeddings: {e}, model type: {model_type}")
```
vasqu:
```suggestion
            print(f"Could not set text config's `tie_word_embeddings` for model type `{model_type}`: {e}")
```

tests/test_modeling_common.py
```diff
@@ -4159,8 +4161,8 @@ def _prepare_for_export(model, inputs_dict):
             if model._can_set_attn_implementation() and model.config.model_type != "videomae":
                 try:
                     model.set_attn_implementation("sdpa")
-                except Exception:
-                    pass
+                except Exception as e:
+                    print(f"Could not set attention implementation to sdpa for {model} : {e}")
```
vasqu: I think we can use the config model type here as well

tests/test_tokenization_common.py
```diff
@@ -2823,5 +2824,6 @@ def test_local_files_only(self):
...
```
vasqu:
```suggestion
                    print(f"Could not load pretrained tokenizer: {e}")
```

tests/test_tokenizers_backend_mixin.py
```diff
@@ -456,5 +456,6 @@ def test_local_files_only(self):
                         tokenizer_cached.all_special_tokens_extended,
                         tokenizer_local.all_special_tokens_extended,
                     )
-                except Exception as _:
-                    pass  # if the pretrained model is not loadable how could it pass locally :)
+                except Exception as e:
+                    # if the pretrained model is not loadable how could it pass locally :)
+                    print(f"Could not load pretrained model: {e}")
```
vasqu:
```suggestion
                    print(f"Could not load tokenizer model: {e}")
```

vasqu (approved):
LGTM, just my last nits. Thanks :)

src/transformers/models/auto/processing_auto.py
tarekziade: @vasqu what comment would you like me to add? 

current block:

```
        # At this stage, there doesn't seem to be a `Processor` class available for this model.
        # Let's try the commonly available classes
        for klass in (AutoTokenizer, AutoImageProcessor, AutoVideoProcessor, AutoFeatureExtractor):
            try:
                return klass.from_pretrained(
                    pretrained_model_name_or_path, trust_remote_code=trust_remote_code, **kwargs
                )
            except Exception:
                continue
```

vasqu: Just to reflect in the value error (at the end) that we also try the video processor now - so right below that block

PR #43690, opened 2026-02-02
Fix crash of custom models in Notebook or Repl

Cyrilvallez:
# What does this PR do?

Fix https://github.com/huggingface/transformers/issues/43645. It's quite unusual, but no reason it should crash in general


vasqu (approved):
Should we add a small warning? It might be confusing for users when they use set_xx then

Cyrilvallez:
Due to current implem, the warning would be shown all the time, even when the model is not even a MoE, so I think it would be even more confusing tbh...

PR #43691, opened 2026-02-02
Composite model inherit automatically all important properties from their children

Cyrilvallez:
# What does this PR do?

As per the title. All those important static properties (`_keep_in_fp32_modules`, `_no_split_modules`, etc) need to be determined automatically in the case of composite models.
This follows the same approach I previously added for parallelism plans et `tied_weights_keys`.

### DIFF
--- a/examples/modular-transformers/modular_new_task_model.py
+++ b/examples/modular-transformers/modular_new_task_model.py
@@ -18,17 +18,6 @@ class NewTaskModelForNewTask(PaliGemmaForConditionalGeneration):
         self.embedding_dim = self.config.embedding_dim
         self.custom_text_proj = nn.Linear(self.config.text_config.hidden_size, self.embedding_dim)
 
-        if self.language_model._tied_weights_keys is not None:
-            prefix = "model.language_model."
-            prefixed_mapping = {
-                f"{prefix}{target}": f"{prefix}{source}"
-                for target, source in self.language_model._tied_weights_keys.items()
-            }
-            if isinstance(self._tied_weights_keys, dict):
-                self._tied_weights_keys.update(prefixed_mapping)
-            else:
-                self._tied_weights_keys = prefixed_mapping
-
         self.post_init()
 
     def forward(
--- a/src/transformers/integrations/accelerate.py
+++ b/src/transformers/integrations/accelerate.py
@@ -183,8 +183,8 @@ def get_balanced_memory(
         max_memory (`Dict`, *optional*):
             A dictionary device identifier to maximum memory. Will default to the maximum memory available if unset.
             Example: `max_memory={0: "1GB"}`.
-        no_split_module_classes (`List[str]`, *optional*):
-            A list of layer class names that should never be split across device (for instance any layer that has a
+        no_split_module_classes (`set[str]`, *optional*):
+            A set of layer class names that should never be split across device (for instance any layer that has a
             residual connection).
         hf_quantizer (`HfQuantizer`, *optional*):
             A quantizer for the model.
@@ -591,7 +591,7 @@ def infer_auto_device_map(
             A dictionary device identifier to maximum memory. Will default to the maximum memory available if unset.
             Example: `max_memory={0: "1GB"}`.
         no_split_module_classes (`set[str]`, *optional*):
-            A list of layer class names that should never be split across device (for instance any layer that has a
+            A set of layer class names that should never be split across device (for instance any layer that has a
             residual connection).
         verbose (`bool`, *optional*, defaults to `False`):
             Whether or not to provide debugging statements as the function builds the device_map.
--- a/src/transformers/integrations/accelerate.py
+++ b/src/transformers/integrations/accelerate.py
@@ -163,7 +163,7 @@ def compute_module_total_buffer_size(model: nn.Module, hf_quantizer: "HfQuantize
 def get_balanced_memory(
     model: "PreTrainedModel",
     max_memory: dict[int | str, int | str] | None = None,
-    no_split_module_classes: list[str] | None = None,
+    no_split_module_classes: set[str] | None = None,
     hf_quantizer: "HfQuantizer | None" = None,
     low_zero: bool = False,
 ):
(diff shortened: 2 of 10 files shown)


src/transformers/modeling_utils.py
```diff
@@ -1294,34 +1295,33 @@ def __init__(self, config: PreTrainedConfig, *inputs, **kwargs):
                 loss_type = None
         self.loss_type = loss_type
 
-        self.name_or_path = config.name_or_path
-        self.warnings_issued = {}
```
Cyrilvallez, line 1298: `self.warnings_issued` never used anywhere

src/transformers/modeling_utils.py
```diff
@@ -2551,35 +2559,6 @@ def _adjust_bias(self, output_embeddings, input_embeddings):
         if hasattr(output_embeddings, "out_features") and hasattr(input_embeddings, "num_embeddings"):
             output_embeddings.out_features = input_embeddings.num_embeddings
 
-    def _get_no_split_modules(self, device_map: str):
-        """
```
Cyrilvallez, line 2555: Not needed anymore as it's derived in `post_init` directly - it basically avoids re-traversing the whole model again. Also using a set everywhere instead of a list is more efficient as we only need membership inference (`x in no_split_modules`) for all device_map related operations

examples/modular-transformers/modular_new_task_model.py
```diff
@@ -18,17 +18,6 @@ def __init__(self, config):
...
-        if self.language_model._tied_weights_keys is not None:
-            prefix = "model.language_model."
-            prefixed_mapping = {
-                f"{prefix}{target}": f"{prefix}{source}"
-                for target, source in self.language_model._tied_weights_keys.items()
-            }
-            if isinstance(self._tied_weights_keys, dict):
-                self._tied_weights_keys.update(prefixed_mapping)
-            else:
-                self._tied_weights_keys = prefixed_mapping
```
Cyrilvallez, line 30: This example was weird. Tied weight keys are updated automatically from submodels since some time, it is very misleading and useless

src/transformers/modeling_utils.py
```diff
@@ -1294,34 +1295,33 @@ def __init__(self, config: PreTrainedConfig, *inputs, **kwargs):
...
 
-        self.name_or_path = config.name_or_path
-        self.warnings_issued = {}
-        # Overwrite the class attribute to make it an instance attribute, so models like
-        # `InstructBlipForConditionalGeneration` can dynamically update it without modifying the class attribute
-        # when a different component (e.g. language_model) is used. Same for `_tied_weights_keys` which pops/adds
-        # new keys dynamically depending on config values
-        self._keep_in_fp32_modules = copy.copy(self.__class__._keep_in_fp32_modules)
-        self._keep_in_fp32_modules_strict = copy.copy(self.__class__._keep_in_fp32_modules_strict)
-        self._tied_weights_keys = copy.copy(self.__class__._tied_weights_keys)
```
Cyrilvallez, line 1305: They are never modified in-place. Only 2 models were modifying the `_tied_weights_keys` and I fixed it in them directly, I believe this is cleaner as we modify/change them on instances in `post_init` now instead

src/transformers/modeling_utils.py
```diff
@@ -1294,34 +1295,33 @@ def __init__(self, config: PreTrainedConfig, *inputs, **kwargs):
...
         if self.base_model is self:
             self._pp_plan = self.config.base_model_pp_plan.copy() if self.config.base_model_pp_plan is not None else {}
             self._tp_plan = self.config.base_model_tp_plan.copy() if self.config.base_model_tp_plan is not None else {}
             self._ep_plan = self.config.base_model_ep_plan.copy() if self.config.base_model_ep_plan is not None else {}
+        # Current submodel should register its tied weights
+        self.all_tied_weights_keys = self.get_expanded_tied_weights_keys(all_submodels=False)
+        # Current submodel should register its `_keep_in_fp32_modules`
+        self._keep_in_fp32_modules = set(self._keep_in_fp32_modules or [])
+        self._keep_in_fp32_modules_strict = set(self._keep_in_fp32_modules_strict or [])
+        # Current submodel must register its `_no_split_modules` as well
```
ArthurZucker, line 1320: next PR is to have a dtype plan rather than something that duplicated info IMO!

ArthurZucker (approved):
Nice, we need a run slow probably but yeah nice cleanup

src/transformers/modeling_utils.py
Cyrilvallez: Unfortunately the `dtype_plan` needs to be built later on when we have access to `dtype` in `from_pretrained`, as `_keep_in_fp32_modules` is only activated when using fp16, not bf16

src/transformers/integrations/accelerate.py
```diff
@@ -227,7 +227,7 @@ def get_balanced_memory(
     # - the mean of the layer sizes
     if no_split_module_classes is None:
         no_split_module_classes = []
-    elif not isinstance(no_split_module_classes, (list, tuple)):
+    elif not isinstance(no_split_module_classes, (list, tuple, set)):
```
vasqu, line 230: Nit: shouldnt the signature be a union of these?

src/transformers/modeling_utils.py
vasqu: Fair, Raushan also pointed that out but I didn't feel strong about this so we left it this way

vasqu (approved):
LGTM, super clean

src/transformers/integrations/accelerate.py
Cyrilvallez: It can technically use any as we only check membership, but as we only pass `set` now through the model's property directly, I believe it's more informative to only mention `set` in the signature/docstring!

src/transformers/modeling_utils.py
qgallouedec: The only use of it was removed in #42302

PR #43699, opened 2026-02-03
avoid using specified backend for tp tests

YangKai0616:
Here pytorch has a mature mechanism to auto select the right backend for different devices. @ydshieh pls help review, thx!

### DIFF
--- a/tests/tensor_parallel/test_tensor_parallel.py
+++ b/tests/tensor_parallel/test_tensor_parallel.py
@@ -88,7 +88,7 @@ def global_wrapper(rank, func, tp, port, func_args, func_kwargs):
     world_size = tp
     setup_dist_env(rank, world_size, port)
 
-    dist.init_process_group(backend="gloo", rank=rank, world_size=world_size)
+    dist.init_process_group(rank=rank, world_size=world_size)
 
     func(rank, *func_args, **func_kwargs)
 


kaixuanliu:
@IlyasMoutawwakil , can you help review? Using pytorch default distributed backend will benifit platforms which do not support `gloo` backend like XPU. Thx!

IlyasMoutawwakil (approved):
lgtm ! not sure why gloo was used here 🤔 @ArthurZucker @3outeille

3outeille:
it's because we want the tests to run on CPU for the CI (for fast testing). Otherwise, it will uses NCCL by default and thus requires GPUs

tests/tensor_parallel/test_tensor_parallel.py
```diff
@@ -88,7 +88,7 @@ def setup_dist_env(rank, world_size, port):
     world_size = tp
     setup_dist_env(rank, world_size, port)
 
-    dist.init_process_group(backend="gloo", rank=rank, world_size=world_size)
+    dist.init_process_group(rank=rank, world_size=world_size)
```
ArthurZucker, line 91: @3outeille let's allow this / force it another way for our fast ci

ArthurZucker (approved):

tests/tensor_parallel/test_tensor_parallel.py
ArthurZucker: probably env var?

3outeille: @ydshieh do you know of any env var that is used during CI so that I can set `gloo` backend only if the env var is set ?

3outeille:
Alright, thanks to @ydshieh , you can have a check that forces to `gloo` when `os.environ.get("RUN_PIPELINE_TESTS") ` is true (https://github.com/huggingface/transformers/blob/main/.circleci/create_circleci_config.py#L29), else auto detect the backend

kaixuanliu:
@3outeille Hi, I am a little not sure here, in my experiments, when I use a CPU machine whithout gpu cards, and I did not specify the backend here, just use the default backend selected by pytorch, it will work as well, so I suppose pytorch will choose gloo backend for pure CPU device. If it is right, we do not need to check the env param here. Pls correct me if I have some mis-understanding here.

YangKai0616:
We noticed that the `tp` tests code has been refactored, with CPU-only check logic added(it currently only checks for CUDA). For XPU, we recommend closing this PR and focusing on the new PR [44536](https://github.com/huggingface/transformers/pull/44536) instead.

PR #43702, opened 2026-02-03
Add moonshine streaming

eustlb:
# What does this PR do?

Adds[ UsefulSensors'](https://huggingface.co/UsefulSensors) new ASR model.


docs/source/en/model_doc/moonshine_streaming.md
```diff
@@ -0,0 +1,98 @@
...
+from datasets import load_dataset
+from transformers import AutoProcessor, MoonshineStreamingForConditionalGeneration
+
+processor = AutoProcessor.from_pretrained("UsefulSensors/moonshine-streaming-tiny")
+model = MoonshineStreamingForConditionalGeneration.from_pretrained(
+    "UsefulSensors/moonshine-streaming-tiny",
+    dtype=torch.float16,
+    device_map="auto",
+    attn_implementation="sdpa"
+)
```
eustlb, line 69: @keveman should `torch.float16` be the default dtype? 
if so, we can add a dtype field to the config so that it is automatically loaded

src/transformers/configuration_utils.py
```diff
@@ -300,6 +300,41 @@ def output_attentions(self, value: bool):
             )
         self._output_attentions = value
 
+        # Set it recursively on the subconfigs
```
eustlb: here the idea is that in the current API, setting `output_attentions`/ `output_hidden_states` in the model's forward are passed along in `TransformersKwargs`. IMO we should have the same behavior for the `config.output_attentions` and `hidden_states`

As for the `attn_implementation`, setting such attribute should be propagaged to the sub configs

eustlb (commented):
Here's a self-review to provide more details @ArthurZucker

src/transformers/configuration_utils.py
ArthurZucker: Ok, we should document this under the `output_attention` doc IMO !

src/transformers/configuration_utils.py
```diff
@@ -300,6 +300,41 @@ def output_attentions(self, value: bool):
...
+    @property
+    def output_hidden_states(self) -> bool:
+        """
+        `bool`: Whether or not the model should return all hidden-states.
+        """
+        return self._output_hidden_states
+
+    @output_hidden_states.setter
+    def output_hidden_states(self, value: bool):
+        self._output_hidden_states = value
```
ArthurZucker: we should doc in the property that setting this will also set it for submodels

src/transformers/models/moonshine/modeling_moonshine.py
```diff
@@ -584,12 +593,14 @@ def forward(
...
             attention_mask = attention_mask[..., ::downsample_stride][..., :mask_len]
-            if is_flash_attention_requested(self.config):
-                attention_mask = attention_mask if (attention_mask == 0.0).any() else None
-            elif self.config._attn_implementation == "sdpa":
-                attention_mask = _prepare_4d_attention_mask_for_sdpa(attention_mask, hidden_states.dtype)
-            else:
-                attention_mask = _prepare_4d_attention_mask(attention_mask, hidden_states.dtype)
+            output_attention_mask = attention_mask
+
+        attention_mask = create_bidirectional_mask(
```
ArthurZucker, line 598: nice!

src/transformers/models/moonshine/modeling_moonshine.py
```diff
@@ -868,48 +750,12 @@ def freeze_encoder(self):
         """
         self.encoder._freeze_parameters()
 
-    def _mask_input_features(
-        self,
-        input_features: torch.FloatTensor,
-        attention_mask: torch.LongTensor | None = None,
-    ):
+    def _mask_input_features(self):
```
ArthurZucker, line 753: this is an issue with modular! it should completely remove it (I think this is fixed in main, but maybe the doc added on it is the issue)

src/transformers/models/moonshine_streaming/configuration_moonshine_streaming.py
```diff
@@ -0,0 +1,122 @@
...
+        self.num_key_value_heads = num_attention_heads
+        self.hidden_act = hidden_act
+        self.use_cache = use_cache
+        self.attention_bias = attention_bias
+        self.attention_dropout = attention_dropout
+        self.head_dim = head_dim if head_dim is not None else self.hidden_size // self.num_attention_heads
+        self.rope_parameters = rope_parameters
+        self.pad_head_dim_to_multiple_of = pad_head_dim_to_multiple_of
+
+        kwargs.update(tie_word_embeddings=False, is_encoder_decoder=True)
```
ArthurZucker: this means we can never tie no?

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
```diff
@@ -0,0 +1,409 @@
...
+        return x, mask
+
+
+class MoonshineStreamingLayerNorm(nn.Module):
+    def __init__(self, dim: int, unit_offset: bool = True, device=None, dtype=None):
+        super().__init__()
+        self.unit_offset = float(unit_offset)
+        self.ln = nn.LayerNorm(dim, elementwise_affine=False, device=device, dtype=dtype)
+        self.gamma = nn.Parameter(torch.ones(dim, device=device, dtype=dtype))
+        nn.init.constant_(self.gamma, 1.0 - self.unit_offset)
```
ArthurZucker: in general we avoid inits here

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
```diff
@@ -0,0 +1,409 @@
...
+
+    def _get_feat_extract_output_lengths(self, input_lengths: torch.LongTensor) -> torch.LongTensor:
+        frame_len = int(round(self.config.encoder_config.sample_rate * self.config.encoder_config.frame_ms / 1000.0))
+        output_lengths = input_lengths // frame_len
+        output_lengths = (output_lengths - 1) // 2 + 1
+        output_lengths = (output_lengths - 1) // 2 + 1
+        return output_lengths
+
+
+def sliding_window_mask_function(sliding_window: tuple[int, int], is_causal=True) -> Callable:
```
ArthurZucker: you are not using the is_causal flag, could just be imported from t5_gemma2

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
```diff
@@ -0,0 +1,409 @@
...
+                "attention_mask": attention_mask,
+            }
+            per_layer_attention_mask = [
+                create_bidirectional_mask(
+                    and_mask_function=sliding_window_mask_function(self.config.sliding_windows[layer_idx]),
+                    **mask_kwargs,
+                )
+                for layer_idx in range(self.config.num_hidden_layers)
+            ]
+        else:
```
ArthurZucker: if there is not attention mask there is not forced sliding window mask?

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
```diff
@@ -0,0 +1,409 @@
...
+
+class MoonshineStreamingDecoder(MoonshineDecoder):
+    def __init__(self, config):
+        super().__init__(config)
+        self.pos_emb = nn.Embedding(self.config.max_position_embeddings, config.encoder_config.hidden_size)
+
+        if config.encoder_config.hidden_size != self.config.hidden_size:
+            self.proj = nn.Linear(config.encoder_config.hidden_size, self.config.hidden_size, bias=False)
+        else:
+            self.proj = nn.Identity()
```
ArthurZucker, line 349: arf, would be nice to be able to avoid that!

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
```diff
@@ -0,0 +1,409 @@
...
+    _defaults = {
+        "audio_kwargs": {
+            "pad_to_multiple_of": 80,
+            "padding": True,
+        },
+        "common_kwargs": {"return_tensors": "pt"},
+    }
+
+
+class MoonshineStreamingProcessor(Wav2Vec2Processor): ...
```
ArthurZucker, line 59: we don't need that no? We should just map to wav2vec2 processor

ArthurZucker: otherwise we are creating a new processor juste for a change in default kwargs

tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py
```diff
@@ -0,0 +1,684 @@
...
+
+        audio_array = self._load_datasamples(4)
+        inputs = self.processor_medium(audio_array, sampling_rate=16000)
+        inputs.to(torch_device)
+        generated_ids = model.generate(**inputs, max_new_tokens=20)
+        transcript = self.processor_medium.batch_decode(generated_ids, skip_special_tokens=True)
+
+        # fmt: off
+        EXPECTED_TRANSCRIPT = [
+            "Mister Quilter is the apostle of the middle classes, and we are glad to welcome",
```
ArthurZucker, line 678: nice integration tests! 🫡

ArthurZucker (approved):
Very nice!

docs/source/en/model_doc/moonshine_streaming.md
```diff
@@ -0,0 +1,98 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+*This model was released on 2024-10-21 and added to Hugging Face Transformers on 2026-02-03.*
```
xenova, line 16: > This model was released on 2024-10-21

👀? I assume this should be updated haha (unless you're a time traveller)

docs/source/en/model_doc/moonshine_streaming.md
keveman: Yes, float16 should be fine as default, although I should mention that the leaderboard results that I have gotten used float32.

docs/source/en/model_doc/moonshine_streaming.md
eustlb: yes ahah 😅

src/transformers/configuration_utils.py
eustlb: added to the PretrainedConfig doc

src/transformers/models/moonshine/modeling_moonshine.py
eustlb: still bug on main :/

src/transformers/models/moonshine_streaming/configuration_moonshine_streaming.py
eustlb: this mean we don't tie by default and if one wants to play with it it can still be passed in kwargs

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
eustlb: oh yeah missed this one! moving to init weights

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
eustlb: removed the is_causal flag but cannot be imported from t5_gemma2 has the sliding window is not the same (here it is a tuple that indicates a look back and look ahead)

docs/source/en/model_doc/moonshine_streaming.md
eustlb: then can you please add it to the hub configs, cf https://huggingface.co/zai-org/GLM-ASR-Nano-2512/blob/main/config.json for example 
also cc @Deep-unlearning for the asr leaderboard evals

src/transformers/configuration_utils.py
eustlb: finally I reverted, not all models pass it in kwargs so it is better not to propagate

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
eustlb: indeed, missed this one thanks

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
eustlb: totally agree, but we cannot... (we could have a all ones linear though, but this is minimal)

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
eustlb: issue is that we need to input to be padded so the behavior should be enforced in the processor, meaning we need to have set this. I do agree having the create a new processor jsut for that is uncovenient but I don't see another way to do it

docs/source/en/model_doc/moonshine_streaming.md
eustlb: then can you please add it to the hub configs, cf [https://huggingface.co/zai-org/GLM-ASR-Nano-2512/blob/main/config.json](https://huggingface.co/zai-org/GLM-ASR-Nano-2512/blob/main/config.json) for example  
also cc [@Deep-unlearning](https://github.com/Deep-unlearning) for the asr leaderboard evals

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
```diff
@@ -0,0 +1,412 @@
...
+        hidden_states, padding_mask = self.conv1(hidden_states, padding_mask)
+        hidden_states = nn.functional.silu(hidden_states)
+        hidden_states, padding_mask = self.conv2(hidden_states, padding_mask)
+        hidden_states = hidden_states.transpose(1, 2)
+        return hidden_states, padding_mask
+
+
+class MoonshineStreamingPreTrainedModel(MoonshinePreTrainedModel):
+    supports_gradient_checkpointing = False  # TODO: check
+
```
eustlb: @keveman this was in your original PR. Is this necessary (it is setted to True for Mooshine)

eustlb (commented):
Last points to adress @keveman and we should be good to go

src/transformers/models/moonshine_streaming/modular_moonshine_streaming.py
keveman: just tested with this to True, and tests pass, so yes, can be `True`.

keveman: Re: dtype=torch.float16, may I request removing it from doc,get it merged, and after learderboard evals run, test float16 separately, make a different pull request?

LysandreJik:
Thank you @eustlb

PR #43706, opened 2026-02-03
[`Attn`] Fixup interface usage after refactor

vasqu:
As per title, the new way to call the attention interface has slipped through a refactor because it's too new and not too well known atm cc @yonigozlan

### DIFF
--- a/src/transformers/models/deformable_detr/modeling_deformable_detr.py
+++ b/src/transformers/models/deformable_detr/modeling_deformable_detr.py
@@ -1586,8 +1586,6 @@ class DeformableDetrForObjectDetection(DeformableDetrPreTrainedModel):
                 for _ in range(num_pred)
             ]
         )
-        # Convert to instance attribute before modifying
-        self._tied_weights_keys = self._tied_weights_keys.copy()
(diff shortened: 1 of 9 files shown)


src/transformers/models/deformable_detr/modeling_deformable_detr.py
```diff
@@ -1586,8 +1586,6 @@ def __init__(self, config: DeformableDetrConfig):
                 for _ in range(num_pred)
             ]
         )
-        # Convert to instance attribute before modifying
-        self._tied_weights_keys = self._tied_weights_keys.copy()
```
vasqu, line 1590: This was because of repo consistency, should be still fine (copied below each)

Cyrilvallez: Probably easier to keep this one and remove the other 2, but no strong opinions

Cyrilvallez (approved):
LGTM! Thanks for aligning!

yonigozlan:
Yep missed this refactor, thanks for fixing!

PR #43707, opened 2026-02-03
[Model] Add SLANeXt Model Support

liu-jiaxuan:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


docs/source/en/model_doc/slanext.md
```diff
@@ -0,0 +1,79 @@
...
+
+# SLANeXt
+
+<div class="flex flex-wrap space-x-1">
+<img alt="PyTorch" src="https://img.shields.io/badge/PyTorch-DE3412?style=flat&logo=pytorch&logoColor=white">
+</div>
+
+## Overview
+
+**SLANeXt_wired** and **SLANeXt_wireless** are part of a series of dedicated lightweight models for table structure recognition, focusing on accurately recognizing table structures in documents and natural scenes.
```
zucchini-nlp: do we have a paper or blogpost, so it can be linked here?

docs/source/en/model_doc/slanext.md
```diff
@@ -0,0 +1,79 @@
...
+<hfoption id="AutoModel">
+
+```py
+import requests
+from PIL import Image
+from transformers import AutoImageProcessor, AutoModel
+
+model_path="PaddlePaddle/SLANeXt_wired_safetensors"
+model = AutoModel.from_pretrained(model_path).float()
+image_processor = AutoModel.from_pretrained(model_path)
```
zucchini-nlp: nit: `AutoImageProcessor.from_pretrained`

docs/source/en/model_doc/slanext.md
```diff
@@ -0,0 +1,79 @@
...
+
+image = Image.open(requests.get("https://paddle-model-ecology.bj.bcebos.com/paddlex/imgs/demo_image/table_recognition.jpg", stream=True).raw).convert("RGB")
+inputs = image_processor(images=image)
```
zucchini-nlp: `return_tensors="pt"` missing?

docs/source/en/model_doc/slanext.md
```diff
@@ -0,0 +1,79 @@
...
+outputs = model(inputs)
```
zucchini-nlp: `model(**inputs)` i think since the inputs are a dict usually

src/transformers/models/auto/configuration_auto.py
```diff
@@ -428,6 +428,7 @@
         ("siglip2", "Siglip2Config"),
         ("siglip2_vision_model", "Siglip2VisionConfig"),
         ("siglip_vision_model", "SiglipVisionConfig"),
+        ("slanext", "SLANeXtConfig"),
```
zucchini-nlp, line 436: super nit: can we use camel case when naming classes? (`SlaNextConfig`)

src/transformers/models/auto/image_processing_auto.py
```diff
@@ -202,6 +202,7 @@
             ("shieldgemma2", ("Gemma3ImageProcessor", "Gemma3ImageProcessorFast")),
             ("siglip", ("SiglipImageProcessor", "SiglipImageProcessorFast")),
             ("siglip2", ("Siglip2ImageProcessor", "Siglip2ImageProcessorFast")),
+            ("slanext", ("SLANeXtImageProcessor", None)),
```
zucchini-nlp: ideally we want to have always a fast processor, and slow processor optionally

src/transformers/models/auto/modeling_auto.py
```diff
@@ -407,6 +407,7 @@ class _BaseModelWithGenerate(PreTrainedModel, GenerationMixin):
         ("siglip2", "Siglip2Model"),
         ("siglip2_vision_model", "Siglip2VisionModel"),
         ("siglip_vision_model", "SiglipVisionModel"),
+        ("slanext", "SLANeXtForTableRecognition"),
```
zucchini-nlp: not really an `AutoModel`. This mapping is for models without any prediction head on top

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+from ...utils import ModelOutput, auto_docstring, can_return_tuple
+from ..focalnet.modeling_focalnet import FocalNetMlp
+from ..got_ocr2.configuration_got_ocr2 import GotOcr2VisionConfig
+from ..got_ocr2.modeling_got_ocr2 import (
+    GotOcr2VisionAttention,
+    GotOcr2VisionEncoder,
+)
+
+
+class SLANeXtVisionConfig(GotOcr2VisionConfig):
```
zucchini-nlp, line 52: we'll need a `auto_docstring` and `strict` decorators here so we can use correct checkpoint for SlaNext. Mabe you need to rebase first

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+        Number of regression values predicted per token for bounding box location (e.g., 8 for four corner
+        coordinates).
+    """,
+)
+class SLANeXtConfig(PreTrainedConfig):
+
+    model_type = "slanext"
+    sub_configs = {"vision_config": SLANeXtVisionConfig}
+
+    def __init__(
```
zucchini-nlp: needs to be rebased and turned into a `dataclass`

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+
+
+class SLANeXtVary_VIT_B(nn.Module):
+    def __init__(
+        self,
+        vision_config: SLANeXtConfig,
+        post_conv_in_channels: int=256,
+        post_conv_out_channels: int=512,
+    ):
+        super().__init__()
```
zucchini-nlp: nit: i think this is quite timm-format naming, maybe `SlaNextBackbone`? And the `vision_config` can be simply `config`

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+class SLANeXtVisionEncoder(GotOcr2VisionEncoder):
+    pass
+
+
+class SLANeXtVary_VIT_B(nn.Module):
+    def __init__(
+        self,
+        vision_config: SLANeXtConfig,
+        post_conv_in_channels: int=256,
+        post_conv_out_channels: int=512,
```
zucchini-nlp: these two are already in config, no?

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+        alpha = alpha.transpose(1, 2)
+        context = torch.bmm(alpha, batch_hidden).squeeze(1)
+        concat_context = torch.cat([context, char_onehots], 1)
+
+        cur_hidden = self.rnn(concat_context, prev_hidden)
+
+        return (cur_hidden, cur_hidden), alpha
+
+
+class SLANeXtHWAttention(nn.Module):
```
zucchini-nlp: not used anywhere

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+    img: B C H W
+    """
+    batch_size, height, width, channels = img.shape
+    img_reshape = img.reshape(batch_size, height // H_sp, H_sp, width // W_sp, W_sp, channels)
+    img_perm = img_reshape.permute(0, 1, 3, 2, 4, 5).reshape(-1, H_sp * W_sp, channels)
+    return img_perm
+
+
+def windows2img(img_splits_hw, H_sp, W_sp, H, W):
+    """
```
zucchini-nlp: same, can't find where this is called

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+            nn.Linear(hidden_size, out_channels),
+        )
+
+        dpr = np.linspace(0, 0.1, 2)
+
+        self.loc_generator = nn.Sequential(
+            nn.Linear(self.hidden_size, self.hidden_size),
+            nn.Linear(self.hidden_size, loc_reg_num),
+            nn.Sigmoid(),
+        )
```
zucchini-nlp: instead of sequential, let's define each one separately or split out to a Module (like `SlaNextMlp`). Sequentials make it hard to follow and debug

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+
+class SLANeXtSLAHead(nn.Module):
+    def __init__(
+        self,
+        in_channels=512,
+        hidden_size=512,
+        out_channels=30,
+        max_text_length=500,
+        loc_reg_num=4,
+        fc_decay=0.0,
```
zucchini-nlp: all of these come from config, so we can expect a single arg `config` here

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+        self.loc_reg_num = loc_reg_num
+        self.eos = self.num_embeddings - 1
+
+        self.structure_attention_cell = SLANeXtAttentionGRUCell(in_channels, hidden_size, self.num_embeddings)
+        self.structure_generator = nn.Sequential(
+            nn.Linear(self.hidden_size, self.hidden_size),
+            nn.Linear(hidden_size, out_channels),
+        )
+
+        dpr = np.linspace(0, 0.1, 2)
```
zucchini-nlp: not used

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+
+        self.loc_generator = nn.Sequential(
+            nn.Linear(self.hidden_size, self.hidden_size),
+            nn.Linear(self.hidden_size, loc_reg_num),
+            nn.Sigmoid(),
+        )
+
+    def forward(self, inputs, targets=None):
+        features = inputs
```
zucchini-nlp: no-op

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+        batch_size = features.shape[0]
+
+        hidden = torch.zeros((batch_size, self.hidden_size), device=features.device)
+        structure_preds = torch.zeros(
+            (batch_size, self.max_text_length + 1, self.num_embeddings), device=features.device
+        )
+        loc_preds = torch.zeros((batch_size, self.max_text_length + 1, self.loc_reg_num), device=features.device)
+        structure_preds.requires_grad = False
+        loc_preds.requires_grad = False
```
zucchini-nlp: we shouldn't set or unset grad in model forward. Is it for training-only or do we need to disable grad for the whole decoding?

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+
+        structure_step = self.structure_generator(output)
+        loc_step = self.loc_generator(output)
+        return hidden, structure_step, loc_step
+
+    def _char_to_onehot(self, input_char):
+        return F.one_hot(input_char, self.num_embeddings).float()
+
+
+@auto_docstring(custom_intro="The SLANeXt model.")
```
zucchini-nlp: Ig `custom_intro` was meant to be filled later?

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+        @param in_channels: input shape
+        @param hidden_size: hidden_size for RNN and Embedding
+        @param out_channels: num_classes to rec
+        @param max_text_length: max text pred
+        """
+        super().__init__()
+
+        self.hidden_size = hidden_size
+        self.max_text_length = max_text_length
+        self.emb = self._char_to_onehot
```
zucchini-nlp: i think this is not needed as well

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+            structure_preds[:, i, :] = structure_step
+            loc_preds[:, i, :] = loc_step
+
+            structure_ids[:, i] = pre_chars
+            if (structure_ids == self.eos).any(-1).all():
+                break
+        structure_preds = F.softmax(structure_preds[:, : i + 1], dim=-1)
+        loc_preds = loc_preds[:, : i + 1]
+
+        return structure_preds
```
zucchini-nlp: we don't return `ids` or `loc_preds`, thus no need to create an empty tensor and fill them as we're decoding

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+
+    def _decode(self, pre_chars, features, hidden):
+        """
+        Predict table label and coordinates for each step
+        @param pre_chars: Table label in previous step
+        @param features:
+        @param hidden: hidden status in previous step
+        @return:
+        """
```
zucchini-nlp: nit: dosctring doesn't follow the recommended format. Tbh it can be removed imo

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+        img = self._tablerec_resize(img, [width_resize, height_resize])
+        img = img / 255.0
+        img = normalize(image=img, mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225])
+        height, width = img.shape[:2]
+        pad_right = max(0, self.target_pad_size - width)
+        pad_bottom = max(0, self.target_pad_size - height)
+        img = pad(image=img, padding=((0, pad_bottom), (0, pad_right)))
+        img = img.transpose((2, 0, 1))
+        img = np.expand_dims(img, axis=0)
+        img = torch.tensor(img).float()
```
zucchini-nlp: let's rewrite to follow the image processor API. Preferable with fast processor

The model needs special resizing, normalization and padding, so you need to override `self.resize` method and define correct defaults as in ViT for example

https://github.com/huggingface/transformers/blob/main/src/transformers/models/vit/image_processing_vit_fast.py

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,680 @@
...
+            idx = np.array(self.dict["sos"])
+        elif beg_or_end == "end":
+            idx = np.array(self.dict["eos"])
+        else:
+            assert False, "unsupported type %s in get_beg_end_flag_idx" % beg_or_end
+        return idx
+
+    def post_process_table_recognition(self, outputs):
+        self.pred = outputs.last_hidden_state.detach().cpu()
+        structure_probs = np.array([list(self.pred[0])])
```
zucchini-nlp: this should happen before `post_process_table_recognition` is called, i.e. the users gets the correct pred and passed it to post-process

src/transformers/models/slanext/image_processing_slanext.py
```diff
@@ -0,0 +1,235 @@
...
+        self.td_token = ["<td>", "<td", "<td></td>"]
+
+    def get_beg_end_flag_idx(self, beg_or_end):
+        if beg_or_end == "beg":
+            idx = np.array(self.dict["sos"])
+        elif beg_or_end == "end":
+            idx = np.array(self.dict["eos"])
+        else:
+            assert False, "unsupported type %s in get_beg_end_flag_idx" % beg_or_end
+        return idx
```
zucchini-nlp: we can just save `self.bos_id = np.array(self.dict["sos"])` and use it when needed. A utility seems to be unnecessary

utils/check_repo.py
```diff
@@ -153,6 +153,7 @@
         "FastSpeech2ConformerWithHifiGan",  # Built with two smaller (tested) models.
         "GlmImageVQVAE",  # Building part of bigger (tested) model.
         "GraphormerDecoderHead",  # Building part of bigger (tested) model.
+        "SLANeXtModel",  # Building part of bigger (tested) model.
```
zucchini-nlp: let's remove it. `SLANeXtModel` is same as `SLANeXtTableRecognition`

utils/check_repo.py
```diff
@@ -389,6 +390,7 @@
     "HubertForCTC",
     "SEWForCTC",
     "SEWDForCTC",
+    "SLANeXtModel",
```
zucchini-nlp: same here, we don't have a base model and a separate task model

tests/models/slanext/test_modeling_slanext.py
```diff
@@ -0,0 +1,252 @@
+# coding = utf-8
```
zucchini-nlp, line 1: we also need a test for image processing because SlaNext has its own processor

zucchini-nlp (commented):
Hey, thanks for the PR! Could you rebase and update a few moments to align with the latest API?

docs/source/en/model_doc/slanext.md
liu-jiaxuan: done

docs/source/en/model_doc/slanext.md
liu-jiaxuan: done

docs/source/en/model_doc/slanext.md
liu-jiaxuan: done

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: done

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: done

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: done

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: done

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: done

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: done

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: done

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Yes, these two already exist in the config.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done. Split the two structures defined by `nn.Sequential `into two separate modules: `SLANeXtStructureMLP` and `SLANeXtLocationMLP`.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done. Removed the grad settings.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/image_processing_slanext.py
liu-jiaxuan: Done.

utils/check_repo.py
liu-jiaxuan: Done.

utils/check_repo.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: We rewrite the image processor API, and add the fast processor.

src/transformers/models/auto/modeling_auto.py
liu-jiaxuan: Done. We change the `SLANeXtForTableRecognition` to the `SLANeXtModel`.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/auto/configuration_auto.py
liu-jiaxuan: Hi, could we keep using the model's official name "SLANeXt" here (even though it may not strictly follow camel case)

src/transformers/models/auto/image_processing_auto.py
liu-jiaxuan: We add the fast processor API here.

docs/source/en/model_doc/slanext.md
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: done.

tests/models/slanext/test_modeling_slanext.py
liu-jiaxuan: done.

zhang-prog:
@zucchini-nlp We’ve finished the changes. PTAL.

Discussion:
`BaseImageProcessor.resize` affects accuracy, because this model uses OpenCV-based resizing during both training and inference. To avoid directly relying on `cv2.resize`, we implemented a NumPy-based approximation of `cv2.resize` in the slow image processor to better align with the original accuracy.

However, NumPy-style operations do not seem appropriate for the fast image processor, so we are inclined to think that this model does not need a `SLANeXtImageProcessorFast`.

What do you think? Also, for the sake of code cleanliness and readability, would it be acceptable to use `cv2.resize` directly in the slow image processor?

zucchini-nlp:
> What do you think? Also, for the sake of code cleanliness and readability, would it be acceptable to use cv2.resize directly in the slow image processor?

Unfortunately we don't have cv2 in dependencies list, so for slow processor we need to stick to numpy. In Fast processor, we can use numpy and torchvision, does `torchvision.F.resize` work for you?

src/transformers/models/slanext/modular_slanext.py
zucchini-nlp: not done, I see this auto-generated in config file. Needs a correct checkpoint

```
@auto_docstring(checkpoint="facebook/sam-vit-huge")
@strict(accept_kwargs=True)
```

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,823 @@
...
+        return (cur_hidden, cur_hidden), alpha
+
+
+class SLANeXtStructureMLP(nn.Module):
+    def __init__(self, hidden_size, out_channels):
+        super().__init__()
+        self.linear1 = nn.Linear(hidden_size, hidden_size)
+        self.linear2 = nn.Linear(hidden_size, out_channels)
+
+    def forward(self, x):
```
zucchini-nlp: nit: let's rename all `x` to smth more informative, like `hidden_states`

src/transformers/models/slanext/modular_slanext.py
zucchini-nlp: not done

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,823 @@
...
+        image = make_flat_list_of_images(image)
+        if not valid_images(image):
+            raise ValueError("Invalid image type. Must be of type PIL.Image.Image, numpy.ndarray, or torch.Tensor")
+        image = to_numpy_array(image[0])
+
+        if do_resize:
+            image = self.resize(image=image, size=size)
+
+        if do_normalize:
+            image = image.astype(np.float32) / 255.0
```
zucchini-nlp: this is done if `do_rescale` with `rescale_factor` 

https://github.com/huggingface/transformers/blob/aa57e1cd2fd0ede5ffbc70db3f193943b8f3e720/src/transformers/models/vit/image_processing_vit.py#L260-L264

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,823 @@
...
+        else:
+            squeeze_output = False
+
+        # OpenCV's fixed-point arithmetic constants
+        INTER_RESIZE_COEF_BITS = 11
+        INTER_RESIZE_COEF_SCALE = 1 << INTER_RESIZE_COEF_BITS  # 2048
+
+        # Calculate scale factors using float32 (matching OpenCV)
+        scale_x = np.float32(width) / np.float32(target_width)
+        scale_y = np.float32(height) / np.float32(target_height)
```
zucchini-nlp: can we use PIL.resize as an equivalent?

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,823 @@
...
+            `max(size["height"], size["width"])` while preserving the aspect ratio. Can be overridden by the `size`
+            parameter in the `preprocess` method.
+        do_rescale (`bool`, *optional*, defaults to `True`):
+            Whether to rescale the image by the specified scale `rescale_factor`. Can be overridden by the
+            `do_rescale` parameter in the `preprocess` method.
+        rescale_factor (`int` or `float`, *optional*, defaults to `1/255`):
+            Scale factor to use if rescaling the image. Can be overridden by the `rescale_factor` parameter in the
+            `preprocess` method.
+        do_normalize (`bool`, *optional*, defaults to `True`):
+            Whether to normalize the image. Can be overridden by the `do_normalize` parameter in the `preprocess`
```
zucchini-nlp: not needed, `auto_docstring` adds it for us

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,823 @@
...
+)
+class SLANeXtForTableRecognition(SLANeXtPreTrainedModel):
+    """
+    SLANeXt model for table recognition tasks.
+    """
+
+    def __init__(self, config: SLANeXtConfig):
+        super().__init__(config)
+        self.model = SLANeXtModel(config)
+        self.post_init()
```
zucchini-nlp: Not a fan of wrapper model that does nothing. We can rename `SLANeXtModel` to `SLANeXtForTableRecognition` if needed

zucchini-nlp (commented):
I see a few comments not yet resolved, commented there. As for resize, in slow processor we can use `PIL` and I believe it will work for you case. PIL has all the same resize types as torchvision

Actually we can even delete the slow processor, as fast processor is enough and we're defaulting to load a fast processor anyway. Much easier for you than fixing the code prob

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Sorry for missing this. Removed the unnecessary class variables and now only use `self.config` as the single class variable.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Thanks for the suggestion! After my experiments, both `PIL.resize` and `torchvision.F.resize` introduce non-negligible accuracy loss in SLANeXt's preprocessing. Therefore, I implemented a custom resize method with minimal accuracy loss (without CPU operations), while all other preprocessing steps use operators from image_processing_utils_fast. Additionally, I removed the slow image processor and kept only the fast one.

src/transformers/models/auto/image_processing_auto.py
```diff
@@ -204,7 +204,8 @@
             ("shieldgemma2", ("Gemma3ImageProcessor", "Gemma3ImageProcessorFast")),
             ("siglip", ("SiglipImageProcessor", "SiglipImageProcessorFast")),
             ("siglip2", ("Siglip2ImageProcessor", "Siglip2ImageProcessorFast")),
-            ("smolvlm", ("SmolVLMImageProcessor", "SmolVLMImageProcessorFast")),
+            ("slanext", ("SLANeXtImageProcessor", "SLANeXtImageProcessorFast")),
+            ("slanext", (None, "SLANeXtImageProcessorFast")),
```
zucchini-nlp: nit: duplicate key, the first row has to be deleted ig

src/transformers/models/auto/image_processing_auto.py
```diff
@@ -204,7 +204,8 @@
             ("shieldgemma2", ("Gemma3ImageProcessor", "Gemma3ImageProcessorFast")),
             ("siglip", ("SiglipImageProcessor", "SiglipImageProcessorFast")),
             ("siglip2", ("Siglip2ImageProcessor", "Siglip2ImageProcessorFast")),
-            ("smolvlm", ("SmolVLMImageProcessor", "SmolVLMImageProcessorFast")),
```
zucchini-nlp: accidental delete? to revert

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,520 @@
...
+        image_mean: float | list[float] | None,
+        image_std: float | list[float] | None,
+        do_pad: bool | None,
+        pad_size: SizeDict | None,
+        disable_grouping: bool | None,
+        return_tensors: str | TensorType | None,
+        **kwargs,
+    ) -> BatchFeature:
+        if do_resize:
+            resized_images = [self._resize(img, size) for img in images]
```
zucchini-nlp: Why not group by shape?

zucchini-nlp (commented):
So last comment: 
- `SLANeXtVisionConfig`  has incorrect checkpoint in auto-generated config file I believe. Still hasn't been fixed
- A few typos commented below 👇🏻 
- Most important q: does the tiny difference in resize affect the model's performance or could we adopt `torchvision`?

src/transformers/models/auto/image_processing_auto.py
liu-jiaxuan: My mistake... I accidentally deleted the wrong lines instead of the ones that should have been removed. They have been restored.

src/transformers/models/auto/image_processing_auto.py
liu-jiaxuan: My mistake... I accidentally deleted the wrong lines instead of the ones that should have been removed. They have been restored.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done. This has been modified to follow the same resize procedure as in `BaseImageProcessorFast`.

liu-jiaxuan:
> So last comment:
> 
> * `SLANeXtVisionConfig`  has incorrect checkpoint in auto-generated config file I believe. Still hasn't been fixed
> * A few typos commented below 👇🏻
> * Most important q: does the tiny difference in resize affect the model's performance or could we adopt `torchvision`?


Hi @zucchini-nlp，thanks for your comment!

1. We have corrected the checkpoint for SLANeXtVisionConfig.
2. We have fixed the typo you pointed out.
3. We conducted extensive experiments, and the results show that using the resize method based on `PIL` or `torchvision` introduces difference during the preprocessing of the SLANeXt model. Such difference lead to unacceptable performance degradation for the model.

zucchini-nlp (approved):
Got it, thanks for verifying ❤️  Wanted to make sure so we simplify the code if possible. If it affects the performance and torchvision doesn't have the needed resize, we can leave as is

Thanks a lot for iterating, let's merge @vasqu

docs/source/en/model_doc/slanext.md
```diff
@@ -0,0 +1,79 @@
...
+from PIL import Image
+from transformers import AutoImageProcessor, AutoModel
+
+model_path="PaddlePaddle/SLANeXt_wired_safetensors"
+model = AutoModel.from_pretrained(model_path, dtype=torch.float32)
+image_processor = AutoImageProcessor.from_pretrained(model_path, use_fast=True)
+
+image = Image.open(requests.get("https://paddle-model-ecology.bj.bcebos.com/paddlex/imgs/demo_image/table_recognition.jpg", stream=True).raw).convert("RGB")
+inputs = image_processor(images=image, return_tensors="pt")
+outputs = model(**inputs)
```
vasqu:
```suggestion
model = AutoModel.from_pretrained(model_path, dtype=torch.float32, device_map="auto")
image_processor = AutoImageProcessor.from_pretrained(model_path)

image = Image.open(requests.get("https://paddle-model-ecology.bj.bcebos.com/paddlex/imgs/demo_image/table_recognition.jpg", stream=True).raw)
inputs = image_processor(images=image, return_tensors="pt").to(model.device)
outputs = model(**inputs)
```
A few smaller things
- Do we still need the rgb conversion? Should maybe added to the image processor?
- Let's use a device map
- We should have a separate auto model for `ForTableRecognition`, see https://github.com/huggingface/transformers/pull/44808/changes/d0e841d619bc7028068ceae25f92e7f0e1198a38

src/transformers/models/auto/modeling_auto.py
```diff
@@ -790,6 +791,7 @@ class _BaseModelWithGenerate(PreTrainedModel, GenerationMixin):
         ("resnet", "ResNetModel"),
         ("segformer", "SegformerModel"),
         ("siglip_vision_model", "SiglipVisionModel"),
+        ("slanext", "SLANeXtModel"),
```
vasqu: This is a bit weird, do we need this here? Like I mentioned before I think an auto mapping for table recognition makes sense here

src/transformers/models/slanext/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+from typing import TYPE_CHECKING
+
+from ...utils import _LazyModule
+from ...utils.import_utils import define_import_structure
+
+
+if TYPE_CHECKING:
+    from .configuration_slanext import *
+    from .modeling_slanext import *
```
vasqu, line 24:
```suggestion
    from .configuration_slanext import *
    from .modeling_slanext import *
    from .image_processing_slanext_fast import *
```

src/transformers/models/auto/modeling_auto.py
vasqu: Ok might be needed - just to double check

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        Number of output channels for the post-encoder convolution layer.
+    out_channels (`int`, *optional*, defaults to 50):
+        Number of output token classes for the structure prediction head (i.e., vocabulary size for table structure
+        tokens).
+    max_text_length (`int`, *optional*, defaults to 500):
+        Maximum number of decoding steps (tokens) for the autoregressive structure and location decoder.
+    loc_reg_num (`int`, *optional*, defaults to 8):
+        Number of regression values predicted per token for bounding box location (e.g., 8 for four corner
+        coordinates).
+    """,
```
vasqu: Nit: Would prefer if we moved this into the body instead - more standard with the refactor

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+            self.vision_config = SLANeXtVisionConfig(**self.vision_config)
+        super().__post_init__(**kwargs)
+
+
+class SLANeXtPreTrainedModel(PreTrainedModel):
+    """
+    Base class for all SLANeXt pre-trained models. Handles model initialization,
+    configuration, and loading of pre-trained weights, following the Transformers library conventions.
+    """
+
```
vasqu:
```suggestion
```
not really needed

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        # Initialize relative positional embeddings to zero (SLANeXtVisionAttention holds rel_pos_h/w)
+        if isinstance(module, SLANeXtVisionAttention):
+            if module.use_rel_pos:
+                init.constant_(module.rel_pos_h, 0.0)
+                init.constant_(module.rel_pos_w, 0.0)
+
+        # Initialize GRUCell (replicates PyTorch default reset_parameters)
+        if isinstance(module, nn.GRUCell):
+            stdv = 1.0 / math.sqrt(module.hidden_size) if module.hidden_size > 0 else 0
```
vasqu:
```suggestion
            std = 1.0 / math.sqrt(module.hidden_size) if module.hidden_size > 0 else 0
```
just for consistency as used elsewhere, same below

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        self.post_conv_out_channels = config.post_conv_out_channels
+
+        self.vision_tower = SLANeXtVisionEncoder(self.vision_config)
+        self.post_conv = nn.Conv2d(
+            self.post_conv_in_channels, self.post_conv_out_channels, kernel_size=3, stride=2, padding=1, bias=False
+        )
+
+    def forward(self, hidden_states):
+        if hidden_states.shape[1] == 1:
+            hidden_states = torch.repeat_interleave(hidden_states, repeats=3, dim=1)
```
vasqu: Is this if needed, can we just base call here?

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        hidden_states = self.vision_tower(hidden_states).last_hidden_state
+        hidden_states = self.post_conv(hidden_states)
+        hidden_states = hidden_states.flatten(2).permute(0, 2, 1)
```
vasqu:
```suggestion
        hidden_states = hidden_states.flatten(2).transpose(1, 2)
```
nit

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+
+    def forward(self, prev_hidden, batch_hidden, char_onehots):
+        batch_hidden_proj = self.input_to_hidden(batch_hidden)
+        prev_hidden_proj = self.hidden_to_hidden(prev_hidden).unsqueeze(1)
+
+        attention_scores = batch_hidden_proj + prev_hidden_proj
+        attention_scores = torch.tanh(attention_scores)
+        attention_scores = self.score(attention_scores)
+
+        alpha = F.softmax(attention_scores, dim=1)
```
vasqu:
```suggestion
        alpha = F.softmax(attention_scores, dim=1, dtype=torch.float32).to(attention_scores.dtype)
```
softmax is unstable in fp16 without this

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        alpha = alpha.transpose(1, 2)
+        context = torch.bmm(alpha, batch_hidden).squeeze(1)
```
vasqu: lets use torch matmul instead please

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        concat_context = torch.cat([context, char_onehots], 1)
```
vasqu:
```suggestion
        prefixed_context = torch.cat([context, char_onehots], 1)
```
maybe some different naming, concat is not a nice naming imo

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+
+        cur_hidden = self.rnn(concat_context, prev_hidden)
+
+        return (cur_hidden, cur_hidden), alpha
```
vasqu:
```suggestion
        hidden_states = self.rnn(concat_context, prev_hidden)

        return hidden_states, attn_weights
```
let's rename a bit + why output the same information twice?

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+    def __init__(self, input_size, hidden_size, num_embeddings, use_gru=False):
+        super().__init__()
+
+        self.input_to_hidden = nn.Linear(input_size, hidden_size, bias=False)
+        self.hidden_to_hidden = nn.Linear(hidden_size, hidden_size)
+        self.score = nn.Linear(hidden_size, 1, bias=False)
+
+        self.rnn = nn.GRUCell(input_size + num_embeddings, hidden_size)
+
+    def forward(self, prev_hidden, batch_hidden, char_onehots):
```
vasqu: Let's add typing here

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        cur_hidden = self.rnn(concat_context, prev_hidden)
+
+        return (cur_hidden, cur_hidden), alpha
+
+
+class SLANeXtStructureMLP(nn.Module):
+    def __init__(self, hidden_size, out_channels):
+        super().__init__()
+        self.linear1 = nn.Linear(hidden_size, hidden_size)
+        self.linear2 = nn.Linear(hidden_size, out_channels)
```
vasqu:
```suggestion
        self.fc1 = nn.Linear(hidden_size, hidden_size)
        self.fc2 = nn.Linear(hidden_size, out_channels)
```
just more common across the code base

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        super().__init__()
+        self.linear1 = nn.Linear(hidden_size, hidden_size)
+        self.linear2 = nn.Linear(hidden_size, loc_reg_num)
+        self.sigmoid = nn.Sigmoid()
+
+    def forward(self, hidden_states):
+        hidden_states = self.linear1(hidden_states)
+        hidden_states = self.linear2(hidden_states)
+        hidden_states = self.sigmoid(hidden_states)
+        return hidden_states
```
vasqu: Imo could be one simple MLP combined:
- Rename with fc1/fc2
- Add act_fn - `nn.Identity() if activation is None else ACT2CLS[activation]()`
- Have same forward then

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+
+        self.config = config
+
+        self.structure_attention_cell = SLANeXtAttentionGRUCell(
+            config.post_conv_out_channels, config.hidden_size, config.out_channels
+        )
+        self.structure_generator = SLANeXtStructureMLP(config.hidden_size, config.out_channels)
+        self.loc_generator = SLANeXtLocationMLP(config.hidden_size, config.loc_reg_num)
+
+    def forward(self, features, targets=None):
```
vasqu: typing please

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        hidden = torch.zeros((batch_size, self.config.hidden_size), device=features.device)
+        structure_preds_list = []
+        structure_ids_list = []
+        pre_chars = torch.zeros(size=[batch_size], dtype=torch.long, device=features.device)
+        for _ in range(self.config.max_text_length + 1):
+            hidden, structure_step, loc_step = self._decode(pre_chars, features, hidden)
+            pre_chars = structure_step.argmax(dim=1)
+            structure_preds_list.append(structure_step)
+            structure_ids_list.append(pre_chars)
+            if torch.stack(structure_ids_list, dim=1).eq(self.config.out_channels - 1).any(-1).all():
```
vasqu, line 262: Hmm, not super compile friendly but guess we dont have a choice

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        self.structure_attention_cell = SLANeXtAttentionGRUCell(
+            config.post_conv_out_channels, config.hidden_size, config.out_channels
+        )
+        self.structure_generator = SLANeXtStructureMLP(config.hidden_size, config.out_channels)
+        self.loc_generator = SLANeXtLocationMLP(config.hidden_size, config.loc_reg_num)
+
+    def forward(self, features, targets=None):
+        batch_size = features.shape[0]
+
+        hidden = torch.zeros((batch_size, self.config.hidden_size), device=features.device)
```
vasqu: hidden_states

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        structure_ids_list = []
+        pre_chars = torch.zeros(size=[batch_size], dtype=torch.long, device=features.device)
+        for _ in range(self.config.max_text_length + 1):
+            hidden, structure_step, loc_step = self._decode(pre_chars, features, hidden)
+            pre_chars = structure_step.argmax(dim=1)
+            structure_preds_list.append(structure_step)
+            structure_ids_list.append(pre_chars)
+            if torch.stack(structure_ids_list, dim=1).eq(self.config.out_channels - 1).any(-1).all():
+                break
+        structure_preds = F.softmax(torch.stack(structure_preds_list, dim=1), dim=-1)
```
vasqu: See my comments about softmax in the attn rnn cell

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+
+        return structure_preds
+
+    def _decode(self, pre_chars, features, hidden):
+        emb_feature = self._char_to_onehot(pre_chars)
```
vasqu: Imo, we don't need a separate function, just call it directly here

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        (output, hidden), alpha = self.structure_attention_cell(hidden, features, emb_feature)
```
vasqu:
```suggestion
        hidden_states, _ = self.structure_attention_cell(hidden, features, emb_feature)
        residual = hidden_states
        ....
        return residual, hidden_states, loc_step
```

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        structure_step = self.structure_generator(output)
+        loc_step = self.loc_generator(output)
+        return hidden, structure_step, loc_step
+
+    def _char_to_onehot(self, input_char):
+        return F.one_hot(input_char, self.config.out_channels).float()
+
+
+@auto_docstring
+class SLANeXtModel(SLANeXtPreTrainedModel):
```
vasqu: Imo should be a `ForXXX` model instead (have commented earlier on this)

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+    @auto_docstring
+    def forward(
+        self, pixel_values: torch.FloatTensor, **kwargs: Unpack[TransformersKwargs]
+    ) -> tuple[torch.FloatTensor] | BaseModelOutputWithNoAttention:
+        backbone_states = self.backbone(pixel_values)
+        hidden_states = self.head(backbone_states)
+        return BaseModelOutputWithNoAttention(
+            last_hidden_state=hidden_states,
+            hidden_states=backbone_states,
+        )
```
vasqu: We are not really collecting any hidden states, i.e. hidden states between layers - imo we can collect two sets of hidden states:
- The head hidden states and the backbone hidden states
- Check out `PPOCRV5MobileRecForTextRecognition`, we have done something similar there

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+    image_std = IMAGENET_DEFAULT_STD
+    size = {"height": 512, "width": 512}
+    pad_size = {"height": 512, "width": 512}
+    rescale_factor = 1 / 255
+    do_resize = True
+    do_rescale = True
+    do_normalize = True
+    do_pad = True
+    valid_kwargs = ImagesKwargs
+    model_input_names = ["pixel_values"]
```
vasqu:
```suggestion
    do_resize = True
    do_rescale = True
    do_normalize = True
    do_pad = True
```
should already be defaults

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+
+    def _resize(
+        self,
+        image: torch.Tensor,
```
vasqu:
```suggestion
        image: "torch.Tensor",
```
for typing just to be sure

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        src_x_floor = src_x.floor().to(torch.int32)
+        src_x_frac = src_x - src_x_floor.float()
+        # boundary handling
+        src_x_frac = torch.where(src_x_floor < 0, torch.zeros_like(src_x_frac), src_x_frac)
+        src_x_floor = torch.where(src_x_floor < 0, torch.zeros_like(src_x_floor), src_x_floor)
+        src_x_frac = torch.where(src_x_floor >= width - 1, torch.ones_like(src_x_frac), src_x_frac)
+        src_x_floor = torch.where(src_x_floor >= width - 1, torch.full_like(src_x_floor, width - 2), src_x_floor)
+        # fixed-point weights
+        weight_x_r = (src_x_frac * 2048 + 0.5).floor().to(torch.int32)  # round-to-nearest
+        weight_x_l = 2048 - weight_x_r  # (target_w,)
```
vasqu: let's avoid single letters where possible, e.g. here with r/l

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        target_y = torch.arange(target_height, dtype=torch.float32, device=device)
+        src_y = (target_y + 0.5) * (float(height) / float(target_height)) - 0.5
+        src_y_floor = src_y.floor().to(torch.int32)
+        src_y_frac = src_y - src_y_floor.float()
+        src_y_frac = torch.where(src_y_floor < 0, torch.zeros_like(src_y_frac), src_y_frac)
+        src_y_floor = torch.where(src_y_floor < 0, torch.zeros_like(src_y_floor), src_y_floor)
+        src_y_frac = torch.where(src_y_floor >= height - 1, torch.ones_like(src_y_frac), src_y_frac)
+        src_y_floor = torch.where(src_y_floor >= height - 1, torch.full_like(src_y_floor, height - 2), src_y_floor)
+        weight_y_b = (src_y_frac * 2048 + 0.5).floor().to(torch.int32)
+        weight_y_t = 2048 - weight_y_b  # (target_h,)
```
vasqu: same here

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+
+        img_u8 = image.clamp(0, 255).to(torch.uint8)  # (C, H, W)
+        img_i32 = img_u8.to(torch.int32)  # (C, H, W)
```
vasqu:
```suggestion
        image_u8 = image.clamp(0, 255).to(torch.uint8)  # (C, H, W)
        image_i32 = img_u8.to(torch.int32)  # (C, H, W)
```
nit

src/transformers/models/slanext/modular_slanext.py
```diff
@@ -0,0 +1,516 @@
...
+        # gather 4 neighbours: (C, target_h, target_w)
+        p00 = img_i32[:, y_top[:, None], x_left[None, :]]
+        p10 = img_i32[:, y_top[:, None], x_right[None, :]]
+        p01 = img_i32[:, y_bottom[:, None], x_left[None, :]]
+        p11 = img_i32[:, y_bottom[:, None], x_right[None, :]]
+        # fixed-point bilinear: weights broadcast over (C, target_h, target_w)
+        weight_y_b_ = weight_y_b.view(1, target_height, 1)
+        weight_y_t_ = weight_y_t.view(1, target_height, 1)
+        weight_x_r_ = weight_x_r.view(1, 1, target_width)
+        weight_x_l_ = weight_x_l.view(1, 1, target_width)
```
vasqu: Yea same here re single letters, can we be a bit more expressive

tests/models/slanext/test_modeling_slanext.py
```diff
@@ -0,0 +1,262 @@
...
+
+        return config
+
+
+@require_torch
+class SLANeXtModelTest(ModelTesterMixin, PipelineTesterMixin, unittest.TestCase):
+    all_model_classes = (SLANeXtModel,) if is_torch_available() else ()
+    pipeline_model_mapping = {"image-feature-extraction": SLANeXtModel} if is_torch_available() else {}
+
+    has_attentions = False
```
vasqu: I think `test_resize_embeddings = False` should help reduce the amount of skips we do manually below - not sure which tests exactly sorry 😓

tests/models/slanext/test_modeling_slanext.py
```diff
@@ -0,0 +1,262 @@
...
+    def test_model_get_set_embeddings(self):
+        pass
+
+    @unittest.skip(reason="SLANeXt does not support input and output embeddings")
+    def test_model_common_attributes(self):
+        pass
+
+    @unittest.skip(reason="SLANeXt does not support batching inference")
+    def test_batching_equivalence(self):
+        pass
```
vasqu: This should definitely be raised as an error based on the input

vasqu: Any chance this could be supported, it is usually quite important to have this

tests/models/slanext/test_modeling_slanext.py
```diff
@@ -0,0 +1,262 @@
...
+    def test_inputs_embeds_matches_input_ids(self):
+        pass
+
+    @unittest.skip(reason="SLANeXt does not use tensor output")
+    def test_determinism(self):
+        pass
+
+    @unittest.skip(reason="SLANeXt does not use tensor output")
+    def test_torch_export(self):
+        pass
```
vasqu: `test_torch_exportable = False` as flag instead

tests/models/slanext/test_modeling_slanext.py
```diff
@@ -0,0 +1,262 @@
...
+    def test_resize_tokens_embeddings(self):
+        pass
+
+    @unittest.skip(reason="Feed forward chunking is not implemented")
+    def test_feed_forward_chunking(self):
+        pass
+
+    @unittest.skip(reason="SLANeXt has a minimum parameter count of 1.7M")
+    def test_model_is_small(self):
+        pass
```
vasqu: Really not possible to make smaller? Would be quite important tbh for CI to remain fast

tests/models/slanext/test_modeling_slanext.py
```diff
@@ -0,0 +1,262 @@
...
+
+    @unittest.skip(reason="SLANeXt does not support attention")
+    def test_retain_grad_hidden_states_attentions(self):
+        pass
+
+    @unittest.skip(reason="SLANeXt does not support attention")
+    def test_attention_outputs(self):
+        pass
```
vasqu: Should be skipped via flag already

vasqu (commented):
I've left a few comments re modeling but there is one core limitation that needs to be lifted: Compatibility with batching. If I saw it correctly, then batches are not supported but this is a fundamental requirement for us to add a model

zhang-prog:
@vasqu Sry, we missed the `test_batching_equivalence` test. We have now removed the skip and it is passing successfully, @liu-jiaxuan is also continuing to address the other comments.

vasqu:
Awesome, thank you @zhang-prog

src/transformers/models/slanext/__init__.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

tests/models/slanext/test_modeling_slanext.py
liu-jiaxuan: Done. We have now removed the skip and it is passing successfully.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/auto/modeling_auto.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

tests/models/slanext/test_modeling_slanext.py
liu-jiaxuan: Done.

tests/models/slanext/test_modeling_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Yes, that is indeed the case for now.

docs/source/en/model_doc/slanext.md
liu-jiaxuan: Done.

tests/models/slanext/test_modeling_slanext.py
liu-jiaxuan: Thanks for the suggestion! I have set `test_resize_embeddings = False` and removed the skip  of `test_resize_tokens_embeddings` below.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

src/transformers/models/slanext/modular_slanext.py
liu-jiaxuan: Done.

tests/models/slanext/test_modeling_slanext.py
liu-jiaxuan: Sry, the SLANeXt model cannot be made any smaller (minimum ~1.7M), but I have updated the test config to the model's minimal state.

vasqu:
@liu-jiaxuan @zhang-prog I pushed my fixes:
- We need to recheck the integration tests in general and the keys have changed because the base model prefix was wrong
- I didn't find any usage of `self.loc_generator` which I commented out and added todo, pls check

vasqu:
run-slow: slanext

vasqu (approved):
Checked locally with our runners, everything passes now! Merging

Thanks a lot @liu-jiaxuan and @zhang-prog for sticking through

PR #43709, opened 2026-02-03
fix: `VersionComparison.from_string` return type mismatch

tarekziade:
# What does this PR do?

I had this typing failure

```
$ ty check src/transformers/utils/import_utils.py

[snip]

error[call-non-callable]: Object of type `VersionComparison` is not callable
    --> src/transformers/utils/import_utils.py:2317:16
     |
2316 |       def is_satisfied(self) -> bool:
2317 |           return VersionComparison.from_string(self.version_comparison)(
     |  ________________^
2318 | |             version.parse(self.get_installed_version()), version.parse(self.version)
2319 | |         )
     | |_________^
2320 |
2321 |       def __repr__(self) -> str:
```

Why: 

1. The from_string() method has a return type annotation of -> "VersionComparison" (indicating it should return an enum member)
2. But the implementation was returning .value (the operator function like operator.ge)
  
This patch aligns the behavior with the API signature


ArthurZucker (approved):
ty, was that a "regression" from package version?

tarekziade:
> ty, was that a "regression" from package version?

no, it's simply a mismatch between the typing annotation that claims the function returns a `VersionComparison` instance not a callable, I picked it with `ty`

in main:

```
>>> from transformers.utils.import_utils import VersionComparison
>>> VersionComparison.from_string(">=")
<built-in function ge>
```

in this patch:

```
>>> from transformers.utils.import_utils import VersionComparison
>>> VersionComparison.from_string(">=")
<VersionComparison.GREATER_THAN_OR_EQUAL: <built-in function ge>>
```

So the patch just aligns behavior so `from_string` really returns a `VersionComparison`

PR #43692, opened 2026-02-02
Fix DeepSpeed ZeRO-3 + PEFT + Trainer scheduler mismatch on PyTorch 2.10+

qgallouedec:
## Summary

On PyTorch 2.10+, `Trainer.train()` crashes at the first `lr_scheduler.step()` when using DeepSpeed ZeRO-3 with a PEFT model.

This PR provides fix, alothough I'm sure it's not the ideal one.

The failure only appears with torch 2.10+ probably because PyTorch made the scheduler update strict (`zip(..., strict=True)`), surfacing a latent mismatch between the scheduler’s `base_lrs` and the optimizer’s param groups after DeepSpeed wraps the optimizer. This PR rebuilds the scheduler after DeepSpeed/Accelerate preparation if a mismatch is detected.

## Reproduction

From the `transformers` repo:

```sh
accelerate launch --config_file zero3.yaml mre.py
```

`mre.py`:

```python
from datasets import Dataset
from transformers import AutoTokenizer, AutoModelForCausalLM, Trainer, TrainingArguments
from peft import LoraConfig, get_peft_model

model_id = "Qwen/Qwen3-0.6B"
data = [
    [46518, 374, 2664, 1091, 27261],
    [98923, 374, 2664, 1091, 17995],
    [16374, 374, 2664, 1091, 6351],
    [31137, 374, 2664, 1091, 16965],
]
Trainer(
    model=get_peft_model(AutoModelForCausalLM.from_pretrained(model_id), LoraConfig()),
    args=TrainingArguments(bf16=True),  # not related: bf16=False fails
    train_dataset=Dataset.from_dict({"input_ids": data, "labels": data}),
    processing_class=AutoTokenizer.from_pretrained(model_id),
).train()
```

`zero3.yaml`:

```yaml
distributed_type: DEEPSPEED
deepspeed_config:
  zero_stage: 3
num_processes: 2
```

**Observed error (torch 2.10+)**

```
W0202 21:38:37.706000 2150915 site-packages/torch/distributed/run.py:852] 
W0202 21:38:37.706000 2150915 site-packages/torch/distributed/run.py:852] *****************************************
W0202 21:38:37.706000 2150915 site-packages/torch/distributed/run.py:852] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed. 
W0202 21:38:37.706000 2150915 site-packages/torch/distributed/run.py:852] *****************************************
[...]

PR #43711, opened 2026-02-03
Fix scheduler initialization order

SunMarc:
# What does this PR do?

This PR put the `lr_scheduler` initialization after the optimizer is prepared to deal with any modification that might happen after. This is mainly to deal with deepspeed as the optimizer is modified and starting with torch 2.10, checks are more strict. 

Fixes https://github.com/huggingface/transformers/pull/43692


src/transformers/trainer.py
```diff
@@ -2327,21 +2329,23 @@ def _inner_training_loop(
                 self._fsdp_qlora_plugin_updates()
                 if self.accelerator.mixed_precision != "fp8":
                     self.model = self.accelerator.prepare(self.model)
-            self.create_optimizer_and_scheduler(num_training_steps=max_steps)
```
qgallouedec, line 2325: wanna keep `create_optimizer_and_scheduler` for backward compatibility? Deprecating it is an option

qgallouedec (approved):
thanks, closing #43692

ArthurZucker (approved):
checks are stricter with 2.10, but this does not change anything for 2.4?

SunMarc:
> checks are stricter with 2.10, but this does not change anything for 2.4?

Nope !

src/transformers/trainer.py
SunMarc: Yeah keeping it for BC for now but I might deprecate it slowly !

PR #43719, opened 2026-02-03
Add id and resume support to SwanLab integration

### LINKED ISSUE #43698: SwanLab integration uses outdated swanlab.init() signature
### Feature request

Request:

Thanks.

### Motivation

### Your contribution

None

i-pj:
Fixes #43698
- Add SWANLAB_RUN_ID env var to pass experiment ID for resumption
- Add SWANLAB_RESUME env var to control resume mode
- Default to resume='allow' when resume_from_checkpoint is set"

# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

Anyone in the community is free to review the PR once the tests have passed.


Rocketknight1:
cc @SunMarc

SunMarc (commented):
Can you check if this fixes your use case @TianjiaoTsao ?

TianjiaoTsao:
> Can you check if this fixes your use case @TianjiaoTsao ?

Thanks. Yes, this fixes my use case.

SunMarc (approved):

SunMarc:
Thans @i-pj !

PR #43722, opened 2026-02-03
🚨 fix + tests dense & MoE TP all reduce (decoder only)

3outeille:
Let's make sure it works for decoder only first (We skip VLM + Encoder-decoder for now)

Introduction, forward, backward, generation (with convert mapping triggering) test agains TP vs non-TP baseline

```python
import torch
import os
from torch.distributed.elastic.multiprocessing.errors import record

model_id = "mistralai/Mixtral-8x7B-Instruct-v0.1"
# model_id = "Qwen/Qwen1.5-MoE-A2.7B-Chat"
# model_id = "Qwen/Qwen3-30B-A3B-Instruct-2507"

rank = int(os.environ["RANK"])
world_size = int(os.environ["WORLD_SIZE"])
device = torch.device(f"cuda:{rank}")
# Need to be initialized explicitly to use the `barrier` before loading
torch.distributed.init_process_group(backend="nccl", rank=rank, world_size=world_size, device_id=rank)

@record
def main():

    model = AutoModelForCausalLM.from_pretrained(model_id, dtype=torch.bfloat16, tp_plan="auto")
    # model = AutoModelForCausalLM.from_pretrained(model_id, dtype=torch.bfloat16, device_map="auto")

    messages = [
        {"role": "user", "content": "What do you think about life?"},
    ]
    inputs = tokenizer.apply_chat_template(messages, return_tensors="pt").to(model.device)
    input_size = inputs.input_ids.shape[-1]
    output = model.generate(**inputs, max_new_tokens=100, do_sample=False)
    text = tokenizer.batch_decode(output[:, input_size:])[0]
    print(text)

main()

```
image

- `./run_dense_tests.sh results_dense`
image
- `./run_moe_tests.sh results_moe`
image


run_dense_tests.sh
```diff
@@ -0,0 +1,267 @@
+#!/bin/bash
+
+# Script to run tensor parallel (TP) tests for Dense models
+# Tests are run sequentially as each TP test uses 2 GPUs internally
+# Usage: ./run_dense_tests.sh /path/to/results
+
+# Define colors for output
+GREEN='\033[0;32m'
```
Cyrilvallez: Probably much more practical to have those as proper tests, not 2 bash scripts living a bit randomly in the lib! 🤗

tests/test_tensor_parallel_mixin.py
```diff
@@ -0,0 +1,380 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
```
Cyrilvallez: It's 2026 hehe

Cyrilvallez (commented):
Just a few very early thoughts!

run_dense_tests.sh
3outeille: it's just an util script for me to run tests faster when developping (this will be removed at the end). You mean having them as native support in python ? I think the way tests are launched in the CI is parallel no ?

Cyrilvallez: Yeah I meant we want to run those tests on the slow CI!

3outeille: they all use tiny models by default so no need for slow. Plus, I made a specific test for MoE that force the weight converter to be triggered by first `revert_weight_conversion`, save it, load it back so that weight converter to be triggered

tests/test_tensor_parallel_mixin.py
```diff
@@ -0,0 +1,511 @@
...
+
+            # Step 1: Create model with fused weights (internal representation)
+            model = model_class(config)
+            # Step 2: Get the current state dict (fused format)
+            state_dict = model.state_dict()
+            # Step 3: Revert to unfused format (simulates original checkpoint format, e.g., w1/w3/w2 separate)
+            original_state_dict = revert_weight_conversion(model, state_dict)
+            # Step 4: Save checkpoint files in the original unfused format
+            save_file(original_state_dict, os.path.join(tmp_dir, "model.safetensors"))
+            model.config.save_pretrained(tmp_dir)
```
ArthurZucker: please use model.save_preatrained(save_original_format=False)`

ArthurZucker (commented):
Not a fan of the bash either, mostly because we don't know which attack vectors could potentially be leveraged here :) :) 

I don't really mind it here as we want a quick fix but yes not optimal. 

The tests are not really clear to read / clearly separated. Its probably missing on or two cases like EP, and setting a different expert implementation

src/transformers/models/longcat_flash/modeling_longcat_flash.py
```diff
@@ -210,7 +210,10 @@ def forward(self, hidden_states, top_k_index, top_k_weights):
             current_state = hidden_states[token_idx]
 
             if expert_idx >= self.num_routed_experts or self.gate_up_proj is None:
+                # Zero expert: identity function. in TP case, we need to scale down the output by 1/tp_world_size otherwise it will get summed twice during all-reduce
```
3outeille: flag it for now (just to pass all the tests). We dont want to change the modeling. Our TP implementation doesnt have enough granularity. An easy fix would be to just do a Specific class for the model (but that can easily gets out of hand)

src/transformers/models/glm4_moe_lite/modeling_glm4_moe_lite.py
```diff
@@ -306,6 +306,14 @@ def forward(
 
         k_rot = k_rot.view(batch_size, 1, seq_length, self.qk_rope_head_dim)
 
+        # In TP mode, k_rot bypasses kv_b_proj (colwise) so its gradient from local
```
3outeille: same comments as longcat_flash

.circleci/create_circleci_config.py
```diff
@@ -328,6 +328,15 @@ def job_name(self):
     parallelism=6,
 )
 
+tensor_parallel_ci_job = CircleCIJob(
+    "tensor_parallel_ci",
+    additional_env={"RUN_TENSOR_PARALLEL_TESTS": True},
+    docker_image=[{"image": "huggingface/transformers-torch-light"}],
+    install_steps=["uv pip install ."],
+    marker="is_tensor_parallel_test",
+    parallelism=6,
```
ArthurZucker, line 337: don't know if we need 6, but we can adjust later on

src/transformers/integrations/tensor_parallel.py
```diff
@@ -460,7 +460,7 @@ def backward(ctx, grad_output):
         device_mesh = ctx.device_mesh
         if device_mesh.size() == 1:
             return grad_output, None
-        dist.all_reduce(grad_output, op=dist.ReduceOp.SUM, group=device_mesh.get_group())
+        dist.all_reduce(grad_output, op=dist.ReduceOp.SUM, group=device_mesh.get_group(), async_op=False)
```
ArthurZucker: we don't have to put false here

src/transformers/integrations/tensor_parallel.py
```diff
@@ -603,6 +603,14 @@ def all_reduce_forward(x, device_mesh):
     return _AllReduceForward.apply(x, device_mesh)
 
 
+def _all_reduce_gradient(grad, device_mesh):
+    """All-reduce a parameter gradient across the TP mesh."""
+    if device_mesh.size() == 1:
+        return grad
+    dist.all_reduce(grad, op=dist.ReduceOp.SUM, group=device_mesh.get_group(), async_op=False)
+    return grad
```
ArthurZucker: its the same as the forward above

src/transformers/integrations/tensor_parallel.py
```diff
@@ -724,6 +732,59 @@ def get_expected_sharded_shape(self, full_shape: tuple[int, ...] | torch.Size) -
         return tuple(shape)
 
 
+class AllReduce(TensorParallelLayer):
```
ArthurZucker: if specific to expert, we should not call it all readuce please

src/transformers/integrations/tensor_parallel.py
```diff
@@ -724,6 +732,59 @@ def get_expected_sharded_shape(self, full_shape: tuple[int, ...] | torch.Size) -
...
+
+    def prepare_module_tp(self, module, device_mesh):
+        # Use a module-level backward hook (not param.register_hook) because parameters are replaced during weight loading after this method runs.
+        # Module hooks survive parameter replacement.
+        def _backward_hook(mod, grad_input, grad_output, mesh=device_mesh):
+            for param in mod.parameters():
+                if param.grad is not None:
+                    _all_reduce_gradient(param.grad, mesh)
+
+        module.register_full_backward_hook(_backward_hook)
```
ArthurZucker: this might be worth improving our api to have `prepare_backward_input_fn` or something that generalizes well. 
Remember we need an API that scales!

src/transformers/models/apertus/configuration_apertus.py
```diff
@@ -101,6 +101,8 @@ class ApertusConfig(PreTrainedConfig):
         "layers.*.self_attn.q_proj": "colwise",
         "layers.*.self_attn.k_proj": "colwise",
         "layers.*.self_attn.v_proj": "colwise",
+        "layers.*.self_attn.q_norm": "replicated_with_grad_allreduce",
+        "layers.*.self_attn.k_norm": "replicated_with_grad_allreduce",
```
ArthurZucker, line 105: good, did it have training issues before?

src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
```diff
@@ -360,6 +360,15 @@ def forward(
...
         k_pe = k_pe.view(batch_size, 1, seq_length, self.qk_rope_head_dim)
+
+        # In TP mode, k_pe bypasses kv_b_proj (colwise) so its gradient from local
+        # heads is only a partial sum. all_reduce_backward fixes this in backward.
+        device_mesh = getattr(self.kv_b_proj, "_hf_device_mesh", None)
+        if device_mesh is not None:
+            from ...integrations.tensor_parallel import all_reduce_backward
+
+            k_pe = all_reduce_backward(k_pe, device_mesh)
+
```
ArthurZucker: we should never ever ever have this kind of code here!

ArthurZucker: Here for example the hook should be on `apply_rotary_emb`? or a tensor hook? idk but this is a big no!

src/transformers/models/longcat_flash/modeling_longcat_flash.py
ArthurZucker: i don't understand why you have to do 
```
current_hidden_states /= torch.distributed.get_world_size()
```
? you are normalizing ?

src/transformers/models/longcat_flash/modular_longcat_flash.py
```diff
@@ -131,7 +131,10 @@ def forward(self, hidden_states, top_k_index, top_k_weights):
             current_state = hidden_states[token_idx]
 
             if expert_idx >= self.num_routed_experts or self.gate_up_proj is None:
+                # Zero expert: identity function. in TP case, we need to scale down the output by 1/tp_world_size otherwise it will get summed twice during all-reduce
```
ArthurZucker: we should fix how we sum , or scale in the all reduce but we should never change this

tests/models/flex_olmo/test_modeling_flex_olmo.py
```diff
@@ -41,6 +41,13 @@ class FlexOlmoModelTester(CausalLMModelTester):
     if is_torch_available():
         base_model_class = FlexOlmoModel
 
+    def __init__(self, parent):
+        super().__init__(parent=parent)
+        # NOTE(3outeille): must be 0.0 for TP backward tests. In train mode, non-zero dropout causes
+        # different RNG states between the non-TP and TP model forward passes (they run sequentially),
+        # leading to different dropout masks and mismatched losses.
+        self.attention_probs_dropout_prob = 0.0
```
ArthurZucker, line 49:
```
from transformers import set_seed
set_seed(0)
```
should work

src/transformers/modeling_utils.py
ArthurZucker, line 1: revert whatever is unrelated please

src/transformers/core_model_loading.py
```diff
@@ -847,6 +846,7 @@ def _format_op_name(curr_op: list[ConversionOps] | ConversionOps | None) -> str
             return curr_op.__class__.__name__
 
         op_name = _format_op_name(op)
+        import traceback
```
ArthurZucker: unrelated

ArthurZucker (commented):
ty, can you revert / rebase ?

src/transformers/integrations/tensor_parallel.py
3outeille: need to revert, probably from the merge of `EP post merge` branch

src/transformers/models/apertus/configuration_apertus.py
3outeille: yep, TP gradients were not matching with non-TP gradients

src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
3outeille: > we should never ever ever have this kind of code here!

yeah I flagged this earlier. I just wanted the code to pass all the tests then cleaning afterwards

src/transformers/models/longcat_flash/modeling_longcat_flash.py
3outeille: Since we are in `LongcatFlashExperts.forward()` and that we are going to `all_reduce` at the end of it, that means if we go into the `if expert_idx >= self.num_routed_experts or self.gate_up_proj is None:` branch, current_hidden_states will get `all_reduce` TP-size times which is not correct. That's why we have to scale it down

src/transformers/integrations/tensor_parallel.py
3outeille: yep need to revert merge

tests/models/flex_olmo/test_modeling_flex_olmo.py
3outeille: it doesnt seem to help

3outeille:
run-slow: apertus, deepseek_v2, deepseek_v3, dots1, ernie4_5_moe, exaone4, exaone_moe, flex_olmo, gemma2, gemma3, gemma3n, glm4_moe, glm4_moe_lite, glm_moe_dsa

3outeille:
run slow test fails despite CI working. I guess I need to add the `@require_multiple_accelerators` but that challenge the whole idea of having those tests on CPUs no ? (some models will now find themselves running the tests twice, one on CPUs and one on GPUs with run_slow) @ArthurZucker @Cyrilvallez

src/transformers/core_model_loading.py
```diff
@@ -847,19 +846,15 @@ def _format_op_name(curr_op: list[ConversionOps] | ConversionOps | None) -> str
             return curr_op.__class__.__name__
 
         op_name = _format_op_name(op)
-
-        tb_str = "".join(traceback.format_exception(type(e), e, e.__traceback__))
```
ArthurZucker, line 869: let's keeps this one please!

src/transformers/core_model_loading.py
```diff
@@ -972,8 +967,9 @@ def rename_source_key(
 def convert_and_load_state_dict_in_model(
     model: PreTrainedModel,
     state_dict: dict[str, Any],
-    load_config: LoadStateDictConfig,
+    load_config: Any,
     tp_plan: dict[str, str] | None,
+    dtype_plan: dict | None = None,
```
ArthurZucker: not sure we want to revert this

src/transformers/core_model_loading.py
```diff
@@ -1155,11 +1151,7 @@ def convert_and_load_state_dict_in_model(
                         mapping.distributed_operation = tp_layer(
                             device_mesh=device_mesh, rank=device_mesh.get_local_rank(), empty_param=empty_param.clone()
                         )
-                    shard_index = (
-                        len(mapping.collected_tensors.get(source_pattern, []))
-                        if isinstance(mapping, WeightConverter) and isinstance(mapping.operations[0], MergeModulelist)
-                        else None
-                    )
```
ArthurZucker, line 1187: this is important for "EP" sharding no?

src/transformers/modeling_utils.py
```diff
@@ -277,20 +278,23 @@ def get_state_dict_dtype(state_dict):
...
-    "U64": torch.uint64,
     "F8_E4M3": torch.float8_e4m3fn,
     "F8_E5M2": torch.float8_e5m2,
 }
 
 
+if is_torch_greater_or_equal("2.3.0"):
+    str_to_torch_dtype["U16"] = torch.uint16
+    str_to_torch_dtype["U32"] = torch.uint32
+    str_to_torch_dtype["U64"] = torch.uint64
```
ArthurZucker: we don't support 2.3 only >= 2.4

src/transformers/modeling_utils.py
ArthurZucker, line 1: there is a lot to revert here still (cleanup)

3outeille:
run-slow: apertus, deepseek_v2, deepseek_v3, dots1, ernie4_5_moe, exaone4, exaone_moe, flex_olmo, gemma2, gemma3, gemma3n, glm4_moe, glm4_moe_lite, glm_moe_dsa, gpt_oss

3outeille:
run-slow: apertus, deepseek_v2, deepseek_v3, dots1, ernie4_5_moe, exaone4, exaone_moe, flex_olmo, gemma2, gemma3, gemma3n, glm4_moe, glm4_moe_lite, glm_moe_dsa, gpt_oss

3outeille:
run-slow: apertus, deepseek_v2, deepseek_v3, dots1, ernie4_5_moe, exaone4, exaone_moe, flex_olmo, gemma3

.circleci/create_circleci_config.py
```diff
@@ -328,6 +328,15 @@ def job_name(self):
     parallelism=6,
 )
 
+tensor_parallel_ci_job = CircleCIJob(
+    "tensor_parallel_ci",
```
ArthurZucker, line 332:
```suggestion
    "distributed_ci",
```
as it will probably cover more than just TP

.gitignore
```diff
@@ -34,6 +34,8 @@ wheels/
 *.egg
 MANIFEST
 
+results*
+
```
ArthurZucker: to remove imo

debug_tp.py
```diff
@@ -0,0 +1,81 @@
+"""Quick debug script to understand the TP crash."""
```
ArthurZucker: missing licence, also probably to remove / put in examples rather than here?

dynamic_weight_loading.md
```diff
@@ -0,0 +1,725 @@
+# Dynamic Weight Loading in Transformers
```
ArthurZucker: missing licence + we already have a doc file for this I think: https://github.com/huggingface/transformers/blob/6c777c168057ac6cc6b1636eda6842b58bc4559f/docs/source/en/weightconverter.md#L1

ArthurZucker: would be nice to merge both as yours adds some value IMO (its quite visual)

log.txt
ArthurZucker, line 1: make sure to remove it!

log_ci.txt
ArthurZucker, line 1: same here

mixtral_forward_backward_tp.md
ArthurZucker, line 1: to remove as well!

run_dense_tests.sh
ArthurZucker, line 1: i don't mind using this rather than pytest if its simpler / collects faster. My main concern is to have it dangling around at the root of the repo. let's have them in `scripts` folder or examples or something.

run_dense_tests.sh
```diff
@@ -0,0 +1,420 @@
...
+    ["starcoder2"]="tests/models/starcoder2/test_modeling_starcoder2.py"
+    ["t5"]="tests/models/t5/test_modeling_t5.py"
+    ["t5gemma"]="tests/models/t5gemma/test_modeling_t5gemma.py"
+    ["t5gemma2"]="tests/models/t5gemma2/test_modeling_t5gemma2.py"
+    ["umt5"]="tests/models/umt5/test_modeling_umt5.py"
+    ["vaultgemma"]="tests/models/vaultgemma/test_modeling_vaultgemma.py"
+    ["xglm"]="tests/models/xglm/test_modeling_xglm.py"
+    ["xlstm"]="tests/models/xlstm/test_modeling_xlstm.py"
+    ["youtu"]="tests/models/youtu/test_modeling_youtu.py"
+)
```
ArthurZucker: for future reference, we should extract the list from all models and exclude some no? main concern is if someone adds a new model he has to remember (and the reviewer) that he's gotta update that list. This does not scale well. 

There are a few ways to enforce this, via repo checks (see makefile) or by making the list an exclusion list instead. 

Exclusion list can work as a good objective (let's push it to 0 exclusion) and have justification per skipped model.

src/transformers/integrations/tensor_parallel.py
ArthurZucker, line 1: We need to decide now how much work we do for the user:
- colwise: should it split the input by itself if not split
- rowise: same 
- rowise_split_input: when would you explicitly have that in that case 

We broken this when we removed `DTensor`.

src/transformers/integrations/tensor_parallel.py
```diff
@@ -1087,6 +1143,29 @@ def shard_tensor(
         return param[...].to(device=device, dtype=dtype)
 
 
+class MoeIdentityExpertParallel(TensorParallelLayer):
+    """
+    TP class for zero/identity experts in MoE layers.
+
+    Under TP, the parent MoeTensorParalellExperts does all_reduce_forward (sum)
+    on the expert module output. Identity experts produce the same output on
+    every rank, so the sum gives world_size * output. This class divides the
```
ArthurZucker, line 1177: i don't understand the need for identity experts, can you explain a bit more?

src/transformers/integrations/tensor_parallel.py
```diff
@@ -724,6 +725,61 @@ def get_expected_sharded_shape(self, full_shape: tuple[int, ...] | torch.Size) -
...
+        pass_output, rope_output = output.split([output.shape[-1] - rope_dim, rope_dim], dim=-1)
+        rope_output = all_reduce_backward(rope_output, device_mesh)
+        return torch.cat([pass_output, rope_output], dim=-1)
+
+    def shard_tensor(self, param, tensor_idx=None, device=None, dtype=None):
+        return param[...].to(device=device, dtype=dtype)
+
+    def prepare_module_tp(self, module, device_mesh, model=None, **kwargs):
+        if model is not None and hasattr(model.config, "qk_rope_head_dim"):
+            module._rope_dim = model.config.qk_rope_head_dim
```
ArthurZucker: would be better to pas the config than the model IMO as you don't use the model

src/transformers/integrations/tensor_parallel.py
```diff
@@ -724,6 +725,61 @@ def get_expected_sharded_shape(self, full_shape: tuple[int, ...] | torch.Size) -
...
+                if param.grad is not None:
+                    all_reduce_forward(param.grad, mesh)
+
+        module.register_full_backward_hook(_backward_hook)
+
+
+class MlaKvAProjParallel(TensorParallelLayer):
+    """
+    For MLA attention: kv_a_proj_with_mqa output is [kv_lora_rank + qk_rope_head_dim].
+    The rope portion bypasses kv_b_proj (colwise), so needs all_reduce_backward
```
ArthurZucker: I read this, but I don't understand why because of bypass we have to use an all reduce (its missing the link) let's be as detailed as possible ! 🤗

tmp_gen.py
ArthurZucker, line 1: to remove as well!

tests/test_tensor_parallel_mixin.py
```diff
@@ -0,0 +1,387 @@
...
+
+    Returns:
+        tuple: (model_tp, model_ref, device)
+    """
+    model_tp = model_class.from_pretrained(model_path, tp_plan="auto")
+    dist.barrier()
+
+    device = model_tp.device
+    model_ref = model_class.from_pretrained(model_path)
+    model_ref = model_ref.to(device)
```
ArthurZucker, line 125: IDK how many devices you have / size of the models but device_map = "auto" could help as well. 
But fine to keep as is. 

But model ref is gonna be on all ranks no?
```suggestion
    if device.index==0:
        model_ref = model_class.from_pretrained(model_path)
        model_ref = model_ref.to(device)
```

tests/test_tensor_parallel_mixin.py
```diff
@@ -0,0 +1,387 @@
...
+                        expected_size = (param_full.size(dim) + world_size - 1) // world_size
+                        assert param.size(dim) <= expected_size, (
+                            f"Weight {name} sharding incorrect: expected <= {expected_size}, got {param.size(dim)}"
+                        )
+                    break
+
+    return sharded_params
+
+
+def _test_tp_forward_impl(_rank, model_path, model_class, atol, rtol):
```
ArthurZucker, line 164: nice

tests/test_tensor_parallel_mixin.py
```diff
@@ -0,0 +1,387 @@
...
+    set_seed(0)
+    input_ids = torch.randint(0, vocab_size, (2, 64)).to(device)
+    set_seed(0)
+    labels = torch.randint(0, vocab_size, (2, 64)).to(device)
+
+    loss = model(input_ids, labels=labels).loss
+    loss.backward()
+
+    loss_tp = model_tp(input_ids, labels=labels).loss
+    loss_tp.backward()
```
ArthurZucker, line 209: I did not check but all dropout should be disabled otherwise you are not getting the same hs if you don't reset the seed between the run of the two models right?

tests/test_tensor_parallel_mixin.py
```diff
@@ -0,0 +1,387 @@
...
+                        else:
+                            # regular slicing
+                            shard_size = grad_tp.size(dim)
+                            start = rank * shard_size
+                            grad = grad.narrow(dim, start, shard_size)
+                        break
+
+            assert torch.allclose(grad.cpu(), grad_tp.cpu(), atol=atol, rtol=rtol), (
+                f"Gradients differ for parameter {name}. Max diff: {(grad.cpu() - grad_tp.cpu()).abs().max().item()}"
+            )
```
ArthurZucker: imo here a "subtest" would be nice (you just create a set of the name of the layers that don't pass this assertion -> list of stuff to fix

tests/test_tensor_parallel_mixin.py
```diff
@@ -0,0 +1,387 @@
...
+
+    dist.barrier()
+
+
+def _test_tp_generation_impl(_rank, model_path, model_class, atol, rtol, max_new_tokens):
+    """Implementation for comparing TP and non-TP model generation outputs (direct load path)."""
+    set_seed(0)
```
ArthurZucker, line 252: nice, its heavy but nice

tests/test_tensor_parallel_mixin.py
```diff
@@ -0,0 +1,387 @@
...
+            return self.model_tester.causal_lm_class
+        return self.all_model_classes[0]
+
+    def _skip_if_not_supported(self):
+        """Check and skip test if TP is not supported for this model/environment."""
+        if not is_torch_greater_or_equal("2.9"):
+            self.skipTest("Tensor parallel tests require torch >= 2.9")
+
+        if torch.cuda.is_available():
+            self.skipTest("Tensor parallel mixin tests are CPU-only and should not run on GPU machines")
```
ArthurZucker, line 391: more like these are the fast tests, slow ones are gonna be much more involved but we will ahve tp tests non CPU

tests/test_tensor_parallel_mixin.py
```diff
@@ -0,0 +1,387 @@
...
+            self.skipTest(
+                f"Tensor parallel tests require at least {self.tensor_parallel_size} CPUs, "
+                f"but only {os.cpu_count()} available"
+            )
+
+        if not hasattr(self.model_tester, "causal_lm_class") or self.model_tester.causal_lm_class is None:
+            self.skipTest("Model tester does not have causal_lm_class (not using CausalLMModelTester)")
+
+        if not self._has_tp_plan():
+            self.skipTest("Model does not have a tensor parallel plan (base_model_tp_plan)")
```
ArthurZucker, line 403: we should raise here since you are already filtering the list of models for which this  will run no?

ArthurZucker (commented):
Great work overall! 
Let's cleanup the artifacts, and find a way to make sure new models are tested automatically!

.circleci/create_circleci_config.py
3outeille: distributed CI is another CI but for training: https://github.com/huggingface/transformers/pull/42765

tests/test_tensor_parallel_mixin.py
3outeille: yes and I tried resetting the seed (so that we can enable the dropout) but it didn't change anything. So we are forced to set dropout to 0

src/transformers/integrations/tensor_parallel.py
3outeille: let me add a visual

src/transformers/integrations/tensor_parallel.py
3outeille: Since we are in `LongcatFlashExperts.forward()` and that we are going to all_reduce at the end of it, that means if we go into the `if expert_idx >= self.num_routed_experts or self.gate_up_proj is None:` branch, `current_hidden_states` will get all_reduce TP-size times which is not correct. That's why we have to scale it down

src/transformers/integrations/tensor_parallel.py
3outeille: done

src/transformers/integrations/tensor_parallel.py
3outeille: we can always make `rowise_split_input` handled automatically inside `rowise` by checking shape mismatch but then it will hide the fact that an ops before has replicated the input which is less expressive. As you want, no strong opinion

3outeille:
run-slow: apertus, deepseek_v2, deepseek_v3, dots1, ernie4_5_moe, exaone4, exaone_moe, flex_olmo, gemma3, gemma3n, glm4_moe, glm4_moe_lite, glm_moe_dsa

src/transformers/integrations/tensor_parallel.py
ArthurZucker: but if its the first op the replication comes from just TP, let me think 1 sec its just that its breaking + can be cumbersome but explicitness is good TBH

src/transformers/integrations/tensor_parallel.py
ArthurZucker: `if expert_idx >= self.num_routed_experts or self.gate_up_proj is None:` never happens (unless you are running EP)

3outeille: as discussed offline, it does happen only in the case of `LongcatFlash` model

docs/source/en/weightconverter.md
```diff
@@ -16,21 +16,158 @@ rendered properly in your Markdown viewer.
...
+## Full loading pipeline
+
+All models go through the dynamic weight loading system. Conversion mapping is an **optional step within that system** that only activates when the model has entries in `_MODEL_TO_CONVERSION_PATTERN`.
+
+```
+Checkpoint File → from_pretrained() → convert_and_load_state_dict_in_model()
+                                              ↓
+                         ┌──────────────────────────────────────┐
+                         │  For each weight in checkpoint:      │
+                         │  1. Match key to model parameter     │
```
ArthurZucker:
```suggestion
                         │  For each weight in checkpoint:      │
                         │  1. Match renamed / processed source key to model parameter     │
```
1. process the key
2. Shard the weight (send to device ASYNC)
3. collect all results at the same time: tensors that have the same "source_pattern" together (for example moe but alos gateup proj etc) -> very important
4. then you apply Dequantization / Deserialization 1st if needed
5. Then you apply conversion 
6. then you apply quantization if 4) was not applied
7. then you set. 

This is wrong!

docs/source/en/weightconverter.md
```diff
@@ -16,21 +16,158 @@ rendered properly in your Markdown viewer.
...
+| Step | When it activates |
+|------|-------------------|
+| Dynamic loading | Always, for all models |
+| Conversion mapping | Only when `model_type` is in `_MODEL_TO_CONVERSION_PATTERN` |
+| TP sharding | Only when `tp_plan="auto"` and model has `base_model_tp_plan` |
+| Quantization | Only when a quantization config is provided |
+
+### Dense models (e.g., Llama)
+
+For dense models, the checkpoint format matches the model format directly, so no conversion mapping is needed. TP sharding still applies.
```
ArthurZucker: not always the case

docs/source/en/weightconverter.md
```diff
@@ -16,21 +16,158 @@ rendered properly in your Markdown viewer.
...
+│  │              │    │              │    │ - MergeModulelist│  │
+│  └──────────────┘    └──────────────┘    │ - Transpose      │  │
+│                                          │ - etc.           │  │
+│                                          └──────────────────┘  │
+│                                                                 │
+│  ┌──────────────────────────────────────────────────────────┐  │
+│  │                  ThreadPoolExecutor                       │  │
+│  │           (Async tensor materialization)                  │  │
+│  └──────────────────────────────────────────────────────────┘  │
+└─────────────────────────────────────────────────────────────────┘
```
ArthurZucker: true but this graph is not super super complete WRT how stuff happen see my correction!

docs/source/en/weightconverter.md
```diff
@@ -94,6 +231,160 @@ WeightConverter(
...
+    return []  # Override in subclass
+```
+
+Example for TorchAO:
+```python
+WeightConverter(
+    source_patterns=[":qdata", ":scale"],
+    target_patterns="",
+    operations=[TorchaoDeserialize()],
+)
```
ArthurZucker: more details needed if we add this

docs/source/en/weightconverter.md
```diff
@@ -105,6 +396,43 @@ If your system runs other heavy processes, multiple threads may slow down loadin
...
+```
+1. Checkpoint iteration:
+   - For each key, submit materialization job
+   - Job returns Future (async) or Callable (sync)
+   - Add to WeightConverter.collected_tensors
+
+2. Conversion phase:
+   - materialize_tensors() waits for all Futures
+   - Applies conversion operations
+   - Sets parameters on model
```
ArthurZucker: not perfectly accurate

docs/source/en/weightconverter.md
```diff
@@ -118,6 +446,146 @@ For example, a MoE model using [`MergeModulelist`] for experts on each layer, th
...
+        target_patterns="mlp.experts.gate_up_proj",
+        operations=[MergeModulelist(dim=0), Concatenate(dim=1)],
+    ),
+    WeightConverter(
+        source_patterns="mlp.experts.*.down_proj.weight",
+        target_patterns="mlp.experts.down_proj",
+        operations=[MergeModulelist(dim=0)],
+    ),
+],
+```
```
ArthurZucker: imo not really needed

docs/source/en/weightconverter.md
```diff
@@ -118,6 +446,146 @@ For example, a MoE model using [`MergeModulelist`] for experts on each layer, th
...
+    ),
+
+    # Split experts between text and vision
+    WeightConverter(
+        source_patterns=["experts.*.down_proj.weight"],
+        target_patterns=[
+            "text_moe.experts.down_proj",
+            "vision_moe.experts.down_proj",
+        ],
+        operations=[ErnieFuseAndSplitTextVisionExperts(stack_dim=0, concat_dim=1)],
```
ArthurZucker, line 498: this is interesting tho yes but should be like "see how you can add your custom op"

src/transformers/integrations/tensor_parallel.py
```diff
@@ -724,6 +725,81 @@ def get_expected_sharded_shape(self, full_shape: tuple[int, ...] | torch.Size) -
...
+    (colwise)      |
+        |          |
+        k_pass     k_rot
+            \\      /
+               cat
+                |
+            key_states
+
+    k_pass is passed to kv_b_proj (colwise) which has built-in all_reduce_backward so we don't have a partial gradient for it.
+    However, k_rot goes straight to attention, never touches kv_b_proj. So we need to average gradient across all ranks otherwise we only get gradient for one rank (partial gradient).
```
ArthurZucker, line 785: ty

src/transformers/integrations/tensor_parallel.py
```diff
@@ -724,6 +725,81 @@ def get_expected_sharded_shape(self, full_shape: tuple[int, ...] | torch.Size) -
...
+    """
+
+    def _prepare_output_fn(self, mod, output, device_mesh):
+        rope_dim = mod._rope_dim
```
ArthurZucker: i think we should use config / an attr that already exists if possible

src/transformers/models/deepseek_v2/configuration_deepseek_v2.py
```diff
@@ -123,12 +123,19 @@ class DeepseekV2Config(PreTrainedConfig):
...
         "layers.*.self_attn.kv_b_proj": "colwise",
         "layers.*.self_attn.o_proj": "rowwise",
         "layers.*.mlp.experts.gate_up_proj": "packed_colwise",
         "layers.*.mlp.experts.down_proj": "rowwise",
+        "layers.*.mlp.experts": "moe_tp_experts",
+        "layers.*.mlp.shared_experts.gate_proj": "colwise",
+        "layers.*.mlp.shared_experts.up_proj": "colwise",
+        "layers.*.mlp.shared_experts.down_proj": "rowwise",
+        "layers.*.mlp.gate_proj": "colwise",
+        "layers.*.mlp.up_proj": "colwise",
```
ArthurZucker, line 137: IDK what's the most efficient to avoid having too many coms but LGTM otherwise

src/transformers/models/longcat_flash/modeling_longcat_flash.py
```diff
@@ -183,6 +183,7 @@ def __init__(self, config):
         self.zero_expert_num = config.zero_expert_num or 0
         self.total_experts = self.num_routed_experts + self.zero_expert_num
         self.act_fn = ACT2FN[config.hidden_act]
+        self.identity_expert = nn.Identity()
```
ArthurZucker, line 186: nice!

src/transformers/models/qwen2_moe/modular_qwen2_moe.py
```diff
@@ -125,7 +125,7 @@ def forward(self, hidden_states: torch.Tensor) -> tuple[torch.Tensor, torch.Tens
 
         shared_expert_output = F.sigmoid(self.shared_expert_gate(hidden_states_reshaped)) * shared_expert_output
 
-        expert_output += shared_expert_output
+        expert_output = expert_output + shared_expert_output
```
ArthurZucker, line 128: nice

src/transformers/conversion_mapping.py
ArthurZucker, line 1: these are probably just rebasing issues?

tests/test_tensor_parallel_mixin.py
```diff
@@ -0,0 +1,391 @@
...
+
+    # Compare logits/scores at each generation step
+    scores = torch.stack(output.scores)
+    scores_tp = torch.stack(output_tp.scores)
+
+    diff = (scores - scores_tp).abs()
+    assert torch.allclose(scores, scores_tp, atol=atol, rtol=rtol), (
+        f"TP and non-TP model generation logits differ (direct load path). "
+        f"Max diff: {diff.max().item()} | Mean diff: {diff.mean().item()}"
+    )
```
ArthurZucker, line 282: let's check generated tokens as well! (its the same but let's still do it please)

tests/test_tensor_parallel_mixin.py
```diff
@@ -0,0 +1,391 @@
...
+    def test_tp_generation(self):
+        # Test TP generation: unfused checkpoint → conversion mapping (if needed) → TP sharding → model → generate
+        self._skip_if_not_supported()
+
+        config = self.model_tester.get_config()
+
+        model_class = self._get_tp_model_class()
+        atol = self.tensor_parallel_atol
+        rtol = self.tensor_parallel_rtol
+        max_new_tokens = 10
```
ArthurZucker:
```suggestion
        max_new_tokens = 25
```
,models are small let's make more default

tests/test_tensor_parallel_mixin.py
ArthurZucker, line 1: nice, missing TP+quantize :) :) :) :)

ArthurZucker (commented):
Ok almost ready!!! 
On thing, we did not answer together: 
- do we split inputs by default (ex: rowise?)

Otherwise let's be careful with docs but LGTM

src/transformers/models/deepseek_v2/configuration_deepseek_v2.py
3outeille: until we have EP working properly, i think it's okay to leave it this way

ArthurZucker:
Just missing the doc updates and should be good to go!

src/transformers/integrations/tensor_parallel.py
```diff
@@ -786,7 +786,7 @@ class MlaKvAProjParallel(TensorParallelLayer):
     """
 
     def _prepare_output_fn(self, mod, output, device_mesh):
-        rope_dim = mod._rope_dim
+        rope_dim = mod.config.qk_rope_head_dim
```
ArthurZucker, line 795: We should raise error if attr does not exist to say add it to the auto mapping

src/transformers/integrations/torchao.py
ArthurZucker, line 1: cc @SunMarc this is valid but happy if you can have a look

src/transformers/models/longcat_flash/modular_longcat_flash.py
```diff
@@ -341,6 +342,7 @@ class LongcatFlashPreTrainedModel(PreTrainedModel):
         "attentions": LongcatFlashMLA,
     }
     _keys_to_ignore_on_load_unexpected = [r"model\.mtp.*"]
+    _keep_in_fp32_modules = ["classifier.weight"]
```
ArthurZucker:
```suggestion
    _keep_in_fp32_modules = ["classifier.weight"] # TODO let's make sure orignal code base has this, for now it fixes quantization
```

ArthurZucker (approved):

3outeille:
run-slow: apertus, deepseek_v2, deepseek_v3, dots1, ernie4_5_moe, exaone4, exaone_moe, flex_olmo, gemma3, gemma3n, glm4_moe, glm4_moe_lite, glm_moe_dsa

3outeille:
run-slow: apertus, deepseek_v2, deepseek_v3, dots1, ernie4_5_moe, exaone4, exaone_moe, flex_olmo, gemma3, gemma3n, glm4_moe, glm4_moe_lite, glm_moe_dsa

3outeille:
run-slow: apertus, deepseek_v2, deepseek_v3, dots1, ernie4_5_moe, exaone4, exaone_moe, flex_olmo, gemma3, gemma3n, glm4_moe, glm4_moe_lite, glm_moe_dsa

src/transformers/integrations/torchao.py
SunMarc: SG !

PR #43728, opened 2026-02-04
add `do_sample=False` to qwen2_5_vl model tests to stablize the output

kaixuanliu:
In qwen2_5_vl model's test file, we use `Qwen/Qwen2.5-VL-7B-Instruct` model, which its default generation config's `do_sample` param is `True`.  This causes some of the test cases are not steady, they will pass sometimes while sometimes get failed. In this PR we set `do_sample=False` explicitly to avoid this.

### DIFF
--- a/tests/models/qwen2_5_vl/test_modeling_qwen2_5_vl.py
+++ b/tests/models/qwen2_5_vl/test_modeling_qwen2_5_vl.py
@@ -467,7 +467,7 @@ class Qwen2_5_VLIntegrationTest(unittest.TestCase):
         # verify generation
         inputs = inputs.to(torch_device)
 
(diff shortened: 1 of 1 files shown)


kaixuanliu:
@ydshieh , Can you help review? Thx!

zucchini-nlp:
run-slow: qwen2_5_vl

zucchini-nlp (approved):
The slow CI is ✅  so I think it's ok to merge

kaixuanliu:
@zucchini-nlp Can you help merge if it is OK? Thx!

zucchini-nlp:
run-slow: qwen2_5_vl

zucchini-nlp:
oke, I think we merge because nothing much changes in expectations except for the "flag"

PR #39867, opened 2025-08-02
Fix default values of getenv

cyyever:
# What does this PR do?

The default value of getenv should be a string or None.

PR #43729, opened 2026-02-04
[Doc tests] Fix bug

NielsRogge:
# What does this PR do?

This PR fixes a bug regarding the doc tests. It also enables the doc tests for Eomt-DINOv3.

## Summary

The `SKIP_CUDA_DOCTEST` environment variable check in `src/transformers/testing_utils.py` has a bug that causes it to always evaluate to `True` by default, unintentionally skipping all doctests that contain CUDA-related code.


ydshieh (approved):
Great catch ... it's from 

Fix default values of getenv (#39867)

The fix loos good here! Thanks

BUG.md
ydshieh, line 1: This file should be deleted from the PR before merge?

PR #43730, opened 2026-02-04
Fix EP post merge

ArthurZucker:
# What does this PR do?

## Fix GPT-OSS ???
EP sharding works, the forward is broken for eager (does not work) and output giberissh for grouped_mm. 
Will work on a fix to have in the patch! 

## Fix any TP model 
- fix get shard tensor 
- fix get packed shard weights
- fix `tensor_idx`


src/transformers/integrations/tensor_parallel.py
```diff
@@ -724,6 +724,20 @@ def get_expected_sharded_shape(self, full_shape: tuple[int, ...] | torch.Size) -
         shape[dim] = end - start
         return tuple(shape)
 
+class AllReduce(TensorParallelLayer):
```
3outeille: we should rename it else it becomes hard to understand given we have

```
def all_reduce_forward(x, device_mesh):
    """All-reduce forward, identity backward. Use after rowwise layers."""
    return _AllReduceForward.apply(x, device_mesh)
```

3outeille: we will use moe_tp_experts instead (handle backwards)

src/transformers/models/solar_open/modular_solar_open.py
```diff
@@ -110,7 +110,7 @@ class SolarOpenConfig(Glm4MoeConfig):
         "layers.*.self_attn.o_proj": "rowwise",
         "layers.*.mlp.experts.gate_up_proj": "local_rowwise",
```
3outeille:
```
        "layers.*.mlp.experts.gate_up_proj": "packed_colwise",
        "layers.*.mlp.experts.down_proj": "rowwise",
        "layers.*.mlp.experts": "moe_tp_experts",
```

src/transformers/models/minimax_m2/modular_minimax_m2.py
```diff
@@ -138,7 +138,7 @@ class MiniMaxM2Config(PreTrainedConfig):
         "layers.*.mlp.gate": "colwise_rep",  # we need to replicate here to correctly route experts
         "layers.*.mlp.experts.gate_up_proj": "local_rowwise",
```
3outeille: same as solar_open model

src/transformers/quantizers/quantizer_fbgemm_fp8.py
```diff
@@ -140,7 +140,7 @@ def update_tp_plan(self, config):
                 "layers.*.self_attn.v_proj.weight": "colwise",
                 "layers.*.self_attn.v_proj.weight_scale": "colwise",
                 "layers.*.self_attn.o_proj.weight": "rowwise",
-                "layers.*.self_attn": "gather",
+                "layers.*.self_attn": "all_reduce",
```
3outeille: do we need all_reduce here ? isnt the rowwise from o_proj.weight supposed to trigger the all_reduce ? That would mean here 2 all_reduces

src/transformers/core_model_loading.py
```diff
@@ -846,15 +846,20 @@ def _format_op_name(curr_op: list[ConversionOps] | ConversionOps | None) -> str
             return curr_op.__class__.__name__
 
         op_name = _format_op_name(op)
+        import traceback
```
3outeille: import on top

src/transformers/quantizers/quantizer_fbgemm_fp8.py
ArthurZucker: you are right

3outeille:
lgtm nice job !

3outeille (approved):

PR #43731, opened 2026-02-04
Fix dtype in image-text-to-text pipe

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/43716, as per title

We didn't see errors even if we test `bf16` because most backbone models (e.g. SigLIP) cast pixels to target dtype inside the model code. I fixed it in pipe and also in pixtral model following siglip pattern

### DIFF
--- a/src/transformers/models/pixtral/modeling_pixtral.py
+++ b/src/transformers/models/pixtral/modeling_pixtral.py
@@ -494,7 +494,7 @@ class PixtralVisionModel(PixtralPreTrainedModel):
 
         # pass images through initial convolution independently
         target_dtype = self.patch_conv.weight.dtype
-        patch_embeds = self.patch_conv(pixel_values.to(dtype=patch_embeds))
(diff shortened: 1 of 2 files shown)


zucchini-nlp:
run-slow: pixtral, mistral3

Rocketknight1 (approved):
Fix looks good, so adding an approval early so you can merge it once you can get the CI green 😅

vasqu (approved):
lgtm, can we add a fast test for mixtral maybe?

zucchini-nlp:
I don't think it's really needed because the fix is now also inside `model.forward`. So even if we accidentally "break" the pipe in the future, the test won't catch anything 🙃 

The whole pattern of casting inputs inside model to the correct `dtype` is a bit non-consistent and I don't know why it came to exist in the first place. We gotta figure out if we keep it and update all models, or ask users to always cast to dtype manually

vasqu:
Gotcha, no worries. Asking users to do something won't end well imo (90% ignore instructions like these)

PR #43732, opened 2026-02-04
Fix accelerate integration import 

SunMarc:
# What does this PR do?

We need to move `get_module_size_with_ties` in transformers as the min version of accelerate 1.1.0 don't have this function. It was introduced in 1.2.0 actually . Another solution would be change the minimum version but I prefer not for now.


hmellor (approved):
LGTM, works with vLLM

PR #43734, opened 2026-02-04
Prepare and keep track of position ids in `generate`

### LINKED ISSUE #29149: Generate: support passing position_ids
Thank you @tengomucho, for uncovering this bug.

### The problem

In a nutshell, passing the correct `position_ids` to `generate` should result in exactly the same results as not passing them. In other words, the following test should pass on all models, if added to `GenerationTesterMixin`. We can see that it is failing in general.

```py
    def test_passing_position_ids(self):
        # Check that passing position ids to generate yields the same results as not passing them, if the position ids
        # are correctly built. If the test fails, it means one of two things:
        # 1 - the manual position ids are not being piped correctly; OR
        # 2 - the automated position ids are not being correctly built.
        for model_class in self.all_generative_model_classes:
            config, input_ids, attention_mask, _ = self._get_input_ids_and_config(batch_size=1)
            if config.is_encoder_decoder:
                self.skipTest("This model does not support position_ids")

            # To truly test this property, let's create a batch where the second row corresponds to the test input with
            # left padding of 1.
            pad_token = torch.tensor([[config.pad_token_id or 0]], device=input_ids.device, dtype=input_ids.dtype)
            input_ids = torch.cat((input_ids, torch.cat((pad_token, input_ids[:, 1:]), dim=1)), dim=0)
            pad_mask = torch.zeros((1, 1), dtype=attention_mask.dtype, device=attention_mask.device)
            attention_mask = torch.cat((attention_mask, torch.cat((pad_mask, attention_mask[:, 1:]), dim=1)), dim=0)
            position_ids = torch.clamp(torch.cumsum(attention_mask, dim=-1) - 1, min=0)

            config.use_cache = True
            config.is_decoder = True

            model = model_class(config).to(torch_device).eval()

            try:
                output_position_ids = model.generate(
                    input_ids,
                    position_ids=position_ids,
                    max_new_tokens=10
                )
            except ValueError as exc:
                if "The following `model_kwargs` are not used by the model: ['position_ids']" in str(exc):
                    self.skipTest("This model does not support position_ids")
                else:
                    raise
            output_no_position_ids = model.generate(
                input_ids,
                max_new_tokens=10
            )
            self.assertListEqual(output_no_position_ids.tolist(), output_position_ids.tolist())
```

### The fix

There are two root causes for this:
1. `position_ids` is rejected in some models when it is passed (e.g. see [here](https://github.com/huggingface/transformers/blob/3c00b885b92fbcd0e7451e56ccf424a2d5a19bbb/src/transformers/models/gpt2/modeling_gpt2.py#L1022)). These models often assume no padding when `position_ids` is rejected.
2. `position_ids` is never updated, so it is only correct when created from scratch (=not passed).

As such, a fix to this problem should consist in updating `position_ids` in `generate`, with `prepare_inputs_for_generation` only creating new `position_ids` when they don't exist.

zucchini-nlp:
# What does this PR do?

As per title. lays ground to unifying 3D position ids in qwen-style VLMs 

PR adds a single entrypoint to prepare position ids in `GenerationMixin` which models can override if needed (qwen-vl for ex). This allow users to prepare their own position ids and pass them to `generate()`. In decoding stages, the position ids are simply incremented by one to build the next positions

Along with it, PR starts a light unification on 3D positions by splitting it into its own utility fn. Now we have only two or three models with their own `compute_3d_positions` and all other models copy from there. In the next PR, I will split `get_rope_index` into smaller components allowing us to copy similarities easily. I am working on it locally but it's blocked by current branch

Review starting from `transformers/generation` and models from which we copy (qwen2-vl and ernie4_5_vl)

Fixes https://github.com/huggingface/transformers/issues/29149


zucchini-nlp:
run-slow: colqwen2, ernie4_5_vl_moe, gemma3, glm46v, glm4v, glm4v_moe, glm_image, glm_ocr, gpt_neo, paddleocr_vl, qwen2_5_vl, qwen2_vl, qwen3_vl, qwen3_vl_moe, reformer, video_llama_3

zucchini-nlp:
run-slow: colqwen2, ernie4_5_vl_moe, gemma3, glm46v, glm4v, glm4v_moe, glm_image, glm_ocr, gpt_neo, paddleocr_vl, qwen2_5_vl, qwen2_vl, qwen3_vl, qwen3_vl_moe, reformer, video_llama_3

src/transformers/generation/utils.py
```diff
@@ -653,7 +654,7 @@ def prepare_inputs_for_generation(
             and position_ids_key in set(inspect.signature(self.forward).parameters.keys())
         ):
             position_ids = attention_mask.long().cumsum(-1) - 1
-            position_ids.masked_fill_(attention_mask == 0, 1)
+            position_ids.masked_fill_(attention_mask == 0, 0)
```
zucchini-nlp, line 657: doesn't make diff which value we use, because the token is masked anyway. Using `0` makes more sense because when the seq has only one `unmasked` token, we are getting position ids with max value of `1`, not `0`

src/transformers/generation/utils.py
```diff
@@ -666,7 +667,7 @@ def prepare_inputs_for_generation(
                         if model_inputs.get("inputs_embeds") is not None
                         else model_inputs[input_ids_key].shape[1]
                     )
-                    model_input = model_input[:, -current_input_length:]
+                    model_input = model_input[..., -current_input_length:]
                     model_input = model_input.clone(memory_format=torch.contiguous_format)
```
zucchini-nlp, line 670: 3D positions support

src/transformers/models/colqwen2/modeling_colqwen2.py
```diff
@@ -160,13 +160,6 @@ def forward(
         )
         return_dict = return_dict if return_dict is not None else self.config.use_return_dict
 
-        position_ids, rope_deltas = self.vlm.model.get_rope_index(
-            input_ids=input_ids,
-            image_grid_thw=image_grid_thw,
-            video_grid_thw=None,
-            attention_mask=attention_mask,
-        )
-
```
zucchini-nlp, line 169: why we called it here, no idea. Better to let `self.vlm` handle everything

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
```diff
@@ -1176,95 +1171,66 @@ def get_rope_index(
...
-                    .view(1, 1, -1)
-                    .expand(3, input_ids.shape[0], -1)
-                )
-                mrope_position_deltas = torch.zeros(
-                    [input_ids.shape[0], 1],
-                    device=input_ids.device,
-                    dtype=input_ids.dtype,
-                )
-
-            return position_ids, mrope_position_deltas
```
zucchini-nlp, line 1269: same thing as it was for all `get_rope_index`, just deleted this part

src/transformers/generation/candidate_generator.py
```diff
@@ -1294,6 +1300,32 @@ def _prepare_attention_mask(model_kwargs: dict[str, Any], new_length: int, is_en
...
+        return model_kwargs
+
+    positions = model_kwargs[position_key]
+    position_length_diff = new_length - positions.shape[-1]
+
+    if position_length_diff < 0:
+        model_kwargs[position_key] = positions[:, :position_length_diff]
+    elif position_length_diff > 0:
+        # Works for 2D and 3D position tensors. NOTE; we should be able to rely only on the last position id
+        # in real inference. Test cases though have arbitary positions because the mask is created randomly
```
vasqu: Yup, I also added to the padding-free test because of exactly this. It could explain why these were more susceptible to being flaky

src/transformers/generation/candidate_generator.py
```diff
@@ -1294,6 +1300,32 @@ def _prepare_attention_mask(model_kwargs: dict[str, Any], new_length: int, is_en
...
+        max_position_ids = positions.max(-1, keepdim=True).values + 1
+        next_position_ids = max_position_ids.repeat(*(*[1] * (max_position_ids.ndim - 1), position_length_diff))
```
```suggestion
        next_position_ids = max_position_ids.repeat(*([1] * (max_position_ids.ndim - 1), position_length_diff))
```
Do we need that?

src/transformers/generation/candidate_generator.py
vasqu: Would it make sense to simplify this here and only expand on this via monkey patching in our test suite? This likely has some perf impact as we need to do quite a lot of tensor constructions

src/transformers/generation/utils.py
```diff
@@ -666,7 +667,7 @@ def prepare_inputs_for_generation(
                         if model_inputs.get("inputs_embeds") is not None
                         else model_inputs[input_ids_key].shape[1]
                     )
-                    model_input = model_input[:, -current_input_length:]
+                    model_input = model_input[..., -current_input_length:]
```
vasqu, line 671: Necessary change or just your preference?

src/transformers/generation/utils.py
```diff
@@ -1024,28 +1050,27 @@ def _update_model_kwargs_for_generation(
...
-                decoder_attention_mask = model_kwargs["decoder_attention_mask"]
-                model_kwargs["decoder_attention_mask"] = torch.cat(
-                    [decoder_attention_mask, decoder_attention_mask.new_ones((decoder_attention_mask.shape[0], 1))],
-                    dim=-1,
-                )
+        position_ids_key = "position_ids" if not is_encoder_decoder else "decoder_position_ids"
+        if (position_ids := model_kwargs.get(position_ids_key)) is not None:
+            next_position_ids = position_ids.max(-1, keepdim=True).values + num_new_tokens
+            if not use_cache:
+                next_position_ids = torch.cat([position_ids, next_position_ids], dim=-1)
```
vasqu, line 1065: Why do we need this? Given a rope model for example, won't we always need the full set of positions (cache + current token) or does this not work for other models?

src/transformers/generation/utils.py
```diff
@@ -2536,6 +2560,11 @@ def generate(
             if model_input_name == "input_ids" and len(model_kwargs["attention_mask"].shape) > 2:
                 raise ValueError("`attention_mask` passed to `generate` must be 2D.")
 
+        kwargs_has_position_ids = model_kwargs.get("position_ids", None) is not None
```
vasqu, line 2564: Nit: I like the `"position_ids" in model_kwargs` more

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
```diff
@@ -1117,6 +1111,7 @@ def get_rope_index(
         video_grid_thw: torch.LongTensor | None = None,
         attention_mask: torch.Tensor | None = None,
         mm_token_type_ids: torch.IntTensor | None = None,
+        **kwargs,
```
vasqu, line 1116: Any reason we have kwargs now?

src/transformers/models/ernie4_5_vl_moe/modular_ernie4_5_vl_moe.py
```diff
@@ -1302,6 +1207,45 @@ def get_image_features(
         image_outputs.pooler_output = image_embeds
         return image_outputs
 
+    def compute_3d_position_ids(
```
vasqu, line 1213: Could we not inherit from `Qwen2_5_VLModel`? Or is there something specific, let's avoid rewriting where possible

src/transformers/models/gpt_neo/modeling_gpt_neo.py
```diff
@@ -719,6 +719,7 @@ def forward(
             output_hidden_states=output_hidden_states,
             return_dict=return_dict,
             cache_position=cache_position,
+            **kwargs,
```
vasqu, line 588: Where did this come from :D

tests/generation/test_utils.py
```diff
@@ -2234,6 +2234,50 @@ def test_forward_with_logits_to_keep(self):
...
+        for model_class in self.all_generative_model_classes:
+            config, inputs_dict = self.prepare_config_and_inputs_for_generate()
+            model = model_class(config).to(torch_device).eval()
+            model_forward_args = inspect.signature(model.forward).parameters
+
+            if "position_ids" not in model_forward_args or "input_ids" not in inputs_dict:
+                self.skipTest("This model doesn't use `position_ids`")
+
+            if config.is_encoder_decoder:
+                self.skipTest("This model doesn't prepare `position_ids` in generate")
```
vasqu, line 2280: It does tho, no? At least decoder position ids?

tests/generation/test_utils.py
```diff
@@ -2234,6 +2234,50 @@ def test_forward_with_logits_to_keep(self):
...
+                position_ids = torch.arange(0, seq_length, dtype=torch.long, device=torch_device)
+                position_ids = position_ids.unsqueeze(0)
+
+            out_w_positions = model.generate(
+                **inputs_dict, position_ids=position_ids, max_new_tokens=5, use_cache=True, do_sample=False
+            )
+
+            # The two sets of generated sequences must match, if generate can infer position ids correctly
+            # and can continue adding new ids to the already passed position ids
+            self.assertListEqual(out_wo_positions.tolist(), out_w_positions.tolist())
```
vasqu, line 2312: Nice test!

vasqu (commented):
Just left a few questions / nits, I have a feeling we can use modular a tad more re `compute_3d_position_ids`?

vasqu:
run slow might have been broken so better to rerun after merging with main :eyes:

src/transformers/generation/candidate_generator.py
zucchini-nlp: we can try to re-build an attn mask in tests, tho I'm not sure if it will uncover more issues with test suite. I'll check

src/transformers/generation/candidate_generator.py
zucchini-nlp: yes, we are unpacking the tuple of ones so it can be concated with `position_length_diff`. If the above suggestion of yours works, I'll delete it

Also don't like this pattern

src/transformers/generation/utils.py
zucchini-nlp: necessary, 3D positions need to be cropped on last dim, i.e. seq length. otherwise we're cropping them on 2nd dim, which is batch size

src/transformers/generation/utils.py
zucchini-nlp: in cache we need only the correct queries' position ids because the k/v is already cached after RoPE. So we have to get the next 1 position only

src/transformers/generation/utils.py
zucchini-nlp: necessary in case `model_kwargs.position_ids == None`. We don't technically prohibit passing it as `None`

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
zucchini-nlp: oh this were added to allow modular copy of `_prepare_position_ids_for_generate` across all 3D rope models. Some models have specific kwargs, and I didn't want to override it only because of that

src/transformers/models/ernie4_5_vl_moe/modular_ernie4_5_vl_moe.py
zucchini-nlp: Ernie uses `mm_token_type_ids` but Qwen2-5VL has `second_grid_ts`. We can do it if we hide extra kwargs as `**kwargs`, which basically will look like the above comment

src/transformers/models/gpt_neo/modeling_gpt_neo.py
zucchini-nlp: prob some tests were failing lol, anyway we need to pass `kwargs` down the stream :)

tests/generation/test_utils.py
zucchini-nlp: it actually doesn't. Generate doesn't prepare positions/attn for encoder decoder models because we prepare a `causal`-format input which might not match encoder expectation. Also we don't prepare decoder position ids because there's no point, decoder ids are usually empty before generation starts

src/transformers/generation/candidate_generator.py
vasqu: Yea no worries, might be a bit out of scope for this PR but this feels like we are adjust for the test suite for something where the test suite should be adjusted

src/transformers/generation/candidate_generator.py
vasqu: I just think that one extra * is not necessary, but yea just have to try if it works

src/transformers/generation/utils.py
vasqu: Want to add a small expected shapes above as comment?

src/transformers/generation/utils.py
vasqu: Argh, I'm always getting confused around what's kept full length but 100% you're right.

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
vasqu: Gotcha

src/transformers/models/ernie4_5_vl_moe/modular_ernie4_5_vl_moe.py
vasqu: Ah, yea that's a good point. On another note, do we want to change the other VLMs to use mm token type ids here? Iiirc, it's much faster(?)

tests/generation/test_utils.py
vasqu: > Also we don't prepare decoder position ids because there's no point, decoder ids are usually empty before generation starts

Yea, I think that's the point. Makes sense

src/transformers/models/ernie4_5_vl_moe/modular_ernie4_5_vl_moe.py
zucchini-nlp: we do! That is my next PR after this one is merged. I have some stuff locally :)

src/transformers/generation/candidate_generator.py
zucchini-nlp: just verified, the second one is needed for 3d positions. Tho we can remove the first *, `torch.repeat` can work with it apparently

vasqu: The less *, the happier I am :D wouldn't die on that hill but it's nicer

vasqu (approved):
Forgot to approve, can you also check if qwen3_5(_moe) have inherited as expected? They are essentially coming from qwen3 vl

zucchini-nlp:
Oke, will run a few slow tests and merge

zucchini-nlp:
run-slow: colqwen2, ernie4_5_vl_moe, gemma3, glm46v, glm4v, glm4v_moe, glm_image, glm_ocr, gpt_neo, idefics, paddleocr_vl, qwen2_5_vl, qwen2_vl, qwen3_vl, qwen3_vl_moe, reformer

Cyrilvallez:
Hey! Sorry I'm a bit late to the party! Instead of having a dedicated `_prepare_position_ids_for_generation` method everywhere for a lot of models, wdyt about simply giving the responsability of creating position_ids to the model during the first forward? Similar to what we do with caches. That way, we don't have to maintain such additional methods, and generate reuse the prepared position_ids after first forward
@zucchini-nlp @vasqu

zucchini-nlp:
If we make each model compute their position ids in forward (which already happens now in not-so-correct way), we can't just build upon it by incrementing to the next position. Position ids aren't returned from model like cache so we have to start returning them from forward to be able to re-use. Otherwise we just have to let each model re-compute positions from scratch every time, basically what happens now 

Actually, I thought at first to make each `BaseModel` have their own method to `compute_position_ids` and generation simply calls `base_model.compute_position_ids`. Yet all models compute it the same way except for qwen-vl and paligemma, so why not just get it in generation mixin and override is special cases

zucchini-nlp:
> [7 new failed tests from this PR](https://huggingface.co/datasets/hf-internal-testing/transformers_pr_ci/raw/1106b1d0667e4aa40e9741067b43432eecc52680/2026-02-10/runs/28607-21866836465/ci_results_run_models_gpu/new_failures_with_bad_commit.json) 😭

Ah that was the issue with padding side, supposed to be "left". We don't recompute positions every time thus it doesn't work well with right padding

vasqu:
Updated on the hub! So ernie should behave somewhat normally now, no idea why the default padding changed tbh

zucchini-nlp:
run-slow: colqwen2, ernie4_5_vl_moe, qwen2_5_vl

zucchini-nlp:
run-slow: ernie4_5_vl_moe, qwen2_5_vl

Cyrilvallez:
>Position ids aren't returned from model like cache

Arhhh, you're right 🥲 Nevermind then!

PR #43735, opened 2026-02-04
Chase TokenizersBackend issue

Rocketknight1:
One of the recurring issues we see in the CI is tokenizers occasionally being initialized as instances of `TokenizersBackend` rather than the actual tokenizer class. The error crops up all over, and is annoyingly hard to track down because I can't reproduce it locally. This PR is an attempt to flush it out and run some debug so I can see what's happening in the CI itself.


Rocketknight1:
We finally tracked it down and the fix is here: https://github.com/huggingface/transformers/pull/43791

PR #43737, opened 2026-02-04
Use associative scan HOP mamba recurrentgemma

riccardofelluga:
# What does this PR do?

Adds support for PyTorch's `associative_scan` higher-order op in Mamba and RecurrentGemma models, enabling parallel computation of the recurrent scan during `torch.compile()`.

- Uses `associative_scan` when available and compiling (`is_torchdynamo_compiling()`)
- Falls back to sequential loop for decoding (with cache) or when unavailable
- Selects optimal `combine_mode` based on device type (pointwise for CUDA/XPU, generic otherwise)

**Motivation:** The sequential scan loop is a compilation bottleneck. Using `associative_scan` allows for faster compile time and better perf.

## Before submitting


vasqu:
Heya that's a nice optimization, wasn't aware that torch had this op :D feel free to ping me when the PR is ready

riccardofelluga:
I managed to benchmark the models and here are the results for the forward pass:

**alpindale/recurrentgemma-9b-it BEFORE the change**

| seq_len | eager (ms) | compiled (ms) | speedup | compile time (s) |
|--------:|-----------:|--------------:|--------:|-----------------:|
| 128 | 105.0 | 70.8 | 1.48x | 230.6 |
| 512 | 353.8 | 1164.7 | 0.30x | 991.0 |
| 1024 | 683.9 | 4572.0 | 0.15x | 2227.6 |

**alpindale/recurrentgemma-9b-it AFTER**

| seq_len | eager (ms) | compiled (ms) | speedup | compile time (s) |
|--------:|-----------:|--------------:|--------:|-----------------:|
| 128 | 104.0 | 29.1 | 3.57x | 25.5 |
| 512 | 349.6 | 98.7 | 3.54x | 24.7 |
| 1024 | 690.9 | 194.6 | 3.55x | 25.8 |


**state-spaces/mamba-2.8b-hf BEFORE the change**

| seq_len | eager (ms) | compiled (ms) | speedup | compile time (s) |
|--------:|-----------:|--------------:|--------:|-----------------:|
| 128 | 240.5 | 113.5 | 2.12x | 1075.6 |
| 512 | 920.9 | Time out | Time out | Time out |
| 1024 | 1851.3 | Time out | Time out | Time out |

**state-spaces/mamba-2.8b-hf AFTER**

| seq_len | eager (ms) | compiled (ms) | speedup | compile time (s) |
|--------:|-----------:|--------------:|--------:|-----------------:|
| 128 | 247.7 | 14.6 | 16.97x | 24.2 |
| 512 | 927.5 | 53.0 | 17.52x | 23.8 |
| 1024 | 1840.1 | 102.7 | 17.91x | 24.3 |

Not only the run time is faster but also the compile time does not depend anymore on the sequence length.

Thanks @vasqu, I've added the tests and I think now it's ready for review

src/transformers/models/recurrent_gemma/modeling_recurrent_gemma.py
```diff
@@ -31,12 +31,18 @@
...
 from .configuration_recurrent_gemma import RecurrentGemmaConfig
 
 
 logger = logging.get_logger(__name__)
 _MAX_SQRT_GRADIENT = 1000.0
 
+try:
+    from torch._higher_order_ops.associative_scan import associative_scan
+except ImportError:
+    associative_scan = None
```
ArthurZucker: can we only try if the correct torch version is available?

ArthurZucker (approved):
Great addition. It's simple enough to not necessarily have to hide it imo! Thanks 🤗

src/transformers/models/recurrent_gemma/modeling_recurrent_gemma.py
riccardofelluga: Sure! I've added the check and tested it on that version.

src/transformers/models/mamba/modeling_mamba.py
```diff
@@ -404,12 +410,27 @@ def slow_forward(self, input_states, cache_params: MambaCache | None=None, cache
...
             scan_output = scan_output * self.act(gate)
         else:
-            scan_outputs = []
-            for i in range(seq_len):
-                ssm_state = discrete_A[:, :, i, :] * ssm_state + deltaB_u[:, :, i, :]      # [batch, intermediate_size, ssm_state]
-                scan_output = torch.matmul(ssm_state.to(dtype), C[:, i, :].unsqueeze(-1))  # [batch, intermediate_size, 1]
-                scan_outputs.append(scan_output[:, :, 0])
-            scan_output = torch.stack(scan_outputs, dim=-1)                                # [batch, intermediate_size, seq_len]
+            # Use associative_scan for parallel computation when available
+            if associative_scan is not None and cache_params is None and is_torchdynamo_compiling():
```
vasqu: We have an `is_tracing` helper - let's use that one as it's broader

vasqu (approved):
Really clean, just a small nit

src/transformers/models/mamba/modeling_mamba.py
riccardofelluga: Thanks! Updated

vasqu:
@bot /repo

ArthurZucker (commented):
Forgot to ask but are the results exactly the same? 
Also this is not an opt-out thing, if res are impacted we should allow users to opt-out!

riccardofelluga:
> Forgot to ask but are the results exactly the same? Also this is not an opt-out thing, if res are impacted we should allow users to opt-out!

@ArthurZucker in the test it checks for the results to be the same between eager and compiled(with associative scan). However, it does not test the compiled for-loop vs the compiled scan so I'll add a flag to make it opt out just in case

riccardofelluga:
To add the opt-out, I've initially modified the config but since it applies to other models I've changed it to be specific to these two. Let me know what you guys think if this is an acceptable solution

ArthurZucker (commented):
sounds good, I think you need to update both configs with default `False` probably for both. Documenting is one of the best way we can tell users to leverage that. We can also default to `True` if results do match (which seems to be the case) but the reasoning needs to be explicit

ArthurZucker:
just waiting for you to fix the CI and we can merge!

riccardofelluga:
> just waiting for you to fix the CI and we can merge!

Thanks @ArthurZucker! Sorry I got stuck a little with the code generation. Now it should be good to go :)

vasqu (approved):
Just checking with run-slow in a second but LGTM, can you resolve the merge conflict as well?

vasqu:
run-slow: falcon_mamba, mamba, recurrent_gemma

vasqu:
run-slow: falcon_mamba, mamba, recurrent_gemma

vasqu:
run-slow: recurrent_gemma

vasqu:
run-slow: recurrent_gemma

vasqu:
run-slow: recurrent_gemma

vasqu:
run-slow: falcon_mamba, mamba, recurrent_gemma

vasqu:
run-slow: recurrent_gemma

vasqu:
run-slow: falcon_mamba, mamba, recurrent_gemma

vasqu:
run-slow: recurrent_gemma

vasqu:
@riccardofelluga are you fine with leaving recurrent gemma out for the time being? It seems our CI really doesn't like recurrent gemma and it might be deprecated soon either way. So I'd like to keep it for the mamba improvements alone

riccardofelluga:
>  are you fine with leaving recurrent gemma out for the time being? It seems our CI really doesn't like recurrent gemma and it might be deprecated soon either way. So I'd like to keep it for the mamba improvements alone

@vasqu Sure! that is ok with me, thanks for having took the time to revert 🙏

vasqu:
Perfect, merging then in a sec

PR #43738, opened 2026-02-04
[Trainer] Move optimizer cls init to trainer_optimizer.py

SunMarc:
# What does this PR do?

This PR do the following: 

- Created src/transformers/trainer_optimizer.py with:
    - Helper functions for each optimizer category
- Updated `Trainer.get_optimizer_cls_and_kwargs`
- Simplify tests


SunMarc:
cc @winglian

src/transformers/trainer_optimizer.py
```diff
@@ -0,0 +1,578 @@
...
+    OptimizerNames.ADAMW_TORCH_8BIT,
+]
+
+_SCHEDULE_FREE_OPTIMIZERS = [
+    OptimizerNames.SCHEDULE_FREE_RADAM,
+    OptimizerNames.SCHEDULE_FREE_ADAMW,
+    OptimizerNames.SCHEDULE_FREE_SGD,
+]
+
+_OPTIMIZER_HANDLERS = {
```
winglian: It might be nice to have an api like `register_optimizer_handler(name, optim_fn)`

SunMarc: Added it but I feel like we shouldn't encourage users to register their optimizers using the handler but just pass the `optimizer_cls_and_kwargs` or overwrite `create_optimizer`. Maybe we can add that later on ? 
Removed for now

src/transformers/trainer.py
```diff
@@ -1221,22 +1215,28 @@ def create_optimizer(self):
...
 
-            # Overwrite `params` in case it's created by `get_optimizer_cls_and_kwargs`
-            # e.g. for GaLore optimizer.
-            if "params" in optimizer_kwargs:
-                optimizer_grouped_parameters = optimizer_kwargs.pop("params")
+            # Check if this is a factory (for complex optimizers like Muon, Dion)
+            # Factories are instantiated first, then called with (opt_model, **kwargs)
+            if is_optimizer_factory(optimizer_cls):
+                self.optimizer = optimizer_cls()(opt_model, **optimizer_kwargs)
+            else:
```
SunMarc, line 1181: is this what you meant @winglian ?

SunMarc:
@bot /style

src/transformers/trainer_optimizer.py
```diff
@@ -0,0 +1,608 @@
...
+    Args:
+        optimizer_cls_or_factory: The first element returned by an optimizer handler.
+
+    Returns:
+        `bool`: True if it's not an Optimizer class (i.e., likely a factory), False if it's an Optimizer class.
+    """
+    # If it's a class that's a subclass of torch.optim.Optimizer, it's not a factory
+    if isinstance(optimizer_cls_or_factory, type) and issubclass(optimizer_cls_or_factory, torch.optim.Optimizer):
+        return False
+    return True
```
winglian: Should we have an explicit check that that the class has a call method and raise a value error otherwise? 

issubclass(optimizer_cls_or_factory, collections.abc.Callable)

kashif (approved):

SunMarc:
@bot /style

PR #43744, opened 2026-02-04
Minor changes trainer

SunMarc:
# What does this PR do?

This PR updates a few minor things from trainer. Some of them are breaking but I think it should be safe to do as I don't think anyone is subclassing them or using them separately. 

Breaking: 
- move `propagate_args_to_deepspeed` to standalone function in `integrations.deepspeed`
- move `_fsdp_qlora_plugin_updates` to standalone function in `integrations.fsdp` + rename to `update_fsdp_plugin_peft`
- move `is_attention_mask_causal` to standalone function in `trainer_pt_utils.py`
- move `_nested_gather` to standalone function in `trainer_pt_utils.py`
- remove `_add_sm_patterns_to_gitignore` -> not used at all 
- move `_align_special_tokens`  to standalone function in `trainer_pt_utils.py`
- move `deepspeed_sp_compute_loss` to `integrations.deepspeed` 
- move `deepspeed_sp_compute_loss` to `integrations.deepspeed` 
- move `_save_tpu` to `save_tpu_checkpoint` in `integrations.tpu` file 
- added `wrap_model_xla_fsdp` to move logic in integrations.fsdp` file

Not breaking: 
- move `get_fsdp_ckpt_kwargs` to `integrations.fsdp` 
- move `safe_globals` to `trainer_utils.py`
- move `_get_learning_rate` in `Trainer` as this method is actually used and not only in examples.


qgallouedec:
Thanks, I'll review asap

SunMarc:
cc @winglian

src/transformers/trainer_pt_utils.py
```diff
@@ -932,6 +974,39 @@ def save_state(self):
     self.state.save_to_json(path)
 
 
+def get_num_trainable_parameters(self) -> int:
```
winglian, line 974: might be good to call out that these are helper functions for downstream users as they aren't used except in tests currently.

SunMarc: Yeah, I did put a comment above `log_metrics`, will do something clearer

SunMarc: done !

SunMarc:
@bot /style

src/transformers/trainer.py
```diff
@@ -297,7 +280,7 @@ class Trainer:
             The function to use to form a batch from a list of elements of `train_dataset` or `eval_dataset`. Will
             default to [`default_data_collator`] if no `processing_class` is provided, an instance of
             [`DataCollatorWithPadding`] otherwise if the processing_class is a feature extractor or tokenizer.
-        train_dataset (Union[`torch.utils.data.Dataset`, `torch.utils.data.IterableDataset`, `datasets.Dataset`], *optional*):
+        train_dataset (`torch.utils.data.Dataset` | `torch.utils.data.IterableDataset` | `datasets.Dataset`, *optional*):
```
qgallouedec, line 278: NIT and not for this PR

I find it more readable to use "or" instead of `|` in the docstring.

```python
- train_dataset (`torch.utils.data.Dataset` | `torch.utils.data.IterableDataset` | `datasets.Dataset`, *optional*):
+ train_dataset (`torch.utils.data.Dataset` or `torch.utils.data.IterableDataset` or `datasets.Dataset`, *optional*):
```

Image

src/transformers/trainer.py
```diff
@@ -2029,7 +1868,7 @@ def patched_optimizer_step(optimizer, barrier=False, optimizer_args={}):
     def train(
         self,
         resume_from_checkpoint: str | bool | None = None,
-        trial: Union["optuna.Trial", dict[str, Any], None] = None,
+        trial: Union["optuna.Trial", dict[str, Any]] | None = None,
```
qgallouedec:
```suggestion
        trial: "optuna.Trial | dict[str, Any]] | None" = None,
```

qgallouedec (approved):
this looks good!

winglian (approved):

src/transformers/trainer_utils.py
```diff
@@ -71,6 +71,9 @@ def _is_peft_model(model):
     return False
 
 
+logger = logging.get_logger(__name__)
+
+
```
albertvillanova: The logger is already defined in line 57:
```suggestion
```

SunMarc: thanks for spotting !

src/transformers/trainer.py
SunMarc: Noted ! I will do a PR in order to fix all docstrings + typing later on !

src/transformers/trainer.py
SunMarc: Thx for the trick, I tried earlier with just putting the string between `optuna.Trial` and the CI was complaining ;)

ArthurZucker (approved):
I am down to keep aliases for now + deprecate (but its up to you if you can adds lots of coms about it) 

Trying to refrain from too many breaks from v5 already

SunMarc:
Can you run the TRL CI on this PR @qgallouedec ? Also @winglian, is there a way to run the axolotl CI just like we do for TRL ? One of my next step also would be to create nice tests like TRL so that we don't need to rely on TRL to catch regressions. 

> I am down to keep aliases for now + deprecate (but its up to you if you can adds lots of coms about it)

I think for these, it shouldn't be too breaking. There are other more sensible arg that will indeed require to deprecate if we want to ensure a smooth transition in a future PR. If any of those are too breaking, happy to put it back and do a proper deprecation cycle.

qgallouedec:
https://github.com/huggingface/trl/actions/runs/21873008073

tomaarsen:
This resulted in a regression in Sentence Transformers due to the removal of `_nested_gather`. I'll try to push out a release soon to resolve it. See https://github.com/huggingface/sentence-transformers/pull/3664 for more details.

- Tom Aarsen

PR #43750, opened 2026-02-05
enable tp for benchmark

sywangyi:
enable tp in benchmark_v2, to ensure large model could run.


sywangyi:
@remi-or , pls help review, thx very much.

sywangyi:
hi, @remi-or any thought to enable tp, cp in the benchmark tool?

benchmark_v2/framework/benchmark_config.py
```diff
@@ -142,6 +143,9 @@ def hash(self) -> str:
 
     def infer_name(self, compact: bool = True) -> str:
         """Infer a human-readable name for the benchmark config, either compact or verbose."""
+        tp_str = "no_tp"
```
remi-or: Can be reformatted as a one-liner

benchmark_v2/framework/benchmark_config.py
```diff
@@ -142,6 +143,9 @@ def hash(self) -> str:
...
+        if self.tp_plan is not None:
+            tp_str = f"tp_{self.tp_plan}"
```
remi-or: Could we come up with a more summarized way to include the tp_plan? It seems like this is going to produce a very long name... Maybe "TP" in the compact version and a summarized tp_plan in the non-compact version.

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -206,15 +212,18 @@ def setup_benchmark(self, model_id: str, config: BenchmarkConfig) -> None:
         self.logger.debug(f"Loading model {model_id} on device {config.device}...")
         dtype = getattr(torch, config.dtype.removeprefix("torch."))
         use_kernels = config.kernelize and kernelize is not None and Mode is not None
+        device_map = None if config.tp_plan is not None else config.device
```
remi-or: I think this would be more readable the other way around! 
device_map = config.device if config.tp_plan is None else None

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -401,6 +412,8 @@ def run_benchmarks(
 
     def save_results(self, model_name: str, results: dict, timestamp: str = "", summarized: bool = True) -> str:
         """Save benchmark results to JSON file."""
+        if not self._is_primary_process():
+            return ""
```
remi-or, line 416: I am not sure if the empty string handled in the outet scope?

benchmark_v2/framework/benchmark_runner.py
```diff
@@ -169,6 +169,12 @@ def cleanup(self) -> None:
         self.model = None
         flush_memory()
 
+    @staticmethod
```
remi-or, line 172: nice!

remi-or (commented):
Some nits, but otherwise lgtm. Have you tried out the benchmarking with TP? And if so, how were the results? I am curious what applications you are targeting this for 🙂 !

sywangyi:
> Some nits, but otherwise lgtm. Have you tried out the benchmarking with TP? And if so, how were the results? I am curious what applications you are targeting this for 🙂 !

actually I would like to leverage this benchmark tool in xpu to broader model, so I need to run bigger model(like moe serial). one card could not run such model for memory limitation. so I need to run with ep and tp with multiple-cards. also tp, ep support in kernels path is only in our radar.

benchmark_v2/framework/benchmark_runner.py
sywangyi: the return string is not used in outer scope

remi-or:
Ok, this looks good! I would like to test n my end before merging, will do so soon. My question was: did you manage to run the benchmarker in a distributed setting on your end? Or is this a small change needed for that but not enough to enable the feature? Thanks

sywangyi:
yes, I test by my side using torchrun --nproc-per-node 2 run_benchmarks.py --enable-tp, enough to enable the feature.

remi-or (approved):
LGTM! Thanks for your patience.

PR #43753, opened 2026-02-05
Update KERNELS_MIN_VERSION to 0.10.2 to be the same as setup.py

cyyever:
# What does this PR do?
Update KERNELS_MIN_VERSION to 0.10.2 to be the same as setup.py


Rocketknight1:
Good point - cc @arthurzucker @cyrilvallez @vasqu should we have two sources of truth for these, or should we generate these constants from `setup.py`?

ArthurZucker (approved):
@Rocketknight1 we should put this check in `make fix-repo` imo!

Rocketknight1:
@cyyever do you want to take that? I can make the PR to update `make fix-repo` if you don't want to

cyyever:
@Rocketknight1 You do? Either add a commit to this PR or open a new PR?

Rocketknight1:
@cyyever sure, I'll add it to this PR

Cyrilvallez:
IMO no need for `make fix-repo`: if the min version is lower than setup.py, let's simply delete that variablle and update `is_kernels_available` to never use it

PR #43754, opened 2026-02-05
Remove SDPA workarounds for torch 2.4+

cyyever:
# What does this PR do?
https://github.com/pytorch/pytorch/issues/112577 was fixed in torch 2.2. Let's remove the workarounds and see what happen in the tests.


vasqu (approved):
Should have been removed earlier when we moved to 2.2.; good catch

PR #43758, opened 2026-02-05
stablize the expect output for eomt_dinov3 model test

kaixuanliu:
@ydshieh can you help review? Thx!

### DIFF
--- a/tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
+++ b/tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
@@ -18,7 +18,14 @@ import unittest
 import requests
 
 from transformers import AutoImageProcessor, EomtDinov3Config, EomtDinov3ForUniversalSegmentation, pipeline
-from transformers.testing_utils import require_torch, require_torch_accelerator, slow, torch_device
+from transformers.testing_utils import (
+    Expectations,
+    require_deterministic_for_xpu,
+    require_torch,
+    require_torch_accelerator,
+    slow,
+    torch_device,
+)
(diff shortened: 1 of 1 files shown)


tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
```diff
@@ -258,23 +266,55 @@ def test_inference_bf16(self):
...
-            [ 2.5781,  5.3438,  6.2188],
-            [ 3.7656,  7.1562,  8.1875]
-        ], device=model.device)
+        class_expectations = Expectations(
+            {
+                (None, None): [
+                    [-0.3262, -5.6875, -0.7539],
+                    [ 0.0474, -6.8438, -2.1406],
+                    [-1.4219, -6.0312, -5.4688]
+                ],
```
vasqu: Do we need this?

vasqu (approved):
Thanks, the other tests work out of the box?

tests/models/eomt_dinov3/test_modeling_eomt_dinov3.py
kaixuanliu: I am not sure. This is the original version, I don't have other platforms to check.

vasqu: Let's remove it for now then, if we encounter other platforms, I'd rather we have to explicitly recheck it

kaixuanliu: Done

vasqu:
Merging when CI has run (and turned green) :hugs:

PR #43762, opened 2026-02-05
 Widen match condition for `_can_record_outputs`

molbap:
In case of module reloading, we currently lose tracking hooks for hidden states and attentions . Widening the matching condition a bit.

Should fix https://github.com/huggingface/transformers/issues/43761, also mentioned in https://github.com/linkedin/Liger-Kernel/pull/1061 

Reproducer for CLIP:

```python
import importlib, torch
from transformers import CLIPVisionConfig, CLIPVisionModel
import transformers.models.clip.modeling_clip as modeling_clip

config = CLIPVisionConfig(hidden_size=32, intermediate_size=64, num_hidden_layers=2, num_attention_heads=4, image_size=30, patch_size=10)
pixel_values = torch.randn(1, 3, 30, 30)

m1 = CLIPVisionModel(config)
o1 = m1(pixel_values=pixel_values, output_hidden_states=True)
print("before reload:", o1.hidden_states is None)

importlib.reload(modeling_clip)

m2 = CLIPVisionModel(config)
o2 = m2(pixel_values=pixel_values, output_hidden_states=True)
print("after reload:", o2.hidden_states is None)
```

Second reload will currently fail, and this is a fix proposal.

Added a test as well (specific to CLIP, likely enough)

### DIFF
--- a/src/transformers/utils/generic.py
+++ b/src/transformers/utils/generic.py
@@ -997,9 +997,17 @@ def check_model_inputs(func=None, *, tie_last_hidden_states=True):
                 for name, module in self.named_modules():
                     for key, specs in capture_tasks:
                         # The second check is for multimodals where only backbone layer suffix is available
-                        if (specs.target_class is not None and isinstance(module, specs.target_class)) or (
-                            specs.class_name is not None and name.endswith(specs.class_name)
-                        ):
+                        if specs.target_class is not None:
+                            target_cls = specs.target_class
+                            module_cls = module.__class__
+                            matches_target_class = isinstance(module, target_cls) or (
+                                module_cls.__name__ == target_cls.__name__
+                                and module_cls.__module__ == target_cls.__module__
+                            )  # this is to make sure we attach hooks in case of module reloading
+                        else:
+                            matches_target_class = False
+
+                        if matches_target_class or (specs.class_name is not None and name.endswith(specs.class_name)):
                             if specs.layer_name is not None and specs.layer_name not in name:
                                 continue
                             # Monkey patch forward
--- a/tests/utils/test_modeling_utils.py
+++ b/tests/utils/test_modeling_utils.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import copy
 import glob
+import importlib
 import json
 import os
 import os.path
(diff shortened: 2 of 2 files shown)


src/transformers/utils/generic.py
```diff
@@ -997,9 +997,17 @@ def wrapped_forward(*args, **kwargs):
                 for name, module in self.named_modules():
                     for key, specs in capture_tasks:
                         # The second check is for multimodals where only backbone layer suffix is available
-                        if (specs.target_class is not None and isinstance(module, specs.target_class)) or (
-                            specs.class_name is not None and name.endswith(specs.class_name)
```
zucchini-nlp, line 1001: This line of check for `class_name` was added for AyaVision, but I think it's not needed anymore. At least tests don't complain 😄

tests/utils/test_modeling_utils.py
```diff
@@ -3732,3 +3736,29 @@ def test_vision_language_model(self):
...
+            image_size=30,
+            patch_size=10,
+        )
+        pixel_values = torch.randn(1, 3, 30, 30, device=torch_device)
+
+        model = CLIPVisionModel(config).to(torch_device)
+        outputs = model(pixel_values=pixel_values, output_hidden_states=True)
+        self.assertIsNotNone(outputs.hidden_states)
+
+        importlib.reload(modeling_clip)
```
zucchini-nlp, line 3763: maybe we can move it under `test_modeling_common.py`, for make smth similar to this test and reload the `module` at some point before checking if `config.output_attentions` works?

https://github.com/huggingface/transformers/blob/ace7c370145a4a4861383672daa7095a0c8e43fc/tests/test_modeling_common.py#L1754-L1757

molbap: yeah I hesitated, I didn't want to add yet another test for 400+ models :face_with_peeking_eye: but you may be right

src/transformers/utils/generic.py
molbap: aah :D makes sense, wdyt of the added check though? not sure how to handle reloading more elegantly

zucchini-nlp: I think it's the best we can do with `importlib.reload` and shouldn't backfire with false positives. Lemme approve, forgot to

tests/utils/test_modeling_utils.py
zucchini-nlp: haha right!

zucchini-nlp (approved):

ArthurZucker (approved):
great catch actually! 🤗

molbap:
I want to think that great minds think alike :eyes: @Cyrilvallez 's PR solves more problems with `check_model_inputs` https://github.com/huggingface/transformers/pull/43765

PR #43764, opened 2026-02-05
:rotating_light: Modify ModernBERT's default attention implementation to stop using FA

YangKai0616:
Background: PR https://github.com/huggingface/transformers/pull/43672

This PR modifies the default attention implementation in `modernbert`, switching away from `FlashAttention`. It also restores the related tests that were previously overridden.

@vasqu , please help review again, thank you very much!


src/transformers/models/pe_audio/modular_pe_audio.py
```diff
@@ -220,7 +220,9 @@ def forward(
         text_audio_embeds = self.text_audio_head(text_audio_embeds)
 
         logits_audio_text = audio_embeds @ text_audio_embeds.T
-        logits_audio_text = logits_audio_text * self.text_audio_logit_scale + self.text_audio_logit_bias
+        logits_audio_text = logits_audio_text * self.text_audio_logit_scale.to(
+            logits_audio_text.device
+        ) + self.text_audio_logit_bias.to(logits_audio_text.device)
```
YangKai0616, line 226: Fix this because the test `tests/models/pe_audio/test_modeling_pe_audio.py::PeAudioModelTest::test_model_parallelism` will report error `RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0!`.

vasqu (approved):
Much better, can you add :rotating_light: to the title since it is breaking (in a good way)

vasqu:
Also check modern_bert decoder, it inherited from modernbert (should be a simple removal of the function there in modular) + should we add a small note in the docs / update them for the padding free example

tests/models/modernbert_decoder/test_modeling_modernbert_decoder.py
```diff
@@ -144,7 +144,7 @@ def test_inference_causal_lm(self):
         inputs = tokenizer("Paris is the capital of", return_tensors="pt")
         with torch.no_grad():
             output = model(**inputs)[0]
-        expected_shape = torch.Size((1, 6, model.config.vocab_size))
+        expected_shape = torch.Size((1, 7, model.config.vocab_size))
```
YangKai0616, line 147: The current version of the tokens output template has `add_eos_token=True`, and the actual output here is: `['[CLS]', 'Paris', ' is', ' the', ' capital', ' of', '[SEP]']`. Therefore, `expected_shape` needs to be modified.

YangKai0616:
> Also check modern_bert decoder, it inherited from modernbert (should be a simple removal of the function there in modular) + should we add a small note in the docs / update them for the padding free example

Done.

YangKai0616:
> Much better, can you add 🚨 to the title since it is breaking (in a good way)

It seems I don’t have permission to edit the title. The `Edit` button next to the title is gone.
Could you please help me change the title? Thank you!

vasqu (approved):
Perfect, it's an important step imo. It already started to grow on a few models so this was going to be a nightmare (maintenance-wise) 

No idea why you can't modify the title :sob: I just did, no worries

Issue #42673, opened 2025-12-06
Qwen3ForCausalLM leaks VRAM if used in multiple dataloader threads


dxqb:
### System Info

torch 2.8.0
transformers==4.56.2 or ransformers==4.57.3, both tested

### Who can help?

@ArthurZucker @Cyrilvallez

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

Please see repro code below. In this test case, it runs out of VRAM within only a few steps. In the real use case it takes about 50 iterations.
I hope it's still the same root cause, but I'm not entirely sure

```import torch
import threading

from transformers import Qwen3ForCausalLM


from diffusers import ZImagePipeline

pipe = ZImagePipeline.from_pretrained(
    "Tongyi-MAI/Z-Image-Turbo",
    torch_dtype=torch.bfloat16,
)

text_encoder = pipe.text_encoder
text_encoder.to('cuda')

def run():
    tokens = torch.zeros((1, 512), device='cuda', dtype=torch.int64)
    tokens_attention_mask = torch.ones((1, 512), device='cuda')
    tokens_attention_mask[:, 200:] = 0.0

    i = 0
    while True:
        i += 1
        print(i)
        text_encoder_output = text_encoder(
            tokens,
            attention_mask=tokens_attention_mask,
            output_hidden_states=True,
            return_dict=True,
            use_cache=False,
        )


thread1 = threading.Thread(target=run)
thread2 = threading.Thread(target=run) # <--- comment this to see it working with no issues

thread1.start()
thread2.start() # <--- comment this to see it working with no issues
```

### Expected behavior

see above

Rocketknight1:
Hmmn, is this a memory leak? It feels more like running two instances simultaneously will eventually cause both of them to hit peak memory usage simultaneously and go OOM. Can you try just running a single instance over and over and checking if max VRAM utilization in torch is increasing?

dxqb:
in my real use case it was definitely a leak. vram increased slowly and by iteration ~ 50 I ran out of vram. Note that I was not keeping the output, but saving it to disk and discarding it.
when I tried to make a small test case (see above) it OOMed much faster, by step 2. Maybe my reproduction is mistaken, but the issue I wanted to report is a leak.

[...]

PR #43765, opened 2026-02-05
[core] Faster and thread-safe `check_model_inputs` implementation

Cyrilvallez:
# What does this PR do?

## The problem

Currently, `check_model_inputs` needs to iterate on all modules and monkey-patch all needed submodule's `forward` on-the-fly, before restoring them afterwards. This brings 2 big issues:

- It's NOT thread-safe (see also issue https://github.com/huggingface/transformers/issues/42673): due to the monkey patching, 2 threads running the model's forward concurrently will each try to monkey-patch the global state, resulting in undefined behavior and memory leaks in the collected outputs
- It's not efficient at all: every call to the model's `forward` need to traverse the whole model graph to monkey patch every module

Note that thread-safety is an important aspect for most simple servers, i.e. gradio spaces etc as the model then runs on different threads. So even if the `output_xxx` kwargs are somewhat niche, I believe it's an important issue that we should not overlook.

## The proposal

We can instead use forward hooks, and only activate them when needed. Coupled with `ContextVars` for thread-safety on the `collected_outputs`, this brings a much more safe, clean and efficient implementation. The idea is the following:

- install forward hooks ONCE lazily the first time we need output capture (so we only traverse the graph once) -> no monkey patching of forwards - if we never need output capture (most of the time) the hooks are NOT installed
- These hooks are dormant once installed, and are only awakened when we have the `output_xxx` kwargs
- Since the hooks need to use a global variable to collect the outputs (because we don't want to modify the signature of every module's `forward` of course), use a `ContextVar` inside the hook, so the collected outputs are thread-safe (`ContextVar` basically behaves as a thread-safe global variable and it blazingly fast)
- Since `torch.compile` does not support tracing the `get` of `ContextVar`, use a simple trick to use a simple global variable in those cases. This unfortunately means that `torch.compile` + `return_xxx` are not thread-safe, but works on single thread
- Collect outputs inside lists instead of tuples to avoid creating new objects all the time, and simply append to it

## TLDR - What is solves

|            | Need to traverse the whole graph every time | Thread-safe without compile | Thread-safe with compile |
|-------|:--------------------------------------------:|:-----------------------------:|:-------------------------:|
| Currently |    🚫   |    🚫   |   🚫   |
| This PR |    ✅   |   ✅    |    🚫   | 

## Reproduction code

<details>
<summary>memory_leak.py</summary>

```python
import threading
import time

import torch

model_id = "meta-llama/Llama-3.1-8B-Instruct"
device = 0
WARMUP = 3
TOKS = 100
THREADS = 2

model = AutoModelForCausalLM.from_pretrained(model_id, dtype=torch.bfloat16, device_map=device, attn_implementation="eager")
messages = [
    {"role": "user", "content": "What do you think about life?"},
]
inputs = tokenizer.apply_chat_template(messages, return_tensors="pt").to(device)
input_size = inputs.input_ids.shape[-1]

generation_config = dict(
    min_new_tokens=TOKS,
    max_new_tokens=TOKS,
    do_sample=False,
    eos_token_id=tokenizer.eos_token_id,
    pad_token_id=tokenizer.eos_token_id,
    output_attentions=True,
    output_hidden_states=True,
    return_dict_in_generate=True,
)

def print_memory_peak():
    peak = torch.cuda.max_memory_allocated(device) / 1024 ** 3
    print(f"Memory peak is {peak:.2f} GiB")

def run_a_few():
    t0 = time.time()
    for _ in range(WARMUP):
        output = model.generate(**inputs, **generation_config)
    dt = time.time() - t0
    print_memory_peak()

def run_indefinitely():
    while True:
        output = model.generate(**inputs, **generation_config)
        print_memory_peak()

run_a_few()


threads = []
for _ in range(THREADS):
    threads.append(threading.Thread(target=run_indefinitely))

# Start each thread
for t in threads:
    t.start()

# Wait for all threads to finish
for t in threads:
    t.join()

```

</details>

The above script shows the issue mentionned and the associated memory leak. On current main, memory keeps increasing without bounds with time. On this PR, it is solved and memory is stable.
Moreover, the output gathered are now correct compared to fully random before.

<details>
<summary>benchmark_speed.py</summary>

```python
import numpy as np
import torch
import time

model_id = "meta-llama/Llama-3.1-8B-Instruct"
device = 0
WARMUP = 10
EXPERIENCE = 1000
INPUT_SHAPE = (2, 56)


model = AutoModelForCausalLM.from_pretrained(model_id, dtype=torch.bfloat16, device_map=device, attn_implementation="eager")

torch.manual_seed(12)
input_ids = torch.randint(10, 1000, INPUT_SHAPE, device=device)

# Warmup
tot = []
for _ in range(WARMUP):
    torch.cuda.synchronize(device)
    t0 = time.time()
    with torch.no_grad():
        out = model(input_ids, return_dict=True)
    torch.cuda.synchronize(device)
    dt = time.time() - t0
    tot.append(dt)
assert out.attentions is None
print(f"Warmup took {np.mean(tot):.2e} s")

# Actual test
tot = []
for _ in range(EXPERIENCE):
    torch.cuda.synchronize(device)
    t0 = time.time()
    with torch.no_grad():
        out = model(input_ids, return_dict=True)
    torch.cuda.synchronize(device)
    dt = time.time() - t0
    tot.append(dt)
assert out.attentions is None
print(f"Forward took {np.mean(tot):.2e} s")


# Actual test with return_xxx
tot = []
for _ in range(EXPERIENCE):
    torch.cuda.synchronize(device)
    t0 = time.time()
    with torch.no_grad():
        out = model(input_ids, return_dict=True, output_attentions=True)
    torch.cuda.synchronize(device)
    dt = time.time() - t0
    tot.append(dt)
assert out.attentions is not None
print(f"Return attention forward took {np.mean(tot):.2e} s")

```

</details>

The above script was used to benchmark the speed of both approches when using `output_xxx` or not.
Results on 1 small foward of Llama3 8B (input size (2, 56)):

|            | Without output capture | With output capture | 
|-------|:-----------------------:|:---------------------:|
| Currently |    3.37e-02 s   |     3.53e-02 s |  
| This PR |    3.33e-02 s (**1.2% faster**)|   3.41e-02 s (**3.5% faster**) |

So this PR is slightly faster in both cases.

## Note

Some models did not use the decorator correctly or very redundantly, that's why I had to perform some minimal changes on a few.

### DIFF
--- a/src/transformers/utils/generic.py
+++ b/src/transformers/utils/generic.py
@@ -851,6 +851,60 @@ def can_return_tuple(func):
     return wrapper
 
 
+def merge_with_config_defaults(func):
+    """
+    Decorator using config field (if they exist) as default value for some args and kwargs. Precedence is always
+    given to the args/kwargs that are explicitly passed.
+    """
+
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        args_with_config_defaults = [
+            "use_cache",
+            "vision_feature_layer",
+            "vision_feature_select_strategy",
+            "vision_aspect_ratio",
+        ]
+        for arg_name in args_with_config_defaults:
+            arg_index = None
+            if arg_name in func.__code__.co_varnames:
+                arg_index = func.__code__.co_varnames.index(arg_name) - 1  # -1 for self
+
+            if arg_index is not None and len(args) > arg_index and args[arg_index] is not None:
+                arg_value = args[arg_index]
+            elif kwargs.get(arg_name) is not None:
+                arg_value = kwargs[arg_name]
+            else:
+                arg_value = getattr(self.config, arg_name, None)
+
+            if arg_value is not None:
+                # Arg-specific handling
+                if arg_name == "use_cache":
+                    if getattr(self, "gradient_checkpointing", False) and self.training and arg_value:
+                        logger.warning_once(
+                            "`use_cache=True` is incompatible with gradient checkpointing. Setting `use_cache=False`."
+                        )
+                        arg_value = False
+                elif arg_name == "vision_feature_select_strategy":
+                    valid_strategies = ["default", "full"]
+                    if arg_value not in valid_strategies:
+                        raise ValueError(
+                            f"`Unexpected select feature strategy: {arg_value}. Please select from {valid_strategies}."
+                        )
+
+                if arg_index is not None and len(args) > arg_index:
+                    args = list(args)
+                    args[arg_index] = arg_value
+                    args = tuple(args)
+                else:
+                    kwargs[arg_name] = arg_value
+
+        output = func(self, *args, **kwargs)
+        return output
+
+    return wrapper
+
+
 @dataclass
 @requires(backends=("torch",))
 class OutputRecorder:
--- a/src/transformers/utils/generic.py
+++ b/src/transformers/utils/generic.py
@@ -981,11 +981,11 @@ def install_output_capuring_hook(module: nn.Module, key: str, index: int) -> Non
             return
 
         if key == "hidden_states" and len(collected_outputs[key]) == 0:
-            collected_outputs[key] += (args[0],)
+            collected_outputs[key].append(args[0])
         if not isinstance(output, tuple):
-            collected_outputs[key] += (output,)
+            collected_outputs[key].append(output)
         elif output[index] is not None:
-            collected_outputs[key] += (output[index],)
+            collected_outputs[key].append(output[index])
 
     module.register_forward_hook(output_capturing_hook)
 
@@ -1171,7 +1171,7 @@ def check_model_inputs(func=None, *, tie_last_hidden_states=True):
             if "output_attentions" in recordable_keys:
                 recordable_keys["output_cross_attentions"] = recordable_keys["output_attentions"]
 
-            collected_outputs = {k.replace("output_", ""): () for k, v in recordable_keys.items() if v}
+            collected_outputs = {k.replace("output_", ""): [] for k, v in recordable_keys.items() if v}
             # Make sure hooks are installed if we need to collect outputs
             if len(collected_outputs) > 0:
                 maybe_install_capturing_hooks(self)
@@ -1217,20 +1217,21 @@ def check_model_inputs(func=None, *, tie_last_hidden_states=True):
                         pass
                     elif hasattr(outputs, "vision_hidden_states"):
                         collected_outputs[key] = collected_outputs[key][:-1]
-                        collected_outputs[key] += (outputs.vision_hidden_states,)
+                        collected_outputs[key].append(outputs.vision_hidden_states)
                     elif hasattr(outputs, "last_hidden_state"):
                         collected_outputs[key] = collected_outputs[key][:-1]
-                        collected_outputs[key] += (outputs.last_hidden_state,)
+                        collected_outputs[key].append(outputs.last_hidden_state)
 
-                    outputs[key] = collected_outputs[key]
+                    outputs[key] = tuple(collected_outputs[key])
                 elif key == "attentions":
                     if isinstance(capture_flags[key], list) and len(capture_flags[key]) == 2:
-                        outputs[key] = collected_outputs[key][0::2]
-                        outputs["cross_" + key] = collected_outputs[key][1::2]
+                        outputs[key] = tuple(collected_outputs[key][0::2])
+                        outputs["cross_" + key] = tuple(collected_outputs[key][1::2])
                     else:
-                        outputs[key] = collected_outputs[key]
+                        outputs[key] = tuple(collected_outputs[key])
                 else:
-                    outputs[key] = collected_outputs[key]
+                    outputs[key] = tuple(collected_outputs[key])
+
             if return_dict is False:
                 outputs = outputs.to_tuple()
             return outputs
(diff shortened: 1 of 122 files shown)


utils/check_repo.py
```diff
@@ -110,6 +110,11 @@
     "PeAudioPreTrainedModel",
     "PeAudioVideoPreTrainedModel",
     "PeVideoPreTrainedModel",
+    "CLIPTextTransformer",  # was not a PreTrainedModel originally but needs to be
+    "CLIPVisionTransformer",  # was not a PreTrainedModel originally but needs to be
+    "MetaClip2TextTransformer",  # was not a PreTrainedModel originally but needs to be
+    "MetaClip2VisionTransformer",  # was not a PreTrainedModel originally but needs to be
+    "MLCDVisionTransformer",  # was not a PreTrainedModel originally but needs to be
```
ArthurZucker, line 117: why do they need to be?

src/transformers/utils/generic.py
```diff
@@ -855,12 +918,132 @@ class OutputRecorder:
...
     """
 
-    target_class: "type[torch.nn.Module]"
+    target_class: type[nn.Module]
     index: int = 0
     layer_name: str | None = None
     class_name: str | None = None
 
 
+class CompileableContextVar:
```
ArthurZucker: can we move everything out of `generic.py` and into `output_recording` ?

src/transformers/utils/generic.py
```diff
@@ -855,12 +918,132 @@ class OutputRecorder:
...
+            self.context_var.reset(token)
+
+
+# Thread-safe global variables
+_active_collector = CompileableContextVar("output_collector", default=None)
+_active_keys = CompileableContextVar("keys_to_capture", default=None)
+
+
+def install_output_capuring_hook(module: nn.Module, key: str, index: int) -> None:
+    """Install the forward hook needed to capture the output described by `key` and `index` in `module`."""
```
ArthurZucker: the tradeoff is that now we have installed this on EVERY model, regardless of them needing any recording. Forwardd context / stacktrace is gonna grow quite a lot, which I am not much of a fan TBH.

ArthurZucker (commented):
Nice, we have a big decision to make here tho. 

One thing that I would love to reach is having 0 hooks / a normal forward when we are not recording to avoid having the unreadable stacktrace.


Other than that, let's try to reduce the overhead for inference-heavy workloads that never use capturing.                                                                        
                                                                                                                                                                                                                                                                                                                                                                                      Currently, every hooked module calls `_active_keys.get()` on every forward pass, even when capturing is disabled.`ContextVar` lookup probably is not free.
                                                                                                                                                                                                                       
We can add a fast-path guard using a simple global ref count of the active recorders.

Cyrilvallez:
Yes I agree! I woke up this morning with the same idea basically: I want to defer hook installation as most of the time we don't want any output recording - I believe we can do so with simple locks to avoid installing twice.

As for the `ContextVar.get()`, I benchmarked it carefully and it's actually completely negligible (~2e-6 s) compared to the forward of a module. However, calling the hook itself is not entirely negligible due to pytorch implem (i.e. even an empty hook with only `pass` create a very tiny measurable overhead)

utils/check_repo.py
Cyrilvallez: Due to the super weird/bad structure of the other composite models using them...
Basically the already public `PreTrainedModel` were just very thin wrapper around them doing nothing more, so the composite models use the following in `__init__`:

```
text_model = CLIPTextModel._from_config(text_config)
self.text_model = text_model.text_model   #  <-- this child is `"CLIPTextTransformer"`
```

so it uses the submodel of the public wrapper `CLIPTextModel` (which is `CLIPTextTransformer`) directly. However, since `CLIPTextTransformer` needs to capture output with its own `_can_record_putputs`, it needs to be a `PreTrainedModel`. I have absolutely no clue why `CLIPTextModel` would even exist originally since other models use it to instantiate, but then bypass it by using the child....

vasqu: cc @zucchini-nlp @molbap this is known on the vision side, it's super awkward and is why it needs a refactor

zucchini-nlp: yep, a known issue and easily fixable by re-ordering modules. @yonigozlan had a plan to work on it :)

Cyrilvallez:
Actually I was wrong and it's not the hooks themselves that were slowing down the forward. They basically have no visible cost. Anyway, I benchmarked everything and this PR makes everything slightly faster in both cases.
TLDR -> we go faster and we are muuuuch safer

ArthurZucker (approved):
Should we have `pytest-benchmark` on tests specifc to this?

O-J1:
@Cyrilvallez Was this fix backported into the v4 releases of Transformers? We arent ready to migrate to v5 I suspect for quite a while (OneTrainer, I help with Dxzqb). Would be a big boon for us if it was but I can understand why it may have not

ArthurZucker:
It did not but we can push for that if required! (gonna be a bit hard)

Cyrilvallez:
Yep, not sure how worth it it is, and it's gonna be a mess to pick that commit back to v4

ArthurZucker:
@O-J1 what can we do to help transition to v5?

O-J1:
> @O-J1 what can we do to help transition to v5?

@ArthurZucker 

At the moment not much 🤣 we are swamped with refactors, PyTorch bugs and functionality additions. Once we have the majority of our big ticket PR items merged I think there will be more headspace for us to transition.

Based on what you have both communicated I don’t wanna make your codebase a mess so let’s ignore my request. We will bite the bullet eventually

PR #43766, opened 2026-02-05
Fix `convert_rope_params_to_dict` so it uses `rope_theta` from the config

hmellor:
`convert_rope_params_to_dict` assumes that `rope_theta` and `partial_rotary_factor` will be present in `kwargs`. This is only true if these parameters are not explicit arguments of the config class's `__init__` method.

i.e. `convert_rope_params_to_dict` will fail for a config which looks like:

```py
class MyConfig(PreTrainedConfig):
    def __init__(self, ..., rope_theta, ..., **kwargs):
        ...
        super().__init__(**kwargs)
```

All custom configs to date will look like this so this is not a safe assumption to make.

This PR updates `convert_rope_params_to_dict` so that it checks for the rope parameters in the following locations in order of priority:

1. Values in `rope_parameters` dict (where they should be after standardization)
2. Values in `kwargs` (i.e. it's in config.json but not MyConfig.__init__'s args)
3. Values in the config's attributes (i.e. it's in MyConfig.__init__'s args)
4. Default values (i.e. not present at all but other RoPE parameters are present)

It also simplifies the fallback to `{}` for `rope_parameters`.

### DIFF
--- a/src/transformers/modeling_rope_utils.py
+++ b/src/transformers/modeling_rope_utils.py
@@ -630,9 +630,9 @@ class RotaryEmbeddingConfigMixin:
     default_theta = 10_000.0
 
     def convert_rope_params_to_dict(self, ignore_keys_at_rope_validation: set | None = None, **kwargs):
-        # If `rope_scaling` was present, it will be written to `rope_parameters` automatically by `@rope_scaling.setter`
-        if self.rope_parameters is None:
-            self.rope_parameters = {}
+        rope_scaling = kwargs.pop("rope_scaling", None)
+        self.rope_parameters = rope_scaling or self.rope_parameters
+        self.rope_parameters = self.rope_parameters if self.rope_parameters is not None else {}
 
         # Standardize and validate the correctness of rotary position embeddings parameters. Priority for these parameters is:
         # 1. Values in `rope_parameters` dict (where they should be after standardization)
--- a/src/transformers/modeling_rope_utils.py
+++ b/src/transformers/modeling_rope_utils.py
@@ -630,15 +630,21 @@ class RotaryEmbeddingConfigMixin:
     default_theta = 10_000.0
 
     def convert_rope_params_to_dict(self, ignore_keys_at_rope_validation: set | None = None, **kwargs):
-        rope_scaling = kwargs.pop("rope_scaling", None)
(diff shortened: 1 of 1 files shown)


src/transformers/modeling_rope_utils.py
```diff
@@ -630,15 +630,21 @@ class RotaryEmbeddingConfigMixin:
...
-        self.rope_parameters = self.rope_parameters if self.rope_parameters is not None else {}
-
-        # Standardize and validate the correctness of rotary position embeddings parameters
-        self.rope_parameters.setdefault("rope_theta", kwargs.pop("rope_theta", self.default_theta))
-
-        if "partial_rotary_factor" in kwargs:
-            self.rope_parameters.setdefault("partial_rotary_factor", kwargs["partial_rotary_factor"])
+        # If `rope_scaling` was present, it will be written to `rope_parameters` automatically by `@rope_scaling.setter`
+        if self.rope_parameters is None:
+            self.rope_parameters = {}
```
zucchini-nlp: i think what we had in L633-636 was doing the same, and a bit more by checking `kwargs`. Can we revert it back?

src/transformers/modeling_rope_utils.py
```diff
@@ -630,15 +630,21 @@ class RotaryEmbeddingConfigMixin:
...
+
+        # Standardize and validate the correctness of rotary position embeddings parameters. Priority for these parameters is:
+        # 1. Values in `rope_parameters` dict (where they should be after standardization)
+        # 2. Values in `kwargs` (i.e. it's in config.json but not MyConfig.__init__'s args)
+        # 3. Values in the config's attributes (i.e. it's in MyConfig.__init__'s args)
+        # 4. Default values (i.e. not present at all but other RoPE parameters are present)
```
zucchini-nlp, line 641: nice explanation!

zucchini-nlp (approved):
Let's just revert `rope_scaling` and we can merge!

src/transformers/modeling_rope_utils.py
hmellor: Sure

vasqu:
Looks like the branch is a few commits behind main? Best to merge/rebase re CI

PR #43768, opened 2026-02-05
Fix init weights in remote code

zucchini-nlp:
# What does this PR do?

Helps vLLM to bump to v5


src/transformers/modeling_utils.py
```diff
@@ -2311,6 +2312,9 @@ def _initialize_weights(self, module):
         if getattr(module, "_is_hf_initialized", False):
             return
 
+        if (weight := getattr(module, "weight", None)) is not None and getattr(weight, "_is_hf_initialized", False):
+            return
+
```
zucchini-nlp, line 2322: module's never have an `_is_hf_initialized` attr, ig this is a typo? Otherwise it causes the whole model to be random init when remote code has an old-format `_init_weights` defined and it takes ages for big models

src/transformers/modeling_utils.py
```diff
@@ -4212,12 +4218,11 @@ def _finalize_model_loading(
             load_config.device_mesh,
             load_config.hf_quantizer,
         )
-
         # Correctly initialize the missing (and potentially mismatched) keys (all parameters without the `_is_hf_initialized` flag)
         model._initialize_missing_keys(load_config.is_quantized)
 
         # Tie the weights
-        model.tie_weights(missing_keys=loading_info.missing_keys, recompute_mapping=False)
+        model.tie_weights(missing_keys=loading_info.missing_keys)
```
zucchini-nlp: not sure about this one, it's not BC for models that have `tie_weights` overriden. But we might just say it's supposed to be fixed in remote code

src/transformers/tokenization_python.py
```diff
@@ -433,7 +433,7 @@ def __init__(self, **kwargs):
 
         # 5. Special tokens mask configuration
         # Patterns: "none", "cls_sep", "eos", "bos", "bos_eos", "cls_double_sep", "prefix_suffix"
-        self.special_tokens_pattern = kwargs.pop("special_tokens_pattern", "cls_sep")
+        self.special_tokens_pattern = kwargs.pop("special_tokens_pattern", "bos_eos")
```
zucchini-nlp: cc @itazap @ArthurZucker , i want to clarify this part. Should we default to `None` because cls-sep ids arent always available for all tokenizers. Thus we are getting `[None, 1, 18001, 468, None]` as token ids for those models

Ignore the current change to `bos_eos`

src/transformers/modeling_utils.py
ArthurZucker: you are right they never do now, only the tensors have them. 
The check should only run for remote code I think no?

ArthurZucker (approved):
missing a few tests IMO (especially for the tokenizer fix)

src/transformers/modeling_utils.py
zucchini-nlp: yes, for local models it will not have much effect because we check `if weight._is_hf_initialized` later in the `initialization.py`. So we are never random init weights for already loaded params

zucchini-nlp:
run-slow: llama, mixtral. whisper, bart, mamba, gemma3n, qwen2_vl, llava

src/transformers/modeling_utils.py
Cyrilvallez: Wait, what is this check? Any module with both a weight and something else would be skipped if the "something else" is a missing param!

Cyrilvallez: Modules WILL have the flag when weight init is not called from `from_pretrained`. Image doing raw instantiation with a composite model, such as `model = ModelArch(config)`, then every submodel will run `post_init` and initialize its weights, and set the flag, so that next post_init does not run it again!

zucchini-nlp: It was due to remote code models re-init weights randomly. For ex, this one has a custom `_init_weights` and will random init without checking if weights were loaded from ckpt or not. Or do we not support this type of models on purpose in v5 in which case we can revert and skip these models in vLLM?

https://huggingface.co/TIGER-Lab/VLM2Vec-Full/blob/main/modeling_phi3_v.py#L1237-L1247

Cyrilvallez: v5 broke most of the loading of remote code. For this model in particular, note how e.g. the rope module would have random weights anyway as the non-persistent buffer is not reinitialized explicitly and so would lose its value

Cyrilvallez: Oh actually they initialize it at None and then update it, so would be fine in this case

PR #43770, opened 2026-02-05
Remove unconditional train_batch_size assignment

lordaarush:
# What does this PR do?

Removes the unconditional `self.state.train_batch_size = self._train_batch_size` assignment that was causing issues when resuming from checkpoint with different batch configurations.

The `train_batch_size` should only be saved to `TrainerState` when `auto_find_batch_size` is enabled, which is already handled in the `auto_find_batch_size` block. The unconditional assignment was redundant and caused the bug where `max_steps` was incorrectly calculated when resuming training with a different batch size configuration (even when keeping the same global batch size).

Fixes #43708

## Before submitting
      Pull Request section?
      to it if that's the case. Yes, discussed in #43708

## Who can review?

@SunMarc

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -2146,7 +2146,10 @@ class Trainer:
                 self._load_from_checkpoint(resume_from_checkpoint)
             # In case of repeating the find_executable_batch_size, set `self._train_batch_size` properly
             state = TrainerState.load_from_json(os.path.join(resume_from_checkpoint, TRAINER_STATE_NAME))
-            if state.train_batch_size is not None:
+            # Only restore the checkpoint's train_batch_size when using auto_find_batch_size,
+            # as that feature needs to resume with the automatically-found batch size.
+            # Otherwise, use the current args batch size to allow users to change batch configuration.
+            if state.train_batch_size is not None and args.auto_find_batch_size:
                 self._train_batch_size = state.train_batch_size
 
         # If model was re-initialized, put it on the right device and update self.model_wrapped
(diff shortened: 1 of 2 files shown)


SunMarc (commented):
Nice, can you add a simple test to test that it is indeed not saved when auto_find_batch_size is False ?

lordaarush:
Sure, I'll do just that.

lordaarush:
Done @SunMarc ! Added a test that verifies `train_batch_size` is `None` in the saved checkpoint when `auto_find_batch_size=False`.

SunMarc (approved):
Thanks !

lordaarush:
Hi @SunMarc, looks like simply removing that line broke some tests:
FAILED test_resume_training_with_frozen_params - TypeError: unsupported operand type(s) for //: 'NoneType' and 'int'
FAILED test_resume_training_with_checkpoint - TypeError: unsupported operand type(s) for //: 'NoneType' and 'int'
FAILED test_auto_batch_size_with_resume_from_checkpoint - TypeError: '>=' not supported between instances of 'NoneType' and 'int'
... (9 total trainer resume tests failed)

The issue is that other resume code expects `state.train_batch_size` to have a value for calculations.

I think the correct fix is to:
1. **Keep saving** `train_batch_size` to the checkpoint (as before)
2. **Only restore** it to `_train_batch_size` when `auto_find_batch_size=True`

This way:
- `train_batch_size` is still available in state for other resume calculations
- But it only overwrites the user's batch config when `auto_find_batch_size` is enabled (which needs it)

Should I update the PR with this approach?

SunMarc:
Okay let's do that for now ! Thanks for checking

lordaarush:
Updated @SunMarc ! I've implemented the approach we discussed. Also added `test_resume_training_with_different_batch_size`to verify users can change batch size when resuming. All tests passing now.

SunMarc:
Thanks a lot ! merging

PR #43772, opened 2026-02-05
[`Modular Dependencies`] Fixup qwen rms norms

vasqu:
As per title, this led to weird dependencies where modeling files used direct imports

### DIFF
--- a/src/transformers/models/dots1/modeling_dots1.py
+++ b/src/transformers/models/dots1/modeling_dots1.py
@@ -48,7 +48,7 @@ from .configuration_dots1 import Dots1Config
 
 @use_kernel_forward_from_hub("RMSNorm")
 class Dots1RMSNorm(nn.Module):
-    def __init__(self, hidden_size, eps: float = 1e-6) -> None:
+    def __init__(self, hidden_size, eps=1e-6):
         """
         Dots1RMSNorm is equivalent to T5LayerNorm
         """
@@ -56,7 +56,7 @@ class Dots1RMSNorm(nn.Module):
         self.weight = nn.Parameter(torch.ones(hidden_size))
         self.variance_epsilon = eps
 
-    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+    def forward(self, hidden_states):
         input_dtype = hidden_states.dtype
         hidden_states = hidden_states.to(torch.float32)
         variance = hidden_states.pow(2).mean(-1, keepdim=True)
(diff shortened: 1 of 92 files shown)


src/transformers/models/dots1/modeling_dots1.py
```diff
@@ -48,15 +48,15 @@
 
 @use_kernel_forward_from_hub("RMSNorm")
 class Dots1RMSNorm(nn.Module):
-    def __init__(self, hidden_size, eps: float = 1e-6) -> None:
+    def __init__(self, hidden_size, eps=1e-6):
```
zucchini-nlp: the type hints were pretty, can we add them in llama so it's copied to all models?

src/transformers/models/qwen2_5_omni/modeling_qwen2_5_omni.py
```diff
@@ -1064,8 +1085,8 @@ def forward(self, hidden_state):
 class Qwen2_5OmniVisionBlock(GradientCheckpointingLayer):
     def __init__(self, config: Qwen2_5OmniVisionEncoderConfig) -> None:
         super().__init__()
-        self.norm1 = Qwen2RMSNorm(config.hidden_size, eps=1e-6)
-        self.norm2 = Qwen2RMSNorm(config.hidden_size, eps=1e-6)
+        self.norm1 = Qwen2_5OmniRMSNorm(config.hidden_size, eps=1e-6)
+        self.norm2 = Qwen2_5OmniRMSNorm(config.hidden_size, eps=1e-6)
```
zucchini-nlp, line 1089: finally, have been annoyed by this as well!

zucchini-nlp (approved):
Super nice, happy to get rid of bad modular patterns!

src/transformers/models/dots1/modeling_dots1.py
vasqu: Fair, added it in c8574896c38a0b53ce11fd549099ccbf9e80e81d

src/transformers/models/qwen2/modular_qwen2.py
```diff
@@ -239,7 +207,6 @@ class Qwen2ForQuestionAnswering(LlamaForQuestionAnswering):
     "Qwen2PreTrainedModel",
     "Qwen2Model",
     "Qwen2ForCausalLM",
-    "Qwen2RMSNorm",
```
ArthurZucker, line 242: this is breaking haha

ArthurZucker (approved):
we should still keep the alias IMO

src/transformers/models/qwen2/modular_qwen2.py
vasqu: Fair enough, it shouldn't have been this way in the beginning but it's not worth to break

PR #43776, opened 2026-02-05
Refactor trainer data_collator and callbacks tests

SunMarc:
# What does this PR do?

This PR updates and simplifies tests for trainer. We never really had any issues with those tests, just cleaning a bit. 

1) `tests/trainer/test_data_collator.py` 
- Restructured from 4 large classes (PyTorch/NumPy × Integration/Immutability) into 10
  focused classes by collator type
- Each test class now contains PyTorch, NumPy, and immutability tests together
- Added missing tests (attention_mask generation, flattening immutability)

2) `tests/trainer/test_trainer_callback.py`
  - Split monolithic tests into ~30 focused single-purpose tests
  - Added 4 new test classes: `TrainerStateTest`, `TrainerControlTest`, `CallbackHandlerTest`, `EarlyStoppingCallbackTest`
  - Tests now cover: callback registration, event firing, control flow, state,  persistence, and early stopping logic

### DIFF
--- a/tests/trainer/test_data_collator.py
+++ b/tests/trainer/test_data_collator.py
@@ -536,6 +536,21 @@ class TestDataCollatorForSeq2Seq(DataCollatorTestMixin, unittest.TestCase):
 
         self.assertEqual(batch["labels"][0].tolist(), [0, 1, 2, -1, -1, -1])
 
+    def test_do_not_pad(self):
+        """Test DO_NOT_PAD raises on unequal lengths, works on equal."""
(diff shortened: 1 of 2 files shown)


MekkCyber (approved):
Huge work! Thank you

ArthurZucker (commented):
Is there a specific part I should be review? (there is a loooooot )

SunMarc:
> Is there a specific part I should be review? (there is a loooooot )

Not really haha. Just look at the structure of the tests I would say. But this is not really important as those tests never really failed until now.

Issue #43717, opened 2026-02-03
init_weights usage in Mamba and Mamba-2


kevinli573:
Hi,

`dt_bias` initialization has a pretty non-trivial impact on Mamba training/performance.

In the current implementation, it looks like `dt_bias` is [initialized to all ones](https://github.com/huggingface/transformers/blob/v5.0.0/src/transformers/models/mamba2/modeling_mamba2.py#L253) in the mixer class and then later [reinitialized to the standard range](https://github.com/huggingface/transformers/blob/v5.0.0/src/transformers/models/mamba2/modeling_mamba2.py#L730) in the parent PreTrainedModel class.

We were wondering if this is the right way to structure the code for Mamba models, as those who directly import the layer/mixer without using the full model will have the incorrect initialized `dt_bias` values. If this is not the case, would it be possible to move the initialization into the mixer?

Thanks!

tridao:
I'd like to advocate calling the weight_init code when initializing the Mamba2Mixer class. There were multiple cases of folks using the wrong initialization (for dt_bias) for months because they only import Mamba2Mixer where dt_bias has the wrong initialization (all ones). 
If this sounds fine, I'll work with @kevinli573 on a PR.

ArthurZucker:
Sounds great actually, @Cyrilvallez was gonna refactor `transformers` in general to have a much more granular, module level init scheme! 

Super happy if you can open a PR! 🤗

Cyrilvallez:
Indeed! Weight init schemes have historically been centralized in the `_init_weights` of the `XXXPreTrainedModel`, but we've recently discussed a lot about it and we all feel that it would be much easier/readable to have them at the module-level!

PR #43778, opened 2026-02-05
Mamba-1/-2 init weights in mixer class

kevinli573:
# What does this PR do?

Initializes dt_bias in the mixer class for both Mamba-1 and Mamba-2 (also initialized in PreTrainedModel class for meta device test).

Addresses [#43717](https://github.com/huggingface/transformers/issues/43717)

### DIFF
--- a/src/transformers/models/mamba/modeling_mamba.py
+++ b/src/transformers/models/mamba/modeling_mamba.py
@@ -187,6 +187,15 @@ class MambaMixer(nn.Module):
         # time step projection (discretization)
         self.dt_proj = nn.Linear(self.time_step_rank, self.intermediate_size, bias=True)
 
+        dt = torch.exp(
+            torch.rand(self.intermediate_size)
+            * (math.log(config.time_step_max) - math.log(config.time_step_min))
+            + math.log(config.time_step_min)
+        ).clamp(min=config.time_step_floor)
+        # Inverse of softplus: https://github.com/pytorch/pytorch/issues/72759
+        inv_dt = dt + torch.log(-torch.expm1(-dt))
+        init.copy_(self.dt_proj.bias, inv_dt)
+
         # S4D real initialization. These are not discretized!
         # The core is to load them, compute the discrete states, then write the updated state. Keeps the memory bounded
         A = torch.arange(1, self.ssm_state_size + 1, dtype=torch.float32)[None, :]
--- a/src/transformers/models/mamba/modeling_mamba.py
+++ b/src/transformers/models/mamba/modeling_mamba.py
@@ -188,8 +188,7 @@ class MambaMixer(nn.Module):
         self.dt_proj = nn.Linear(self.time_step_rank, self.intermediate_size, bias=True)
 
         dt = torch.exp(
-            torch.rand(self.intermediate_size)
-            * (math.log(config.time_step_max) - math.log(config.time_step_min))
+            torch.rand(self.intermediate_size) * (math.log(config.time_step_max) - math.log(config.time_step_min))
             + math.log(config.time_step_min)
         ).clamp(min=config.time_step_floor)
         # Inverse of softplus: https://github.com/pytorch/pytorch/issues/72759
--- a/src/transformers/models/mamba/modeling_mamba.py
+++ b/src/transformers/models/mamba/modeling_mamba.py
@@ -156,7 +156,7 @@ class MambaMixer(nn.Module):
     and is why Mamba is called **selective** state spaces)
     """
 
-    def __init__(self, config: MambaConfig, layer_idx: int):
+    def __init__(self, config: MambaConfig, layer_idx: int, initialize_mixer_weights: bool = True):
         super().__init__()
         self.config = config
         self.hidden_size = config.hidden_size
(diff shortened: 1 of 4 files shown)


kevinli573:
> **[For maintainers]** Suggested jobs to run (before merge)
> 
> run-slow: mamba, mamba2

The following tests are failing, but they are unrelated to the changes proposed in this PR and also fail on the main branch.

```
FAILED tests/models/mamba2/test_modeling_mamba2.py::Mamba2ModelTest::test_mamba2_slow_vs_fast_forward - TypeError: 'NoneType' object is not callable
FAILED tests/models/mamba2/test_modeling_mamba2.py::Mamba2ModelTest::test_mamba2_slow_vs_fast_forward_grouped - TypeError: 'NoneType' object is not callable
FAILED tests/models/mamba2/test_modeling_mamba2.py::Mamba2ModelTest::test_resize_tokens_embeddings - AssertionError: Tensor-likes are not close!
```

```
FAILED tests/models/mamba/test_modeling_mamba.py::MambaModelTest::test_resize_tokens_embeddings - AssertionError: Tensor-likes are not close!
```

ArthurZucker (commented):
LGTM, @Cyrilvallez I am not 100% this won't prevent `_init_weight` from overwriting it :
```python

def copy_(tensor: torch.Tensor, other: torch.Tensor) -> torch.Tensor:
    if not getattr(tensor, "_is_hf_initialized", False):
        with torch.no_grad():
            return tensor.copy_(other)
    return tensor
```
does not set it to `True`

kevinli573:
> LGTM, @Cyrilvallez I am not 100% this won't prevent `_init_weight` from overwriting it :
> 
> ```python
> def copy_(tensor: torch.Tensor, other: torch.Tensor) -> torch.Tensor:
>     if not getattr(tensor, "_is_hf_initialized", False):
>         with torch.no_grad():
>             return tensor.copy_(other)
>     return tensor
> ```
> 
> does not set it to `True`

I might be misinterpreting, so please don't hesitate to correct me if so.

I had initially removed the redundant init_weights of A_log and dt_bias in the PreTrainedModel class, but that caused the a test for meta device initialization to fail. So currently, if you instantiate a Mamba2PreTrainedModel, the dt_bias and A_log will be init twice.

src/transformers/models/mamba/modeling_mamba.py
```diff
@@ -187,6 +187,14 @@ def __init__(self, config: MambaConfig, layer_idx: int):
         # time step projection (discretization)
         self.dt_proj = nn.Linear(self.time_step_rank, self.intermediate_size, bias=True)
 
+        dt = torch.exp(
+            torch.rand(self.intermediate_size) * (math.log(config.time_step_max) - math.log(config.time_step_min))
+            + math.log(config.time_step_min)
+        ).clamp(min=config.time_step_floor)
+        # Inverse of softplus: https://github.com/pytorch/pytorch/issues/72759
+        inv_dt = dt + torch.log(-torch.expm1(-dt))
+        init.copy_(self.dt_proj.bias, inv_dt)
```
Cyrilvallez: Indeed, as @ArthurZucker said, this will be overwritten in most case. We start by initializing the models on meta device, so all weights defined in `__init__` do not get instantiated (only their shape etc).
Then, if the weight is missing, or if we want a fully randomly instantiated model, the proper way is to it in `_init_weights` currently, which will be called later on.
So here this should go inside this block of the `_init_weights`:

```python
def _init_weights(self, module):
        """Initialize the weights."""
        std = self.config.initializer_range
        if isinstance(module, MambaMixer):
```

Cyrilvallez: Oh but I see it's already the same there!

Cyrilvallez: Then if you want to avoid (potentially costly) double init of the same weights in case of random initialization, you should declare all nn.Parameter and buffers as `torch.empty` tensor inside `__init__` (the same way it's done in nn.Linear for example)

kevinli573: Hmm I see, so what you are recommended is just declare the weights as torch.empty within the mixer and have a specialized init_weights function for the MambaMixer class?

Cyrilvallez: Yes - well rn it's unfortunately not on the MambaMixer itself that it needs to live, but on the `MambaPreTrainedModel` class. We are thinking about making those module-level though, and will do in the future

kevinli573:
@ArthurZucker @Cyrilvallez sorry for the late follow-up.

I've updated the implementation so the select params are declared as empty with a local mixer-based init helper that runs only when the mixer is instaniated on a non-meta device. To prevent double inits, I had to add a custom `initialize_mixer_weights` argument (which might still cause some issues if people decide to copy over both the mixer and block).

This does seem a bit unwieldy, but correct initialization is really important for users who instantiate/copy the mixer directly. The same issue was also present in the [flash-linear-attention](https://github.com/fla-org/flash-linear-attention/pull/739) repo, where it had a noticable impact on pretrained performance.

If you all have other suggestions on how to handle the initialization, I'd be more than happy to adjust and implement it!

ArthurZucker:
@kevinli573 took the liberty to fix CIs, we'll merge this as is for the release!

ArthurZucker:
Thanks again for the PR and the reports! 🤗

kevinli573:
Fantastic, thank you both!

PR #43781, opened 2026-02-05
fix: Load model with co-located adapter from local path (Granite Speech)

gabe-l-hart:
Only overwrite the pretrained_model_name_or_path if needed with adapter

The check is based on the assumption that if the current value is a path on disk and there is a `config.json` present in that path, the path points to a full model checkpoint with an embedded adapter.

Fixes: https://github.com/huggingface/transformers/issues/43746

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

PEFT: @BenjaminBossan @githubnemo


gabe-l-hart:
The failing tests seem to be unrelated:

```
FAILED tests/models/udop/test_processing_udop.py::UdopProcessorTest::test_unstructured_kwargs_batched - TypeError: TextInputSequence must be str
```

https://app.circleci.com/pipelines/github/huggingface/transformers/163107/workflows/cba5c187-f7e8-4292-9da6-fe1a001af4e8/jobs/2149103

tests/models/auto/test_modeling_auto.py
```diff
@@ -541,3 +544,45 @@ def test_model_with_dotted_name_and_relative_imports(self):
...
+        transformers_test_model = "hf-internal-testing/tiny-random-OPTForCausalLM"
+
+        # Create a temporary directory with a complete adapter model structure
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            tmp_dir = Path(tmp_dir)
+
+            # Save the model and adapter locally
+            config = BertConfig.from_pretrained(transformers_test_model)
+            model = AutoModel.from_pretrained(transformers_test_model)
+            adapter_model = AutoModel.from_pretrained(peft_test_model)
```
Rocketknight1: This is a little strange - why are we initializing a `BertConfig` from an `OPT` checkpoint?

Rocketknight1 (approved):
This looks like a good fix! I have one question about the test, though. Also, this path is only hit when the model is loaded with `AutoModel`. This is by far the most common, but I think if the user loads with a specific model class, they might hit the check in `integrations/peft.py` or even `pipelines/__init__.,py`, which looks like it's copied from here. We should probably make the same fix everywhere, right? I might be wrong, though - I'm not ultra-familiar with the PEFT code.

tests/models/auto/test_modeling_auto.py
gabe-l-hart: Ah, yep, that's some slop I missed cleaning up from the draft test (Claude Code + glm-4.7-flash in Ollama). Thanks for catching this.

gabe-l-hart:
>but I think if the user loads with a specific model class, they might hit the check in integrations/peft.py or even pipelines/__init__.

Good call! I only traced my failure path and didn't notice these other cases. I'll extend to those today.

gabe-l-hart:
@Rocketknight1 One small style question: I see that throughout the codebase, there's a mix of `pathlib.Path` and `os.path` used. I'm completely ambivalent on the two, so I opted to use `os.path` since it doesn't introduce another import. Is there a general community preference for one over the other?

Rocketknight1:
@gabe-l-hart no preference! Both are acceptable for now.

Also cc @BenjaminBossan @githubnemo for review, I don't want to merge a PEFT patch I don't understand that well!

src/transformers/integrations/peft.py
```diff
@@ -954,7 +954,12 @@ def maybe_load_adapters(
     if _adapter_model_path is not None and os.path.isfile(_adapter_model_path):
         with open(_adapter_model_path, "r", encoding="utf-8") as f:
             _adapter_model_path = pretrained_model_name_or_path
-            pretrained_model_name_or_path = json.load(f)["base_model_name_or_path"]
+            # Only override the model name/path if the current value doesn't point to a
+            # complete model with an embedded adapter
```
githubnemo: Let's document the 'why' as well (e.g., "to handle cases where the base model is in the same local directory as the adapter" or something).

tests/models/auto/test_modeling_auto.py
```diff
@@ -541,3 +544,45 @@ def test_model_with_dotted_name_and_relative_imports(self):
...
+        """
+        Test for issue #43746: Only overwrite the pretrained_model_name_or_path if needed with adapter.
+
+        This test ensures that when a model has an adapter config and the pretrained_model_name_or_path
+        points to a complete model with an embedded adapter (i.e., it has a config.json file),
+        the path should NOT be overwritten with the base_model_name_or_path.
+
+        The bug was that the path was being unconditionally overwritten, which would cause
+        incorrect behavior when loading models with adapters that are embedded within the
+        same directory as the base model.
```
githubnemo, line 559: I don't have concrete suggestions on how to improve this explanation but here are some points I noticed while trying to understand it:

- `a model has an adapter` vs. `a complete model` - are they the same model? Then it would make sense to make that clear
- the first paragraph is basically the second one but the second one states that it is 'incorrect behavior' but doesn't state what the incorrect behavior is, maybe we can be a bit clearer here?

githubnemo (approved):
Thanks for the fix :)

If I understand correctly this will break repos that are downloaded into a directory (e.g., via CLI instead of `hub_download`) and contain a model and an adapter where the adapter is not based on the model in the same repo. I think that is niche enough to not be a concern (and it can be fixed by the user using subfolders).

Some nits about the comments but otherwise LGTM

gabe-l-hart:
>and contain a model and an adapter where the adapter is not based on the model in the same repo.

Yes, this is definitely true. I'm not sure those examples would actually work as-is today though since this clause would just redirect to the hub model, download it, and ignore the local (different) model.

src/transformers/integrations/peft.py
gabe-l-hart: Good call, I'll expand this comment in all three places

tests/models/auto/test_modeling_auto.py
gabe-l-hart: Good call, this was definitely confusing. I've clarified a bit (hopefully)

githubnemo:
@Rocketknight1 this is good to go from my side

tests/pipelines/test_pipelines_common.py
```diff
@@ -213,6 +214,76 @@ def test_auto_model_pipeline_registration_from_local_dir(self):
...
+        # Under the hood: we now store a generation config in the pipeline. This generation config stores the
+        # task-specific parameters.
+        self.assertTrue(pipe.generation_config.num_beams == 4)
+
+        # We can confirm that the task-specific parameters have an effect. (In this case, the default is `num_beams=1`,
+        # which would crash when `num_return_sequences=4` is passed.)
+        pipe("Hugging Face doesn't sell hugs.", num_return_sequences=4)
+        with self.assertRaises(ValueError):
+            pipe("Hugging Face doesn't sell hugs.", num_return_sequences=4, num_beams=1)
+
```
Rocketknight1: Remove unused translation test after merge
```suggestion
```

Rocketknight1:
@bot /style

Rocketknight1:
Merging!

PR #43785, opened 2026-02-06
Fix FSDP_CPU_RAM_EFFICIENT_LOADING (#43749)

### LINKED ISSUE #43749: FSDP_CPU_RAM_EFFICIENT_LOADING broken
### System Info

- `transformers` version: 5.0.0
- Platform: Linux-6.14.0-37-generic-x86_64-with-glibc2.39
- Python version: 3.12.3
- Huggingface_hub version: 1.3.4
- Safetensors version: 0.7.0
- Accelerate version: 1.12.0
- Accelerate config: 	not found
- PyTorch version (accelerator?): 2.9.1+cu129 (CUDA)
- Using distributed or parallel set-up in script?: manual launcher
- Using GPU in script?: yes
- GPU type: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition

### Who can help?

@ArthurZucker @Cyrilvallez

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

1. Load a large model using FSDP2 and FSDP_CPU_RAM_EFFICIENT_LOADING=True. I don't think this is model-specific, but I ran into this with Qwen/Qwen3-30B-A3B-Instruct-2507
2. Watch the ram usage of all of the ranks
3. See that all ranks temporarily use an amount of CPU RAM that's the size of the model being loaded
4. See that the system gets stuck for a long time right after the weights are loaded

I think there are two separate but related issues happening here:
- The model loading code doesn't check for FSDP_CPU_RAM_EFFICIENT_LOADING until _move_missing_keys_from_meta_to_device, which is after the model is already loaded. At this point it discards the loaded weights and inserts new empty tensors
- These empty tensors are seen as uninitialized and will all undergo random initialization. Since they're loaded on the CPU this can take a very long time

This diff causes both of these issues to go away for me. Setting _is_hf_initialized on the new tensors seems like it might be the intended way of handling the second. I'm not sure about the first; I just used a kludge where it skips all of the loading entirely. This works but causes a lot of loud warnings about missing keys. I had tried changing get_device() to return "meta" but this crashed quickly and I didn't investigate further.

```diff
--- modeling_utils.py.orig      2026-02-04 06:59:59.107266669 +0000
+++ modeling_utils.py   2026-02-02 05:50:43.453427661 +0000
@@ -497,10 +497,11 @@
 def _load_parameter_into_model(model: "PreTrainedModel", param_name: str, tensor: torch.Tensor):
     """Cast a single parameter or buffer `param_name` into the `model`, with value `tensor`."""
     parent, param_type = get_module_from_name(model, param_name)
     if param_type in parent._parameters and not isinstance(tensor, nn.Parameter):
         tensor = nn.Parameter(tensor, requires_grad=tensor.is_floating_point())
+        tensor._is_hf_initialized = True
     # We need to use setattr here, as we set non-persistent buffers as well with this function (`load_state_dict`
     # does not allow to do it)
     setattr(parent, param_type, tensor)
--- core_model_loading.py.orig  2026-02-04 07:03:09.751218636 +0000
+++ core_model_loading.py       2026-02-04 07:02:54.937066957 +0000
@@ -1134,10 +1134,15 @@
 
     pattern_to_converter = {k: converter for converter in converters for k in converter.source_patterns}
 
     state_dict = sorted(state_dict.items(), key=lambda kv: dot_natural_key(kv[0]))
 
+    from .integrations import is_fsdp_enabled
+    from .modeling_utils import is_local_dist_rank_0
+    if is_fsdp_enabled() and not is_local_dist_rank_0() and hf_quantizer is None:
+        state_dict = []
+
     for original_key, tensor in state_dict:
         # 1. Rename the key according to all renaming pattern and optional weight converter patterns
         renamed_key, source_pattern = rename_source_key(
             original_key, renamings, converters, prefix, meta_model_state_dict
         )
```

### Expected behavior

- Only rank0 allocates a large amount of cpu ram
- The system moves quickly from weights loading to the rest of the initialization

MengAiDev:
- Add _is_hf_initialized flag in _load_parameter_into_model to prevent unnecessary random initialization
- Skip state_dict loading for non-rank0 processes when FSDP is enabled to avoid wasting CPU RAM
- This fixes the issue where all ranks temporarily allocate model-sized CPU RAM and experience long delays

Modified:
- src/transformers/modeling_utils.py: Set _is_hf_initialized=True on new parameters
- src/transformers/core_model_loading.py: Add FSDP check to skip loading on non-rank0

Fixes #43749

@CyrilVallez @ArthurZucker

### DIFF
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -980,6 +980,27 @@ def convert_and_load_state_dict_in_model(
     We build a mapping from the keys obtained by renaming each of the checkpoint keys according to the weight_mapping rules.
     Then we load the tensors into the model, applying any conversion operations as needed.
 
+    # Early return for FSDP CPU RAM efficient loading
+    # When FSDP is enabled with CPU RAM efficient loading, only rank 0 needs to load the weights
+    # for FULL_SHARD and SHARD_GRAD_OP strategies. Non-rank 0 processes will receive the sharded
+    # weights from rank 0 via FSDP's broadcast. This prevents all ranks from temporarily allocating
+    # model-sized CPU RAM.
+    #
+    # For HYBRID_SHARD and HYBRID_SHARD_ZERO2 strategies, all ranks need to load weights as there
+    # are multiple data parallel groups.
+    from .integrations import should_skip_non_rank0_weight_loading
+
+    if should_skip_non_rank0_weight_loading() and load_config.hf_quantizer is None:
+        # Return empty loading info - the weights will be synced from rank 0 via FSDP
+        loading_info = LoadStateDictInfo(
+            missing_keys=set(model.state_dict().keys()),
+            unexpected_keys=set(),
+            mismatched_keys=set(),
+            conversion_errors={},
+            error_msgs=[],
+        )
+        return loading_info, disk_offload_index
+
     The `param_name_to_load` will look like this:
     {
         "model.layers.0.attention.q.weight": # Notice here there is only the first key of the target keys
@@ -1105,12 +1126,6 @@ def convert_and_load_state_dict_in_model(
     pattern_to_converter = {k: converter for converter in converters for k in converter.source_patterns}
 
     state_dict = sorted(state_dict.items(), key=lambda kv: dot_natural_key(kv[0]))
-
-    from .integrations import is_fsdp_enabled
-    from .modeling_utils import is_local_dist_rank_0
-
-    if is_fsdp_enabled() and not is_local_dist_rank_0() and hf_quantizer is None:
-        state_dict = []
 
     for original_key, tensor in state_dict:
         # 1. Rename the key according to all renaming pattern and optional weight converter patterns
--- a/src/transformers/core_model_loading.py
+++ b/src/transformers/core_model_loading.py
@@ -1105,6 +1105,13 @@ def convert_and_load_state_dict_in_model(
     pattern_to_converter = {k: converter for converter in converters for k in converter.source_patterns}
 
     state_dict = sorted(state_dict.items(), key=lambda kv: dot_natural_key(kv[0]))
+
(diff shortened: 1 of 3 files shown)


src/transformers/core_model_loading.py
```diff
@@ -1105,6 +1105,13 @@ def convert_and_load_state_dict_in_model(
     pattern_to_converter = {k: converter for converter in converters for k in converter.source_patterns}
 
     state_dict = sorted(state_dict.items(), key=lambda kv: dot_natural_key(kv[0]))
+
+    from .integrations import is_fsdp_enabled
+    from .modeling_utils import is_local_dist_rank_0
+
+    if is_fsdp_enabled() and not is_local_dist_rank_0() and hf_quantizer is None:
+        state_dict = []
+
```
ArthurZucker: I don't think this is where the skip should happen, you r are creating a thread pool for notthing

src/transformers/modeling_utils.py
```diff
@@ -476,6 +476,10 @@ def _load_parameter_into_model(model: "PreTrainedModel", param_name: str, tensor
     parent, param_type = get_module_from_name(model, param_name)
     if param_type in parent._parameters and not isinstance(tensor, nn.Parameter):
```
ArthurZucker, line 495: this is only used in `_move_missing_keys_from_meta_to_device` for missing keys, it does not really make sense for me can you elaborate on why?

winglian:
I don't think this behavior should happen at this low of a level. There are a whole host of other cases where this would break ND-parallel loading (HSDP for example), where you would have FSDP, but have multiple data parallel meshes that each would need the model weights too.

Cyrilvallez:
Agreed, everything should simply be skipped both in `_initialize_missing_keys`

PR #43787, opened 2026-02-06
Remove unnecessary code or checks for PT 2.4+

cyyever:
# What does this PR do?

This PR removes or replaces code to use APIs from PyTorch 2.4+.


src/transformers/pytorch_utils.py
```diff
@@ -49,17 +49,6 @@
...
-def softmax_backward_data(parent, grad_output, output):
-    """
-    A function that calls the internal `_softmax_backward_data` PyTorch method and that adjusts the arguments according
-    to the torch version detected.
-    """
-
-    from torch import _softmax_backward_data
-
-    return _softmax_backward_data(grad_output, output, parent.dim, output.dtype)
-
```
Rocketknight1, line 61: Even though we don't use them in the codebase anymore, these utils should probably be left for remote code models that copied library code!

cyyever: I see

src/transformers/pytorch_utils.py
```diff
@@ -201,15 +201,6 @@ def forward(self, hidden_states):
     return forward_fn(*input_tensors)
 
 
-def meshgrid(*tensors: torch.Tensor | list[torch.Tensor], indexing: str | None = None) -> tuple[torch.Tensor, ...]:
-    """
-    Wrapper around torch.meshgrid to avoid warning messages about the introduced `indexing` argument.
-
-    Reference: https://pytorch.org/docs/1.13/generated/torch.meshgrid.html
-    """
-    return torch.meshgrid(*tensors, indexing=indexing)
```
Rocketknight1, line 210: Should have been clearer - this one should probably stay too, for the same remote model reasons!

cyyever: Sure, restore it.

Rocketknight1 (approved):
Yes, LGTM now. Thank you!

PR #43791, opened 2026-02-06
fix: Prevent AutoTokenizer type mismatch from directory name substrin…

tarekziade:
When saving a tokenizer to a local directory and reloading it, the tokenizer type could change to an incorrect class (or fall back to TokenizersBackend) if the directory name contained a model type substring.

Example:
```python
tokenizer = AutoTokenizer.from_pretrained("bert-base-cased")
tokenizer.save_pretrained("./dumptruck")  # Contains "mpt"
new_tokenizer = AutoTokenizer.from_pretrained("./dumptruck")
type(new_tokenizer)  # TokenizersBackend (WRONG! Should be BertTokenizer)
```

This affected any directory name containing model type substrings like:
- "dumptruck" → matched "mpt"
- "gpt2-test" → matched "gpt2"
- "roberta_v2" → matched "roberta"

A cascade failure involving two components:

1. **AutoConfig substring matching**: When loading a config without an explicit model_type field, AutoConfig would perform substring matching on the path to infer the model type. This caused false positives for local paths.

2. **AutoTokenizer mismatch handling**: When the (incorrectly) inferred model_type didn't match the saved tokenizer_class, AutoTokenizer would fall back to the generic TokenizersBackend instead of trusting the explicitly saved tokenizer_class.

### DIFF
--- a/src/transformers/models/auto/configuration_auto.py
+++ b/src/transformers/models/auto/configuration_auto.py
@@ -20,6 +20,8 @@ from collections import OrderedDict
 from collections.abc import Callable, Iterator, KeysView, ValuesView
 from typing import Any, TypeVar
 
+from huggingface_hub import is_offline_mode, model_info
+
 from ...configuration_utils import PreTrainedConfig
 from ...dynamic_module_utils import get_class_from_dynamic_module, resolve_trust_remote_code
 from ...utils import CONFIG_NAME, logging
@@ -1270,6 +1272,39 @@ def replace_list_option_in_docstrings(
     return docstring_decorator
 
 
+def _infer_model_type_from_tags(model_id: str) -> str | None:
+    """
+    Infer model type from repository tags on the Hub.
+
+    Args:
+        model_id: The model repository ID (e.g., "org/model-name")
+
+    Returns:
+        The inferred model type if found in tags, None otherwise
+    """
+    try:
+        model = model_info(model_id)
+        if model.tags is not None:
+            # Check for base_model tag pattern (e.g., "base_model:mistralai/Mistral-7B")
+            for tag in model.tags:
+                if tag.startswith("base_model:"):
+                    # Extract the base model and try to infer type from it
+                    base_model = tag.split(":", 1)[1]
+                    # Check if any model type appears in the base model tag
+                    for model_type in sorted(CONFIG_MAPPING.keys(), key=len, reverse=True):
+                        if model_type in base_model.lower():
+                            return model_type
+
+                # Also check for direct model type tags
+                if tag in CONFIG_MAPPING.keys():
+                    return tag
+    except Exception as e:
+        # If we can't fetch model info (offline, network error, etc.), return None
+        logger.debug(f"Can't fetch model info for {model_id}: {e}")
+
+    return None
+
+
 class AutoConfig:
     r"""
     This is a generic configuration class that will be instantiated as one of the configuration classes of the library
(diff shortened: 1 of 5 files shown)


src/transformers/models/auto/configuration_auto.py
```diff
@@ -1421,10 +1421,18 @@ def from_pretrained(cls, pretrained_model_name_or_path: str | os.PathLike[str],
             return config_class.from_dict(config_dict, **unused_kwargs)
         else:
             # Fallback: use pattern matching on the string.
-            # We go from longer names to shorter names to catch roberta before bert (for instance)
-            for pattern in sorted(CONFIG_MAPPING.keys(), key=len, reverse=True):
-                if pattern in str(pretrained_model_name_or_path):
-                    return CONFIG_MAPPING[pattern].from_dict(config_dict, **unused_kwargs)
+            # Only apply this to remote repository identifiers (containing '/') to avoid false matches
+            # on local directory names like "dumptruck" containing "mpt" or "gpt2-test" containing "gpt2"
+            pretrained_str = str(pretrained_model_name_or_path)
```
ArthurZucker: I am not sure we event want this feature anymore 😢 Usually the remote repos have tags, we should better check them instead maybe? (like for the patch mistral regex thing)

tarekziade: good point

Do you want to remove the substring matching entirely and just fail if there's no model_type in config,
or call the hub api for tags here? like mistral

tarekziade: I tried the tag checking way

tarekziade:
run-slow: auto

src/transformers/models/auto/configuration_auto.py
tarekziade: I don't know how safe tags string are so the code is extra defensive here, let me know if its the right approach, or maybe this snippet already exists somewhere as a util?

ArthurZucker: yeah I mean it might not be the best either as we don't want to do more requests to the hub that the original one in general + `is_local` flags and etc !

ArthurZucker: but yes sub string matching should go away IMO

tarekziade:
run-slow: auto

ArthurZucker (approved):
Aria's failure just seems to be default padding side being different but its probably just something we can change in expected values

tarekziade:
run-slow: aria, auto

tests/models/auto/test_tokenization_auto.py
```diff
@@ -621,14 +621,16 @@ def test_tokenization_class_priority(self):
             tok2 = AutoTokenizer.from_pretrained(tmp_dir)
             self.assertTrue(tok2.__class__ == HerbertTokenizer)
 
+        # SmolLM2 has tokenizer_class=GPT2Tokenizer explicitly saved, so we should use that
+        # instead of falling back to TokenizersBackend
         tok = AutoTokenizer.from_pretrained("HuggingFaceTB/SmolLM2-135M-Instruct")
-        self.assertTrue(tok.__class__ == TokenizersBackend)
+        self.assertTrue(tok.__class__ == GPT2Tokenizer)
```
vasqu: Isn't that exactly what we wanted to avoid? A lot of configs save the "wrong" class in a sense that they meant to use the tokenizers backend while pointing out to GPT2 or Llama for example

The goal was to avoid these exceptions when we put them into auto ourselves so that we are 100% sure it's that explicit tokenizer

ArthurZucker: yep 💯

tests/models/auto/test_tokenization_auto.py
```diff
@@ -696,3 +698,80 @@ def __init__(self, **kwargs):
...
             finally:
                 os.chdir(prev_dir)
+
+    def test_tokenizer_save_load_with_substring_patterns(self):
+        # Load a BERT tokenizer
+        original_tokenizer = AutoTokenizer.from_pretrained("google-bert/bert-base-cased")
+        original_class = type(original_tokenizer)
+
+        # Test cases: directory names that contain model type substrings
+        test_cases = [
```
Rocketknight1: I don't think we need the regression tests if we're deleting that block! It seems very unlikely that we'll regress on this when we're totally removing the relevant code, and adding unnecessary tests might create more CI problems later 😅

Rocketknight1 (approved):
Overall looks great! We can probably remove the changes in `test_tokenization_auto.py` though, right?

src/transformers/models/auto/configuration_auto.py
```diff
@@ -1421,17 +1421,10 @@ def from_pretrained(cls, pretrained_model_name_or_path: str | os.PathLike[str],
                     "`pip install git+https://github.com/huggingface/transformers.git`"
                 )
             return config_class.from_dict(config_dict, **unused_kwargs)
-        else:
-            # Fallback: use pattern matching on the string.
-            # We go from longer names to shorter names to catch roberta before bert (for instance)
-            for pattern in sorted(CONFIG_MAPPING.keys(), key=len, reverse=True):
-                if pattern in str(pretrained_model_name_or_path):
-                    return CONFIG_MAPPING[pattern].from_dict(config_dict, **unused_kwargs)
```
tomaarsen, line 1429: This is breaking, right? We need to mention this explicitly in https://github.com/huggingface/transformers/releases/tag/v5.2.0, cc @ArthurZucker @LysandreJik 

For example, this now fails:
```python

model = AutoModel.from_pretrained("prajjwal1/bert-tiny")
```

ArthurZucker: yeah, for a super small minority of models but yeah. Let's put it in front sorry

ArthurZucker: image

PR #43098, opened 2026-01-04
[Model] Add PP-DocLayoutV3 Model Support

zhang-prog:

PR #43793, opened 2026-02-06
[Model] Add PP-OCRV5_mobile_rec Model Support

liu-jiaxuan:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


src/transformers/models/pp_ocrv5_mobile_rec/modular_pp_ocrv5_mobile_rec.py
```diff
@@ -0,0 +1,2153 @@
...
+        self.drop_prob = drop_prob
+
+    def forward(self, x):
+        if self.drop_prob == 0.0 or not self.training:
+            return x
+        keep_prob = 1 - self.drop_prob
+        shape = (x.shape[0],) + (1,) * (x.ndim - 1)
+        random_tensor = keep_prob + torch.rand(shape, dtype=x.dtype, device=x.device)
+        random_tensor.floor_()
+        return x.div(keep_prob) * random_tensor
```
yonigozlan: Is this really used anywhere? seems to always be 0, with no way to change it. If that's the case let's remove entirely

src/transformers/models/pp_ocrv5_mobile_rec/modular_pp_ocrv5_mobile_rec.py
```diff
@@ -0,0 +1,2153 @@
...
+            is_remove_duplicate=True,
+        )
+
+        texts = []
+        scores = []
+        for t in text:
+            texts.append(t[0])
+            scores.append(t[1])
+
+        return texts, scores
```
yonigozlan, line 2187: Let's avoid cv2 as a dependency. We use PiL for "slow" image processors, and torch/torchvision for fast ones

yonigozlan (commented):
Hello @liu-jiaxuan! Thanks for opening this PR, however there is quite a bit to change here to fit the standards of the Transformers library.

The biggest issue is that you've written everything from scratch without inheriting from existing models. The modular file should maximize inheritance. Even if this is a novel architecture (especially the Conv modules part, which might not exist elsewhere in the library), components like MLP blocks, attention, and layer norms should use standard library patterns by inheriting form an existing model's module in modular.

The novel modules that can't be inherited through modular should also follow library standards in terms of naming, formatting, structure and good-practices ("PPOCRV5MobileRec" prefix for all module names, weight names standardized with other similar modules in the library, no single letter variables, type hints, docstrings when args are not standards or obvious, never use "eval()" etc.), and the model should support as much transformers features as possible, such as the attention interface through flags in PreTrainedModel( _supports_attention_backend, _supports_sdpa, _supports_flash_attn etc.)

Some other big things wrong or missing:
- We shouldn't have a cv2 dependency in image processors, "slow" should use PiL/numpy functions, fast torch/torchvision. 
- Weight initialization shouldn't be scattered in individual module constructors but centralized in _init_weights() on the PreTrainedModel class, and use the transfromers "init" module.
- Attention modules are standardized across models in the transformers library, so using modular for attention modules is a must.

Before we go deeper in reviewing this new model addition (and other Paddle Paddle ones open recently that are very similar), please have a good look at how other models are implemented in the library. Notably, you can have a look at the recently merged [PP-DocLayoutV3 PR](https://github.com/huggingface/transformers/pull/43098) (here's its [modular file](https://github.com/huggingface/transformers/blob/5ba28289d7f04f1b0d0c24bb596ab972b32f0d49/src/transformers/models/pp_doclayout_v3/modular_pp_doclayout_v3.py).
We also have resources to learn more about how to contribute a new model and how to use modular: [Contributing a new model](https://huggingface.co/docs/transformers/en/modular_transformers), [using modular](https://huggingface.co/docs/transformers/v4.45.1/en/modular_transformers).

Also as the multiple Paddle Paddle models that have a new model addition PR open currently seem to be quite similar, I'd recommend focusing on one (the simplest) for now, then we'll be able to leverage modular to easily add the other models.

Happy to answer any questions you may have!

liu-jiaxuan:
Hello @yonigozlan , thank you very much for your detailed review and valuable guidance!
We have revised the three models (pp_ocrv5_mobile_rec, [pp_ocrv5_server_rec](https://github.com/huggingface/transformers/pull/43795), and [slanext](https://github.com/huggingface/transformers/pull/43707)) to address the issues you mentioned.

Specifically, we have implemented the following improvements:
1. Removed cv2 dependency: We replaced cv2 in image preprocessing and switched to numpy-based image processing. However, since all three models are designed for text or table recognition tasks and are highly sensitive to pixel-level perturbations in images, replacing cv2 operations has caused some impact on model accuracy.
2. Centralized weight initialization: We moved all weight initialization to the _init_weights() method of the PreTrainedModel class, and removed the initialization code from the constructors of individual modules.
3. Refactored inheritance and naming: Modified classes that could be modularly inherited to extend existing models, removed standalone implementations for modules that PyTorch supports directly, and added model prefixes (e.g., PPOCRV5MobileRec) to classes that cannot be directly inherited in a modular fashion.
4. Removed unused functionality from various model modules, such as the DropPath class you pointed out.

We will continue to refine the code according to the transformers library standards and conventions. Please let us know if you have any further comments or suggestions.
Thank you again for your help!

src/transformers/models/pp_ocrv5_mobile_rec/modular_pp_ocrv5_mobile_rec.py
liu-jiaxuan: Done

src/transformers/models/pp_ocrv5_mobile_rec/modular_pp_ocrv5_mobile_rec.py
liu-jiaxuan: Done.

yonigozlan:
Thanks a lot for iterating @liu-jiaxuan ! I'll have a look in the coming days.

> We replaced cv2 in image preprocessing and switched to numpy-based image processing. However, since all three models are designed for text or table recognition tasks and are highly sensitive to pixel-level perturbations in images, replacing cv2 operations has caused some impact on model accuracy.

We do support using PIL.resize in slow processor and torchvision resize in fast, maybe you'll get closer results with this when choosing the equivalent interpolation than the one in cv2 than with custom numpy code?

liu-jiaxuan:
> Thanks a lot for iterating @liu-jiaxuan ! I'll have a look in the coming days.
> 
> > We replaced cv2 in image preprocessing and switched to numpy-based image processing. However, since all three models are designed for text or table recognition tasks and are highly sensitive to pixel-level perturbations in images, replacing cv2 operations has caused some impact on model accuracy.
> 
> We do support using PIL.resize in slow processor and torchvision resize in fast, maybe you'll get closer results with this when choosing the equivalent interpolation than the one in cv2 than with custom numpy code?

Hi @yonigozlan, thank you very much for your suggestion! Based on our current experiments, using PIL/torchvision for image preprocessing in these three models results in a larger accuracy loss compared to numpy. Therefore, we have chosen the numpy-based approach. We will continue to iterate on the PIL/torchvision-based preprocessing method, and we will update the PR immediately if we achieve a better version.

vasqu:
Closing in favor of #44808

PR #43794, opened 2026-02-06
Enforce manual seed to reduce flakiness

tarekziade:
This patch aims to reduce flakiness in CI tests. We identified two causes of nondeterministic behavior:

- Some tests were not using a fixed RNG seed, which reduced determinism.
- The cli tests were occasionally triggering I/O errors due to writes on a closed stdout.

This branch was run multiple times and appears to reduce flakiness in all previously unseeded tests. While there’s no deterministic way to prove the improvement, using fixed seeds is still a best practice.

As a follow-up, we could centralize the seed initialization in a shared test fixture, avoiding the need to set it explicitly in individual tests.


tarekziade:
run-slow: doge, donut, esm, fastspeech2_conformer, mimi, minimax_m2, mistral, mixtral, musicgen, musicgen_melody, nllb_moe, qwen2, qwen2_moe, qwen3, qwen3_moe, recurrent_gemma

tests/models/seamless_m4t/test_modeling_seamless_m4t.py
```diff
@@ -515,10 +515,60 @@ def test_attention_outputs(self):
             )
 
     def test_retain_grad_hidden_states_attentions(self):
-        # When training the model, the first speech encoder layer is sometimes skipped.
-        # Setting the seed to always have the first layer.
-        set_seed(0)
-        super().test_retain_grad_hidden_states_attentions()
+        # When training the model, speech encoder layerdrop can skip layers and return None attentions.
+        # Disable layerdrop for this test to ensure retain_grad targets are present.
```
Rocketknight1: This fix is fine, but maybe we could just disable layerdrop in the test config for this model to reduce LOC changes? It seems like an obvious source of massive variability, and layerdrop is never used in modern models.

Rocketknight1 (approved):
Fix LGTM! I'm in favour of blindly setting seeds everywhere, especially in tests where we compare two model outputs.

PR #43797, opened 2026-02-06
preventing I/O errors on closed streams in the `cli` helper

tarekziade:
Let's see if we can fix that flaky


tarekziade:
(this branch inherits from `tarekziade-flaky-test_generate` ) 

only the last commit matters here

tarekziade:
merged back in https://github.com/huggingface/transformers/pull/43794

PR #43799, opened 2026-02-06
[Trainer] use deepspeed SP process group when Accelerate doesn’t build a mesh

kashif:
# What does this PR do?

With this change: https://github.com/huggingface/accelerate/pull/3916 there is no torch device mesh when sp_backend="deepspeed", and transformers currently assumes it exists.
  That’s why you get:

```
  sp_group = self.accelerator.torch_device_mesh["sp"].get_group()
  TypeError: 'NoneType' object is not subscriptable
```

With this change, it uses DeepSpeed’s SP process group when Accelerate doesn’t build a mesh.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


kashif:
cc @stas00 if you can kindly check?

src/transformers/trainer.py
```diff
@@ -3919,8 +3953,20 @@ def _deepspeed_sp_compute_loss(self, model, inputs, return_outputs, pc):
...
+        # Use torch device mesh when available; fallback to DeepSpeed SP groups only for Ulysses SP
+        if self.accelerator.torch_device_mesh is not None:
+            sp_group = self.accelerator.torch_device_mesh["sp"].get_group()
+            sp_world_size = pc.sp_size
+        elif pc.sp_backend == "deepspeed" and pc.sp_size > 1:
+            from deepspeed.utils import groups
+
+            sp_group = groups._get_sequence_parallel_group()
+            sp_world_size = groups._get_sequence_parallel_world_size()
+        else:
```
SunMarc: Maybe switch the order so that we first check if we are using deepspeed ? I'm afraid of silent bug where `torch_device_mesh` is actually defined.

SunMarc (approved):
Thanks for fixing this ! Just a nit

src/transformers/trainer.py
kashif: ok sure fixing

winglian: in that particular case where both are defined, would it be preferable to lot an error or raise an exception?

PR #43801, opened 2026-02-06
Fix lifted_tensor in Gemma3n export which dynamo can't reason about

robell:
When using the torch.export path with this model we hit an issue on correction_coefs being modified with unknown side-effects the dynamo can't catch. So it lifts this tensor unnecessarily, and fails assuming mutation is needed RuntimeError: Constant model.model.language_model.layers.0.lifted_tensor_4 is mutated in the forward method. Pls register it as buffer.

This change makes the subsequent use recognisable as an out variant to prevent this.

As this is a minor adjustment with no functionality added I've just checked that the basic flow and gemma3n tests pass as in the guide:
```
pip install -e ".[dev]"
python utils/modular_model_converter.py gemma3n
RUN_SLOW=1 pytest tests/models/gemma3n/
make style
make check-repo
```

Hope this needs attn of:
- multimodal models: @zucchini-nlp


Rocketknight1 (approved):
Yes, thank you! Mutating a weight tensor like that is kinda gross, and this is obviously preferable. Shouldn't we keep the `self.training` check too, though?

src/transformers/models/gemma3n/modeling_gemma3n.py
```diff
@@ -1142,13 +1142,15 @@ def correct(self, predictions: torch.Tensor, activated: torch.Tensor) -> torch.T
         innovation = activated - predictions[self.config.altup_active_idx]  # (batch, num_tokens, hidden_size)
         innovation = innovation.repeat(self.config.altup_num_inputs, 1, 1, 1)  # Repeat on dim0 to match predictions
 
-        if self.training and self.config.altup_coef_clip is not None:
-            self.correction_coefs.weight.data.clamp_(-self.config.altup_coef_clip, self.config.altup_coef_clip)
+        if self.config.altup_coef_clip is not None:
+            weight = self.correction_coefs.weight.clamp(-self.config.altup_coef_clip, self.config.altup_coef_clip)
+            all_coefs = torch.nn.functional.linear(modalities, weight, bias=None) + 1.0
+        else:
+            all_coefs = self.correction_coefs(modalities) + 1.0
```
zucchini-nlp, line 1151: iirc deepmind team wanted to keep `correction_coefs` within a range when training, tho I don't remember why it wasn't updated to smth better before release. I am not sure if ti anyhow affects training tho, happy to keep this diff

robell:
Thanks - I've reintroduced the training check just to be on the safe side.

src/transformers/models/gemma3n/modeling_gemma3n.py
robell: addressed this to be on the safe side. Anything else to do or can this be merged?

zucchini-nlp:
run-slow: gemma3n

src/transformers/models/gemma3n/modeling_gemma3n.py
zucchini-nlp: we can merge when slow CI checks pass :)

robell:
hi @Rocketknight1, can we merge this now? Thanks. :)

zucchini-nlp:
Yep, let's merge!

PR #43803, opened 2026-02-06
[`Repo Consistency`] Fix rms norm

vasqu:
As per title, I merged #43772 after #43400 which caused this mismatch - causes red CI


ebezzam (commented):
👍

ebezzam (approved):

PR #43804, opened 2026-02-06
fix(flaky): Different approach to make sure loss exists

tarekziade:
# What does this PR do?

The difference check was returning False 40% of the times and was reproducible locally

tested with 

```
pytest -svx tests/models/ernie4_5_moe/test_modeling_ernie4_5_moe.py -k test_load_balancing_loss --flake-finder
```
moved to a deterministic approach


tarekziade:
run-slow: ernie4_5_moe

tests/models/ernie4_5_moe/test_modeling_ernie4_5_moe.py
```diff
@@ -111,22 +114,39 @@ def test_load_balancing_loss(self):
...
         padded_result = model(padded_input_ids, attention_mask=padded_attention_mask)
         torch.testing.assert_close(result.aux_loss.cpu(), padded_result.aux_loss.cpu(), rtol=1e-4, atol=1e-4)
 
-        # We make sure that the loss of including padding tokens != the loss without padding tokens
-        # if attention_mask=None --> we don't exclude padding tokens
-        include_padding_result = model(padded_input_ids, attention_mask=None)
-
-        # This is to mimic torch.testing.assert_not_close
-        self.assertNotAlmostEqual(include_padding_result.aux_loss.item(), result.aux_loss.item())
+        # We make sure that masking can change the loss using a deterministic synthetic example.
```
tarekziade, line 133: I assume calling `load_balancing_loss_func` is good enough to validate loss diff, and it could be its own test. But it also ignore the `model()` path. is this ok?

tarekziade:
may be related https://github.com/huggingface/transformers/pull/43775

tests/models/ernie4_5_moe/test_modeling_ernie4_5_moe.py
ArthurZucker: This looks alright!

ArthurZucker (approved):
ty

tarekziade:
run-slow: ernie4_5_moe

PR #43568, opened 2026-01-28
Simplify TrainingArguments docstring

SunMarc:
# What does this PR do?

This PR makes it easier to go through TrainingArguments ! This should help making trainer less bloated for the users. A nice first step to make Trainer simpler to use.

PR #43806, opened 2026-02-06
Update TrainingArguments

SunMarc:
# What does this PR do?

This PR updates TrainingArguments in several places:
- More tests
- Simplify the post_init  
- update and reorder __init__

For the review, just check that the post_init is correctly modified


src/transformers/training_args.py
```diff
@@ -107,26 +107,6 @@
...
-    """
-    Returns the xla device type (CPU|GPU|TPU) or None if the device is a non-xla device.
-    """
-    if is_torch_xla_available():
-        if device.type == "cpu":
-            return "CPU"
-        return xm.xla_real_devices([device])[0].split(":")[0]
-    return None
-
-
```
SunMarc, line 129: not used anywhere

qgallouedec:
https://github.com/huggingface/trl/actions/runs/21761955843

src/transformers/training_args.py
```diff
@@ -1617,85 +1586,117 @@ def __post_init__(self):
...
+            if not is_torch_available():
+                raise ValueError("lr_scheduler_type reduce_lr_on_plateau requires torch>=0.2.0")
+
+        if self.warmup_steps < 0:
+            raise ValueError("warmup_steps must be an integer or a float")
 
         if self.dataloader_num_workers == 0 and self.dataloader_prefetch_factor is not None:
             raise ValueError(
                 "--dataloader_prefetch_factor can only be set when data is loaded in a different process, i.e."
                 " when --dataloader_num_workers > 1."
```
qgallouedec:
```suggestion
                "--dataloader_prefetch_factor can only be set when data is loaded in a different process, i.e."
                " when --dataloader_num_workers > 0."
```

qgallouedec (approved):
lgtm, just one suggestion

qgallouedec:
https://github.com/huggingface/trl/actions/runs/21765706593

src/transformers/training_args.py
SunMarc: thanks !

src/transformers/training_args.py
```diff
@@ -1617,83 +1586,115 @@ def __post_init__(self):
...
-            # Import at runtime to avoid a circular import.
-            from .integrations import get_available_reporting_integrations
-
-            self.report_to = get_available_reporting_integrations()
-
-            if "codecarbon" in self.report_to and torch.version.hip:
-                logger.warning(
-                    "When using the Trainer, CodeCarbonCallback requires the `codecarbon` package, which is not compatible with AMD ROCm (https://github.com/mlco2/codecarbon/pull/490). Automatically disabling the codecarbon callback. Reference: https://huggingface.co/docs/transformers/v4.39.3/en/main_classes/trainer#transformers.TrainingArguments.report_to."
-                )
-                self.report_to.remove("codecarbon")
```
tomaarsen, line 1632: Why did we remove this? `report_to="all"` went from the default in #41375 to now no longer possible. The updated docs from #43568 also state that `"all"` should still be possible. We need to be a bit careful with these refactors/changes.

SunMarc: Yeah I also saw the docs, we need to update them ! We changed `report_to="all"` to `"none"` as this was something that was supposed to be done in v5. 

```python 
        if self.report_to is None:
            logger.info(
                "The default value for the training argument `--report_to` will change in v5 (from all installed "
                "integrations to none). In v5, you will need to use `--report_to all` to get the same behavior as "
                "now. You should start updating your code and make this info disappear :-)."
            )
            self.report_to = "all"
```

tomaarsen: I'm not sure if this has been fixed since my comment, but I believe `report_to="all"` fails currently, due to the diff I commented on being removed. I think it might have been accident!

SunMarc: Oh okay sorry I misunderstood. Indeed, with this, "all" don't work anymore due to this incident. I'll add it back ;/

PR #43807, opened 2026-02-06
Refactor trainer init

SunMarc:
# What does this PR do?

This PR simplifies Trainer `__init__`: 

- Quantization validation extracted 
- PEFT unwrapping deduplicated
-  Liger Kernel extracted — apply_liger_kernel
-  Label smoother simplified
- Validations grouped — `_validate_args()` method consolidates three scattered validation blocks (arg checks, optimizer checks, dataset checks) into one place
- and more


src/transformers/integrations/liger.py
```diff
@@ -0,0 +1,53 @@
...
+
+    from liger_kernel.transformers import _apply_liger_kernel_to_instance
+
+    kernel_config = args.liger_kernel_config if args.liger_kernel_config is not None else {}
+    base_model = unwrap_peft_model(model)
+
+    if isinstance(base_model, PreTrainedModel):
+        _apply_liger_kernel_to_instance(model=base_model, **kernel_config)
+    else:
+        logger.warning("The model is not an instance of PreTrainedModel. No liger kernels will be applied.")
```
SunMarc, line 53: simpler now

src/transformers/trainer.py
```diff
@@ -546,6 +485,19 @@ def __init__(
...
+                raise ValueError("SageMaker Model Parallelism does not support BF16 yet. Please use FP16 instead ")
+            if args.fp16 != smp.state.cfg.fp16:
+                logger.warning(
+                    f"FP16 provided in SM_HP_MP_PARAMETERS is {smp.state.cfg.fp16}, "
+                    f"but FP16 provided in trainer argument is {args.fp16}, "
+                    f"setting to {smp.state.cfg.fp16}"
+                )
+                args.fp16 = smp.state.cfg.fp16
+
+        # ---- 5. Data ----------------------------------------------------------------
```
winglian: this feels more like an args validation check that should go up to #1

qgallouedec:
> /trl-ci

PAT is not yet approved for this bot, I ran manually here https://github.com/huggingface/trl/actions/runs/21760373311

src/transformers/trainer.py
```diff
@@ -546,18 +484,20 @@ def __init__(
         ):
             self.place_model_on_device = False
```
winglian, line 481: this pattern feels a bit off. `args.place_model_on_device` is a `@property` on args that calculates based on `not is_sagemaker_mp_enabled()`. A developer reading this might assume they would have some control over this arg and would want to manually affect this state. I'm pretty sure there are a whole host of edge cases as there are also about 5 `or` cases above that we currently account for, so making this settable from the `TrainingArguments` would be a nice to have.

winglian (approved):
Generally, I love seeing this method being more compact 🤗. Added some additional thoughts, but can get addressed separate from this PR as well

src/transformers/integrations/liger.py
```diff
@@ -0,0 +1,53 @@
...
+
+from ..modeling_utils import PreTrainedModel
+from ..trainer_utils import unwrap_peft_model
+from ..utils import is_liger_kernel_available, logging
+
+
+logger = logging.get_logger(__name__)
+
+
+def apply_liger_kernel(model, args):
```
qgallouedec:
```suggestion
def apply_liger_kernel(model, kernel_config):
```

src/transformers/integrations/liger.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Apply Liger Kernel optimizations to a model instance.
+
+    Liger Kernel provides optimized Triton kernels for common transformer operations.
+    This function patches the model in-place with those kernels.
+
+    Args:
+        model: The model to patch. Must be a `PreTrainedModel` or a PEFT wrapper around one.
+        args: Training arguments. Uses `args.liger_kernel_config` for kernel configuration.
```
qgallouedec:
```suggestion
        kernel_config: Kernel configuration.
```

src/transformers/integrations/liger.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    if not is_liger_kernel_available():
+        raise ImportError(
+            "You have set `use_liger_kernel` to `True` but liger-kernel >= 0.3.0 is not available. "
+            "Please install it with `pip install liger-kernel`"
+        )
+
+    from liger_kernel.transformers import _apply_liger_kernel_to_instance
+
+    kernel_config = args.liger_kernel_config if args.liger_kernel_config is not None else {}
```
qgallouedec:
```suggestion
    kernel_config = kernel_config or {}
```

src/transformers/trainer.py
```diff
@@ -458,6 +450,12 @@ def __init__(
                 "https://huggingface.co/docs/transformers/model_doc/auto"
             )
 
+        if self.args.use_liger_kernel:
+            apply_liger_kernel(model, self.args)
```
qgallouedec:
```suggestion
            apply_liger_kernel(model, self.args.liger_kernel_config)
```

src/transformers/trainer.py
```diff
@@ -2066,14 +2012,9 @@ def train(
...
-        # the following is a workaround:
-        if (
-            (args.fp16_full_eval or args.bf16_full_eval)
-            and not args.do_train
-            and not self.is_model_parallel
-            and self.model_init is None
-        ):
+        # When fp16/bf16 full eval is enabled, __init__ skips device placement so that
+        # evaluation_loop can cast dtype and move in one step. Move the model now for training.
+        if (args.fp16_full_eval or args.bf16_full_eval) and not self.is_model_parallel and self.model_init is None:
```
qgallouedec, line 1597: we don't need `not args.do_train` annymore?

src/transformers/trainer.py
SunMarc: Indeed, this is a bit off. Let me fix this ! Note that for now, I'm only trying to do simple changes to make it easier to read. But I will start soon cleaning things that feels a bit off in Trainer =)

qgallouedec (approved):
lgtm, just a question and suggestions

src/transformers/trainer.py
SunMarc: `args.do_train` is supposed to be used by external scripts and not Trainer itself. So, I prefer to remove it ! It should be fine to remove as `do_train` default value is `False` and most users don't always set this arg to True (meaning that since there were no issues from the community until now, it should be safe to remove)

src/transformers/trainer.py
```diff
@@ -529,35 +473,35 @@ def __init__(
...
-        # 2. fp16-enabled DeepSpeed loads the model in half the size and it doesn't need .to() anyway,
-        #    and we only use deepspeed for training at the moment
-        # 3. full bf16 or fp16 eval - since the model needs to be cast to the right dtype first
-        # 4. FSDP - same as MP
-        self.place_model_on_device = args.place_model_on_device
-        if (
+        # Postpone switching model to cuda when MP, DeepSpeed, full bf16/fp16 eval, or FSDP
+        if args.place_model_on_device is not None:
+            self.place_model_on_device = args.place_model_on_device
+        elif (
```
SunMarc, line 473: lmk if this is better @winglian. The changes should be BC as the user can still overwrite `place_model_on_device` as a property

qgallouedec:
(just a test, now that the PAT is approved)
/trl-ci

src/transformers/trainer.py
SunMarc: Moved to validation check ! thanks

SunMarc: Moved to validation check ! thanks

SunMarc:
@bot /style

PR #43808, opened 2026-02-06
[docs] reveal more in toctree

stevhliu:
expands the base classes, inference, and training sections as these are quite important and will give readers a better idea of what's available


vasqu (approved):
Nice lgtm

PR #43780, opened 2026-02-05
Fix DeepSpeed model preparation logic in Trainer class

qgallouedec:
The changes in #43711 caused the model to be never prepared when using DeepSpeed. When training you hit for example:

```
[rank0]: Traceback (most recent call last):
[rank0]:   File "/fsx/qgallouedec/trl/trl/scripts/grpo.py", line 193, in <module>
[rank0]:     main(script_args, training_args, model_args, dataset_args)
[rank0]:     ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/fsx/qgallouedec/trl/trl/scripts/grpo.py", line 162, in main
[rank0]:     trainer.train()
[rank0]:     ~~~~~~~~~~~~~^^
[rank0]:   File "/fsx/qgallouedec/transformers/src/transformers/trainer.py", line 2170, in train
[rank0]:     return inner_training_loop(
[rank0]:         args=args,
[rank0]:     ...<2 lines>...
[rank0]:         ignore_keys_for_eval=ignore_keys_for_eval,
[rank0]:     )
[rank0]:   File "/fsx/qgallouedec/transformers/src/transformers/trainer.py", line 2537, in _inner_training_loop
[rank0]:     tr_loss_step = self.training_step(model, inputs, num_items_in_batch)
[rank0]:   File "/fsx/qgallouedec/trl/trl/trainer/grpo_trainer.py", line 1024, in training_step
[rank0]:     output = super().training_step(model, inputs, num_items_in_batch)
[rank0]:   File "/fsx/qgallouedec/transformers/src/transformers/trainer.py", line 3804, in training_step
[rank0]:     inputs = self._prepare_inputs(inputs)
[rank0]:   File "/fsx/qgallouedec/trl/trl/extras/profiling.py", line 202, in wrapper
[rank0]:     return func(self, *args, **kwargs)
[rank0]:   File "/fsx/qgallouedec/trl/trl/trainer/grpo_trainer.py", line 1053, in _prepare_inputs
[rank0]:     generation_batch = self._generate_and_score_completions(generation_batch)
[rank0]:   File "/fsx/qgallouedec/trl/trl/trainer/grpo_trainer.py", line 1551, in _generate_and_score_completions
[rank0]:     ) = self._generate(prompts)
[rank0]:         ~~~~~~~~~~~~~~^^^^^^^^^
[...]

PR #43809, opened 2026-02-06
Add TRL CI bot workflow to trigger tests on PR comments

qgallouedec:
Trainer is going to be heavily refactored over the coming weeks/months (see #43595). To avoid the recurring pattern:

PR merges in `transformers` → TRL CI breaks → follow-up PR in `transformers` to fix something we could have caught earlier
(e.g. #43711 → #43780),

I propose adding a small "slash-command" bot that lets us run TRL’s CI against the exact commit/branch from a `transformers` PR, so we can detect downstream breakages before merging.

To use it, just comment on the PR:

> `/trl-ci`


SunMarc (approved):
Nice idea ! Thanks

PR #43810, opened 2026-02-06
add bidirectional attention to qwen and llama configs

michaelfeil:
# What does this PR do?

Background: I want to remove custom code from some files, e.g. this one https://huggingface.co/nvidia/llama-embed-nemotron-8b/blob/main/llama_bidirectional_model.py or voyageai/voyage-4-nano model. 

They have custom code, which is not necessary.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/transformers/models/llama/configuration_llama.py
+++ b/src/transformers/models/llama/configuration_llama.py
@@ -90,7 +90,8 @@ class LlamaConfig(PreTrainedConfig):
             Whether to use a bias in up_proj, down_proj and gate_proj layers in the MLP layers.
         head_dim (`int`, *optional*):
             The attention head dimension. If None, it will default to hidden_size // num_attention_heads
-
+        use_bidirectional_attention (`bool`, *optional*):
+            If True, the model will attend to all text tokens instead of using a causal mask.
     ```python
     >>> from transformers import LlamaModel, LlamaConfig
 
@@ -145,6 +146,7 @@ class LlamaConfig(PreTrainedConfig):
         attention_dropout: float | None = 0.0,
         mlp_bias: bool | None = False,
         head_dim: int | None = None,
+        use_bidirectional_attention: bool | None = False,
         **kwargs,
     ):
         self.vocab_size = vocab_size
@@ -174,6 +176,7 @@ class LlamaConfig(PreTrainedConfig):
         self.pad_token_id = pad_token_id
         self.bos_token_id = bos_token_id
         self.eos_token_id = eos_token_id
+        self.use_bidirectional_attention = use_bidirectional_attention
         super().__init__(**kwargs)
 
 
--- a/src/transformers/models/llama/modeling_llama.py
+++ b/src/transformers/models/llama/modeling_llama.py
@@ -233,7 +233,7 @@ class LlamaAttention(nn.Module):
         self.num_key_value_groups = config.num_attention_heads // config.num_key_value_heads
         self.scaling = self.head_dim**-0.5
         self.attention_dropout = config.attention_dropout
-        self.is_causal = True
+        self.is_causal = not getattr(config, "use_bidirectional_attention", False)
 
         self.q_proj = nn.Linear(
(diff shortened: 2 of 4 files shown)


vasqu:
Can you check #43705? It should enable your usecase just by passing `is_causal=False` as kwarg

michaelfeil:
@vasqu https://huggingface.co/nvidia/llama-embed-nemotron-8b/blob/main/llama_bidirectional_model.py + https://huggingface.co/nvidia/llama-embed-nemotron-8b/blob/main/config.json#L40 

Background: Embedding models are starting to use this flag more and more, but they are also forking the modeling code for the sake of having bidrectrional attention. Similar to gemma3, i want a custom flag.

vasqu:
Does `is_causal` not suffice? You can either
- pass it as kwarg to the forward
- or set it in the config (and the forward afterwards will be adjusted accordingly)

I'd like to avoid causal specific flags, gemma3 is a bit of a special case. `is_causal` could be extended to most models

vasqu:
I.e. it would only need to update the config to that of the respective llama and set `is_causal=False` as attribute and it will work out of the box without custom code

vasqu:
But also cc @tomaarsen who wants to use this in ST

michaelfeil:
also https://github.com/huggingface/text-embeddings-inference/pull/802 and https://github.com/huggingface/text-embeddings-inference/pull/805 for reference. 

Would be nice if, similar to gemma, `use_bidirectional_attention` would gain more traction, IMO, e.g. for TEI, this is hugley prefered. e.g. gemmaembeddings. Adding support for it rather sooner than later would be awesome!

michaelfeil:
I'm mostly looking at it from a integration point into Rust, even if this is in python. I hope this makes @tomaarsen's job easier too + smoother integration into TEI. Embedding models are also super easy to serve with the wrong attention, e.g. the above models will load well, and TEI will just use the causal attention instead of full attention.

vasqu:
I have the feeling you haven't really looked at #43705? 

Passing `is_causal` as kwarg or setting it as config attribute does everything for you:
- It exchanges the mask for a bidirectional one
- Overwrites the attention module's `is_causal` as kwarg takes precedence over the attr (there were a few prs to enable this) aka what you do by exchanging at init time (which is also brittle as we cannot dynamically change behavior).

Correctness has also been checked by checking such a remote code (with this new PR) to produce similar result (similar due to sdpa backends defaulting to either xformers or fa).

Hence, this is pretty much doing everything that you do here without modifying the modeling code. The only difference is that we use `is_causal` instead of `use_bidirectional_attention`. You can keep BC for remote code by just adding `is_causal: false` to the configs. 

The standard behind `is_causal` is to keep things as general as possible and make it extendable to future flags we can use for most models. `use_bidirectional_attention` is super specific to causal models and doesn't take into account 
1. That are more unique models like VLMs (paligemma's mask comes into my mind), encoder-decoder models etc
2. How specific and unscalable this is - every causal model that now subsequently gets a bidirectional model has to be modified after model addition.

The goal is to eventually refactor the gemma embedding model as well to follow the `is_causal` standard (with BC behavior).

michaelfeil:
awesome, so we better add a PR to require latest transformers + `is_causal` to the config.json? I'm happy to contribute it, once we have a transformers version. Sounds great.

tomaarsen:
> awesome, so we better add a PR to require latest transformers + `is_causal` to the config.json? I'm happy to contribute it, once we have a transformers version. Sounds great.

Bingo! In short, the PR @vasqu mentions was introduced exactly because we wanted it to be easier to turn a decoder into an encoder, as it's a very common strategy for modern embedding and retrieval models.

Most models should soon accept `is_causal=False` in either the config or the forward.

- Tom Aarsen

vasqu:
Our minor release cycle should be much faster now so you hopefully won't have to wait long ;) also we should add more docs 😭

michaelfeil:
This is awesome! Makes so much sense, thank you all!

PR #43811, opened 2026-02-06
Fix video interpolation in pe_audio_video

Rocketknight1:
The test `tests/models/pe_audio_video/test_modeling_pe_audio_video.py::PeAudioVideoEncoderTest::test_model_forward_default_config_values` is flaky in the CI. In local testing, it failed in 5 out of 100 runs for me. After some digging, I figured out that in the failing runs, it takes a fast path when all the audio and video samples are the same length. However, this fast path is incorrect, because it interpolates along `seq_dim` instead of the `timestep` dimension, so we get an output with the wrong shape and the test fails.

The fix is just to correct the fast path to interpolate the right dimension! cc @zucchini-nlp @tarekziade 

Example failing run: https://app.circleci.com/pipelines/github/huggingface/transformers/163294/workflows/b07f04c5-f240-4601-aace-47f970ed7c1c/jobs/2151900

### DIFF
--- a/src/transformers/models/pe_audio_video/modeling_pe_audio_video.py
+++ b/src/transformers/models/pe_audio_video/modeling_pe_audio_video.py
@@ -195,7 +195,9 @@ class PeAudioVideoEncoderEmbedder(nn.Module):
             # note: when one of the above is true, we can expect the other to be true as there is no reason
             # to have masked audio without masked video and vice versa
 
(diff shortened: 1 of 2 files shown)


Rocketknight1:
run-slow: pe_audio_video

tarekziade:
nice finding!

zucchini-nlp (approved):
Great catch, thanks

PR #43812, opened 2026-02-06
Fix: Import torch.nn for type hints even when PyTorch is disabled

### LINKED ISSUE #43784: NameError: name 'nn' is not defined when importing sentence-transformers with latest transformers
## System Info

```
transformers version: latest (installed via pip today, 2026-02-05)
sentence-transformers version: latest
torch version: 2.x (from pytorch/pytorch Docker image)
Python version: 3.11+
OS: Linux (Docker container)
```

## Who can help?

@ArthurZucker @Rocketknight1

## Information

- [x] My own modified scripts

## Tasks

- [ ] An officially supported task in the `examples` folder
- [x] My own task or dataset

## Reproduction

Minimal reproduction:

```python
from sentence_transformers import SentenceTransformer

model = SentenceTransformer("all-MiniLM-L6-v2")
```

Or directly:

```python
```

Results in:

```
NameError: name 'nn' is not defined
```

The error occurs during the import chain when `transformers` attempts to load a module that references `nn.Module` or similar `torch.nn` constructs without first importing `torch.nn as nn`.

## Expected behavior

The import should succeed without errors. This worked with previous versions of `transformers`.

## Additional context

This appears to be a similar regression to #38269 ("NameError: name 'Replicate' is not defined"), which was caused by a missing import in `transformers/integrations/tensor_parallel.py`. 

The current error (`nn` not defined) suggests another module has a similar missing import issue, likely referencing `nn.Module` without the corresponding `import torch.nn as nn` statement.

**Workaround**: Pinning `transformers<4.53.0` (or to the last known working version) resolves the issue.

tobyliu2004:
# What does this PR do?

Fixes #43784

Fixes `NameError: name 'nn' is not defined` when importing transformers with PyTorch < 2.4.

## The Issue

When PyTorch < 2.4 is detected, transformers disables PyTorch by making `is_torch_available()` return `False`. This causes the import of `torch.nn as nn` to be skipped (line 42).

However, `nn.Module` is used in type hints throughout the file (lines 155, 491, 504, 561). Type hints are evaluated at module import time, before any runtime checks, causing:
```python
NameError: name 'nn' is not defined
```

## The Fix

Import `torch.nn as nn` in the `TYPE_CHECKING` block to ensure it's always available for type hints, regardless of PyTorch version detection.
```python
if TYPE_CHECKING:
    import torch.nn as nn

if is_torch_available():
    import torch
    if not TYPE_CHECKING:  # avoid duplicate import
        import torch.nn as nn
```

This follows the standard pattern for making imports available to type checkers without affecting runtime behavior.

## Testing

**Reproduced the bug condition:**
- Monkey-patched `is_torch_available()` to return `False`
- Verified import succeeds with the fix

**Verified no regression:**
- Import works normally with PyTorch 2.10
- All type hints remain valid

Tested on macOS with Python 3.12 and PyTorch 2.10.

## Before submitting
- [x] This PR fixes a bug
- [x] Tested both bug condition and normal usage
- [x] Code follows existing patterns

cc @ArthurZucker @Rocketknight1 @tomaarsen

### DIFF
--- a/src/transformers/integrations/accelerate.py
+++ b/src/transformers/integrations/accelerate.py
@@ -37,9 +37,13 @@ from .deepspeed import is_deepspeed_zero3_enabled
 from .fsdp import is_fsdp_enabled
 
 
+if TYPE_CHECKING:
+    import torch.nn as nn
+
 if is_torch_available():
     import torch
-    import torch.nn as nn
+    if not TYPE_CHECKING:  # avoid duplicate import
+        import torch.nn as nn
 
 if is_accelerate_available():
     from accelerate import dispatch_model


tomaarsen:
Hello!

I'm not sure we'll want to go in this direction. One of my colleagues recently deprecated torch <2.4, and it does not make much sense to re-introduce compatibility for older versions. 

By requiring torch 2.4 or higher, we can start using newer functionality from it.

I'd recommend either upgrading your `torch`, or sticking with a lower `transformers` version until you're ready to upgrade `torch`.

I don't do a lot of maintenance on `transformers`, though, so I can leave the final decision to one of my colleagues.

- Tom Aarsen

notrudyyy:
@tomaarsen Is there any update on this? If not, I suppose it can be closed?

tomaarsen:
Yes, it should be closed. Thanks for the ping.

- Tom Aarsen

PR #43814, opened 2026-02-07
Delete batch_split from EncoderDecoderCache

cyyever:
# What does this PR do?

This EncoderDecoderCache.batch_split is a remaining method from previous refactoring and is not used by other code.


vasqu (approved):
Yup, thanks for taking care of this 🫡

PR #43817, opened 2026-02-07
fix(flaky): use a fixture for `set_seed` and single-threading [WIP]

tarekziade:
This is a follow-up work on trying to fix flakiness. Adding Global Deterministic Testing

# Deterministic Testing Infrastructure - Summary

N.B. this is for CPU-only tests

## Problem

The test suite has flaky tests that failed intermittently due to non-deterministic behavior:
- Random number generation (RNG) producing different values across runs
- PyTorch algorithm selection varying based on availability
- Non-deterministic parallel computation order
- Threading causing race conditions

**Original flaky tests:**
1. `test_generate_continue_from_past_key_values` (qwen3_omni_moe, qwen2_5_omni)
2. `test_assisted_decoding_matches_greedy_search_1_same` (kosmos2)

### 1. Global Deterministic Fixture

Added a global pytest fixture in `tests/conftest.py` that runs before every test:

```python
@pytest.fixture(autouse=True)
def setup_deterministic_testing():
    if is_torch_available():
        import torch
        torch.set_num_threads(1)
        set_seed(42, deterministic=True)
    else:
        set_seed(42, deterministic=False)
```

**Impact:**
- All tests now run with deterministic algorithms enabled
- Single-threaded PyTorch execution eliminates threading non-determinism
- Consistent random seed across all tests

### 2. Fixed Flaky Generation Tests

Added strategic `set_seed(42, deterministic=True)` calls before `model.generate()` in tests comparing multiple generation outputs:

**Files:**
- `tests/generation/test_utils.py`:
  - `test_assisted_decoding_matches_greedy_search`
  - `test_generate_continue_from_past_key_values`

**Pattern:**
```python
set_seed(42, deterministic=True)
output1 = model.generate(...)

set_seed(42, deterministic=True)  # Reset to ensure same RNG state
output2 = model.generate(...)
```

### 3. Fixed test_feed_forward_chunking Weight Mismatch

**Problem:** 32 models failed because models were created at different times with different random weights.

**Solution:** Create both models first, then copy weights:

```python
model_no_chunk = model_class(config_no_chunk)
model_with_chunk = model_class(config_with_chunk)
model_with_chunk.load_state_dict(model_no_chunk.state_dict())  # Ensure identical weights
```

**Files:**
- `tests/test_modeling_common.py::test_feed_forward_chunking`

### 4. Cleanup: Removed Redundant set_seed() Calls

Removed 24+ redundant `set_seed(42)` calls from test files since the global fixture now handles initialization.

**Files cleaned:**
- `tests/generation/test_utils.py` (5 calls)
- `tests/trainer/test_trainer.py` (2 calls)
- `tests/causal_lm_tester.py` (1 call)
- Various model test files

### 5. Identified Pre-existing Issues

Deterministic mode exposed pre-existing bugs in some models:

**Chunking implementation bugs (4 models):**
- `vilt`: Chunked output differs from non-chunked with identical weights
- `SpeechT5ForTextToSpeech`: Borderline numerical precision (14% over tolerance)
- `vits`: Shape mismatch (2208 vs 2240 elements)
- `vit_mae`: Catastrophic differences (99.7% mismatched, 3353x over tolerance)

**Deterministic mode incompatibilities (2 models):**
- `EsmFold`: Creates empty tensors
- `Tapas`: Produces NaN values in logits

**Network flakiness (1 test):**
- `video_llama_3`: Downloads 10MB video - marked with `@is_flaky`

All incompatible tests are now properly documented and skipped.

### Before (Flaky)
```python
def test_generation(self):
    output1 = model.generate(**inputs)
    output2 = model.generate(**inputs)
    assert outputs_match(output1, output2)  # Fails intermittently!
```

### After (Deterministic)
```python
def test_generation(self):
    set_seed(42, deterministic=True)
    output1 = model.generate(**inputs)

    set_seed(42, deterministic=True)  # Reset RNG state
    output2 = model.generate(**inputs)

    assert outputs_match(output1, output2)  # Always passes!
```

### Key Rules
1. **Don't add `set_seed()` at test start** - Global fixture handles this
2. **Do add `set_seed()` before each `generate()` call** - When comparing outputs
3. **Always use `deterministic=True`** - Be explicit
4. **Copy weights when comparing models** - Use `load_state_dict()`
5. **Document skips** - Explain why a test can't be deterministic

## Future Work

- [ ] Fix chunking bugs in vilt, vits, vit_mae, SpeechT5
- [ ] Investigate EsmFold and Tapas initialization issues with deterministic mode
- [ ] Consider adding determinism checks to CI to catch new flaky tests early
- [ ] Explore per-model deterministic mode overrides for models that need it

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -86,11 +86,10 @@ warnings.simplefilter(action="ignore", category=FutureWarning)
 def setup_deterministic_testing():
     """
     Global fixture to ensure deterministic behavior across all tests.
-    Sets PyTorch to single-threaded mode and resets the random seed.
+    Sets PyTorch to single-threaded mode and enables deterministic algorithms.
 
-    Note: We don't use deterministic=True globally as it changes numerical behavior
-    in ways that break many tests (e.g., test_feed_forward_chunking). Instead, we rely
-    on single-threading and seed resetting for reproducibility.
+    This ensures maximum reproducibility and eliminates flaky tests caused by
+    non-deterministic operations (e.g., in generation tests).
     """
     if is_torch_available():
         import torch
@@ -98,8 +97,8 @@ def setup_deterministic_testing():
         # Force single-threaded execution to avoid non-deterministic thread scheduling
         torch.set_num_threads(1)
 
-        # Set seed but don't force deterministic algorithms (would break many tests)
-        set_seed(42, deterministic=False)
+        # Enable deterministic algorithms for maximum reproducibility
+        set_seed(42, deterministic=True)
     else:
         # Still set seed for random and numpy even without torch
         set_seed(42, deterministic=False)
--- a/conftest.py
+++ b/conftest.py
@@ -31,6 +31,7 @@ from transformers.testing_utils import (
     patch_testing_methods_to_collect_info,
     patch_torch_compile_force_graph,
 )
+from transformers.trainer_utils import set_seed
 from transformers.utils import enable_tf32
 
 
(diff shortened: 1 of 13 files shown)


tarekziade:
dropped with the recent work that was done in this area

PR #43822, opened 2026-02-07
Fix NameError: name 'nn' is not defined when PyTorch < 2.4

### LINKED ISSUE #43784: NameError: name 'nn' is not defined when importing sentence-transformers with latest transformers

veeceey:
## What does this PR do?

Fixes #43784

When PyTorch < 2.4 is installed, transformers v5.x disables PyTorch by making `is_torch_available()` return `False`. This causes the conditional import of `torch.nn as nn` (line 42) to be skipped. However, `nn.Module` is used in type annotations throughout `accelerate.py` (function signatures at lines 62, 191, 527, 540, 597), which are evaluated at module import time, causing:

```
NameError: name 'nn' is not defined
```

## The Fix

Add `from __future__ import annotations` (PEP 563) to make all annotations lazily evaluated as string literals rather than being evaluated at import time. This means `nn.Module` in function signatures becomes a string and is never resolved during import.

This follows the same pattern already used in sibling modules:
- `transformers/integrations/tensor_parallel.py` (line 14)
- `transformers/integrations/fsdp.py`

The runtime `isinstance()` calls using `nn.Module` and `nn.Parameter` (lines 745, 801) are unaffected because they are inside functions that are only called when PyTorch is available.

## Reproduction

As reported in the issue, this can be reproduced with PyTorch < 2.4 (e.g., 2.2.1):

```python
from sentence_transformers import SentenceTransformer
# NameError: name 'nn' is not defined
```

## Before submitting

- [x] This PR fixes a bug (NameError on import with old PyTorch)
- [x] Follows existing patterns in the codebase (`tensor_parallel.py`, `fsdp.py`)
- [x] Minimal, single-line change with no side effects

### DIFF
--- a/src/transformers/integrations/accelerate.py
+++ b/src/transformers/integrations/accelerate.py
@@ -26,7 +26,6 @@ from collections import OrderedDict, defaultdict
 from typing import TYPE_CHECKING
 
 from safetensors import safe_open
-from safetensors.torch import save_file
 
 from ..utils import (
     is_accelerate_available,
@@ -42,6 +41,7 @@ from .fsdp import is_fsdp_enabled
 if is_torch_available():
     import torch
     import torch.nn as nn
+    from safetensors.torch import save_file
 
 if is_accelerate_available():
     from accelerate import dispatch_model
--- a/src/transformers/integrations/accelerate.py
+++ b/src/transformers/integrations/accelerate.py
@@ -16,6 +16,8 @@ Some of the functions here are derived from the `accelerate` library, with some 
 and simplicity/ease of use.
 """
 
+from __future__ import annotations
+
 import copy
 import inspect
 import os
--- a/src/transformers/integrations/accelerate.py
+++ b/src/transformers/integrations/accelerate.py
@@ -92,7 +92,7 @@ def get_module_size_with_ties(
     return module_size_with_ties, tied_module_names, tied_modules
 
 
-def check_and_set_device_map(device_map: "torch.device | int | str | dict | None") -> dict | str | None:
+def check_and_set_device_map(device_map: torch.device | int | str | dict | None) -> dict | str | None:
     from ..modeling_utils import get_torch_context_manager_or_global_device
 
     # Potentially detect context manager or global device, and use it (only if no device_map was provided)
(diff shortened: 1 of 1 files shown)


veeceey:
Fixed the failing  CI check.

The issue was that a previous commit removed the null check for `TokenizersBackend` before calling `.from_pretrained()`. When the tokenizers library is not available, `TokenizersBackend` is set to `None` (see line 50 of `tokenization_auto.py`), which would cause a `TypeError` when trying to call `None.from_pretrained()`.

The fix restores the null check and properly falls back to `tokenizer_class_from_name()` when:
1. `TokenizersBackend` is `None` (tokenizers library not installed)
2. `TokenizersBackend.from_pretrained()` raises an exception

This ensures the code works correctly with or without the tokenizers library installed.

vasqu:
We have bumped the min torch version to 2.4 in #41307 so while the fix is (likely) valid, I don't think we need this cc @Rocketknight1

Rocketknight1:
Yes, even if you "fix" this, there is lots of other code that will be broken for PyTorch < 2.4, so you'll just hit a bug somewhere else. The right solution here is to update your PyTorch version!

PR #43830, opened 2026-02-08
Adding Support for Qwen3.5

bozheng-hit:
This PR adds the support of codes for the upcoming Qwen3.5 series models. For information about Qwen, please visit:
👉https://qwen.ai

Special thanks to @JJJYmmm for helping complete the code in this PR. We also appreciate the valuable feedback and thorough review provided by @vasqu and @ArthurZucker ! 🙏


fpshuang:
Merge this PR only after Feb 23th, *PLEASE! WE NEED A FINE SPRING FESTIVAL*

1-bytes:
尽快审批～ 期待

vasqu (approved):

MaoJianwei:
wait for a long time

BestJuly:
Hi @bozheng-hit , thanks for the PR. I have a simple question, that we noticed that the dtypes for `A_log` and `out_norm` in the HF released ckpt are different from Qwen3-Next. Should the storage of these two parameters also be in FP32? Or computation is necessary in FP32 but storage could remain the same as Qwen3-Next, in BF16.
image
image

Current implementation in MCore is using BF16 for storage and computation for these parts are in FP32. But if the storage dtypes are also required in FP32, there require make some changes ([draft PR](https://github.com/NVIDIA/Megatron-LM/pull/3634)). Thank you.

adithyabsk:
I have a similar question as @BestJuly and a follow up. @bozheng-hit wanted to better understand the choice of dtype here for `A_log` and `out_norm` and it also seems we force qkv, and a to float as well. 

https://github.com/huggingface/transformers/blob/fa6c8308e22dade298c10c72d44937e41b962353/src/transformers/models/qwen3_5/modeling_qwen3_5.py#L262-L264

https://github.com/huggingface/transformers/blob/fa6c8308e22dade298c10c72d44937e41b962353/src/transformers/models/qwen3_5/modeling_qwen3_5.py#L518

Empirically, I see the choice of numerics for these ops matters a lot. It seems that there is a perf / capability trade off here. Has there been any investigations into different dtypes for these key ops and the subsequent impact for SFT / RL?

PR #43847, opened 2026-02-08
 Fix _from_config silently skipping weight initialization under DeepSpeed ZeRO-3

tohtana:
# What does this PR do?

When using `HfDeepSpeedConfig` + ZeRO-3 with `from_config()`, the model's custom weight initialization (`_init_weights`) is silently skipped because parameters are already partitioned by `deepspeed.zero.Init`. The model retains PyTorch's default `kaiming_uniform_` initialization instead of the intended initialization (e.g., `normal_(std=0.02)`). The abnormal gradients reported in the issue are a downstream symptom of this wrong init.

**Root cause:** `_from_config()` enters `deepspeed.zero.Init` without `no_init_weights()`. Under this context, parameters are immediately partitioned after each module's `__init__` (`param.data = empty(0)`). When `init_weights()` → `smart_apply()` → `_init_weights()` later runs, it operates on zero-sized tensors — a silent no-op.

**Fix:** Add `no_init_weights()` to `_from_config`'s context managers when ZeRO-3 is active (same pattern `get_init_context()` uses for `from_pretrained`). After construction, re-initialize weights using a `smart_apply`-style traversal with `GatheredParameters(modifier_rank=0)` — gathering each module's params, running `_init_weights` on rank 0, then broadcasting. Finally call `tie_weights()` to restore weight tying that was skipped during the suppressed init.

This mirrors the established pattern for `from_pretrained` (which uses `no_init_weights()` + loads pretrained weights via `GatheredParameters`), adapted for random initialization.

Fixes #43844

## Before submitting
      Pull Request section?
      to it if that's the case. — #43844

## Verified

 Tested with `SmolLM2-135M`, 1 GPU, ZeRO-3, single training step.

| Metric | Before fix (buggy) | After fix | Correct path (`from_config` w/o DS) |
|--------|-------------------|-----------|-------------------------------------|
| Weight std (q_proj) | 0.024 | 0.0417 | 0.0414 |
| Loss | 381.83 | 11.33 | 11.17 |
| Grad std (v_proj) | 531.65 | 53.48 | 51.44 |

- `from_pretrained` + ZeRO-3 still works correctly (no regression)
- `_is_hf_initialized` is `True` on all modules after fix
- Weight tying preserved (`lm_head.weight is embed_tokens.weight`)

## Who can review?

@3outeille @SunMarc — DeepSpeed / distributed / Big Model Inference integration


Cyrilvallez:
Hey @tohtana! Thanks for the PR. I'm not an expert on deepspeed, but I assume the issue here is that the modules keep their `_is_hf_initialized` flag to True even after deepspeed changed the params to empty? Otherwise, `no_init_weights` is actually not needed. And I don't get how this could work if every rank gets params replaced by empty tensor? When is the base torch init scheme called then? A few pointers would be helpful

@SunMarc could you double check here? Mostly the fact that deepspeed will indeed set params to empty before re-gathering then and somehow call torch inits?

src/transformers/modeling_utils.py
```diff
@@ -1461,18 +1461,56 @@ def _from_config(cls, config, **kwargs):
...
+                    else:
+                        _reinit_zero3_weights(child, init_fn)
+
+                params = list(model_or_module.parameters(recurse=False))
+                if params:
+                    with GatheredParameters(params, modifier_rank=0):
+                        if deepspeed.comm.get_rank() == 0:
+                            init_fn(model_or_module)
+                else:
+                    init_fn(model_or_module)
```
Cyrilvallez, line 1509: Not needed here if we don't have any params anyway

tohtana:
Hi @Cyrilvallez, I'm not an HF expert either. I appreciate your feedback on this PR's approach.

> the issue here is that the modules keep their _is_hf_initialized flag to True even after deepspeed changed the params to empty? Otherwise, no_init_weights is actually not needed.

Are you suggesting we manually reset `_is_hf_initialized` to False after exiting the ContextManagers?
If we go that route, we'd need to handle the flag reset consistently across all submodules. We'd also be running the initialization twice, which could affect RNG states. I thought  `no_init_weights` is the standard way to prevent these issues.

> And I don't get how this could work if every rank gets params replaced by empty tensor? When is the base torch init scheme called then?

In `GatheredParameters` context, we unshard tensors. The weight initialization occurs while the parameters are gathered, and those updated weights are then broadcasted across the ranks.

src/transformers/modeling_utils.py
tohtana: Thank you for the suggestion. I wondered if `init_fn` covers more than parameters (buffers and other states).

SunMarc:
> Root cause: _from_config() enters deepspeed.zero.Init without no_init_weights(). Under this context, parameters are immediately partitioned after each module's __init__ (param.data = empty(0)). When init_weights() → smart_apply() → _init_weights() later runs, it operates on zero-sized tensors — a silent no-op.

This happens because we are not gathering the parameters no ? In `from_pretrained`, we have the following to make sure we correctly initialize the missing weights:

```python
    def _initialize_missing_keys(self, is_quantized: bool) -> None:
        """
        Initialize the missing keys (keys that are part of the model parameters, but were NOT found in the loaded state dicts), according to
        `_initialize_weights`. Indeed, since the corresponding weights are missing from the state dict, they will not be replaced and need to
        be initialized correctly (i.e. weight initialization distribution).

        Params that are not missing have the `is_hf_initialized` flag.
        """
        # This will only initialize submodules that are not marked as initialized by the line above.
        if is_deepspeed_zero3_enabled() and not is_quantized:
            import deepspeed

            # keep_vars=True as we need the original tensors, so that the "_is_hf_initialized" is present on them
            not_initialized_parameters = list(
                {v for v in self.state_dict(keep_vars=True).values() if not getattr(v, "_is_hf_initialized", False)}
            )
            with deepspeed.zero.GatheredParameters(not_initialized_parameters, modifier_rank=0):
                self.initialize_weights()
        else:
            self.initialize_weights()
```

When using `from_config`, it only triggers the `post_init` where we have: 

```python
def init_weights(self):
        """
        Initialize and tie the weights if needed. If using a custom `PreTrainedModel`, you need to implement any
        initialization logic in `_init_weights`.
        """
        # If we are initializing on meta device, there is no point in trying to run inits
        if get_torch_context_manager_or_global_device() != torch.device("meta"):
            # Initialize weights
            self.initialize_weights()
        # Tie weights needs to be called here, but it can use the pre-computed `all_tied_weights_keys`
        self.tie_weights(recompute_mapping=False)
```

SunMarc:
Can you share your script so that I can have a look ? @tohtana

tohtana:
Thank you for the feedback, @SunMarc!

Yes, I think the core issue is that `from_config` + ZeRO-3 currently misses a gathered initialization step.
As you pointed out, `from_pretrained` has `_initialize_missing_keys()` under `GatheredParameters`, while `from_config` goes through `post_init -> init_weights` without an equivalent gather step.

I agree we should avoid introducing a separate custom recursive init path if we can reuse the existing `init_weights` flow.
One caveat versus `from_pretrained`: for `from_config`, we need to initialize essentially all params, so gathering the full model at once can create a large memory spike. We likely still need module-wise gathered init to keep peak memory bounded.

Here is a minimal repro script. On my side, the `use_hfds=True` case shows clearly different init stats from the non-
DeepSpeed `from_config` baseline (It produced `nan` in my environment).

```python
# repro_from_config_zero3_init.py
import os
import torch
import torch.distributed as dist

from transformers import AutoConfig, AutoModelForCausalLM
from transformers.integrations.deepspeed import HfDeepSpeedConfig

def init_dist():
    if dist.is_initialized():
        return
    os.environ.setdefault("MASTER_ADDR", "127.0.0.1")
    os.environ.setdefault("MASTER_PORT", "29617")  # non-default
    os.environ.setdefault("RANK", "0")
    os.environ.setdefault("WORLD_SIZE", "1")
    os.environ.setdefault("LOCAL_RANK", "0")
    dist.init_process_group(backend="nccl")
    torch.cuda.set_device(0)

def qproj_std(model):
    return float(model.model.layers[0].self_attn.q_proj.weight.detach().float().std().cpu())

def run_case(use_hfds):
    torch.manual_seed(1234)
    cfg = AutoConfig.from_pretrained("HuggingFaceTB/SmolLM2-135M")

    if use_hfds:
        ds_cfg = {
            "train_batch_size": 1,
            "zero_optimization": {"stage": 3},
            "bf16": {"enabled": True},
        }
        _ = HfDeepSpeedConfig(ds_cfg)  # activates ZeRO-3 path in from_config

    model = AutoModelForCausalLM.from_config(cfg).cuda()
    std = qproj_std(model)
    print(f"use_hfds={use_hfds} q_proj.std={std:.6f}")
    return model

if __name__ == "__main__":
    init_dist()
    run_case(False)  # expected HF init distribution
    run_case(True)   # currently drifts (issue case)
```

SunMarc:
Thanks for the reproducer ! I've opened a PR to fast track this PR ! Added you as co-author

tohtana:
Closing as #44236 has been merged.

PR #43848, opened 2026-02-09
Fix SwanLab callback to forward resume init args

### LINKED ISSUE #43698: SwanLab integration uses outdated swanlab.init() signature

weiguangli-io:
## Summary

Fixes #43698 by adding SwanLab resume init argument support in the Trainer integration.

## What changed

- `SwanLabCallback.setup()` now forwards two optional env-based init args to `swanlab.init(...)`:
  - `SWANLAB_RUN_ID` -> `id`
  - `SWANLAB_RESUME` -> `resume`
- Added callback docs for both env vars in the SwanLab integration section.
- Added unit tests to cover:
  - default behavior (no `id` / `resume` forwarded)
  - env-driven behavior (`id` / `resume` forwarded)

## Why env vars?

This keeps the integration change minimal and backward-compatible while providing a direct way to pass SwanLab's resume parameters in resumed training workflows.

## Tests

- `PYTHONPATH=src uv run --with pytest pytest -q tests/trainer/test_trainer_callback.py -k SwanLabCallbackTest`
- `uvx ruff check tests/trainer/test_trainer_callback.py`

### DIFF
--- a/tests/trainer/test_trainer_callback.py
+++ b/tests/trainer/test_trainer_callback.py
@@ -27,7 +27,8 @@ import os
 import shutil
 import tempfile
 import unittest
-from unittest.mock import patch
+from types import SimpleNamespace
+from unittest.mock import Mock, patch
 
 from transformers import (
     DefaultFlowCallback,
@@ -41,6 +42,7 @@ from transformers import (
     TrainingArguments,
     is_torch_available,
 )
+from transformers.integrations.integration_utils import SwanLabCallback
 from transformers.testing_utils import require_torch
 from transformers.trainer_callback import CallbackHandler, ExportableState, TrainerControl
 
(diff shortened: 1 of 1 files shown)


MekkCyber (commented):
Thank you @OiPunk  ! sorry this pr is superseded by https://github.com/huggingface/transformers/pull/43719, but we can keep the test I guess, wdyt @SunMarc

SunMarc (approved):
We can keep the tests yes !

weiguangli-io:
Thanks both for the guidance.

I have now trimmed this PR to **tests only** so it does not modify `integration_utils.py` anymore (functional change already merged in #43719).

What remains in this PR:
- `tests/trainer/test_trainer_callback.py`
- coverage for `SWANLAB_RUN_ID` / `SWANLAB_RESUME` forwarding behavior

I also aligned the test setup with current `main` behavior (`resume_from_checkpoint`) to avoid unintended side effects.

Local verification:
`PYTHONPATH=src uv run --with pytest pytest -q tests/trainer/test_trainer_callback.py -k SwanLabCallbackTest`

If you want, I can further narrow the test scope.

PR #43852, opened 2026-02-09
don't merge check

ydshieh:
# What does this PR do?

Currently, if slack reporting (re-usable) workflow or `check new failure` workflow fails, the CI triggered via a comment in a pull request will send a comment back to the PR page with

✅ No failing test specific to this PR 🎉 👏 !

This is very wrong and will cause a lot of issues.

This PR tries to detect such scenario and if this happens, the feedback comment would be something like

 ⚠️ Model CI Infrastructure Error

The test failure analysis could not be completed. Please check the [workflow run](https://github.com/huggingface/transformers/actions/runs/21855542500) for details.

### DIFF
--- a/.github/workflows/check_failed_tests.yml
+++ b/.github/workflows/check_failed_tests.yml
@@ -24,6 +24,10 @@ on:
       pr_number:
         required: false
         type: string
+    outputs:
+      is_check_failures_ok:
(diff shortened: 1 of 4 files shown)


ydshieh:
run-slow: vit

PR #43855, opened 2026-02-09
delete unnecessary code to make moe compatible to full graph compile

kaixuanliu:
When all routing weights are zero, in this line [out_per_sample = out_per_sample * sample_weights.unsqueeze(-1)](https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/moe.py#L153) , it will also output all zero tensor for this func, we can delete theses lines of code to make it work for full graph compile. Here is a test case:
`tests/models/glm4_moe_lite/test_modeling_glm4_moe_lite.py::Glm4MoeModelTest::test_generate_compile_model_forward_fullgraph`


kaixuanliu:
@ArthurZucker pls help review, thx!

vasqu (approved):
Thanks, it was meant for EP but we need to undo this either way cc @3outeille

ArthurZucker (commented):
😉 it's not useless for EP, but EP is not working anyways so yep. 
Sorry about that and ty for the prompt fix!

PR #43862, opened 2026-02-09
Separate `check_model_inputs` into `capture_outputs` and `merge_with_config_defaults` + ensure correctness

Cyrilvallez:
# What does this PR do?

As per the title.
`check_model_inputs` is becoming very complex and doing more than what it should. Let's separate into 2 clear decorators:
- `capture_outputs`: everything related to capture outputs
- `merge_with_config_defaults`: everything related to handling of some `kwargs`

Also, add a test to ensure that `capture_outputs` is not called several time in a chain wrongly for the same output, and fix affected models.
This is very important for correctness and needs a proper test to ensure it.

### DIFF
--- a/src/transformers/utils/generic.py
+++ b/src/transformers/utils/generic.py
@@ -33,7 +33,6 @@ import numpy as np
 
 from ..utils import logging
 from .import_utils import is_mlx_available, is_torch_available, is_torch_fx_proxy
-from .output_capturing import _CAN_RECORD_REGISTRY, _active_collector, maybe_install_capturing_hooks
 
 
 _is_torch_available = False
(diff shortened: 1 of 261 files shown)


src/transformers/utils/output_capturing.py
```diff
@@ -200,3 +201,90 @@ def maybe_install_capturing_hooks(model: PreTrainedModel) -> None:
...
+            # For BC as cross-attentions used to be captured with `output_attentions`
+            if "cross_attentions" in capturable_flags:
+                recordable_keys["output_cross_attentions"] = kwargs.get(
+                    "output_attentions", getattr(self.config, "output_attentions", False)
+                )
+            # The sam model variants need this annoying exception as well...
+            if "mask_decoder_attentions" in capturable_flags:
+                recordable_keys["output_mask_decoder_attentions"] = kwargs.get(
+                    "output_attentions", getattr(self.config, "output_attentions", False)
+                )
```
vasqu, line 242: Not super relevant to this PR as it goes a bit beyond the intended scope here but I've seen a few more exceptions I think with `image_hidden_states` etc. Could be worthwhile to checkout at some point

src/transformers/utils/generic.py
```diff
@@ -895,185 +894,36 @@ def wrapper(self, *args, **kwargs):
...
-
-    def wrapped_fn(func):
-        @wraps(func)
-        def wrapper(self, *args, **kwargs):
-            args_with_config_defaults = [
-                "use_cache",
-                "vision_feature_layer",
-                "vision_feature_select_strategy",
-                "vision_aspect_ratio",
-            ]
```
vasqu, line 925: This is now `merge_with_config_defaults`? Do we always apply both decorators now? E.g. llama needs it for `use_cache` no?

vasqu: Nvm, yes yes we do

vasqu (approved):
LGTM overall, does the chain order have any influence?

src/transformers/utils/output_capturing.py
Cyrilvallez: If a Model uses a submodel that captures "hidden_states", and then later wants to rename it by doing `return ModelOutput(image_hidden_states=vision_model_output.hidden_states,...)` it's not a problem.

This issue is when `"mask_decoder_attentions"` is the key provided in `_can_record_outputs` directly. I cannot find any occurence of `"image_hidden_states"` as a direct key in any `_can_record_outputs` so should be fine here - and there was not any exception for this case either before

Cyrilvallez:
No, none at all! (except obviously how each model is handling the outputs its submodels may capture)

PR #43865, opened 2026-02-09
update somes audio models

Deep-unlearning:


README.md
```diff
@@ -266,7 +266,7 @@ Expand each modality below to see a few example models for various use cases.
 <summary>Audio</summary>
 
 - Audio classification with [Whisper](https://huggingface.co/openai/whisper-large-v3-turbo)
```
ebezzam: I would lean towards removing Whisper here as it's not intended for ASR and they even have a warning on their model card: "In particular, we caution against ... purporting to use these models for any kind of subjective classification...The models are intended to transcribe and translate speech, use of the model for classification is not only not evaluated but also not appropriate, particularly to infer human attributes."

README.md
ebezzam: Replacing Bark with Dia and CSM? Not sure how reliable Bark is...
https://huggingface.co/sesame/csm-1b
https://huggingface.co/nari-labs/Dia-1.6B-0626

README.md
```diff
@@ -266,7 +266,7 @@ Expand each modality below to see a few example models for various use cases.
...
-- Automatic speech recognition with [Moonshine](https://huggingface.co/UsefulSensors/moonshine)
+- Automatic speech recognition with [Parakeet](https://huggingface.co/nvidia/parakeet-tdt-0.6b-v3)
```
ebezzam: Rather this link? https://huggingface.co/nvidia/parakeet-ctc-1.1b#transcribing-using-transformers-%F0%9F%A4%97

As the goal is users to Transformers usage

ebezzam: And we can also add Whisper and GLM ASR to point users to multiple options? For example:

- Automatic speech recognition with [Parakeet](https://huggingface.co/nvidia/parakeet-ctc-1.1b#transcribing-using-transformers-%F0%9F%A4%97),  [Whisper](https://huggingface.co/openai/whisper-large-v3-turbo), and [GLM-ASR](https://huggingface.co/zai-org/GLM-ASR-Nano-2512)

README.md
```diff
@@ -266,7 +266,7 @@ Expand each modality below to see a few example models for various use cases.
...
 - Keyword spotting with [Wav2Vec2](https://huggingface.co/superb/wav2vec2-base-superb-ks)
 - Speech to speech generation with [Moshi](https://huggingface.co/kyutai/moshiko-pytorch-bf16)
```
ebezzam, line 271: Is this supported by transformers? I don't see transformers usage at the model card

README.md
```diff
@@ -292,7 +292,7 @@ Expand each modality below to see a few example models for various use cases.
 <details>
 <summary>Multimodal</summary>
 
-- Audio or text to text with [Qwen2-Audio](https://huggingface.co/Qwen/Qwen2-Audio-7B)
+- Audio or text to text with [Voxtral](https://huggingface.co/mistralai/Voxtral-Mini-3B-2507)
```
ebezzam: Can also add:
- Audio Flamingo 3: https://huggingface.co/nvidia/audio-flamingo-3-hf

README.md
ebezzam, line 274: How about adding codec models?
- https://huggingface.co/descript/dac_44khz
- https://huggingface.co/facebook/encodec_24khz
- https://huggingface.co/microsoft/VibeVoice-AcousticTokenizer

ebezzam (commented):
Thanks @Deep-unlearning I've mentioned a few models possibilities

Do we have to only mention one model per task? 
cc @vasqu if you know?

README.md
eustlb: clearly! does not make sense to restrict to only one model here IMO

README.md
eustlb: not sure this make sense to have on the main readme

README.md
eustlb: yep

vasqu:
Nope, let's mention a few popular ones! We shouldn't restrict

README.md
Deep-unlearning: lets just add csm then

README.md
Deep-unlearning: Which model do you think we should replace?

README.md
Deep-unlearning: no its not, only parakeet-ctc, thanks for pointing that out!

README.md
ebezzam: These ones could be relevant to link:
- Most recent which can be used for zero-shot (similar to CLAP): https://huggingface.co/facebook/pe-av-small
- Another popular zero-shot: https://huggingface.co/laion/clap-htsat-fused
- Emotion: https://huggingface.co/superb/wav2vec2-base-superb-er
- Language: https://huggingface.co/facebook/mms-lid-1024

Deep-unlearning: lets use CLAP then

README.md
```diff
@@ -265,12 +265,12 @@ Expand each modality below to see a few example models for various use cases.
 <details>
 <summary>Audio</summary>
 
-- Audio classification with [Whisper](https://huggingface.co/openai/whisper-large-v3-turbo)
-- Automatic speech recognition with [Moonshine](https://huggingface.co/UsefulSensors/moonshine)
+- Audio classification with [CLAP](https://huggingface.co/laion/clap-htsat-fused)
+- Automatic speech recognition with [Parakeet](https://huggingface.co/nvidia/parakeet-ctc-1.1b), [Whisper](https://huggingface.co/openai/whisper-large-v3-turbo), and [GLM-ASR](https://huggingface.co/zai-org/GLM-ASR-Nano-2512)
```
ebezzam: For Parakeet can we link to the Transformers usage? https://huggingface.co/nvidia/parakeet-ctc-1.1b#transcribing-using-transformers-%F0%9F%A4%97

ebezzam: And since we're doing multiple models, can we also add [Moonshine Streaming](https://huggingface.co/UsefulSensors/moonshine-streaming-medium)? As it's another type of ASR that could be interesting for streaming usage

ebezzam (approved):
Thanks for updating the models, just two more small requests! Otherwise looks good 👍

PR #43876, opened 2026-02-10
Fix GlmMoeDsaConfig default mlp_layer_types in modular conversion

### LINKED ISSUE #43864: GlmMoeDsaConfig: mlp_layer_types default overwritten by inlined parent init
## Bug Description

`GlmMoeDsaConfig` ends up with the wrong default `mlp_layer_types`. The intended default is `["dense"]*3 + ["sparse"]*75` (3 initial dense layers), but the actual default at runtime is `["dense"] + ["sparse"]*77` (1 dense layer).

## Root Cause

The generated `configuration_glm_moe_dsa.py` inlines both the child (`GlmMoeDsaConfig`) and parent (`Glm4MoeLiteConfig`) `__init__` bodies sequentially. Both contain this pattern:

```python
self.mlp_layer_types = mlp_layer_types  # from __init__ param, which is None
if self.mlp_layer_types is None:
    self.mlp_layer_types = ["dense"] * N + ["sparse"] * M
```

The child's block (lines ~211-215) correctly computes `["dense"]*3 + ["sparse"]*75`. But then the parent's inlined block (lines ~254-256) runs afterward, resets `self.mlp_layer_types` back to `None` (from the original parameter), and recomputes it as `["dense"] + ["sparse"]*77`.

The modular source (`modular_glm_moe_dsa.py`) has the same issue — the child sets the correct value at lines 214-217, then `super().__init__(**kwargs)` at line 250 calls the parent which overwrites it.

## Reproduction

```python
from transformers import GlmMoeDsaConfig

config = GlmMoeDsaConfig()
print(config.mlp_layer_types[:5])
# Actual:   ['dense', 'sparse', 'sparse', 'sparse', 'sparse']
# Expected: ['dense', 'dense', 'dense', 'sparse', 'sparse']
```

## Impact

Layers 1 and 2 will use sparse MoE instead of dense MLP when using the default config. This likely does not affect pretrained model loading (where `mlp_layer_types` is read from `config.json`), but it affects anyone instantiating a default config.

## Affected Files

- `src/transformers/models/glm_moe_dsa/modular_glm_moe_dsa.py` (source of truth)
- `src/transformers/models/glm_moe_dsa/configuration_glm_moe_dsa.py` (generated, runtime)

## Suggested Fix

In `modular_glm_moe_dsa.py`, move the `mlp_layer_types` default computation to **after** the `super().__init__()` call, so the child's value is set last:

```python
def __init__(self, ..., mlp_layer_types=None, **kwargs):
    ...
    super().__init__(**kwargs)
    # Set after super() so parent doesn't overwrite
    if mlp_layer_types is None:
        mlp_layer_types = ["dense"] * 3 + ["sparse"] * (self.num_hidden_layers - 3)
    self.mlp_layer_types = mlp_layer_types
```

Then regenerate `configuration_glm_moe_dsa.py`.

Introduced in #43858.

weiguangli-io:
## Summary
This PR fixes #43864 by preserving the `GlmMoeDsaConfig` default `mlp_layer_types` from the modular source.

`GlmMoeDsaConfig` should default to dense MLP for the first 3 layers and sparse afterward. During modular conversion, the parent init body was being inlined and overwrote that default with the parent pattern (`["dense"] + ["sparse"] * ...`).

## Changes
- In `modular_glm_moe_dsa.py`, call `PreTrainedConfig.__init__(self, **kwargs)` instead of `super().__init__(**kwargs)` to avoid inlining parent init logic.
- Regenerated `configuration_glm_moe_dsa.py` via modular converter, which removes the duplicated parent default block.
- Added a regression test in `tests/models/glm_moe_dsa/test_configuration_glm_moe_dsa.py` to assert the expected default pattern for `num_hidden_layers=8`.

## Validation
- `PYTHONPATH=src python3 utils/modular_model_converter.py glm_moe_dsa`
- `PYTHONPATH=src python3 utils/check_modular_conversion.py --files src/transformers/models/glm_moe_dsa/modular_glm_moe_dsa.py`
- `PYTHONPATH=src python3 -m pytest tests/models/glm_moe_dsa/test_configuration_glm_moe_dsa.py -q`
- `PYTHONPATH=src python3 -m trace --count --summary --module unittest tests.models.glm_moe_dsa.test_configuration_glm_moe_dsa | grep -E "configuration_glm_moe_dsa|test_configuration_glm_moe_dsa"`
  - output: `configuration_glm_moe_dsa ... 100%`


src/transformers/models/glm_moe_dsa/configuration_glm_moe_dsa.py
```diff
@@ -215,47 +215,6 @@ def __init__(
             )
         layer_type_validation(self.mlp_layer_types, self.num_hidden_layers, attention=False)
 
-        self.moe_intermediate_size = moe_intermediate_size
-        self.num_attention_heads = num_attention_heads
-        self.n_shared_experts = n_shared_experts
-        self.n_routed_experts = n_routed_experts
-        self.routed_scaling_factor = routed_scaling_factor
-        self.kv_lora_rank = kv_lora_rank
-        self.q_lora_rank = q_lora_rank
```
zucchini-nlp, line 224: nice, so many duplicates. I also see that we still have `self.n_shared_experts = n_shared_experts` or others duplicated twice in modular file. Can you check and make sure each attr is assigned only once

And at last check that we didn't accidentally delete any config attr in aut-generated `configuration_xxx.py` file

tests/models/glm_moe_dsa/test_configuration_glm_moe_dsa.py
```diff
@@ -0,0 +1,30 @@
...
+
+from transformers import GlmMoeDsaConfig
+
+
+class GlmMoeDsaConfigTest(unittest.TestCase):
+    def test_default_mlp_layer_types(self):
+        config = GlmMoeDsaConfig(num_hidden_layers=8)
+        self.assertEqual(
+            config.mlp_layer_types, ["dense", "dense", "dense", "sparse", "sparse", "sparse", "sparse", "sparse"]
+        )
```
zucchini-nlp: tests have to be under `test_modeling_glm_moe_daa.py::GlmMoeDsaModelTest`

weiguangli-io:
Thanks for the detailed review. I pushed commit `a10f430` to address the requested changes.

What I changed:
- Removed duplicate attribute assignments in `modular_glm_moe_dsa.py` so each config field is assigned once.
- Regenerated `configuration_glm_moe_dsa.py` from the modular source to keep generated output in sync.
- Moved the default `mlp_layer_types` regression test into `test_modeling_glm_moe_dsa.py::GlmMoeDsaModelTest` and removed the standalone configuration test file.

Validation run locally:
- `PYTHONPATH=src python3 utils/modular_model_converter.py glm_moe_dsa`
- `PYTHONPATH=src python3 utils/check_modular_conversion.py --files src/transformers/models/glm_moe_dsa/modular_glm_moe_dsa.py`
- `PYTHONPATH=src python3 -m ruff check src/transformers/models/glm_moe_dsa/modular_glm_moe_dsa.py src/transformers/models/glm_moe_dsa/configuration_glm_moe_dsa.py tests/models/glm_moe_dsa/test_modeling_glm_moe_dsa.py`
- `PYTHONPATH=src pytest -q tests/models/glm_moe_dsa/test_modeling_glm_moe_dsa.py -k default_mlp_layer_types`

I also verified both `mlp_layer_types` paths (`None` and explicit list) execute in the config initializer.

zucchini-nlp:
run-slow: glm_moe_dsa

zucchini-nlp:
@bot /style

zucchini-nlp (approved):
Thanks

PR #43887, opened 2026-02-10
[MistralCommonBackend] fix loading proc

eustlb:
# What does this PR do?

MistralCommonBackend does not implement `added_tokens_decoder` so any call to PreTrainedTokenizerBase's `__repr__ `fails.


ArthurZucker (approved):
ty

PR #43897, opened 2026-02-10
Decorate cache updates with no_grad, just in case

Rocketknight1:
Although our cache update methods are usually used in inference, when grad is disabled anyway, there seem to be some edge cases where they cause problems with compilation and gradient computation. Since we never want to propagate gradient through these steps, adding `@no_grad` decorators shouldn't hurt, and fixes the edge cases.

Fixes #43010

### DIFF
--- a/src/transformers/cache_utils.py
+++ b/src/transformers/cache_utils.py
@@ -96,6 +96,7 @@ class DynamicLayer(CacheLayerMixin):
         self.values = torch.tensor([], dtype=self.dtype, device=self.device)
         self.is_initialized = True
 
+    @torch.no_grad()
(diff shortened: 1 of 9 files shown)


Rocketknight1:
Not sure who the right person to ping for generate/caches is now so cc @remi-or @ArthurZucker @McPatate

ArthurZucker (approved):
ty

Cyrilvallez:
Once again the test fetcher bites us... This fails a few `test_retain_grad_hidden_states_attentions` and `test_training_ci`
cc @ArthurZucker @ydshieh @tarekziade

PR #43904, opened 2026-02-11
Remove unused dependencies

cyyever:
# What does this PR do?

removes two unused dependencies.


Rocketknight1:
Are you sure about the `importlib_metadata` dependency? I thought that was still in use - when did we stop?

cyyever:
> Are you sure about the `importlib_metadata` dependency? I thought that was still in use - when did we stop?

Sure, not used. At least they are not direct dependencies.

Rocketknight1:
Seems okay in that case, cc core maintainers @arthurzucker @cyrilvallez @vasqu since we're touching core dependencies!

setup.py
```diff
@@ -149,8 +147,8 @@
     "torchaudio",
     "torchvision",
     "pyctcdecode>=0.4.0",
-    "tqdm>=4.27",
-    "typer-slim",
+    "tqdm>=4.60",
+    "typer-slim>=0.9.0",
```
Cyrilvallez: Why are we changing versions here?

Cyrilvallez (commented):
It's alright for me to remove `importlib_metadata` as it's indeed superseded by built-in for all Python versions we support, and `diffusers` should not really be a dep. But why did you change the versions of other packages?

setup.py
cyyever: The code base uses features starting from minimum versions of these dependencies.

Cyrilvallez (approved):
LGTM thanks!

PR #43913, opened 2026-02-11
Add a sentinel mechanism in Transpose and fix qwen3_vl_moe weight mapping

Cyrilvallez:
# What does this PR do?

As per the title. Fixes https://github.com/huggingface/transformers/issues/43299#issuecomment-3849688073


ArthurZucker (commented):
Not sure that fairs well with TP no? config shape is never exact.

zucchini-nlp:
IIUC current PR tried to transpose it back, i..e revert the transpose which makes it same as other qwens.

So imo we can just map qwen3-vl-moe to qwen2-moe. I see that originally transpose was added because the model had a slightly different impl of experts, but now it support `expert_implementation` and is standard. So deleting transpose op from mapping would work?

Cyrilvallez:
No no, their saved weight format are different. See https://github.com/huggingface/transformers/pull/43953#issuecomment-3891771939, qwen3_vl_moe already merged the experts, but reverted the dims

ArthurZucker (commented):
let's try it out. Not a fan of the "sentinel" and also the TP plan for gate_up needs to be rowise_packed and for down colwise (if you transpose) if you don't it should not. 
That's a bit tricky with transposition. 

For the name can we use something more explicit and if possible not rely on the config hidden size?

ArthurZucker (commented):
hahah seing it again, do you mind trying another PR with passing the empty tensor? I am really not a fan afterall 😓

PR #43917, opened 2026-02-11
Model patching API

### LINKED ISSUE #43284: Customize Quantization-Friendly Backward Compatibility
### Feature request

Hi guys,
It’s great to see Transformers moving to V5 with a modular design and improved performance! We’ve started adapting it with the RC branch, but noticed that some of the changes are not very friendly for quantization tools.
Here’s the context:
Given a BF16 model, quantization tools typically detect `torch.nn.Linear` layers and then quantize them to the target dtypes. However, for some MoE models like DeepSeek, the experts are concatenated into a large tensor. 

https://github.com/huggingface/transformers/blob/e8ae373133be1eff2254c5dd71fcd628445cb4a4/src/transformers/models/mixtral/modular_mixtral.py#L144-L146

This introduces two potential issues:

- Detection Issue
The quantization tool can’t detect the individual linear layers and requires maintaining a lot of patching code to enable proper quantization.

- Memory Issue
Since the expert concatenation logic happens at the beginning of model loading, it materializes all weights in RAM. This means the peak memory usage exceeds the model size. Previously (before V5), we could quantize layer by layer and offload the quantized layers to disk, so peak memory was only slightly higher than the size of a single decoding layer.

I’m wondering if you have any solution to make these changes more backward-compatible or if there’s a way to provide an interface to disable this behavior?
Thanks!

### Motivation

Better backward compatible

### Your contribution

Instead of concatenating all experts at load time, use lazy loading or chunked concatenation?

IlyasMoutawwakil:
# What does this PR do?

Integrates a library wide monkey patching API to fix #43284 and allow things like restructuring the expert weights and fusing the qkv projections. A concrete and complete example:

<details> 

```py
from typing import Unpack

import torch
import torch.nn as nn

from transformers import AutoModelForCausalLM, Concatenate, WeightConverter
from transformers.activations import ACT2FN
from transformers.cache_utils import Cache
from transformers.conversion_mapping import register_checkpoint_conversion_mapping
from transformers.integrations.sdpa_attention import sdpa_attention_forward
from transformers.models.qwen2_moe.modeling_qwen2_moe import apply_rotary_pos_emb
from transformers.monkey_patch import register_monkey_patch_mapping
from transformers.utils.generic import TransformersKwargs


class MoeMLP(nn.Module):
    def __init__(self, config):
        super().__init__()
        self.config = config
        self.num_experts = config.num_experts
        self.intermediate_size = config.moe_intermediate_size
        self.gate_proj = nn.Linear(self.hidden_size, self.intermediate_size, bias=False)
        self.up_proj = nn.Linear(self.hidden_size, self.intermediate_size, bias=False)
        self.down_proj = nn.Linear(self.intermediate_size, self.hidden_size, bias=False)

    def forward(self, x):
        down_proj = self.down_proj(self.act_fn(self.gate_proj(x)) * self.up_proj(x))
        return down_proj


class ModuleListExperts(nn.ModuleList):
    def __init__(self, config):
        super().__init__()
        self.config = config
        self.num_experts = config.num_experts
        for _ in range(self.num_experts):
            self.append(MoeMLP(config))

    def forward(
        self, hidden_states: torch.Tensor, top_k_index: torch.Tensor, top_k_weights: torch.Tensor
    ) -> torch.Tensor:
        final_hidden_states = torch.zeros_like(hidden_states)
        with torch.no_grad():
            expert_mask = torch.nn.functional.one_hot(top_k_index, num_classes=self.num_experts)
            expert_mask = expert_mask.permute(2, 1, 0)

        for expert_idx in range(self.num_experts):
            top_k_pos, token_idx = torch.where(expert_mask[expert_idx])
            current_hidden_states = self[expert_idx](current_state)
            current_hidden_states = current_hidden_states * top_k_weights[token_idx, top_k_pos, None]
            final_hidden_states.index_add_(0, token_idx, current_hidden_states.to(final_hidden_states.dtype))
        return final_hidden_states


class FusedQKVAttention(nn.Module):
    def __init__(self, config, layer_idx: int):
        super().__init__()
        self.config = config
        self.layer_idx = layer_idx
        self.head_dim = getattr(config, "head_dim", config.hidden_size // config.num_attention_heads)
        self.num_key_value_groups = config.num_attention_heads // config.num_key_value_heads
        self.scaling = self.head_dim**-0.5
        self.attention_dropout = config.attention_dropout
        self.is_causal = True

        self.qkv_proj = nn.Linear(config.hidden_size, 3 * config.num_attention_heads * self.head_dim, bias=True)
        self.o_proj = nn.Linear(config.num_attention_heads * self.head_dim, config.hidden_size, bias=False)

        if self.config.layer_types[layer_idx] == "sliding_attention":
            self.sliding_window = config.sliding_window

    def forward(
        self,
        hidden_states: torch.Tensor,
        position_embeddings: tuple[torch.Tensor, torch.Tensor] | None = None,
        past_key_values: Cache | None = None,
        cache_position: torch.LongTensor | None = None,
    ) -> tuple[torch.Tensor, torch.Tensor]:
        input_shape = hidden_states.shape[:-1]
        hidden_shape = (*input_shape, -1, self.head_dim)

        query_states, key_states, value_states = self.qkv_proj(hidden_states).chunk(3, dim=-1)

        query_states = query_states.view(hidden_shape).transpose(1, 2)
        key_states = key_states.view(hidden_shape).transpose(1, 2)
        value_states = value_states.view(hidden_shape).transpose(1, 2)

        cos, sin = position_embeddings
        query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)

            # sin and cos are specific to RoPE models; cache_position needed for the static cache

        attn_output, attn_weights = sdpa_attention_forward(
            self,
            query_states,
            key_states,
            value_states,
            attention_mask,
            scaling=self.scaling,
            **kwargs,
        )

        attn_output = attn_output.reshape(*input_shape, -1).contiguous()
        attn_output = self.o_proj(attn_output)


# Registering monkey patches for the new attention and experts modules.
register_monkey_patch_mapping(
    mapping={
        "Qwen2MoeExperts": ModuleListExperts,
        "Qwen2MoeAttention": FusedQKVAttention,
    }
)

# Registering weight conversion mappings adapted for the new modules. This registration will:
# - Override the original conversion mapping for qwen2_moe which packed the experts into a single parameter format.
# - Convert the q_proj, k_proj, v_proj weights/biases into a single qkv_proj weight/bias for the new fused attention module.
register_checkpoint_conversion_mapping(
    model_type="qwen2_moe",
    mapping=[
        WeightConverter(
            source_patterns=["q_proj", "k_proj", "v_proj"],
            target_patterns=["qkv_proj"],
            operations=[Concatenate(dim=0)],
        ),
    ],
    overwrite=True,
)

model = AutoModelForCausalLM.from_pretrained("Qwen/Qwen1.5-MoE-A2.7B")
```

</details>

@ArthurZucker @yiliu30

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


docs/source/en/model_patching.md
```diff
@@ -0,0 +1,240 @@
...
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# Model patching
+
+Model patching allows you to temporarily replace model components during loading without modifying the original model code. This enables you to restructure models for specific requirements like quantization compatibility, apply optimizations, or experiment with architectural variants.
```
ArthurZucker: we should support `from_config` and `LlamaModel(config, monkey_patch)` maybe no? 

The idea is to allow maybe training cases where you don't load anything.

docs/source/en/model_patching.md
```diff
@@ -0,0 +1,240 @@
...
+Here's a complete example restructuring a Mixture-of-Experts (MoE) model. While this example focuses on quantization compatibility, the same approach works for other optimizations:
+
+```python
+from transformers import AutoModelForCausalLM, Patcher, WeightConverter, Concatenate
+import torch.nn as nn
+
+# Define restructured expert module for quantization compatibility
+class SequentialExperts(nn.ModuleList):
+    """MoE experts as a ModuleList instead of custom container.
+
```
ArthurZucker: imo registering what conversion is needed here is "explicit" about which conversion is needed by which class for example.

docs/source/en/model_patching.md
```diff
@@ -0,0 +1,240 @@
...
+        WeightConverter(
+            source_patterns=["q_proj.weight", "k_proj.weight", "v_proj.weight"],
+            target_patterns=["qkv_proj.weight"],
+            operations=[Concatenate(dim=0)],
+        ),
+        WeightConverter(
+            source_patterns=["q_proj.bias", "k_proj.bias", "v_proj.bias"],
+            target_patterns=["qkv_proj.bias"],
+            operations=[Concatenate(dim=0)],
+        ),
```
ArthurZucker: it would allow a more minimal call here once you define the `class_mapping`.

docs/source/en/model_patching.md
```diff
@@ -0,0 +1,240 @@
...
+        # ... rest of implementation
+
+# Create patcher with both class replacements and weight conversions
+patcher = Patcher(
+    class_mapping={
+        "Qwen2MoeExperts": SequentialExperts,
+        "Qwen2MoeAttention": FusedQKVAttention,
+    },
+    # Filter out existing conversions incompatible with new structure
+    filtered_weight_conversion_patterns=[".gate_up_proj", ".down_proj"],
```
ArthurZucker: I don't even think you have to do that since the replace pattern will create a key that does not exist -> will no be appllied

docs/source/en/model_patching.md
```diff
@@ -0,0 +1,240 @@
...
+            target_patterns=["qkv_proj.weight"],
+            operations=[Concatenate(dim=0)],
+        ),
+        WeightConverter(
+            source_patterns=["q_proj.bias", "k_proj.bias", "v_proj.bias"],
+            target_patterns=["qkv_proj.bias"],
+            operations=[Concatenate(dim=0)],
+        ),
+    ]
+)
```
ArthurZucker: would be nice, to stress the api, to add this for ALL models we have. Maybe in a more "progamatic" manner ?

ArthurZucker: this would allow to see how people who want to enable a feature more globally would do

src/transformers/modeling_utils.py
```diff
@@ -4024,9 +4025,12 @@ def from_pretrained(
 
         config.name_or_path = pretrained_model_name_or_path
         model_init_context = cls.get_init_context(dtype, is_quantized, _is_ds_init_called)
+
+        if patcher is not None:
+            model_init_context.append(patcher.get_patching_context(cls))
```
ArthurZucker: that's a good take

docs/source/en/model_patching.md
IlyasMoutawwakil: I can see it for `from_config` but not `LlamaModel(config, monkey_patch)` because currently LlamaModel (and all models) are defined this way:
```py
class LlamaModel(LlamaPreTrainedModel):
    def __init__(self, config: LlamaConfig):
        super().__init__(config)
....
```
the only way to pass the patcher would be to make all models inits take additional args/kwargs.

IlyasMoutawwakil: if weight loading is not an issue (in the case of instantiation withuot pretrained weights), the patcher could be used directly wih:
```py
with patcher.get_patching_context(LlamaModel):
      model = LlamaModel(config)
```

docs/source/en/model_patching.md
IlyasMoutawwakil: that's what i assumed but it seems the current weight loader can create keys that don't exist in the model and deletes the keys that were used to create them so you end up with gate_up_proj as UNEXPECTED and the individual expert weights as MISSING if we don't remove those conversions.

ArthurZucker: ah okay, that's something that might be unexpected

docs/source/en/model_patching.md
ArthurZucker:
```
with patcher.get_patching_context(LlamaModel):
      model = LlamaModel(config)
```
sounds great

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,114 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# Monkey patching
```
ArthurZucker:
```suggestion
# Monkey patching (experimental feature)
```
might want to mention?

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,114 @@
...
+
+Monkey patching allows you to replace model components globally without modifying the original model code. Once registered, patches are automatically applied when loading any model with [`~PreTrainedModel.from_pretrained`] or [`~PreTrainedModel.from_config`]. This enables you to restructure models for specific requirements like quantization compatibility, apply optimizations, or experiment with architectural variants.
+
+> [!WARNING]
+> **Monkey patching should be used as a last resort** when you need to change the layout and structure of a module and or its weights. For many customization and optimization needs, try using the [Attention interface](./attention_interface), [Experts interface](./experts_interface), or [Kernels registry](./kernel_doc/overview) instead. Only use monkey patching when you need structural changes that can't be achieved through custom forward implementations alone (e.g., for quantization library compatibility, fusing layers, or architectural experiments).
```
ArthurZucker, line 22: that's very nice!

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Clear all patches
+clear_monkey_patch_mapping()
+```
+
+## Important notes
+
+- **Weight handling**: Monkey patching only replaces classes, not weights. If your patched class has a different weights layout, you'll need to handle [weight conversions](./weightconverter) separately to ensure compatibility with pretrained weights.
+- **Global effect**: Patches are applied globally to all models loaded after registration. Be cautious when registering patches that may affect multiple models.
+
```
ArthurZucker: I would probably add a `usage` here with the quantization example from the issue, and finally the way to monkey patch all models to fuse QKV!

ArthurZucker: basically a small script that shows how to just fetch all AttentionLayers, create the converters automatically for q,k,v (and q_proj, k_proj, v_proj etc cases).

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,114 @@
...
+import torch.nn as nn
+
+class ModuleListExperts(nn.ModuleList):
+    """MoE experts as a ModuleList instead of packed parameters."""
+    def __init__(self, config):
+        super().__init__()
+        self.config = config
+        self.num_experts = config.num_experts
+        for _ in range(self.num_experts):
+            self.append(MoeMLP(config))
```
ArthurZucker, line 320: define this

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,114 @@
...
+            top_k_pos, token_idx = torch.where(expert_mask[expert_idx])
+            current_state = hidden_states[token_idx]
+            current_hidden_states = self[expert_idx](current_state)
+            current_hidden_states = current_hidden_states * top_k_weights[token_idx, top_k_pos, None]
+            final_hidden_states.index_add_(0, token_idx, current_hidden_states.to(final_hidden_states.dtype))
+        return final_hidden_states
+
+# Register the patch globally - it will be automatically applied
+register_monkey_patch_mapping(
+    mapping={"Qwen2MoeExperts": SequentialExperts}
```
ArthurZucker: you forgot the conversion mapping in tthis script

ArthurZucker: add the full `register_checkpoint_conversion_mapping` call as well here!

src/transformers/monkey_patch.py
```diff
@@ -0,0 +1,153 @@
...
+    modeling_module = importlib.import_module(model_class.__module__)
+    original_classes = {}
+
+    try:
+        for module_name, replacement_class in mapping.items():
+            if hasattr(modeling_module, module_name):
+                original_classes[module_name] = getattr(modeling_module, module_name)
+                setattr(modeling_module, module_name, replacement_class)
+            else:
+                logger.debug(f"Skipping patch for '{module_name}': not found in module '{modeling_module.__name__}'.")
```
ArthurZucker: okay, the only issue I see is for `output_recorder`! 
We probably need to check / update the target class / keep a reference to the old object so that: 
```python
    _can_record_outputs = {
        "hidden_states": LlamaDecoderLayer,
        "attentions": LlamaAttention,
    }
```
`LlamaAttention` will actually be of type `FusedQKVAttention` ?

ArthurZucker (commented):
A test with both `register_checkpoint_conversion_mapping` and the register for monkey patch is missing!

src/transformers/monkey_patch.py
IlyasMoutawwakil: very good catch !

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,424 @@
...
+You can use regular expressions to match multiple classes with a single pattern:
+
+```python
+from transformers.monkey_patch import register_monkey_patch_mapping
+
+# Match all classes containing "Attention"
+register_monkey_patch_mapping(
+    mapping={".*Attention": CustomAttention}
+)
+
```
ArthurZucker: that's very nice! 

I "think" we kinda want to introduce a flag for `from_pretrained` that's gonna be exactly this. 
`fuse_qkv`? 

For now we can just define it here with the correct mapping just to have a very thorough example.

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,424 @@
...
+from transformers.conversion_mapping import register_checkpoint_conversion_mapping, WeightConverter
+
+register_checkpoint_conversion_mapping(
+    model_type="bert",
+    mapping=[WeightConverter(
+        source_patterns=["query", "key", "value"],
+        target_patterns=["qkv"],
+        operations=[Concatenate(dim=0)]
+    )],
+    overwrite=True
```
ArthurZucker: indeed. I think we can also work with how that would look like when we want to monkey patch all attention. If no model type or model type is `"all"` maybe we can force use it? (in the sense that we do that for attention no?

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,424 @@
...
+            top_k_pos, token_idx = torch.where(expert_mask[expert_idx])
+            current_state = hidden_states[token_idx]
+            current_hidden_states = self[expert_idx](current_state)
+            current_hidden_states = current_hidden_states * top_k_weights[token_idx, top_k_pos, None]
+            final_hidden_states.index_add_(0, token_idx, current_hidden_states.to(final_hidden_states.dtype))
+        return final_hidden_states
+
+
+# Adapted from the original Qwen2MoeAttention
+class FusedQKVAttention(nn.Module):
```
ArthurZucker, line 340: ah ok it appears here

ArthurZucker (commented):
LGTM, I asked @Cyrilvallez for his input as well but very happy with it!

ArthurZucker:
cc @fxmarty and @fxmarty-amd that might interest you 😉

docs/source/en/monkey_patching.md
IlyasMoutawwakil: hmm actually this is gonna be very tricky, because attention modules do look different if you look closely.
for example with bert there are three classes, BertSelfAttention, BertCrossAttention and BertAttention which contains either of them. patching with a regular expression ".*Attention" is dangerous in this case because patches all three with the custom attention layer, which results in a messed up model 😅.

ArthurZucker: yep agreed agreed

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,436 @@
...
+rendered properly in your Markdown viewer.
+
+-->
+
+# Monkey patching (experimental feature)
+
+Monkey patching allows you to replace model components globally without modifying the original model code. Once registered, patches are automatically applied when loading any model with [`~PreTrainedModel.from_pretrained`] or [`~PreTrainedModel.from_config`]. This enables you to restructure models for specific requirements like quantization compatibility, apply optimizations, or experiment with architectural variants.
+
+> [!WARNING]
+> **Monkey patching should be used as a last resort** when you need to change the layout and structure of a module and or its weights. For many customization and optimization needs, try using the [Attention interface](./attention_interface), [Experts interface](./experts_interface), or [Kernels registry](./kernel_doc/overview) instead. Only use monkey patching when you need structural changes that can't be achieved through custom forward implementations alone (e.g., for quantization library compatibility, fusing layers, or architectural experiments).
```
Cyrilvallez, line 22: Thanks for emphasizing this!! ❤️

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,436 @@
...
+
+## Quick start
+
+Here's a simple example showing how to replace a model component:
+
+```python
+from transformers import AutoModelForCausalLM
+from transformers.models.gpt2.modeling_gpt2 import GPT2Attention
+from transformers.monkey_patch import register_monkey_patch_mapping
```
Cyrilvallez: It's extremely nit, but wdyt about making this a bit more recent by e.g. using Llama instead of the out-dated gpt2/bert for the examples?

docs/source/en/monkey_patching.md
Cyrilvallez: Indeed, not so easy in general!

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,436 @@
...
+        "^Llama\\d+Attention$": CustomAttention,  # Llama2Attention, Llama3Attention, etc.
+    }
+)
+```
+
+**Important**: Exact matches take precedence over patterns. If you register both `"BertAttention"` and `".*Attention"`, classes named `BertAttention` will use the exact-match replacement, while other matching classes will use the pattern-match replacement.
+
+> [!WARNING]
+> **Regex patterns can silently break models.** A broad pattern like `".*Attention"` will match *every* class whose name contains "Attention" — including container classes that wrap the attention you actually want to replace. For example, BERT has three attention-related classes: `BertSelfAttention` and `BertCrossAttention` (the inner attention implementations) and `BertAttention` (an outer module that *contains* one of those inner classes). Patching all three with the same custom attention layer produces a broken model because the outer `BertAttention` no longer wraps the inner one — it *is* one, eliminating expected sub-modules like `self` and `output`. Prefer narrow patterns (e.g., `".*SelfAttention$"`) or exact class names to avoid unintended matches.
```
Cyrilvallez, line 116: Again, thanks for being thorough on the potential issues!

src/transformers/monkey_patching.py
```diff
@@ -0,0 +1,357 @@
+# Copyright 2026 The HuggingFace Inc. team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
Cyrilvallez, line 3: Again, super super nit, but maybe let's rename this file `monkey_patching.py`? 🤓

src/transformers/monkey_patching.py
```diff
@@ -0,0 +1,357 @@
...
+        Compiled regex pattern or None if invalid
+    """
+    if pattern in _compiled_patterns_cache:
+        return _compiled_patterns_cache[pattern]
+
+    try:
+        compiled = re.compile(pattern)
+        _compiled_patterns_cache[pattern] = compiled
+        return compiled
+    except re.error as e:
```
Cyrilvallez, line 51: Is this the right exception type to catch here?

Cyrilvallez (approved):
Very happy about this overall!
Thanks a lot for putting the emphasis on everything that can go wrong, and thoroughly explain that this is for experimental purposes only!
I still see a few potential issues  apart from the `_can_record_outputs`, for example the `_init_weights`! I guess, one of the more reason why we want the `_init_weights` to become module-level functions @ArthurZucker! That way it would work out-of-the-box
Not an issue for pretrained models, so can be worked out later

docs/source/en/monkey_patching.md
IlyasMoutawwakil: will do ! using gpt2/bert is a bad habit of mine 😅

src/transformers/monkey_patching.py
IlyasMoutawwakil: makes sense !

src/transformers/monkey_patching.py
IlyasMoutawwakil: yess it's the general regex error

docs/source/en/monkey_patching.md
```diff
@@ -0,0 +1,435 @@
...
+rendered properly in your Markdown viewer.
+
+-->
+
+# Monkey patching (experimental feature)
+
+Monkey patching allows you to replace model components globally without modifying the original model code. Once registered, patches are automatically applied when loading any model with [`~PreTrainedModel.from_pretrained`] or [`~PreTrainedModel.from_config`]. This enables you to restructure models for specific requirements like quantization compatibility, apply optimizations, or experiment with architectural variants.
+
+> [!WARNING]
+> **Monkey patching should be used as a last resort** when you need to change the layout and structure of a module and or its weights. For many customization and optimization needs, try using the [Attention interface](./attention_interface), [Experts interface](./experts_interface), or [Kernels registry](./kernel_doc/overview) instead. Only use monkey patching when you need structural changes that can't be achieved through custom forward implementations alone (e.g., for quantization library compatibility, fusing layers, or architectural experiments).
```
fxmarty-amd, line 22: [Experts interface](https://github.com/huggingface/transformers/blob/54c838680d9605d170b92b4e2e7533b7e0ec3360/docs/source/en/experts_interface) link is dead (maybe not meant to be accessed through github https://github.com/huggingface/transformers/blob/v5.2.0/docs/source/en/experts_interface.md)

IlyasMoutawwakil: does it not link to https://huggingface.co/docs/transformers/main/en/experts_interface ?

fxmarty-amd: in github UI at https://github.com/huggingface/transformers/blob/54c838680d9605d170b92b4e2e7533b7e0ec3360/docs/source/en/monkey_patching.md no, but nevermind, this is meant to be accessed through hf.co/docs, not github ui.

PR #43919, opened 2026-02-11
Fix sync gradient

SunMarc:
# What does this PR do?

This PR enables to use `sync_each_batch` argument when passing `gradient_accumulation_kwargs` in `AcceleratorConfig`. I'm also removing `adjust_scheduler` docstring as it is not used/enabled for now as we don't prepare the scheduler with accelerate in Trainer. 

Fixes https://github.com/huggingface/transformers/issues/43899


ojh31 (approved):
LGTM, thanks for the quick fix!

qgallouedec (approved):

ArthurZucker (approved):

PR #43921, opened 2026-02-11
Depend on `typer` instead of `typer-slim`

svlandeg:
Hi 👋  Typer maintainer here.

# What does this PR do?

We used to offer `typer-slim` as a more light-weight package, which didn't depend on `rich` and `shellingham`. Unfortuntately, the way it was set up meant that we ran into issues with package managers (e.g. https://github.com/fastapi/typer/issues/1503). We ran a user survey and eventually decided to drop support for `typer-slim` altogether (since [0.22.0](https://github.com/fastapi/typer/releases/tag/0.22.0)). Going forward, it's better to just use `typer` as dependency.

What this means for this repo:
- `shellingham` already was an explicit requirement of `huggingface_hub` so was being pulled in anyway.
- `rich` will now be installed by default (about 12MB). This means that any Typer app will automatically set `rich_markup_mode` to `"rich"` by default (it was `None` when Rich wasn't installed). This can be overwritten when creating a `Typer()` app by setting `rich_markup_mode` to `None`. Alternatively, since Typer 0.22.0 there is a new env var `TYPER_USE_RICH` that you can set to `False` to disable Rich entirely everywhere and always. I'm not super sure what the best solution is for this repo, I did see that `rich` is already included in the `_deps` of `setup.py` though.

A similar [PR](https://github.com/huggingface/huggingface_hub/pull/3797) to `huggingface-hub` has already been merged.

Let me know if there's any concerns/questions!

## Before submitting
- [x] No related Github issue
- [x] I don't think there are documentation changes needed
- [x] I don't think new tests are needed?

## Who can review?


svlandeg:
Hm, I'm not sure why the CI fails, the tests flagged there don't seem related to this particular change 🤔

Rocketknight1:
Thank you for the heads-up, and the PR! A core dep change probably needs a core maintainer so cc @arthurzucker @vasqu @cyrilvallez

Cyrilvallez (approved):
LGTM thanks a lot for the PR!
Tests should pass now with pulling main - otherwise I'll force merge this one!

PR #43923, opened 2026-02-11
Revert #43897

Rocketknight1:
The `no_grad` decorators created problems, especially when new tensors are created! Reverting that PR until we can narrow down exactly which areas are problematic.


Cyrilvallez (approved):
Yup, thanks a lot!
As discussed offline, a backward with an empty Cache should have the same gradient as without Cache - however, due to `torch.cat` inside the Cache `update` method (which will return the same as it is cat'ing with an empty tensor as the Cache is empty), it will return a new tensor with `requires_grad=False` if it is under a `no_grad` context!

PR #43924, opened 2026-02-11
:rotating_light: [`Attn`] Remove all old mask APIs from modeling

vasqu:
As per title, removes the last remains of old mask API usage which means that now everything relies on the new mask API (except for a few exceptions that use 3D masks like idefics)

Adding a deprecation cylce to remove it from modeling utils


vasqu:
run-slow: big_bird,blip_2,bridgetower,clap,flava,ibert,instructblip,instructblipvideo,tapas,vilt

vasqu:
run-slow: align,altclip,big_bird,blip,blip_2,bridgetower,bros,canine,chinese_clip,clap,convbert,flava,ibert,idefics,imagegpt,instructblip,instructblipvideo,layoutlmv3,lightglue,lilt,longformer,longt5,luke,megatron_bert,mpnet,mra,nystromformer,perceiver,pix2struct,pop2piano,rembert,roformer,splinter,squeezebert,superglue,switch_transformers,tapas,tvp,udop,umt5,vilt,visual_bert

vasqu:
run-slow: align,altclip,big_bird,blip,blip_2,bridgetower,bros,canine,chinese_clip,clap,convbert,flava,ibert,idefics,imagegpt,instructblip,instructblipvideo,layoutlmv3,lightglue,lilt,longformer,longt5,luke,megatron_bert,mpnet,mra,nystromformer,perceiver,pix2struct,pop2piano,rembert,roformer,splinter,squeezebert,superglue,switch_transformers,tapas,tvp,udop,umt5,vilt,visual_bert

src/transformers/models/blip/modeling_blip_text.py
```diff
@@ -483,68 +484,6 @@ def get_input_embeddings(self):
     def set_input_embeddings(self, value):
         self.embeddings.word_embeddings = value
 
-    def get_extended_attention_mask(
```
vasqu, line 486: Shouldnt be breaking tbh because the module alternative still exists for now (with the same name) - I think this is a legacy remnant when module utils did not have these yet

src/transformers/models/blip/modeling_blip_text.py
```diff
@@ -661,6 +566,39 @@ def forward(
             cross_attentions=encoder_outputs.cross_attentions,
         )
 
+    def _create_attention_masks(
```
vasqu: I don't like this but it follows the bert pattern where it can be used as either encoder or decoder (with optional cross attn)

src/transformers/models/bridgetower/modeling_bridgetower.py
```diff
@@ -1218,8 +1218,11 @@ def forward(
 
         if attention_mask is None:
             attention_mask = torch.ones(input_shape, dtype=torch.long, device=input_ids.device)
-        extend_text_masks = self.text_model.get_extended_attention_mask(attention_mask, input_shape).to(
-            input_ids.device
+
+        extend_text_masks = create_bidirectional_mask(
+            config=self.config,
+            input_embeds=text_embeds[:, 0:1, :],  # weird case where the mask always wants q_len == 1
```
vasqu: happens a few times, I guess they just like to broadcast internally instead

src/transformers/models/git/modeling_git.py
```diff
@@ -70,34 +70,6 @@ class GitVisionModelOutput(ModelOutput):
     attentions: tuple[torch.FloatTensor, ...] | None = None
 
 
-# Copied from transformers.models.gemma3.modeling_gemma3.token_type_ids_mask_function
```
vasqu: this is not used anymore, not sure why this still existed tbh

src/transformers/models/idefics/modeling_idefics.py
```diff
@@ -1017,30 +1017,26 @@ def forward(
...
 
-        if image_hidden_states is not None:
-            image_batch_size, image_sequence_length, _ = image_hidden_states.size()
-            image_hidden_shape = (image_batch_size, image_sequence_length)
-            if image_attention_mask is None:
-                image_attention_mask = torch.ones(image_hidden_shape, device=device)
-            image_attention_mask = self.invert_attention_mask(image_attention_mask)
-        else:
-            image_attention_mask = None
+        # Mask is in 3D (incompatible with our mask API --> manual expansion)
```
vasqu, line 1021: this is the only case where we cannot avoid doing things manually because our mask API doesnt support 3D masks and tbh I dont think it's worth it for that (one) outlier

src/transformers/models/longformer/modeling_longformer.py
```diff
@@ -1498,16 +1499,22 @@ def forward(
...
+            config=self.config,
+            input_embeds=embedding_output,
+            attention_mask=torch.clamp(attention_mask, min=0, max=1),  # we have a mask of 0, 1, 2 (see bug below)
+            # Force mask creation
+            and_mask_function=lambda *args: torch.tensor(True, dtype=torch.bool),
+        )
+        # Cuts off back to 2D
+        extended_attention_mask = extended_attention_mask[:, 0, 0, :]
+        # Bug in the old mask API converted global masks (==2) to max dtype
+        extended_attention_mask[attention_mask == 2] = torch.finfo(embedding_output.dtype).max
```
vasqu, line 1516: buggy implementation, followed old behavior here

src/transformers/models/mra/modeling_mra.py
```diff
@@ -858,19 +859,24 @@ def forward(
...
             token_type_ids=token_type_ids,
             inputs_embeds=inputs_embeds,
         )
+
+        attention_mask = create_bidirectional_mask(
+            config=self.config,
+            input_embeds=embedding_output[:, 0:1, :],  # Force q_len == 1
+            attention_mask=attention_mask,
+            # Force mask creation
+            and_mask_function=lambda *args: torch.tensor(True, dtype=torch.bool),
```
vasqu, line 874: sometimes forcing the mask if we do expect it downstream

tests/models/bridgetower/test_modeling_bridgetower.py
```diff
@@ -307,7 +307,7 @@ class BridgeTowerModelTest(ModelTesterMixin, PipelineTesterMixin, unittest.TestC
     pipeline_model_mapping = {"feature-extraction": BridgeTowerModel} if is_torch_available() else {}
 
     is_training = False
-
+    test_torch_exportable = False
```
vasqu, line 310: few new failures re export but not worth fixing imo (at least for now)

src/transformers/models/blip/modeling_blip_text.py
Cyrilvallez: Agreed that it's not ideal (especially for creating the mask outside of the main forward, i.e. `create_mask_for_generate`), but still much cleaner than before.
Maybe we can have the paths directly in the `forward` instead of separate function though, wdyt? Oftentimes, I feel like it's easier to follow if we don't use separate functions (since the function is quite simple in itself, just 2 `if`s + mask primitives)

src/transformers/models/bridgetower/modeling_bridgetower.py
Cyrilvallez: If it's only for broadcasting, IMO we should drop this and use full length as always - resulting mask should not change

src/transformers/models/idefics/modeling_idefics.py
Cyrilvallez: Hard agree, let's not bloat general APIs for one old and peculiar model

src/transformers/models/longformer/modeling_longformer.py
Cyrilvallez: Humm, I don't see where the behavior was bugged here before based on the diffs?

src/transformers/modeling_utils.py
```diff
@@ -979,6 +989,11 @@ def get_extended_attention_mask(
         Returns:
             `torch.Tensor` The extended attention mask, with a the same dtype as `attention_mask.dtype`.
         """
+        logger.warning_once(
+            "Detected the usage of `get_extended_attention_mask`: This function is deprecated and will be removed in v5.12.0. "
+            "Please use the new API in `transformers.masking_utils`"
+        )
```
Cyrilvallez, line 995: Can't wait to remove them 🙏

tests/models/bridgetower/test_modeling_bridgetower.py
Cyrilvallez: Alright

Cyrilvallez (approved):
God's work 🙏 Thanks a lot for that, will simplify our lives so much in the future 🤗
I answered directly on most of your comments. but recommenting here to be sure:
- The `def _create_attention_masks` with only 2 `if`/`else` IMO we should remove and put them directly in the forward to ease readability
- For `bridgetower` (and others you mention) where we broadcast the q_len, we should be able to not slice and use usual logic without slicing then broadcasting no? Would be easier
- Not sure about `longformer` -> was it really a bug before? I don't see any issues about 2s in the mask from diffs

src/transformers/models/blip/modeling_blip_text.py
vasqu: Yup, moved them back into the main forward and removed the extra fns

src/transformers/models/longformer/modeling_longformer.py
vasqu: You can only see that it is bugged if you actual look into the mask values (either 0,1, or 2) and what `get_extended_attention_mask` does https://github.com/huggingface/transformers/blob/b75feb2af64c3e29cbbc1bd859958c5432cc7ed4/src/transformers/modeling_utils.py#L1010

Based on that line if you have a value of 2, you get `-1 * -(max dtype) == max dtype`

src/transformers/models/bridgetower/modeling_bridgetower.py
vasqu: This is tied to the old API https://github.com/huggingface/transformers/blob/b75feb2af64c3e29cbbc1bd859958c5432cc7ed4/src/transformers/modeling_utils.py#L998

It always broadcasted along q no matter what and it enabled bad practices imo because the downstream emebedding with the correct shape is probably created at a later time. I don't think it's worth to rewrite the model just for that :sweat_smile:

vasqu:
@Cyrilvallez I think I answered everything:
1. Done, moved it inside and removed the extra fns
2. https://github.com/huggingface/transformers/pull/43924#discussion_r3219213031 - this is tied to the fact that the actual embedding with correct shapes is created at a later time, imo not worth to rewrite the model for that
3. https://github.com/huggingface/transformers/pull/43924#discussion_r3219080392 - that is an actual bug, tried to explain it in that comment but it's not visible through the diff alone

src/transformers/models/longformer/modeling_longformer.py
Cyrilvallez: Not sure why we would have `2`s inside the mask, but trusting you!

Cyrilvallez:
ALright, thanks for the added explanations! All good to me!

PR #43928, opened 2026-02-11
fix(models): Preserve custom token IDs through DiaConfig save and load

harshaljanjani:
### What does this PR do?

The following failing Dia use case was identified and fixed in this PR:

→ Tests that created `DiaConfig` with custom token IDs (`eos_token_id=97` for a `vocab_size=100`) failed because saving then reloading the config would reset these values to defaults (`eos_token_id=1024`). The reason being that [DiaConfig](https://github.com/huggingface/transformers/blob/main/src/transformers/models/dia/configuration_dia.py#L249-L281) only set the attrs on the sub `decoder_config`, not on the main parent config, also leading to an IndexError during generation.
→ For more details on reproducing the bug and the output screenshots, please visit the linked issue!

Fixes #43927.

### Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/src/transformers/models/dia/configuration_dia.py
+++ b/src/transformers/models/dia/configuration_dia.py
@@ -145,6 +145,12 @@ class DiaDecoderConfig(PreTrainedConfig):
             Whether or not the model should return the last key/values attentions (not used by all models).
         is_encoder_decoder (`bool`, *optional*, defaults to `True`):
             Indicating that this model is part of an encoder-decoder architecture.
+        pad_token_id (`int`, *optional*, defaults to 1025):
+            The token id used for padding sequences to the same length within a batch.
+        eos_token_id (`int`, *optional*, defaults to 1024):
+            The token id representing the end-of-sequence token, indicating that generation should stop.
+        bos_token_id (`int`, *optional*, defaults to 1026):
+            The token id representing the beginning-of-sequence token, used to initialize decoding.
     """
 
     model_type = "dia_decoder"
@@ -170,6 +176,9 @@ class DiaDecoderConfig(PreTrainedConfig):
         initializer_range: float = 0.02,
         use_cache: bool = True,
         is_encoder_decoder: bool = True,
+        pad_token_id: int = 1025,
+        eos_token_id: int = 1024,
+        bos_token_id: int = 1026,
         **kwargs,
     ):
         self.max_position_embeddings = max_position_embeddings
@@ -190,6 +199,9 @@ class DiaDecoderConfig(PreTrainedConfig):
         self.initializer_range = initializer_range
         self.use_cache = use_cache
         self.rope_parameters = rope_parameters
+        self.pad_token_id = pad_token_id
+        self.eos_token_id = eos_token_id
+        self.bos_token_id = bos_token_id
 
         super().__init__(is_encoder_decoder=is_encoder_decoder, **kwargs)
 
@@ -213,12 +225,15 @@ class DiaConfig(PreTrainedConfig):
             The epsilon used by the normalization layers.
         is_encoder_decoder (`bool`, *optional*, defaults to `True`):
             Indicating that this model uses an encoder-decoder architecture.
-        pad_token_id (`int`, *optional*, defaults to 1025):
-            Padding token id.
-        eos_token_id (`int`, *optional*, defaults to 1024):
-            End of stream token id.
-        bos_token_id (`int`, *optional*, defaults to 1026):
-            Beginning of stream token id.
+        pad_token_id (`int`, *optional*):
+            Deprecated. Please set this on `DiaDecoderConfig` directly. If provided, it will be forwarded
+            to `decoder_config`.
+        eos_token_id (`int`, *optional*):
+            Deprecated. Please set this on `DiaDecoderConfig` directly. If provided, it will be forwarded
+            to `decoder_config`.
+        bos_token_id (`int`, *optional*):
+            Deprecated. Please set this on `DiaDecoderConfig` directly. If provided, it will be forwarded
+            to `decoder_config`.
         delay_pattern (`list[int]`, *optional*, defaults to `[0, 8, 9, 10, 11, 12, 13, 14, 15]`):
             The delay pattern for the decoder. The length of this list must match `decoder_config.num_channels`.
         initializer_range (`float`, *optional*, defaults to 0.02):
@@ -252,9 +267,9 @@ class DiaConfig(PreTrainedConfig):
         decoder_config: DiaDecoderConfig | None = None,
         norm_eps: float = 1e-5,
         is_encoder_decoder: bool = True,
-        pad_token_id: int = 1025,
-        eos_token_id: int = 1024,
-        bos_token_id: int = 1026,
+        pad_token_id: int | None = None,
+        eos_token_id: int | None = None,
+        bos_token_id: int | None = None,
         delay_pattern: list[int] | None = None,
         initializer_range: float = 0.02,
         use_cache: bool = True,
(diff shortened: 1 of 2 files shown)


src/transformers/models/dia/configuration_dia.py
```diff
@@ -270,6 +270,9 @@ def __init__(
         self.delay_pattern = delay_pattern if delay_pattern is not None else [0, 8, 9, 10, 11, 12, 13, 14, 15]
         self.initializer_range = initializer_range
         self.use_cache = use_cache
+        self.pad_token_id = pad_token_id
+        self.eos_token_id = eos_token_id
+        self.bos_token_id = bos_token_id
```
zucchini-nlp: imo we need to remote `pad_token_id: int = 1025, eos_token_id: int = 1024, bos_token_id: int = 1026` in longer term. They have to be in the config where it's actually set as attr

For BC we can keep it and set default to `None`, that should work since `save_pretrained` wasn't saving it anyway in the main config. So smth like

python
```
def __init__(self,
pad_token_id: int = None,
        eos_token_id: int = None,
        bos_token_id: int = None,

# We could raise a deprecation warning here, but first we need to update the official ckpt in `nari-labs` org
if pad_token_id is not None:
    logger.maybe_warn("Please pass you pad token to the config where it belongs!")
    self.decoder_config.pad_token_id = pad_token_id

```

zucchini-nlp (commented):
Thanks for catching this, I've also been thinking to get rid of this pattern across the repo haha. I'm not 100% sure but there were a few more similar models, t5gemma and maybe some more. Would be great to check if you have bandwidth. or I'll just let you know later in the comments :)

harshaljanjani:
@zucchini-nlp That makes sense! I've left a comment here with the findings within the time I had for your perusal.
For now, I'll resolve this according to your review and ensured it’s BC, maybe on an as-needed basis, I could open a separate issue to address the points below if they are indeed unwanted anti-patterns, lmk. Happy to help fix them down the line as well.
Just a nit: Dia is the only one that "broke" because of the anti-pattern. I've listed what I found in T5Gemma and others as well, but the others seem to be working just fine in terms of token ID preservation, again repro-ed this for the listed models as well :)

→ I've listed the hyperlinks with the correct line numbers for each reference; feel free to cross-check my findings :)

**What was I searching for** → Accept pad_token_id/eos_token_id/bos_token_id in `__init__`
→ Forward them to a sub-config (`self.decoder_config.eos_token_id = ..`)
→ Fail to persist them on the main config obj or pass them to `super().__init__()`

**What I found** → Wasn't able to find other models other than Dia with the actual save/load bug; other composite configs set token IDs on a sub-config, but make sure it persists in the main config, speaking of which, the patterns are:

| Pattern | Models | Buggy?? |
|----------|--------|--------|
| A. Force-set on sub-config only: [Takes token IDs](https://github.com/huggingface/transformers/blob/main/src/transformers/models/dia/configuration_dia.py#L255-L257) and [sets them on `self.decoder_config.*`](https://github.com/huggingface/transformers/blob/main/src/transformers/models/dia/configuration_dia.py#L273-L275), but never on `self` → lost on save / load and defaults overwrite sub-config values. | Could only find DiaConfig (before fix), happy to know if I missed any others | Yes |
| B. Extract from sub-config → inject into kwargs: Does not accept token IDs, instead pulls them from the decoder sub-config and injects into `**kwargs` before calling `super().__init__()`. | [T5GemmaConfig](https://github.com/huggingface/transformers/blob/main/src/transformers/models/t5gemma/configuration_t5gemma.py#L271-L275), [T5Gemma2Config](https://github.com/huggingface/transformers/blob/main/src/transformers/models/t5gemma2/configuration_t5gemma2.py#L600-L610) | No (works, but this is quite like the anti-pattern you rightly mentioned) |
| C. Set on self directly → Configs that accept token IDs then set `self` (`pad_token_id = pad_token_id` before / after `super().__init__()`); [`PreTrainedConfig.__init__` does not explicitly handle these attributes](https://github.com/huggingface/transformers/blob/main/src/transformers/configuration_utils.py#L264-L267) and they persist. | [SeamlessM4TConfig](https://github.com/huggingface/transformers/blob/main/src/transformers/models/seamless_m4t/configuration_seamless_m4t.py#L405-L411), [SeamlessM4Tv2Config](https://github.com/huggingface/transformers/blob/main/src/transformers/models/seamless_m4t_v2/configuration_seamless_m4t_v2.py#L414-L420), [MoshiConfig](https://github.com/huggingface/transformers/blob/main/src/transformers/models/moshi/configuration_moshi.py#L333-L336), [CsmConfig](https://github.com/huggingface/transformers/blob/main/src/transformers/models/csm/configuration_csm.py#L343-L347), [KyutaiSpeechToTextConfig](https://github.com/huggingface/transformers/blob/main/src/transformers/models/kyutai_speech_to_text/configuration_kyutai_speech_to_text.py#L189-L193) | No |

→ **Additional:** [`Pix2StructConfig`](https://github.com/huggingface/transformers/blob/main/src/transformers/models/pix2struct/configuration_pix2struct.py#L330-L331) reads token IDs from `text_config` (`self.eos_token_id = self.text_config.eos_token_id`) which also works correctly, but I'm not too sure if that's warranted.

Would love to hear your suggestions for which ones you're looking to fix asap and probably we could move to writing issues for them, happy to help :)

src/transformers/models/dia/configuration_dia.py
harshaljanjani: Resolved; I've updated the tests and docs, and added a TODO to avoid missing the `nari-labs` org config update (happy to remove if unnecessary). Tests pass without the change to the testing file as well, but maybe this should make the pattern clearer for readers (? :))

zucchini-nlp:
@harshaljanjani thanks a lot for your research. The T5gemma seems to work because it actually assigns the attr in subconfig and also in the main config, thus it will save it in both places. Let's fix Dia for now and I will think of unifying these later, I am not 100% sure if it'll disrupt downstream usage for other models if we just delete the attr

src/transformers/models/dia/configuration_dia.py
```diff
@@ -270,9 +285,27 @@ def __init__(
         self.delay_pattern = delay_pattern if delay_pattern is not None else [0, 8, 9, 10, 11, 12, 13, 14, 15]
         self.initializer_range = initializer_range
         self.use_cache = use_cache
-        self.decoder_config.pad_token_id = pad_token_id
-        self.decoder_config.eos_token_id = eos_token_id
-        self.decoder_config.bos_token_id = bos_token_id
+
+        # TODO: Remove token ID forwarding once the `nari-labs/Dia-1.6B`
+        # checkpoint is updated
```
zucchini-nlp, line 290: Can you open a PR in nari-labs repo and link to this issue?

zucchini-nlp (approved):
Nice, thanks a lot!

harshaljanjani:
Opened a PR at https://github.com/nari-labs/dia with the change; links the issue and the PR.

zucchini-nlp:
I meant more like the hub one 😅  Opened https://huggingface.co/nari-labs/Dia-1.6B-0626/discussions/7 as well :)

harshaljanjani:
> I meant more like the hub one 😅  Opened https://huggingface.co/nari-labs/Dia-1.6B-0626/discussions/7 as well :)

Oops, my bad! Should I close the other PR then? 😅

zucchini-nlp:
No, you can leave it. They would need to update their GH repo if hub config changes are merged!

harshaljanjani:
> No, you can leave it. They would need to update their GH repo if hub config changes are merged!

Perfect, sounds good!

PR #43932, opened 2026-02-12
Inclusion of process_group in the gather_full_tensor function in tensor_parallel.py

quic-meetkuma:
# What does this PR do?

It fixes minor issue in the gather_full_tensor function. The existing implementation tries to perform all_gather across all the ranks of distributed training. But when we talk about the case of TP+DP then we only need to perform all_gather across TP ranks, not the DP ranks. This change makes use of group argument in dist.all_gather API of pytorch to perform it correctly.

## Before submitting
      Pull Request section?
      to it if that's the case.

CC: @SunMarc @ArthurZucker @ydshieh


ArthurZucker (approved):
LGTM

PR #43934, opened 2026-02-12
[Quantization] Add metal quantization for MPS devices!

MekkCyber:
# What does this PR do?

Adds mlx quantization for mps devices leveraging the `kernels` library for pre-built kernels !!


src/transformers/integrations/mlx_quantization.py
```diff
@@ -0,0 +1,334 @@
...
+    w_flat = torch.zeros(N, K, dtype=torch.float32, device=w_packed.device)
+    for i in range(elems_per_int):
+        w_flat[:, i::elems_per_int] = ((w_packed_i >> (bits * i)) & max_val).float()
+
+    w_grouped = w_flat.reshape(N, -1, group_size)
+    w_deq = w_grouped * scales.float().unsqueeze(-1) + biases.float().unsqueeze(-1)
+    return w_deq.reshape(N, K)
+
+
+class MlxQuantize(ConversionOps):
```
ArthurZucker: Nice!!!

src/transformers/integrations/metal_quantization.py
```diff
@@ -0,0 +1,334 @@
...
+
+        qbiases = input_dict.get("qbiases")
+        qbiases = qbiases[0] if isinstance(qbiases, list) else qbiases
+
+        if scales is None or qbiases is None:
+            # Only the weight was found; return as-is (probably already float)
+            return {full_layer_name: quantized}
+
+        w_deq = _affine_dequantize_tensor(quantized, scales, qbiases, group_size, bits)
+        return {full_layer_name: w_deq}
```
ArthurZucker: nice as well!

src/transformers/utils/quantization_config.py
```diff
@@ -1902,3 +1903,57 @@ def to_dict(self) -> dict[str, Any]:
...
+        self.post_init()
+
+    def post_init(self):
+        if self.bits not in (2, 4, 8):
+            raise ValueError(f"MLX quantization only supports bits in {{2, 4, 8}}, got {self.bits}")
+        if self.group_size <= 0:
+            raise ValueError(f"group_size must be positive, got {self.group_size}")
+
+    def get_loading_attributes(self):
+        return {"dequantize": self.dequantize}
```
ArthurZucker, line 1946:
```suggestion
        return {"dequantize": self.dequantize, "modules_to_not_convert": modules_to_not_convert}
```
this applies to all methods actually but otherwise you can't overwrite that in a from pretrained call

ArthurZucker (commented):
Super nice, missing some tests tho!!

src/transformers/integrations/metal_quantization.py
```diff
@@ -0,0 +1,334 @@
...
+            # (dtype=None → PyTorch default float). The MlxQuantize ConversionOp
+            # will transform this into packed uint32 + scales + qbiases.
+            self.weight = nn.Parameter(torch.zeros(out_features, in_features, dtype=dtype), requires_grad=False)
+
+        # For pre-quantized (dtype==uint32): scales/qbiases will be loaded
+        # from the checkpoint in whatever dtype they were saved as.
+        # For quantize-on-the-fly (dtype==None/float): MlxQuantize will set
+        # them in the original weight's dtype (e.g. bfloat16).
+        # We use float16 as the placeholder — it gets overwritten either way.
+        scales_dtype = torch.float16 if dtype == torch.uint32 else (dtype or torch.float16)
```
SunMarc: I don't think so no ? It will try to use this dtype and not the one from the checkpoint.

src/transformers/integrations/metal_quantization.py
```diff
@@ -0,0 +1,334 @@
...
+        self.qbiases = nn.Parameter(torch.zeros(out_features, n_groups, dtype=scales_dtype), requires_grad=False)
+
+        if bias:
+            self.bias = nn.Parameter(torch.zeros(out_features))
+        else:
+            self.register_parameter("bias", None)
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        if self.weight.dtype not in (torch.uint32, torch.int32):
+            return nn.functional.linear(input, self.weight, self.bias)
```
SunMarc: the weight can be either in torch.int32 or torch.uint32 ?

src/transformers/integrations/metal_quantization.py
```diff
@@ -0,0 +1,334 @@
...
+
+        if isinstance(module, nn.Linear):
+            # Following the FP8 pattern: when pre-quantized, don't pass dtype
+            # so the default (torch.uint32) is used → packed uint32 weights.
+            # When quantize-on-the-fly, pass dtype=None so the weight is
+            # created with full shape in the default float dtype, allowing the
+            # loader to fill it before MlxQuantize converts it.
+            module_kwargs = {} if pre_quantized else {"dtype": None}
+
+            with torch.device("meta"):
```
SunMarc: we don't need with context manager

src/transformers/integrations/metal_quantization.py
```diff
@@ -0,0 +1,334 @@
...
+    elems_per_int = 32 // bits
+    max_val = (1 << bits) - 1
+    n_groups = K // group_size
+
+    w_grouped = weight.float().reshape(N, n_groups, group_size)
+    w_min = w_grouped.min(dim=-1).values  # [N, n_groups]
+    w_max = w_grouped.max(dim=-1).values
+
+    scales = ((w_max - w_min) / max_val).clamp(min=1e-8)
+    biases = w_min
```
SunMarc: for scales and biases, they are not necessarily in fp32

src/transformers/integrations/mlx_quantization.py
```diff
@@ -0,0 +1,334 @@
...
+        bias_key = f"{base}.qbiases" if base else "qbiases"
+
+        # Store scales/qbiases in the same dtype as the original weight so the
+        # Metal kernel (which templates on x.scalar_type()) can use them
+        # directly without a per-forward cast.
+        orig_dtype = value.dtype  # e.g. bfloat16 for Llama
+        return {
+            target_key: w_packed,
+            scale_key: scales.to(orig_dtype),
+            bias_key: biases.to(orig_dtype),
```
SunMarc: fine but I think `_affine_quantize_tensor` should return them in the right dtype already

src/transformers/integrations/metal_quantization.py
```diff
@@ -0,0 +1,334 @@
...
+        quantized = input_dict.get("weight$")
+        if quantized is None:
+            quantized = next(iter(input_dict.values()))
+        quantized = quantized[0] if isinstance(quantized, list) else quantized
+
+        scales = input_dict.get("scales")
+        scales = scales[0] if isinstance(scales, list) else scales
+
+        qbiases = input_dict.get("qbiases")
+        qbiases = qbiases[0] if isinstance(qbiases, list) else qbiases
```
SunMarc: simplify this 
we can just do 
```python
quantized = input_dict["weight$"][0]
scales =input_dict["scales"][0]
qbiases =input_dict["qbiases"][0]
```

src/transformers/integrations/metal_quantization.py
```diff
@@ -0,0 +1,334 @@
...
+
+        if scales is None or qbiases is None:
+            # Only the weight was found; return as-is (probably already float)
+            return {full_layer_name: quantized}
```
SunMarc: we can do as in fp8
```python
if len(input_dict) < 2:
            # case where we only got weights, need to check for "weight$"
            return {full_layer_name: input_dict["weight$"]}
```

src/transformers/integrations/metal_quantization.py
```diff
@@ -0,0 +1,334 @@
...
+
+        w_deq = _affine_dequantize_tensor(quantized, scales, qbiases, group_size, bits)
```
SunMarc, line 295: so this is returning a fp32 value but it the user specified bf16, it think there will be an issue no ?

src/transformers/quantizers/quantizer_mlx.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+
+    requires_calibration = False
+
+    def __init__(self, quantization_config, **kwargs):
+        super().__init__(quantization_config, **kwargs)
+
+    def validate_environment(self, *args, **kwargs):
+        if not is_torch_available():
+            raise ImportError("MLX quantization requires PyTorch. Please install torch.")
```
SunMarc: we can remove that no ?

src/transformers/quantizers/quantizer_metal.py
```diff
@@ -0,0 +1,133 @@
...
+
+        if not is_accelerate_available():
+            raise ImportError("MLX quantization requires accelerate: `pip install accelerate`")
+
+        device_map = kwargs.get("device_map")
+        if device_map is None:
+            logger.warning_once(
+                "You have loaded an MLX quantized model on CPU and have an MPS device available. "
+                "Set device_map='mps' to use the Metal kernels."
+            )
```
SunMarc, line 67: we can maybe overwrite ?

src/transformers/quantizers/quantizer_mlx.py
```diff
@@ -0,0 +1,133 @@
...
+                self.quantization_config.dequantize = True
+                return
+            else:
+                raise RuntimeError("MLX quantization requires an Apple Silicon GPU (MPS). No MPS device found.")
+
+        if not is_kernels_available():
+            raise ImportError("MLX quantization requires kernels: `pip install kernels`")
+
+        if not is_accelerate_available():
+            raise ImportError("MLX quantization requires accelerate: `pip install accelerate`")
```
SunMarc: I think we can remove this

src/transformers/quantizers/quantizer_metal.py
```diff
@@ -0,0 +1,133 @@
...
+
+        model = replace_with_mlx_linear(
+            model,
+            modules_to_not_convert=self.modules_to_not_convert,
+            quantization_config=self.quantization_config,
+            pre_quantized=self.pre_quantized,
+        )
+
+    def _process_model_after_weight_loading(self, model: "PreTrainedModel", **kwargs):
+        pass
```
SunMarc: remove

src/transformers/utils/quantization_config.py
```diff
@@ -1902,3 +1903,57 @@ def to_dict(self) -> dict[str, Any]:
...
+    This quantization method uses the ``quantization-mlx`` Metal kernels from the Hugging Face Hub
+    to perform affine quantization (scales + qbiases) with configurable bit-width and group size.
+    The quantized weights are packed into ``uint32`` tensors and the forward pass uses fused
+    dequantization + matmul Metal kernels.
+
+    Args:
+        bits (`int`, *optional*, defaults to `4`):
+            Number of bits per weight element. Supported values are 2, 4, and 8.
+        group_size (`int`, *optional*, defaults to `128`):
+            Number of elements sharing the same scale/bias pair.
```
SunMarc: maybe we can choose a better default for this one, either 64 or 32.

src/transformers/utils/quantization_config.py
SunMarc: this is when we reload the model, so we shouldn't be able to change this. Otherwise, it will lead to errors

SunMarc (commented):
Thanks ! Maybe we can change the name to Metal insteal of Mlx as it can create confusion ? In the future, we might have mlx if we add compatibility with mlx models. Please add some e2e tests + add tests to check that we have the right dtype after quantization and dequantization

src/transformers/integrations/metal_quantization.py
```diff
@@ -0,0 +1,334 @@
...
+    def __init__(self, hf_quantizer):
+        self.hf_quantizer = hf_quantizer
+
+    def convert(self, input_dict: dict, full_layer_name: str | None = None, **kwargs) -> dict:
+        bits = self.hf_quantizer.quantization_config.bits
+        group_size = self.hf_quantizer.quantization_config.group_size
+
+        # The WeightConverter collects tensors matching source_patterns
+        quantized = input_dict.get("weight$")
+        if quantized is None:
```
MekkCyber: check if quantized is in bf/fp or in unit32 here

src/transformers/integrations/mlx_quantization.py
MekkCyber: not sure about this, since when we quantize we keep the scales dtype the same as the weight dtype before quantization which is `float32`

src/transformers/integrations/metal_quantization.py
MekkCyber: shouldn't be a problem since we are casting to the correct dtype inside the forward pass kernel

SunMarc (commented):
Thanks a lot ! Can you just update the overview docs to add this quantization method ?

SunMarc (approved):

PR #43936, opened 2026-02-12
Fix failed unit tests for moonshine_streaming model

kaixuanliu:
@ydshieh , pls help review, thx!


kaixuanliu:
Failed cases listed:
```
FAILED tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py::MoonshineStreamingModelTest::test_cpu_offload - AttributeError: 'MoonshineStreamingModel' object has no attribute 'hf_device_map'
FAILED tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py::MoonshineStreamingModelTest::test_disk_offload_bin - AttributeError: 'MoonshineStreamingModel' object has no attribute 'hf_device_map'
FAILED tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py::MoonshineStreamingModelTest::test_disk_offload_safetensors - AttributeError: 'MoonshineStreamingModel' object has no attribute 'hf_device_map'
FAILED tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py::MoonshineStreamingModelTest::test_model_parallelism - AttributeError: 'MoonshineStreamingModel' object has no attribute 'hf_device_map'
FAILED tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py::MoonshineStreamingModelIntegrationTests::test_medium_logits_batch - AssertionError: Tensor-likes are not close!
FAILED tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py::MoonshineStreamingModelIntegrationTests::test_small_logits_batch - AssertionError: Tensor-likes are not close! 
```

kaixuanliu:
@IlyasMoutawwakil , could you help review? Thx!

tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py
```diff
@@ -162,6 +162,22 @@ def test_can_init_all_missing_weights(self):
     def test_init_weights_can_init_buffers(self):
         self.skipTest("MoonshineStreaming uses special buffer initialization that conflicts with this test")
 
+    @unittest.skip(reason="Does not work on the tiny model as we keep hitting edge cases.")
```
IlyasMoutawwakil: not sure what this comment mean, what's the edge case ?

kaixuanliu: The edge case here means when the input model is very tiny(~161KB), [infer_auto_device_map](https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/accelerate.py#L351) is called with [max_memory={0: 112865, 'cpu': 322472}], and it causes the `device_map` here is `OrderedDict({'': 'cpu'})`, which causes `hf_device_map` is never created.

kaixuanliu: I have updated the comments to be more clear here.

kaixuanliu:
@IlyasMoutawwakil can you help review it again? Thx!

tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py
IlyasMoutawwakil: are you sure about this ? doesn't make a lot of sense tbh, all models used for test are tiny, it only needs two parameters to be splitable across cpu and gpu

kaixuanliu: I double checked the debug output, it is correct. But maybe some root cause reason of why `device_map` here is `OrderedDict({'': 'cpu'})` should be added: In the test model, the single `decoder.pos_emb` layer occupies 81% of the weight param, which exceeds the size defined in [L2936](https://github.com/huggingface/transformers/blob/main/tests/test_modeling_common.py#L2936) (self.model_split_percents: [0.5, 0.7, 0.9]), hence it puts the entire model on CPU side, and hence cause the crash. It seems it is better to update the skip reason again.

kaixuanliu: There are 2 options: 1. update the skip reason to make it more accurate 2. modify the `model_split_percents` value for this model. Which one do you prefer?

IlyasMoutawwakil: thanks for investigating ! makes sense for me to update the model split precents

kaixuanliu: Updated the code and fixed 1 failed test case:
`pytest -rA tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py::MoonshineStreamingModelTest::test_model_parallelism`

kaixuanliu:
@IlyasMoutawwakil Can you help review again? Thx!

tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py
IlyasMoutawwakil: thanks for investigating and fixing !

src/transformers/models/moonshine_streaming/modeling_moonshine_streaming.py
```diff
@@ -822,7 +822,7 @@ def forward(
         position_embeddings = self.pos_emb(
             torch.arange(encoder_hidden_states.shape[1], device=encoder_hidden_states.device)
         )
-        encoder_hidden_states += position_embeddings
+        encoder_hidden_states = encoder_hidden_states + position_embeddings.to(encoder_hidden_states.device)
```
IlyasMoutawwakil: can keep the += here, why would the position_embeddings have a different device ?

tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py
```diff
@@ -162,6 +162,10 @@ def test_can_init_all_missing_weights(self):
     def test_init_weights_can_init_buffers(self):
         self.skipTest("MoonshineStreaming uses special buffer initialization that conflicts with this test")
 
+    # MoonshineStreaming test model's pos_emb layer is large relative to total model size,
+    # requiring higher GPU allocation percentage to enable device splitting
+    model_split_percents = [0.5, 0.9, 0.95]
```
IlyasMoutawwakil, line 167: nice fix !

src/transformers/models/moonshine_streaming/modeling_moonshine_streaming.py
kaixuanliu: When using model parallism, self.pos_emb may be on a different device than encoder_hidden_states.

kaixuanliu: You can reproduce following:
```
export RUN_SLOW=1
export CUDA_VISIBLE_DEVICES=0,1
pytest -rA tests/models/moonshine_streaming/test_modeling_moonshine_streaming.py::MoonshineStreamingModelTest::test_model_parallelism
```

IlyasMoutawwakil (approved):
lgtm!

kaixuanliu:
@IlyasMoutawwakil Hi, can we merge it?

PR #43944, opened 2026-02-12
Fix flaky test for multimodal LLMs

Rocketknight1:
We have flaky test failures in `tests/models/qwen3_omni_moe/test_modeling_qwen3_omni_moe.py::Qwen3OmniMoeThinkerForConditionalGenerationModelTest::test_generate_continue_from_past_key_values`. The cause is that the logic in this test drops multimodal inputs (because it's passing `past_key_values` to generate instead), but it fails to drop **audio** inputs correctly. This causes flaky failures in LLMs like Qwen3OmniMoE, which I think occur when the audio encoder creates audio features that overwrite the embeddings for generated tokens (maybe when the tokens have the `audio_token_id` value?)

I'm not 100% sure of the cause, but the test fails 3% of the time without this fix and 0% with it, so I'm pretty sure dropping inputs correctly is the solution!

cc @zucchini-nlp

### DIFF
--- a/tests/generation/test_utils.py
+++ b/tests/generation/test_utils.py
@@ -1343,7 +1343,10 @@ class GenerationTesterMixin:
             # if there are multimodal data which don't belong anywhere inside `text_tokens`
             keys_to_pop = []
             for key in inputs:
-                if ("pixel" in key or key in ["image_patches", "input_feature"]) and key != model.main_input_name:
+                if (
+                    "pixel" in key
+                    or key in ["image_patches", "input_feature", "input_features", "feature_attention_mask"]
+                ) and key != model.main_input_name:
                     keys_to_pop.append(key)
             for key in keys_to_pop:
                 inputs.pop(key)


tests/generation/test_utils.py
```diff
@@ -1343,7 +1331,10 @@ def test_generate_continue_from_past_key_values(self):
             # if there are multimodal data which don't belong anywhere inside `text_tokens`
             keys_to_pop = []
             for key in inputs:
-                if ("pixel" in key or key in ["image_patches", "input_feature"]) and key != model.main_input_name:
+                if (
+                    "pixel" in key
+                    or key in ["image_patches", "input_feature", "input_features", "feature_attention_mask"]
```
zucchini-nlp, line 1348: `input_features` is a quite common name across audio-models like whisper. Can you verify that the test passes for all models?
We can as well override it in qwen3-omni-moe test file, if it fails for other audio models

zucchini-nlp (approved):
Approving so you can merge after checking that the test passes for other audio models :)

zucchini-nlp:
Btw, the failing Bark was fixed on `main` :)

tests/generation/test_utils.py
Rocketknight1: Yes! Dropping that input is intentional even when it's present on the model, unless it's the `model_main_input`. Because we're passing `past_key_values` to generation, we don't actually want to run multimodal inputs through the encoder again, and doing that is what caused the failure in the past.

Rocketknight1:
Ran the test for every model, and this change didn't cause any new failures, but did fix the Qwen3 test!

PR #43951, opened 2026-02-12
Fix AttributeError for remote models with trust_remote_code=True

lordaarush:
# What does this PR do?

Fixes #43883

After #42270, `all_tied_weights_keys` is initialized in `post_init()`, but remote models loaded with `trust_remote_code=True` don't always call `post_init()` properly, causing `AttributeError` when loading models like Molmo.

This fix adds defensive checks in two methods:
- `_adjust_tied_keys_with_tied_pointers()`: Initialize empty dict if missing, then detect tied weights via data pointers
- `mark_tied_weights_as_initialized()`: Return early if attribute missing

This allows remote models to load successfully while maintaining tied weight detection.

## Before submitting

## Who can review?

@Rocketknight1 @Cyrilvallez

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -4436,6 +4436,8 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         If we don't add them back in `self.all_tied_weights_keys`, they will be re-initialized
         and all params in tied group get random weights.
         """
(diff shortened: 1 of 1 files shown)


Cyrilvallez:
See my answer here https://github.com/huggingface/transformers/issues/43883#issuecomment-3891672810! In the meantime, I'm closing the PR! Let me know if you need anything else 🤗

PR #43953, opened 2026-02-12
Fix conversion mapping for Qwen3VL-MoE

zucchini-nlp:
Fixes https://github.com/huggingface/transformers/issues/43931, no transpose needed after standardizing the model impl to inherit from Qwen3-MoE

### DIFF
--- a/src/transformers/conversion_mapping.py
+++ b/src/transformers/conversion_mapping.py
@@ -52,6 +52,7 @@ _MODEL_TO_CONVERSION_PATTERN = {
     "longcat_flash": "qwen2_moe",
     "solar_open": "qwen2_moe",
     "qwen3_moe": "qwen2_moe",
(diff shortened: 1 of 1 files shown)


Cyrilvallez:
Nop, this is not the right conversion. Weights are already merged into the final experts, but need transposition, see https://github.com/huggingface/transformers/pull/43913!

Cyrilvallez:
Though if all weights ONLY need the transpose, that PR would be much simpler actually - gonna check

zucchini-nlp:
>  Weights are already merged 

Ah right, didn't see that. Indeed, we shouldn't need to merge. In that case no conversion is needed ig

Cyrilvallez:
They still need to be transposed 🥲 Finally fixed it in https://github.com/huggingface/transformers/pull/44037

PR #43954, opened 2026-02-12
Fix LASR test layerdrop issue

Rocketknight1:
The LASR model uses `layerdrop`, which we forgot to disable in the tests. Since the tests only have 2 hidden layers, and the default layerdrop chance is `0.1`, this means there's a `0.1^2 = 1%` chance in any test that doesn't call `model.eval()` that the model will drop **all** of the hidden layers, which breaks lots of tests (for example, there will be no output attentions when this happens).

This PR disables layerdrop in the test config, which resolves the test flakiness (example failing run: https://app.circleci.com/pipelines/github/huggingface/transformers/163931/workflows/272c0cd9-df40-4b38-8e9b-21b12dff3386/jobs/2160291)

cc @tarekziade

### DIFF
--- a/tests/models/lasr/test_modeling_lasr.py
+++ b/tests/models/lasr/test_modeling_lasr.py
@@ -54,6 +54,7 @@ class LasrEncoderModelTester:
         subsampling_conv_channels=32,
         subsampling_conv_kernel_size=5,
         subsampling_conv_stride=2,
+        layerdrop=0.0,
     ):
         # testing suite parameters
         self.parent = parent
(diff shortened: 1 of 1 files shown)


Rocketknight1:
run-slow: lasr

tarekziade:
Nice finding. Shouldn't we fix `speecht5` similarly? Looking at this pattern, I think we have a few other models like this

Rocketknight1:
@tarekziade although the test name appears for a lot of models, it's actually mostly just to override and skip it! If you check the SpeechT5 tests, the test is skipped there.

I think there are relatively few models that support layerdrop and that have the test active, so we're probably okay to just fix this one and worry about other cases only if they show up in test runs.

tarekziade (approved):

PR #43955, opened 2026-02-12
[kernels] fix kernel versions 

MekkCyber:
# What does this PR do?

Fixes kernel versions for new builds with torch 2.10


ArthurZucker (approved):
ty

PR #43960, opened 2026-02-12
[CB] [Major] Asynchronous batching

remi-or:
# Summary

This PR adds the asynchronous batching feature to continuous batching (CB). Asynchronous batching, through the use of more VRAM and CUDA streams and events, greatly reduces the CPU overhead of preparing and updating batches by hiding it with GPU compute. In practice, the GPU can run uninterrupted for the workloads we tested the feature on (8B model, lots of generated tokens). 
The PR also:
- reduces the number of statuses for requests, thereby simplifying requests life cycles
- optimizes greatly the generation of read and write indices for full attention groups, which benefits both synchronous and asynchronous workflows
- caps the number of CUDA graphs instantiated at the same time
- modifies the way batches are padded: the user can now define the size of interval of padding for both Q and KV dimensions rather than the total number of intervals. The new way still allows the user to strike a balance between the amount of padding and the frequency of recording graphs, without penalizing users that choose a large `num_blocks` or `max_batch_tokens`

This makes for a beefy PR, but it was needed in order to reach the intended performances for async workflows.

## Performance
| Arguments                                          | Throughput on main | Throughput with PR | Delta    |
|----------------------------------------------------|--------------------|--------------------|----------|
| --samples 10                                       | 548.31             | 552.18             | +3.87    |
| --samples 20 --num-blocks 20                       | 131.16             | 135.98             | +4.82    |
| --samples 50                                       | 1416.88            | 1449.68            | +32.80   |
| --samples 100                                      | 2361.93            | 2429.70            | +67.77   |
| --samples 100 --attn flash_attention_2             | 2028.53            | 2080.67            | +52.14   |
| --samples 100 --attn sdpa                          | 834.30             | 846.19             | +11.89   |
| --samples 500                                      | 5396.69            | 5763.99            | +367.30  |
| --samples 500 --use-async                          | N/A                | 6331.60            | +5396.69*|
| --samples 500 --add-prefix --compile               | 6934.67            | 6802.36            | -132.31^ |
| --samples 50 --num-return-sequences 8 --do-sample  | 650.73             | 694.97             | +44.24   |
| --samples 100 --num-return-sequences 4 --do-sample | 1218.37            | 1291.48            | +73.11   |

\* compared with main, without async
^ this is the margins of error for a run with `compile`. I think it's better we address `compile` reproducibility and optimization in a downstream PR.

Another perf table, this time only about async:

| Arguments                 | Sync throughput | Async throughput | Delta              |
|---------------------------|-----------------|------------------|--------------------|
| --samples 10              | 547.35          | 556.75           | +1.72%             |
| --samples 50              | 1436.22         | 1562.68          | +8.81%             |
| --samples 100             | 2410.69         | 2672.14          | +10.85%            |
| --samples 250             | 4483.55         | 4636.34          | +3.41%             |
| --samples 500             | 5769.94         | 6302.8           | +9.23%             |
| --samples 32 --max-new-tokens 2048 --force-max-length | 1692.42| 1983.27 | + 17.19% |

## Tests
Tests pass and we added new test for async and read / write indices.

## Sanity checks
Generation looks ok.


mxchampagne:
Hi @remi-or,

Great work on the async batching. Quick question: is there any plan to support multimodal inputs (e.g. pixel values, audio features) in CB down the line? Just asking because if it's not on the roadmap, I'd be happy to take a stab at it myself.

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -122,6 +122,7 @@ def __init__(
         dtype: torch.dtype = torch.float16,
         tp_size: int | None = None,
         allow_block_sharing: bool = True,
+        use_async: bool = False,
```
ArthurZucker:
```suggestion
        use_async_batching: bool = False,
```
let's find a better name

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -242,10 +244,13 @@ def __init__(
 
         # We only use prefix sharing if the whole model has only full attention layers and block sharing is allowed
         self.use_prefix_sharing = allow_block_sharing and group_types == ["full_attention"]
-        self._block_manager = BlockManager(num_blocks, self.block_size)
-        self.blocks_to_complete: dict[str, int] = {}
+        self._block_manager = BlockManager(num_blocks, self.block_size, use_async=use_async)
```
ArthurZucker: should it be a different block manager?

src/transformers/generation/continuous_batching/cache.py
```diff
@@ -242,10 +244,13 @@ def __init__(
...
         self._total_prefix_length: int = 0  # a counter to measure the impact of prefix sharing, also used in tests
 
+    def finalize_block_uninitialization(self) -> None:
+        if self._block_manager is not None:
+            self._block_manager.finalize_block_uninitialization()
```
ArthurZucker: is this necessary?

ArthurZucker: to have it here

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -71,14 +71,17 @@ class BlockManager:
...
-    def __init__(self, num_blocks: int, block_size: int) -> None:
+    def __init__(self, num_blocks: int, block_size: int, use_async: bool) -> None:
         """Initializes the block manager with a given number of blocks (num_blocks) of size (block_size)."""
         self.num_blocks = num_blocks
         self.block_size = block_size
         self._uninit_block_ids = deque(range(num_blocks))
         self._init_block_ids: dict[int, None] = {}  # effectively act as an ordered set
         self._hash_to_id: dict[int, int] = {}
         self._id_to_block: dict[int, Block] = {}
+        # If async is used, there is a one batch period of waiting needed before uninitializing them
```
ArthurZucker: why?

```suggestion
        # If async is used, there is a one forward period of waiting needed before uninitializing them
```
?

src/transformers/generation/continuous_batching/cache_manager.py
```diff
@@ -214,7 +217,13 @@ def uninitialize_unshared_block(self, block_id: int) -> None:
         if block.ref_count > 1:
             raise RuntimeError(f"Block {block_id} has more than one reference: {block.ref_count = }")
         # Add the block to the uninitialized blocks queue
-        self._uninit_block_ids.append(block_id)
+        self._blocks_to_uninit.append(block_id)
+
+    def finalize_block_uninitialization(self) -> None:
```
ArthurZucker: appears twice no?

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -148,7 +147,17 @@ def __init__(
...
+        self.use_async = use_async
+        if self.use_async:
+            # Since in async there are 2 IO pairs, there are also 2 graph buffers: we divide the max_cached_graphs by 2
+            max_cached_graphs = ceil(max_cached_graphs / 2)
+            self.inputs_and_outputs = ContinuousBatchingAsyncIOs(
+                cache, config, model_device, model_dtype, max_cached_graphs
+            )
+        else:
+            self.inputs_and_outputs = ContinuousBatchingIOs(
+                cache, config, model_device, model_dtype, max_cached_graphs
```
ArthurZucker, line 159: that's nice

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -374,50 +385,60 @@ def _generation_step(self, model: nn.Module, logit_processor: LogitsProcessorLis
...
-            self._forward_process_and_sample(model, batch_data, logit_processor, do_sample)
-        torch.cuda.current_stream().wait_stream(stream)
-        # Catpure
-        graph = torch.cuda.CUDAGraph()
-        with torch.cuda.graph(graph, stream=stream):
-            self._forward_process_and_sample(model, batch_data, logit_processor, do_sample)
-        self._graphs[(padded_q, padded_read_index_size)] = graph
+        # Otherwise, we use create or replay the graph
+        else:
+            graph = self.inputs_and_outputs.graphs.get_graph(padded_q, padded_read_index_size)
```
ArthurZucker, line 409: that's nice

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -522,16 +547,23 @@ def __init__(
...
-            num_q_padding_intervals=num_q_padding_intervals,
-            num_kv_padding_intervals=num_kv_padding_intervals,
+            user_specified_param=bool(q_padding_interval_size or kv_padding_interval_size or max_cached_graphs),
             compile_config=getattr(generation_config, "compile_config", None),
         )
+        # If no behavior is given for the async API, it's turned on if cuda graphs are on and there is no attention mask
+        if use_async is not None:
+            self.use_async = use_async
+        else:
+            self.use_async = self.use_cuda_graph and not attn_mask_is_needed(self.model.config)
```
ArthurZucker: and a certain amount of requests no? (better defaults / criteria)

src/transformers/generation/continuous_batching/input_outputs.py
```diff
@@ -0,0 +1,612 @@
...
+        other.max_seqlen_k = dict(self.max_seqlen_k.items())
+        # Transfer static tensors
+        context = torch.cuda.stream(stream) if stream is not None else nullcontext()
+        with context:
+            other._bulk_input_tensor.copy_(self._bulk_input_tensor, non_blocking=non_blocking)
+            other.write_index_storage.copy_(self.write_index_storage, non_blocking=non_blocking)
+            other.read_index_storage.copy_(self.read_index_storage, non_blocking=non_blocking)
+            if self.attention_mask is not None and other.attention_mask is not None:
+                for layer_type in self.attention_mask.keys():
+                    other.attention_mask[layer_type].copy_(self.attention_mask[layer_type], non_blocking=non_blocking)
```
ArthurZucker: nice nice

remi-or:
Hey @mxchampagne ! It is on the roadmap, don't worry :slightly_smiling_face: but right now the main focus is performance /  documentation. Multimodality is coming afterwards.

src/transformers/generation/continuous_batching/scheduler.py
```diff
@@ -295,14 +297,14 @@ def __init__(self, cache: PagedAttentionCache, retain_cache_on_finish: bool = Fa
...
-    def schedule_batch(self, token_budget: int, cache_budget: int) -> list[RequestState] | None:
+    def schedule_batch(self, token_budget: int, cache_budget: int) -> list[FutureRequestState] | None:
         priority_states: list[RequestState] = []
         second_priority_states: list[RequestState] = []
 
         for state in self.active_requests.values():
             if state.status == RequestStatus.DECODING:
                 priority_states.append(state)
-            if state.status in [RequestStatus.SPLIT_PENDING_REMAINDER, RequestStatus.PREFILLING_SPLIT]:
+            elif state.status == RequestStatus.PREFILLING:
```
ArthurZucker, line 307: nice

src/transformers/generation/continuous_batching/requests.py
```diff
@@ -72,16 +72,14 @@ def get_device_and_memory_breakdown() -> tuple[torch.device, int, int, int]:
     return device, total_memory, reserved_memory, allocated_memory
 
 
-class RequestStatus(Enum):
+class RequestStatus(IntEnum):
```
ArthurZucker, line 75: nice as well

src/transformers/generation/continuous_batching/input_outputs.py
```diff
@@ -0,0 +1,612 @@
...
+
+    Asynchronous batching works by creating two pairs of host - device inputs and ouputs:
+
+                                    inputs
+                      ┌──────────┐ ────────► ┌────────────┐
+    IO pair object:   │ Host IOs │           │ Device IOs │       (for a CUDA sytem, Host = CPU and Device = GPU)
+                      └──────────┘ ◄──────── └────────────┘
+                                    outputs
+
+    Each pair is separate from the other. This means that each pairs has its own CUDA graphs set. But the CUDA streams
```
ArthurZucker: can't each pair use the same graph? or is it because we pad to closest graph?

src/transformers/generation/continuous_batching/input_outputs.py
```diff
@@ -0,0 +1,612 @@
...
+    ) -> None:
+        # IO pairs used to avoid race conditions
+        self.current_pair = 0
+        self.io_pairs = [HostDeviceIOPair(cache, config, device, model_dtype, max_graphs) for _ in range(2)]
+        # CUDA streams
+        self.h2d_stream = torch.cuda.Stream(device=device)
+        self.d2h_stream = torch.cuda.Stream(device=device)
+        self.compute_stream = torch.cuda.Stream(device=device)
+        # Set all unused compute streams to None
+        self.io_pairs[0].host_io.compute_stream = None
```
ArthurZucker, line 532: nice

src/transformers/generation/continuous_batching/input_outputs.py
```diff
@@ -0,0 +1,612 @@
...
+    def get_actual_lengths(self) -> tuple[int, int, int, list[int], list[int]]:
+        return self.io_pairs[self.current_pair].host_io.get_actual_lengths()
+
+    # The prepare_batch_tensor method also has to prepare the carry over ids
+    def prepare_batch_tensors(self, requests_in_batch: list[FutureRequestState]) -> None:
+        io_pair = self.io_pairs[self.current_pair]
+        io_pair.host_io.prepare_batch_tensors(requests_in_batch)
+        io_pair.host_io.carry_over_ids.copy_(self.infer_carry_over_ids())
+
+    def infer_carry_over_ids(self) -> torch.Tensor:
```
ArthurZucker, line 552: missing doc its not eveident for me waht this is etc

src/transformers/generation/continuous_batching/input_outputs.py
```diff
@@ -0,0 +1,612 @@
...
+
+    # The get_model_kwargs method is where the H2D transfer happens
+    def get_model_kwargs(self, padded_q_size: int = 0, padded_kv_cache_size: int = 0) -> dict[str, Any]:
+        io_pair = self.io_pairs[self.current_pair]
+        io_pair.transfer_inputs_h2d(self.h2d_stream)
+        self.h2d_stream.record_event(io_pair.h2d_over)
+        self.compute_stream.wait_event(io_pair.h2d_over)
+        return io_pair.device_io.get_model_kwargs(padded_q_size, padded_kv_cache_size)
+
+    def carry_over_tokens(self, input_ids: torch.Tensor) -> None:
```
ArthurZucker, line 579: same here, core logic / function need doc!

src/transformers/generation/continuous_batching/input_outputs.py
```diff
@@ -0,0 +1,612 @@
...
+class HostDeviceIOPair:
+    def __init__(
+        self,
+        cache: PagedAttentionCache,
+        config: PretrainedConfig,
+        device: torch.device,
+        model_dtype: torch.dtype,
+        max_graphs: int = 32,
+    ) -> None:
+        self.host_io = ContinuousBatchingIOs(cache, config, torch.device("cpu"), model_dtype, max_graphs)
```
ArthurZucker, line 456: pinned?

src/transformers/generation/continuous_batching/input_outputs.py
```diff
@@ -0,0 +1,612 @@
...
+        self.write_index_storage = torch.empty(
+            (num_groups, max_batch_tokens), dtype=torch.int32, device=self.device, pin_memory=pin_memory
+        )
+        self.read_index_storage = torch.empty(
+            (num_groups, num_pages + max_batch_tokens), dtype=torch.int32, device=self.device, pin_memory=pin_memory
+        )
+        # For read index, the +T is because there are -1 for seqlen_q when model uses a sliding window
+
+    def _transfer_inputs(
+        self, other: "ContinuousBatchingIOs", stream: torch.cuda.Stream | None = None, non_blocking: bool = False
```
ArthurZucker:
```suggestion
        self, target_io: "ContinuousBatchingIOs", stream: torch.cuda.Stream | None = None, non_blocking: bool = False
```

ArthurZucker: nit

src/transformers/generation/continuous_batching/input_outputs.py
```diff
@@ -0,0 +1,612 @@
...
+        other.actual_key_length = self.actual_key_length
+        other.actual_batch_size = self.actual_batch_size
+        other.actual_read_sizes = self.actual_read_sizes[:]
+        other.actual_write_sizes = self.actual_write_sizes[:]
+        # Transfer scalar attributes
+        other.total_seqlen_q = self.total_seqlen_q
+        other.max_seqlen_q = self.max_seqlen_q
+        other.max_seqlen_k = dict(self.max_seqlen_k.items())
+        # Transfer static tensors
+        context = torch.cuda.stream(stream) if stream is not None else nullcontext()
```
ArthurZucker: there's always gonna be a stream no?

src/transformers/generation/continuous_batching/input_outputs.py
```diff
@@ -0,0 +1,612 @@
...
+        # For other kwargs, we need a list of tensors with as many tensors as there are groups
+        self.write_index_storage = torch.empty(
+            (num_groups, max_batch_tokens), dtype=torch.int32, device=self.device, pin_memory=pin_memory
+        )
+        self.read_index_storage = torch.empty(
+            (num_groups, num_pages + max_batch_tokens), dtype=torch.int32, device=self.device, pin_memory=pin_memory
+        )
+        # For read index, the +T is because there are -1 for seqlen_q when model uses a sliding window
+
+    def _transfer_inputs(
```
ArthurZucker, line 193: Wow actually nice, I would write somewhere that we do the bulk tensor specifically to have somewhat of a single copy_ vs many for each tensor!

src/transformers/generation/continuous_batching/input_outputs.py
```diff
@@ -0,0 +1,612 @@
...
+        """Allocates static tensors for generation inputs and outputs. This is called only once at init time, to avoid
+        repeated allocations and enable CUDA graphs. All tensors are allocated with maximum possible sizes.
+        The allocated tensors are:
+
+        - `input_ids` and `position_ids`: Query token information
+        - `cumulative_seqlens_q` and `cumulative_seqlens_k`: Sequence length tracking for FlashAttention-style batching
+        - `attention_mask`: Optional attention masks (only for eager/SDPA implementations)
+        - `write_index` and `read_index` storage: Cache indexing tensors for each attention group
+        - `output_ids`: Storage for generated token IDs
+        """
```
ArthurZucker, line 133: `_bulk_input_tensor` is also allocated and we just create pointers (`input_ids`, `position_ids` etc).

ArthurZucker (approved):
Huge work! 🫡 
I am weirded out by the input_and_output, but it most probably makes sense! 

Great doc, great work on leveraging cuda steram, io sync etc its super super nice!

ArthurZucker:
@mxchampagne I think its also very very core, we probably want to do that specifically internally, but you can take a stab at beam-search if you have time / want to contribute!!! 🤗

mxchampagne:
Got it. I'll implement multimodal support on my side since that's what I need for my current use case. If I come up with something clean and reliable, i'll let you know. Thanks.

src/transformers/generation/continuous_batching/cache.py
remi-or: done!

src/transformers/generation/continuous_batching/cache.py
remi-or: No, the BlockManager stays unique, as the cache. Its job does not change: ensure coherency and share cache across all requests.

src/transformers/generation/continuous_batching/cache.py
remi-or: Good point, it seems it's actually useless. It would have been useful for triple-buffering or more, but here since we update batch N while batch N+1 is in flight, there is no need to delay the un-initialization: the block will be overwritten in batch N+2 at the earliest.

src/transformers/generation/continuous_batching/cache_manager.py
remi-or: cf. above

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: thanks!

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: Thanks

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: Actually, it seems async provices a benefit in all cases. I just removed the cases where attn_mask is needed because that adds a lot of D2H transfers and the case where CUDA graphs are off because the only reason to do so (once no attn mask is used) is to save memory, which async does not help with. I will add a comment and we can workshop this along with other defaults later.

src/transformers/generation/continuous_batching/input_outputs.py
remi-or: thx thx

src/transformers/generation/continuous_batching/input_outputs.py
remi-or: No, because CUDA graphs track the address of the input and output tensors. Hence, if we want to have a single set of graph, we would need a third GPU-side buffer to transfer the data of the batch we are about to run to. Memory-wise, idk which would be best, we would need to compute the size of one vs the other, but speed-wise it will always be a loss. So i went with this solution, I am adding a comment on this.

src/transformers/generation/continuous_batching/input_outputs.py
remi-or: added docstring

src/transformers/generation/continuous_batching/input_outputs.py
remi-or: done

src/transformers/generation/continuous_batching/input_outputs.py
remi-or: Automatically! Added a comment

src/transformers/generation/continuous_batching/input_outputs.py
remi-or: Actually, I think it's common to have "other" as an argument name when it's another instance of the same class. Like in `__add__` or `__mul__`, so I am thinking we can keep it that way. Otherwise I will change it in another PR, no probs.

src/transformers/generation/continuous_batching/input_outputs.py
remi-or: You are right, this is a legacy from when we had a third device-side buffer. Removing this.

src/transformers/generation/continuous_batching/input_outputs.py
remi-or: Done, added more context at tensor creation and a comment in the transfer function

src/transformers/generation/continuous_batching/input_outputs.py
remi-or: Done

PR #43961, opened 2026-02-12
Replace mutable default arguments with None

DimiChatzipavlis:
# What does this PR do?

Fixes a common Python pitfall regarding **mutable default arguments**. 

In Python, default arguments are evaluated only once at function definition time. If a mutable object (like a `list`) is used as a default, that same list instance is shared across all calls to the function. This can lead to insidious bugs where the list grows unexpectedly between function calls or instances.

## Modifications
Refactored the following method signatures to use `None` as the default value and initialize the list inside the function:

- **`src/transformers/models/idefics/modeling_idefics.py`**:
  - `freeze_model(..., module_exceptions=[])` -> `module_exceptions=None`
  - `freeze_text_layers(..., module_exceptions=[])` -> `module_exceptions=None`
  - `freeze_vision_layers(..., module_exceptions=[])` -> `module_exceptions=None`

- **`src/transformers/debug_utils.py`**:
  - `DebugUnderflowOverflow.__init__(..., trace_batch_nums=[])` -> `trace_batch_nums=None`

## Verification
- Verified that the logic inside the functions correctly initializes the list if the argument is `None` (`if arg is None: arg = []`).
- Verified that existing calls passing explicit lists are unaffected.

## Who can review?
@sgugger 
@amyeroberts
@ArthurZucker

## Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/src/transformers/debug_utils.py
+++ b/src/transformers/debug_utils.py
@@ -142,7 +142,9 @@ class DebugUnderflowOverflow:
             Whether to abort after a certain batch number has finished
     """
 
-    def __init__(self, model, max_frames_to_save=21, trace_batch_nums=[], abort_after_batch_num=None):
+    def __init__(self, model, max_frames_to_save=21, trace_batch_nums=None, abort_after_batch_num=None):
+        if trace_batch_nums is None:
+            trace_batch_nums = []
         self.model = model
         self.trace_batch_nums = trace_batch_nums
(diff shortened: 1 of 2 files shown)


Rocketknight1:
These methods are all very short and none of them mutate the input argument, though! I agree that it's an antipattern, which is why your code agent picked it up when asked to "improve the codebase" or however you prompted it, but it seems too insignificant to fix right now

PR #43964, opened 2026-02-12
Fix ZeRO-3 conversion-mapped checkpoint loading

tohtana:
This PR fixes a ZeRO-3 checkpoint loading failure in Transformers’ conversion-mapped loading path. In affected cases, many parameters are reported as missing and are actually not restored from checkpoint (they get reinitialized).

`transformers/integrations/deepspeed.py` used a `WeightConverter.convert()` return assumption that is no longer stable across versions (legacy tuple vs current dict), which broke correctness/compatibility in the ZeRO-3 conversion helper path.

Chages:
  - Made ZeRO-3 conversion loading robust to both WeightConverter.convert() return shapes:
      - legacy: (converted_tensors, conversion_errors)
      - current: converted_tensors
  - Added a regression test:

Links: huggingface/transformers#43940, deepspeedai/DeepSpeed#7848

### DIFF
--- a/src/transformers/integrations/deepspeed.py
+++ b/src/transformers/integrations/deepspeed.py
@@ -383,11 +383,15 @@ def _apply_weight_conversions_to_state_dict(model, state_dict, weight_mapping):
             # Only WeightConverter needs convert(); WeightRenaming is just a simple rename
             if not isinstance(mapping, WeightConverter):
                 continue
-            realized_value, _ = mapping.convert(
+            converted = mapping.convert(
                 renamed_key,
                 model=model,
                 config=model.config,
             )
+            # Be robust to both conversion APIs:
+            # - legacy: (converted_tensors, conversion_errors)
+            # - current: converted_tensors
+            realized_value = converted[0] if isinstance(converted, tuple) else converted
             for target_name, param in realized_value.items():
                 param = param[0] if isinstance(param, list) else param
                 new_state_dict[target_name] = param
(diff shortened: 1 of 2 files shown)


Rocketknight1:
cc @sunmarc

src/transformers/integrations/deepspeed.py
```diff
@@ -383,11 +383,15 @@ def _apply_weight_conversions_to_state_dict(model, state_dict, weight_mapping):
...
-            realized_value, _ = mapping.convert(
+            converted = mapping.convert(
                 renamed_key,
                 model=model,
                 config=model.config,
             )
+            # Be robust to both conversion APIs:
+            # - legacy: (converted_tensors, conversion_errors)
+            # - current: converted_tensors
+            realized_value = converted[0] if isinstance(converted, tuple) else converted
```
SunMarc, line 394: Do we still care about the legacy API @ArthurZucker ? It should always be a dict no ?

SunMarc (approved):
Thanks a lot @tohtana ! Just a quick question for @ArthurZucker

kashif:
thanks @tohtana I was thinking perhaps something like: https://github.com/huggingface/transformers/pull/43926 the issue was that my initial test was wrong and did not catch this issue which i have fixed in this PR

tohtana:
Thank you for the update! It looks like you have a thorough solution in place, so I’ll go ahead and close this PR. I really appreciate how quickly you fixed the DeepSpeed issue.

PR #43968, opened 2026-02-13
Enhance JSON schema generation to support instance, static, and class methods

qgallouedec:
# What does this PR do?

Enhances `get_json_schema()` and `render_jinja_template()` to support instance methods, class methods, and static methods, not just plain functions.

Previously, `get_json_schema()` only worked with standalone functions. Passing a method (bound or unbound) would either fail or incorrectly include `self`/`cls` as a required parameter in the generated schema. Similarly, `render_jinja_template()` would reject bound methods with a `ValueError` because `inspect.isfunction()` returns `False` for them.

This is needed for stateful tool use, where tools are naturally defined as methods on a class instance. For example, an agent may need tools that share state (e.g., a database connection, a session, or accumulated context), and the idiomatic way to do this in Python is to group them as methods on an object:

```python
class IncrementEnv:
    def __init__(self):
        self._counter = 0

    def increment(self, a: int) -> int:
        """
        Increment the internal counter by the input value.

        Args:
            a: The amount to increment the counter by

        Returns:
            The new value of the counter after incrementing
        """
        self._counter += a
        return self._counter

my_env = IncrementEnv()
schema = get_json_schema(my_env.increment)  # Now works correctly
```

```
{'type': 'function', 'function': {'name': 'increment', 'description': 'Increment the internal counter by the input value.', 'parameters': {'type': 'object', 'properties': {'a': {'type': 'integer', 'description': 'The amount to increment the counter by'}}, 'required': ['a']}, 'return': {'type': 'integer', 'description': 'The new value of the counter after incrementing'}}}
```
## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/transformers/utils/chat_template_utils.py
+++ b/src/transformers/utils/chat_template_utils.py
@@ -181,8 +181,21 @@ def _parse_type_hint(hint: str) -> dict:
 def _convert_type_hints_to_json_schema(func: Callable) -> dict:
     type_hints = get_type_hints(func)
     signature = inspect.signature(func)
+    # For methods, we need to ignore the first "self" or "cls" parameter. However, since unbound methods are just
+    # functions, we need to check the signature to see if it looks like a method rather than only relying on 
+    # inspect.ismethod(), which returns False for unbound methods.
+    qualname = getattr(func, "__qualname__", "")
+    qualname_parts = qualname.split(".")
+    has_unbound_method_signature = isfunction(func) and len(qualname_parts) >= 2 and qualname_parts[-2] != "<locals>"
+    first_param_name = next(iter(signature.parameters), None)
+    implicit_arg_name = None
+    if first_param_name in {"self", "cls"} and (inspect.ismethod(func) or has_unbound_method_signature):
+        implicit_arg_name = first_param_name
+
     required = []
     for param_name, param in signature.parameters.items():
+        if param_name == implicit_arg_name:
+            continue
         if param.annotation == inspect.Parameter.empty:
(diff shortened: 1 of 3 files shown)


src/transformers/utils/chat_template_utils.py
```diff
@@ -181,15 +181,30 @@ def _parse_type_hint(hint: str) -> dict:
...
+    # For methods, we need to ignore the first "self" or "cls" parameter. However, since unbound methods are just
+    # functions, we need to check the signature to see if it looks like a method rather than only relying on 
+    # inspect.ismethod(), which returns False for unbound methods.
+    qualname = getattr(func, "__qualname__", "")
+    qualname_parts = qualname.split(".")
+    has_unbound_method_signature = isfunction(func) and len(qualname_parts) >= 2 and qualname_parts[-2] != "<locals>"
+    first_param_name = next(iter(signature.parameters), None)
+    implicit_arg_name = None
+    if first_param_name in {"self", "cls"} and (inspect.ismethod(func) or has_unbound_method_signature):
+        implicit_arg_name = first_param_name
```
Rocketknight1: I'm a little scared of the `qualname` access - it's a fairly reliable heuristic to detect when something is an unbound method and not a function, but it seems a bit too complex and I worry something will break it later.

How about just "if the first param name is `self` or `cls` and it doesn't have a type hint"? That makes the check a lot simpler, avoids `qualname` hackery, and will only fail in very pathological cases where the user is doing something silly, lol

qgallouedec: fair enough, I'm updating this

Rocketknight1 (approved):

PR #43969, opened 2026-02-13
Fix for 2D weight error in embedding layer with ZeRO3

jp1924:
# What does this PR do?

When training a model with ZeRO3, a 2D weight error occurs in the embedding layer. This happens because, although the model parameters are sharded across GPUs, the DeepSpeed engine does not group them, so it fails to collect the parameters. https://github.com/huggingface/transformers/pull/43711

A recent PR added a branch in self.accelerator.prepare to handle the scheduler separately for DeepSpeed. In some conditions this causes prepare to be bypassed, preventing DeepSpeed training from running correctly, so the trainer was modified.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc

### DIFF
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -1563,16 +1563,13 @@ class Trainer:
 
         # prepare using `accelerator` prepare
         if use_accelerator_prepare:
-            self.model.train()
-            if self.is_deepspeed_enabled:
-                from accelerate.utils import DummyScheduler
+            from accelerate.utils import DummyScheduler
 
-                if isinstance(self.lr_scheduler, DummyScheduler):
-                    model, self.optimizer, self.lr_scheduler = self.accelerator.prepare(
-                        self.model, self.optimizer, self.lr_scheduler
-                    )
-                else:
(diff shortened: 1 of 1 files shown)


src/transformers/trainer.py
```diff
@@ -1563,16 +1563,13 @@ def _inner_training_loop(
...
-                    model, self.optimizer, self.lr_scheduler = self.accelerator.prepare(
-                        self.model, self.optimizer, self.lr_scheduler
-                    )
-                else:
-                    model, self.optimizer = self.accelerator.prepare(self.model, self.optimizer)
+            self.model.train()
+            if self.is_deepspeed_enabled and isinstance(self.lr_scheduler, DummyScheduler):
+                model, self.optimizer, self.lr_scheduler = self.accelerator.prepare(
+                    self.model, self.optimizer, self.lr_scheduler
+                )
```
SunMarc, line 1572: the logic is the same no ? happy to still have this pr to simplify it but you will need to rename the PR

SunMarc (commented):
Left a comment

winglian:
The fix was already merged into `main` with https://github.com/huggingface/transformers/pull/43780

PR #43970, opened 2026-02-13
Add Four Over Six quantization integration

jackcook:
# What does this PR do?

This PR adds support for quantization with [Four Over Six (4/6)](https://github.com/mit-han-lab/fouroversix). Our library currently focuses on quantizing linear layers to NVFP4, including weight, activation, and gradient tensors. After quantization, these layers can be trained as well, as described in our docs.

Before running our code, you should install `fouroversix`:

```python
pip install fouroversix --no-build-isolation
```

And once installed, our library has efficient kernels that can be run on NVIDIA Blackwell GPUs, but will also fall back to a reference implementation that supports older GPUs and CPUs as well.

@SunMarc @MekkCyber


docs/source/en/quantization/fouroversix.md
```diff
@@ -0,0 +1,64 @@
...
+
+```python
+from transformers import AutoModelForCausalLM, FourOverSixConfig
+
+model = AutoModelForCausalLM.from_pretrained(
+    "Qwen/Qwen3-8B",
+    device_map="auto",
+    quantization_config=FourOverSixConfig(
+        keep_master_weights=True,
+        exclude_layers=[
```
ArthurZucker: we use `modules_to_not_convert` I don't mind changing the name but we need standardization

sayakpaul:
@jackcook thanks for this PR. I am from the Diffusers library. Would you be interested in contributing this to library as well? I would be keen on helping you as much as I can.

src/transformers/integrations/fouroversix.py
```diff
@@ -0,0 +1,85 @@
...
+
+        if module.weight.device.type != "meta":
+            module.apply_ptq()
+
+        return {
+            f"{layer_name}.quantized_weight_amax": module.quantized_weight_amax,
+            f"{layer_name}.quantized_weight_scale_factors": module.quantized_weight_scale_factors,
+            f"{layer_name}.quantized_weight_values": module.quantized_weight_values,
+            f"{layer_name}.quantized_weight_metadata": module.quantized_weight_metadata,
+        }
```
ArthurZucker, line 57: we gotta make sure they are deleted from the module afterwards right?

src/transformers/integrations/fouroversix.py
```diff
@@ -0,0 +1,85 @@
...
+        **kwargs,
+    ) -> dict[str, torch.Tensor]:
+        """
+        we need to store some parameters to create the quantized weight. For example, fouroversix requires 6 values that are stored in the checkpoint to recover the quantized weight. So we store them in a dict that it stored in hf_quantizer for now as we can't save it in the op since we create an op per tensor.
+        """
+        value = list(input_dict.values())[0]
+        value = value[0]
+        module, _ = get_module_from_name(model, full_layer_name)
+        module = module.to_empty(device=value.device)
+        module.weight = torch.nn.Parameter(value, requires_grad=False)
```
ArthurZucker: in general we don't change the module here. We should have 1 func that changes the module first for the weight converter to filter the keys that are unexpected etc

src/transformers/quantizers/quantizer_fouroversix.py
```diff
@@ -0,0 +1,108 @@
...
+        **kwargs,
+    ):
+        from fouroversix import quantize_model
+
+        from ..integrations.fouroversix import adapt_fouroversix_config
+
+        quantize_model(
+            model,
+            adapt_fouroversix_config(self.quantization_config),
+        )
```
ArthurZucker, line 82: right so this is where we should do any module replacement etc in general!

src/transformers/utils/quantization_config.py
```diff
@@ -1902,3 +1903,87 @@ def to_dict(self) -> dict[str, Any]:
...
+            Four Over Six, `"static_6"` for default NVFP4 quantization, or `"static_4"` to scale
+            all blocks to a maximum value of 4.
+        weight_scale_2d (`bool`, *optional*, defaults to `False`):
+            Whether to compute scale factors on weight tensors in 2D blocks. This should be done
+            during training.
+        weight_scale_rule (`str`, *optional*, defaults to `None`):
+            Scaling rule to use when selecting a scale for blocks in weight tensors. If not
+            provided, `scale_rule` is used.
+        exclude_layers (`list[str]`, *optional*, defaults to `["lm_head"]`):
+            The list of layers to exclude from quantization. By default, the `lm_head` is excluded.
```
ArthurZucker: let's standardize whatever we can!

ArthurZucker (commented):
Super nice to have!!!!

src/transformers/integrations/fouroversix.py
```diff
@@ -0,0 +1,85 @@
...
+
+        layer_name = full_layer_name.rsplit(".", 1)[0]
+        missing_keys.discard(f"{layer_name}.weight")
+
+        if module.weight.device.type != "meta":
```
SunMarc: In which case the module weight device is still on `meta` ?

src/transformers/integrations/fouroversix.py
```diff
@@ -0,0 +1,85 @@
...
+        model: torch.nn.Module | None = None,
+        full_layer_name: str | None = None,
+        missing_keys: list[str] | None = None,
+        **kwargs,
+    ) -> dict[str, torch.Tensor]:
+        layer_name = full_layer_name.rsplit(".", 1)[0]
+        layer = model.get_submodule(layer_name)
+
+        # Delete the randomly initialized high-precision weight before loading the quantized weight
+        del layer.weight
```
SunMarc: the model is initialized on `meta` device, so there shouldn't need to do that no ? 
This happens because _process_model_before_weight_loading is executed in those context manager: 
```python
with ContextManagers(model_init_context):
            # Let's make sure we don't run the init function of buffer modules
            model = cls(config, *model_args, **model_kwargs)

            if hf_quantizer is not None:  # replace module with quantized modules (does not touch weights)
                hf_quantizer.preprocess_model(
                    model=model,
                    dtype=dtype,
                    device_map=device_map,
                    checkpoint_files=checkpoint_files,
                    use_kernels=use_kernels,
                )
```

src/transformers/utils/quantization_config.py
```diff
@@ -1902,3 +1903,87 @@ def to_dict(self) -> dict[str, Any]:
...
+        self.dtype = dtype
+        self.gradient_scale_rule = gradient_scale_rule
+        self.keep_master_weights = keep_master_weights
+        self.matmul_backend = matmul_backend
+        self.quantize_backend = quantize_backend
+        self.output_dtype = output_dtype
+        self.scale_rule = scale_rule
+        self.weight_scale_2d = weight_scale_2d
+        self.weight_scale_rule = weight_scale_rule
+        self.exclude_layers = exclude_layers
```
SunMarc: as @ArthurZucker suggested, it would be nice to use `modules_to_not_convert` instead

src/transformers/quantizers/quantizer_fouroversix.py
SunMarc, line 82: Up to you but we can put the code on how the layers are replaced here like most other quants methods.

SunMarc (commented):
Thanks for this integration ! Just a few nits

jackcook:
Thanks for all the quick feedback @ArthurZucker @SunMarc! I think I understand the HF quantizer pattern a little better now, and I've pushed some changes that reflect this. Most importantly, the quantized parameters are created in our `quantize_model` function now and the module is no longer updated in `FourOverSixQuantize`.

I still feel like I'm missing something though — where is the correct place to remove the high-precision `weight` parameter once the quantized parameters have been set? If I don't do it in `_process_model_after_weight_loading` (where I've put it for now), I get an error during `save_pretrained` because the `weight` tensor remains on the `meta` device and can't be saved to disk. But as the code is currently written, I'm failing the `FourOverSixMSETest::test_quantized_model_multi_accelerator` test, because the `weight` tensor is not removed before the model is split up between devices:

```
ValueError: weight is on the meta device, we need a `value` to put in on 0
```

Thanks again for the help, and yes @sayakpaul I'd be happy to add this to diffusers as well, I have a few things I need to take care of in the next week or two but I'll get back to you about that soon!

sayakpaul:
> Thanks again for the help, and yes @sayakpaul I'd be happy to add this to diffusers as well, I have a few things I need to take care of in the next week or two but I'll get back to you about that soon!

Sure, I am looking forward to seeing that happen!

src/transformers/integrations/fouroversix.py
```diff
@@ -0,0 +1,95 @@
...
+        self,
+        input_dict: dict[str, torch.Tensor],
+        model: torch.nn.Module | None = None,
+        full_layer_name: str | None = None,
+        missing_keys: list[str] | None = None,
+        **kwargs,
+    ) -> dict[str, torch.Tensor]:
+        module_name = full_layer_name.rsplit(".", 1)[0]
+        missing_keys.discard(f"{module_name}.weight")
+        return {full_layer_name: input_dict["quantized_weight_values"][0].data}
```
MekkCyber: not sure why we are still discarding from missing keys while returning the tensor, we shouldn't be expecting the `weight` attribute in the `FourOverSix` quantized linear, we should only have the quantization states I think that's what causing the issue with the weight on meta device

src/transformers/quantizers/quantizer_fouroversix.py
```diff
@@ -0,0 +1,114 @@
...
+
+        if self.pre_quantized:
+            return [
+                WeightConverter(
+                    source_patterns=["quantized_weight_values"],
+                    target_patterns="quantized_weight_values",
+                    operations=[FourOverSixDeserialize(self)],
+                )
+            ]
+
```
MekkCyber: we can add dequantization here with a flag like `dequantize` that can be set to true or false in the config, you can check `mxfp4` quantization for example for reference

src/transformers/utils/quantization_config.py
```diff
@@ -1902,3 +1903,87 @@ def to_dict(self) -> dict[str, Any]:
...
+        gradient_scale_rule: str | None = None,
+        keep_master_weights: bool = False,
+        matmul_backend: str | None = None,
+        output_dtype: str | None = "bfloat16",
+        quantize_backend: str | None = None,
+        scale_rule: str = "mse",
+        weight_scale_2d: bool = False,
+        weight_scale_rule: str | None = None,
+        module_config_overrides: dict[str, dict[str, Any]] | None = None,
+        modules_to_not_convert: list[str] | None = ["lm_head"],
```
MekkCyber, line 1978: maybe we can add a `post_init` to check that the params passed by the user are correct ?

tests/quantization/fouroversix_integration/test_fouroversix.py
```diff
@@ -0,0 +1,195 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
+#
```
MekkCyber:
```suggestion
# Copyright 2026 The HuggingFace Team. All rights reserved.
#
```

src/transformers/integrations/fouroversix.py
```diff
@@ -0,0 +1,95 @@
...
+        since we create an op per tensor.
+        """
+
+        target_keys, value = list(input_dict.items())[0]
+        value = value[0]
+        module, _ = get_module_from_name(model, full_layer_name)
+
+        module_name = full_layer_name.rsplit(".", 1)[0]
+        missing_keys.discard(target_keys)
+        quantized_params = module.get_quantized_parameters(value)
```
MekkCyber: could deleting the `weight` be done here instead after getting the quantized_params ?

MekkCyber (commented):
Thanks for the contribution @jackcook ! It looks very clean, I just have some few comments

SunMarc:
> I still feel like I'm missing something though — where is the correct place to remove the high-precision weight parameter once the quantized parameters have been set? If I don't do it in _process_model_after_weight_loading (where I've put it for now), I get an error during save_pretrained because the weight tensor remains on the meta device and can't be saved to disk. 

Usually, In the quantized module, there is a param that is still named `weight`. So we don't really need to remove the high-precision weight. Otherwise, you could remove it in `FourOverSixQuantize` as @MekkCyber suggested https://github.com/huggingface/transformers/pull/43970/changes#r2811773058 or as you did in `_process_model_after_weight_loading`.

src/transformers/integrations/fouroversix.py
jackcook: Sorry I'm not sure I understand what you mean by this, the quantized parameters need to be deleted from the module?

src/transformers/integrations/fouroversix.py
jackcook: This didn't end up being what was causing the issue with the meta device, but you're right that this was essentially a no-op. I've removed it

src/transformers/quantizers/quantizer_fouroversix.py
jackcook: I'm having some trouble implementing it as the FourOverSixLinear layer doesn't support being run with a high precision weight, so the modules would need to be replaced. Do you think this flag is necessary? Our implementation will automatically perform dequantization and do high-precision matmuls if it detects that the user's GPU doesn't support NVFP4 operations.

src/transformers/utils/quantization_config.py
jackcook: This is handled by our framework's code when the config is converted in `adapt_fouroversix_config`. I can still do some checks here if you'd like but I think it may be better to keep them in one place

jackcook:
All tests are passing with the most recent change, thanks again everyone for the feedback!

SunMarc:
@bot /style

SunMarc:
Can you fix the conflicts + run `fix-repo` ?

src/transformers/utils/quantization_config.py
MekkCyber: nw sounds good if the checks are handled by the framework

src/transformers/quantizers/quantizer_fouroversix.py
MekkCyber: nice if that's done automatically then no need to add the flag for now, but would be awesome to add this to the docs

src/transformers/integrations/fouroversix.py
ArthurZucker: the new changes properly delete the attr 😉

jackcook:
Done @SunMarc, let me know if any other changes are needed, thanks again for the help with everything!

src/transformers/integrations/fouroversix.py
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        quantized_params = module.get_quantized_parameters(**high_precision_parameters)
+
+        # Delete the high-precision parameters from the module now that we've used them to create
+        # the quantized parameters
+        for parameter_name in module.high_precision_parameter_names:
+            delattr(module, parameter_name)
+
+        # Remove these keys from the missing_keys list since we've deleted them from hte model
```
MekkCyber:
```suggestion
        # Remove these keys from the missing_keys list since we've deleted them from the model
```

src/transformers/integrations/fouroversix.py
```diff
@@ -0,0 +1,70 @@
...
+
+        module, _ = get_module_from_name(model, full_layer_name)
+        module_name = full_layer_name.rsplit(".", 1)[0]
+        high_precision_parameters = {
+            key.replace(f"{module_name}.", "", 1): value[0] for key, value in input_dict.items()
+        }
+
+        quantized_params = module.get_quantized_parameters(**high_precision_parameters)
+
+        # Delete the high-precision parameters from the module now that we've used them to create
```
MekkCyber:
```suggestion
        # Delete the high-precision parameters from the module after we used them to create
```

src/transformers/quantizers/quantizer_fouroversix.py
```diff
@@ -0,0 +1,114 @@
...
+            adapt_fouroversix_config(self.quantization_config),
+        )
+
+        # If the model has already been quantized, we need to delete the weight tensor here so that
+        # it's not expected when parameters are loaded from the checkpoint.
+        if self.pre_quantized:
+            for _, module in model.named_modules():
+                if QuantizedModule.is_quantized_module_type(type(module)):
+                    for parameter_name in module.high_precision_parameter_names:
+                        delattr(module, parameter_name)
```
MekkCyber: I think we should add some logic about `keep_master_weights`

src/transformers/quantizers/quantizer_fouroversix.py
```diff
@@ -0,0 +1,114 @@
...
+        # from ..integrations.fouroversix import FourOverSixDeserialize
+
+        # if self.pre_quantized:
+        #     return [
+        #         WeightConverter(
+        #             source_patterns=["quantized_weight_values"],
+        #             target_patterns="quantized_weight_values",
+        #             operations=[FourOverSixDeserialize(self)],
+        #         )
+        #     ]
```
MekkCyber: why is this commented ?

MekkCyber (commented):
Thanks @jackcook for iterating quickly on this! I just have some few nits about the master weights, and the deserialisation part

src/transformers/quantizers/quantizer_fouroversix.py
jackcook: My apologies, I meant to delete this since we're not adding the `dequantize` flag

src/transformers/quantizers/quantizer_fouroversix.py
jackcook: You're right, thank you for pointing this out. This is partly handled by our framework but it needs to be handled here as well, I've fixed it and added a unit test for this just to check it in the future.

jackcook:
All fixed, thanks again for the feedback @MekkCyber

SunMarc:
@bot /style

src/transformers/quantizers/quantizer_fouroversix.py
```diff
@@ -78,6 +76,14 @@ def _process_model_before_weight_loading(
             adapt_fouroversix_config(self.quantization_config),
         )
 
+        # If the model has already been quantized, we need to delete the weight tensor here so that
+        # it's not expected when parameters are loaded from the checkpoint.
+        if self.pre_quantized and not self.quantization_config.keep_master_weights:
+            for _, module in model.named_modules():
+                if QuantizedModule.is_quantized_module_type(type(module)):
+                    for parameter_name in module.high_precision_parameter_names:
+                        delattr(module, parameter_name)
```
SunMarc, line 90: ok I see, this way we don't need the deserialize

SunMarc (approved):
This looks nice, thanks a lot ! If you can just fix the CI. There is the following issue: 

```
  File "/root/transformers/utils/check_docstrings.py", line 2033, in <module>
    check_docstrings(overwrite=args.fix_and_overwrite, check_all=args.check_all)
  File "/root/transformers/utils/check_docstrings.py", line 2022, in check_docstrings
    raise ValueError(error_message)
ValueError: There was at least one problem when checking docstrings of public objects.
The following objects docstrings do not match their signature. Run `make fix-repo` to fix this. In some cases, this error may be raised incorrectly by the docstring checker. If you think this is the case, you can manually check the docstrings and then add the object name to `OBJECTS_TO_IGNORE` in `utils/check_docstrings.py`.
- FourOverSixConfig

Exited with code exit status 1
```

jackcook:
Just fixed the docstring! Thanks again for all the help

MekkCyber (approved):
Let's goooo 🔥 ! Thanks for this huge contribution @jackcook

sayakpaul:
@jackcook thank you for landing this PR. Just ping me whenever you think we're ready to land it in Diffusers :)

PR #43972, opened 2026-02-13
:rotating_light: Unify 3D position ids

zucchini-nlp:
# What does this PR do?

Following Ernie, we build 3d positions based on `mm_token_type_ids` and the models will return them by default from `processor`. 

We have a unified `get_vision_position` in the qwen2-vl model file, all other models just copy it from there. The utility build vision ids as the name suggests, and the models are free to manipulate on top as they wish. In most cases, the only thing that changes is the presence of new modalities or kwargs


zucchini-nlp:
run-slow: ernie4_5_vl_moe, qwen2_5_vl, qwen2_vl

zucchini-nlp:
run-slow: glm46v, glm4v, glm4v_moe, glm_ocr, paddleocr_vl, qwen2_5_vl, qwen2_vl, qwen3_5, qwen3_5_moe, qwen3_vl, qwen3_vl_moe

zucchini-nlp:
run-slow: ernie4_5_vl_moe, glm46v, glm4v, glm4v_moe, glm_image, glm_ocr, paddleocr_vl, qwen2_5_vl, qwen2_vl, qwen3_5, qwen3_5_moe, qwen3_vl, qwen3_vl_moe

zucchini-nlp:
run-slow: ernie4_5_vl_moe, glm46v, glm4v, glm4v_moe, glm_image, glm_ocr, paddleocr_vl, qwen2_5_vl, qwen2_vl, qwen3_5, qwen3_5_moe, qwen3_vl, qwen3_vl_moe

zucchini-nlp:
run-slow: ernie4_5_vl_moe, glm46v, glm4v, glm4v_moe, glm_image, glm_ocr, paddleocr_vl, qwen2_5_vl, qwen2_vl, qwen3_5, qwen3_5_moe, qwen3_vl, qwen3_vl_moe

zucchini-nlp:
run-slow: ernie4_5_vl_moe, glm46v, glm4v, glm4v_moe, glm_image, glm_ocr, paddleocr_vl, qwen2_5_vl, qwen2_vl, qwen3_5, qwen3_5_moe, qwen3_vl, qwen3_vl_moe

zucchini-nlp:
The slow CI takes really long now, it took the whole work-day for 13 models 🙃

zucchini-nlp:
run-slow: qwen3_vl, glm_image, qwen2_5_vl, glm4v, ernie_4_5vl_moe

zucchini-nlp:
run-slow: glm4v

zucchini-nlp:
retriggering later, it took a bad commit to run CI!

vasqu:
No worries, should I still take a look?

zucchini-nlp:
Yes please, the rest is to adjust processor tests and re-run slow CI

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
```diff
@@ -990,44 +991,69 @@ def get_input_embeddings(self):
...
+            Each vision input results in (temporal x height × width) positions. Here: 3 x 2 × 2 = 12 positions total.
+
+            Temporal position IDs are spaced by:
+                `interval = tokens_per_second * temporal_patch_size / fps`
+
+                If fps = 1; tokens_per_second = 25; temporal_patch_size = 2, temporal IDs increase by 50 for each temporal patch:
+                `[0, 0, 0, 0, 50, 50, 50, 50, 100, 100, 100, 100]`
+
+            Height IDs repeat per row: `[0, 0, 1, 1, ...]`
+            Width IDs alternate per column: `[0, 1, 0, 1, ...]`
```
vasqu, line 1076: Just for my understanding, for each I construct some random things
- Temporal
  - If we had 1 height patch instead we would go
  - `[0, 0, 50, 50, 100, 100]` --> essentially "equally distributing" for every temporal patch
- Height
  -  If we had 3 height patches instead we would go
  - `[0, 0, 0, 1, 1, 1, 2, 2, 2, ...]` --> essentially cyclic repetition then
- Width
  -  If we had 3 width patches instead we would go
  - `[0, 1, 2, 0, 1, 2, ...]` --> essentially cyclic repetition then

Ofc everything would need to match to have this divisible etc. Way more understandable than the original version, nice work

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
```diff
@@ -1043,79 +1069,58 @@ def get_rope_index(
...
-                        image_grid_thw[image_index][0],
-                        image_grid_thw[image_index][1],
-                        image_grid_thw[image_index][2],
+        grid_iters = {
+            1: iter(image_grid_thw) if image_grid_thw is not None else None,
+            2: iter(video_grid_thw) if video_grid_thw is not None else None,
+        }
+
+        for batch_idx, current_input_ids in enumerate(input_ids):
+            input_token_type = mm_token_type_ids[batch_idx]
```
vasqu, line 1115: Now that we require `mm_token_type_ids` I suspect this might be breaking for a lot of outside users?

Unsure how many really use our internals here tbh, would still mark it tho with 🚨 how does vllm handle this? Do they rely on our processors or do they provide the position ids on their own?

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
```diff
@@ -1043,79 +1069,58 @@ def get_rope_index(
...
-                    t, h, w = (
-                        image_grid_thw[image_index][0],
-                        image_grid_thw[image_index][1],
-                        image_grid_thw[image_index][2],
+        grid_iters = {
+            1: iter(image_grid_thw) if image_grid_thw is not None else None,
+            2: iter(video_grid_thw) if video_grid_thw is not None else None,
+        }
+
+        for batch_idx, current_input_ids in enumerate(input_ids):
```
vasqu, line 1114: I think this can also be improved although not super easy. We would need to track offsets per batch sequences somehow and remove them from the found groups respectively. (tl;dr: One long flattened sequence to iterate over instead of per batch sample)

If you have the motivation, would be nice but not essential to this PR - I'd rather focus on the unification first.

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
```diff
@@ -1043,79 +1069,58 @@ def get_rope_index(
...
+                # image == 1, video == 2
                 else:
-                    t, h, w = (
-                        video_grid_thw[video_index][0],
-                        video_grid_thw[video_index][1],
-                        video_grid_thw[video_index][2],
+                    grid_thw = next(grid_iters[modality_type])
+                    vision_position_ids = self.get_vision_position_ids(
+                        current_pos, grid_thw, 1, spatial_merge_size, device=input_ids.device
                     )
```
vasqu, line 1142: This is really neat

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
```diff
@@ -1234,7 +1245,7 @@ def compute_3d_position_ids(
                 position_ids = torch.arange(past_key_values_length, past_key_values_length + seq_length)
                 position_ids = position_ids.view(1, 1, -1).expand(3, batch_size, -1).to(inputs_embeds.device)
             delta = self.rope_deltas.repeat_interleave(batch_size // self.rope_deltas.shape[0], dim=0)
-            position_ids = position_ids + delta.to(device=position_ids.device)
+            position_ids = position_ids + delta.to(device=inputs_embeds.device)
```
vasqu, line 1276: Nit: any reason this was changed?

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
```diff
@@ -1552,8 +1567,10 @@ def _prepare_position_ids_for_generation(self, inputs_tensor, model_kwargs):
             inputs_tensor = model_kwargs["input_ids"]
 
         is_input_ids = len(inputs_tensor.shape) == 2 and inputs_tensor.dtype in [torch.int, torch.long]
-        if is_input_ids and (
-            model_kwargs.get("image_grid_thw") is not None or model_kwargs.get("video_grid_thw") is not None
+        if (
+            is_input_ids
+            and model_kwargs.get("mm_token_type_ids") is not None
```
vasqu, line 1600: Do we really want to be that strict? We could also init with zeros otherwise? Unaware whether this would bring more silent issues instead

src/transformers/models/ernie4_5_vl_moe/modular_ernie4_5_vl_moe.py
```diff
@@ -1210,45 +1189,6 @@ def get_image_features(
         image_outputs.pooler_output = image_embeds
         return image_outputs
 
-    def compute_3d_position_ids(
```
vasqu, line 1213: This is now inherited instead 👀 ?

src/transformers/models/ernie4_5_vl_moe/modular_ernie4_5_vl_moe.py
vasqu, line 1: Let me check the local big tests before merging tho

src/transformers/models/glm4v/modular_glm4v.py
```diff
@@ -1178,6 +1029,123 @@ def get_placeholder_mask(
...
+                    # GLM4V splits video into segments per frame but there's only one `grid_thw`
+                    # per whole video. We can't exhaus the iterator and have to re-use the grid
+                    # while processing the same video!
+                    if modality_type == 2:
+                        if video_group_index == 0:
+                            grid_thw = next(grid_iters[modality_type])
+                        video_group_index += 1
+                        video_group_index = 0 if video_group_index >= grid_thw[0] else video_group_index
+                    else:
+                        grid_thw = next(grid_iters[modality_type])
```
vasqu, line 1131: This feels like an opportunity for a helper function? So much is the same, it's a shame if we could not use modular a bit more

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
```diff
@@ -990,44 +991,69 @@ def get_input_embeddings(self):
     def set_input_embeddings(self, value):
         self.language_model.set_input_embeddings(value)
 
+    def get_vision_position_ids(
```
vasqu, line 994: docstring?

vasqu (approved):
Overall LGTM, just added a few smaller questions. Something that we could further improve is to not iterate over each sample within the batch but dw, probably better suited for a future PR that builds on this

(trusting that the tests will be fixed :D)

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
zucchini-nlp: yep, height and width will `repeat` or `repeat_interleave`, while the temporal stays constant per each frame. Sometimes it stays constant per video, depending on how the model is configured

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
zucchini-nlp: Good point, vllm indeed does call `get_rope_index`. I will coordinate with Harry on that

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
zucchini-nlp: Hm, I am not sure if that will be justified since usually we get relatively small batch sizes, and iterating over each shouldn't need much time

I will see what we can do after merging the PR, but I'll put it low prio for now

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
zucchini-nlp: no idea, accidental change which doesn't affect anything since positions are in the same device anyway :)

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
zucchini-nlp: I think without `mm_token_types` we will end up with garbage positions in any case. I added this line so we don't have to check inside `rope_index` and early return

This way, we fallback to basic incremental position ids from `generate` utils, and we don't have to duplocate that code inside `get_rope_index`

src/transformers/models/ernie4_5_vl_moe/modular_ernie4_5_vl_moe.py
zucchini-nlp: yep, thanks to `mm_token_type_ids`

src/transformers/models/qwen2_vl/modeling_qwen2_vl.py
vasqu: Yea no worries, just a thought on improvement since it could add up when we want to improve on big batch inference.

zucchini-nlp:
run-slow: ernie4_5_vl_moe, glm46v, glm4v, glm4v_moe, glm_image, glm_ocr, paddleocr_vl, qwen2_5_vl, qwen2_vl, qwen3_5, qwen3_5_moe, qwen3_vl, qwen3_vl_moe, video_llama_3

zucchini-nlp:
Same tests failing as in `main` for slow CI, verified manually with the latest run. I'll merge when vLLM is oke with this change

tomaarsen:
This PR is introducing a regression for video inputs for Qwen models, see https://github.com/huggingface/transformers/issues/44479

- Tom Aarsen

PR #43981, opened 2026-02-13
Fix early image processor return not raising error

Rocketknight1:
Some of our image processors have a fast return for images that are already square. However, this fast return skips the `background_color` check, which causes flaky test failures because the `test_padding` test uses `self.assertRaises()` to check for that error.

This PR does the check first, ensuring that the right error is raised even when images are square!

Sample error log: https://app.circleci.com/pipelines/github/huggingface/transformers/164053/workflows/a8cc5920-b834-4c5f-bc26-e5fd420c9e18/jobs/2161839

### DIFF
--- a/src/transformers/models/deepseek_vl/image_processing_deepseek_vl.py
+++ b/src/transformers/models/deepseek_vl/image_processing_deepseek_vl.py
@@ -387,6 +387,14 @@ class DeepseekVLImageProcessor(BaseImageProcessor):
         height, width = get_image_size(image, input_data_format)
         num_channels = image.shape[0] if input_data_format == ChannelDimension.FIRST else image.shape[-1]
 
+        # Ensure background_color is the correct shape
+        if isinstance(background_color, int):
+            background_color = [background_color]
+        elif len(background_color) != num_channels:
+            raise ValueError(
+                f"background_color must have no more than {num_channels} elements to match the number of channels"
+            )
+
         if height == width:
             image = (
                 to_channel_dimension_format(image, data_format, input_data_format)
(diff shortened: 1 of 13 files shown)


Rocketknight1:
cc @zucchini-nlp

Rocketknight1:
run-slow: deepseek_vl, deepseek_vl_hybrid, janus, llava, llava_onevision

src/transformers/models/deepseek_vl/image_processing_deepseek_vl.py
```diff
@@ -387,6 +387,14 @@ def pad_to_square(
...
         num_channels = image.shape[0] if input_data_format == ChannelDimension.FIRST else image.shape[-1]
 
+        # Ensure background_color is the correct shape
+        if isinstance(background_color, int):
+            background_color = [background_color]
+        elif len(background_color) != num_channels:
+            raise ValueError(
+                f"background_color must have no more than {num_channels} elements to match the number of channels"
+            )
+
```
zucchini-nlp: I wonder if we really want to raise an error for `background_color` if it's not being used. Maybe update the test to check only with non-square image shapes and add a comment there why?

Rocketknight1: Yeah, maybe that makes more sense. I'll update it!

Rocketknight1:
@zucchini-nlp updated to just fix the test

tests/models/llava/test_image_processing_llava.py
```diff
@@ -225,6 +225,11 @@ def pad_to_square_original(
                 np.testing.assert_allclose(padded_image, padded_image_original)
 
             # background color length should match channel length
+            if image_inputs[0].shape[0] == image_inputs[0].shape[1]:
+                # This avoids a source of test flakiness - if the image is already square
+                # no padding is done and background length is not chacked.
```
Rocketknight1:
```suggestion
                # no padding is done and background colour is not checked.
```

zucchini-nlp (approved):
Nice, thanks!

PR #43914, opened 2026-02-11
Reorder Trainer methods

SunMarc:
# What does this PR do?

This PR reorder all methods from Trainer. No modifications was done to the methods itself. With this final PR, all the major changes should be done and we can focus on simplifying the core logic + testing. 

Reordering makes it hard to go back to previous PRs with git blame so a follow-up would be to create a `.git-blame-ignore-revs `file as this enables github to automatically skip those commits when using blame. 

Here's the structure for now: 

```
# ---- Initialization & Validation ----
 - __init__
 - _validate_args
 - create_accelerator_and_postprocess

# ---- Data Loading ----
 - get_train_dataloader
 - get_eval_dataloader
 - get_test_dataloader
 - num_examples
 - _get_dataloader
 - _get_train_sampler
 - _get_eval_sampler
 - _set_signature_columns_if_needed
 - _remove_unused_columns
 - _get_collator_with_removed_columns

# ---- Optimizer & Scheduler & Learning rate ----
 - create_optimizer_and_scheduler
 - create_optimizer
 - create_scheduler
 - get_optimizer_cls_and_kwargs
 - get_decay_parameter_names
 - _get_learning_rate

# ---- Training ----
 - train
 - _inner_training_loop
 - training_step
 - compute_loss
 - compute_loss_context_manager
 - autocast_smart_context_manager
 - _maybe_log_save_evaluate

# ---- Training Utilites ----
 - get_batch_samples
 - _get_num_items_in_batch
 - _prepare_input
 - _prepare_inputs
 - _prepare_context_parallel_inputs
 - set_initial_training_values
 - get_total_train_batch_size
 - get_sp_size
 - get_cp_size
 - get_tp_size
 - _wrap_model

# ---- Evaluation & Prediction ----
 - evaluate
 - evaluation_loop
 - predict
 - prediction_step
 - _evaluate

# ---- Checkpoint Saving ----
 - _get_output_dir
 - _save_checkpoint
 - _determine_best_metric
 - _save_rng_state
 - _save_optimizer_and_scheduler
 - _save_scaler

# ---- Checkpoint Resuming ----
 - _load_from_checkpoint
 - _load_best_model
 - _load_rng_state
 - _load_optimizer_and_scheduler
 - _load_scaler
 - _load_callback_state
 - _issue_warnings_after_load

# ---- Saving & Serialization ----
 - save_model
 - _save

# ---- Logging & Metrics ----
 - log
 - store_flos
 - floating_point_ops

# ---- Hub Integration ----
[...]

PR #43982, opened 2026-02-13
create .git-blame-ignore-revs file 

SunMarc:
# What does this PR do?

This PR create a `.git-blame-ignore-revs` file to ignore the following commit https://github.com/huggingface/transformers/pull/43914 when using git blame.


winglian (approved):

ArthurZucker (approved):
Sounds good ty!

PR #43989, opened 2026-02-13
Fix AutoVideoProcessor class lookup when torchvision is unavailable

### LINKED ISSUE #43986: Confusing crash when loading a video model through AutoProcessor without torchvision installed
### System Info

The problem exists on the main branch

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

When trying to load a processor using `AutoProcessor` that requires a video processor there's a confusing error message if torchvision is not installed:

```
.../video_processing_auto.py", line 98, in video_processor_class_from_name
    if class_name in extractors:
       ^^^^^^^^^^^^^^^^^^^^^^^^
```

I ran into this when trying to run the latest GLM-OCR model.

The reason this happens is that all values in the global `VIDEO_PROCESSOR_MAPPING_NAMES` is set to `None` when torchvision is not installed.

```py
for model_type, video_processors in VIDEO_PROCESSOR_MAPPING_NAMES.items():
    fast_video_processor_class = video_processors

    # If the torchvision is not available, we set it to None
    if not is_torchvision_available():
        fast_video_processor_class = None

    VIDEO_PROCESSOR_MAPPING_NAMES[model_type] = fast_video_processor_class
```

The value is then used in an if-statement, where it crashes:
```py
        if class_name in extractors:
            ...
```

### Expected behavior

I think the crash message should be improved to mention that user must install torchvision.

Something like

```py
        if extractors is None:
            raise SomeError("No video processors available, make sure to install torchvision...")
        if class_name in extractors:
            ...
```

Abhijeetsingh610:
## What does this PR do?

Fixes a crash in `AutoVideoProcessor` when `torchvision` is unavailable.

`VIDEO_PROCESSOR_MAPPING_NAMES` can contain `None`, and `video_processor_class_from_name` was doing `if class_name in extractors`, which raises `TypeError` for `None`.

This PR:
- skips `None` entries in `video_processor_class_from_name`
- adds a regression test in `tests/models/auto/test_video_processing_auto.py`

Fixes #43986   https://github.com/huggingface/transformers/issues/43986


winglian (approved):

christopher5106:
Hi,

Does this PR also solve this issue: https://github.com/huggingface/diffusers/issues/13140#issuecomment-3899096061

?

Thanks

Abhijeetsingh610:
Thanks for checking. Yes, this PR addresses the same root cause as huggingface/diffusers#13140.
In that issue, the crash happens in transformers.models.auto.video_processing_auto.video_processor_class_from_name when a mapping entry is None (typically when torchvision is unavailable) and code checks class_name in extractors.
This PR adds a None guard for that lookup and includes a regression test.
So once this is merged and released in transformers, that diffusers issue should be resolved when using the updated transformers version.

Frojdholm:
@christopher5106 note that the root cause for the crash is you not having torchvision installed.

If your model requires a video processor it will not work since this will not create one. If the video processor is optional this might solve your issue.

christopher5106:
@Frojdholm the bug appears on an image model, Qwen-Image and Qwen-Image-Edit. In that case, there is no video in inputs and outputs.

zucchini-nlp (commented):
I don't think this is the fix we want. Video processors were originally copied from image processor that have slow/fast versions. But since videos entirely rely on torchvision, we need to raise an error if it's not installed and not skip until the last error message

Same might apply for image processors, only in case when the model has a fast-only processor implemented (we have a couple models only). We need to check if torchvision is installed in L363 `elif video_processor_class is None and not is_torchvision_available(): raise ValueError(informative and actioanable message)`

Issue #43979, opened 2026-02-13
Call to contributions: refactor output tracing in transformers


molbap:
Following #43590 that updates 112 models, we want to finish migrating all models to the standardized output collection interface. So, this is a call to contributors to open PRs to link to this meta-issue and learn about the codebase! 

Two decorators replace the old manual boilerplate:

- **`@capture_outputs`**: goes on the base model forward. Automatically collects `hidden_states`/`attentions` via hooks on submodules declared in `_can_record_outputs`. Removes the need for explicit `output_attentions`/`output_hidden_states`/`return_dict` handling.
- **`@can_return_tuple`**: goes on wrapper forwards (`ForCausalLM`, etc.). Handles `return_dict` only without hook installation.

### How to refactor a model

1. Add `_can_record_outputs = {"hidden_states": DecoderLayer, "attentions": Attention}` on the `PreTrainedModel` subclass. Use **class references**, not strings
2. Add `@capture_outputs` on the base model forward (the one with the layer loop)
3. Add `@can_return_tuple` on higher-level forwards
4. Drop `output_attentions`, `output_hidden_states`, `return_dict` from signatures
5. Remove parameter resolution lines and manual collection loops
6. Decoder layers return `hidden_states` directly; attention always returns `(attn_output, attn_weights)`

You can use `llama`, `mistral`, or `qwen2` as reference implementations. Modern agents are of some help in this issue, but correctly placing the decorators can be tricky. 

### Models to update

Please claim one or several models by commenting before starting. One PR per model (or small group of related models) could be ideal, for instance!

As of today here's the remaining list of models: 

- [ ] `bark`
- [ ] `beit`
- [ ] `bit`
- [ ] `blip_2`
- [ ] `bloom`
- [ ] `canine`
- [ ] `codegen`
- [ ] `colqwen2`
- [ ] `cpmant`
- [ ] `ctrl`
- [ ] `cvt`
- [ ] `dab_detr`
- [ ] `dac`
- [ ] `data2vec`
- [ ] `deberta`
- [ ] `deberta_v2`
- [ ] `decision_transformer`
- [ ] `depth_anything`
- [ ] `depth_pro`
- [ ] `dinat`
- [ ] `donut`
- [ ] `dpr`
- [ ] `efficientnet`
- [ ] `encodec`
- [ ] `falcon`
- [ ] `falcon_mamba`
- [ ] `fastspeech2_conformer`
- [ ] `flaubert`
- [ ] `flava`
[...]

PR #43999, opened 2026-02-14
refactor output tracing in `mobilenet_v1`

omkar-334:
This PR refactors the `mobilenet_v1` model as per #43979

cc @molbap


omkar-334:
I changed the import and updated the test to fix the failing tests, but 1 test is still failing.

PR #44003, opened 2026-02-14
refactor output tracing in `mamba`

omkar-334:
This PR refactors the `mamba`  model as per #43979

cc @molbap
Screenshot 2026-02-15 at 5 12 43 AM

Note - Only 46 tests were passing initially, the rest were failing due to `TypeError: 'int' object is not iterable in SpecifierSet`. I examined this and found out this is because of the library - `"kernels>=0.10.2,<0.11"`. When i upgraded this library, all tests pass.  The latest version is - 
Screenshot 2026-02-15 at 5 15 46 AM  
Should we update `setup.py`?


omkar-334:
@molbap - here I think falconmamba is copied from make fix-repo.  What is the general convention - do we include the overwritten falconmamba over here itself, or do we merge without it and then focus on falconmamba in its own pr?

PR #44005, opened 2026-02-15
Reduce reduce CUDA sync

cyyever:
# What does this PR do?
This PR transfers grid_thw to a python list at the beginning of some functions to reduce later CUDA sync calls. Therefore, several sync calls are merged into one call.

### DIFF
--- a/src/transformers/models/glm_image/modeling_glm_image.py
+++ b/src/transformers/models/glm_image/modeling_glm_image.py
@@ -1100,10 +1100,11 @@ class GlmImageModel(GlmImagePreTrainedModel):
                 decode_height_list = []
                 decode_width_list = []
 
+                curr_grids_list = curr_grids.tolist()
                 for i in range(1, num_decode_grids + 1):
                     grid_idx = -i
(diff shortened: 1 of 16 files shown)


src/transformers/models/glm_image/modeling_glm_image.py
```diff
@@ -1191,12 +1192,12 @@ def get_placeholder_mask(
         """
 
         special_image_mask = input_ids == self.config.image_token_id
-        n_placeholder_tokens = special_image_mask.sum().item()
+        n_placeholder_tokens = special_image_mask.sum()
         n_image_tokens = image_ids.shape[0]
 
         if n_placeholder_tokens != n_image_tokens:
```
Rocketknight1, line 1198: This one doesn't actually avoid a sync, since it has to be implicitly converted to a Python bool for the `if` statement! It's fine to leave it, though.

Rocketknight1 (approved):
Overall this looks good, and thank you for the PR!

ArthurZucker:
are there any benches on any of the files "fixed"

PR #44006, opened 2026-02-15
Use torch.xlogy 

cyyever:
# What does this PR do?

This PR uses torch.xlogy for better numerical handling.


Rocketknight1 (approved):
Agree that this is cleaner than avoiding NaNs with an epsilon!

PR #44009, opened 2026-02-15
update python requirement to 3.10+ to match codebase

mariam851:
Fixes #43976

Updated the documentation to reflect the actual Python requirement (3.10+) as defined in setup.py.

Changes:

Updated README.md .


Rocketknight1 (approved):
Thanks for the fix!

PR #44014, opened 2026-02-15
[UMT5] Ignore tied encoder embedding missing-key warning

### LINKED ISSUE #43992: UMT5Encoder.from_pretrained misses `embed_tokens.weight`
### System Info

```
self.text_encoder = UMT5EncoderModel.from_pretrained(
            "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
            cache_dir=os.environ.get("HF_HOME", None),
            subfolder="text_encoder",
            local_files_only=str2bool(os.getenv("LOCAL_FILES_ONLY", "false")),
        )

# This ensures that the input embeddings are tied to the shared embeddings.
self.text_encoder.set_input_embeddings(self.text_encoder.get_input_embeddings())
```

This will throw a warning
```
Key                         | Status  | 
----------------------------+---------+-
encoder.embed_tokens.weight | MISSING | 
```

Since the weights are copied over to the right parameter by doing **self.text_encoder.set_input_embeddings(self.text_encoder.get_input_embeddings())**, the warning won't be gone due to the fact its coming from `from_pretrained`. The correct behavior should be to tie the weights if users were to use it as a pretrained encoder to encode text. 

Maybe it is better to have an option to indicate whether users want to train its own token embedding vs loading the pretrained one to use it as it is?

### Who can help?

_No response_

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

1. self.text_encoder = UMT5EncoderModel.from_pretrained(
            "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
            cache_dir=os.environ.get("HF_HOME", None),
            subfolder="text_encoder",
            local_files_only=str2bool(os.getenv("LOCAL_FILES_ONLY", "false")),
        )

### Expected behavior

Expected behavior is that it doesn't have any missing weights.

weiguangli-io:
## Summary
Fixes #43992 by preventing a false missing-key report for `UMT5EncoderModel` when `encoder.embed_tokens.weight` is tied to `shared.weight`.

`UMT5EncoderModel` already declares tied weights, but loading checkpoints that only carry the shared embedding can still surface a misleading missing key warning for `encoder.embed_tokens.weight`. This change marks that key as ignorable-on-load for this encoder-only class, matching the tied-embedding semantics.

## Changes
- `src/transformers/models/umt5/modeling_umt5.py`
  - Add `_keys_to_ignore_on_load_missing = [r"encoder.embed_tokens.weight"]` to `UMT5EncoderModel`.
- `tests/models/umt5/test_encoder_missing_keys.py`
  - Add a regression test that loads `UMT5EncoderModel` from a state dict missing `encoder.embed_tokens.weight` and asserts the key is not reported in `missing_keys`.

## Validation
- `PYTHONPATH=src python3 -m ruff check src/transformers/models/umt5/modeling_umt5.py tests/models/umt5/test_encoder_missing_keys.py`
- `PYTHONPATH=src /tmp/tfpycheck/bin/pytest -q tests/models/umt5/test_encoder_missing_keys.py`
- `PYTHONPATH=src /tmp/tfpycheck/bin/coverage run -m pytest -q tests/models/umt5/test_encoder_missing_keys.py`
- `PYTHONPATH=src /tmp/tfpycheck/bin/coverage report -m src/transformers/models/umt5/modeling_umt5.py`

Coverage note:
- The added regression test hits the new load-ignore behavior directly; full-file coverage remains low because the model file contains many unrelated execution paths.

### DIFF
--- a/src/transformers/models/umt5/modeling_umt5.py
+++ b/src/transformers/models/umt5/modeling_umt5.py
@@ -1161,6 +1161,7 @@ class UMT5EncoderModel(UMT5PreTrainedModel):
 
     model_type = "umt5"
     # config_class = UMT5Config
+    _keys_to_ignore_on_load_missing = [r"encoder.embed_tokens.weight"]
     _tied_weights_keys = {
         "encoder.embed_tokens.weight": "shared.weight",
     }
--- a/tests/models/umt5/test_encoder_missing_keys.py
+++ b/tests/models/umt5/test_encoder_missing_keys.py
@@ -1,0 +1,44 @@
+# Copyright 2026 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+from transformers import UMT5Config, UMT5EncoderModel
+from transformers.testing_utils import require_torch
+
+
+@require_torch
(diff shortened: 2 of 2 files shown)


Rocketknight1:
I think this is a duplicate of #43880!

Issue #38071, opened 2025-05-11
transformers showing decoder model architecture detected so padding should be left


sleepingcat4:
### System Info

I had implemented some batching both using pipeline and model.generate method and for Qwen3 and Llama4 models it shows decoder only model detected so padding should mentioned in left. Which is weird as I inspected the model output and it was totally fine. 

However when I mentioned padding to be left, Qwen3 class returned an error. it is very weird. Can this waring/bug be fixed?

### Who can help?

@ArthurZucker 

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

To reproduce please use my below code:

```
def load_pipeline(model_name: str):
    pipe = pipeline(
        "text-generation",
        model=model_name,
        model_kwargs={"torch_dtype": torch.bfloat16, "attn_implementation": "flash_attention_2"},
        device_map="auto"
    )
    tokenizer = AutoTokenizer.from_pretrained(model_name, trust_remote_code=True)
    return pipe, tokenizer

system_prompt = (
    "Think step by step to answer the following question.\n"
    "Return the answer for the MC (Multiple Choice) question at the end of the response after a separator #####.\n"
)

def zero_eval(df, pipe, tokenizer, agieval=False, batch_size=4):
    full_prompts = []

    for _, row in tqdm(df.iterrows(), total=len(df)):
        question = row["question"]
        if agieval:
            choices_str = row["answer"]
        else:
            choices = eval(row["options"])
            choices_str = "\n".join([f"({chr(65+i)}) {c}" for i, c in enumerate(choices)])
        user_prompt = f"---\nQ: {question}\nChoices:\n{choices_str}\nA:"
        full_prompts.append(user_prompt)

    batch_messages = [[
        {"role": "system", "content": system_prompt},
        {"role": "user", "content": prompt}
    ] for prompt in full_prompts]

    prompts = [
        tokenizer.apply_chat_template(msg, tokenize=False, enable_thinking=False, add_generation_prompt=True)
        for msg in batch_messages
    ]

    outputs = []
    for i in tqdm(range(0, len(prompts), batch_size)):
        batch = prompts[i:i+batch_size]
        result = pipe(batch, batch_size=batch_size, max_new_tokens=2035)
        outputs.extend([r[0]["generated_text"] for r in result])

[...]

PR #44021, opened 2026-02-15
Fix false positive right-padding warning for decoder-only models in pipeline

### LINKED ISSUE #43906: Isolated reproduction of https://github.com/huggingface/transformers/issues/38071
### System Info

name = "accelerate"
version = "1.12.0"
name = "transformers"
version = "4.57.3"

Python 3.11

### Who can help?

@gante  @ArthurZucker  Related to warning from https://github.com/huggingface/transformers/issues/38071 for `Qwen/Qwen3-Next-80B-A3B-Instruct` model

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

```
from tqdm import tqdm
LANGUAGE_MODEL = "Qwen/Qwen3-Next-80B-A3B-Instruct"
def main():

    pipe = pipeline("text-generation", model=LANGUAGE_MODEL, device_map="auto")
    messages = [[
        {"role": "system", "content": "hi"},
        {"role": "user", "content": "hdi"},
    ],[
        {"role": "system", "content": "hi"},
        {"role": "user", "content": "hddi"},
    ],[
        {"role": "system", "content": "hi"},
        {"role": "user", "content": "hasdasi"},
    ],[
        {"role": "system", "content": "hi"},
        {"role": "user", "content": "hiasdsad"},
    ],[
        {"role": "system", "content": "hi"},
        {"role": "user", "content": "hiasd"},
    ],[
        {"role": "system", "content": "hi"},
        {"role": "user", "content": "hiasd"},
    ]]

    BATCH_SIZE = 3
    for out in tqdm(
        pipe(messages, max_new_tokens=4, batch_size=BATCH_SIZE),
        total=len(messages),
        desc="Batched inference",
    ):
        response = out[0]["generated_text"][-1]["content"].strip()
        print(response)

if __name__ == "__main__":
    main()
```

### Expected behavior

```
(ecg-preprocess) (ecg-encoder) -bash-4.4$ CUDA_VISIBLE_DEVICES=4,5,6,7 uv run src/test.py
The fast path is not available because one of the required library is not installed. Falling back to torch implementation. To install follow https://github.com/fla-org/flash-linear-attention#installation and https://github.com/Dao-AILab/causal-conv1d
Loading checkpoint shards: 100%|█████████████████| 41/41 [00:31<00:00,  1.28it/s]
Device set to use cuda:0
A decoder-only architecture is being used, but right-padding was detected! For correct generation results, please set `padding_side='left'` when initializing the tokenizer.
A decoder-only architecture is being used, but right-padding was detected! For correct generation results, please set `padding_side='left'` when initializing the tokenizer.
Batched inference:   0%|                                   | 0/6 [00:00<?, ?it/s])  
Hi!
Hi! It looks
Hello! It seems
Hello! It seems
)
) I'm here
Batched inference: 100%|████████████████████████| 6/6 [00:00<00:00, 82782.32it/s]
```

ManasVardhan:
## What does this PR do?

Fixes #43906 (related to #38071)

### Problem
When using `pipeline('text-generation')` with batched inference on Qwen3 (and other models where `pad_token_id == bos_token_id`), a spurious warning is emitted:

> A decoder-only architecture is being used, but right-padding was detected!

This happens for two reasons:
1. The `TextGenerationPipeline` doesn't set `padding_side='left'` for decoder-only models, so the default `'right'` padding is used during batch collation
2. The right-padding detection heuristic in `generate()` only checks if the last token equals `pad_token_id`, which can produce false positives when `pad_token_id` equals other special tokens

### Fix
1. **`TextGenerationPipeline.__init__`**: Automatically set `tokenizer.padding_side = 'left'` for decoder-only models (since `TextGenerationPipeline` is exclusively for causal LM)
2. **`GenerationMixin.generate`**: When an `attention_mask` is available, use it to detect right-padding (check if last position has mask=0) instead of relying solely on the token id heuristic. Falls back to the original check when no attention mask is provided.

### Root Cause Analysis
For Qwen3, `pad_token_id = bos_token_id = 151643` (`<|endoftext|>`). The tokenizer's default `padding_side='right'` means shorter sequences in a batch get right-padded. The existing check `inputs_tensor[:, -1] == pad_token_tensor` then correctly detects this — but the real issue is that the pipeline should have been left-padding all along.

Even after fixing the pipeline, the attention-mask-based detection is more robust for cases where users call `model.generate()` directly with properly left-padded inputs whose content happens to end with the pad token.

## Who can review?
@gante @ArthurZucker


zucchini-nlp (approved):
Thanks, can you chekc failing whisper pipeline tests?

https://app.circleci.com/pipelines/github/huggingface/transformers/164182/workflows/d3394016-7983-47ee-963d-6592873bdf01/jobs/2163511/tests

ManasVardhan:
Thanks for flagging! The Whisper pipeline tests were failing because `WhisperForCausalLM` has both a tokenizer and a feature_extractor — my change set `tokenizer.padding_side='left'` which conflicted with the feature_extractor's `padding_side='right'`, causing a `ValueError` in `pad_collate_fn`.

Fixed in 3c63b39 — now we only override `padding_side` when no `feature_extractor` is present.

zucchini-nlp:
Thanks, I pushed a fix so we can merge soon. We don't want to check `self.feature_extractor` in this pipe, the pipe generates text from text and thus is expected to not load a `feature_extractor`. There was an issue in how test is written

PR #44030, opened 2026-02-16
refactor output tracing in `dpr`

omkar-334:
This PR refactors the`dpr` model as per #43979

cc @molbap
Screenshot 2026-02-16 at 9 13 30 AM


src/transformers/models/dpr/modeling_dpr.py
```diff
@@ -561,7 +545,6 @@ def forward(
             inputs_embeds=inputs_embeds,
             output_attentions=output_attentions,
             output_hidden_states=output_hidden_states,
```
molbap: Thanks for the contrib! 
Here and everywhere in the modeling file,, `output_attentions` and `output_hidden_states` should not be explicitly mentioned anywhere anymore,`capture_outputs` ought to be added instead.

PR #44032, opened 2026-02-16
[Misc][vlms] Use text_config when initializing the fine-grained FP8Expert

JJJYmmm:
# What does this PR do?

Update FP8 expert replacement to use `model.config.text_config` when available (VLMs), falling back to model.config if it's text-only models.

### DIFF
--- a/src/transformers/integrations/finegrained_fp8.py
+++ b/src/transformers/integrations/finegrained_fp8.py
@@ -745,7 +745,9 @@ def replace_with_fp8_linear(
         with torch.device("meta"):
             if module_name.endswith(".experts"):
                 new_module = FP8Expert(
-                    config=model.config, block_size=quantization_config.weight_block_size, **module_kwargs
+                    config=model.config.get_text_config(),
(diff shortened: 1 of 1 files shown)


src/transformers/integrations/finegrained_fp8.py
```diff
@@ -745,7 +745,7 @@ def replace_with_fp8_linear(
         with torch.device("meta"):
             if module_name.endswith(".experts"):
                 new_module = FP8Expert(
-                    config=model.config, block_size=quantization_config.weight_block_size, **module_kwargs
+                    config=getattr(model.config, "text_config", model.config), block_size=quantization_config.weight_block_size, **module_kwargs
```
zucchini-nlp: nit: we can use `config.get_text_config()`

JJJYmmm: got it🫡

zucchini-nlp (approved):
thanks, lgtm

PR #44035, opened 2026-02-16
bring back our demons: clean_up_tokenization_spaces

ArthurZucker:
# What does this PR do?
We already brought it back with: 
```python
        if clean_up_tokenization_spaces:
            # Call custom cleanup method if it exists (e.g., for CLVP's [SPACE] token replacement)
            if hasattr(self, "clean_up_tokenization") and callable(self.clean_up_tokenization):
                text = self.clean_up_tokenization(text)
            else:
                # Otherwise apply standard cleanup
                text = (
                    text.replace(" .", ".")
                    .replace(" ?", "?")
                    .replace(" !", "!")
                    .replace(" ,", ",")
                    .replace(" ' ", "'")
                    .replace(" n't", "n't")
                    .replace(" 'm", "'m")
                    .replace(" 's", "'s")
                    .replace(" 've", "'ve")
                    .replace(" 're", "'re")
                )
```
which basically always calls it. 
We are doing this now because it allows to fix custom remote model that expect this to still exist.

### DIFF
--- a/src/transformers/models/luke/tokenization_luke.py
+++ b/src/transformers/models/luke/tokenization_luke.py
@@ -418,18 +418,7 @@ class LukeTokenizer(TokenizersBackend):
             else self.clean_up_tokenization_spaces
         )
         if clean_up_tokenization_spaces:
-            text = (
-                text.replace(" .", ".")
-                .replace(" ?", "?")
-                .replace(" !", "!")
-                .replace(" ,", ",")
-                .replace(" ' ", "'")
-                .replace(" n't", "n't")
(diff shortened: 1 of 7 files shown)


itazap:
nice thanks - we can also then rm this code then from `tokenization_utils_tokenizers.py`, `toknization_python.py`, and `tokenization_mistral_common.py` !

hmellor (approved):
This fixes the issues with remote models in vLLM.

Let's take @itazap's suggestion to deduplicate this BC code.

PR #44037, opened 2026-02-16
Add a dim check mechanism in Transpose and fix qwen3_vl_moe weight mapping

Cyrilvallez:
# What does this PR do?

As per the title. Supersedes https://github.com/huggingface/transformers/pull/43913

### DIFF
--- a/src/transformers/conversion_mapping.py
+++ b/src/transformers/conversion_mapping.py
@@ -143,12 +143,12 @@ def _build_checkpoint_conversion_mapping():
             WeightConverter(
                 source_patterns="mlp.experts.gate_up_proj",
                 target_patterns="mlp.experts.gate_up_proj",
-                operations=[Transpose(1, 2, sentinel_dim=2), Force16BytesAlignment()],
+                operations=[Transpose(1, 2, sentinel=True), Force16BytesAlignment()],
             ),
             WeightConverter(
                 source_patterns="mlp.experts.down_proj",
                 target_patterns="mlp.experts.down_proj",
-                operations=[Transpose(1, 2, sentinel_dim=1), Force16BytesAlignment()],
+                operations=[Transpose(1, 2, sentinel=True), Force16BytesAlignment()],
             ),
         ],
         "phimoe": [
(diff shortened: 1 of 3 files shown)


src/transformers/conversion_mapping.py
```diff
@@ -141,17 +141,14 @@ def _build_checkpoint_conversion_mapping():
...
         "qwen3_vl_moe": [
             WeightConverter(
-                source_patterns=[
-                    "mlp.experts.*.gate_proj.weight",
-                    "mlp.experts.*.up_proj.weight",
-                ],
+                source_patterns="mlp.experts.gate_up_proj",
                 target_patterns="mlp.experts.gate_up_proj",
-                operations=[MergeModulelist(dim=0), Concatenate(dim=1), Transpose(1, 2)],
+                operations=[Transpose(1, 2, sentinel=True), Force16BytesAlignment()],
```
ArthurZucker:
```suggestion
                operations=[Transpose(1, 2, conditional_check=True), Force16BytesAlignment()],
```
or something like that?

src/transformers/core_model_loading.py
```diff
@@ -260,7 +261,18 @@ def convert(
...
         tensors = next(iter(input_dict.values()))
         tensor = tensors[0] if isinstance(tensors, list) else tensors
-        return {target_pattern: torch.transpose(tensor, dim0=self.dim0, dim1=self.dim1).contiguous()}
+        # In this case, always transpose
+        if not self.sentinel:
+            return {target_pattern: torch.transpose(tensor, dim0=self.dim0, dim1=self.dim1).contiguous()}
+        # In this case, check the shapes before transposing
+        else:
+            # NOTE: this rely on the first param name, so cannot be used for many-to-one operation
+            expected_shape = kwargs["model"].get_parameter(kwargs["full_layer_name"]).shape
```
ArthurZucker, line 270: good, we should probably make the TP op overwrite the shape ? or actually the shape you check is never TPed?

ArthurZucker (approved):
IMO much better!

src/transformers/core_model_loading.py
Cyrilvallez: TP is still not defined for only a Transpose anyway... IMO we should wait for support before taking a decision on this

PR #44040, opened 2026-02-16
Fix gemma3n `get_audio_features`

### LINKED ISSUE #44008: [Gemma 3n][modular] AttributeError: 'Tensor' object has no attribute 'audio_mel_mask' — variable name collision in Gemma3nModel.forward()
## System Info

- `transformers` version: 4.53.0 (also confirmed on latest `main` as of 2026-02-15)
- Python: 3.12
- PyTorch: 2.7+
- Platform: Linux - x86_64
- Model: `google/gemma-3n-E2B-it` (also affects `google/gemma-3n-E4B-it`)
- Reported by: @reedmayhew18

## Who can help?

@ArthurZucker @zucchini-nlp @Rocketknight1

## Bug Description

When performing audio inference with Gemma 3n using `AutoModelForImageTextToText`, the model crashes with:

```
AttributeError: 'Tensor' object has no attribute 'audio_mel_mask'
```

This is caused by a **variable name collision** in `Gemma3nModel.forward()` (in `modular_gemma3n.py`) where `audio_features` (a `Gemma3nAudioEncoderModelOutput` dataclass) is overwritten by its own `.pooler_output` attribute (a plain `Tensor`), and then `.audio_mel_mask` is accessed on the now-plain Tensor.

## Root Cause

In [`modular_gemma3n.py`, inside `Gemma3nModel.forward()` at **lines 2388-2392**](https://github.com/huggingface/transformers/blob/c9ea365a7b56326418769a4ba4682864d407ed63/src/transformers/models/gemma3n/modular_gemma3n.py#L2389C1-L2392C55):

```python
# Merge text and audio
if input_features is not None and input_features_mask is not None:
    audio_features = self.get_audio_features(input_features, ~input_features_mask, return_dict=True)
    audio_features = audio_features.pooler_output      # ← Line 2391: Overwrites dataclass with plain Tensor
    audio_mask = audio_features.audio_mel_mask          # ← Line 2392: CRASH - Tensor has no .audio_mel_mask
```

After **line 2391**, `audio_features` is no longer the `Gemma3nAudioEncoderModelOutput` dataclass — it's been reassigned to `.pooler_output` which is a `torch.Tensor`. **Line 2392** then tries to access `.audio_mel_mask` on that Tensor, which doesn't exist.

**Note:** `get_audio_features()` correctly returns a `Gemma3nAudioEncoderModelOutput` with both `.pooler_output` and `.audio_mel_mask`. The issue is purely the variable name reuse in the calling code.

## Suggested Fix

Use a separate variable name to preserve the full output object:

```python
# Merge text and audio
if input_features is not None and input_features_mask is not None:
    audio_outputs = self.get_audio_features(input_features, ~input_features_mask, return_dict=True)
    audio_features = audio_outputs.pooler_output       # Tensor — the embedded audio
    audio_mask = audio_outputs.audio_mel_mask           # BoolTensor — preserved from dataclass
```

This same pattern is already used correctly for vision in the same method (where `image_features` is extracted via `get_image_features(...).pooler_output` without overwriting the original output object).

**Important:** Since `modeling_gemma3n.py` is **auto-generated** from `modular_gemma3n.py`, this fix **must** be applied to `modular_gemma3n.py` (the source file) to prevent the bug from being regenerated.

## Steps to Reproduce

```python
from transformers import AutoModelForImageTextToText, AutoProcessor

model_id = "google/gemma-3n-E2B-it"
processor = AutoProcessor.from_pretrained(model_id)
model = AutoModelForImageTextToText.from_pretrained(model_id, torch_dtype="auto", device_map="auto")

messages = [
    {
        "role": "user",
        "content": [
            {"type": "audio", "audio": "any_valid_audio_file.wav"},
            {"type": "text", "text": "What is being said in this audio?"},
        ],
    }
]

inputs = processor.apply_chat_template(
    messages,
    add_generation_prompt=True,
    tokenize=True,
    return_dict=True,
    return_tensors="pt",
).to(model.device)

# This crashes:
outputs = model.generate(**inputs, max_new_tokens=256)
```

## Full Traceback

```
  File ".../transformers/generation/utils.py", line 2638, in generate
    result = decoding_method(...)
  File ".../transformers/generation/utils.py", line 2833, in _sample
  File ".../transformers/generation/utils.py", line 3822, in _prefill
    return self(**model_inputs, return_dict=True)
  File ".../transformers/models/gemma3n/modeling_gemma3n.py", line 2284, in forward
    outputs = self.model(...)
  File ".../transformers/models/gemma3n/modular_gemma3n.py", line 2392, in forward
    audio_mask = audio_features.audio_mel_mask
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Tensor' object has no attribute 'audio_mel_mask'
```

## Additional Context

- **Image-only** and **text-only** inference work fine — only **audio** triggers this bug.
- The bug exists in both:
  - `modular_gemma3n.py` (the **source** file, **lines 2388-2392**)
  - `modeling_gemma3n.py` (the **auto-generated** file from modular source)
- The [official Google Colab notebook](https://colab.research.google.com/github/google-gemini/gemma-cookbook/blob/main/Gemma/%5BGemma_3n%5DMultimodal_understanding_with_HF.ipynb) may appear to work because it uses a pinned/pre-release version of transformers or processes audio differently.
- The same variable-naming pattern is already implemented correctly for vision inputs in the same method.

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/44008 and re-enables tests


zucchini-nlp:
run-slow: gemma3n

tests/models/gemma3n/test_modeling_gemma3n.py
```diff
@@ -781,17 +808,143 @@ def test_eager_padding_matches_padding_free_with_position_ids(self):
...
+
+    @unittest.skip("Audio modality is not tested here")
+    def test_get_audio_features_attentions(self, return_dict: bool | None):
+        pass
+
+    @unittest.skip(
+        "Timm backbone buffers can't be init correctly on loading back! Specifically `vision_tower.timm_model.rope.pos_embed`"
+    )
+    def test_save_load(self):
+        pass
```
zucchini-nlp: important imo, cc @Cyrilvallez in case you don't yet have it in your todo's

tests/models/gemma3n/test_modeling_gemma3n.py
```diff
@@ -1009,16 +1161,11 @@ def test_model_1b_text_only(self):
         EXPECTED_TEXTS = ['Write a poem about Machine Learning.\n\n---\n\nThe data flows, a river deep,\nWith patterns hidden, secrets sleep.\nA neural net, a watchful eye,\nLearning']  # fmt: skip
         self.assertEqual(output_text, EXPECTED_TEXTS)
 
-    @parameterized.expand([("sdpa",), ("eager",), ("flash_attention_2",)])
-    def test_generation_beyond_sliding_window(self, attn_implementation: str):
+    def test_generation_beyond_sliding_window(self):
```
zucchini-nlp, line 1154: model doesn't support attn impl on the timm vision backbone, so we leave only "eager"

zucchini-nlp:
run-slow: gemma3n

tests/models/gemma3n/test_modeling_gemma3n.py
Cyrilvallez: Humm, they should be able to 🤔 We worked with the timm team for this https://github.com/huggingface/transformers/blob/main/src/transformers/models/timm_wrapper/modeling_timm_wrapper.py#L115-L117

Cyrilvallez: I believe something else is probably wrong here

zucchini-nlp: maybe, tho it's definitely from `timm`. PerceptionLM is another model with timm backbone where this test fails

src/transformers/models/gemma3n/modeling_gemma3n.py
```diff
@@ -1503,6 +1422,90 @@ def _init_weights(self, module):
         if hasattr(module, "gradient_clipping"):
             init.constant_(module.gradient_clipping, self.config.gradient_clipping)
 
+        if hasattr(module, "pos_embed"):
+            init.constant_(module.pos_embed, self.config.gradient_clipping)
```
Cyrilvallez: cc @zucchini-nlp this has the same name as the param that apparently does not init correctly, so may be the culprit even though it should not - in any case, I don't see amy `pos_embed` anywhere in the modleing

zucchini-nlp: ah wait, it's the timm version. Since you updated that recently, I realized that I don't have the latest version 🤦🏻

zucchini-nlp: can you review the rest pls?

Cyrilvallez: Ha indeed! We bumped the requirement, but if you had it already installed that's why!

tests/models/gemma3n/test_modeling_gemma3n.py
```diff
@@ -649,39 +647,62 @@ def __init__(
...
+        self.eoi_token_id = eoi_token_id
+        self.boa_token_id = boa_token_id
+        self.eoa_token_id = eoa_token_id
+        self.audio_token_id = audio_token_id
         self.llm_tester = Gemma3nTextModelTester(self.parent)
         self.text_config = self.llm_tester.get_config()
+        self.audio_tester = Gemma3nAudioModelTester(self.parent)
+        self.audio_config = self.audio_tester.get_audio_encoder_config()
+        # NOTE: gemma3n uses mobilenet backbone but timm doens't let us
+        # create a tiny MobileNet. So we use a random ViT backbone for testing!
```
Cyrilvallez, line 677: No way to tweak the sizes to get a tiny one?

tests/models/gemma3n/test_modeling_gemma3n.py
```diff
@@ -781,17 +807,139 @@ def test_eager_padding_matches_padding_free_with_position_ids(self):
...
+    def test_training_gradient_checkpointing(self):
+        pass
+
+    @unittest.skip("timm model has no gradient")
+    def test_training_gradient_checkpointing_use_reentrant_true(self):
+        pass
+
+    @unittest.skip("timm model has no gradient")
+    def test_training_gradient_checkpointing_use_reentrant_false(self):
+        pass
```
Cyrilvallez, line 832: That's weird that they have no gradients - are they all frozen by default?

tests/models/gemma3n/test_modeling_gemma3n.py
```diff
@@ -781,17 +807,139 @@ def test_eager_padding_matches_padding_free_with_position_ids(self):
...
+
+    @unittest.skip("Audio modality is not tested here")
+    def test_get_audio_features_attentions(self, return_dict: bool | None):
+        pass
+
+    @unittest.skip(
+        "Timm backbone buffers can't be init correctly on loading back! Specifically `vision_tower.timm_model.rope.pos_embed`"
+    )
+    def test_save_load(self):
+        pass
```
Cyrilvallez: SHould not be needed with the right timm version - let's remove!

Cyrilvallez (approved):
Very nice, thanks for taking care of the tests at the same time ❤️ Just a few remarks, feel free to merge if not possible

tests/models/gemma3n/test_modeling_gemma3n.py
zucchini-nlp: I couldn't find a way, timm has hardcoded `num_stages` depending on variant and there's no tiny variant 😢

tests/models/gemma3n/test_modeling_gemma3n.py
zucchini-nlp: maybe it's me again with timm version, gotta re-check XD

PR #43985, opened 2026-02-13
Refactor inner training loop

winglian:
# What does this PR do?

The `_inner_training_loop` method has a lot going on which makes it hard to extend for downstream developers/libraries. This PR breaks it up into smaller well described methods that are chained in the training loop.

Additionally, this adds the concept of a DataProducer that can wrap the dataset/sampler/dataloader which allows for better handling of Iterable data and Async data for use cases like async RL downstream (e.g. trl)

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #44041, opened 2026-02-16
refactor _inner_training_loop to smaller methods

winglian:
# What does this PR do?

Alternate PR to #43985 to be a reorder only PR.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


src/transformers/trainer.py
SunMarc: can't we move that into `set_initial_training_values` ?

SunMarc: could it be worth to put in a separate method ?

SunMarc, line 1447: Note but it would be nice to upstream fsdpv2 xla in accelerate so that we can get rid of those

SunMarc: maybe we should move this somewhere else ?

SunMarc: same for this, maybe we can move this somewhere else ? Also not sure how much this is used. I was in favor to deprecate that

src/transformers/trainer.py
```diff
@@ -1462,6 +1462,44 @@ def _inner_training_loop(
...
+                epochs_trained=epochs_trained,
+                steps_trained_in_current_epoch=steps_trained_in_current_epoch,
+            )
+            if self.control.should_training_stop:
+                break
+
+        return self._finalize_training(model, trial, num_train_samples, start_time)
+
+    def _setup_training(self, args, max_steps, resume_from_checkpoint, train_dataloader):
+        """Create optimizer, wrap model, load checkpoint. Returns (wrapped_model, train_dataloader)."""
```
SunMarc: Update a bit the docstring here. We also create/update the lr_scheduler.

src/transformers/trainer.py
```diff
@@ -1583,6 +1620,26 @@ def _inner_training_loop(
...
+        num_examples,
+        len_dataloader,
+        train_dataloader,
+        resume_from_checkpoint,
+        trial,
+        ignore_keys_for_eval,
+    ):
+        """Initialize training loop state. Returns (epochs_trained, steps_trained_in_current_epoch, start_time)."""
+        self.state.is_hyper_param_search = trial is not None
+
```
SunMarc: not sure to move that here

src/transformers/trainer.py
SunMarc: I saw your other PR, we can clean/upgrade the logging later on

src/transformers/trainer.py
```diff
@@ -1632,240 +1689,257 @@ def _inner_training_loop(
...
         self._globalstep_last_logged = self.state.global_step
         model.zero_grad()
-        grad_norm: float | None = None
-        learning_rate = None
+        self._grad_norm: float | None = None
+        self._learning_rate = None
         self.control = self.callback_handler.on_train_begin(args, self.state, self.control)
 
         if args.eval_on_start:
             self._evaluate(trial, ignore_keys_for_eval, skip_scheduler=True)
```
SunMarc: I think we can move this out ?

src/transformers/trainer.py
```diff
@@ -1583,6 +1620,26 @@ def _inner_training_loop(
...
+        max_steps,
+        total_train_batch_size,
+        num_examples,
+        len_dataloader,
+        train_dataloader,
+        resume_from_checkpoint,
+        trial,
+        ignore_keys_for_eval,
+    ):
+        """Initialize training loop state. Returns (epochs_trained, steps_trained_in_current_epoch, start_time)."""
```
SunMarc: Maybe we can try to further clean this and try to make sense of what should be here or not.

SunMarc (commented):
I saw that you mostly split the _inner_training_loop into three separate private methods. I was thinking maybe it could the chance to try simplifying and reorganizing `_inner_training_loop` ? Like move things that feel out of place somewhere else and simplify a logic that feels a bit strange/complicated.

src/transformers/trainer.py
winglian: should we decide what to do with this in a separate PR?

SunMarc:
I've tried to simplify/reorganize a bunch of things to make things a bit clearer. I tweaked a bit with the internals but I will probably stop there and first get this merged before it gets too complicated to follow all the changes.
cc @kashif @qgallouedec @winglian

SunMarc:
Some FSDPv2 tests are not passing on trl , investigating. The issue was the following: 

The crash only happens when `accelerator._models` contains both an unwrapped model (regular Tensors) and an FSDP2-wrapped model (`DTensors`). `GRPOTrainer` registers a reward model via `accelerator.prepare_model(evaluation_mode=True)` during `__init__`, which adds it to `_models` without FSDP-wrapping. Regular Trainer tests don't register extra models. This only happens with GRPO and RLOO.

qgallouedec:
@SunMarc 

is it something that is done wrong in grpo/rloo?

src/transformers/testing_utils.py
```diff
@@ -1468,12 +1468,8 @@ def get_steps_per_epoch(trainer: Trainer) -> int:
...
     train_dataloader = trainer.get_train_dataloader()
 
-    initial_training_values = trainer.set_initial_training_values(
-        args=training_args,
-        dataloader=train_dataloader,
-        total_train_batch_size=training_args.per_device_train_batch_size,
-    )
-    steps_per_epoch = initial_training_values[1]
+    initial_training_values = trainer.set_initial_training_values(args=training_args, dataloader=train_dataloader)
+    steps_per_epoch = initial_training_values[5]
```
qgallouedec, line 1472: can you explain this change? the `get_steps_per_epoch` function changed from returning `num_update_steps_per_epoch` (index 1) to `steps_in_epoch` (index 5), which are semantically different values.

winglian: `steps_per_epoch` looks to be number of mini-steps, whereas `num_update_steps_per_epoch` is the steps we usually see in the progress bar which encapsulates `gradient_accumulation_steps` mini-steps.

winglian: see https://github.com/huggingface/transformers/pull/44041/changes/daec1155e2d9ef3685057917f742a04462e9e316#diff-ed55888e6665791fe92cc8fc0c499da54f4ace6738551cd9a2591881cda076deL1658-L1662

qgallouedec:
trl tests look good: https://github.com/huggingface/trl/actions/runs/22246318776 (distributed smoke test is in queue, I don't know why it doesn't start)

winglian (commented):
Latest changes all look good. I can't approve the PR since I opened it.

SunMarc (approved):
Let's get this merged ! thanks !

SunMarc:
> is it something that is done wrong in grpo/rloo?

No it was just something that I deleted `self.accelerator.free_memory()` as I thought we shouldn't need it but it turns out we needed that.

PR #44042, opened 2026-02-16
Update assignee for generate in bug report template

Rocketknight1:
Joao is regrettably no longer with us :saluting_face:  so we should really stop getting users to ping him! This PR makes @cyrilvallez responsible for `generate` issues outside of VLMs.


Cyrilvallez (approved):
Thanks

PR #44043, opened 2026-02-16
`grouped_mm` fallback

IlyasMoutawwakil:
# What does this PR do?

With this PR, grouped_mm impl will always be default when the model uses experts impl (even in old torch versions) and will no longer fail if you are torch compiling an fp16/fp32 model, the impl will succeed in all cases, except cuda-graphs (common failure case in all grouped_mm impls).

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


src/transformers/modeling_utils.py
```diff
@@ -110,7 +110,6 @@
     is_env_variable_true,
     is_flash_attn_2_available,
     is_flash_attn_3_available,
-    is_grouped_mm_available,
```
IlyasMoutawwakil, line 113: should we keep the function `is_grouped_mm_available` ? i didn't erase it for bc

ArthurZucker: for bc for now ?

src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
```diff
@@ -452,9 +452,8 @@ class DeepseekV2PreTrainedModel(PreTrainedModel):
     _supports_flash_attn = True
     _supports_sdpa = True
     _supports_flex_attn = True
-    _can_compile_fullgraph = (
-        is_grouped_mm_available()
-    )  # https://huggingface.co/docs/transformers/experts_interface#torchcompile
+
+    _can_compile_fullgraph = True
```
ArthurZucker, line 456: actually eager is not no?

ArthurZucker (approved):
Very nice work!

src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
IlyasMoutawwakil: so before we had "if grouped_mm is available then its fullgraph compilable" because we default to grouped_mm when available. and now its always available so it makes sense to set it to True since all models that can use the experts impl will use it.

src/transformers/modeling_utils.py
IlyasMoutawwakil: maybe a small warning when it's used ? my worry is if someone uses it like:
```
if is_grouped_m_available():
    model.set_experts_impl("grouped_mm")
```
thinking that grouped_mm can only be used if is_grouped_m_available() evaluates to true

src/transformers/models/deepseek_v2/modeling_deepseek_v2.py
ArthurZucker: okay!

src/transformers/integrations/moe.py
```diff
@@ -156,44 +156,108 @@ def batched_mm_experts_forward(
...
+        input (`torch.Tensor`):
+            Input tensor of shape (S, input_dim).
+        weight (`torch.Tensor`):
+            Weight tensor of shape (num_experts, input_dim, output_dim).
+        offs (`torch.Tensor`):
+            Offsets tensor indicating the boundaries of each group in the input tensor.
+    Returns:
+        `bool`: True if grouped_mm can be used, False otherwise.
+    """
+    if torch.compiler.is_compiling() and weight.dtype != torch.bfloat16:
```
vasqu, line 218: Should we have used the `is_tracing` helper instead?

docs/source/en/experts_interface.md
```diff
@@ -81,11 +81,12 @@ model = AutoModelForCausalLM.from_pretrained(
...
-| -------------- | ------------------------------------ | -------------------------------- | ---------------- |
-| `grouped_mm`   | `None`, `max-autotune-no-cudagraphs` | `bfloat16`                       | Yes              |
-| `batched_mm`   | all                                  | `bfloat16`, `float16`, `float32` | Yes              |
-| `eager`        | all                                  | `bfloat16`, `float16`, `float32` | No               |
+| Implementation          | compilation modes                    | dtypes                           | `fullgraph=True` |
+| ----------------------- | ------------------------------------ | -------------------------------- | ---------------- |
+| `grouped_mm`            | `None`, `max-autotune-no-cudagraphs` | `bfloat16`                       | Yes              |
+| `grouped_mm` (fallback) | `None`, `max-autotune-no-cudagraphs` | `bfloat16`, `float16`, `float32` | Yes              |
+| `batched_mm`            | all                                  | `bfloat16`, `float16`, `float32` | Yes              |
+| `eager`                 | all                                  | `bfloat16`, `float16`, `float32` | No               |
```
vasqu, line 89: More for my interest but now that we have the "naive" grouped mm fallback, how does it compare to the compiled eager mode, e.g. in fp16, speed/memory wise?

tests/test_modeling_common.py
```diff
@@ -4064,11 +4064,6 @@ def test_torch_compile_for_training(self):
 
         model = cls(config).to(device=torch_device)
 
-        # torch.nn.functional.grouped_mm still only supports bfloat16 when used with torch.compile
-        # bfloat16 is problematic with precisions so we use an implementation with full precision
-        if model.config._experts_implementation == "grouped_mm":
-            model.set_experts_implementation("batched_mm")
```
vasqu, line 4070: Glad to see these gone :)

vasqu (commented):
Late to the party, not relevant for the merge now but 2 quick questions/nudges

tests/test_modeling_common.py
IlyasMoutawwakil: knew you would be happy haha

docs/source/en/experts_interface.md
IlyasMoutawwakil: my intuition is that it's overall better, notice that with the fallback we can apply the gating mechanism once (on a bigger tensor). it's also compatible with fullgraph (though it does sync cpu-gpu, but only once per call in https://github.com/huggingface/transformers/pull/44152). I'll run a benchmark to be sure.

src/transformers/integrations/moe.py
IlyasMoutawwakil: hmm i don't think so, the issue here is that when compiling, torch replaces the grouped_mm operator with a triton implementation that only supports bf16. `is_tracing` also evaluates to true when stream capturing (cuda graphs), where the fp32 torch.grouped_mm works fine.

vasqu: Fair enough, I guess the compiler api is stable for torch 2.4+? I'd prefer to use the `is_torchdynamo_compiling` just to be sure

docs/source/en/experts_interface.md
IlyasMoutawwakil: - very competitive prefill ; way better than eager but not as fast as grouped_mm kernel.
- competetive decoding, especially when compiled ; not as fast as batched_mm since it also can't use cuda graphs but beats grouped_mm kernel.

[very balanced !](https://media.tenor.com/uIXA-brBP9UAAAAe/thanos-perfectlybalanced.png)

fallback

vasqu: Neat! Always pro updating the docs then

PR #44061, opened 2026-02-16
Fix: flaky `Kosmos2ModelTest` test

tarekziade:
Reproduced locally with 

```
pytest -q -m generate --random-order-bucket=none --flake-finder --flake-runs=200 tests/models/kosmos2/test_modeling_kosmos2.py -k test_assisted_decoding_matches_greedy_search
```

Root cause:
- prepare_config_and_inputs_for_generate() uses helpers (ids_tensor, random_attention_mask) backed by global_rng.
- So each repetition used different random inputs. Some prompts pass, some fail in same assisted decoding.

Fix:
- The test now calls `gobal_rng()` before `prepare_config_and_inputs_for_generate()`


tarekziade:
run-slow: kosmos2

Rocketknight1:
Simpler fix: Why don't we just add `global_rng.seed(0)` to the `setUp` for `Kosmos2ModelTest`? It keeps the diff much smaller but should resolve this!

tarekziade:
run-slow: kosmos2

Rocketknight1 (approved):
LGTM!

PR #44069, opened 2026-02-17
Use torch.isfinite

cyyever:
# What does this PR do?

This PR uses ``torch.isfinite`` to simplify conditions, and the CUDA sync calls may also be reduced.


Rocketknight1 (approved):
Yep, LGTM! Thanks for the simplification

PR #44078, opened 2026-02-17
[`fix`] Set input_modalities on various architectures that aren't just text

tomaarsen:
# What does this PR do?

* Set `input_modalities` on various architectures that aren't just text

Sentence Transformers would like to rely on `input_modalities` in the future to determine what modalities can be used. However, it's not quite robust enough yet. This is a small step in the right direction with some architectures that I noticed failing with my tests. 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@zucchini-nlp

- Tom Aarsen

### DIFF
--- a/src/transformers/models/qwen3_omni_moe/modeling_qwen3_omni_moe.py
+++ b/src/transformers/models/qwen3_omni_moe/modeling_qwen3_omni_moe.py
@@ -2931,7 +2931,7 @@ class Qwen3OmniMoeTalkerDecoderLayer(GradientCheckpointingLayer):
 )
 class Qwen3OmniMoeTalkerModel(Qwen3OmniMoePreTrainedModel):
     config: Qwen3OmniMoeTextConfig
-    input_modalities = ("text",)
+    input_modalities = ("audio",)
     _no_split_modules = ["Qwen3OmniMoeTalkerDecoderLayer"]
     config_class = Qwen3OmniMoeTalkerTextConfig
     base_model_prefix = "talker.model"
--- a/src/transformers/models/qwen3_omni_moe/modeling_qwen3_omni_moe.py
+++ b/src/transformers/models/qwen3_omni_moe/modeling_qwen3_omni_moe.py
@@ -1040,6 +1040,7 @@ class Qwen3OmniMoeVisionPatchEmbed(nn.Module):
 
 class Qwen3OmniMoeVisionEncoder(Qwen3OmniMoePreTrainedModel):
     config: Qwen3OmniMoeVisionEncoderConfig
+    input_modalities = ("image", "video")
     _no_split_modules = ["Qwen3OmniMoeVisionBlock"]
     _can_record_outputs = {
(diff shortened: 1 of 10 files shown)


src/transformers/models/speech_to_text/modeling_speech_to_text.py
```diff
@@ -909,7 +910,7 @@ def forward(
         return_dict: bool | None = None,
         cache_position: torch.Tensor | None = None,
         **kwargs,
-    ) -> tuple[torch.FloatTensor] | Seq2SeqLMOutput:
+    ) -> tuple[torch.FloatTensor] | Seq2SeqModelOutput:
```
tomaarsen, line 913: Hmm, this fix slipped in there. Should I move it to a different PR? There's currently a mismatch between the type hint and actual return type that gets fixed by this line.

src/transformers/models/qwen3_vl_moe/modeling_qwen3_vl_moe.py
```diff
@@ -380,6 +380,7 @@ class Qwen3VLMoePreTrainedModel(PreTrainedModel):
         "hidden_states": Qwen3VLMoeTextDecoderLayer,
         "attentions": Qwen3VLMoeTextAttention,
     }
+    input_modalities = ("text", "image", "video")
```
tomaarsen, line 383: Sadly this was placed at the bottom after I updated the modular.

src/transformers/models/qwen3_omni_moe/modeling_qwen3_omni_moe.py
```diff
@@ -2929,6 +2931,7 @@ def forward(
 )
 class Qwen3OmniMoeTalkerModel(Qwen3OmniMoePreTrainedModel):
     config: Qwen3OmniMoeTextConfig
+    input_modalities = ("text",)
```
zucchini-nlp: nit: this is not really text, model expects 'audio' but encoded in discrete ids

src/transformers/models/speech_to_text/modeling_speech_to_text.py
zucchini-nlp: nah, it's fine

zucchini-nlp (approved):
Thanks, one tiny comment on Omni-model and we can merge

src/transformers/models/qwen3_omni_moe/modeling_qwen3_omni_moe.py
tomaarsen: Ah yes. This was set via the underlying `Qwen3VLMoeTextModel` via modular. I'll set it to `input_modalities=("audio",)` for `Qwen3OmniMoeTalkerModel`.

tomaarsen: Done in 47345d316a8785160bf653ccd638e5e33852fbd5

tomaarsen:
I think this should be ready (for final review)!

- Tom Aarsen

zucchini-nlp (approved):
Nice!

PR #44081, opened 2026-02-17
add default flash impl

ArthurZucker:
# What does this PR do?
Fixes #42533 by introducing default flash implementations. 

cc @vasqu and @cyrilvallez

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -1145,6 +1145,7 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
     _supports_sdpa: bool = False
     _supports_flash_attn: bool = False
     _supports_flex_attn: bool = False
+    _default_flash_implementation: str | None = None  # Model's preferred flash kernel (e.g., "kernels-community/flash-mla")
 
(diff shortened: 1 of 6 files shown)


Cyrilvallez (approved):
Makes a lot of sense for models with special requirements!

src/transformers/modeling_utils.py
```diff
@@ -1785,6 +1788,20 @@ def _check_and_adjust_attn_implementation(
...
+            base_impl = attn_implementation.removeprefix("paged|")
+
+            if base_impl in ("flash_attention_2", "flash_attention_3"):
+                default_flash = getattr(self, "_default_flash_implementation", None)
+                if default_flash is not None:
+                    logger.warning_once(
+                        f"This model is optimized for '{default_flash}'. "
+                        f"Automatically using '{default_flash}' instead of '{attn_implementation}'."
+                    )
+                    attn_implementation = f"paged|{default_flash}" if is_paged else default_flash
```
vasqu, line 1803: Sorry a bit late but wdyt about changing this to something along `_compatible_flash_implementations` (and making it a list)? I'm expecting that at some point other FA implementations will also work e.g. sinks for FA4 + we should only warn if we encounter that the requested FA is not one of the compatible ones

PR #44082, opened 2026-02-17
Fix patchtsmixer call to post_init

### LINKED ISSUE #44077: `patchtsmixer` has optional `post_init`, should no longer be allowed
### System Info

- `transformers` version: 5.2.0.dev0
- Platform: Windows-10-10.0.26200-SP0
- Python version: 3.11.13
- Huggingface_hub version: 1.3.1
- Safetensors version: 0.6.2
- Accelerate version: 1.11.0
- PyTorch version (accelerator?): 2.9.0+cu126 (CUDA)
- Using GPU in script?: no
- GPU type: NVIDIA GeForce RTX 3090

### Who can help?

@Cyrilvallez 

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

The `patchtsmixer` architecture has a `post_init` argument in its configuration, which when set to `False` (which is the default!), will not call `post_init()`. Since https://github.com/huggingface/transformers/pull/42873 by @Cyrilvallez, we should not be doing this anymore.

```python
from transformers import AutoModel, AutoConfig

model_name = "hf-internal-testing/tiny-random-PatchTSMixerModel"
config = AutoConfig.from_pretrained(model_name)
print(f"{config.post_init=}")
# config.post_init=False

model = AutoModel.from_pretrained(model_name)
# AttributeError: 'PatchTSMixerModel' object has no attribute 'all_tied_weights_keys'. Did you mean: '_tied_weights_keys'?
```

### Expected behavior

I would expect the `post_init` argument to be deprecated, and for `post_init` to no longer be optional. A somewhat related PR is https://github.com/huggingface/transformers/pull/42497, which asks whether the `post_init` weight initialization could be optional: I'm not sure when someone would want this `post_init` to ever be False.

- Tom Aarsen

Cyrilvallez:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/44077.
Indeed, the call is not optional. This is slightly breaking as the defaut used to be False, so fresh model instantiation will now use a different init (the one described in `_init_weights`)

### DIFF
--- a/src/transformers/models/patchtsmixer/configuration_patchtsmixer.py
+++ b/src/transformers/models/patchtsmixer/configuration_patchtsmixer.py
@@ -82,9 +82,6 @@ class PatchTSMixerConfig(PreTrainedConfig):
             error "mse".
         init_std (`float`, *optional*, defaults to 0.02):
             The standard deviation of the truncated normal weight initialization distribution.
-        post_init (`bool`, *optional*, defaults to `False`):
-            Whether to use custom weight initialization from `transformers` library, or the default initialization in
-            `PyTorch`. Setting it to `False` performs `PyTorch` weight initialization.
         norm_eps (`float`, *optional*, defaults to 1e-05):
             A value added to the denominator for numerical stability of normalization.
         mask_type (`str`, *optional*, defaults to `"random"`):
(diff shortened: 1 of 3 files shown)


ArthurZucker (approved):
were there any issues with previous?

Cyrilvallez:
Yes, see the linked issue. The pretrained model cannot be loaded correctly as the default is False!

PR #44083, opened 2026-02-17
FSDP2 native support in transformers 

3outeille:
cf https://github.com/huggingface/transformers/pull/45028/changes

---
This PR introduces **first-class FSDP2 (Fully Sharded Data Parallel v2) support** directly in Transformers, bypassing the need for Accelerate's FSDP wrapper. It covers the full lifecycle: model distribution, training, checkpointing, and CI testing across dozens of models.

A **standalone script for usage and expected throughput** will be available at https://github.com/huggingface/distributed-training-cookbook

```
TODO: add example train_fsdp.py in the PR
```

The correctness has been tested to match the Torchtitan implementation (cf https://github.com/huggingface/torchtitan/blob/sanity-check-fsdp/torchtitan/experiments/debug_fsdp/README.md) 

image

NOTE: Transformers modeling do take more memory for some reasons. To investigate later

### 1. Native FSDP2 Integration (`src/transformers/integrations/fsdp.py`)

The core addition is a new FSDP integration module that provides:

- **`initialize_fsdp()`** -- Sets up the `DeviceMesh` and process group for FSDP2 (requires PyTorch >= 2.5). Handles automatic backend detection (NCCL, GLOO, XCCL, etc.) and device assignment.

- **`apply_fsdp2()` with two modes:**
  - **Auto mode** (`{"mode": "auto"}`) -- Automatically discovers transformer block classes (DecoderLayer, EncoderLayer, etc.), shards input embeddings, all transformer blocks, and groups the final norm + output head together. Supports optional `cpu_offload` and `mixed_precision` policies.
  - **Manual mode** (`{"mode": "manual", "modules": {...}}`) -- Lets users specify exactly which modules to shard, with per-module options like `"free_full_weight"`, `"keep_full_weight"`, `"cpu_offload"`, and `"mixed_precision"`.

- **Smart block detection** (`get_transformer_block_classes()`) -- Finds transformer block classes by name pattern and filters out nested blocks (e.g., MoeBlock inside DecoderLayer) to only FSDP-wrap the outermost ones. This enables **MoE model support**.

- **Tied weight handling** -- Properly handles weight tying (e.g., `lm_head.weight` == `embed_tokens.weight`) by grouping tied modules and re-tying after `fully_shard` replaces parameters with DTensors.

### 2. FSDP2-Aware Save/Load via DCP + Safetensors

- **`save_fsdp_model()`** -- Saves FSDP2 model weights using PyTorch's Distributed Checkpoint (DCP) with `HuggingFaceStorageWriter`, enabling parallel distributed save with automatic consolidation into standard HF-compatible safetensors files.

- **`save_pretrained()` integration** -- `PreTrainedModel.save_pretrained()` now detects FSDP2 models (`_is_fsdp_managed_module`) and automatically routes to the DCP save path.

- **`from_pretrained()` integration** -- Accepts new `fsdp_plan` and `fsdp_device_mesh` kwargs. After loading weights, it applies FSDP2 distribution via `distribute_fsdp_model()`.

### 3. Comprehensive FSDP Test Suite (`tests/test_fsdp_mixin.py`)

A new `FSDPTesterMixin` class is added to the standard test infrastructure, automatically inherited by all `CausalLMModelTest` classes. It includes **7 batched subtests** per model, all run on CPU through `gloo` backend + `mp.spawn`: 

| Test | What it validates |
|---|---|
| `sharding_structure_untied/tied` | Correct FSDP wrapping targets match expectations |
| `auto_plan_vs_ddp` (untied/tied) | FSDP2 auto mode produces identical losses, grad norms, and final weights as DDP |
| `manual_plan_vs_ddp` (untied/tied) | FSDP2 manual mode matches DDP |
| `save_load` | Save via `save_pretrained` + reload via `from_pretrained` produces bit-exact weights |

The tests also validate **checkpoint resumability**: train for N/2 steps, save checkpoint (model via DCP+safetensors, optimizer+RNG via distcp), load into a fresh model, continue training, and verify the full trace matches an uninterrupted DDP run.

### 4. CI Test for Broad Model Coverage

Two bash scripts run the FSDP mixin tests across many models in parallel:

- **Dense models**: Tests ~10 active models (GPT-2, Qwen3, Phi, Llama, ModernBERT-decoder, OLMo3, Phi3, Mistral, LFM2, Qwen3.5) out of 40 total, ranked by HuggingFace Hub downloads.

- **MoE models**: Tests ~10 active MoE models (GPT-OSS, GLM-MoE-DSA, Qwen3-MoE, GLM4-MoE-Lite, Qwen3.5-MoE, DeepSeek-V2, Qwen3-Next, Mixtral, Qwen2-MoE, PhiMoE) out of 24 total.


src/transformers/modeling_utils.py
```diff
@@ -3938,6 +3939,8 @@ def from_pretrained(
         gguf_file = kwargs.pop("gguf_file", None)
         tp_plan = kwargs.pop("tp_plan", None)
         tp_size = kwargs.pop("tp_size", None)
+        fsdp_plan = kwargs.pop("fsdp_plan", None)
+        fsdp_device_mesh = kwargs.pop("fsdp_device_mesh", None)
```
SunMarc: let's just reuse device_mesh no instead of reintroducing it just for fsdp no ?

tests/test_fsdp_mixin.py
```diff
@@ -0,0 +1,861 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
```
SunMarc, line 1: let's create a new folder called tests/training and put those there instead. It will be better I think

tests/fsdp/run_fsdp_mixin_dense_tests.sh
```diff
@@ -0,0 +1,554 @@
+#!/bin/bash
+
+# Script to run all FSDP mixin tests for dense models in parallel.
+# Work in tandem with a special test_fsdp_mixin.py that batches all 11 distributed tests in a single mp.spawn. (will not be committed)
+# Uses concurrency-limited dispatch: multiple models share GPU pairs since test models are tiny (~7 MiB).
```
SunMarc: even the fsdp folder we can move that there

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,512 @@
...
+
+    Returns:
+        The FSDP-distributed model.
+    """
+    if fsdp_plan is None or device_mesh is None:
+        return model
+
+    model = apply_fsdp2(model, device_mesh, fsdp_plan)
 
+    return model
```
SunMarc: let's use apply_fsdp2 instead of calling distribute_fsdp_model ?

src/transformers/modeling_utils.py
```diff
@@ -4142,6 +4145,12 @@ def from_pretrained(
         model.eval()  # Set model in evaluation mode to deactivate Dropout modules by default
         model.set_use_kernels(use_kernels, kernel_config)
 
+        # Apply FSDP2 if configured (must be after weight loading)
+        if fsdp_plan is not None:
+            fsdp_mesh = fsdp_device_mesh if fsdp_device_mesh is not None else device_mesh
+            if fsdp_mesh is not None:
```
SunMarc: hmmm then we should force the user to pass fsdp_mesh, otherwise nothing will be applied and this without any warning

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,512 @@
...
+            "mode": "manual",
+            "modules": {
+                "model.embed_tokens": ["free_full_weight"],
+                "model.layers.0.self_attn": ["free_full_weight", "cpu_offload", "mixed_precision"],
+                "model.layers.0.mlp": ["free_full_weight"],
+                "model.norm": ["keep_full_weight"],
+                "lm_head": ["keep_full_weight"],
+            },
+        }
+    """
```
SunMarc, line 394: maybe it could make sense to have a nice `dataclass` for `fsdp_plan` instead of a dict ?

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,512 @@
...
+    if not isinstance(modules, dict):
+        raise ValueError("Manual fsdp_plan must define a 'modules' dict.")
+    return modules
+
+def apply_fsdp2(
+    model,
+    device_mesh,
+    fsdp_plan: dict[str, Any] | None,
+):
+    """
```
SunMarc, line 368: It would be nice to check how it integrates well with trainer. We can pass `fsdp` and `fsdp_config` in training_args and we would have to do the mapping to create the correct `fsdp_plan` + we need to call `apply_fsdp2` instead of `prepare()` on the model.

SunMarc (commented):
Thanks, left a couple of comments !

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,567 @@
...
         return False
+    try:
+        from torch.distributed.fsdp import FullyShardedDataParallel
+    except ImportError:
+        return False
+    return isinstance(module, FullyShardedDataParallel) or getattr(module, "_is_fsdp_managed_module", False)
+
 
-    import torch.distributed.fsdp
+def is_fsdp_enabled():
```
ArthurZucker: we usually don't define is xxx available here! (import utils)

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,567 @@
...
+
+    Returns:
+        Tuple of (device_map, device_mesh, fsdp_size)
+    """
+    if not is_torch_available():
+        raise ImportError("PyTorch is required for FSDP support")
+
+    import torch
+    import torch.distributed as dist
+
```
ArthurZucker: we should import top level!

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,567 @@
...
+        return device_map, device_mesh, None
+
+    if not is_torch_greater_or_equal("2.5"):
+        raise OSError("FSDP2 is only supported for `torch>=2.5`.")
+
+    if device_mesh is None:
+        # Detect the accelerator on the machine
+        device_type = torch._C._get_accelerator().type
+        if device_type == "mps":
+            device_type = "cpu"  # fallback
```
ArthurZucker: DTensor supports mps tho no?

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,567 @@
...
+            current_device.set_device(int(os.environ["LOCAL_RANK"]))
+            index = current_device.current_device()
+            fsdp_device = torch.device(device_type, index)
+            device_map = fsdp_device
+        else:
+            fsdp_device = torch.device(device_type)
+            device_map = device_type or {}
+
+        fsdp_size = dist.get_world_size()
+        device_mesh = torch.distributed.init_device_mesh(fsdp_device.type, (fsdp_size,), mesh_dim_names=("dp_shard",))
```
ArthurZucker, line 102: can't we re-use the func we defined in `tensor_parallel` ?

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,567 @@
...
+    return device_map, device_mesh, fsdp_size
+
+
+def get_transformer_block_classes(model):
+    """
+    Identifies transformer block classes in a model for FSDP wrapping.
+    These are typically the repeated layers that benefit from FSDP sharding.
+
+    Returns a set of module classes that should be wrapped with fully_shard().
+    """
```
ArthurZucker, line 124: pretty sure you can just check if the layer is `GradientCheckpointingLayer` 😉 https://github.com/huggingface/transformers/blob/3491f116a0ec4def708c2fbc141ea08df16ba29b/src/transformers/modeling_layers.py#L34 we make all blocks inherit from this!

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,567 @@
...
+        class_name = module.__class__.__name__
+        # Use endswith to avoid false positives (e.g. "Layer" matching "LayerNorm")
+        for block_name in block_names:
+            if class_name.endswith(block_name):
+                block_classes.add(type(module))
+                break
+
+    # Filter out nested block classes (e.g. SparseMoeBlock inside DecoderLayer).
+    # We only want to FSDP-wrap the outermost block classes. If a class like
+    # MoeBlock only ever appears inside a DecoderLayer, we skip it.
```
ArthurZucker, line 146: using `GradientCheckpointingLayer` allows us to remove this

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,567 @@
...
+    if fsdp_plan.get("mixed_precision"):
+        policy_kwargs["mp_policy"] = MixedPrecisionPolicy(
+            param_dtype=torch.bfloat16,
+            reduce_dtype=torch.float32,
+            output_dtype=None,
         )
+    return policy_kwargs
+
+
+def _shard_auto_input_embedding(input_embed, is_weights_tied: bool, device_mesh, auto_policy_kwargs):
```
ArthurZucker:
```suggestion
def _auto_shard_input_embedding(input_embed, is_weights_tied: bool, device_mesh, auto_policy_kwargs):
```

ArthurZucker: idk sounds more natural

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,567 @@
...
+
+
+def _shard_auto_transformer_blocks(model, block_classes, device_mesh, auto_policy_kwargs):
+    for name, module in model.named_modules():
+        if type(module) in block_classes:
+            fully_shard(module, mesh=device_mesh, reshard_after_forward=True, **auto_policy_kwargs)
+            logger.debug(f"Applied fully_shard to {name} ({type(module).__name__})")
+
+
+def _find_final_norm(model, decoder_layer_names):
```
ArthurZucker, line 210:
```python
    base_model_pp_plan = {
        "embed_tokens": (["input_ids"], ["inputs_embeds"]),
        "layers": (["hidden_states", "attention_mask"], ["hidden_states"]),
        "norm": (["hidden_states"], ["hidden_states"]),
    }
```
this looks like something we can define in metadata / take from PP no?

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,567 @@
...
+    return final_norm
+
+
+def _get_auto_tail_modules(model, decoder_layer_names, input_embed, output_embed, is_weights_tied: bool) -> list:
+    # Group final norm + output head.
+    # NOTE(3outeille): Small optimization by forcing reshard_after_forward=False for the final norm and output head.
+    # Otherwise, that would mean reshard/freeing full params after the last forward and immediately re-all-gathering
+    # them in the backward pass, which is wasteful. Better to keep them gathered for reuse.
+    # Untied: [final_norm, lm_head]
+    # Tied:   [final_norm, embed_tokens] - embed_tokens.weight IS lm_head.weight.
```
ArthurZucker, line 237: can be taken from:
```
    _pp_plan = {"lm_head": (["hidden_states"], ["logits"])}
```
https://github.com/huggingface/transformers/blob/3491f116a0ec4def708c2fbc141ea08df16ba29b/src/transformers/models/llama/modeling_llama.py#L432

No?

src/transformers/integrations/fsdp.py
```diff
@@ -15,46 +15,567 @@
...
+                f"{item!r}. Supported: 'free_full_weight', 'keep_full_weight', 'cpu_offload', 'mixed_precision'."
+            )
+
+    if strategy is None:
+        strategy = "free_full_weight"
+
+    return strategy != "keep_full_weight", mp_policy, offload_policy
+
+
+def _iter_manual_plan_targets(model, pattern, name_to_module, already_sharded_names):
```
ArthurZucker, line 314: can you document what this does? 
not entirely sure we need it, but if we do, probably something TP plan or PP plan or Dtype plan are gonna be using ?

src/transformers/integrations/fsdp.py
ArthurZucker: Yep, I think the idea was to stay simple like the pp / tp plan. But for fsdp we might want more control.

ArthurZucker (commented):
Will look at the tests next time but my main comment:

You are not integrated with the `core-model-loading` API noo?

My understanding is that you should shard the weights exactly the same way we do for TP, maybe push further as you want to shard ALL layers, then when running the forward, each process materialize the weights locally (GP0-8 end up with the same full tensor) and discard it keeping only its slice. 

We need 2 points:
1. TENSOR_PARALLEL_LAYERS integration in `core_model_loading` needs to support `fsdp`. This is what's gonna be responsible for loading the weights
2. distribute_module, which needs tto happen before the load, and is responsible for attaching tthe appropriate hooks for fsdp2. 
THis also needs to be explained like the above comment: what is fsdp? -> 1 sharding plan 2. hook plan. 

Now if you have to rely on DTensor, you might need to change set_param? or you just apply dtensor conversion post all the loading. 

The most important is to test say MIxtral with the dynamic weight loader. The way I see it you'll load all weights on all layers, then shard (discard) some of them.

kashif:
Based on my CP experiments:
1.  Accept `fsdp_plan="auto"` string shorthand:  Currently `fsdp_plan={"mode": "auto"}` works but `fsdp_plan="auto"` raises a confusing error. Since `tp_plan="auto"` is accepted as a string, `fsdp_plan` should be consistent:
```py
  # _parse_fsdp_plan_mode could normalize strings:
  if isinstance(fsdp_plan, str):
      fsdp_plan = {"mode": fsdp_plan}
```

2. Document CP + FSDP2 device mesh pattern:  When using Context Parallelism with native FSDP2, the `fsdp_device_mesh` must be the flattened `dp_cp` mesh, not just the `dp` submesh otherwise FSDP2 doesn't shard parameters across CP ranks and  you OOM. This is non-obvious:
```py
  # Users need to do this:
  world_mesh = init_device_mesh("cuda", (dp_size, cp_size), mesh_dim_names=("dp", "cp"))
  fsdp_mesh = world_mesh["dp", "cp"]._flatten(mesh_dim_name="dp_cp")
  model = AutoModelForCausalLM.from_pretrained(..., fsdp_device_mesh=fsdp_mesh, fsdp_plan={"mode": "auto"})
``` 
The PR's docstring or the `3D_parallel.py` example should show this pattern.

3. Add a CP example using native FSDP2:  The current `3D_parallel.py` uses FSDP1 (`FullyShardedDataParallel` with `NO_SHARD`). An example showing native FSDP2 (`fsdp_plan`) + CP would be good to have.

4. Note about CP rotation method:  The default CP rotation (`allgather`) requires `allgather_into_tensor_coalesced` which may not be available in all PyTorch builds. The `alltoall` rotation works universally. Worth noting in the docs  or defaulting to `alltoall`.

src/transformers/integrations/fsdp.py
3outeille: im not sure but better not fallback and having people raise an error

PR #44090, opened 2026-02-17
Update post proc

itazap:
bos and eos behaviour should match when updating post processor

setting `add_bos_token=True` when `bos_token=None` should silently disables `add_bos_token`. (was already the behavior for `eos_token`)


ArthurZucker (approved):

ArthurZucker:
SGTM, erroring out is fine as well IMO but it does make sense setting to None -> disables

PR #44095, opened 2026-02-17
Fix loading logic issue

Cyrilvallez:
# What does this PR do?

As per the title. The check that was added in https://github.com/huggingface/transformers/pull/43768 is wrong, as a missing weight would NOT be reinitialized in some cases!

As for the pointers check, it is actually not needed at all since when modules (modules, not parameters) are shared (tied), loading only 1 param from the checkpoints will load both weights correctly as it calls `setattr` on the module which is the object shared already. So we simply have to remove them from missing keys, instead of an expensive (and sometimes wrong) check of the pointers.
This change fixes https://github.com/huggingface/transformers/issues/44060

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -2332,10 +2332,14 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         # This check is for remote code that does NOT use either `torch.init` or `transformers.initialization` in `_init_weights`
         # which allow to check the flag directly on param. As they don't and write the params in-place, params would be reinitialized
         # otherwise
-        if all(getattr(param, "_is_hf_initialized", False) for param in module.parameters(recurse=False)) and all(
-            getattr(buffer, "_is_hf_initialized", False)
-            for buffer in module.buffers(recurse=False)
-            if buffer is not None
+        if (
+            self.is_remote_code()
+            and all(getattr(param, "_is_hf_initialized", False) for param in module.parameters(recurse=False))
+            and all(
+                getattr(buffer, "_is_hf_initialized", False)
+                for buffer in module.buffers(recurse=False)
+                if buffer is not None
+            )
         ):
             module._is_hf_initialized = True
             return
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -4230,9 +4230,6 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
         missing keys from meta device to their expected device, reinitializing missing weights according to proper
         distributions, tying the weights and logging the loading report."""
         try:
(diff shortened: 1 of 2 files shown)


src/transformers/modeling_utils.py
```diff
@@ -2329,10 +2329,13 @@ def _initialize_weights(self, module):
...
-            (weight := getattr(module, "weight", None)) is not None
-            and getattr(weight, "_is_hf_initialized", False)
-            and not list(module.named_buffers())
+        # This check is for remote code that does NOT use either `torch.init` or `transformers.initialization` in `_init_weights`
+        # which allow to check the flag directly on param. As they don't and write the params in-place, params would be reinitialized
+        # otherwise
+        if all(getattr(param, "_is_hf_initialized", False) for param in module.parameters(recurse=False)) and all(
+            getattr(buffer, "_is_hf_initialized", False)
+            for buffer in module.buffers(recurse=False)
+            if buffer is not None
```
zucchini-nlp: nice, forgot that we can get bigger modules

zucchini-nlp (approved):
Thanks, that works much better tbh

foreverlms:
Have you tested the case that:
1. local model, which means we have to set `trust_remote_code` = True
2. tied_word_embeddings is set to True?

I found in this case, missing_keys will be cleared by this logic and the lm_head weight won't be tied anymore.
Please correct me if I was wrong @Cyrilvallez

src/transformers/modeling_utils.py
```diff
@@ -4583,6 +4558,23 @@ def mark_tied_weights_as_initialized(self):
...
+        # as they are inherently tied because the 2 modules are the same object. In this case, once we load a parameter
+        # inside one of the 2 modules, the other will also automatically be loaded and will have the `_is_hf_initialized`
+        # flag (because we call `setattr` with the loaded param on the module, which is the same object), but its counterpart
+        # will still appear as a missing key as we never get it out of the set (because it appears in the state_dict as well).
+        # So we remove it now - otherwise it's considered missing and will be wrongly reinitialized
+        # Note: this is never an issue in main Transformers, as we never do module-tying, only parameter-tying, and we know
+        # which params are supposed to be tied to which other params
+        if self.is_remote_code():
+            # Remove those that are already initialized, but appear as missing due to module tying
+            loading_info.missing_keys = {
```
foreverlms, line 4572: Have you tested the case that:

1. local model, which means we have to set trust_remote_code = True
2. tied_word_embeddings is set to True

I found in this case, missing_keys will be cleared by this logic and the lm_head weight won't be tied anymore. It seems under meta device and the model will output non-sense text.
Please correct me if I was wrong @Cyrilvallez

Cyrilvallez: This was fixed in https://github.com/huggingface/transformers/pull/44469!

foreverlms: Thanks! It's greate to see this is fixed.

PR #44097, opened 2026-02-17
Merge test_keep_in_fp32_modules and test_keep_in_fp32_modules_strict

Rocketknight1:
The `test_keep_in_fp32_modules` issues in #44052 are because the test assumes a model has **either** `_keep_in_fp32_modules` or `_keep_in_fp32_modules_strict` **but not both.**

The only model that uses both is `glm_moe_dsa`, so this is the only model that fails the tests. This PR refactors things and merges both tests into a single `test_keep_in_fp32_modules` test. This test checks that:

1) When we load in `fp16`, the layers in `_keep_in_fp32_modules` and `_keep_in_fp32_modules_strict` are upcasted, and no other layers
2) When we load in `bf16`, the layers in `_keep_in_fp32_modules_strict` are upcasted, and no other layes.
3) There is no overlap between `_keep_in_fp32_modules` and `_keep_in_fp32_modules_strict` (because allowing this just creates weird edge cases, and there's no reason to)

This was always the intended behaviour, and now the test covers it correctly!

Fixes #44052, cc @arthurzucker

### DIFF
--- a/tests/test_modeling_common.py
+++ b/tests/test_modeling_common.py
@@ -928,6 +928,10 @@ class ModelTesterMixin:
                     model.save_pretrained(tmpdirname)
 
                     model = model_class.from_pretrained(tmpdirname, dtype=torch.float16)
+                    self.assertFalse(
+                        model._keep_in_fp32_modules & model._keep_in_fp32_modules_strict,
+                        "We found a layer in both the `_keep_in_fp32_modules` and `_keep_in_fp32_modules_strict` lists. Please remove it from one of the two lists.",
+                    )
                     # When reloading in fp16, keep_in_fp32_modules AND keep_in_fp32_modules_strict should be upcasted
                     all_fp32_modules = model._keep_in_fp32_modules | model._keep_in_fp32_modules_strict
                     for name, param in model.state_dict().items():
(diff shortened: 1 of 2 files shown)


Rocketknight1:
run-slow: glm_moe_dsa

tests/test_modeling_common.py
```diff
@@ -914,61 +914,39 @@ def test_keep_in_fp32_modules_exist(self):
...
-        config, _ = self.model_tester.prepare_config_and_inputs_for_common()
-        for model_class in self.all_model_classes:
-            with self.subTest(model_class.__name__):
-                model = model_class(copy.deepcopy(config))
-                if len(model._keep_in_fp32_modules_strict) == 0:
-                    self.skipTest(
-                        reason=f"{model_class.__name__} class has no _keep_in_fp32_modules_strict attribute defined"
+                    self.assertFalse(
+                        model._keep_in_fp32_modules & model._keep_in_fp32_modules_strict,
+                        "We found a layer in both the `_keep_in_fp32_modules` and `_keep_in_fp32_modules_strict` lists. Please remove it from one of the two lists.",
```
ArthurZucker, line 933: both can have a different path I think (absolute / relative?) but its fine

ArthurZucker (approved):
ty!

tests/test_modeling_common.py
Rocketknight1: Should be okay for now, since it's rare to use both anyway! We can fix it later if it becomes a problem

PR #43980, opened 2026-02-13
Escape % in help for ArgumentParser.add_argument to fix TypeError

albertvillanova:
# What does this PR do?

Escape `%` in `help` for `ArgumentParser.add_argument` to fix TypeError:
> TypeError: not enough arguments for format string

Context: https://docs.python.org/3/library/argparse.html#help
> As the help string supports %-formatting, if you want a literal % to appear in the help string, you must escape it as %%.

### Code to reproduce the bug
```python
from transformers import HfArgumentParser, TrainingArguments

parser = HfArgumentParser(TrainingArguments)
parser.parse_args(['--help'])
```

### Additional context
Related issue in `trl`:
- https://github.com/huggingface/trl/issues/5099

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#create-a-pull-request),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #44100, opened 2026-02-17
Fix percentage formatting in help messages for gradient checkpointing, Liger Kernel, and empty cache steps

qgallouedec:
In https://github.com/huggingface/trl/pull/5112 a user reported that `trl sft --help` fails

It's because three inherited args from `TrainingArguments` (`torch_empty_cache_steps`, `gradient_checkpointing` and `use_liger_kernel`)help strings include bare `%` which break argparse formatting on Python 3.13. This PR fixes it.

### DIFF
--- a/src/transformers/training_args.py
+++ b/src/transformers/training_args.py
@@ -896,7 +896,7 @@ class TrainingArguments:
     gradient_checkpointing: bool = field(
         default=False,
         metadata={
-            "help": "Enable gradient checkpointing to trade compute for memory. Reduces memory at the cost of ~20% slower training."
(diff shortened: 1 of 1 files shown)


stevhliu (approved):
thanks for fixing! this should also close #43980

qgallouedec:
ah right sorry for the duplicate

albertvillanova:
Thanks, glad to see this has been already fixed!

For future reference, in most OSS workflows the usual practice is to keep the earlier PR open and mark the later one as a duplicate before closing it. That helps preserve the original discussion context and avoids losing review history on the first contribution.

Not a big deal at all, just mentioning it for consistency going forward.

PR #44119, opened 2026-02-17
fix: handle None value from TOKENIZER_MAPPING_NAMES.get() in AutoTokenizer

### LINKED ISSUE #44117: TOKENIZER_MAPPING_NAMES sometimes returns None, but from_pretrained assumes otherwise
### System Info

When loading a tokenizer with AutoTokenizer (src/transformers/models/auto/tokenization_auto.py), on L652 in from_pretained the code tries to remove "Fast" from the tokenizer mapping for old-style models:
```
if (
            tokenizer_auto_map is None
            and tokenizer_config_class is not None
            and config_model_type is not None
            and config_model_type != ""
            # Here:
            and TOKENIZER_MAPPING_NAMES.get(config_model_type, "").replace("Fast", "")
            != tokenizer_config_class.replace("Fast", "")
        ):
```
Permalink: https://github.com/huggingface/transformers/blob/ffbea2db927e4a0fb4bc8615c909ae277d8df0e6/src/transformers/models/auto/tokenization_auto.py#L652

However, sometimes ` TOKENIZER_MAPPING_NAMES.get(config_model_type, "")` returns `None`, instead of an empty string (if the config_model_type is in the dict, but has value None), leading to models failing to load with an attribute error
```
AttributeError: 'NoneType' object has no attribute 'replace'
```

This could be fixed as follows:
```
and (TOKENIZER_MAPPING_NAMES.get(config_model_type, "") or "")
```

Or by some other reasonable check.

### Who can help?

Looks like this was introduced in #42894 - @ArthurZucker  and @itazap 

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

Run:

```
processor = AutoProcessor.from_pretrained('google/siglip2-so400m-patch14-384')
```

### Expected behavior

It shouldn't throw an attribute error :)

tirth8205:
Fixes #44117

`TOKENIZER_MAPPING_NAMES.get(config_model_type, "")` returns `None` when the key exists with value `None`, causing `AttributeError: 'NoneType' object has no attribute 'replace'` when loading models like `google/siglip2-so400m-patch14-384`.

Changed `.get(key, "")` to `(.get(key) or "")` to handle both missing keys and `None` values.

### DIFF
--- a/src/transformers/models/auto/tokenization_auto.py
+++ b/src/transformers/models/auto/tokenization_auto.py
@@ -649,7 +649,7 @@ class AutoTokenizer:
             and tokenizer_config_class is not None
             and config_model_type is not None
             and config_model_type != ""
-            and TOKENIZER_MAPPING_NAMES.get(config_model_type, "").replace("Fast", "")
+            and (TOKENIZER_MAPPING_NAMES.get(config_model_type) or "").replace("Fast", "")
             != tokenizer_config_class.replace("Fast", "")
         ):
             # new model, but we ignore it unless the model type is the same


Rocketknight1:
Fixed already!

PR #37905, opened 2025-05-01
Break weight tying when quantizing input embedding

jerryzh168:
Summary:
Currently when we try to quantize input_embedding for some models, the output embedding (lm_head) will also be quantized the same way, since they are tied, and this may not be what we want. To break the tie, we added the option to allow people to
1. load unquantized weight
2. tie weights
3. quantize

so that the tie will be broken

Test Plan:
```
from transformers import (
  AutoModelForCausalLM,
  AutoProcessor,
  AutoTokenizer,
  TorchAoConfig,
)
from torchao.quantization.quant_api import (
    IntxWeightOnlyConfig,
    Int8DynamicActivationIntxWeightConfig,
    AOPerModuleConfig
)
from torchao.quantization.granularity import PerGroup, PerAxis
import torch

model_id = "microsoft/Phi-4-mini-instruct"

embedding_config = IntxWeightOnlyConfig(
    weight_dtype=torch.int8,
    granularity=PerAxis(0),
)
linear_config = Int8DynamicActivationIntxWeightConfig(
    weight_dtype=torch.int4,
    weight_granularity=PerGroup(32),
    weight_scale_dtype=torch.bfloat16,
)
quant_config = AOPerModuleConfig({"_default": linear_config, "model.embed_tokens": embedding_config})
quantization_config = TorchAoConfig(quant_type=quant_config, include_embedding=True, untie_embedding_weights=True)
quantized_model = AutoModelForCausalLM.from_pretrained(model_id, torch_dtype=torch.float32, device_map="auto", quantization_config=quantization_config)
tokenizer = AutoTokenizer.from_pretrained(model_id)

print(quantized_model)
print("embed_tokens.weight:", quantized_model.model.embed_tokens.weight)
print("lm head weight:", quantized_model.lm_head.weight)
from transformers.modeling_utils import find_tied_parameters
print(find_tied_parameters(quantized_model))

# Manual Testing
prompt = "Hey, are you conscious? Can you talk to me?"
messages = [
    {
        "role": "system",
        "content": "",
    },
    {"role": "user", "content": prompt},
]
templated_prompt = tokenizer.apply_chat_template(
    messages,
    tokenize=False,
    add_generation_prompt=True,
)
print("Prompt:", prompt)
print("Templated prompt:", templated_prompt)
inputs = tokenizer(
    templated_prompt,
    return_tensors="pt",
).to("cuda")
generated_ids = quantized_model.generate(**inputs, max_new_tokens=128)
[...]

PR #37802, opened 2025-04-26
Support `AOPerModuleConfig` and `include_embedding`

jerryzh168:
Summary:
This PR adds support per module configuration for torchao Also added per module quantization examples:

1. Quantizing different layers with different quantization configs
2. Skip quantization for certain layers

Test Plan:
python tests/quantization/torchao_integration/test_torchao.py -k test_include_embedding python tests/quantization/torchao_integration/test_torchao.py -k test_per_module_config_skip

Reviewers:

Subscribers:

Tasks:

Tags:

PR #44122, opened 2026-02-18
Add BnB 4-bit embedding quantization support

### LINKED ISSUE #31474: Quantization support for heads and embeddings
### Feature request

Hi! I’ve been researching LLM quantization recently ([this paper](https://arxiv.org/abs/2405.14852)), and noticed a potentially improtant issue that arises when using LLMs with 1-2 bit quantization.

### Problem description :mag:

Transformers supports several great ways for quantizing transformer ‘body’, but it seems that there is no built-in way
to quantize embeddings and/or lm head.

The reason why this is important is that some of the recent LLMs have very large vocabularies, and as a result,
their embeddings and heads can get massive. For instance, [Llama 3](https://huggingface.co/meta-llama/Meta-Llama-3-8B)
has 128K token vocabulary, [Qwen 2](https://huggingface.co/Qwen/Qwen2-72B-Instruct) has over 150K, [Gemma 2b](https://huggingface.co/google/gemma-2b) has 256K

As a result, if you load NF4 or AQLM quantized models, their **embeddings can take up 50% or more of the model footprint**. 
This is even more critical for lower bitwidth quantization:

### Feature Request :rocket:

It would be great if transformers had a flag to quantize embeddings and heads using some of the existing quantization methods. One simple way would be to use LLM.int8 or NF4 by Tim Dettmers since transformers already supports this.

I’ve investigated how quantizing embeddings with these methods affects common models. Below is model perplexity for [Llama 3 8B using AQLM+PV 2-bit quantization](https://huggingface.co/ISTA-DASLab/Meta-Llama-3-8B-AQLM-PV-2Bit-1x16). I measured three configurations: fp16 embeddings, int8 embeddings and NF4 embeddings with the same parameters that transformers uses for linear layers.

The values represent perplexity on [WikiText-2 test set](https://huggingface.co/datasets/Salesforce/wikitext/viewer/wikitext-2-v1)
measured with the same protocol used in [GPTQ](https://arxiv.org/abs/2210.17323) / [AQLM](https://arxiv.org/abs/2401.06118)
/ [QuIP#](https://arxiv.org/pdf/2402.04396.pdf) papers.
The code for these measurements can be found [here](https://gist.github.com/galqiwi/cb896f39052d1f4f718cb772040f3088).

Overall, 8-bit compression looks nearly lossless, the increase in perplexity does not exceed the error you get when 
quantizing the transformer with the same LLM int8 codec. In turn, NF4 introduces some error (within 0.05 for Llama 3),
but I would argue that this trade-off makes sense for low memory applications. Also, embeddings appear
easier to quantize than heads.

### Implementation details :gear:

There are multiple obstacles on the way to implementing this feature:
#### No support for mixed quantization
Currently, transformers does not support quantizing with multiple `HfQuantizer`s. IMO this is a good behaviour, as interactions between different quantizators can be messy. The problem is that this feature requires for transformers library to use different compression methods for body and heads/embeddings. I think that can be solved by extending `HfQuantizer` interface by adding embedding/head quantization methods and adding new `[embed,head]_quantization_config` arguments to `QuantizationConfigMixin` or something in this area.
#### No support for embedding quantization in bitsandbytes
As far as I know, no quantization method supports `nn.Embedding`-like interface. I can ask bitsandbytes maintainers if they would accept a PR that fixes that.

Also, there is a caveat that some models use tied embeddings/heads, while implementing, one need to be mindful of them.

### Cool things that this can enable :trophy:

If we can implement 4-bit embeddings, it will be possible to write a colab notebook that runs [Llama 3 70B model](https://huggingface.co/meta-llama/Meta-Llama-3-70B)
on the free tier T4 GPU without offoading, by combining embedding/heads quantization and the  PV-tuned model 
https://huggingface.co/ISTA-DASLab/Meta-Llama-3-70B-AQLM-PV-1Bit-1x16 .

Another use case is running quantized LLMs on smartphones or embedded devices: for instance, the [gemma-2b](https://huggingface.co/google/gemma-2b) can fit into 1GB RAM, but only if you quantize embeddings/heads in addition to transformer weights.

If you’re interested in making a demo out of this, I’d be excited to implement this with your review / recommendations if you prefer, or wait for you to implement it your way.

What do you think?

### Motivation

We are faced with a new bottleneck in model quantization. I think we can manage to fix it

### Your contribution

I can allocate my time to submitting PR, but we need to figure out what to do first

adityuhkapoor:
# What does this PR do?

  Adds 4-bit embedding quantization for BitsAndBytes, mirroring TorchAO's existing
  `include_input_output_embeddings` and `untie_embedding_weights` pattern (PRs #37802, #37905, #37935).

  Large-vocabulary models (Llama 3 128K, Qwen 2 150K, Gemma 256K) have embeddings that can be 50%+ of a
   quantized model's footprint. BnB quantizes `nn.Linear` layers but skips `nn.Embedding` entirely.
  `bnb.nn.Embedding4bit` exists upstream but transformers doesn't use it. Builds on @galqiwi's upstream work adding bnb.nn.Embedding4bit to bitsandbytes(bitsandbytes-foundation/bitsandbytes#1292).

  This PR adds two flags to `BitsAndBytesConfig`:
  - `include_input_output_embeddings` — replaces `nn.Embedding` with `bnb.nn.Embedding4bit`
  - `untie_embedding_weights` — unties embed_tokens ↔ lm_head before quantization (needed for models
  with tied weights like Llama)

  4-bit only — not supported for 8-bit quantization. Serialization is not included (aligned with
  @SunMarc's guidance on the issue).

  Also fixes a pre-existing `NameError` in `dequantize_and_replace` where `has_been_replaced` was used
  without initialization.

  Fixes #31474

  ## Before submitting
  - [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the
  case).
  - [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/
  CONTRIBUTING.md#create-a-pull-request), Pull Request section?
  - [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)?
   Please add a link to it if that's the case. https://github.com/huggingface/transformers/issues/31474

  ## Who can review?

  @SunMarc @MekkCyber — quantization

### DIFF
--- a/src/transformers/integrations/bitsandbytes.py
+++ b/src/transformers/integrations/bitsandbytes.py
@@ -47,6 +47,15 @@ class Bnb4bitQuantize(ConversionOps):
         # update param name to get the weights instead of the quantized stats
         module, _ = get_module_from_name(model, full_layer_name)
 
+        # Handle untying of embedding weights before quantization
+        input_embed = model.get_input_embeddings()
+        is_embedding_param = id(module) == id(input_embed)
+        untie = getattr(self.hf_quantizer.quantization_config, "untie_embedding_weights", False)
+
+        if untie and is_embedding_param:
+            setattr(model.config.get_text_config(decoder=True), "tie_word_embeddings", False)
+            lm_head = value.clone()
+
         # Support models using `Conv1D` in place of `nn.Linear` (e.g. openai-community/gpt2) by transposing the weight matrix prior to quantization.
         # Since weights are saved in the correct "orientation", we skip transposing when loading.
         if issubclass(module.source_cls, Conv1D):
@@ -55,7 +64,10 @@ class Bnb4bitQuantize(ConversionOps):
         old_value = model.get_parameter_or_buffer(full_layer_name)
         new_value = bnb.nn.Params4bit(value, requires_grad=False, **old_value.__dict__).to(value.device)
         module._is_hf_initialized = True
-        return {full_layer_name: new_value}
+        result = {full_layer_name: new_value}
+        if untie and is_embedding_param:
+            result["lm_head.weight"] = lm_head
+        return result
 
 
 class Bnb4bitDeserialize(ConversionOps):
@@ -221,6 +233,21 @@ def replace_with_bnb_linear(
                     new_module.requires_grad_(False)
                     model.set_submodule(module_name, new_module)
                     has_been_replaced = True
+            elif (
+                isinstance(module, nn.Embedding)
+                and quantization_config.quantization_method() != "llm_int8"
+                and getattr(quantization_config, "include_input_output_embeddings", False)
+            ):
+                new_module = bnb.nn.Embedding4bit(
+                    module.num_embeddings,
+                    module.embedding_dim,
+                    quant_type=quantization_config.bnb_4bit_quant_type,
+                    quant_storage=quantization_config.bnb_4bit_quant_storage,
+                )
+                new_module.source_cls = type(module)
+                new_module.requires_grad_(False)
+                model.set_submodule(module_name, new_module)
+                has_been_replaced = True
 
     if not has_been_replaced:
         logger.warning(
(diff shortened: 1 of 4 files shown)


Rocketknight1:
please do not tell your code agent to "find issues and fix them" with no further instruction because no-one had posted on that one in a year and a half

PR #44123, opened 2026-02-18
Avoid device sync in training loss accumulation

cyyever:
# What does this PR do?
This PR avoids device sync in training loss accumulation by ```torch.where```. The `is_torch_xla_available` condition is also removed.


Rocketknight1:
Hey @cyyever! We did some internal discussion about CUDA sync PRs - would you be able to make a benchmark or profile graph showing the reduction in syncs? We might make those mandatory in future just because it can sometimes be hard to tell from reading the code where exactly syncs would occur

PR #44125, opened 2026-02-18
Raise informative error when loading video processors

### LINKED ISSUE #43986: Confusing crash when loading a video model through AutoProcessor without torchvision installed

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/43986

### DIFF
--- a/src/transformers/models/auto/video_processing_auto.py
+++ b/src/transformers/models/auto/video_processing_auto.py
@@ -94,8 +94,8 @@ VIDEO_PROCESSOR_MAPPING = _LazyAutoMapping(CONFIG_MAPPING_NAMES, VIDEO_PROCESSOR
 
 
 def video_processor_class_from_name(class_name: str):
-    for module_name, extractors in VIDEO_PROCESSOR_MAPPING_NAMES.items():
-        if class_name in extractors:
+    for module_name, extractor in VIDEO_PROCESSOR_MAPPING_NAMES.items():
+        if class_name == extractor:
             module_name = model_type_to_module_name(module_name)
 
             module = importlib.import_module(f".{module_name}", "transformers.models")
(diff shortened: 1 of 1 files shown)


src/transformers/models/auto/video_processing_auto.py
```diff
@@ -360,6 +360,10 @@ def from_pretrained(cls, pretrained_model_name_or_path, *inputs, **kwargs):
 
         if video_processor_class is not None:
             video_processor_class = video_processor_class_from_name(video_processor_class)
+        elif not is_torchvision_available():
+            raise ValueError(
+                f"{pretrained_model_name_or_path} requires `torchvision` to be installed. Please install `torchvision` and try again."
+            )
```
molbap: sounds good, maybe it happens too early though? like a remote processor has only `auto_map` it might break it, maybe after the has_remote_code path is better?

molbap (approved):
Sounds ok to me! just a question about the ordering of checks

src/transformers/models/auto/video_processing_auto.py
zucchini-nlp: oh right, good catch

PR #44127, opened 2026-02-18
AutoTokenizer ignores config when model_type is None

itazap:
when the model_type isn't in `TOKENIZER_MAPPING_NAMES` (ex. "llama"), `TOKENIZER_MAPPING_NAMES.get("llama", "")` -->  "". Then we compare `"" != "LlamaTokenizer"`  (the `tokenizer_class` in `tokenizer_config.json`). Since that's true we early exit and fallback to `TokenizerBackend`


ArthurZucker (approved):
ty

itazap:
run-slow: auto

PR #44130, opened 2026-02-18
[generate] Completely stop relying on `cache_position` to prepare inputs

Cyrilvallez:
# What does this PR do?

As per the title.

This PR is the first big step towards removing the `cache_position` everywhere, as they are not needed in general and everything can be inferred from the cache itself.
The major changes are the following:

- `prepare_inputs_for_generation` now receives the already sliced `input_ids` or `inputs_embeds` according to the cache -> this sequence length is absolute and everything else is sliced based on it. The idea is that in every generation loop, we always know how many tokens we need to keep for next step with cache (usually a single one), so let's use this information instead of very convoluted logic. Currently, it takes as input the full sequence, then relies on `cache_position` to slice but this is extremely brittle versus taking the correct new input all the time as immediate input

- align `position_ids` and `cache_position` to contain the values from all past tokens. Then they are only sliced to input length in `prepare_inputs_for_generation`. This was done because the `attention_mask` needs to contain the values of all tokens all the time (because we have padding), so it makes more sense to align the behavior of all important common inputs (`position_ids`, `cache_position`, `attention_mask`), so that we don't have to always remember what is already sliced and was is not) -- fixes some broken logic for that in `_update_model_kwargs_for_generation` as well

- Revert most of https://github.com/huggingface/transformers/pull/42088 as it's not needed to avoid running prefill for assistants - prefill can simply take a kwarg `is_first_iteration` to know if it's the assistant calling it several time. This is needed as when restarting from cache, we have to pass all the preceding `input_ids` as well and `_prefill` is now responsible to check the cache and slice the inputs to prefill with only the new tokens. We cannot easily slice before calling `_prefill`, as we potentially need to recreate the full `attention_mask` and `positions_ids` based on the FULL SEQUENCE (including preceding tokens)

## Next steps

Now that input preparation is decorrelated from the `cache_position`, next PR will decorrelate both the cache update and the masking logic, and `cache_position` will not be useful anymore.

Note that currently, `cache_position` are not used anymore for general input preparation, but are still created and forwarded to the `forward`s for downstream use in the cache and masking.

cc @zucchini-nlp @vasqu @ArthurZucker


src/transformers/models/csm/generation_csm.py
```diff
@@ -209,18 +209,19 @@ def _sample(
...
+            generation_config,
+            model_kwargs,
+            is_first_iteration=not generation_config.is_assistant,
+        )
 
         while self._has_unfinished_sequences(this_peer_finished, synced_gpus, device=input_ids.device):
             if prefill_consumed:
-                model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
+                use_cache = model_kwargs.get("use_cache", True)
+                new_inputs_ids = input_ids[:, -1:] if use_cache else input_ids
```
zucchini-nlp, line 223: why do we need to slice outside of `prepare_inputs_for_generation`?

src/transformers/generation/utils.py
```diff
@@ -828,15 +700,14 @@ def _prepare_position_ids_for_generation(self, inputs_tensor, model_kwargs):
...
             position_ids = position_ids.masked_fill(attention_mask == 0, 0)
         else:
             past_length = 0
             if (cache := model_kwargs.get("past_key_values")) is not None:
                 past_length = cache.get_seq_length()
 
-            position_ids = torch.arange(
-                past_length, seq_length + past_length, dtype=torch.long, device=inputs_tensor.device
-            )
+            position_ids = torch.arange(seq_length + past_length, dtype=torch.long, device=inputs_tensor.device)
```
zucchini-nlp, line 710: since we trust in slicing during `prepare_inputs_for_generation`? I'd also check models with custom positions, that build upon `super()`, like qwen-vl

src/transformers/generation/utils.py
```diff
@@ -3871,7 +3784,7 @@ def _prefill(self, input_ids: torch.LongTensor, generation_config: GenerationCon
                 past_length = current_length
 
             model_kwargs["attention_mask"] = attention_mask
-            model_kwargs["cache_position"] = model_kwargs["cache_position"][-1:] + 1
+            _ = model_kwargs.pop("cache_position", None)
```
zucchini-nlp: why is it popped? so chunked prefill doesn't use `cache_position` when decoding, but the full prefill keeps it

src/transformers/generation/utils.py
```diff
@@ -2839,18 +2719,19 @@ def _sample(
...
+            generation_config,
+            model_kwargs,
+            is_first_iteration=not generation_config.is_assistant,
+        )
 
         while self._has_unfinished_sequences(this_peer_finished, synced_gpus, device=input_ids.device):
             if prefill_consumed:
-                model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
+                use_cache = model_kwargs.get("use_cache", True)
+                new_inputs_ids = input_ids[:, -1:] if use_cache else input_ids
```
zucchini-nlp, line 2733: ah I see now what is the goal. If we're slicing input ids at decode loop, why not stop cat'ting next tokens with olf tokens and just re-use next tokens

This line `input_ids = torch.cat([input_ids, next_tokens[:, None]], dim=-1)`

src/transformers/generation/utils.py
```diff
@@ -3832,12 +3716,41 @@ def _assisted_decoding(
...
+        Note that usually, the prefill stage is always the first iteration of a new input batch, and thus multimodal inputs etc
+        should be treated as if it's the first iteration. However, for assisted decoding, assistants call `generate`
+        several time in a row for a same batch of inputs, so we need to pass `is_first_iteration` here for such cases.
+        """
+        # When restarting from previous cache, the `input_ids` or `inputs_embeds` are always the FULL sequence,
+        # including previous inputs, so slice them to get only new tokens
+        if (cache := model_kwargs.get("past_key_values")) is not None:
+            past_length = cache.get_seq_length()
+            input_ids = input_ids[:, past_length:]
+            if "inputs_embeds" in model_kwargs:
```
zucchini-nlp: this will fail for cases when users pass already sliced input ids, and custom cache position. If we slice by past length, we end up with zero inputs. See `test_generate_custom_cache_position` `in test_utils.py`

I don't mind changing it, but we really need to communicate it well and change slowly. We already had a lot of issues and questions when `cache_position` appeared, about how and why to continue from existing cache. If smth "breaks", we might get another wave of issues

zucchini-nlp (commented):
I love seeing hacky code gone and everything simplified! My only concern is BC with users' expectations. As I said below, there was a point when we got new issues in GH every day saying that "cache doesn't work" or users struggling to re-use system-prompt cache. So let's make sure that we communicate it well, add and update the docs with small snippets

Also if we are to deprecate `cache_position` in the next few releases, I'd prefer a tiny warning in `_prepare_inputs`. We can check if users passed `cache_position` explicitly and raise if there is anything. We need to explain how to craft their inputs correctly without cache position and link to docs

Finally let's run slow integration tests in `test_utils.py`. I know that some of those tests are kinda out-dated, gotta find a day and fix/clean them 🫠  On my list, and will start today since it's quite important to keep generation working

src/transformers/models/csm/generation_csm.py
Cyrilvallez: It's easier as different generation methods may need to keep different number of new tokens, and `_prefill` has to rely on the cache and not on the method itself. We could otherwise pass an additional arg to `prepare_inputs_...` such as `tokens_to_keep` but I thought this way was easier

src/transformers/generation/utils.py
Cyrilvallez: Because we want to keep them full length in-between the `forward` for simplicity (so that it's always the same with/without a cache), and only slice them in `prepare_inputs` if we have a cache

src/transformers/generation/utils.py
Cyrilvallez: Just so that it will be recreated correctly with full length later on, as chunked prefill takes inputs block-by-block

src/transformers/generation/utils.py
Cyrilvallez: Yes we could do that, but then we have to separate cache/no cache again... So we need an if/else somewhere. We also want to return the full sequence at the end, so we need to `cat` something anyway

src/transformers/generation/utils.py
zucchini-nlp: makes sense! I checked qwens, it will work fine because 3D positions are not re-computed based only on `use_cache`

src/transformers/generation/utils.py
zucchini-nlp: ahh right, i forgot about caching

src/transformers/generation/utils.py
zucchini-nlp: We don't recreate it later in decoding loop, afaiu. It won't cause issues downstream imo, because models that rely on `cache_position` are always MLLMs and those don't support chunked prefill

Just want to make sure we have chunked and non-chunked inputs in same format for consistency

src/transformers/generation/utils.py
Cyrilvallez: Yeah I agree, in general our way of doing here is very weird - first I assumed we were restarting from only new tokens as well, then thought that we only support restarting from full inputs. It can make sense since we have the information of the past potential padding tokens as well with full inputs.
And I wanted to change that later to also support just new tokens haha - did not know it could already be done
Only issue with restarting from only new tokens is that we may "loose" the information of previously padding tokens
However, it looks like we restart from only new tokens but fully incremented `attention_mask` in this test @zucchini-nlp, so it's not an issue at all and we can slice based on the mask as well. Can you confirm that?

Cyrilvallez: (the test is not easy to understand, they do weird stuff 😅 For example, they seem to cat new tokens in the wrong order as it does `cat(new_token, old_token)` instead of `cat(old_tokens, new_tokens)` 😅

src/transformers/generation/utils.py
Cyrilvallez: Ha you're right I thought it was! Let me quickly recreate them based on full past length! Same for pos_ids

src/transformers/generation/utils.py
Cyrilvallez: Ok, so added support for already sliced inputs + FULL mask, and completely rewrote the test to make it understandable 🤗

Cyrilvallez:
Thanks for the review!!

>As I said below, there was a point when we got new issues in GH every day saying that "cache doesn't work" or users struggling to re-use system-prompt cache. So let's make sure that we communicate it well, add and update the docs with small snippets

For next reviews, this was updated (it was only a small edge case), and there are no regression at all on what the inputs should look like from a user perspective!

zucchini-nlp (approved):
Nice, thanks!

Cyrilvallez:
run-slow: llama, mixtral, llava, whisper, mistral, qwen3_vl, gpt2

Cyrilvallez:
pixtral test was actually failing and is now passing, not sure why it gets flagged as failing
whisper longform tests are known to be failing and will be checked by @eustlb! (thanks a lot haha)

Merging!

PR #44131, opened 2026-02-18
docs: fix typo 'AuoQuant' → 'AutoQuant' and clarify FINEGRAINED_FP8 library column

cluster2600:
## What

Two small corrections in `docs/source/en/quantization/overview.md`:

1. **Typo fix**: `AuoQuant Notebook` → `AutoQuant Notebook` in the *User-Friendly Quantization Tools* section. The letter `t` was missing from the link text.

2. **Empty table cell**: The `FINEGRAINED_FP8` row had an empty *Link to library* cell. Since Fine-grained FP8 quantization is implemented natively within Transformers (no external library required), the cell now reads `Built-in` to make it clear this is not a missing link.

## Why

- The `AuoQuant` typo is confusing and makes the link text harder to find when scanning the page.
- The empty cell in the `FINEGRAINED_FP8` row looks like a mistake or missing information; clarifying it as `Built-in` is consistent with how Transformers-native features are typically presented in documentation.

## Testing

- Verified the Colab URL in the corrected link is still valid
- Confirmed `FineGrainedFP8` is implemented in `src/transformers/quantizers/quantizer_finegrained_fp8.py` and `src/transformers/integrations/finegrained_fp8.py` with no external package dependency

## Checklist

- [x] Documentation-only change
- [x] Fixes the `AuoQuant` typo
- [x] Adds `Built-in` note for `FINEGRAINED_FP8` to clarify the empty library column


stevhliu (approved):
thanks for fixing!

PR #43674, opened 2026-02-02
Use correct sampling rate in chat template

zucchini-nlp:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/43262

Supersedes https://github.com/huggingface/transformers/pull/43273 and https://github.com/huggingface/transformers/pull/43660

I don't think the prev PR will be updated by contributor, since it's a code agent PR. Also I am seeing that it attracts more code agents as long as the issue is open, so let's merge a fix here

I also made a small clean-up of `ChatTemplateKwargs` and deleted unused/duplicate kwargs, as we discussed in above linked PR review comments

PR #44132, opened 2026-02-18
[voxtral] fix voxtral proc

eustlb:
# What does this PR do?

#43674 broke voxtral processor

### DIFF
--- a/src/transformers/models/voxtral/processing_voxtral.py
+++ b/src/transformers/models/voxtral/processing_voxtral.py
@@ -147,23 +147,6 @@ class VoxtralProcessor(ProcessorMixin):
             if kwargs.get("return_assistant_tokens_mask", False):
                 raise ValueError("continue_final_message is not compatible with return_assistant_tokens_mask.")
 
-        # Fill sets of kwargs that should be used by different parts of template
-        processed_kwargs = {
-            "mm_load_kwargs": {},
-            "template_kwargs": {},
-        }
-
-        for kwarg_type in processed_kwargs:
-            for key in AllKwargsForChatTemplate.__annotations__[kwarg_type].__annotations__:
-                kwarg_type_defaults = AllKwargsForChatTemplate.__annotations__[kwarg_type]
-                default_value = getattr(kwarg_type_defaults, key, None)
-                value = kwargs.pop(key, default_value)
-                if value is not None and not isinstance(value, dict):
-                    processed_kwargs[kwarg_type][key] = value
-
-        # Pass unprocessed custom kwargs
-        processed_kwargs["template_kwargs"].update(kwargs)
-
         if isinstance(conversation, (list, tuple)) and (
             isinstance(conversation[0], (list, tuple)) or hasattr(conversation[0], "content")
         ):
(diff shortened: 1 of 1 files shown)


src/transformers/models/voxtral/processing_voxtral.py
```diff
@@ -147,23 +147,6 @@ def apply_chat_template(
...
-            for key in AllKwargsForChatTemplate.__annotations__[kwarg_type].__annotations__:
-                kwarg_type_defaults = AllKwargsForChatTemplate.__annotations__[kwarg_type]
-                default_value = getattr(kwarg_type_defaults, key, None)
-                value = kwargs.pop(key, default_value)
-                if value is not None and not isinstance(value, dict):
-                    processed_kwargs[kwarg_type][key] = value
-
-        # Pass unprocessed custom kwargs
-        processed_kwargs["template_kwargs"].update(kwargs)
-
```
eustlb, line 166: all this logic (copied from applied_chat_template) was not really required anyway and is clearer with the below

eustlb:
run-slow: voxtral

eustlb:
run-slow: voxtral

src/transformers/models/voxtral/processing_voxtral.py
```diff
@@ -192,20 +175,12 @@ def apply_chat_template(
...
-        tokenize = tokenizer_kwargs.pop("tokenize", False)
-        return_dict = tokenizer_kwargs.pop("return_dict", True)
-
-        encoded_instruct_inputs = self.tokenizer.apply_chat_template(
-            conversations,
-            tokenize=tokenize,
-            return_dict=return_dict,
-            **tokenizer_kwargs,
-        )
+        encoded_instruct_inputs = self.tokenizer.apply_chat_template(conversations, **tokenizer_kwargs)
```
ebezzam, line 184: use `text_kwargs` instead of creating new `tokenizer_kwargs`?

```python
text_kwargs = output_kwargs["text_kwargs"]
...

text_kwargs["return_tensors"] = None  # let's not return tensors here
encoded_instruct_inputs = self.tokenizer.apply_chat_template(conversations, **text_kwargs)
```

src/transformers/models/voxtral/processing_voxtral.py
```diff
@@ -173,12 +156,12 @@ def apply_chat_template(
             is_batched = False
             conversations = [conversation]
 
-        # Check for any overlapping keys between mm_load_kwargs and kwargs
-        mm_load_kwargs = processed_kwargs["mm_load_kwargs"]
-        if any(key in kwargs for key in mm_load_kwargs):
-            overlapping_keys = [key for key in mm_load_kwargs if key in kwargs]
+        unsupported_keys = {"sampling_rate", "load_audio_from_video"} & kwargs.keys()
```
ebezzam, line 163: Comment to explain why these keys?

ebezzam (approved):

src/transformers/models/voxtral/processing_voxtral.py
eustlb: Nope this would be misleading.  `text_kwargs` usually refers to what we pass to the tokenizers' call, not`apply_chat_template` (the annoying stuff about having to use mistral common backend)

naming it `template_kwargs` would have been possible, but same thing it is conflicting this this name is used for the jinja template

src/transformers/models/voxtral/processing_voxtral.py
eustlb: yep agree, adding one

src/transformers/models/voxtral/processing_voxtral.py
eustlb: I actually agree the logic is inefficient here but it will fall the in AudioProcessor refacto anyway

PR #44145, opened 2026-02-19
Add THD support in ESM

balvisio:
# What does this PR do?

This PR adds support for sequence packing in the ESM2 model. Currently, the RotaryEmbedding class of the ESM2 model supports BSHD format. This PR makes the RotayEmbedding class aware of the`position_ids` and builds the cos and sin tables accordingly.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/transformers/models/esm/configuration_esm.py
+++ b/src/transformers/models/esm/configuration_esm.py
@@ -218,7 +218,6 @@ class EsmConfig(PreTrainedConfig):
 
     def __post_init__(self, **kwargs):
         if self.is_folding_model:
-            self.model_type = "esmfold"
             if self.esmfold_config is None:
                 logger.info("No esmfold_config supplied for folding model, using default values.")
                 self.esmfold_config = EsmFoldConfig()
--- a/src/transformers/models/esm/configuration_esm.py
+++ b/src/transformers/models/esm/configuration_esm.py
@@ -218,6 +218,7 @@ class EsmConfig(PreTrainedConfig):
 
     def __post_init__(self, **kwargs):
         if self.is_folding_model:
+            self.model_type = "esmfold"
             if self.esmfold_config is None:
                 logger.info("No esmfold_config supplied for folding model, using default values.")
                 self.esmfold_config = EsmFoldConfig()
--- a/src/transformers/models/esm/configuration_esm.py
+++ b/src/transformers/models/esm/configuration_esm.py
@@ -215,6 +215,8 @@ class EsmConfig(PreTrainedConfig):
     is_decoder: bool | None = False
     add_cross_attention: bool | None = False
     tie_word_embeddings: bool = True
+    bos_token_id: int | None = None
+    eos_token_id: int | list[int] | None = 2
 
     def __post_init__(self, **kwargs):
         if self.is_folding_model:
--- a/src/transformers/models/esm/configuration_esm.py
+++ b/src/transformers/models/esm/configuration_esm.py
@@ -163,6 +163,8 @@ class EsmConfig(PreTrainedConfig):
         Type of position embedding. Choose either `"absolute"` or "rotary"`.
     emb_layer_norm_before (`bool`, *optional*):
         Whether to apply layer normalization after embeddings but before the main stem of the network.
+    rope_theta (`float`, defaults to 10000.0):
+        The base period of the RoPE embeddings. Only used when `position_embedding_type` is set to `"rotary"`.
     token_dropout (`bool`, defaults to `False`):
         When this is enabled, masked tokens are treated as if they had been dropped out by input dropout.
     is_folding_model (`bool`, defaults to `False`):
@@ -200,6 +202,7 @@ class EsmConfig(PreTrainedConfig):
     hidden_dropout_prob: float | None = 0.1
     attention_probs_dropout_prob: float | None = 0.1
     max_position_embeddings: int = 1026
+    rope_theta: float = 10000.0
     initializer_range: float = 0.02
     layer_norm_eps: float | None = 1e-12
     position_embedding_type: str | None = "absolute"
--- a/src/transformers/models/esm/modeling_esm.py
+++ b/src/transformers/models/esm/modeling_esm.py
@@ -726,12 +726,6 @@ class EsmModel(EsmPreTrainedModel):
         if (input_ids is None) ^ (inputs_embeds is not None):
             raise ValueError("You must specify exactly one of input_ids or inputs_embeds")
 
-        seq_len = inputs_embeds.shape[1] if inputs_embeds is not None else input_ids.shape[1]
-        device = input_ids.device if input_ids is not None else inputs_embeds.device
-
-        if position_ids is None:
-            position_ids = torch.arange(seq_len, device=device).unsqueeze(0)
-
         if inputs_embeds is None:
             # Important, attention_mask must be passed to the embedding class
             # This effects how the token_dropout is calculated
@@ -749,9 +743,13 @@ class EsmModel(EsmPreTrainedModel):
             past_key_values=None,
         )
 
-        position_embeddings = (
-            self.rotary_embeddings(inputs_embeds, position_ids) if self.position_embedding_type == "rotary" else None
-        )
+        if self.position_embedding_type == "rotary":
+            if position_ids is None:
+                seq_len = inputs_embeds.shape[1]
+                position_ids = torch.arange(seq_len, device=inputs_embeds.device).unsqueeze(0)
+            position_embeddings = self.rotary_embeddings(inputs_embeds, position_ids)
+        else:
+            position_embeddings = None
 
         encoder_outputs = self.encoder(
             inputs_embeds,
--- a/src/transformers/models/esm/modeling_esm.py
+++ b/src/transformers/models/esm/modeling_esm.py
@@ -103,54 +103,6 @@ def average_product_correct(x):
     return normalized
 
 
-# class RotaryEmbedding(torch.nn.Module):
-#     """
-#     Rotary position embeddings based on those in
-#     [RoFormer](https://huggingface.co/docs/transformers/model_doc/roformer). Query and keys are transformed by rotation
-#     matrices which depend on their relative positions.
-#     """
-
-#     inv_freq: torch.Tensor  # fix linting for `register_buffer`
-
-#     def __init__(self, dim: int):
-#         super().__init__()
-#         self.dim = dim
-#         # Generate and save the inverse frequency buffer (non trainable)
-#         inv_freq = 1.0 / (10000 ** (torch.arange(0, dim, 2, dtype=torch.int64).float() / dim))
-#         self.register_buffer("inv_freq", inv_freq)
-
-#     def _update_cos_sin_tables(self, x, seq_dimension=2, seq_len=None, position_ids=None):
-#         if seq_len is None:
-#             seq_len = x.shape[seq_dimension]
-
-#         if position_ids is None:
-#             t = torch.arange(x.shape[seq_dimension], device=x.device).type_as(self.inv_freq)
-#         else:
-#             t = position_ids.squeeze()
-#             if t.dim() != 1:
-#                 raise ValueError(
-#                     f"position_ids must be 1-D after squeezing, got shape {position_ids.shape}. "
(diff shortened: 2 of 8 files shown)


Rocketknight1:
Hmm, this is an interesting PR! In general I agree that the ESM attention code is old and could use a refactor, but rather than doing it this way, could we refactor it along the lines of a modern masked LM? `modernbert` in particular has a clean and well-tested implementation, and it might be possible to use most of it.

One issue though - I remember from the original ESM PR that `self.inv_freq` is saved in the state dict and should not be recomputed because it was (iirc?) computed in `float16`. Therefore, if you recompute it with `float32` values, the model output will be subtly different. Even though `float32` is more accurate, we need to match what the model was trained with.

balvisio:
Thanks for taking a look at this. Understood; I can duplicate the implementation of ModernBertRotaryEmbedding in this file and adapt it as needed. If you’re okay with it, we could merge this once the tests are fixed, and then I can follow up with a proper refactor in a few weeks. Of course, happy to proceed however you prefer

Rocketknight1:
I'd prefer one single PR rather than this one plus a follow-up, if that's okay with you!

balvisio:
Sounds good. I’ll take care of it in a few days

balvisio:
Hi @Rocketknight1 , following your suggestion, I refactored `ESMModel` making it more aligned with the `ModernBERT` implementation. I have also took care of the `inv_freq` buffer initialization so that it is backwards compatible; one of the tests assets that casting the buffer to `float16` givesthe exact result.
Thank you for reviewing.

Rocketknight1:
Hi @balvisio, please check the CI first! There might be issues here, but I can review after it's green.

balvisio:
Hi @Rocketknight1 : CI is green now. Thanks!

Rocketknight1:
run-slow: esm, evolla

src/transformers/models/esm/modeling_esm.py
```diff
@@ -643,6 +726,12 @@ def forward(
         if (input_ids is None) ^ (inputs_embeds is not None):
             raise ValueError("You must specify exactly one of input_ids or inputs_embeds")
 
+        seq_len = inputs_embeds.shape[1] if inputs_embeds is not None else input_ids.shape[1]
+        device = input_ids.device if input_ids is not None else inputs_embeds.device
+
+        if position_ids is None:
+            position_ids = torch.arange(seq_len, device=device).unsqueeze(0)
```
Rocketknight1: One nit I'm not sure about: Does this need to be padding-aware? I think the ESM code also covers the ESM-1 models, which used absolute embeddings, but I worry padding tokens might disrupt the position IDs here. Possibly an issue in training if MLM random-masking selects padding tokens?

balvisio: @Rocketknight1 : You are correct. This was breaking for absolute embeddings. Fixed it in last commit.

src/transformers/models/esm/configuration_esm.py
```diff
@@ -215,6 +218,7 @@ class EsmConfig(PreTrainedConfig):
 
     def __post_init__(self, **kwargs):
         if self.is_folding_model:
+            self.model_type = "esmfold"
```
Rocketknight1: I missed this one the first time - is there a reason we want this line, when it wasn't there before? Do we use that model type elsewhere in the code?

Rocketknight1:
@balvisio almost ready to go, one final thing I missed the first time around!

src/transformers/models/esm/configuration_esm.py
balvisio: Hi @Rocketknight1 , the reason is because the the two models need different conversion behavior.
1. I added a conversion mapping for "esm" in conversion_mapping.py to handle the inv_freq key rename.
2. Since the default `EsmForProteinFolding` doesn't have rotary embeddings the test

By setting `self.model_type = "esmfold"` when i`s_folding_model=True`, `get_checkpoint_conversion_mapping("esmfold")` returns `None`. 

Let me know if this should be handled in a different way.

Rocketknight1: I think it's fine because if the mapping doesn't match a weight, then nothing happens and we're okay, right? So having a mapping entry that isn't valid for our model isn't a problem, and we can just remove the `self.model_type` line

balvisio: The problem is with the `test_reverse_loading_mapping`. With `EsmFold`:

```
tests/models/esm/test_modeling_esmfold.py::EsmFoldModelTest::test_reverse_loading_mapping FAILED                                                              [100%]

============================================================================= FAILURES ==============================================================================
___________________________________________________________ EsmFoldModelTest.test_reverse_loading_mapping ___________________________________________________________

self = <tests.models.esm.test_modeling_esmfold.EsmFoldModelTest testMethod=test_reverse_loading_mapping>, check_keys_were_modified = True

    def test_reverse_loading_mapping(self, check_keys_were_modified=True):
        """Make sure we can load and save correctly the models having any weight renaming mapping or weight conversion
        mapping.
        Note that this test would be better if we could start from the serialized keys, and check that the model
        keys correspond to the weight converions. However, when instantiating a model, it already has the "target"
        keys (or modified keys after mapping) of the conversion mapping, so we have to do it the other way, i.e.
        reverse the conversion and then check that those converted keys match correctly the conversions.
    
        However, all the checks performed here should ensure everything is going as it should.
    
        Args:
            check_keys_were_modified (`bool`, *optional*, defaults to `True`):
                Whether to expect keys being modified or not. In some cases, models do not change keys but
                their weights, e.g. via transpose, memory alignment, etc.
        """
        config, _ = self.model_tester.prepare_config_and_inputs_for_common()
    
        #  Some MoE models alternate between a classic MLP and a MoE layer, in which case we want to have at
        # lest one MoE layer here to check the mapping
        config_to_set = config.get_text_config(decoder=True)
        config_to_set.first_k_dense_replace = 1  # means that the first layer (idx 0) will be MLP, then MoE
        config_to_set.moe_layer_start_index = 1  # same as above but for Ernie 4.5...
        config_to_set.mlp_only_layers = [0]  # same but for qwens
        config_to_set.num_dense_layers = 1  # lfm2_moe
    
        for model_class in self.all_model_classes:
            # Each individual model is a subtest
            with self.subTest(model_class.__name__):
                model = model_class(copy.deepcopy(config))
                # Skip if no conversions
                conversions = get_model_conversion_mapping(model, add_legacy=False)
                if len(conversions) == 0:
                    self.skipTest("No conversion found for this model")
    
                # Find the model keys, so the targets according to the conversions
                model_keys = list(model.state_dict().keys())
    
                with tempfile.TemporaryDirectory() as tmpdirname:
                    # Serialize with reverse mapping
                    model.save_pretrained(tmpdirname)
                    state_dict = load_file(os.path.join(tmpdirname, "model.safetensors"))
                    # Get all the serialized keys that we just saved according to the reverse mapping
                    serialized_keys = list(state_dict.keys())
    
                if check_keys_were_modified:
                    # They should be different, otherwise we did not perform any mapping
                    self.assertNotEqual(sorted(serialized_keys), sorted(model_keys), "No key mapping was performed!")
    
                # Check that for each conversion entry, we at least map to one key
                for conversion in conversions:
                    for source_pattern in conversion.source_patterns:
                        # Some patterns are written for gen-model only and won't be applied on base model
                        if "lm_head" in source_pattern and model_class not in [
                            *get_values(MODEL_FOR_CAUSAL_LM_MAPPING_NAMES),
                            *get_values(MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES),
                        ]:
                            continue
    
                        # Sometimes the mappings specify keys that are tied, so absent from the saved state dict
                        if isinstance(conversion, WeightRenaming):
                            # We need to revert the target pattern to make it compatible with regex search
                            target_pattern_reversed = conversion.target_patterns[0]
                            captured_group = process_target_pattern(source_pattern)[1]
                            if captured_group:
                                target_pattern_reversed = target_pattern_reversed.replace(r"\1", captured_group)
                            if any(re.search(target_pattern_reversed, k) for k in model.all_tied_weights_keys.keys()):
                                continue
                        num_matches = sum(re.search(source_pattern, key) is not None for key in serialized_keys)
>                       self.assertTrue(
                            num_matches > 0,
                            f"`{source_pattern}` in `{conversion}` did not match any of the source keys. "
                            "This indicates whether that the pattern is not properly written, ot that it could not be reversed correctly",
                        )
E                       AssertionError: False is not true : `encoder.layer.*.attention.self.rotary_embeddings.inv_freq` in `WeightRenaming(source_patterns=['encoder.layer.*.attention.self.rotary_embeddings.inv_freq'], target_patterns=['rotary_embeddings.inv_freq'], compiled_sources=re.compile('(?P<g0>encoder.layer\\..*\\.attention.self.rotary_embeddings.inv_freq)'), distributed_operation=None, quantization_operation=None, collected_tensors=defaultdict(<class 'list'>, {}), layer_targets=defaultdict(<class 'set'>, {}))` did not match any of the source keys. This indicates whether that the pattern is not properly written, ot that it could not be reversed correctly

tests/test_modeling_common.py:4754: AssertionError
```

I think the problem is that since by default `EsmFold` uses absolute embeddings the `rotary_embeddings.inv_freq` is not found.

balvisio: I could override `test_reverse_loading_mapping` to use a modfied `EsmFoldConfig` with rotary position embeddings instead.

Rocketknight1: Yes, I think the test override makes sense here. The reason this is a little fiddly is that we generally prefer to have one model class correspond to one architecture, with even subtle changes being split into another class. However, we didn't do that with `esm` / `esmfold`, mostly because of a rushed release!

The result is that some test assumptions are violated for this class, which I didn't realize at first so thanks for pointing it out! The test assumes, as you said, that the mappings all match something. I think adding an override to either skip the test or modify it for ESM/ESMFold is correct!

balvisio: Sounds good. Removed `self.model_type = "esmfold"` and changed the test to use a config with rotary embeddings.

balvisio:
@Rocketknight1 The CI is failing. The failing test doesn't seem related to my change and it passes locally

Rocketknight1:
@balvisio can you try rebasing and resolving the merge conflict? Hopefully tests pass after that, if not just ping me and I'll see what I can do with the CI.

balvisio:
@Rocketknight1 CI looks good!

Rocketknight1:
run-slow: esm, evolla

Rocketknight1:
Happy to merge at this point, and thanks for iterating!

Rocketknight1 (approved):

Rocketknight1:
CI issues, will fix with https://github.com/huggingface/transformers/pull/45123

balvisio:
@Rocketknight1 : I added one more commit that adds missing 'bos_token_id' and 'eos_token_id' fields to the ESM Config. Since they are not there I am reaching this line `https://github.com/huggingface/transformers/blob/8213e0d920d52cb00dcade16b6d1f6e952ac0a8c/src/transformers/modeling_utils.py#L4372` and that throws the following exception:

```
self.warn_if_padding_and_no_attention_mask(input_ids, attention_mask)
  File "/usr/local/lib/python3.12/dist-packages/transformers/modeling_utils.py", line 4379, in warn_if_padding_and_no_attention_mask
    (self.config.bos_token_id is not None and self.config.bos_token_id == self.config.pad_token_id)
     ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/transformers/configuration_utils.py", line 164, in __getattribute__
    return super().__getattribute__(key)
```

Rocketknight1:
@balvisio nice, thank you! Trying to merge again, not sure why the CI is complaining.

balvisio:
Hi @Rocketknight1 ping on this :)  Is there something still going on with the CI?

Rocketknight1:
@balvisio yeah, I'll investigate when I get a chance. In the meantime I'll keep merging and hope we get through!

Rocketknight1:
@balvisio CI defeated at last. Thank you again for the PR!

balvisio:
Thank you!

PR #44156, opened 2026-02-19
Fix aqml `modules_to_not_convert`

SunMarc:
# What does this PR do?

This PR fixes a minor error when using aqml quantization. We specified the wrong argument.


LalithaMV (approved):

BenjaminBossan:
@SunMarc Why was the PR closed, is the fix no longer necessary?

SunMarc:
@tarekziade fixed it in a pr !

PR #44157, opened 2026-02-19
Use correct mask for packed inputs in Qwen-VL 

zucchini-nlp:
# What does this PR do?

As per title, gets rid of `if/else` per attn implementation

### DIFF
--- a/src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
+++ b/src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
@@ -31,7 +31,7 @@ from ...activations import ACT2FN
 from ...cache_utils import Cache, DynamicCache
 from ...generation import GenerationMixin
 from ...integrations import use_experts_implementation, use_kernel_forward_from_hub, use_kernelized_func
-from ...masking_utils import create_causal_mask
+from ...masking_utils import create_bidirectional_mask, create_causal_mask, packed_sequence_mask_function
 from ...modeling_flash_attention_utils import FlashAttentionKwargs
 from ...modeling_layers import GradientCheckpointingLayer
 from ...modeling_outputs import BaseModelOutputWithPooling, MoeCausalLMOutputWithPast, MoeModelOutputWithPast
@@ -39,7 +39,7 @@ from ...modeling_rope_utils import dynamic_rope_update
 from ...modeling_utils import ALL_ATTENTION_FUNCTIONS, PreTrainedModel
 from ...processing_utils import Unpack
 from ...utils import TransformersKwargs, auto_docstring, can_return_tuple, torch_compilable_check
-from ...utils.generic import is_flash_attention_requested, maybe_autocast, merge_with_config_defaults
+from ...utils.generic import maybe_autocast, merge_with_config_defaults
 from ...utils.output_capturing import OutputRecorder, capture_outputs
 from .configuration_ernie4_5_vl_moe import (
     Ernie4_5_VL_MoeConfig,
@@ -585,11 +585,11 @@ class Ernie4_5_VL_MoeVisionAttention(nn.Module):
         self,
         hidden_states: torch.Tensor,
         cu_seqlens: torch.Tensor,
-        rotary_pos_emb: torch.Tensor | None = None,
+        attention_mask: torch.Tensor | None = None,
         position_embeddings: tuple[torch.Tensor, torch.Tensor] | None = None,
         **kwargs,
     ) -> torch.Tensor:
-        seq_length = hidden_states.shape[0]
+        seq_length = hidden_states.shape[1]
         query_states, key_states, value_states = (
             self.qkv(hidden_states).reshape(seq_length, 3, self.num_heads, -1).permute(1, 0, 2, 3).unbind(0)
         )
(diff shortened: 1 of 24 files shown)


zucchini-nlp:
run-slow: qwen2_vl, qwen2_5_vl

zucchini-nlp:
run-slow: ernie4_5_vl_moe, glm4v, glm4v_moe, glm_ocr, paddleocr_vl, qwen2_5_omni, qwen2_5_vl, qwen2_vl, qwen3_5, qwen3_5_moe, qwen3_omni_moe, qwen3_vl, qwen3_vl_moe, video_llama_3

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
```diff
@@ -585,7 +585,7 @@ def forward(
         self,
         hidden_states: torch.Tensor,
         cu_seqlens: torch.Tensor,
-        rotary_pos_emb: torch.Tensor | None = None,
+        attention_mask: torch.Tensor | None = None,
```
zucchini-nlp, line 588: unused arg, prob was a bad copy from other models. We use only `position_embeddings`

vasqu: Oops, very likely yes 😓 I started from qwen vl and opted to make it a bit cleaner back then iirc

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
```diff
@@ -603,47 +603,22 @@ def forward(
...
-                    scaling=self.scaling,
-                    dropout=0.0 if not self.training else self.attention_dropout,
-                    is_causal=False,
-                    **kwargs,
-                )[0]
-                for q, k, v in zip(*splits)
-            ]
-            attn_output = torch.cat(attn_outputs, dim=1)
+        max_seqlen = (cu_seqlens[1:] - cu_seqlens[:-1]).max()  # FA-kwargs
+        attn_output, _ = attention_interface(
```
vasqu, line 607: Just noticing it now, but we never collected the attention weights of the vision side 👀

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
```diff
@@ -603,47 +603,22 @@ def forward(
...
+            key_states,
+            value_states,
+            attention_mask=attention_mask,
+            scaling=self.scaling,
+            dropout=0.0 if not self.training else self.attention_dropout,
+            cu_seq_lens_q=cu_seqlens,
+            cu_seq_lens_k=cu_seqlens,
+            max_length_q=max_seqlen,
+            max_length_k=max_seqlen,
+            is_causal=False,
```
vasqu, line 619:
```suggestion
```
I know this comes from previous implementations, but we really shouldn't pass this manually ourselves but rely on `self.is_causal`

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
```diff
@@ -952,13 +927,27 @@ def forward(
...
+        seq_lengths = cu_seqlens[1:] - cu_seqlens[:-1]
+        packed_sequence = torch.repeat_interleave(
+            torch.arange(len(seq_lengths), device=hidden_states.device), seq_lengths
+        ).unsqueeze(0)
+
+        attention_mask = create_bidirectional_mask(
+            config=self.config,
+            inputs_embeds=hidden_states[None, ...],
+            attention_mask=None,
+            and_mask_function=packed_sequence_mask_function(packed_sequence),
```
vasqu, line 939: `and_mask_function` is quite expensive on runtime. I'd prefer if we could natively integrate it into our mask API instead, not to use vmap (should work OOB)

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
```diff
@@ -603,47 +603,22 @@ def forward(
...
-                    attention_mask=None,
-                    scaling=self.scaling,
-                    dropout=0.0 if not self.training else self.attention_dropout,
-                    is_causal=False,
-                    **kwargs,
-                )[0]
-                for q, k, v in zip(*splits)
-            ]
-            attn_output = torch.cat(attn_outputs, dim=1)
+        max_seqlen = (cu_seqlens[1:] - cu_seqlens[:-1]).max()  # FA-kwargs
```
vasqu, line 606: Unsure how / where max_seqlen would be called down the line, but I remember a lot of `.item()` to match FA signature (e.g. for compile); so might be smarter to call it once and before entering each attention module

vasqu (commented):
Some comments from my side, definitely the right way to go! We should aim to natively support this in our mask API (without vmap, i.e. the and mask fn) --> otherwise we lose quite a lot of perf

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
zucchini-nlp: in qwen-vl we did collect as a list of 4D tensors per layer. Prob it wasn't standard and was not copied everywhere

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
zucchini-nlp: hmm, what do you mean by "natively"? I could prepare by looping over cu seq lens and un-masking each block and keep it as a small fn in model file. In that case we don't use `create_bidiractional_mask`

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
zucchini-nlp: yeah, might do in the `VisionModel` once. IIRC fa wouldn't prepare max length for us so we have to pass it explciitly

src/transformers/models/ernie4_5_vl_moe/modeling_ernie4_5_vl_moe.py
vasqu: If you check the new mask API, we force vmap when we pass and/or mask functions https://github.com/huggingface/transformers/blob/efdcbc77ea10f4eb5106b9a4d6df80c541f10a07/src/transformers/masking_utils.py#L1003-L1017

Natively in this case would mean either a new kwarg that we can use to control packed sequences as well (instead of pos ids) or a new function similar to how sliding window is extended https://github.com/huggingface/transformers/blob/efdcbc77ea10f4eb5106b9a4d6df80c541f10a07/src/transformers/masking_utils.py#L1154

The problem with and/or masks is that we can have no idea about their functionality and vmaping is powerful enough to support almost anything, so we are kinda forced to vmap for compatibility/safety

zucchini-nlp:
Discussed internally, we'll add support for torch varlen

PR #44160, opened 2026-02-19
Add model lerobot PI0 to transformers

molbap:
# What does this PR do?

Hey yall, I started porting the pi0 model so Transformers can be a backend for VLAs directly.  

I checked it against LeRobot on fix/lerobot_openpi: outputs seem to match and for sure lerobot/pi0_base loads cleanly (no missing/unexpected keys). The model is PI0ForConditionalGeneration: PaliGemma (Gemma 2B) + an action expert (Gemma 300M), merged cross-attn, flow-matching for continuous action generation, built in modular_pi0.py.

Main divergence vs basic PaliGemma is scaling/caching: get_image_features returns projector output without the sqrt(hidden_size) factor; prefix/suffix pre-divide inputs_embeds to cancel Gemma’s internal scaling; training mask makes prefix + action tokens bidirectional; inference fills prefix KV once and crops back to prefix length each denoise step so suffix doesn’t accumulate.

Still missing: tests, processor wiring, and making sample_actions a proper pluggable .generate(). I think there's a simple way to make it work with the mask interface we have in a simpler way as well. There's also bits of pi0fast but I think I'll remodularize after, based on pi0. It's a bit ugly still. 

cc @ArthurZucker @pkooij , will try and push the final version here tomorrow, already down to a 600 LoC modular (minus the processing but it can belong in either repo?)


src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+        raise ValueError("The time tensor is expected to be of shape `(batch_size, )`.")
+
+    device = time.device
+    dtype = torch.float64 if device.type not in ("mps", "cpu") else torch.float32
+    fraction = torch.linspace(0.0, 1.0, dimension // 2, dtype=dtype, device=device)
+    period = min_period * (max_period / min_period) ** fraction
+
+    scaling_factor = 1.0 / period * 2 * math.pi
+    sin_input = scaling_factor[None, :] * time[:, None]
+    return torch.cat([torch.sin(sin_input), torch.cos(sin_input)], dim=1)
```
molbap: this is likely rope-abstractable

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+
+
+def sample_beta(alpha: float, beta: float, batch_size: int, device: torch.device) -> torch.Tensor:
+    alpha_t = torch.tensor(alpha, dtype=torch.float32)
+    beta_t = torch.tensor(beta, dtype=torch.float32)
+    dist = torch.distributions.Beta(alpha_t, beta_t)
+    return dist.sample((batch_size,)).to(device)
```
molbap: this is new and a kind of .generate helper

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+        self.action_time_mlp_in = nn.Linear(2 * expert_hidden_size, expert_hidden_size)
+        self.action_time_mlp_out = nn.Linear(expert_hidden_size, expert_hidden_size)
+
+        self.post_init()
+
+    def get_input_embeddings(self):
+        return self.model.get_input_embeddings()
+
+    def set_input_embeddings(self, value):
+        self.model.set_input_embeddings(value)
```
molbap: will be removed

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+        else:
+            num_image_tokens = image_features.shape[1]
+            embs.append(image_features)
+            pad_masks.append(torch.ones(batch_size, num_image_tokens, dtype=torch.bool, device=image_features.device))
+
+        lang_emb = self.model.embed_language_tokens(input_ids)
+        embs.append(lang_emb)
+        pad_masks.append(attention_mask.bool())
+
+        return torch.cat(embs, dim=1), torch.cat(pad_masks, dim=1)
```
molbap: that and the suffix are the hard harder things to implement cleanly, likely easier to offload to a processor.

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+
+        attention_mask_4d = create_causal_mask(
+            config=self.config.text_config,
+            inputs_embeds=prefix_embs,
+            attention_mask=combined_pad_masks,
+            cache_position=cache_position,
+            past_key_values=None,
+            position_ids=position_ids,
+            or_mask_function=or_masks(prefix_bidirectional, suffix_bidirectional),
+        )
```
molbap: this seems to work but needs a lot of computing that hopefully we can do with `masking_utils.py`, main bit IMO that can be tricky

zucchini-nlp:
Very interesting PR, subscribed 👀 

Also had long in my mind that we might want to start supporting VLAs if that becomes a thing! Though it seems like most good ones are not autoregressive, but more a hybrid with diffusion

molbap:
Hello @zucchini-nlp ! :D yes, these three models seem to be the closest to preexisting archs we have, so I think we can support it

molbap:
Still WIP, but managed to take down modular a notch. the hidden states collection is the tricky part (to get it clean). Works though, got a good match with lerobot impl.

zucchini-nlp:
I will try to wrap this PR up in Pablo's absence, will start working on it today

ArthurZucker:
Hhaha okayyy here to help as well 🤗

zucchini-nlp:
Yay, 157 lines of modular! 🎉 

@pkooij  I made a model following proper standards and separate DiT from VLM. VLM is a simple Paligemma and DiT is a Gemma, no monkey-patching required from our side. I realized that Pi0 is just trying to do prefix tuning when training, which I fixed by passing cache around

Currently there is one issue with how norm is applied on text embedding which I am fixing on a separate PR (it's linked to this PR). Apart from that, the checkpoints are matched and the suffix/prefix embeddings are matching. I will look at norm and pacth the norm so the whole pipeline outputs also match. I may also need to update the bidirectional attention mask

Do you think smth like this would work for LeRobot? We can also add a proper processor because the VLM now requires that input ids have "image placeholder tokens'

src/transformers/models/pi0/modeling_pi0.py
```diff
@@ -0,0 +1,387 @@
...
+    def forward(self, time):
+        device_type = time.device.type if isinstance(time.device.type, str) and time.device.type != "mps" else "cpu"
+        with maybe_autocast(device_type=device_type, enabled=False):  # Force float32
+            sinusoid_freq = self.sinusoid_freq[None, :]
+            emb = sinusoid_freq * time[:, None]
+            time_embeds = torch.cat([emb.sin(), emb.cos()], dim=1)
+        return time_embeds
+
+
+class ActionTimeEmbedding(nn.Module):
```
ArthurZucker:
```suggestion
class PI0PActionTimeEmbedding(nn.Module):
```

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+        self.time_sampling_beta_alpha = time_sampling_beta_alpha
+        self.time_sampling_beta_beta = time_sampling_beta_beta
+        self.time_sampling_scale = time_sampling_scale
+        self.time_sampling_offset = time_sampling_offset
+        self.min_period = min_period
+        self.max_period = max_period
+        super().__init__(**kwargs)
+
+
+class TimestepEmbeddings(nn.Module):
```
ArthurZucker:
```suggestion
class PI0PTimestepEmbeddings(nn.Module):
```

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+    def forward(self, time):
+        device_type = time.device.type if isinstance(time.device.type, str) and time.device.type != "mps" else "cpu"
+        with maybe_autocast(device_type=device_type, enabled=False):  # Force float32
+            sinusoid_freq = self.sinusoid_freq[None, :]
+            emb = sinusoid_freq * time[:, None]
+            time_embeds = torch.cat([emb.sin(), emb.cos()], dim=1)
+        return time_embeds
+
+
+class ActionTimeEmbedding(nn.Module):
```
ArthurZucker:
```suggestion
class PI0PActionTimeEmbedding(nn.Module):
```

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+        self.register_buffer("sinusoid_freq", sinusoid_freq, persistent=False)
+
+    @staticmethod
+    def compute_freqs(config):
+        fraction = torch.linspace(0.0, 1.0, config.dit_config.hidden_size // 2, dtype=torch.float32)
+        period = config.min_period * (config.max_period / config.min_period) ** fraction
+        sinusoid_freq = 1.0 / period * 2 * math.pi
+        return sinusoid_freq
+
+    def forward(self, time):
```
ArthurZucker, line 327: also this is just rope no? LlamaRotaryEmbedding is the same :)

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+                inputs_embeds = self.embed_prefix(input_ids, pixel_values, pixel_attention_mask)
+                inputs_embeds = inputs_embeds / math.sqrt(2048)
+
+            past_key_values = self.vlm(
+                inputs_embeds=inputs_embeds,
+                attention_mask=attention_mask,
+                position_ids=position_ids,
+                past_key_values=None,  # `None` on purpose
+                use_cache=True,
+            ).past_key_values
```
ArthurZucker, line 451: this prevents us from getting ouptut attentions from the vlm no?

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+            dit_attention_mask = torch.cat([attention_mask, noise_mask], dim=1)
+            dit_position_ids = (torch.cumsum(dit_attention_mask, dim=1) - 1)[:, -action_embeds.shape[1] :]
+
+        def blockwise_bidirectional_mask(block_boundaries: torch.Tensor):
+            def inner_mask(b_idx, h_idx, q_idx, kv_idx):
+                q_block = torch.bucketize(q_idx, block_boundaries)
+                kv_block = torch.bucketize(kv_idx, block_boundaries)
+                return kv_block <= q_block
+
+            return inner_mask
```
ArthurZucker: should be define outside this scope pelase!

tests/models/pi0/test_modeling_pi0.py
```diff
@@ -0,0 +1,280 @@
...
+@require_torch
+@slow
+class PI0ModelIntegrationTest(unittest.TestCase):
+    def test_pi0_base_reference_values(self):
+        model = PI0ForConditionalGeneration.from_pretrained("lerobot/pi0_base", torch_dtype=torch.float32).eval()
+        processor = PI0Processor.from_pretrained("google/paligemma-3b-pt-224")
+
+        inputs = processor(
+            text=["Pick up the object\n"],
+            images=load_image("/raid/raushan/image.png"),
```
ArthurZucker: to change

tests/models/pi0/test_modeling_pi0.py
ArthurZucker, line 1: imo, lots of people are gonna train this model, make sure its added to the list of imporant models / models tests in the training CI and distributed CI

src/transformers/models/pi0/configuration_pi0.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Example:
+    ```python
+    >>> from transformers import PI0ForConditionalGeneration, PI0Config
+
+    >>> config = PI0Config()
+    >>> model = PI0ForConditionalGeneration(config)
+    ```
+    """
+
```
ArthurZucker, line 65: would be nice to have a TP plan here!

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+            position_ids=dit_position_ids,
+            past_key_values=past_key_values,
+            **kwargs,
+        )
+        return dit_output
+
+
+class PI0ForConditionalGeneration(PI0PreTrainedModel):
+    """PI0 model with action projection heads and flow matching."""
+
```
ArthurZucker, line 493:
```suggestion
    _tp_plan = {"action_out_proj": "colwise_gather_output"}

```
this at least is needed IMO

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,604 @@
...
+        return CausalLMOutputWithPast(
+            loss=loss,
+            logits=predicted_velocity,
+            past_key_values=outputs.past_key_values,
+            hidden_states=outputs.hidden_states,
+            attentions=outputs.attentions,
+        )
+
+    @torch.no_grad()
+    def sample_actions(
```
ArthurZucker, line 593: is this the common API used in lerobot / standard?

ArthurZucker (commented):
Nice! As its the first of such models, we should push a bit more the testing!

src/transformers/models/pi0/modular_pi0.py
zucchini-nlp: no same RoPE as in Llama from what I see in `compute_sinusoids`, but the idea is similar. We can copy the forward from llama prob and re-create `__init__` manually

src/transformers/models/pi0/modular_pi0.py
zucchini-nlp: we shouldn't need it afaiu. In VLMs we also don't return attn and hidden states of the backbone vision model, so same here

zucchini-nlp: the main model is DiT used for denoising and if ppl are interested in vision or text attentions, we can re-direct then to `get_image_features`  or other utilities. That looks more like a researchy use-case

src/transformers/models/pi0/modular_pi0.py
zucchini-nlp: oke, actually I've been running on CPU-only as the model is quite lightweight. We can add TP as well

src/transformers/models/pi0/modular_pi0.py
zucchini-nlp: we can say so, should be standard for VLAs. Also we can call it `self.denoise` to generalize across any denoising model

I'm still not 100% sure if we will keep adding more in transformers though. We can't support various scheduling techniques without bloating the code, so if more models are needed we can move all code to `diffusers`. I made sure the "moving" will be easy and straightforward

src/transformers/models/pi0/modular_pi0.py
ArthurZucker: yeah the inv_freq period is different !

src/transformers/models/pi0/modular_pi0.py
ArthurZucker: ok SG

src/transformers/models/pi0/configuration_pi0.py
zucchini-nlp: I checked that TP-plan works recursively for subconfigs and the tests are passing

zucchini-nlp:
Added a small training tests with small chunk from libero dataset, and made sure it runs and loss goes down. One thing is that we don't support GC because of prefix caching

To support GC, we'll need to add DiT as a new model class that processes actions and images in two separate streams, up to attention. During attn, it would concat all inputs into a bigger tensor. We can tie weights for this one with VLM's language model. I didn't do this because it's +600 LOC code at least to add the whole gemma with `encoder_hidden_states aka prefix`

src/transformers/models/pi0/modular_pi0.py
zucchini-nlp: I decided to not copy, since llama positions are 2D tensors. So we'd need to squeeze an extra dim in timestep and unsqueeze it back to be able to use LlamaRoPE

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,635 @@
...
+
+
+@auto_docstring
+class PI0Processor(PaligemmaProcessor):
+    def __init__(self, image_processor=None, tokenizer=None, chat_template=None, **kwargs):
+        self.height, self.width = image_processor.size["height"], image_processor.size["width"]
+        state_mean = kwargs.get("state_mean", [-0.0419, 0.0354, 0.8257, 2.9083, -0.5562, -0.1665, 0.0283, -0.0286])
+        state_std = kwargs.get("state_std", [0.1074, 0.1442, 0.2572, 0.3441, 1.2344, 0.3580, 0.0133, 0.0132])
+        actions_mean = kwargs.get("actions_mean", [0.0182, 0.0586, -0.0559, 0.0046, 0.0029, -0.0077, -0.0916])
+        actions_std = kwargs.get("actions_std", [0.2825, 0.3590, 0.3674, 0.0377, 0.0543, 0.0872, 0.9958])
```
zucchini-nlp, line 69: would be great to have its own processor confuig in `pi0` repo with all these values. Currently we load PG which doesn't have action/state related params

Ig we can update all official pi-0 repos in `lerobot` org, but we still need these for users' finetuned models

src/transformers/models/pi0/image_processing_pi0_fast.py
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+from ...siglip.image_processor_siglip import SiglipImageProcessorFast
+from ...utils import auto_docstring
+
+
+@auto_docstring
+class PI0ImageProcessorFast(SiglipImageProcessorFast):
+    size = {"max_height": 224, "max_width": 224}
+    pad_size = {"height": 224, "width": 224}
+    do_pad = True
```
ArthurZucker, line 28: we really should not have this! its only config attr, let's push a PR with correct ones but not have a new when we don't need

src/transformers/models/pi0/modular_pi0.py
```diff
@@ -0,0 +1,635 @@
...
+        past_key_values: Cache | None = None,
+        actions: torch.FloatTensor = None,  # aka labels
+        **kwargs,
+    ) -> CausalLMOutputWithPast:
+        r"""
+        actions (`torch.Tensor`, *optional*): args description placeholder
+        pixel_attention_mask (`torch.Tensor`, *optional*): args description placeholder
+        state  (`torch.Tensor`, *optional*): args description placeholder
+        noise  (`torch.Tensor`, *optional*): args description placeholder
+        timestep  (`torch.Tensor`, *optional*): args description placeholder
```
ArthurZucker: 😉 hehe

src/transformers/models/pi0/modular_pi0.py
ArthurZucker: `self.denoise` could be nice if its standard yes!

tests/models/pi0/test_modeling_pi0.py
```diff
@@ -0,0 +1,389 @@
...
+    def test_training_gradient_checkpointing_use_reentrant_false(self):
+        pass
+
+    @unittest.skip("Prefix tuning doesn't work with GC and the model uses prefix tuning to fuse VLM outputs")
+    def test_training_gradient_checkpointing_use_reentrant_true(self):
+        pass
+
+
+@require_torch
+class PI0ModelSmokeTest(unittest.TestCase):
```
ArthurZucker: to remove!@

ArthurZucker (approved):
Nice!

src/transformers/models/pi0/image_processing_pi0_fast.py
zucchini-nlp: lerobot has no processor so it is https://github.com/huggingface/transformers/pull/44160#discussion_r2920310348

We can support official PI0 but the finetunes ones won't work then :(

zucchini-nlp: unless the owners of repos update and create a processor config with correct defaults

zucchini-nlp:
run-slow: pi0

zucchini-nlp:
oke, on monday then. Will need to ssh and re-check all values

PR #44163, opened 2026-02-19
ESM2 attention_mask and token_dropout fix

lhallee:
Fixes:
https://github.com/huggingface/transformers/issues/44162

@ArthurZucker @Cyrilvallez


zucchini-nlp (approved):
Thanks, lgtm! Can you check why the `embeds` test is failing, I think we might have to skip it or disable `token_dropout` for this particular test

Rocketknight1:
I think that test is invalid when embeddings are scaled because of token dropout, so I'll add a skip so we can merge this quickly

Rocketknight1:
run-slow: esm

PR #44165, opened 2026-02-20
Fix how PreTrainedModel checks annotations on Python 3.14+

alexmalyshev:
With Python 3.14 and PEP 649, you can no longer expect `cls.__dict__["__annotations__"]` to contain annotations for the exact class, it will be loaded lazily and can make it seem like the class doesn't have any annotations.

The recommended tools all load annotations for the entire class hierarchy.  What we can do is to compute the annotation for a class and for its parent, then compare them.  If they're different, then we know we have a dedicated annotation on the child class.

### DIFF
--- a/src/transformers/modeling_utils.py
+++ b/src/transformers/modeling_utils.py
@@ -1222,27 +1222,42 @@ class PreTrainedModel(nn.Module, EmbeddingAccessMixin, ModuleUtilsMixin, PushToH
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
+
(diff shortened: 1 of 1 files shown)


Rocketknight1:
This one touches some very core code so cc maintainers @arthurzucker @cyrilvallez @vasqu

alexmalyshev:
@ArthurZucker @Cyrilvallez @vasqu -- Gentle nudge to take a look at this, thanks!

Cyrilvallez:
I see it has been superseded by https://github.com/huggingface/transformers/pull/44672 which provided a simpler fix!

PR #44166, opened 2026-02-20
Improve `has_similar_generate_outputs` assertions

tarekziade:
The CI does not output useful info on this flaky test - `tests.models.olmo.test_modeling_olmo.OlmoModelTest testMethod=test_generate_with_static_cache` and makes it harder to determine the root problem when not reproducible locally.

This patch improves this assertion, so we get full details when it fails.

Before (CI output):

```
AssertionError: False is not true
```

After (CI output would look like):

```
AssertionError: Generate outputs are not similar enough (atol=1e-05, rtol=1e-05).
    Sequence mismatch: 3/20 tokens differ (first at position 12).
    Batch index: 0
    Token at mismatch — output_1: 1542, output_2: 8903
    Score diff at first mismatch — max: 2.384186e-04, mean: 1.127943e-05
```

This will tell us immediately whether it's a tolerance issue (small diffs suggesting we might just need a slightly larger atol), a completely different generation path, or a specific batch element problem.

While I've changed all the callers in our tests, I kept the original function intact for backward compact (even though it might be overkill)


tarekziade:
run-slow: dia, gemma3n, glm_ocr, kosmos2, kyutai_speech_to_text, t5gemma, t5gemma2

tests/generation/test_utils.py
```diff
@@ -1452,6 +1448,18 @@ def test_generate_with_static_cache(self):
...
+            # left- nor right-padded, which can cause unstable static-vs-dynamic comparisons.
+            if "attention_mask" in inputs_dict:
+                attention_mask = inputs_dict["attention_mask"]
+                seq_len = attention_mask.shape[1]
+                valid_lengths = attention_mask.sum(dim=1).clamp_(min=1, max=seq_len)
+                position_ids = torch.arange(seq_len, device=attention_mask.device).unsqueeze(0)
+                inputs_dict["attention_mask"] = (position_ids >= (seq_len - valid_lengths).unsqueeze(1)).to(
+                    attention_mask.dtype
+                )
+                self.assertFalse(torch.any(inputs_dict["attention_mask"][:, -1] == 0))
```
Rocketknight1: This block feels like a significant change! Maybe it shouldn't be merged with a PR that just changes error reporting?

tarekziade: oops that leaked from another patch, good catch thx

tests/generation/test_utils.py
```diff
@@ -4924,25 +4917,78 @@ def has_similar_generate_outputs(output_1, output_2, atol=1e-5, rtol=1e-5) -> bo
...
+
+    Args:
+        output_1 (`GenerateOutput`): The first `generate` call output.
+        output_2 (`GenerateOutput`): The second `generate` call output.
+        atol (`float`, *optional*, defaults to 1e-5): The absolute tolerance for the scores.
+        rtol (`float`, *optional*, defaults to 1e-5): The relative tolerance for the scores.
+
+    Returns:
+        A boolean indicating whether the two generate outputs are similar.
+    """
```
Rocketknight1, line 4948: This seems ready to go, but we can probably drop the big docstrings for one-line functions, right? Especially in internal utils.

Rocketknight1 (approved):
One nit about the docstrings but this seems ready!

tests/generation/test_utils.py
tarekziade: besides the args the rest was the original doctsring. happy to reduce it though

tarekziade:
run-slow: dia, gemma3n, glm_ocr, kosmos2, kyutai_speech_to_text, t5gemma, t5gemma2

Issue #21110, opened 2023-01-13
Add support for BLIP and GIT in image-to-text and VQA pipelines


NielsRogge:
### Feature request

BLIP and GIT are 2 recent additions in the library, providing state-of-the-art performance for tasks like image captioning and visual question answering (VQA). GIT is even capable of video captioning and video QA.

Hence it makes sense to support them in our image-to-text and VQA pipelines.

### Motivation

Having support for better models in pipelines is very desired!

See also a request for it here: https://discuss.huggingface.co/t/support-for-different-models-in-text-to-image-pipeline/29504

### Your contribution

I can assist in adding support, see #18446 as a very similar case

atturaioe:
Hi @NielsRogge , can work on it?

NielsRogge:
Sure @atturaioe feel free to start working on it

pikaduck:
@NielsRogge is this issue still open for contribution?

NielsRogge:
Yes

RaghavPrabhakar66:
@NielsRogge If nobody is working on it, I would like to pick up the issue.

strankid:
I would like to pick the issue if its still available.

Tanmaypatil123:
@NielsRogge is this issue still open to contribute . I would like to work on it

NielsRogge:
Support for BLIP in the image-to-text pipeline has been added in #21904. GIT can be added as explained in [this comment](https://github.com/huggingface/transformers/issues/21514#issuecomment-1446420536), feel free to open a PR.

Support for the VQA pipeline still needs to be added for both, also there contributions are welcome.

sushmanthreddy:
@NielsRogge  can I work on this issue??

marechaux:
Hello @NielsRogge ! 

I would like to work on this issue (add support for VQA to GIT model) as a first contribution.

But before I start, I have a question :

Currently the only model implementing the VQA pipeline is `ViltForQuestionAnswering`, it does the task using [classification](https://github.com/huggingface/transformers/blob/4baa34c18f18274fe028ad5a5511ea3fba9eeece/src/transformers/models/vilt/modeling_vilt.py#L1079) 

However in [GIT paper](https://arxiv.org/abs/2205.14100) they say that : 

> For VQA, the input question is treated as a text prefix, and the answer is generated in an auto-regressive way. Furthermore, we present a new generation-based scheme for ImageNet classification, where the predicted labels come directly from our generative model without pre-defining the vocabulary.

[...]

PR #43325, opened 2026-01-16
🚨 More V5 pipeline cleanup

Rocketknight1:
More V5 pipeline cleanup, followup to #43256 and #43306:

(changelist updated Jan 26th)

- `question-answering` and `visual-question-answering` removed
- `image-to-image` removed
- Updated the default `text-generation` and `image-text-to-text` models
- Removed some references to `image-to-text`, `summarization`, `translation` and `text2text` in the tests that I missed in previous PRs
- Updated the migration guide to explain all of this!

**Note: This was intended for v5, but ended up a little late!**

PR #44170, opened 2026-02-20
Add GIT model support in VQA pipeline

veeceey:
## Summary
- Adds `GitForCausalLM` to `MODEL_FOR_VISUAL_QUESTION_ANSWERING_MAPPING_NAMES` so GIT can be used with the `visual-question-answering` pipeline
- Filters tokenizer outputs in VQA pipeline `preprocess` to only pass keys accepted by the model's `forward` method (fixes `token_type_ids` error with BERT-based tokenizers used by GIT)
- Adds `visual-question-answering` to GIT's `pipeline_model_mapping` in tests
- Adds both a fast unit test (tiny model from config) and a slow integration test for GIT VQA

Fixes part of #21110

## Motivation

GIT (Generative Image-to-text Transformer) supports VQA as a generative task where the question is used as a text prefix and the answer is generated auto-regressively. The VQA pipeline already handles generative models via `can_generate()`, so the main change is registering GIT in the model mapping and ensuring tokenizer compatibility.

## Test plan
- [x] Existing VQA pipeline tests pass (ViLT, BLIP2)
- [x] New `test_small_model_pt_git` test passes with tiny model created from config
- [x] New `test_large_model_pt_git` slow test added for `microsoft/git-base-textvqa`
- [x] All 6 non-slow VQA pipeline tests pass

### DIFF
--- a/src/transformers/models/auto/modeling_auto.py
+++ b/src/transformers/models/auto/modeling_auto.py
@@ -1337,6 +1337,7 @@ MODEL_FOR_VISUAL_QUESTION_ANSWERING_MAPPING_NAMES = OrderedDict(
     [
         ("blip", "BlipForQuestionAnswering"),
         ("blip-2", "Blip2ForConditionalGeneration"),
+        ("git", "GitForCausalLM"),
         ("vilt", "ViltForQuestionAnswering"),
     ]
 )
--- a/src/transformers/pipelines/visual_question_answering.py
+++ b/src/transformers/pipelines/visual_question_answering.py
@@ -1,3 +1,4 @@
+import inspect
 from typing import Union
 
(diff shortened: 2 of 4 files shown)


zucchini-nlp (commented):
Thanks for the PR, but we don't want to add generative VLMs to vqa task. They are alreayd supported in `image-text-to-text` and can be used there for 'captioning, vqa, etc.'

Ideally this pipeline can be deleted if there are no non-generative models in the mapping, maybe @Rocketknight1 is already planning to deprecate it 👀

zucchini-nlp:
Indeed, removing in https://github.com/huggingface/transformers/pull/43325. I'll close as we don't accept PRs for pipelines that will be removed soon

PR #44172, opened 2026-02-20
perf: Optimize SynthID logits processor batch index construction

tarekziade:
## Summary

I've noticed `test_synthidtext_watermark_processor_distributional_convergence_*` was our slowest tests in CircleCI 

This PR speeds up the slowest SynthID distributional convergence tests by optimizing a hot path in SynthIDTextWatermarkLogitsProcessor.

  ### What changed

  - File: src/transformers/generation/logits_process.py
  - In SynthIDTextWatermarkLogitsProcessor.__call__, replaced per-batch construction of all_indices:
      - Before: stack of batch_size separate torch.arange(...) tensors
      - After: single torch.arange(...) broadcast with .unsqueeze(0).expand(batch_size, -1)

  This is behavior-preserving and only changes how the dense continuation index tensor is built.

  ## Why

  test_synthidtext_watermark_processor_distributional_convergence_{0,1,2,3} in tests/generation/test_logits_process.py repeatedly call the processor in a tight loop (num_keys=1000, 5
  calls per key).
  The previous implementation rebuilt all_indices inefficiently on every call, causing millions of redundant torch.arange calls.

  ## Performance impact

  Side-by-side on the same machine:

  - pytest -q tests/generation/test_logits_process.py -k "distributional_convergence" --durations=10
      - Before: 39.73s
      - After: 20.03s
      - ~1.98x faster

  Microbenchmark (matching test loop shape):

  - Before: 10.704338s
  - After: 5.321214s
  - ~2.01x faster


ArthurZucker (approved):
good catch indeed! expand shares memory

PR #44158, opened 2026-02-19
fix bug with position_ids on qwen3-vl models, such that position_ids include text position

leopold-tzafon:
# Fix issue where `use_cache=False`, corrupts model Qwen3vl output.
Tested with: 
```
import torch
from transformers import Qwen3VLForConditionalGeneration, AutoProcessor

MODEL_NAME = "Qwen/Qwen3-VL-4B-Instruct"
DEVICE = "cuda"
DTYPE = torch.bfloat16
processor = AutoProcessor.from_pretrained(MODEL_NAME)
messages = [
    {"role": "user", "content": [{"type": "image", "image": "https://picsum.photos/200/300"}]},
]
inputs = processor.apply_chat_template(
    messages,
    tokenize=True,
    add_generation_prompt=True,
    return_dict=True,
    return_tensors="pt"
).to(DEVICE)


model = Qwen3VLForConditionalGeneration.from_pretrained(MODEL_NAME).to(DEVICE)

with torch.inference_mode():
    outputs_true = model(input_ids=inputs["input_ids"], pixel_values=inputs["pixel_values"], image_grid_thw=inputs["image_grid_thw"], use_cache=True)
    outputs_false = model(input_ids=inputs["input_ids"], pixel_values=inputs["pixel_values"], image_grid_thw=inputs["image_grid_thw"], use_cache=False)

kl = torch.nn.KLDivLoss(reduction="none", log_target=True)
print("KL divergence between outputs with use_cache=True and use_cache=False:")
print(kl(torch.nn.functional.log_softmax(outputs_true.logits, dim=-1), torch.nn.functional.log_softmax(outputs_false.logits, dim=-1)).mean().item())
```

# The issue
In certain situations, such as when use_cache is disabled, the Qwen3VLTextModel derives an attention mask based on the position_ids from the vision encoder. But, the issue is that text model uses the temporal dimension of the 3d position as a text position if on a 3d position is pass through. Which, is not compatible with `find_packed_sequence_indices` that assumes the positional ids are increase sequence of number with diff 1.
```
def find_packed_sequence_indices(position_ids: torch.Tensor) -> torch.Tensor | None:
    """
    Find the indices of the sequence to which each new query token in the sequence belongs when using packed
    tensor format (i.e. several sequences packed in the same batch dimension).

    Args:
        position_ids (`torch.Tensor`)
            A 2D tensor of shape (batch_size, query_length) indicating the positions of each token in the sequences.

    Returns:
[...]

PR #44173, opened 2026-02-20
fix(flaky): `test_generate_with_and_without_position_ids` in GLM ORC

tarekziade:
Fixes flaky GLM OCR generation behavior when 2D `position_ids` are passed explicitly.

Reproducible locally with:

```
pytest tests/models/glm_ocr/test_modeling_glm_ocr.py::GlmOcrModelTest::test_generate_with_and_without_position_ids --flake-finder --flake-runs=500
```

## Fix

We skip 2D from `GenerationTesterMixin.test_generate_with_and_without_position_ids` when model uses `get_rope_index`

## 15 models impacted

1. glm_ocr
2. glm4v
3. glm46v,
4. glm_image
5. qwen2_vl
6. qwen2_5_vl
7. qwen2_5_omni
8. qwen3_vl
9. qwen3_vl_moe
10. qwen3_5
11. qwen3_5_moe
12. qwen3_omni_moe
13. paddleocr_vl
14. ernie4_5_vl_moe
15. glm4v_moe


tarekziade:
run-slow: glm_ocr

src/transformers/models/glm_ocr/modeling_glm_ocr.py
```diff
@@ -1392,42 +1392,54 @@ def forward(
...
+                model_inputs["position_ids"] = torch.cat([text_positions, vision_positions], dim=0)
+            else:
+                can_compute_mrope = image_grid_thw is not None or video_grid_thw is not None
+                if can_compute_mrope:
+                    vision_positions, rope_deltas = self.model.get_rope_index(
+                        input_ids=model_inputs.get("input_ids"),
+                        image_grid_thw=image_grid_thw,
+                        video_grid_thw=video_grid_thw,
+                        attention_mask=model_inputs.get("attention_mask"),
+                    )
```
zucchini-nlp: For position ids, we need to manipulate in `self._prepare_position_ids_for_generate`. Usually position ids are prepared only once and at each decoding step we increment by 1 more position. We don't want to revert back this huge block of code

I am not 100% sure, but qwen-vl style models can pick up from text-only positions obtained from `super()._prepare_position_ids_for_generate`  and add vision positions on top

tarekziade: Right, thanks, this is completely overkill, I'll dig into a simpler fix

zucchini-nlp: Thanks! Maybe we can override the test for 3D rope models, if the test setup is completely off

tarekziade: I tweaked a bit the approach, I think it's cleaner that way

zucchini-nlp: btw, maybe this is same as https://github.com/huggingface/transformers/pull/44158? The user is fixing packed inputs in 3D positions as well

taking a look at the new diff 👁️

tarekziade:
run-slow: glm_ocr

src/transformers/models/glm_ocr/modeling_glm_ocr.py
```diff
@@ -1406,6 +1406,22 @@ def prepare_inputs_for_generation(
...
+                prep_kwargs = {
+                    "input_ids": input_ids,
+                    "attention_mask": attention_mask,
+                    "past_key_values": past_key_values,
+                    "image_grid_thw": image_grid_thw,
+                    "video_grid_thw": video_grid_thw,
+                }
+                position_ids = self._prepare_position_ids_for_generation(input_ids, prep_kwargs)
+            else:
+                position_ids = self._pack_position_ids_with_rope_deltas(position_ids)
```
zucchini-nlp: ah this is much cleaner indeed! I got the root reason now, and imo we just need to override the test for 3D rope  models. There are around 10 such models, everything that has a `self.get_rope_index` in modeling code

These models do not expect 2D positions and if users pass it as 2D, model is expected to output "garbage". We leave it on users to take care of their positions if they want custom `position_ids`

tarekziade: ah thanks that makes a lot of sense indeed, will refactor the patch

tarekziade: @zucchini-nlp I am thinking about a central skip in `GenerationTesterMixin.test_generate_with_and_without_position_ids` for models that rely on get_rope_index (3D RoPE path) -- WDYT?

or do we still want to exercise those paths even if it produces garbage ?

zucchini-nlp: skip also works for me, not super relevant feat for these models to support. I mostly rely on slow tests for them because dummy models anyway will produce garbage

tarekziade:
run-slow:  glm_ocr now, glm4v, glm46v, glm_image, qwen2_vl, qwen2_5_vl, qwen2_5_omni, qwen3_vl, qwen3_vl_moe, qwen3_5, qwen3_5_moe, qwen3_omni_moe, paddleocr_vl, ernie4_5_vl_moe, glm4v_moe

tarekziade:
run-slow: glm_ocr, glm4v, glm46v, glm_image, qwen2_vl, qwen2_5_vl, qwen2_5_omni, qwen3_vl, qwen3_vl_moe, qwen3_5, qwen3_5_moe, qwen3_omni_moe, paddleocr_vl, ernie4_5_vl_moe, glm4v_moe

tarekziade:
note that the idefics failure is yet another different flake I will fix in a separate branch

tarekziade:
run-slow: glm_ocr, glm4v, glm46v

tests/generation/test_utils.py
```diff
@@ -2261,17 +2261,31 @@ def test_forward_with_logits_to_keep(self):
...
+                hasattr(module, "get_rope_index")
+                for module in (
+                    model,
+                    getattr(model, "model", None),
+                    getattr(model, "language_model", None),
+                    getattr(model, "text_model", None),
+                )
+            )
+            if has_3d_rope_positions:
+                continue
```
zucchini-nlp, line 2280: nit: we can skip right away because usually all models in the set will be the same arch

zucchini-nlp (approved):
Thanks!

PR #44174, opened 2026-02-20
Expand model-structure lint rules with a fast AST-based, ruff-like framework

tarekziade:
This draft expands `utils/check_modeling_structure.py` into a rule-driven linter for model code, with new checks and tests, while keeping runtime very low. 

Key features:

- The checker is intentionally AST-only (no heavy imports/execution), so it stays fast (<3 seconds against all models)
- It's built in a CI/PR-friendly way, so when devs proposes new models it's easy to provide feedback automatically
- It provides a nice CLI UI to explain every rules, like how Ruff does with `ruff rule XXX`
- it's agentic-friendly

See `utils/check_modeling_structure_rules.toml` for the set of rules


tarekziade:
@ArthurZucker @ydshieh @Cyrilvallez  this is a draft for adding model linting

tarekziade:
listing and understanding rules:

```
✗ python utils/check_modeling_structure.py --list-rules
TRF001: PreTrainedModel subclasses that define __init__ must initialize a PreTrainedModel parent. (default: enabled)
TRF002: When modular_<name>.py exists, modeling_<name>.py must keep the generated banner markers. (default: enabled)
TRF003: Class-level config_class on <Model>PreTrainedModel should match <Model>Config naming. (default: enabled)
TRF004: base_model_prefix should be a non-empty canonical string when defined on PreTrainedModel classes. (default: enabled)
TRF005: forward(return_dict=...) should use return_dict in control flow or output construction. (default: enabled)
TRF006: supports_gradient_checkpointing=True should be paired with gradient checkpointing implementation hooks. (default: enabled)
TRF007: Custom tie_weights should route through output-embedding helpers or delegate to super().tie_weights(). (default: enabled)
TRF008: _no_split_modules, when defined, should be a list/tuple of non-empty strings. (default: enabled)
TRF009: Classes directly mixing GenerationMixin should provide generation input preparation hooks. (default: enabled)
TRF010: forward with cache arguments should reference cache control/state variables consistently. (default: enabled)
TRF011: self.post_init() in __init__ should remain at the end of initialization for PreTrainedModel classes. (default: enabled)
TRF012: Doc decorators on PreTrainedModel classes should avoid empty add_start_docstrings usage. (default: enabled)
TRF013: modeling_<name>.py should avoid importing implementation code from another model package. (default: disabled)
```

```
✗ python utils/check_modeling_structure.py --rule TRF010
TRF010

Summary: forward with cache arguments should reference cache control/state variables consistently.
Default enabled: yes

What it does

Checks forward signatures that expose cache arguments for usage of those arguments in method body.

Why is this bad?

Unused cache arguments can indicate incomplete caching support and inconsistent API behavior.

Example

def forward(self, x, past_key_values=None, use_cache=False):
    return x


Use instead:

def forward(self, x, past_key_values=None, use_cache=False):
    if use_cache:
        ...
    return x
```

Running all rules on all models:

```
✗ time make check-model-rules-all
python utils/check_modeling_structure.py
OK
make check-model-rules-all  2.84s user 0.13s system 94% cpu 3.148 total
```

Rocketknight1:
Hmm, I think some of these are unnecessary! TRF002 is redundant because the modular check will always notice when the `modeling` file is not equivalent to the generated output. Also, I'm unsure about TRF006/TRF009, I think this line:

```python
if has_external_non_generation_base or non_generation_bases:
      continue
```

will basically skip any class that inherits from `PreTrainedModel`, which is all generating classes, right?

tarekziade:
> `TRF002` - the modular check will always notice when the modeling file is not equivalent to the generated output

are you thinking about `utils/check_modular_conversion.py` - if so, wouldn't it make sense to convert that script as well into a standardized lint step that would run alongside others?

tarekziade:
> Also, I'm unsure about TRF006/TRF009,

good point,  I tightened them a bit, they now only skips when the non-generation parent is external/unknown

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+why_bad = "Editing generated files directly creates drift from modular sources and future regeneration will overwrite changes."
+bad_example = '''
+# Copyright ...
+# Licensed under the Apache License...
+# (no generated-file banner present)
+'''
+good_example = '''
+# This file was automatically generated from src/transformers/models/foo/modular_foo.py.
+# Do NOT edit this file manually as any edits will be overwritten ...
+'''
```
zucchini-nlp: modular already auto-generates here as well, imo a bit redundant

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+what_it_does = "Checks naming consistency between <Model>PreTrainedModel and config_class."
+why_bad = "Mismatched config_class can break loading, auto classes, and developer expectations."
+bad_example = '''
+class FooPreTrainedModel(PreTrainedModel):
+    config_class = BarConfig
+'''
+good_example = '''
+class FooPreTrainedModel(PreTrainedModel):
+    config_class = FooConfig
+'''
```
zucchini-nlp, line 16: not sure if we can enforce this. Most nested models have several configs and each of its backbones use different configs (e.g. Gemma3TextConfig vs Gemma3Config)

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+good_example = '''
+class FooPreTrainedModel(PreTrainedModel):
+    base_model_prefix = "model"
+'''
+
+[rules.TRF005]
+description = "forward(return_dict=...) should use return_dict in control flow or output construction."
+default_enabled = true
+allowlist_models = ["chinese_clip", "ovis2", "pixtral"]
```
zucchini-nlp: tested with `test_return_tuple` or smth, no? Do we want to lint-check it as well?

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+bad_example = '''
+def forward(self, x, return_dict=None):
+    return FooOutput(last_hidden_state=x)
+'''
+good_example = '''
+def forward(self, x, return_dict=None):
+    if not return_dict:
+        return (x,)
+    return FooOutput(last_hidden_state=x)
+'''
```
zucchini-nlp: i think this might be totally delegated to `capture_output/can_return_tuple` decorators in a few releases

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+why_bad = "Ad-hoc tying can miss edge cases and diverge from shared loading/saving behavior."
+bad_example = '''
+def tie_weights(self):
+    self.lm_head.weight = self.emb.weight
+'''
+good_example = '''
+def tie_weights(self):
+    super().tie_weights()
+    # or use get_output_embeddings/set_output_embeddings helpers
+'''
```
zucchini-nlp: how often does this happen? Haven't seen outstanding cases yet, since contrib would copy-paste from existing models in codebase

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+
+[rules.TRF008]
+description = "_no_split_modules, when defined, should be a list/tuple of non-empty strings."
```
zucchini-nlp, line 73: looks like a typing check, could be enforced to strict with hub-validators maybe 👀  (if that goes well with configs)

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+class FooLM(FooBackbone, GenerationMixin):
+    pass
+'''
+good_example = '''
+class FooBackbone:
+    def prepare_inputs_for_generation(self, input_ids, **kwargs):
+        return {"input_ids": input_ids}
+
+class FooLM(FooBackbone, GenerationMixin):
+    pass
```
zucchini-nlp: not really clean why, we have been pushing to delegate generation methods to the Mixin. So that each class doesn't have to override it and we can decrease maintaining costs

zucchini-nlp (commented):
Agreed with Matt that some of these checks are already covered by general unitests or `check_modular`, which makes it a bit redundant. I am not sure if we want to enforce checks with linter instead of unit-testing, because linter doesn't actually run the code -> less reliable than what we have now. And model contribs have to make CI 🟢  before actually marking the PR ready (in a perfect world)

Personally I liked the part about test file rules (`TRF101-TRF10`) which can be more useful. I am not sure if reviewers always look at test files, and we can flag those out if tests are skipped unnecessarily or not even added properly. Sometimes contribs can try to be smart by simply skipping tests to make CI green, which is also not good 🙃

tarekziade:
> Agreed with Matt that some of these checks are already covered by general unitests or `check_modular`, which makes it a bit redundant. I am not sure if we want to enforce checks with linter instead of unit-testing, because linter doesn't actually run the code -> less reliable than what we have now. And model contribs have to make CI 🟢 before actually marking the PR ready (in a perfect world)
[snip]
> Sometimes contribs can try to be smart by simply skipping tests to make CI green, which is also not good 🙃

I agree we shouldn’t replace runtime guarantees with static checks.

The intent here isn’t to move validation out of unit tests, but to add a faster static feedback loop. Linters don’t execute code, but they’re useful for catching structural issues early. In that sense, `check_modular_conversion` already behaves like a linter. This would just standardize it as a rule alongside others, likely reusing the same underlying logic.

Tests would remain the source of truth. The goal is additive: shorten the feedback loop without weakening safeguards.

And on the contributor side, I think we should assume good faith, but also design the system so that making CI green by skipping checks isn’t an easy path. Earlier static feedback helps there too.

utils/check_modeling_structure_rules.toml
tarekziade: ah! interesting. I got this rule from reading our guidelines. Do we have markers to know if a model is nested?

utils/check_modeling_structure_rules.toml
tarekziade: oh yeah, @tomaarsen pinged me about this. It's definitely something we can leverage.

But what happens if contributors don't use the new decorators? maybe the linter can control and suggest that if missing

utils/check_modeling_structure_rules.toml
tarekziade: This rule was inferred from looking at the API hierarchy and noticing that overriding `tie_weights()` can easily bypass the shared helpers or `super().tie_weights()`, which could lead to subtle inconsistencies.

I’m not aware of concrete past bugs caused by this, it’s more of a preventative structural safeguard.

Even if this hasn’t happened often in practice (likely because contributors start from existing models), it feels like the kind of thing that could silently drift over time. Since lint rules are relatively cheap, I thought it was reasonable to encode that expectation explicitly.

utils/check_modeling_structure_rules.toml
zucchini-nlp: yeah, we can insteed suggest to use decorators when neeeded. Maybe same can be for `autodocstring` instead of adding a docstring in `model.forward`

utils/check_modeling_structure_rules.toml
tarekziade: I agree with the direction of delegating generation behavior to GenerationMixin to reduce maintenance. The intent of this rule isn’t to force every model to re-implement generation hooks. It’s to prevent classes from exposing `.generate()` when they haven’t clearly satisfied the generation contract, which can lead to runtime failures or “half-working” generation paths.

Maybe the rule should be reframed to accept delegation, e.g. only require a “proof of generation readiness” somewhere in the local hierarchy (a base class / marker / known generation-ready parent), rather than requiring direct overrides on each model class.

maybe the why can become:

```
Mixing GenerationMixin should be done only when the class clearly declares generation readiness in its local hierarchy, otherwise .generate() may be exposed on classes that don’t meet the generation contract (cache handling, input prep, etc.), leading to runtime errors or partial support.
```

and the good example:
```
class FooBackbone(PreTrainedModel):  # already generation-ready base
    pass

class FooLM(FooBackbone, GenerationMixin):
    pass
```

utils/check_modeling_structure_rules.toml
tarekziade: I think it’s similar to the `capture_output` / `can_return_tuple case`:
we already rely on tests to validate behavior, but the linter can catch the structural requirement earlier and provide a more explicit hint (e.g. suggesting `@can_return_tuple` if it’s missing).

That way we fail fast with a clearer error message rather than relying solely on the test suite.

utils/check_modeling_structure.py
```diff
@@ -118,32 +256,1057 @@ def check_post_init(node: ast.AST, violations: list[str], file_path: str) -> lis
...
+    if enabled_rules is None:
+        enabled_rules = DEFAULT_ENABLED_TRF_RULES
+
+    violations: list[Violation] = []
+    source_lines = text.splitlines()
+    tree = ast.parse(text, filename=str(file_path))
+
+    for node in ast.walk(tree):
+        violations = check_init_weights(node, violations, file_path)
+        violations = check_post_init(node, violations, file_path)
```
Cyrilvallez, line 824: Those are the most important ones, and should become part of the main rules!

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
+[rules.TRF001]
+description = "PreTrainedModel subclasses that define __init__ must initialize a PreTrainedModel parent."
+default_enabled = true
```
Cyrilvallez: You mean call `super().__init__`? In general, any nn.Module has to call `super().__init__` otherwise `torch` will complain, so not sure this rule is needed - or it should be checked on almost every class. But we never had this issue anyway I believe, so could be removed for now

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+bad_example = '''
+class FooPreTrainedModel(PreTrainedModel):
+    def __init__(self, config):
+        self.config = config
+'''
+good_example = '''
+class FooPreTrainedModel(PreTrainedModel):
+    def __init__(self, config):
+        super().__init__(config)
+        self.config = config
```
Cyrilvallez: BTW, `self.config = config` is already called in `PreTrainedModel.__init__`, so redundant here

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+'''
+
+[rules.TRF002]
+description = "When modular_<name>.py exists, modeling_<name>.py must keep the generated banner markers."
+default_enabled = true
+allowlist_models = ["colpali", "dpt", "efficientloftr", "grounding_dino", "mask2former", "owlv2", "sam3", "segformer", "yolos"]
```
Cyrilvallez: Let's remove this one - modular consistency is already enforced (+ we have a large allowlist here, due to models using modular only for precossor or something, which makes it a bit incomplete)

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+'''
+good_example = '''
+class FooPreTrainedModel(PreTrainedModel):
+    config_class = FooConfig
+'''
+
+[rules.TRF004]
+description = "base_model_prefix should be a non-empty canonical string when defined on PreTrainedModel classes."
+default_enabled = true
+allowlist_models = ["lighton_ocr"]
```
Cyrilvallez: Here it's a nice one, and we should also check that if defined, there is a module with the same name defined such as `self.model = blabla` in the `__init__`!

I don't think we have a test for this actually, and it can lead to small bugs sometimes (but rarely). It will need to read modeling though, as __init__ may be omitted in modulars, and need to cross-ref for when it's defined only in the XXXPreTraiendModel itself, and not in the XXXModel(XXXPreTrainedModel)

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+    base_model_prefix = ""
+'''
+good_example = '''
+class FooPreTrainedModel(PreTrainedModel):
+    base_model_prefix = "model"
+'''
+
+[rules.TRF005]
+description = "forward(return_dict=...) should use return_dict in control flow or output construction."
+default_enabled = true
```
Cyrilvallez: Actually, we don't want that rule! We should instead forbid 
```
if not return_dict:
        return (x,)
    return FooOutput(last_hidden_state=x)
```

and force the usage of either `capture_ouput` or `can_return_tuple` instead!

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+    supports_gradient_checkpointing = True
+'''
+good_example = '''
+class FooPreTrainedModel(PreTrainedModel):
+    def _set_gradient_checkpointing(self, module, value=False):
+        module.gradient_checkpointing = value
+
+class FooModel(FooPreTrainedModel):
+    supports_gradient_checkpointing = True
+'''
```
Cyrilvallez: We don't want this one, a simple

```
class FooPreTrainedModel(PreTrainedModel):
```  

should be enough! We really don't want to override all the time

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+
+[rules.TRF007]
+description = "Custom tie_weights should route through output-embedding helpers or delegate to super().tie_weights()."
+default_enabled = true
+allowlist_models = ["data2vec", "hubert", "sew", "sew_d", "unispeech", "unispeech_sat", "wav2vec2", "wav2vec2_conformer", "wavlm"]
+
```
Cyrilvallez: Actually, here the rule we want is to NEVER override `tie_weights`. It's a good rule to enforce. Never allow overriding, and instead link to using the usual `_tied_weights_keys = {"target": "source"}` property to declare tied weights.
Any override of `tie_weights` will lead to many very bad consequences in term of loading, device_map computation etc in general

(A few older models override it for other purposes as they don'0t have tied weights, it's known and very bad and we really don't want to keep on doing that)

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+good_example = '''
+def tie_weights(self):
+    super().tie_weights()
+    # or use get_output_embeddings/set_output_embeddings helpers
+'''
+
+[rules.TRF008]
+description = "_no_split_modules, when defined, should be a list/tuple of non-empty strings."
+default_enabled = true
+allowlist_models = ["d_fine", "deformable_detr", "glm46v", "lw_detr", "pp_doclayout_v3", "rt_detr", "rt_detr_v2", "voxtral", "voxtral_realtime"]
```
Cyrilvallez, line 75: Why do we have so many skips here? We should probably patch them directly instead of skipping

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+
+[rules.TRF008.explanation]
+what_it_does = "Checks the shape of _no_split_modules when present."
+why_bad = "Malformed values can break device-map partitioning and sharding behavior."
+bad_example = '''
+_no_split_modules = [SomeLayerClass, ""]
+'''
+good_example = '''
+_no_split_modules = ["FooDecoderLayer", "FooAttention"]
+'''
```
Cyrilvallez: Nice

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+class FooBackbone:
+    pass
+
+class FooLM(FooBackbone, GenerationMixin):
+    pass
+'''
+good_example = '''
+class FooBackbone:
+    def prepare_inputs_for_generation(self, input_ids, **kwargs):
+        return {"input_ids": input_ids}
```
Cyrilvallez: Same as the gradient checkpointing one! We don't want that, there should not have needs to override in general! And if need there is, it should be done only on the class inheriting `GenerationMixin`, not the other ones!

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+def forward(self, x, past_key_values=None, use_cache=False):
+    if use_cache:
+        ...
+    return x
+'''
+
+[rules.TRF011]
+description = "self.post_init() in __init__ should remain at the end of initialization for PreTrainedModel classes."
+default_enabled = true
+allowlist_models = ["distilbert", "lxmert", "mt5", "pix2struct", "pop2piano", "switch_transformers", "t5"]
```
Cyrilvallez: Nice! Similar to before, instead of the allowlist, we should fix the bad models (can be done in another pr though)

utils/check_modeling_structure_rules.toml
```diff
@@ -0,0 +1,242 @@
...
+good_example = '''
+@add_start_docstrings("The Foo model.")
+class FooModel(FooPreTrainedModel):
+    ...
+'''
+
+[rules.TRF013]
+description = "modeling_<name>.py should avoid importing implementation code from another model package."
+default_enabled = false
+allowlist_models = []
```
Cyrilvallez: That's very nice, we had a few instances where this slipped through (modular is sometimes too good for our own good 😆). We then have to fix the modular structure etc when we stumble upon such cases

Cyrilvallez (commented):
Alright! Since this will become a very important tool for model addition and repo health, I'm trying to five feedback as frankly as I can (so sorry if it appears too blunt haha).

Similarly to other maintainer's review, I want to emphasize that this is NOT supposed to replace tests etc, but to give proper rules for *structure* and *good standards*. Those aspects are implicit for us maintainers, but are not for contributors. They come from our vision, and how the library evolved recently. Technically, everything could be covered by tests, but tests can always be skipped, and reviewers can miss that they were skipped. So, for essential aspects, it's much better to enforce as `consistency` as we do here, as this cannot be skipped and it gives much earlier/better feedback to contributors. For example, the 2 rules that already exist in this file (must call `post_init`, and `_init_weights` cannot use in-place ops) are absolutely ESSENTIAL, and we want something even stricter than tests.

For now, let's try to not add rules that are not really essential, or can clash a bit too much with current library state. In general concerning the PR, I really like the linter-like structure, and the clear rules with examples etc. But I think in the current state there are maybe too many rules to start with, and some of them are not helping us.

What I think we want for now:
- Every PreTrainedModel MUST call `post_init` (already there) -> no way to override, i.e. no `noqa TRF001` for this one)
- `_init_weights` cannot use in-place ops (already there) -> no way to override, i.e. no `noqa TRF002` for this one)
- `base_model_prefix` one (current TRF004), with the tweaks I commented
- never allow to override `tied_weights` (current TRF007 with the tweaks I commented) -> no way to override this one
- Current TRF008
- Current TRF010
- Current TRF011
- Current TRF012
- Current TRF013

A very important rule I would like to add if possible (maybe in another PR for simplicity):
- ensure that all classes in `modeling.py` are either in `__all__`, or are used somewhere. Indeed, sometimes using modular wrongly (i.e. forgetting to overwrite a tiny class attribute or something) can lead to silence creation of a few classes in modeling, that are never used, and they bloat the modeling like crazy. This should be doable to check, and to link back to model-adder to tell them to double check their modular, so that those classes are not automatically imported from the dependency graph.
- If `output_xxx` in the signature -> replace with `capture_output` decorator!

For this 2 new proposed rules, the benefits of a linter vs simple tests are obvious, as for the 1st rule those classes are never used and we cannot really test if they should be there or not, but the linter can very easily detect them and warn that the modular should be updated! For the second rule, it's the same, we cannot easily test for something that should or should not be there, but we can tell contributors that we WANT the `capture_output` decoratir whenever possible instead of old `capture_xxx` argument in every signature!

tarekziade:
> Alright! Since this will become a very important tool for model addition and repo health, I'm trying to give feedback as frankly as I can (so sorry if it appears too blunt haha).

Thanks for the thorough review @Cyrilvallez — very helpful feedback.

Your comments make sense: the rules should enforce best practices and prevent real issues, not duplicate safeguards already provided by the base classes or existing tooling.

Here's where I land on each point:

# Rules to drop

- **TRF001** / Agreed.  
I added it to guard against the subtle case where someone calls `nn.Module.__init__` directly and skips `PreTrainedModel.__init__`, but in practice torch already catches missing `super().__init__`, and the bypass scenario hasn't happened in the wild. Removing.
 
- **TRF002**  / Agreed.  
This was meant as a cheap early check to catch edits to generated files before `check_modular_conversion`, but it's redundant with the existing infrastructure and the allowlist already shows it doesn’t fit cleanly. Removing.

- **TRF006** / Agreed.  
I assumed `supports_gradient_checkpointing = True` without explicit hooks could break at runtime, but you're right that `PreTrainedModel` provides sensible defaults. Forcing overrides is the wrong pattern. Removing.

- **TRF009** / Same reasoning.  
I added it to ensure `prepare_inputs_for_generation` was defined when mixing `GenerationMixin`, but the mixin already provides working defaults. Removing.

# Rules to rework / strengthen

**TRF005**

I like the direction you suggest. Instead of checking that `return_dict` is respected, the rule should forbid the old `if not return_dict: return (...)` pattern and enforce the `capture_output` / `can_return_tuple` decorators. I’ll rework it accordingly.

**TRF007**

Will strengthen this to a hard ban on overriding `tie_weights`, with no `noqa` escape. Contributors should use `_tied_weights_keys` instead.

**TRF004**

Will add the check that `self.<prefix>` is actually assigned in `__init__`.

**TRF008 / TRF010 / TRF011**

Agreed on fixing the offending models instead of relying on allowlists. I’ll start addressing those (possibly in a follow-up PR if it gets too large).

# New rules

These sound valuable, thanks.

- Check for unused classes in `modeling.py` (not in `__all__` and not referenced elsewhere) to catch modular bloat.
- Enforce `capture_output` when `output_*` arguments appear in `forward` signatures.

utils/check_modeling_structure.py
tarekziade: Agreed, these remain as core, always-on rules that cannot be disabled or noqa'd.

utils/check_modeling_structure_rules.toml
tarekziade: Removed , torch already enforces this, so the rule was unnecessary.

utils/check_modeling_structure_rules.toml
tarekziade: Good catch, the whole rule was removed, so this is no longer relevant.

utils/check_modeling_structure_rules.toml
tarekziade: Removed as suggested

utils/check_modeling_structure_rules.toml
tarekziade: The basic validation is in place as `TRF002`. The cross-reference check (verifying `self.<prefix>` exists in `\_\_init\_\_`) is a great idea,  will add in a follow-up PR as it requires reading the modeling file and handling cases where `_\_\init\_\_` is only in the `PreTrainedModel` vs the `Model` subclass.

utils/check_modeling_structure_rules.toml
tarekziade: Thanks! This should catch modular mishaps early.

utils/check_modeling_structure_rules.toml
tarekziade: Agreed, will fix in a follow-up PR.

utils/check_modeling_structure_rules.toml
tarekziade: Removed as suggested.

utils/check_modeling_structure_rules.toml
tarekziade: The allowlist is a temporary measure, these models have `_no_split_modules = None` or other invalid values. 
Fixing them directly is planned as a follow-up PR to avoid bloating this one. (or we could relax and allow `None`)

utils/check_modeling_structure_rules.toml
ArthurZucker: we do, the config has subconfigs!

utils/check_modeling_structure_rules.toml
ArthurZucker, line 1: I would add:

- All public classes should start with `<Model>` but maybe it was discussed not to enforce that?
- Test files HAVE to have integration test class
- einsum are to be avoided in modeling code (maybe too many exceptions?)
- kwargs types with transformers kwargs 
- kwargs passed to attention for vllm support at least for all language models

but with your work should be easy to add!

utils/check_modeling_structure.py
```diff
@@ -118,32 +248,700 @@ def check_post_init(node: ast.AST, violations: list[str], file_path: str) -> lis
...
+    if TRF005 in enabled_rules:
+        violations = check_no_split_modules_shape(tree, file_path, source_lines, violations)
+    if TRF006 in enabled_rules:
+        violations = check_cache_argument_usage(tree, file_path, source_lines, violations)
+    if TRF007 in enabled_rules:
+        violations = check_post_init_order(tree, file_path, source_lines, violations)
+    if TRF008 in enabled_rules:
+        violations = check_doc_decorator_usage(tree, file_path, source_lines, violations)
+    if TRF009 in enabled_rules:
+        violations = check_single_file_policy_imports(tree, file_path, source_lines, violations)
```
ArthurZucker: we could just iterate on enabled rules and have a mapping from ENUM -> check func to scale ?

ArthurZucker (approved):
That's actually very nice: empowering the dev to add more rules themselves!! 

A few of them are missing IMO, but as I review I can just add them, this is great, simple and elegant! 🤗 TY

PR #44176, opened 2026-02-20
[`Mamba`] Fix kernel loading

vasqu:
Our kernel loading is incompatible with the original packages as they do not expose the same import structure:
- Kernels seem to expose things in the init (and not in the original path)
- Original packages seem to expose only within their original path (and not in the init, e.g. [see here](https://github.com/state-spaces/mamba/blob/v2.3.0/mamba_ssm/__init__.py) does not contain `selective_state_update`)

This is hard to test as you need to check within multiple different environments what packages are installed and whether things are properly loaded --> mamba2 was silently ignored even though the og pkg was there, mamba1 only luckily forced the issue leading to this discovery

### DIFF
--- a/src/transformers/models/mamba/modeling_mamba.py
+++ b/src/transformers/models/mamba/modeling_mamba.py
@@ -33,7 +33,7 @@ from ...utils import (
     auto_docstring,
     logging,
 )
-from ...utils.import_utils import is_mambapy_available, is_torchdynamo_compiling
+from ...utils.import_utils import is_mambapy_available, is_torchdynamo_compiling, resolve_internal_import
 from .configuration_mamba import MambaConfig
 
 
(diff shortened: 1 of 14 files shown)


src/transformers/models/mamba/modeling_mamba.py
```diff
@@ -199,18 +199,14 @@ def __init__(self, config: MambaConfig, layer_idx: int):
...
         global mamba_ssm, selective_state_update, selective_scan_fn, mamba_inner_fn
         mamba_ssm = lazy_load_kernel("mamba-ssm")
-        selective_state_update, selective_scan_fn, mamba_inner_fn = (
-            (mamba_ssm.selective_state_update, mamba_ssm.selective_scan_fn, mamba_ssm.mamba_inner_fn)
-            if mamba_ssm is not None
-            else (None, None, None)
-        )
+        selective_state_update = getattr(mamba_ssm, "selective_state_update", None)
+        selective_scan_fn = getattr(mamba_ssm, "selective_scan_fn", None)
+        mamba_inner_fn = getattr(mamba_ssm, "mamba_inner_fn", None)
```
vasqu: Same pattern as in mamba2 https://github.com/huggingface/transformers/blob/9a7d817885aab0728cd4894798790e47607920c7/src/transformers/models/mamba2/modeling_mamba2.py#L261-L270

Not sure why this was done differently here cc @MekkCyber fyi

src/transformers/utils/import_utils.py
```diff
@@ -77,6 +77,36 @@ def _is_package_available(pkg_name: str, return_version: bool = False) -> tuple[
         return package_exists
 
 
+def resolve_internal_import(module: ModuleType | None, chained_path: str) -> Callable | ModuleType | None:
+    """
+    Check if a given `module` has an internal import path as defined by the `chained_path`.
+    This can either be the full path (not exposed in `__init__`) OR the last part of the chain (exposed in `__init__`).
```
vasqu, line 83: We are stuck with this either way now because we need BC

MekkCyber (approved):
Thanks for fxing 🤗

PR #44177, opened 2026-02-20
[`Flash Attn`] Enable compatible implementations

vasqu:
As per title, spiritual successor to #44081

Why? Because as is
- Only defaults for fa2/fa3, not on other requested kernels
- Limits implementations to one kernel/implementation while I suspect that there will be multiple viable versions (in the future), e.g. FA4 integrates attention sinks natively

### DIFF
--- a/src/transformers/models/gpt_oss/configuration_gpt_oss.py
+++ b/src/transformers/models/gpt_oss/configuration_gpt_oss.py
@@ -118,22 +118,5 @@ class GptOssConfig(PreTrainedConfig):
         self.eos_token_id = eos_token_id
         super().__init__(**kwargs)
 
-    def __setattr__(self, key, value):
-        """
-        Overwritten to allow checking for the proper attention implementation to be used.
-
-        Due to `set_attn_implementation` which internally assigns `_attn_implementation_internal = "..."`, simply overwriting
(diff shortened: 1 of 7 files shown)


src/transformers/models/gpt_oss/configuration_gpt_oss.py
```diff
@@ -118,22 +118,5 @@ def __init__(
         self.eos_token_id = eos_token_id
         super().__init__(**kwargs)
 
-    def __setattr__(self, key, value):
```
vasqu, line 121: No longer needed because we have this natively with the fallback instead

tests/models/gpt_oss/test_modeling_gpt_oss.py
```diff
@@ -71,44 +70,13 @@ class GptOssModelTest(CausalLMModelTest, unittest.TestCase):
     model_split_percents = [0.5, 0.6]
     model_tester_class = GptOssModelTester
 
-    @require_kernels
-    @require_flash_attn
-    @pytest.mark.flash_attn_test
-    @require_torch_gpu
-    def test_initialization_raises_error_for_flash_attn(self):
```
vasqu, line 78: Same here, we properly fallback instead now

ArthurZucker (approved):
sgtm

PR #44180, opened 2026-02-20
fix(flaky): idefics generate cache flake

tarekziade:
# What does this PR do?

Fixes a flaky test in IdeficsForVisionText2TextTest::test_generate_continue_from_inputs_embeds.

The flakiness can be reproduced with:

```
pytest -q -p no:rerunfailures --flake-finder --flake-runs=20 \
  tests/models/idefics/test_modeling_idefics.py::IdeficsForVisionText2TextTest::test_generate_continue_from_inputs_embeds
```

There are three cache-equality checks in test_modeling_idefics.py. This is the only one exhibiting nondeterministic behavior.

This PR relaxes the assertion for this specific test to validate the cache structure rather than strict equality, which appears to be sensitive to non-deterministic internals while still preserving the functional guarantees we care about.

@Rocketknight1 does this approach look reasonable to you?


tarekziade:
run-slow: idefics

Rocketknight1:
Hmm, this one seems a bit weird to me - it's just verifying the cache shape! Is it really normal that the cache should have totally different values without affecting the output?

tarekziade:
> Hmm, this one seems a bit weird to me - it's just verifying the cache shape! Is it really normal that the cache should have totally different values without affecting the output?

Yeah this was mostly to relax this specific test. After more digging, I can see that the cache is initially completely similar then there's a drifting with the forward passes.

tweaking the test so it's strict, then it adds more tolerance is another approach (see the changes I've made) - but I don't know if this just hides an issue or if this drift is expected with tiny floating-point diffs?

Rocketknight1 (approved):
I'd still like to go back and dig into exactly what's happening with idefics, but in the short term, fixing the CI is probably more valuable! The model's been around enough that it's probably safe.

tarekziade:
run-slow: idefics

PR #44181, opened 2026-02-20
[core] 🚨 Completely remove cache positions

Cyrilvallez:
# What does this PR do?

As per the title! Follow up of https://github.com/huggingface/transformers/pull/44130 and https://github.com/huggingface/transformers/pull/44226.
Finally remove the `cache_position` everywhere (not ALL models, but all the recent/most important models that use modular). This means we still create them and pass them around in generate, but they are ignored by the models. I will gradually remove them from all models, and remove them from `generate` once this is done.

This PR basically tweaks the logic of `cache_utils.py` and `masking_utils.py` to not rely on the `cache_position`, and then remove the argument from the models' `forward`s (not a regression as they are absorbed by the `**kwargs` for all those models).

## Motivation

`cache_position` are basically not needed, as our Cache classes already and masking primitives contains all the necessary informations. We can simply recreate them quickly in the StaticCache when needed. Removing them will make all modeling files much easier to read and understand. Indeed, people are often confused by `cache_position` that seem to only be a 1D version of `position_ids` (thus it seems fully redundant) when you read modeling files, even though it's not exactly the case and `cache_position` don't take padding into account. Also, will allow to make `generate` much easier for input preparation once they will be removed from all models.

## Compile compatibilty

I made EXTRA SURE that we do not have any regresions in terms of `compile`-compatibility: we have exactly the same scope of compatibility as before, i.e. full compatibility (fullgraph=True, dynamic=False, cudagraphs) for StaticLayer without any recompiles, and (fullgraph=True, cudagraphs) for StaticSlidingWindowLayer (it can work with fullgraph=False, but will recompile every iteration - otherwise it recompiles only once to make the internal int a dynamic `Symint`, and once again if it changes cache-regime (i.e. cache becomes full etc). There is no way around that as StaticSlidingWindowLayer has data-dependent control flows that are unavoidable. It's exactly the same currently on `main`.

## Breaking changes

This PR is not really breaking. The 🚨 marker is only for the following detail:

The only breaking change in this PR is in the `masking_utils` API, as `cache_position` become optional everywhere in the `create_xxx_mask` functions (they are not used anymore), and thus `past_key_values` need to be passed as kwarg now due to the position of the args in the signature, as we cannot have an arg without default value following an arg with default value (this was already the case everywhere in Transformers).

The internal functions `sdpa_mask`, `eager_mask` etc as well move from having `cache_position` in their signature to having `q_length` and `q_offset` instead. Those should be private anyway, but you never know.

## Review pointers

The easiest way to review this PR is to only look at the very few files that are not modeling files, especially `cache_utils.py` and `masking_utils.py`. Then, all modeling files are basically the same: just remove the arg `cache_position` everywhere (they are absorbed in the `**kwargs` so no regression, they can still be externally passed, and they are still technically passed by `generate`, just not used)


Cyrilvallez:
run-slow: afmoe, apertus, arcee, aria, audioflamingo3, bamba, bitnet, cohere, cohere2, csm, cwm, deepseek_v2, deepseek_v3, dia, diffllama, llama, t5gemma, lfm2

Cyrilvallez:
Bamba and deepseek_v3 tests are failing the exact same way on this PR than on main. I'm not sure why the CI report say they fail in a different way. I checked locally and everything is perfectly identical!

src/transformers/integrations/executorch.py
```diff
@@ -521,14 +523,22 @@ def __init__(
         # Initialize the static cache
         self.model = model
         self.static_cache = StaticCache(max_cache_len=max_cache_len, config=config)
+        # Since StaticSlidingWindow have dynamic control flow that cannot be avoided, we have to replace them here by
+        # simple StaticLayer... It means that any generation beyond the window is unfortunately unsupported
+        for i, layer in enumerate(self.static_cache.layers):
+            if isinstance(layer, StaticSlidingWindowLayer):
+                self.static_cache.layers[i] = StaticLayer(layer.max_cache_len)
```
vasqu, line 530: Do we add something in the docs to clarify, this seems like something unseemingly hard to catch by outsiders - and I don't think we will remove this limitation any time soon 😓

src/transformers/cache_utils.py
```diff
@@ -42,11 +41,11 @@ def lazy_initialization(self, key_states: torch.Tensor, value_states: torch.Tens
 
     @abstractmethod
     def update(
-        self, key_states: torch.Tensor, value_states: torch.Tensor, cache_kwargs: dict[str, Any] | None = None
+        self, key_states: torch.Tensor, value_states: torch.Tensor, *args, **kwargs
```
vasqu, line 44: Do we really need the args/kwargs signature here? Does kwargs suffice to be BC? 

Just not a fan of the *args tbh

src/transformers/cache_utils.py
```diff
@@ -73,7 +72,11 @@ def reset(self) -> None:
             self.values.zero_()
         # This attribute is set on several Layers
         if hasattr(self, "cumulative_length"):
-            self.cumulative_length = 0
+            # It can either be an int for dynamic layers, or a tensor for static layers
+            if isinstance(self.cumulative_length, int):
+                self.cumulative_length = 0
+            else:
+                self.cumulative_length.zero_()
```
vasqu, line 80: Could we default to tensor instead or too bothersome?

vasqu: Or python ints always, not sure about the impacts.

More of a nit tbh

src/transformers/cache_utils.py
```diff
@@ -484,7 +473,11 @@ def get_mask_sizes(self, cache_position: torch.Tensor) -> tuple[int, int]:
 
     def get_seq_length(self) -> int:
         """Returns the sequence length of the cached states."""
-        return self.cumulative_length
+        return self.cumulative_length_int
+
+    def reset(self):
+        super().reset()
+        self.cumulative_length_int = 0
```
vasqu, line 481: The `_int` suffix is a bit weird, I know you mean to make it explicit but then it won't match with a lot of the other layer types.

src/transformers/masking_utils.py
```diff
@@ -366,8 +366,9 @@ def _non_vmap_expansion_sdpa(
 
 def sdpa_mask(
     batch_size: int,
-    cache_position: torch.Tensor,
+    q_length: int,
     kv_length: int,
+    q_offset: int = 0,
     kv_offset: int = 0,
```
vasqu, line 372: Even if it is internal, we should definitely mark with 🚨 as it's breaking signature

src/transformers/masking_utils.py
```diff
@@ -743,10 +761,10 @@ def _preprocess_mask_arguments(
     config: PreTrainedConfig,
     inputs_embeds: torch.Tensor,
     attention_mask: torch.Tensor | BlockMask | None,
-    cache_position: torch.Tensor,
     past_key_values: Cache | None,
     position_ids: torch.Tensor | None,
     layer_idx: int | None,
+    encoder_hidden_states: torch.Tensor | None = None,
```
vasqu, line 783: docs missing for this

src/transformers/masking_utils.py
```diff
@@ -779,35 +795,47 @@ def _preprocess_mask_arguments(
...
+        kv_length, kv_offset = past_key_values.get_mask_sizes(q_length, layer_idx)
     # Otherwise, we infer based on our input
     else:
+        q_offset = 0
         # 1. Rely on input directly
         if attention_mask is None:
-            kv_length, kv_offset = inputs_embeds.shape[1], 0
+            # For encoder-decoders, use encoder_hidden_states to infer kv_length if provided
+            kv_length = encoder_hidden_states.shape[1] if encoder_hidden_states is not None else q_length
+            kv_offset = 0
```
vasqu, line 857: Yea that's essentially the trick I used indirectly. A bit awkward but it has to be done 👀

src/transformers/masking_utils.py
```diff
@@ -827,15 +855,16 @@ def _preprocess_mask_arguments(
...
 
 
 @deprecate_kwarg("input_embeds", version="5.6.0", new_name="inputs_embeds")
 def create_causal_mask(
     config: PreTrainedConfig,
     inputs_embeds: torch.Tensor,
     attention_mask: torch.Tensor | None,
-    cache_position: torch.Tensor,
+    cache_position: torch.Tensor | None = None,  # not used anymore but kept for BC
+    *,
```
vasqu, line 886:
```suggestion
```
don't see a reason to add it?

src/transformers/masking_utils.py
```diff
@@ -977,19 +1008,15 @@ def create_bidirectional_mask(
...
-        config, embeds, attention_mask, cache_position, None, None, 0
+    early_exit, attention_mask, _, q_length, kv_length, q_offset, kv_offset = _preprocess_mask_arguments(
+        config, inputs_embeds, attention_mask, None, None, 0, encoder_hidden_states
     )
     if early_exit:
         return attention_mask
 
-    batch_size, dtype = embeds.shape[0], embeds.dtype
+    embeds = encoder_hidden_states if encoder_hidden_states is not None else inputs_embeds
+    batch_size, dtype, device = embeds.shape[0], embeds.dtype, embeds.device
```
vasqu, line 1038: I dont think we need this ternary at all anymore? Both embeds should have the same factory data, e.g. device,dtype,batch

src/transformers/masking_utils.py
vasqu: happens elsewhere, keeping it here only

src/transformers/models/afmoe/modular_afmoe.py
vasqu, line 1: I love it, feels good to see

vasqu (commented):
Since I'm on the train, it's a bit hard to review but iiuc we gradually go to remove cache positions and focused on a subset of important models for now

Can you run slow for bert/bart: just me being a bit too anxious about these

src/transformers/integrations/executorch.py
Cyrilvallez: We could, but note that this is ALREADY the case on main, and it looks like nobody noticed/raised issue... It would generate garbage beyond the sliding window, it's just explicit in the code now!
To solve it, we could either have a mimimal version of sliding cache for export, which would work ONLY for decoding (i.e. 1 new tokens all the time, without being able to feed more than 1 token after prefill), or we could juste use full cache with proper sliding masking, but we loose some compute

src/transformers/cache_utils.py
Cyrilvallez: Yes we need it - a lot of models pass `cache_kwarg` as an arg unfortunately, and other as a kwarg like `cache_kwargs=cache_kwargs` 🥲

src/transformers/cache_utils.py
Cyrilvallez: It's easier with ints for dynamic caches, and static ones REALLY need tensors for compile with cudagraphs!

src/transformers/integrations/executorch.py
vasqu: Yea, still think it would be nice to clarify somewhere. True, it's also broken in main, maybe out of scope for this PR

src/transformers/cache_utils.py
Cyrilvallez: It's because this LayerCache has both - the base `cumulative_length` is a Tensor (to be able to use cudagraph in the regime when the cache is not yet full),  and `cumulative_length_int` is the equivalent but as a python int, to avoid data-dependent branching

src/transformers/masking_utils.py
Cyrilvallez: It's because now we want it to be optional, but the next argument `past_key_values` was not optional either, and we don't really want to make it optional (args with default value cannot be in front of other args without default 🥲) - this way it makes it clear that ? past_key_value` has to be passed over

src/transformers/cache_utils.py
vasqu: Ah gotcha, didn't really came through the diff for me but makes sense

src/transformers/masking_utils.py
Cyrilvallez: Probably not, wasn't entirely sure as sometimes with the device_map they can end up on different devices - let's keep it for now and see after the PR is merged if we can remove!

Cyrilvallez:
run-slow: bert, bart

src/transformers/models/audioflamingo3/modeling_audioflamingo3.py
```diff
@@ -582,22 +581,20 @@ def forward(
...
         # Overwritten -- we should not pass input_features when we are in cached decoding stage
 
         input_features = kwargs.pop("input_features", None)
         input_features_mask = kwargs.pop("input_features_mask", None)
-        cache_position = kwargs.get("cache_position")
 
         model_inputs = super().prepare_inputs_for_generation(*args, **kwargs)
 
-        if cache_position is not None and model_inputs["cache_position"][0] == 0:
+        if is_first_iteration:
```
zucchini-nlp: or if not using cache, because when generating without cache we always need the audio

zucchini-nlp:
Can we also update/delete docs where cache position is mentioned, such as https://huggingface.co/docs/transformers/v5.2.0/en/cache_explanation#cache-position?

Cyrilvallez:
Updated audioflamingo, thanks the heads-up @zucchini-nlp! For the doc, I think it's best to wait for more models to remove them before deleting!

src/transformers/models/t5gemma/modular_t5gemma.py
ArthurZucker, line 1: well nice, would be better in another PR but works!

src/transformers/cache_utils.py
ArthurZucker: what about dynamic + compile 🙃

src/transformers/cache_utils.py
```diff
@@ -297,29 +292,27 @@ def lazy_initialization(self, key_states: torch.Tensor, value_states: torch.Tens
...
-        # breaks when updating the cache. However, it is not supported when tracing the graph, so we skip it in this case.
-        # As prefill should never be compiled, this is not an issue and it will still be run (except when users compile
-        # prefill explicitly, but this should be avoided!)
+        # breaks or cudagraph skips due to inplace mutations when updating the cache. However, it is not supported when
+        # tracing the graph, so we skip it in this case. As prefill should never be compiled, this is not an issue and it
+        # will still be run (except when users compile prefill explicitly, but this should be avoided!)
         if not is_torchdynamo_compiling():
             torch._dynamo.mark_static_address(self.keys)
             torch._dynamo.mark_static_address(self.values)
+            torch._dynamo.mark_static_address(self.cumulative_length)
```
ArthurZucker, line 304:
```suggestion
            torch._dynamo.mark_static_address(self.cumulative_length) # MEGA IMPORTANT
```

src/transformers/cache_utils.py
```diff
@@ -328,12 +321,11 @@ def update(
...
-        # in which case we should copy the whole Layer (key_states.shape[-2] == self.max_cache_len)
-        cache_position = cache_kwargs.get("cache_position") if cache_kwargs is not None else None
-        cache_position = (
-            cache_position if cache_position is not None else torch.arange(key_states.shape[-2], device=self.device)
-        )
+        # Create a tensor to slice the static kv at the correct indices
+        kv_length = key_states.shape[-2]
+        cache_position = torch.arange(kv_length, device=self.device) + self.cumulative_length
+        # Note that has to be performed in-place, as we have a static address that we need to keep
+        self.cumulative_length.add_(kv_length)
```
ArthurZucker, line 329: We should almost have AST rules for anything that touches `self.cumulative_length` to use add_ and etc

src/transformers/cache_utils.py
```diff
@@ -431,29 +413,37 @@ def update(
...
                 full_value_states = value_states
             else:
-                full_key_states = torch.cat((self.keys[:, :, :cumulative_length, :], key_states), dim=-2)
-                full_value_states = torch.cat((self.values[:, :, :cumulative_length, :], value_states), dim=-2)
+                full_key_states = torch.cat((self.keys[:, :, :current_length, :], key_states), dim=-2)
+                full_value_states = torch.cat((self.values[:, :, :current_length, :], value_states), dim=-2)
         else:
+            # Note: very important to use the tensor version of the cumulative length here, as otherwise cudagraphs
+            # (triggered by mode="reduced_overhead") will lead to random crashes, as the int would be overwritten
+            cache_position = torch.arange(kv_length, device=self.device) + self.cumulative_length
```
ArthurZucker, line 436: its annoying to me that we have to allocate new memory all the time here..... 😿

src/transformers/masking_utils.py
```diff
@@ -569,21 +576,25 @@ def eager_mask(
         use_vmap (`bool`, optional):
             Whether to use `vmap` during the mask construction or not. Allows powerful custom patterns that may not be
             index-based (for the cost of speed performance). By default `False`.
+        device (`torch.device` or `str`, optional):
+            An optional device to create the mask on.
     """
     # The masks for eager attention are simply boolean mask from sdpa, casted to 0 and -inf
     _ = kwargs.pop("allow_is_causal_skip", None)
     _ = kwargs.pop("allow_torch_fix", None)
```
ArthurZucker, line 592: should we just deprecate it for 2 releases at lease (cache postition as arg?)

src/transformers/models/zamba2/modular_zamba2.py
```diff
@@ -122,18 +122,6 @@ def __init__(
...
-        self, layer_idx: int, new_conv_state: torch.Tensor, cache_position: torch.LongTensor
-    ) -> torch.Tensor:
-        conv_state = self.conv_states[layer_idx]
-        cache_position = cache_position.clamp(0, self.conv_kernel_size - 1)
-
-        conv_state = conv_state.roll(shifts=-1, dims=-1)
-        conv_state[:, :, cache_position] = new_conv_state.to(conv_state.device)
-        self.conv_states[layer_idx].zero_()
-        self.conv_states[layer_idx] += conv_state
-        return self.conv_states[layer_idx]
```
ArthurZucker, line 135: this is a good catch but unrelated no?

ArthurZucker (commented):
I like this, but lets update the PR so the community understands why we are doing this breaking change (it is kinda breaking for the mask API). let's add our strong motivations please!

Also we could deprecate without breaking for some of the changes

src/transformers/cache_utils.py
Cyrilvallez: It's basically free - it's a tensor of 1 element 99% of the time, and even when it's not it's always super small!

src/transformers/models/zamba2/modular_zamba2.py
Cyrilvallez: Yes, fully unrelated, but I stumbled upon it and it's never used anywhere...

src/transformers/masking_utils.py
Cyrilvallez: Done! Until v5.6!

src/transformers/cache_utils.py
Cyrilvallez: Dynamic layers will never be able to use cudagraphs anyway as the kv tensors will change shape dynamically! However, I have strong hopes that we will be able to make them compile-compatible with dynamic=True/None and mode="default"!

ArthurZucker (approved):
LGTM 🤗

PR #44185, opened 2026-02-20
[docs] trainer part 1

stevhliu:
part 1 of refactoring the `Trainer` docs

- restructure the `toctree` a bit to accommodate new sections and docs
- slim down `trainer.md` to be a clearer entry point (will expand the `## Next steps` section as we continue for better navigation). everything else here is either moved to their relevant sections or removed because its duplicate content
- updates `training.md` tutorial to show training a language model instead of BERT lol
- adds new `trainer_customize.md` guide showing how to subclass `get_train_dataloader` and `compute_loss` using real-world examples from TRL (we can add more examples here later)


docs/source/en/trainer_customize.md
```diff
@@ -0,0 +1,107 @@
...
+| [`~Trainer.log`] | log information about the training process |
+| [`~Trainer.create_optimizer_and_scheduler`] | create an optimizer and learning rate scheduler (can also be separately customized with [`~Trainer.create_optimizer`] and [`~Trainer.create_scheduler`] if they weren't passed in `__init__`) |
+| [`~Trainer.compute_loss`] | compute the loss of a batch of training inputs |
+| [`~Trainer.training_step`] | perform the training step |
+| [`~Trainer.prediction_step`] | perform the prediction and test step |
+| [`~Trainer.evaluate`] | evaluate the model and return the evaluation metric |
+
+The examples below show how to subclass some of these methods.
+
+## get_train_dataloader
```
stevhliu, line 26: maybe you can review these sections @qgallouedec since they're TRL examples? 🙏

docs/source/en/trainer_callbacks.md
```diff
@@ -0,0 +1,30 @@
+# Callbacks
```
stevhliu, line 1: you can ignore this section as its covered in https://github.com/huggingface/transformers/pull/44239

docs/source/en/perf_train_gpu_one.md
```diff
@@ -293,3 +293,148 @@ from transformers import AutoModelForCausalLM
...
+[Gradient Low-Rank Projection (GaLore)](https://hf.co/papers/2403.03507) significantly reduces memory usage when training large language models (LLMs). One of GaLores key benefits is *full-parameter* learning, unlike low-rank adaptation methods like [LoRA](https://hf.co/papers/2106.09685), which produces better model performance.
+
+Install the [GaLore](https://github.com/jiaweizzhao/GaLore) and [TRL](https://hf.co/docs/trl/index) libraries.
+
+```bash
+pip install galore-torch trl
+```
+
+Pick a GaLore optimizer (`"galore_adamw"`, `"galore_adafactor"`, `"galore_adamw_8bit`") and pass it to the `optim` parameter in [`trl.SFTConfig`]. Use the `optim_target_modules` parameter to specify which modules to adapt (can be a list of strings, regex, or a full path).
+
```
SunMarc: Not sure if I would put that here. The doc here is quite nice as it is super light. So adding this galore section will incite ppl in the future to put everything there as we ded in trainer.md.

docs/source/en/perf_train_gpu_one.md
```diff
@@ -293,3 +293,148 @@ from transformers import AutoModelForCausalLM
...
+Only linear layers that are considered GaLore layers can be trained with low-rank decomposition. The rest of the model layers are optimized in the usual way.
+
+## Liger
+
+[Liger Kernel](https://github.com/linkedin/Liger-Kernel) is a collection of layers such as RMSNorm, RoPE, SwiGLU, CrossEntropy, FusedLinearCrossEntropy, and more that have been fused into a single Triton kernel for training LLMs. These kernels are also compatible with FlashAttention, FSDP, and DeepSpeed. As a result, Liger Kernel can increase multi-GPU training throughput and reduce memory usage. This is useful for multi-head training and supporting larger vocabulary sizes, larger batch sizes, and longer context lengths.
+
+```bash
+pip install liger-kernel
+```
+
```
SunMarc, line 304: this is quite used to it should be okay to leave it here

docs/source/en/trainer_customize.md
```diff
@@ -0,0 +1,107 @@
...
+
+-->
+
+# Customizing the Trainer
+
+Subclass [`Trainer`] methods to change training behavior without rewriting the entire loop. Subclassing modifies the *training loop*, for example the forward pass or loss computation.
+
+Before subclassing, consider whether you need to change *what* [`Trainer`] computes or *when* and *whether* it acts. For timing and conditional logic, use a [Callback](./trainer_callbacks) instead. Callbacks control when things happen (logging, evaluation, early stopping) and subclassing changes what happens (loss computation, data loading, optimization).
+
+The table below lists methods available for subclassing.
```
SunMarc: maybe also say that they are private methods that they can subclass but we don't guarantee any breaking changes on those

docs/source/en/trainer_customize.md
```diff
@@ -0,0 +1,107 @@
...
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+
+-->
+
+# Customizing the Trainer
+
+Subclass [`Trainer`] methods to change training behavior without rewriting the entire loop. Subclassing modifies the *training loop*, for example the forward pass or loss computation.
+
+Before subclassing, consider whether you need to change *what* [`Trainer`] computes or *when* and *whether* it acts. For timing and conditional logic, use a [Callback](./trainer_callbacks) instead. Callbacks control when things happen (logging, evaluation, early stopping) and subclassing changes what happens (loss computation, data loading, optimization).
```
SunMarc, line 21: maybe add some examples of trainer in trl or axolotl

docs/source/en/training.md
```diff
@@ -16,119 +16,140 @@ rendered properly in your Markdown viewer.
...
+```
 
-<Youtube id="nvBXf7s7vTI"/>
+## Loading a model
 
-[Trainer](./trainer) is an optimized training loop for Transformers models, making it easy to start training right away without manually writing your own training code. Pick and choose from a wide range of training features in [`TrainingArguments`] such as gradient accumulation, mixed precision, and options for reporting and logging training metrics.
+Load a pretrained checkpoint to fine-tune (see the [Loading models](./models) guide for more details about loading models).
 
-Load a model and provide the number of expected labels (you can find this information on the Yelp Review [dataset card](https://huggingface.co/datasets/Yelp/yelp_review_full#data-fields)).
+- Set `dtype="auto"` to load the weights in their saved dtype. Without it, PyTorch loads weights in `torch.float32`, which doubles memory usage if the weights are originally `torch.bfloat16`.
```
SunMarc, line 78: it is not the needed anymore. we switched to "auto" in v5. But we can keep it in the example in case

SunMarc (approved):
Thanks a lot ! Left a couple of comment but overall really good !

docs/source/en/perf_train_gpu_one.md
```diff
@@ -293,3 +293,148 @@ from transformers import AutoModelForCausalLM
...
+    output_dir="your-model",
+    learning_rate=2e-5,
+    per_device_train_batch_size=16,
+    per_device_eval_batch_size=16,
+    num_train_epochs=2,
+    weight_decay=0.01,
+    eval_strategy="epoch",
+    save_strategy="epoch",
+    load_best_model_at_end=True,
+    push_to_hub=True,
```
qgallouedec:
```suggestion
```

docs/source/en/perf_train_gpu_one.md
```diff
@@ -293,3 +293,148 @@ from transformers import AutoModelForCausalLM
...
```
qgallouedec:
```suggestion
```

docs/source/en/trainer_customize.md
```diff
@@ -0,0 +1,107 @@
...
+|---|---|
+| [`~Trainer.get_train_dataloader`] | create a training DataLoader |
+| [`~Trainer.get_eval_dataloader`] | create an evaluation DataLoader |
+| [`~Trainer.get_test_dataloader`] | create a test DataLoader |
+| [`~Trainer.log`] | log information about the training process |
+| [`~Trainer.create_optimizer_and_scheduler`] | create an optimizer and learning rate scheduler (can also be separately customized with [`~Trainer.create_optimizer`] and [`~Trainer.create_scheduler`] if they weren't passed in `__init__`) |
+| [`~Trainer.compute_loss`] | compute the loss of a batch of training inputs |
+| [`~Trainer.training_step`] | perform the training step |
+| [`~Trainer.prediction_step`] | perform the prediction and test step |
+| [`~Trainer.evaluate`] | evaluate the model and return the evaluation metric |
```
qgallouedec: not convinced by the added value of this table. it seems redundant with the `[[autodoc]] Trainer`

docs/source/en/trainer_customize.md
```diff
@@ -0,0 +1,107 @@
...
+
+```py
+def get_train_dataloader(self):
+    return self._get_dataloader(
+        batch_size=self._train_batch_size,
+        ...
+)
+```
+
+GRPO needs to generate completions before training on them. Generating completions every step is expensive because it's autoregressive. A 512-token completion requires ~512 sequential forward passes compared to one forward pass for a training step. [`~trl.GRPOTrainer`] subclasses [`~Trainer.get_train_dataloader`] to batch generation across multiple steps.
```
qgallouedec:
```suggestion
[GRPO](https://huggingface.co/docs/trl/en/grpo_trainer) is an online reinforcement learning algorithm, which means it must first generate completions and then train on those same samples. Generating completions every step is expensive because it's autoregressive. A 512-token completion requires ~512 sequential forward passes compared to one forward pass for a training step. [`~trl.GRPOTrainer`] subclasses [`~Trainer.get_train_dataloader`] to batch generation across multiple steps.
```

docs/source/en/trainer_customize.md
```diff
@@ -0,0 +1,107 @@
...
+## compute_loss
+
+[`~Trainer.compute_loss`] returns the cross-entropy loss calculated by the model.
+
+```py
+def compute_loss(self, model, inputs, return_outputs=False, num_items_in_batch=None):
+    ...
+    outputs = model(**inputs)
+    ...
+        loss = outputs["loss"] # get loss from model
```
qgallouedec:
```suggestion
    loss = outputs["loss"] # get loss from model
```

docs/source/en/trainer_customize.md
```diff
@@ -0,0 +1,107 @@
...
+
+    return (loss, outputs) if return_outputs else loss
+```
+
+DPO measures how much more likely the policy model is to prefer a chosen response versus a rejected one, relative to a reference model. [`~trl.DPOTrainer`] subclasses [`~Trainer.compute_loss`] because the loss computation differs from standard cross-entropy in several ways:
```
qgallouedec:
```suggestion
[DPO](https://huggingface.co/docs/trl/en/dpo_trainer) measures how much more likely the policy model is to prefer a chosen response versus a rejected one, relative to a reference model. [`~trl.DPOTrainer`] subclasses [`~Trainer.compute_loss`] because the loss computation differs from standard cross-entropy in several ways:
```

docs/source/en/trainer_customize.md
```diff
@@ -0,0 +1,107 @@
...
+        # 2. compute reference model log-probs (or use precomputed values from the batch)
+        # 3. dpo_loss: loss = -log σ(β * (log π(chosen)/π_ref(chosen) - log π(rejected)/π_ref(rejected)))
+        loss, metrics = self.get_batch_loss_metrics(model, inputs, train_eval="train")
+
+    ...
+
+    if return_outputs:
+        return loss, metrics
+
+    return loss
```
qgallouedec: We refactored it recently, maybe it's easier to understand like this:


```suggestion
        outputs = model(**inputs)
        logits = outputs.logits
        logps = get_logps(logits, inputs)
        chosen_logps, rejected_logps = logps.chunk(2, dim=0)  # batch is [chosen, rejected]
        ref_logits = self.ref_model(**inputs).logits
        ref_logps = get_logps(ref_logits, inputs)
        ref_chosen_logps, ref_rejected_logps = ref_logps.chunk(2, dim=0)  # batch is [chosen, rejected]
        chosen_scores = chosen_logps - ref_chosen_logps
        rejected_scores = rejected_logps - ref_rejected_logps
        per_sequence_loss = -F.logsigmoid(self.beta * chosen_scores - rejected_scores)
        loss = per_sequence_loss.mean()
        return (loss, outputs) if return_outputs else loss
```

docs/source/en/training.md
```diff
@@ -16,119 +16,140 @@ rendered properly in your Markdown viewer.
...
 
-```py
-import numpy as np
-import evaluate
+</hfoption>
+<hfoption id="training optimizations">
 
-metric = evaluate.load("accuracy")
+- Set `bf16=True` for fast mixed precision training if your hardware supports it (Ampere+ GPUs). Otherwise, fall back to `fp16=True` on older hardware.
+- Enable `gradient_accumulation_steps` and `gradient_checkpointing` to simulate training on larger batches and reduce memory usage.
```
qgallouedec: mixing these two seems a bit weird. Maybe you meant `per_device_train_batch_size` instead of `gradient_checkpointing`?

qgallouedec (approved):
nice, I shared some suggestions :)

docs/source/en/perf_train_gpu_one.md
stevhliu: good point!

i'll move galore --> `optimizers.md` since thats what it actually is
move liger --> performance > speed optimizations > kernels (will do in future pr once i have those sections)
move neftune --> trainer cookbook recipes (will do in future pr once i have those sections)

this way we can keep this doc here more light and it focuses on general GPU training techniques?

docs/source/en/trainer_customize.md
stevhliu: added link to examples in trl/axolotl at the bottom

docs/source/en/trainer_customize.md
stevhliu: removed in favor of `[[autodoc]] Trainer` !

docs/source/en/training.md
stevhliu: splitted into two separate points to avoid conflating the two!

PR #44187, opened 2026-02-20
fix(models): Fix LayoutLMv2 NER crash and broken batched truncation/padding

### LINKED ISSUE #44186: [BUG] LayoutLMv2Tokenizer crashes on NER inputs and batched padding/truncation
### System Info

* `transformers` version: `5.0.0.dev0`
* Platform: `Linux-5.15.167.4-microsoft-standard-WSL2-x86_64-with-glibc2.39`
* Python version: `3.12.3`
* `huggingface_hub` version: `1.3.2`
* `safetensors` version: `0.7.0`
* `accelerate` version: `1.12.0`
* Accelerate config: `not installed`
* DeepSpeed version: `not installed`
* PyTorch version (accelerator?): `2.9.1+cu128 (CUDA)`
* GPU type: `NVIDIA L4`
* NVIDIA driver version: `550.90.07`
* CUDA version: `12.4`

### Who can help?

@zucchini-nlp (multimodal model)
@ArthurZucker (tokenizer)

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

**NER use case:**

```python
from transformers import LayoutLMv2Tokenizer

tokenizer = LayoutLMv2Tokenizer.from_pretrained("microsoft/layoutlmv2-base-uncased")
words = ["Total", "Amount", ":", "$1,234.56"]
boxes = [[100, 200, 300, 250], [310, 200, 450, 250], [460, 200, 480, 250], [490, 200, 650, 250]]
word_labels = [0, 0, 0, 1]

try:
    encoding = tokenizer(words, boxes=boxes, word_labels=word_labels)
    print(encoding["labels"])
except Exception as e:
    print(e)
```

**Batched training data prep with truncation/padding:**

```python
from transformers import LayoutLMv2Processor
from datasets import load_dataset
import textwrap

try:
    processor = LayoutLMv2Processor.from_pretrained(
        "microsoft/layoutlmv2-base-uncased",
        apply_ocr=False
    )
    dataset = load_dataset("nielsr/funsd", split="train")
    images = [img.convert("RGB") for img in dataset["image"]]
    words = list(dataset["words"])
    boxes = list(dataset["bboxes"])
    word_labels = list(dataset["ner_tags"])
    encoding = processor(
        images,
        words,
        boxes=boxes,
        word_labels=word_labels,
        padding="max_length",
        truncation=True,
        return_tensors="pt",
    )
    print(encoding["input_ids"].shape)
except Exception as e:
    print("\n".join(textwrap.wrap(str(e), width=160)))
```

[LayoutLMv2Tokenizer](https://github.com/huggingface/transformers/blob/main/src/transformers/models/layoutlmv2/tokenization_layoutlmv2.py#L112) crash with an `AttributeError` when `word_labels` is passed for NER token classification. In a different use case, calling the processor with `padding="max_length"` and `truncation=True` raises a downstream `ValueError` asking to set the aforementioned flags (more details in the PR; the screenshots in the PR show what happens after the first attr issue is fixed but before the second fix is made), despite both flags being set correctly.

**Current Repro Output:**

Image

### Expected behavior

→ `encoding["labels"]` should return a list in which subword tokens are masked with the [default ignore_index](https://docs.pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html) (`-100`) in `nn.CrossEntropyLoss`
→ `encoding["input_ids"].shape` should return the expected `torch.Size()`.

harshaljanjani:
### What does this PR do?

The following issues were identified and fixed in this PR:

→ The NER/token classification issue and the downstream bug uncovered in the batched preprocessing use case with `LayoutLMv2Tokenizer`.
→ **Reasoning:** The NER use case makes it apparent that the error is hit at [this line](https://github.com/harshaljanjani/transformers/blob/main/src/transformers/models/layoutlmv2/tokenization_layoutlmv2.py#L661) in [LayoutLMv2Tokenizer](https://github.com/huggingface/transformers/blob/main/src/transformers/models/layoutlmv2/tokenization_layoutlmv2.py#L112), which is missing `self.only_label_first_subword`. `PretrainedTokenizerBase` doesn't create `self.only_label_first_subword` either, so this must be added along with the [other custom attributes](https://github.com/harshaljanjani/transformers/blob/main/src/transformers/models/layoutlmv2/tokenization_layoutlmv2.py#L220-L223), and this directly resolved the NER use case as shown in the screenshot.
→ For the second fix; [any padding="max_length" or truncation=True call without an explicit max_length arg compares self.model_max_length > LARGE_INTEGER (1e20)](https://github.com/harshaljanjani/transformers/blob/main/src/transformers/tokenization_utils_base.py#L2348-L2359), which in this case evaluates to `True` ([since model_max_length falls back to VERY_LARGE_INTEGER](https://github.com/harshaljanjani/transformers/blob/main/src/transformers/tokenization_utils_base.py#L129)), and both get translated into no-ops. Sequences in the batch are of different lengths and can't be tensorized, and the misleading `ValueError` tells the user to add `padding=True` and `truncation=True`, but they already did? The fix restores `model_max_length=512`. I confirmed both the [base](https://huggingface.co/microsoft/layoutlmv2-base-uncased/blob/main/config.json) and [large](https://huggingface.co/microsoft/layoutlmv2-large-uncased/blob/main/config.json) model configs have `max_position_embeddings=512`, so 512 as a default is correct, and followed the same pattern as [MarianTokenizer](https://github.com/harshaljanjani/transformers/blob/main/src/transformers/models/marian/tokenization_marian.py#L118) and [TapasTokenizer](https://github.com/harshaljanjani/transformers/blob/main/src/transformers/models/tapas/tokenization_tapas.py#L256) :)

Originally removed in #42894, just wanted to double-check if this was intentional; happy to adjust this fix if I’ve missed something :)

Fixes #44186.

**Before both fixes applied:**

<br>

**Attribute fix resolves NER; batched use case still fails (feel free to cross-check; the errors are reproducible):**

<br>

**After both fixes applied; NER + batched use case work (feel free to cross-check):**

<br>

### Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/src/transformers/models/layoutlmv2/tokenization_layoutlmv2.py
+++ b/src/transformers/models/layoutlmv2/tokenization_layoutlmv2.py
@@ -175,6 +175,7 @@ class LayoutLMv2Tokenizer(TokenizersBackend):
         only_label_first_subword=True,
         tokenize_chinese_chars=True,
         strip_accents=None,
+        model_max_length=512,
         **kwargs,
     ):
         self.do_lower_case = do_lower_case
@@ -214,6 +215,7 @@ class LayoutLMv2Tokenizer(TokenizersBackend):
             only_label_first_subword=only_label_first_subword,
             tokenize_chinese_chars=tokenize_chinese_chars,
             strip_accents=strip_accents,
+            model_max_length=model_max_length,
             **kwargs,
         )
 
@@ -221,6 +223,7 @@ class LayoutLMv2Tokenizer(TokenizersBackend):
         self.sep_token_box = sep_token_box
         self.pad_token_box = pad_token_box
         self.pad_token_label = pad_token_label
+        self.only_label_first_subword = only_label_first_subword
 
         # Now set post_processor with actual token IDs
         cls = str(self.cls_token)


LalithaMV (approved):

zucchini-nlp (approved):
Thanks for the detailed explanation, lgtm! Btw, @ArthurZucker , I think we should raise a warning/error asking to pass a `max_length` arg when `model_max_length > VERY_LARGE_INT` because the current error message is misleading

zucchini-nlp:
run-slow: layoutlmv2

PR #44201, opened 2026-02-21
fix: HiggsAudioV2 cached decode inputs in compiled generation

tarekziade:
# What does this PR do?

main is currently failing with 

```
FAILED tests/models/higgs_audio_v2/test_modeling_higgs_audio_v2.py::HiggsAudioV2ModelTest::test_generate_compilation_all_outputs - AssertionError: Lists differ: [torch.Size([2, 15, 32])] != [(2, 1, 32)]
```

likely introduced by [1618d44b92](https://github.com/huggingface/transformers/commit/1618d44b92) or a more recent conflictual change (I did not bisect)

Fix: Changed to slice last token when `use_cache=True`

I am not 100% sure my fix is right


tarekziade:
run-slow: higgs_audio_v2

src/transformers/models/higgs_audio_v2/generation_higgs_audio_v2.py
```diff
@@ -276,7 +276,11 @@ def _sample(
 
         while self._has_unfinished_sequences(this_peer_finished, synced_gpus, device=input_ids.device):
             if prefill_consumed:
-                model_inputs = self.prepare_inputs_for_generation(input_ids, **model_kwargs)
+                use_cache = model_kwargs.get("use_cache", True)
+                new_input_ids = input_ids[:, -1:] if use_cache else input_ids
+                model_inputs = self.prepare_inputs_for_generation(
+                    new_input_ids, full_input_ids=input_ids, **model_kwargs
```
vasqu, line 281: I guess we need the full input ids for the custom mask logic? It's a bit weird but otherwise aligns with the rest of the new changes on generate

vasqu (commented):
The conflict comes from a bad timing between the model addition and #44130 so waiting for @Cyrilvallez to take a look (the PR at least seems to resolve the issues, so not to double the effort)

src/transformers/models/higgs_audio_v2/generation_higgs_audio_v2.py
Cyrilvallez: cc @eustlb, do we absolutely need all previous input_ids? It's indeed a bit awkward to have to do pass both

Cyrilvallez (approved):
Just pushed a missing block for consistency. 
LGTM, even though I'm not a fan oh having to send both sliced inputs and full inputs. Maybe @eustlb knows if we can avoid it?
We can merge like this in the meantime though I guess as a fix

tarekziade:
run-slow: higgs_audio_v2

src/transformers/models/higgs_audio_v2/generation_higgs_audio_v2.py
eustlb: wrote this quite some time ago but this strange logic is intended to allow generating from past_key_values from a previous run

eustlb:
Having a look quickly before merging

eustlb (approved):
For higgs_audio_v2, we have
- `input_ids` (classical, with eg **N** audio tokens placeholders)
- `audio_input_ids`, shape `[bs, N, 8]`

Now let's say we generate `max_new_tokens=3`, and continue from the `past_key_values`, and from the same inputs (API enforced in `test_generate_continue_from_past_key_values`), issue is that to mask audio_input_ids correctly we need to know how much audio_input_ids tokens were consumed in the previous run. This was done before _prefill was introduced so let's merge like this as a hit fix for now and I'll raise another PR to clean it.

PR #44202, opened 2026-02-21
Fix: bidirectional mask skip when attention dropout is active (#44188)

### LINKED ISSUE #44188: Diverging attention kernels due to `allow_is_bidirectional_skip` branching on torch.compile
### System Info

Hi, while we were updating the PyTorch transformers pin to v5.2.0, our regression tests caught a numerics issue between eager and compiled, the difference is very substantial (3.3 vs the typical e-4 accepted difference). Digging into it: https://github.com/pytorch/pytorch/pull/175274#issuecomment-3930952666, we found the cause to be in these lines (added in https://github.com/huggingface/transformers/pull/41265): 

https://github.com/huggingface/transformers/blob/147b7aa040812b079f467e777a2d2e1284167de0/src/transformers/masking_utils.py#L490-L491

We set `allow_is_bidirectional_skip=True` in a few places:
https://github.com/huggingface/transformers/blob/147b7aa040812b079f467e777a2d2e1284167de0/src/transformers/masking_utils.py#L996-L997

And in `_ignore_bidirectional_mask_sdpa`, we branch logic on whether we compile or not:
https://github.com/huggingface/transformers/blob/147b7aa040812b079f467e777a2d2e1284167de0/src/transformers/masking_utils.py#L324-L332

This issue was found on BERT but it seems like it would affect other models too.

We've also verified that removing the branching fixes the numerical difference. I'm creating this issue to ask about the best way forward here. From the PR that added it, it looks like this was necessary specifically for executorch, but the algorithm difference is also affected all other APIs that fall under `is_tracing` . Can we restrict the check?

### Who can help?

@vasqu @ArthurZucker @Cyrilvallez 

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

I believe the description is enough, but I can provide a simpler repro on request

### Expected behavior

transformers users probably shouldn't run into large numeric differences when compiling, at least not by default

GS-GOAT:
# What does this PR do?

When `torch.compile` is used, [_ignore_bidirectional_mask_sdpa](cci:1://file:///c:/Users/BIT/Desktop/proj/gitrepo_clones/transformers/src/transformers/masking_utils.py:303:0-338:16) behaves differently than in eager mode due to the [is_tracing()](cci:1://file:///c:/Users/BIT/Desktop/proj/gitrepo_clones/transformers/src/transformers/utils/import_utils.py:1390:0-1402:22) check. In eager mode (no tracing), the bidirectional mask is skipped (returns `None`), allowing SDPA to dispatch to flash attention. Under `torch.compile` ([is_tracing()](cci:1://file:///c:/Users/BIT/Desktop/proj/gitrepo_clones/transformers/src/transformers/utils/import_utils.py:1390:0-1402:22) returns `True`), the mask is materialized, causing SDPA to use the memory-efficient backend instead.

When attention dropout is active (like, BERT's default `attention_probs_dropout_prob=0.1`), these two backends handle dropout RNG differently, leading to large numerical divergences.

**Fix:** adding a [dropout](cci:1://file:///c:/Users/BIT/Desktop/proj/gitrepo_clones/transformers/tests/utils/test_masking_utils.py:388:4-423:48) condition to [_ignore_bidirectional_mask_sdpa](cci:1://file:///c:/Users/BIT/Desktop/proj/gitrepo_clones/transformers/src/transformers/masking_utils.py:303:0-338:16) - when `dropout > 0`, mask creation is never skipped, ensuring the same SDPA backend is used in both eager and compiled modes. The dropout value is extracted from the model-config (supporting both `attention_probs_dropout_prob` and `attention_dropout` attribute names).

Fixes #44188

## Before submitting
      Pull Request section?
      to it if that's the case.
      - https://github.com/huggingface/transformers/issues/44188

## Who can review?

### DIFF
--- a/src/transformers/masking_utils.py
+++ b/src/transformers/masking_utils.py
@@ -305,6 +305,7 @@ def _ignore_bidirectional_mask_sdpa(
     padding_mask: torch.Tensor | None,
     kv_length: int,
     local_attention_size: int | None = None,
+    dropout: float = 0.0,
 ) -> bool:
     """
     Detects whether the bidirectional mask can be ignored in case PyTorch's SDPA is used.
@@ -314,6 +315,12 @@ def _ignore_bidirectional_mask_sdpa(
     allowing to dispatch to the flash attention kernel (that can otherwise not be used if a custom `attn_mask` is
     passed).
     """
+    # when dropout is used, different sdpa backends handle dropout rng differently;
+    # to ensure consistent numerics between eager and compiled modes, we should not,
+    # skip mask creation when dropout is active. [see: https://github.com/huggingface/transformers/issues/44188]
+    if dropout > 0.0:
+        return False
+
     if _is_torch_xpu_available:
         # XPU devices have special handling for mask skipping:
         # - Skip if no padding and no local attention constraint
@@ -376,6 +383,7 @@ def sdpa_mask(
     allow_is_bidirectional_skip: bool = False,
     allow_torch_fix: bool = True,
     use_vmap: bool = False,
+    dropout: float = 0.0,
     **kwargs,
 ) -> torch.Tensor | None:
     """
@@ -487,7 +495,7 @@ def sdpa_mask(
     #   2. Bidirectional do not need any further processing (no bias)
     if allow_is_causal_skip and _ignore_causal_mask_sdpa(padding_mask, q_length, kv_length, kv_offset, local_size):
         return None
-    if allow_is_bidirectional_skip and _ignore_bidirectional_mask_sdpa(padding_mask, kv_length, local_size):
+    if allow_is_bidirectional_skip and _ignore_bidirectional_mask_sdpa(padding_mask, kv_length, local_size, dropout):
         return None
 
     # Potentially add the padding 2D mask
@@ -541,6 +549,7 @@ def eager_mask(
     dtype: torch.dtype = torch.float32,
     allow_is_bidirectional_skip: bool = False,
     use_vmap: bool = False,
+    dropout: float = 0.0,
     **kwargs,
 ) -> torch.Tensor:
     """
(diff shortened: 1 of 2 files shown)


vasqu:
Referencing my comment just in case https://github.com/huggingface/transformers/issues/44188#issuecomment-3944038471 - not sure if we proceed with this now, let's wait on a response from the torch side

redpanda1995 (approved):

Cyrilvallez:
Agreed with @vasqu here! I don't think we need this, as it simply stems from different rng in dropout!
Piling this kind of "fixes" eventually makes our code much harder to understand and read, as we already have several different pathes etc due to upstream issues based on different conditions, torch versions etc, so I would really like to avoid this one, as it's not really a bug.

Cyrilvallez:
Unrelated, @redpanda1995 I've seen you randomly approve several PRs recently. If you wish to boost your GitHub contributions, please do it on another repo - we do not tolerate this kind of behavior here, as it adds noise to all of us reviewing PRs. This is your first and last warning, if I notice you doing it again, I will be forced to ban you from this repo

GS-GOAT:
@Cyrilvallez In that case, i will close this pr.

TransformerTitan:
hi @Cyrilvallez ,  I truly apologize for this. Again I truly apologize, I did not know it was violating policy. @redpanda1995  is a honest and dedicated contributor and has no intention of adding noise.  It appears @redpanda1995  is already blocked now and can't contribute, although you mentioned it was a warning. Could you please unblock @redpanda1995  ?  Again, it was just a minor misunderstanding, it will not happen again. @redpanda1995  is a strong ML Engineer and only has good intentions. cc @Cyrilvallez 

Thanks.

Cyrilvallez:
Alright, closing then! If it truly becomes an issue (other than "somewhat expected" rng differences), feel free to reopen later on! Thanks for flagging it in any case!

PR #44182, opened 2026-02-20
Remove refs to grouped_entities

Rocketknight1:
Our code has some references to the `grouped_entities` arg to the token classification pipeline, but this is no longer usable. This PR cleans them up entirely!

Fixes #44016

PR #44217, opened 2026-02-23
Docs(it): update deprecated grouped_entities reference

matisgagneux21:
## Summary
- Update the Italian migration guide to avoid pointing users to the deprecated `grouped_entities` flag.
- Clarify that `aggregation_strategy` is the current option (with a note that it was previously `grouped_entities`).

## Why
Recent reports (e.g. around NER examples) show confusion from users hitting `TypeError` when using `grouped_entities`. This keeps the migration docs aligned with current pipeline parameters.


LalithaMV (approved):

Rocketknight1:
Should be fixed by #44182 as soon as that's merged!

PR #44223, opened 2026-02-23
Fix type of TrainingArguments.logging_steps

albertvillanova:
Fix type of `TrainingArguments.logging_steps`.

This PR makes a minor update to the `TrainingArguments` class, so `logging_steps` parameter accepts both integers and floats, rather than only floats. Note these are the expected types in the implementation and the documentation.

- Configuration improvement:
  * Fixed the type of `logging_steps` from `float` to `int | float` in the `TrainingArguments` class to allow both input types.


albertvillanova:
I see argparse only supports a single type per argument.

PR #44225, opened 2026-02-23
Fix generation integration tests

zucchini-nlp:
# What does this PR do?

As per title, we weren't running these test for some time because they were being filtered into `non-model` tests. But `non-model` doesn't run tests that are marked as generation, so I moved it back to `generation` and fixed failing ones


tests/generation/test_utils.py
```diff
@@ -3357,16 +3356,6 @@ def test_length_warning_assisted_generation(self):
             )
             self.assertEqual(len(warning_list), 0)
 
-    def test_default_assisted_generation(self):
-        # Initialize the GenerationConfig object
-        config = GenerationConfig()
-
```
zucchini-nlp, line 3363: irrelevant test, all defaults are tested in `test_confgiuration_utils` and are `None`

tests/generation/test_utils.py
```diff
@@ -3924,30 +3923,6 @@ def test_prepare_inputs_for_generation_encoder_decoder_llm(self):
         self.assertTrue(model_inputs["encoder_outputs"] == "foo")
         # See the decoder-only test for more corner cases. The code is the same, so we don't repeat it here.
 
-    @pytest.mark.torch_compile_test
-    def test_generate_compile_fullgraph_tiny(self):
-        """
-        Tests that we can call end-to-end generation with a tiny model (i.e. doesn't crash)
-        NOTE: this test is quite slow (~20s on a consumer desktop), but it is important that we keep it as part of the
```
zucchini-nlp, line 3931: generate is not and will not be fullgraph compilable, we compile only `model.forward`

tests/generation/test_utils.py
```diff
@@ -4008,45 +3983,6 @@ def test_beam_search_advanced_stopping_criteria(self):
         self.assertTrue(":" in output_text[-5:])
         self.assertTrue(":" in last_non_special_token_decoded)
 
-    def test_max_time(self):
-        tokenizer = GPT2Tokenizer.from_pretrained("openai-community/gpt2")
-        model = GPT2LMHeadModel.from_pretrained("openai-community/gpt2")
-        model.to(torch_device)
-
```
zucchini-nlp, line 4015: hard to test because generation time is affected by many factors, and I think max time stopping is not used at all.

tests/generation/test_utils.py
```diff
@@ -4358,57 +4296,19 @@ def test_generate_inputs_and_encoder_kwargs(self):
...
         input_ids = tokenizer(article, return_tensors="pt").input_ids
         with self.assertRaises(ValueError):
-            model.generate(input_ids=input_ids, inputs_embeds=input_ids)
-
-    def test_generate_input_features_as_encoder_kwarg(self):
-        """Test that non-`input_ids` main model inputs are correctly handled as positional arguments"""
-        input_features = floats_tensor((3, 80, 60))
-        model = AutoModelForSpeechSeq2Seq.from_pretrained(
-            "hf-internal-testing/tiny-random-WhisperForConditionalGeneration"
-        )
```
zucchini-nlp, line 4380: irrelevant tests here as well

zucchini-nlp:
Failing tests on HiggsAudio are not related

tests/utils/test_cache_utils.py
```diff
@@ -64,7 +64,7 @@
     cache_name
     for cache_name in ALL_CACHE_IMPLEMENTATIONS
     # TODO (joao): offloaded_hybrid == offloaded_hybrid_chunked, deprecate one of them
-    if cache_name not in ["offloaded_hybrid", "offloaded_static", "offloaded_hybrid_chunked"]
+    if cache_name not in ["offloaded", "offloaded_hybrid", "offloaded_static", "offloaded_hybrid_chunked"]
```
zucchini-nlp, line 67: as explained in L61

tests/generation/test_utils.py
```diff
@@ -3567,33 +3554,38 @@ def test_compare_unprocessed_logit_scores(self):
         torch.testing.assert_allclose(logits_fwd.tolist(), logits_gen.tolist())
 
     def test_return_unprocessed_logit_scores(self):
-        # tell model to generate text and return unprocessed/unwarped logit scores
+        "Test that the returned logits and scores are different, i.e. unproceessed vs processed scores"
```
tarekziade: typo

tests/generation/test_utils.py
```diff
@@ -3567,33 +3554,38 @@ def test_compare_unprocessed_logit_scores(self):
...
 
         outputs = model.generate(
-            input_ids=input_ids, return_dict_in_generate=True, output_logits=True, max_new_tokens=3
+            input_ids=input_ids,
+            return_dict_in_generate=True,
+            output_logits=True,
+            output_scores=True,
+            max_new_tokens=3,
+            do_sample=False,
+            exponential_decay_length_penalty=(0, 4.6),
```
tarekziade, line 3570: what the rationale behind 4.6?

tests/generation/test_utils.py
```diff
@@ -3817,80 +3809,86 @@ def test_prepare_inputs_for_generation_decoder_llm(self):
...
+        init_cache_position = cache_position[dynamic_cache.get_seq_length() :]
+        model_inputs = model.prepare_inputs_for_generation(
+            input_ids,
+            past_key_values=init_dynamic_cache,
+            cache_position=init_cache_position,
+            attention_mask=attention_mask,
+            position_ids=position_ids,
+        )
+        self.assertTrue("past_key_values" in model_inputs)
+        self.assertTrue(torch.all(model_inputs["cache_position"] == init_cache_position))
```
tarekziade: for this block of assertions I would use `assertEqual` so it if it fails we get the two values explicitely displayed

tests/generation/test_utils.py
```diff
@@ -3817,80 +3809,86 @@ def test_prepare_inputs_for_generation_decoder_llm(self):
...
+            attention_mask=attention_mask,
+            position_ids=position_ids,
         )
         self.assertTrue("past_key_values" in model_inputs)
-        self.assertTrue(torch.all(model_inputs["cache_position"] == cache_position))
-        self.assertTrue(model_inputs["input_ids"].shape[-1] == 1)  # 1 = 3 fed tokens - 2 tokens in the cache
-        self.assertTrue(model_inputs["position_ids"].shape[-1] == 1)
-        self.assertTrue(model_inputs["attention_mask"].shape[-1] == 3)  # we still need the full attention mask!
+        self.assertTrue(model_inputs["cache_position"] is None)
+        self.assertTrue(model_inputs["input_ids"].shape[-1] == init_input_ids.shape[1])
```
tarekziade: same remark than earlier

tests/generation/test_utils.py
tarekziade: could it be more predictable on AWS runners? (so behind a `@slow` decorator)

tests/generation/test_utils.py
```diff
@@ -4072,15 +4007,22 @@ def test_custom_logits_processor(self):
...
 
-        # it should not be allowed to both define `min_length` via config and `logits_processor` list
-        with self.assertRaises(ValueError):
+        # it should not be allowed to both define `min_length` via kwargs and `logits_processor` list
+        logger = logging.get_logger("transformers.generation.utils")
+        logger.warning_once.cache_clear()
+        with CaptureLogger(logger) as cl:
             bart_model.generate(input_ids, logits_processor=logits_processor, min_length=10)
+        self.assertTrue(
+            "custom logits processor of type <class 'transformers.generation.logits_process.MinLengthLogitsProcessor'> has been passed to `.generate()`, but it was also created in `.generate()"
```
tarekziade, line 4024: are these emited by a warning? if so it would be more robust to catch the exception on warning instead of relying on a log that may change over time

tests/generation/test_utils.py
```diff
@@ -4308,25 +4247,23 @@ def test_encoder_decoder_generate_attention_mask(self):
...
-        batched_out = output_sequences_batched.sequences_scores
-        out = output_sequences.sequences_scores
-        batched_out = batched_out.cpu().numpy()
-        out = out.cpu().numpy()
+        batched_out = output_sequences_batched.sequences_scores.cpu().numpy()
+        out = output_sequences.sequences_scores.cpu().numpy()
 
         diff = np.abs(np.sum(batched_out[:5]) - np.sum(out))
-        self.assertTrue(diff < 1e-4)
+        self.assertTrue(diff < 5e-4)
```
tarekziade: use a less comparison

tarekziade (changes requested):
I have just some remarks to improve assertions so we get better feedback on failures

I also wonder if the max time test could be improved/saved.

tests/generation/test_utils.py
zucchini-nlp: nothing, just a random penalty so the logits are altered

tests/generation/test_utils.py
zucchini-nlp: yes, they are logged with `logger.warning_once`. What do you mean by "exception on warning", happy to improve?

tarekziade: see https://docs.pytest.org/en/stable/how-to/capture-warnings.html#warns
this allows you to treat warning as exceptions in your test

zucchini-nlp: Ah, this ones are for `warning` package, and the one I am trying to catch comes from `logger`. Not sure if that works, can try though

tarekziade: On our side that function just adds an lru cache on the real `warning` function 

https://github.com/huggingface/transformers/blob/main/src/transformers/utils/logging.py#L321-L330

so it should work unless pytest does something crazy :D

tarekziade: oh sorry, I misread, you are definitely right, sorry ! ignore this one

tarekziade (approved):
LGTM for the tests side. I can't comment much on the model side for now

redpanda1995 (approved):

zucchini-nlp:
Will merge it then, generation changes aren't global and mostly are just nitty-picks for edge-cases in tests

PR #44226, opened 2026-02-23
[generate] Always pass full input_ids in `prepare_inputs_for_generation`

Cyrilvallez:
# What does this PR do?

As per the title. It looks like some models (xlnet and kosmos2_5) and most audio models sometimes rely on the full previous input_ids to prepare inputs. Note that this cannot be compatible with restarting generation from a previously filled cache and new inputs, so those models are not well-behaved in general (if they use a cache, they should be able to restart from it with any input). However, it's a simple fix to forward the full inputs and slice inside `prepare_inputs_for_generation` for those models.

This should fix all audio models cc @eustlb. Please note my comment about how that means audio models cannot restart from old cache, not sure if it's known/intended/fixable.
As for the other only 2 non-audio models requiring past input_ids (xlnet and kosmos2_5), it's in general the same issue. Kosmos could be fixed, the other not sure.

See https://huggingface.co/datasets/hf-internal-testing/transformers_daily_ci/raw/8785954cca2fdca181de0b9567059471bcadb959/2026-02-21/ci_results_run_models_gpu/new_failures_with_bad_commit_grouped_by_authors.json for details on the failing tests.

cc @zucchini-nlp @vasqu

### DIFF
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -3744,7 +3744,7 @@ class GenerationMixin(ContinuousMixin):
             use_inputs_embeds = True
         if (cache := model_kwargs.get("past_key_values")) is not None:
             past_length = cache.get_seq_length()
-            # Always directly slice the inputs_embeds if present, as `prepare_inputs_for_generation` never need them full and _get_initial_cache_position
+            # Always directly slice the inputs_embeds if present, as `prepare_inputs_for_generation` never need them full and `_get_initial_cache_position`
             # rely on its size explicitly. For input_ids, we need to use `next_sequence_length` to slice later instead of explicit slicing,
             # as some model need them full for correct input preparation inside `prepare_inputs_for_generation` (i.e. audio models)
             if use_inputs_embeds:
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -539,15 +539,12 @@ class GenerationMixin(ContinuousMixin):
             model_inputs["token_type_ids"] = token_type_ids
 
         # 3. Slice model inputs if it's an input that should have the same length as `input_ids`
-        # We check `use_cache` below because some stateful models (like `recurrent_gemma`) expect input slicing if
-        # their caching mechanism is used. To define `use_cache`, the user-defined argument takes precedence.
-        if past_key_values is not None or kwargs["use_cache"]:
-            for model_input_name in [position_ids_key, "cache_position", "token_type_ids"]:
-                model_input = model_inputs.get(model_input_name)
-                if model_input is not None:
-                    # Input can be 2D or 3D, and we always slice on `seq-length` (last dim)
-                    model_input = model_input[..., -sequence_length:].clone(memory_format=torch.contiguous_format)
-                    model_inputs[model_input_name] = model_input
+        for model_input_name in [position_ids_key, "cache_position", "token_type_ids"]:
+            model_input = model_inputs.get(model_input_name)
+            if model_input is not None and model_input.shape[-1] != sequence_length:
+                # Input can be 2D or 3D, and we always slice on `seq-length` (last dim)
+                model_input = model_input[..., -sequence_length:].clone(memory_format=torch.contiguous_format)
+                model_inputs[model_input_name] = model_input
 
         # 4. Create 4D attention mask is we are using a compilable cache (important for performant compiled forward
         # pass)
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -539,7 +539,9 @@ class GenerationMixin(ContinuousMixin):
             model_inputs["token_type_ids"] = token_type_ids
 
         # 3. Slice model inputs if it's an input that should have the same length as `input_ids`
-        if kwargs["use_cache"]:
+        # We check `use_cache` below because some stateful models (like `recurrent_gemma`) expect input slicing if
+        # their caching mechanism is used. To define `use_cache`, the user-defined argument takes precedence.
+        if past_key_values is not None or kwargs["use_cache"]:
             for model_input_name in [position_ids_key, "cache_position", "token_type_ids"]:
                 model_input = model_inputs.get(model_input_name)
                 if model_input is not None:
--- a/src/transformers/generation/utils.py
+++ b/src/transformers/generation/utils.py
@@ -493,6 +493,7 @@ class GenerationMixin(ContinuousMixin):
     def prepare_inputs_for_generation(
         self,
         input_ids: torch.LongTensor,
+        next_sequence_length: int | None = None,
         past_key_values: Cache | None = None,
         attention_mask: torch.LongTensor | None = None,
         inputs_embeds: torch.FloatTensor | None = None,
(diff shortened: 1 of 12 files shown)


src/transformers/generation/utils.py
```diff
@@ -2729,9 +2728,10 @@ def _sample(
 
         while self._has_unfinished_sequences(this_peer_finished, synced_gpus, device=input_ids.device):
             if prefill_consumed:
-                use_cache = model_kwargs.get("use_cache", True)
-                new_inputs_ids = input_ids[:, -1:] if use_cache else input_ids
-                model_inputs = self.prepare_inputs_for_generation(new_inputs_ids, **model_kwargs)
+                next_sequence_length = 1 if model_kwargs.get("use_cache", True) else None
```
zucchini-nlp: nit: is it oke to default to `use_cache=True`? Prev we had a fallback to config attr

Cyrilvallez: It's actually guaranteed to exist in the model_kwargs. I updated to show it explicitly

Cyrilvallez:
run-slow: kosmos2_5

src/transformers/generation/utils.py
```diff
@@ -3730,30 +3733,40 @@ def _prefill(
...
+                # In this case we need to slice - if it's smaller than the mask, only the new inputs were passed -> no need to do anything
+                if attention_mask is not None and input_ids.shape[1] == attention_mask.shape[1]:
+                    # inputs will be sliced as `input_ids[:, -next_sequence_length :]` in `prepare_inputs_for_generation`
+                    next_sequence_length = input_ids.shape[1] - past_length
 
         # Usual prefill
         if generation_config.prefill_chunk_size is None:
-            model_kwargs = self._get_initial_cache_position(input_ids.shape[1], input_ids.device, model_kwargs)
+            # The cache is already taken into account in `_get_initial_cache_position`, so the length is only the new tokens if we slice
+            effective_input_length = next_sequence_length if next_sequence_length is not None else input_ids.shape[1]
```
zucchini-nlp, line 3763: Lost about this part, why inputs embeds cannot be sliced as is and does this work when both ids and embeds are passed?

src/transformers/models/kosmos2_5/modeling_kosmos2_5.py
```diff
@@ -1495,13 +1495,10 @@ def prepare_inputs_for_generation(
...
-                Kosmos2_5TextSinusoidalPositionalEmbedding.create_position_ids_from_input_ids(
-                    input_ids,
-                    padding_idx=self.config.pad_token_id,
-                    past_key_values_length=0,
-                )[:, -cache_position.shape[0] :]
-            )
+
+            # Kosmos2.5 starts position_ids at `pad_token_id`...
+            if model_inputs.get("position_ids") is not None:
+                model_inputs["position_ids"] = model_inputs["position_ids"] + 1 + self.config.pad_token_id
```
zucchini-nlp, line 1502: oh, maybe then we can just override `self._prepare_position_ids` for this model to account for offset

src/transformers/generation/utils.py
zucchini-nlp, line 500: imo now we have `next_sequence_length` which has similar functionality as `cache_position` or even just `past_key_values` in simple cases. Let's clean up redundant arg and deprecate them properly in subsequent PRs

zucchini-nlp (approved):
The only question I have is about input embeddings. Not really clear why we need to slice them before prefill, otherwise lgtm

src/transformers/generation/utils.py
Cyrilvallez: Yeah the whole goal of those PRs was to remove cache_position - even though those 2 kwargs are not really the same as `next_sequence_length` is not intended to be used in modeling code at all, I've already started removing cache_position everywhere here https://github.com/huggingface/transformers/pull/44181 🤗

src/transformers/generation/utils.py
Cyrilvallez: We never use them both simultaneously in `prepare_inputs_for_generation`. The issue is that `_get_initial_cache_position` will override the given `sequence_length` if inputs_embeds are present, so if we want the correct position we need to slice!

src/transformers/models/kosmos2_5/modeling_kosmos2_5.py
Cyrilvallez: Tried it but then I had to overwrite the with/without position_ids test, and it was working correctly when running only this one, but failing when running all of kosmos2.5 tests.... Don't really have time to investigate why the tester probably mutate its state, so went back 🥲

src/transformers/generation/utils.py
```diff
@@ -512,15 +513,18 @@ def prepare_inputs_for_generation(
...
+            inputs_embeds = (
+                inputs_embeds[:, -next_sequence_length:, :] if next_sequence_length is not None else inputs_embeds
+            )
             model_inputs["inputs_embeds"] = inputs_embeds.clone(memory_format=torch.contiguous_format)
             batch_size, sequence_length = inputs_embeds.shape[:2]
         else:
             # `clone` calls in this function ensure a consistent stride. See #32227
+            input_ids = input_ids[:, -next_sequence_length:] if next_sequence_length is not None else input_ids
             model_inputs[input_ids_key] = input_ids.clone(memory_format=torch.contiguous_format)
             batch_size, sequence_length = input_ids.shape[:2]  # we slice here as some models may have them 3D
```
vasqu, line 530: Not super relevant to this PR but this is probably a limitation to audio models, no? We can have 3D input ids with the codebook channel dim, so something along `[bsz, channels, seq_len]`

src/transformers/generation/utils.py
```diff
@@ -535,17 +539,12 @@ def prepare_inputs_for_generation(
...
-                    # Input can be 2D or 3D, and we always slice on `seq-length` (last dim)
-                    model_input = model_input[..., -sequence_length:]
-                    model_input = model_input.clone(memory_format=torch.contiguous_format)
-                    model_inputs[model_input_name] = model_input
+        for model_input_name in [position_ids_key, "cache_position", "token_type_ids"]:
+            model_input = model_inputs.get(model_input_name)
+            if model_input is not None and model_input.shape[-1] != sequence_length:
+                # Input can be 2D or 3D, and we always slice on `seq-length` (last dim)
+                model_input = model_input[..., -sequence_length:].clone(memory_format=torch.contiguous_format)
+                model_inputs[model_input_name] = model_input
```
vasqu, line 547:
```suggestion
            if model_input is not None:
                # Input can be 2D or 3D, and we always slice on `seq-length` (last dim)
                model_input = model_input[..., -sequence_length:].clone(memory_format=torch.contiguous_format)
                model_inputs[model_input_name] = model_input
```
Not a fan of a data dependent control flow here; should work in any case, no?

src/transformers/generation/utils.py
```diff
@@ -2729,9 +2729,10 @@ def _sample(
 
         while self._has_unfinished_sequences(this_peer_finished, synced_gpus, device=input_ids.device):
             if prefill_consumed:
-                use_cache = model_kwargs.get("use_cache", True)
-                new_inputs_ids = input_ids[:, -1:] if use_cache else input_ids
-                model_inputs = self.prepare_inputs_for_generation(new_inputs_ids, **model_kwargs)
+                next_sequence_length = 1 if model_kwargs["use_cache"] else None
```
vasqu, line 2732: `use_cache` always exists? I.e. changed from `.get`

src/transformers/generation/utils.py
zucchini-nlp: ahh I see, it's surprising that `_get_initial_cache_position` treats embeds and ids differently

Cyrilvallez: Yes, way too surprising IMO. It should only take the sequence_length and/or cache and that's it imo, but for another time

src/transformers/models/paligemma/modeling_paligemma.py
```diff
@@ -577,7 +577,7 @@ def prepare_inputs_for_generation(
 
         # position_ids in Paligemma are 1-indexed
         if model_inputs.get("position_ids") is not None:
-            model_inputs["position_ids"] += 1
+            model_inputs["position_ids"] = model_inputs["position_ids"] + 1
```
vasqu, line 581: Wanna add a small comment here that we need this to be explicitly out of place

vasqu (approved):
LGTM, just some smaller comments. I guess this is a pre-step to removing cache positions in #44181?

src/transformers/generation/utils.py
Cyrilvallez: It's actually not, as omitting the dim is equivalent to slicing it all with `:`!

src/transformers/generation/utils.py
Cyrilvallez: I don't think shape checks are data-dependent flows in terms of dynamo - we can remove anyway but then we always clone which is unnecessary (though not really a big deal)

src/transformers/generation/utils.py
Cyrilvallez: Yup!

Cyrilvallez:
Well, technically https://github.com/huggingface/transformers/pull/44130 was the pre-step for removing cache_position - this is a fix after said PR as audio models need to access full input_ids in `prepare_inputs_for_generation` unfortunately, so we have to delay the slicing a little bit!

PR #44227, opened 2026-02-23
[CB] Small fixes

remi-or:
This PR adds a logging message when infering the behavior of use async and fixes an error when evicting a graph from the graph buffer.


src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -58,7 +59,7 @@
 """
 Q_PADDING_INTERVAL_SIZE = 64
 KV_PADDING_INTERVAL_SIZE = 512 * 32  # 512 blocks of 32 tokens (interval size is in tokens for both Q and KV)
-MAX_CACHED_GRAPHS = 32
+MAX_CACHED_GRAPHS = 16
```
ArthurZucker: why?

ArthurZucker: did you check distribution of graph cache / miss? 😐

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -559,6 +566,11 @@ def __init__(
         # transfers, and if CUDA graphs are not turned off, because that would mean we are trying to save memory
         else:
             self.use_async_batching = self.use_cuda_graph and not attn_mask_is_needed(self.model.config)
+            logger.info(
+                f"No behavior specified for use_async_batching, choosing {self.use_async_batching = } because CUDA "
+                "graphs are turned on and no attention mask is needed. If you want to save memory, you can disable "
+                "asynchronous batching but it will degrade performance."
```
ArthurZucker: by an approximate x%

src/transformers/generation/continuous_batching/continuous_api.py
```diff
@@ -646,6 +658,19 @@ def stop(self, block: bool = True, timeout: float | None = None) -> None:
...
+                with open(self.timing_output_file, "w") as f:
+                    json.dump(
+                        {
+                            "cpu_prepare_times": time_tracker.cpu_prepare_times,
+                            "gpu_compute_times": time_tracker.gpu_compute_times,
+                            "idle_transfer_times": time_tracker.idle_transfer_times,
+                            "total_time": perf_counter() - self._start_time,
+                        },
+                        f,
+                    )
```
ArthurZucker: I did not check in details, I find this a bit invasive, would love to see an external lib being used (viztracer?) or anything.


This will clutter the code for just timing TBH!

src/transformers/generation/continuous_batching/utils.py
```diff
@@ -36,11 +39,79 @@ def get_graph(self, q_len: int, kv_len: int) -> torch.cuda.CUDAGraph | None:
...
 
 
+class CpuGpuTimeTracker:
+    """Tracks CPU and GPU time spans for performance analysis. GPU times use async CUDA events for no added overhead.
+    There are three monitored spans:
+     - CPU prepare time: corresponds to the time spent preparing and updating batches on CPU
+     - GPU compute time: corresponds to the time spent computing the forward pass on GPU
+     - idle transfer time: only in async batching, corresponds to the time spent waiting for H2D transfer to complete
+    """
+
```
ArthurZucker: are there no exising packages that can do that for us? (we don't have to do everything ourselves)

(let's document motivation for each choice we do as it impacts a lot of downstream libs.) Here for example you are adding code that is "optional" like an optional dep 😉

ArthurZucker (commented):
Thanks! 

Did you try other different tracking libs / what other libs depend on / what's the motivation for having it inhouse ? 

mostly afraid of adding optional code when potentially a decorator can do it for us (even for stream stuff)

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: lowered because I was OOMing, but I will revert as this is fixed in the `pged-attention` PR

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: done!

src/transformers/generation/continuous_batching/continuous_api.py
remi-or: Agreed this is misplaced, I will add it inside the tracer. I have checked out other libs, like viztracer or torch-simple-timer, but they either don't handle CUDa at all or introduce overhead. So a custom solution seems better, but I will add justification why.

src/transformers/generation/continuous_batching/utils.py
remi-or: Adding motivation there, good point

remi-or:
Following this discussion and offline discussion, the tracking tools that was originally planned for this PR is too niche to be added to the code base. This PR now only consists of two minor improvements that were planned for the previous PR.

ArthurZucker (approved):

PR #44232, opened 2026-02-23
chore: added CLAUDE.md alias

tarekziade:
# What does this PR do?

per https://code.claude.com/docs/en/claude-code-on-the-web#best-practices 

`CLAUDE.md` can alias directly into `AGENTS.md`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,32 +1,14 @@
-# AGENTS.md — Hugging Face Transformers
-
-## Commands
-- `make style` — ruff format + repo formatters (run before finishing)
-- `make fix-repo` — auto-fixes copies, modular conversions, doc TOCs, docstrings
+## Useful commands
+- `make style`: runs formatters and linters, necessary to pass code style checks
+- `make fix-repo`: auto-fixes copies, modular conversions, doc TOCs, docstrings in addition to the `make style` fixes
 - `make check-repo` — CI-style consistency checks
-- Slow tests: `RUN_SLOW=1 pytest ...`
+- Many tests are marked as 'slow' and skipped by default in the CI. To run them, use: `RUN_SLOW=1 pytest ...`
 
-## Workflow
-- After any code change: `make style` -> targeted `pytest tests/models/<name>/...`
-- After touching copied code, modular files, docs, or docstrings: `make fix-repo` instead
+`make style` or `make fix-repo` should be run as the final step before opening a PR. The CI will run `make check-repo` and fail if any issues are found.
 
 ## Copies and Modular Models
 
-We try to avoid direct inheritance between model-specific files. We have two mechanisms to manage the resulting code duplication:
+We try to avoid direct inheritance between model-specific files in `src/transformers/models/`. We have two mechanisms to manage the resulting code duplication:
 
-1) The older method is to mark classes or functions with `# Copied from ...`. Copies are kept in sync by linters and `make fix-repo`. Do not edit a `# Copied from` block, as it will be reverted by `make fix-repo`. Ideally you should edit the code it's copying from and propagate the change, but you can break the `# Copied from` link if needed.
-2) The newer method is to add a file named `modular_<name>.py` in the model directory. `modular` files **can** inherit from other models. `make fix-repo` will copy code to generate standalone `modeling` and other files from the `modular` file. When a `modular` file is present, that file should be edited directly. Changes made to the generated files will be overwritten by `make fix-repo`!
-
-## Adding a Model (modular way)
-- Create `src/transformers/models/<name>/modular_<name>.py`, inherit from a similar model, define only what changes.
-- `super().__init__(...)` unravels the parent body; use `del self.attr` after to drop unwanted attributes.
-- Run `make fix-repo` to regenerate model files from modular files.
-- Full guide: [docs/source/en/modular_transformers.md](docs/source/en/modular_transformers.md)
-
-## CI Test Selection
-- CI runs only tests impacted by the diff.
-- Reproduce locally: `python utils/tests_fetcher.py`, then `python -m pytest -n 8 --dist=loadfile -rA -s $(cat test_list.txt)`
-
-## Boundaries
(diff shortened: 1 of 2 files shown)


tarekziade:
even if AGENTS.md is eventually discovered, it's not read by default

see https://github.com/anthropics/claude-code/issues/6235

Rocketknight1:
Hmm - I'm scared of this PR! The reason is that a lot of the content in the `AGENTS.md` file is stuff Claude doesn't have a problem with, so I'm worried that we'll actually hurt performance with this. Explaining commands like `make fix-repo`, `make style` and the `modular` conventions are probably good, but I don't think it needs much else, we can probably keep the file quite short.

Any `CLAUDE.md` file should probably be tested and edited only when we update the repo (like when we added `fix-repo`) or when there's a clear A/B test or thing that Claude is visibly struggling with. Just packing a lot of explanation in there without testing will probably hurt more than help!

tarekziade:
@Rocketknight1 good points

I took this section, https://code.claude.com/docs/en/best-practices#write-an-effective-claude-md and ask Claude to analyze our AGENTS.md. It found it's very good but 2/3x too long. 

Then I asked it to tweak it to what it did not know.

```
⏺ Cut from 114 lines to 40. Here's what was removed and why:

  ┌─────────────────────────┬─────────────────────────────────────────────────────────────────┐
  │         Removed         │                               Why                               │
  ├─────────────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Environment Setup       │ Standard pip install -e ".[dev]" — Claude knows this            │
  ├─────────────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Tech Stack              │ pytest + ruff + PyTorch — inferrable from pyproject.toml        │
  ├─────────────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Repo Map                │ src/transformers/models etc. — obvious from directory structure │
  ├─────────────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Git Workflow            │ "Keep PRs small" — self-evident                                 │
  ├─────────────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Troubleshooting section │ Claude can read error output itself                             │
  ├─────────────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Common Errors section   │ Same — Claude doesn't need pre-canned error lookup              │
  ├─────────────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Output Example          │ Prescribing output format adds no value                         │
  ├─────────────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Docs MDX note           │ "keep formatting minimal" — vague and self-evident              │
  ├─────────────────────────┼─────────────────────────────────────────────────────────────────┤
  │ Makefile Truth          │ Merged the one non-obvious part (fix-repo scope) into Workflow  │
  └─────────────────────────┴─────────────────────────────────────────────────────────────────┘

  What was kept: the make commands, the copies/modular rules (genuinely non-obvious and critical), the CI test fetcher (can't guess this exists), slow test env var, the modular model key patterns, and the toctree requirement.
```

AGENTS.md
```diff
@@ -1,113 +1,30 @@
...
+- After touching copied code, modular files, docs, or docstrings: `make fix-repo` instead
 
-**Makefile Truth**
-- `make style` runs ruff format/check plus repo formatters.
-- `make check-repo` runs CI-style consistency checks.
-- `make fix-repo` auto-fixes copies, modular conversions, doc TOCs, doctest lists, and docstrings.
+## Copies and Modular Models (critical)
+- Files with `# Copied from ...`: edit the source file, then run `make fix-repo` to propagate. Never edit the copy directly.
+- Files with `modular_<name>.py`: never edit generated `modeling_*.py` directly. Edit the modular file, then run `make fix-repo` or `python utils/modular_model_converter.py <name>`.
+- Generated files are identified by `This file was automatically generated from` on the second line, or by `_pb2.py` suffix.
```
Rocketknight1: One suggestion, but feel free to ask Claude its opinions, or if it thinks this is unnecessarily verbose.

```suggestion
## Copies and Modular Models

We try to avoid direct inheritance between model-specific files. We have two mechanisms to manage the resulting code duplication:

1) The older method is to mark classes or functions with `# Copied from ...`. Copies are kept in sync by linters and `make fix-repo`. Do not edit a `# Copied from` block, as it will be reverted by `make fix-repo`. Ideally you should edit the code it's copying from and propagate the change, but you can break the `# Copied from` link if needed.
2) The newer method is to add a file named `modular_<name>.py` in the model directory. `modular` files **can** inherit from other models. `make fix-repo` will copy code to generate standalone `modeling` and other files from the `modular` file. When a `modular` file is present, that file should be edited directly. Changes made to the generated files will be overwritten by `make fix-repo`!
```

AGENTS.md
```diff
@@ -1,113 +1,32 @@
...
-**Troubleshooting**
-- `make style` failures: read output carefully—most are auto-fixed, some need manual fixes. Common: import ordering, docstring formatting.
-- `make check-repo` failures: check specific error message. If it mentions copies/modular/TOC/docstrings, run `make fix-repo`.
-- Test failures: check test output for assertion errors. Verify changes didn't break existing behavior.
-- Import errors after install: ensure you ran `pip install -e ".[dev]"` with `-e` flag for editable mode.
-- Slow tests timing out: use `RUN_SLOW=1` environment variable only when needed.
+## Adding a Model (modular way)
+- Create `src/transformers/models/<name>/modular_<name>.py`, inherit from a similar model, define only what changes.
+- `super().__init__(...)` unravels the parent body; use `del self.attr` after to drop unwanted attributes.
+- Run `python utils/modular_model_converter.py <name>` to generate files, then `make check-repo`.
```
Rocketknight1:
```suggestion
- Run `make fix-repo` to regenerate model files from modular files.
```

Rocketknight1 (approved):
I trimmed this one down a little more - I think we should start by keeping it as minimal as possible, and add more from there only when we see clear failure modes in the agents!

PR #44236, opened 2026-02-23
fix zero3 init config

SunMarc:
# What does this PR do?

Supersedes https://github.com/huggingface/transformers/pull/43847

When using zero3 + from_config, the model was incorrectly initialized as we were not gathering the params. Added a test also. 

cc @tohtana


SunMarc:
@bot /style

winglian (approved):

PR #44240, opened 2026-02-23
Fix return value - fixes #44238

tarekziade:
# What does this PR do?

Makes sure `find_bad_commit` always return the result `dict`


ydshieh (approved):
Thanks a lot 🙏

PR #44241, opened 2026-02-23
fix: add missing return type annotations to type-checking utilities in generic.py

yushiran:
## Summary

Adds missing `-> bool`, `-> int`, and `-> str | None` return type annotations to public utility functions in `utils/generic.py`, making them consistent with the newer `is_timm_config_dict` and `is_timm_local_checkpoint` functions already annotated in the same file.

**Functions updated:**
- `strtobool` → `-> int`
- `infer_framework_from_repr` → `-> str | None`
- `is_tensor` → `-> bool`
- `is_numpy_array` → `-> bool`
- `is_torch_tensor` → `-> bool`
- `is_torch_device` → `-> bool`
- `is_torch_dtype` → `-> bool`
- `is_mlx_array` → `-> bool`
- `is_flash_attention_requested` → `-> bool`

Zero behavior change — pure annotations only.

## Motivation

The file already has `from __future__ import annotations` and the newer helpers in the same file (`is_timm_config_dict`, `is_timm_local_checkpoint`, ~line 774) already carry return type hints. This PR brings the older foundational utilities in line with that convention. These functions are imported across hundreds of model files, so annotating them improves IDE support and static analysis throughout the entire codebase.

## Test plan

- [x] `ruff check` passes with no errors on the modified file
- [x] `ruff format` reports file unchanged (no style issues)
- [ ] CI tests pass

### DIFF
--- a/src/transformers/utils/generic.py
+++ b/src/transformers/utils/generic.py
@@ -60,7 +60,7 @@ if is_mlx_available():
 
 
 # vendored from distutils.util
-def strtobool(val):
+def strtobool(val) -> int:
     """Convert a string representation of truth to true (1) or false (0).
 
     True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values are 'n', 'no', 'f', 'false', 'off', and '0'.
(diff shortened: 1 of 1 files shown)


Rocketknight1 (approved):
LGTM

PR #44243, opened 2026-02-23
fix: add load balancing loss when output_router_logits=False

thakoreh:
## Summary

Fixes #44242

Load balancing loss was not being added when `output_router_logits=False` in Mixtral models.

## Changes

- Fixed loss calculation to include load balancing even when router logits are not output
- Added test case

## Testing

Added test in `test_modeling_mixtral.py`


thakoreh:
Closing - needs more investigation to fix CI issues properly.

PR #44250, opened 2026-02-24
fix regression report_to "all"

SunMarc:
# What does this PR do?

This PR fixes https://github.com/huggingface/transformers/pull/43806#discussion_r2834269455. We removed `self.report_to == "all"` functionality by mistake. Adding it back !


tomaarsen (approved):
Looks solid to me, thanks!

cc @qgallouedec as well

